PR #457, opened 2014-08-04
Ww3 feature/user settings

pstaabp:
A new view for changing the settings of a user (email, password and clear history).


pstaabp:
Relooked at this and fixed the issue about matching the passwords.  About your comment on Aug. 12, I decided with a submit password button to indicated when the user is done typing in old and new passwords. I could submit the request after the user leaves the confirm password input cell.  Do you think that would be clearer?

I also just looked into adding a password strength indicator.  The good ones download a large file to match the current password, which is a downside, but I can probably do it dynamically so that it's not too bad.

goehle:
- Show saved answers doesn't seem to work.  If I open up User Settings in WeBWorK 2 and uncheck show saved answers in WeBWorK 3 the state doesn't change in WeBWorK 2 (even though there is a success message)
-  Changing the email address doesn't work.  Typing and address and hitting enter or clicking on the page doesn't trigger a saved message.  
-  The page doesn't support the Equation Editor option. 
-  When I clicked Change Password I got a "You should not set a short password" message even though I hadn't entered anything.  I also got a random error about the original password not matching.  (This was not the first time I clicked Change Password.  This was on my 3rd or 4th run through.)  
-  I misspoke about blank password.  Its a problem because WeBWorK won't allow you to log in with a blank password so you shouldn't be able to set one.  You could get a password strength indicator if you want, but I don't consider it essential. 
-  The error message about the original password not matching is oddly persistent.  It stayed around after I had entered the correct password and the text field remained red until I reloaded the page.

pstaabp:
I think I have the errors cleaned up although I couldn't replicated your first one.  I've been playing with the backbone validator method of validating models and it didn't work very well for this situation so I abandoned it mostly. 

What do you mean by the "Equation Editor" option?

goehle:
Equation editor is a current option in develop in the user settings area that turns mathview on or off (depending on if mathview is enabled for the course).

pstaabp:
I see it now.  I didn't have the "Use MathView" global setting on.

pstaabp:
I've added the Use MathView (Equation Editor) as well.

goehle:
You haven't pushed your latest changes to your github repo.

pstaabp:
This is now updated.

goehle:
-  Reset History doesn't seem to work. 
-  I'm still having trouble with the checkbox settings.  Clicking the equation editor checkbox doesn't change it on the ww2 side.  The show saved answer checkbox is reversed for me (if its checked, then its Yes on the ww2 side, and if its not checked then its No) although changing the check box will affect the ww2 side.  Could this be an artifact of the "0"/"1"/"true"/"false" troubles that were around before?  I've seen some 0/1's mixed in with the true/falses when viewing the js console.

pstaabp:
* Fixed the reset history. 
* on the saveOldAnswers, I'm seeing what I think is correct behavior.  The check on ww3 is No on ww2.  Also, looking in the database, check is a 1 in the saveOldAnswers column and a uncheck is a 0. 
* Also, I getting the same behavior for useMathView.  

What are you seeing in the DB when you change the check boxes?

What 0/1 vs. true/false are you seeing in the js console?  What I've tried to do to stay consistent is for boolean variables, use them as booleans on the client/js side.  On the perl side dancer has some support for handling the discrepancy, but I have routines that will convert the perl true/false to a boolean in the JSON sense before sending to the client.  This is the convertObjectToHash method in most of the routes and I pass an array of the boolean fields.

goehle:
I figured out the issue.   The problem is that there is a "default" value for "showOldAnswers" and "useMathView" which ww3 is not respecting.  If the db value of showOldAnswers is NULL then the config variable `$pg{options}{showOldAnswers}` is used as a default.  If its not null then its value is used.  Similarly if the db value is null for useMathView then `$pg{options}{useMathView}` is used as the default.  However when ww3 is saving an "uncheck" it sets the db value to NULL and not 0, causing the default value to be used in ww2.

pstaabp:
I was taking NULL from the DB to be false.  Now if it is null it uses the default settings as you stated above.

pstaabp:
Yep.  Jumped the gun again without testing.  Sorry.

goehle:
Now setting the values to checked sets them to 0 and unchecking them sets them to NULL. 

What should happen is that the values will start out as null, which causes the checkbox to be checked or unchecked as appropriate.  Then if the checkbox is changed the value gets set to 0 or 1, as appropriate.  It never goes back to null after that.

pstaabp:
What should the values be?  I guess I'm confused.  What I'm seeing seems to be consistent with ww2.   Is your version of this branch switching back to NULL?

goehle:
Yes.  When the boxes are checked the value in the db is 0 and when the boxes are unchecked the value in the db is NULL.  (I just double checked and its not left over js in the cache or anything like that.)

pstaabp:
So this is not what I'm seeing.   I'm getting 0 and 1 as expected and as consistent with ww2.  I'm going to try to test it on a different setup to try to figure this out.

pstaabp:
What versions of JSON   JSON::PP or JSON::XS  are on your system?  Dancer uses JSON to convert to booleans and there appear to be some incompatibilities.

pstaabp:
I changed the way that this handles the booleans.  I'm thinking this will work.  If so, I'm going to make a utility subroutine to handle this more generally.

goehle:
The versions I have installed are: 

JSON : 2.61
JSON::PP: 2.27203
JSON::XS: N/A


Your changes have fixed things on my end so I'm putting this branch to bed.

PR #507, opened 2014-10-19
The nav menu now collapses when the window is resized to a small enough ...

aorinevo:
...width and a toggle button appears to allow access to the nav menu content.


drgrice1:
This has been here long enough, and clearly is not going to be merged.  It will still be listed in the closed pull requests for historical purposes.

PR #508, opened 2014-10-22
Auto hide side menu when resizing window below 768px and auto show when ...

aorinevo:
...resizing menu to beyond 768px.  Respects toggle when resizing.  That is if the menu is toggled off by clicking the toggle button and then resizing beyond 768px, the menu will not auto show.  Similarly for when the toggle button has been clicked to show the menu prior to resizing below 768px.  The threshold can be set in the math4.js in the variable ANSM.threshold. Also, margins were fixed. The masthead would overflow outside of the contair-fluid div when resizing below 768px.  The content has class span12 the menu is toggled off.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -122,8 +122,6 @@ ul.nav li a:hover {
 #breadcrumb-navigation {
     float:left;
     display:inline-block;
-    width:90%;
-    width:calc(100% - 35px);
 }
 
 /* Main Content */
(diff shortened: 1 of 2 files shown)


pstaabp:
This looks interesting, however, it doesn't merge right now.  I think you'll need to repull openwebwork/ww3 into this branch and resolve any conflicts. 

Also, it looks like you might be using the bootstrap2 classes to do some of this.  We've switched to bootstrap 3 moving forward and instead of span2 there are classes like col-md-2.

goehle:
ported over to another pull.

PR #514, opened 2014-10-30
Beta version of a Gradebook.

aorinevo:
Gradebook view renders student login names vs. assignment scores. Columns are all sortable asc/desc.

Some more things to do:
- Link the column heading name to the student progress view containing statistics for all students for the set corresponding to the column heading name.
- Link the row label name (login name) to the student progress view containing statistics for all sets for the student with login name corresponding to the row label name.
- Column headings for sets should be sorted, at least alphabetically.  Now it just returns some random order.
- Add a column for totals (in percent?)

Questions:
- Should we make certain CRUD functions available on this page, like delete assignment or delete student?
  - for example highlighting a column and pressing a delete button removes the set for all users, or have an icon in the column heading that when pressed, deletes the set for all users.  More precisely, for columns, we can add an icon that gives a dropdown menu of column specific functions like sorting and including some CRUD functions.  For example, consider an instructor viewing the GradeBook that decides they want to delete a set.  They'll click an icon next to that set and select delete from corresponding dropdown menu.  This would involve less clicks than navigating to the problem set manager, and clicking the delete button.
  - Similarly for deleting students (GradeBookRows). Perhaps something like this:
    rowcrud
  - Can we extend the CollectionTableView to include a footer for say, set averages? But this is column specific, so maybe include it in the dropdown menu suggested above.

### DIFF
--- a/webwork3/bin/app.pl
+++ b/webwork3/bin/app.pl
@@ -14,6 +14,7 @@ use WeBWorK::Authen;
 use Routes::Authentication qw/buildSession setCourseEnvironment setCookie/; 
 use Routes::Course;
 use Routes::Library;
+use Routes::GradeBook;
 use Routes::ProblemSets;
 use Routes::User;
 use Routes::Settings;
--- a/webwork3/lib/Routes/GradeBook.pm
+++ b/webwork3/lib/Routes/GradeBook.pm
@@ -1,0 +1,64 @@
+### ProblemSet routes
+##
+#  These are the routes for related GradeBook functions in the RESTful webservice
+#
+##
+
+package Routes::GradeBook;
+
+use strict;
+use warnings;
+use Dancer ':syntax';
+use Utils::Convert qw/convertObjectToHash convertArrayOfObjectsToHash convertBooleans/;
+use Utils::ProblemSets qw/reorderProblems addGlobalProblems addUserSet addUserProblems deleteProblems createNewUserProblem/;
+use WeBWorK::Utils qw/parseDateTime decodeAnswers/;
+use Array::Utils qw(array_minus); 
+use Routes::Authentication qw/checkPermissions setCourseEnvironment/;
+use Utils::CourseUtils qw/getCourseSettings/;
+use Dancer::Plugin::Database;
+use Dancer::Plugin::Ajax;
+use List::Util qw/first max/;
+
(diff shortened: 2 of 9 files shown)


aorinevo:
@pstaabp, has anyone looked at this yet?

pstaabp:
Aori,

Sorry, I haven’t had much time to work on webwork in the past few months.  I hope to free up some time soon and will take a look at it.  Nudge me again if you haven’t heard back in a couple of weeks. 

Peter

aorinevo:
No worries, when you get a chance.   

Can you also take a look at my gradebook plugin?  My goal is to integrate [AN_GradeBook](http://wordpress.org/plugins/an-gradebook) into WeBWorK3.

pstaabp:
I'm finally getting around to this.  Sorry for the delay.  It's fanstastic that you've been able to wade through the code to get a new tool integrated. 

I had one error in the GradeBook.pm file that may have been an old perl issue.  I now have it running.  I was testing on a course with about 30 Problem sets and only 2 of them show up.  I think because each set wasn't assigned to each user.    Here's a few general thoughts:

* This looks similar to the StudentProgress Pane currently.  (I don't recall when I put that in and if you had it when you wrote this). 
* If this is to be equivalent to the ww2 version of the Scoring Tools, we need to have some way to make a grid of all users and all problem sets as well as be able to download the file.  You can look at the Classlist manager which has a way to download the classlist file as an example. 
* My thoughts on a StudentProgress/GradeBook is to have flexibility is viewing and downloading grades, which this has. 
* I've made of lot of changes recently to the CollectionTableView and I just submitted that for a pull request.  That may break your code. 
* On the data transfer to the server: it appears to be fetching the data at startup.  My goal has been to load up the problem set and user data at startup (since nearly every tool needs one of these) and then only load other data as needed.  I can't seem to find the place where you make this call though. I would recommend to make only after this tool is loaded. 

* Also on the data fetch.  I would recommend a different data format.  I think it will be more robust if you have a JSON object like:
[{ user_id: XXXX, scores: [HW1: XX, HW2, XX, HW3, XX, ...]},
 {user_ID: YYYY, scores: [HW1: XX, HW2, XX, HW3, XX, ...]}
...
]

and then parse on the client side.   I'll take a look at the AN_Gradebook as well to see what else can/should be included.

pstaabp:
It's embarrassing that we left this so long.  I have pick up webwork3 a lot this summer and planned to get this pull request going.  Earlier, I switched to Dancer2, since Dancer1 is no longer supported, which is one of the reasons that this won't work as is. 

I have just pulled this into one of my branches (pstaabp/ww3-features/dancer2), which will be pulled into develop this fall (I hope).  I've been making a few changes to get it working with the current code, but this should make it in.

PR #521, opened 2014-12-11
Set defs search depth

Alex-Jordan:
Prior to this pull request, when importing `.def` files, only the `templates` folder (and no subfolder) are scanned for `.def` files. This adds `$options{setDefSearchDepth}=0;` and `$options{useOPLdefFiles}=0;` to `defaults.conf` which can be changed in `localOverrides.conf` or `course.conf`. `$options{setDefSearchDepth}` sets a search depth within subfolders (and symlinks) within `templates`. The idea is now you can store `.def` files in a subfolder of `templates` or in a central location on the server via a symlink.

Because this can take you into the OPL (where there are some `.def` files), you may want to exclude the OPL with `$options{useOPLdefFiles}`. (In fact every `.def` file I encountered in the OPL was broken with bad file paths and/or answer dates that came before the due date.)

I tested this with Hmwk Sets Editor2. When I went to test it with Hmwk Sets Editor, I had error warnings because of `.def` files that included a `description` line, and it appears Hmwk Sets Editor can't deal with that. Is this an oversight or are we not maintaining Hmwk Sets Editor anymore?

Test by restarting Apache, going to Hmwk Sets Editor2 in a course, and trying to import problem sets with various combinations of:
* different values for `$options{setDefSearchDepth}` and `$options{useOPLdefFiles}`. Setting the depth to 4 should make some OPL `.def` files appear. (But keep in mind that the OPL set.def files are probably not actually good if you then try to visit the set.) Also you might try putting some `.def` files in `templates` at depth 1 or 2.
* importing one or many sets.

And maybe do it with both Hmwk Sets Editor and Hmwk Sets Editor2.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -821,6 +821,14 @@ $options{useDateTimePicker} = 1;
 
 $options{enableConditionalRelease} = 0;
 
+# In the hmwk sets editor, how deep to search within templates for .def files;
+# 0 means only within templates
+$options{setDefSearchDepth} = 0;
+
+# In the hmwk sets editor, delve into the OPL while searching for .def files?
+$options{useOPLdefFiles} = 0;
+
+
 ##########################################################################################
 #### Default settings for the problem grader page
 ##########################################################################################
--- a/conf/localOverrides.conf.dist
+++ b/conf/localOverrides.conf.dist
@@ -371,4 +371,15 @@ $problemDefaults{showMeAnother} = -1;
 #$allowFacebooking = 1;
 #$facebookAppId = your_app_id_here;
 
+################################################################################
+# Searching for set.def files to import 
+################################################################################
+## Uncomment below so that when the homework sets editor searches for set def
+## files, it searches beyond templates; it can search deeper subfolders of 
+## templates, and optionally also descend into Library
+
+#$options{setDefSearchDepth}=4;
+#$options{useOPLdefFiles}=1;
+
+
 1; #final line of the file to reassure perl that it was read properly.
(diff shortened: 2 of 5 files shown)


goehle:
This now has conflicts which need to be resolved before merging.

Alex-Jordan:
OK, sorry for being so slow on this. So many other things going on all the
time. I'll prioritize this after Tuesday's webinar.

On Sat, Jan 31, 2015 at 1:52 PM, Geoff Goehle <notifications@github.com>
wrote:

> This now has conflicts which need to be resolved before merging.
>
> —
> <https://github.com/openwebwork/webwork2/pull/521#issuecomment-72338718>.
>



-- 
Alex Jordan
Mathematics Instructor
Portland Community College

goehle:
No worries, just letting you know.

Alex-Jordan:
Warning: I got exasperated trying to handle this properly with git. In the end I deleted my local branch, created a new one from the current develop branch, and pasted back in the changes from this pull request. Then I force pushed it. So if you have a local copy of the earlier branch, it will have nothing but merge conflicts with this one. Maybe delete whatever local branch you have, or force pull this one over it. 

I know it's bad practice, but I was going crazy trying to deal with the conflicts. It's what I get for letting this sit so long.

I added that line to the `get_set_defs_wanted` subroutine as suggested. I tested by setting `$options{setDefSearchDepth}` and `$options{useOPLdefFiles}` to various values and reloading the Hmwk Sets editor to see what was available. All seems well.

I didn't check the prune flag is working as intended, but it looks like you did that and found that it was.

goehle:
Sorry I let this one slip through the cracks.  I've tested it and it works with all the settings I can think of.  It also doesn't dive unnecessarily deep.  The default behavior reproduced how things work currently.  The only thing that people should be aware of is that if you enable the opl search and set the depth to be too high then you can pretty easily bog down your server.

PR #525, opened 2014-12-13
Labels for PG fields.  

goehle:
This pull request is for Alex, mostly. 

After thinking about it I put together a bit of a bandaid that adds screen reader only label for PG problem input fields.  In practice this doesn't really add anything that aria-label doesn't do, but it does help validate against automated accessiblity checkers.  I dont *think* it causes any other problems with rendering or the like.  Does this get most of the way to addressing your concerns?  It doesn't do selects, or checkboxes, and the label text isn't particularly friendly.


goehle:
Have you had a chance to look at this?  I want to merge release/2.10 in the next few days and need to know if this is part of it.  Thanks

PR #530, opened 2014-12-28
Icons in LB1 to signal problems with math objects and static problems

jwj61:
There are new tags in OPL problems to signal if the problem uses math objects and if it is static.  This change puts the information in the OPL database.  

If a problem uses math objects, an icon appears next to the rerandomize icon to signal that (in Library Browser 1).  The icon has hover text to explain what it is, and clicking on it opens a help pop-up to explain the significance of math objects to an instructor browsing problems.

If a problem is tagged as static, the rerandomize icon is simply not shown.

If a problem is not part of the OPL, (e.g., browsing problems in the templates directory), the problem is checked directly for the tags and acts accordingly.  In all cases, the default action (i.e., if there are no special tags), is to not show the math object icon and to show the rerandomize icon.


goehle:
I tracked that bug down and fixed it in #531.  I couldn't find any problems with the icons, though.

jwj61:
The tagging for math objects and static problems should be pretty thorough.  To see the MO icon, I start with Single variable Calc/Limits and Cont/Rules of limits - basic.  Restricting to the first 5, of them have the icon.

For static problems, there should not be a randomize icon at all.  If you can rerandomize a problem marked as static, then that is not showing.

Do you have the icon, it is htdocs/images/pibox.png .

goehle:
I needed the latest OPL to get it to work.  I have it going now.  I will say that the pi icon would look better without the black border, at least in math4.  None of the other icons have borders so it really stands out.

jwj61:
The thinking was that pi is a symbol of math, and the border represents it being an object.  Other suggestions people had for the icon were to use a tiny picture of Davide, or to have the 3-d pi spinning (to make it clear it was an object and not just a math symbol).  Compared to those, the current icon is probably looking pretty good now.

htdocs/helpFiles/UsesMathObjects.html
```diff
@@ -0,0 +1,31 @@
...
+################################################################################
+-->
+<head>
+<title>Uses Math Objects</title>
+</head>
+
+<body><br>
+<p>
+Some WeBWorK are programmed using what are called <i>Math Objects</i>.
+These problems are preferred by some people because they
```
goehle: Should be "Some WeBWorK problems"

goehle:
I kind of like a 3d spinning gif of Davide's head.  In any case, other questions comments. The prev/next thing really needs to be fixed.  The other stuff is a bit more grey. 

-  Lots of problems have the MathObjects icon but dont actually use math objects (/Library/WHFreeman/Rogawski_Calculus_Early_Transcendentals_Second_Edition/3_Differentiation/3.6_Trigonometric_Functions/3.6.37.pg)  I think it may be because they load the MathObjects macro but dont actually do anything with it.  Its a problem with tagging and not this pull, in any case.  
-  For some MathObject problems I also get the blue MathObject message and for others I dont.  (e.g. Library/Michigan/Chap8Sec5/Q23.pg)  It seems like if we have the icon we also dont need the blue messages. 
-  Clicking "Next Page" or "Previous Page" causes all problems to get MathObject icons. 
-  Clicking "Next Page" and then "Previous Page" will cause problems which didn't have the randomize icon to have the randomize icon. I think this is a problem associated to the `getDBextras` stuff.
-  Clicking "M" and then hiding the top problem causes the L and the Pi icon to switch places (if the next problem has a pi, that is).

jwj61:
Some of these can be addressed now.  I will get to the others in time.

* The missing word in the help file has been added (from earlier comment)
* I hope erroneous MO tags are not in _many_ files.  As you say, this should be considered a bug in that individual problem.  The bulk were done by a script which looked for "->cmp" (not loading mathobjects).  People at OPL IV were then converting individual problems to math objects, and the one pointed out was poorly done.
* The blue math object comments will be removed at some point in the future.  If I remove those comments before this pull becomes a part of a webwork release, people may be unhappy.

On to the interaction with Next/Previous/M.

jwj61:
The problems with next/previous buttons and the more-like-this M/L should now be fixed.

lib/WeBWorK/Utils/ListingDB.pm
```diff
@@ -135,11 +136,13 @@ sub setProblemTags {
...
 		my $sect = shift;
 		my $level = shift;
+		my $status = shift || 0;
 		my $tags = WeBWorK::Utils::Tags->new($path);
 		$tags->settag('DBsubject', $subj, 1);
 		$tags->settag('DBchapter', $chap, 1);
 		$tags->settag('DBsection', $sect, 1);
 		$tags->settag('Level', $level, 1);
+		$tags->settag('Status', $status, 0) if $status;
 		eval {
```
goehle: Out of curiosity what is the status tag?

jwj61:
Oops.  Harmless for now, but used in my next contribution which will be tools for OPL editors (so they can review contributions to the OPL through webwork).  The status is whether the problem under review has been accepted/rejected/marked for further consideration.

PR #532, opened 2015-01-05
Add status menu to tagwidget for OPL editors

jwj61:
This creates a new menu in the tagging widget which provides drop-down menus for tagging problems.  The new menu is for the Status of an OPL submission.  The overall process is as follows
   1. Person adds to Contrib, makes a pull request, it gets accepted 
   2. Managing editor copies files to the Pending directory in the OPL git repository
   3. OPL editors look at problems and use the status menu to accept/reject the problem (if accepted, also set taxonomy menus and level)
   4. Managing editor then moves accepted problems to the main OPL, deletes rejected ones from Pending and notes them in a file of problems which are not accepted (so they don't go back into Pending unless it is being forced)

By default, we don't want other users using this feature, so it presumes the course where the editor works has a link called Pending pointing to the OPL Pending directory on their server (with server having write permissions to those files).  Ideally, the server is configured with a "special button" for Pending.

In the end, this pull request just makes the menu and will save the results.  If the problem is not under templates/Pending, the new menu should not appear, otherwise it will.

### DIFF
--- a/lib/WeBWorK/Utils/ListingDB.pm
+++ b/lib/WeBWorK/Utils/ListingDB.pm
@@ -142,7 +142,7 @@ sub setProblemTags {
 		$tags->settag('DBchapter', $chap, 1);
 		$tags->settag('DBsection', $sect, 1);
 		$tags->settag('Level', $level, 1);
-		$tags->settag('Status', $status, 0) if $status;
+		$tags->settag('Status', $status, 1);
 		eval {
 			$tags->write();
 			1;
--- a/htdocs/js/legacy/tagwidget.js
+++ b/htdocs/js/legacy/tagwidget.js
@@ -57,6 +57,18 @@ function tag_widget(id, path) {
   for (var j=1; j<7; j++) {
     levels.append('<option value="'+j+'">'+j+'</option>');
   }
+  // Only show the status menu if we are looking at something in Pending
+  var shortpath = path.replace(/^.*templates\//,'');
+  if(/^Pending\//.test(shortpath)) {
+  
+    $el.append('<select id="'+id+'stat"></select>');
+    var stat = $('#'+id+'stat');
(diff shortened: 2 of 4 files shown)


lib/WeBWorK/Utils/ListingDB.pm
```diff
@@ -142,7 +142,7 @@ sub setProblemTags {
 		$tags->settag('DBchapter', $chap, 1);
 		$tags->settag('DBsection', $sect, 1);
 		$tags->settag('Level', $level, 1);
-		$tags->settag('Status', $status, 0) if $status;
+		$tags->settag('Status', $status, 1);
 		eval {
```
goehle: why is this setting status to be 1 for everything?

goehle: Sorry, I should have looked at what settag actually does.  The 1 is for the "force" variable.  This will force the "status" tag to be empty, even if it already had stuff in it.

goehle:
I tested that the menu will appear if you are in the Pending directory, but not otherwise and that it will save the status tag properly (including removing it if you set it back to "review"). 

I didn't realize this was holding up the whole review process for the OPL.  This pull doesn't seem like it would impact a non-editor (most people don't even have the tag widget around).  If you think its important we could probably rebase this to master.  Otherwise I'll just pull it.

jwj61:
On the force, the point is that it would not put a status of 0 if it currently had a non-zero status.  So, if someone set something to "Accept" and then want to go back to unrated, it would not do it without the force.

You are right, I don't think this will have any visible effect for most people  You need a pretty contrived situation to even see the status menu, so rebasing it to master should work.  I am not sure I know how to do that, but you could explain it to me tomorrow.

goehle:
We decided that this can go into develop for now.

PR #538, opened 2015-01-08
Set 0 Bug

goehle:
Post release bugfixing continues.  Fixed a problem where the set name 0 was interpreted as null by xmlrpc.  This allows you to add problems to sets named "0" via the library browser now.


mgage:
Checks out.

PR #542, opened 2015-01-11
SendMail Bug

goehle:
Refactored SendMail.  Previous setup used the "value" of the submit buttons to figure out the action, which broke when the value changed, e.g. in a different locale. Current setup uses the name of the button to determine action.


goehle:
To test - go to Send Mail and click  on all the buttons and make sure they do what they should.  To test broken behavior change locales to something not english.

mgage:
All of the buttons work.  There is a lot of code in this module that has not been localized but that is another story.  Possibly we'll just wait and do a good job of localizing the webwork3 version of send mail.

PR #544, opened 2015-01-11
Radio/Math4 Bug   (Develop)

goehle:
Develop version of #543


jwj61:
I tested this one and it works.  I would merge it, but am unsure if you that will cause trouble when the master one is merged, and ultimately master is merged with develop

goehle:
We generally pull at the same time, although in theory there isn't any trouble if we pull this one first.  They have the same commit numbers so there won't be conflicts.

PR #548, opened 2015-01-17
Apache Version Conf Variable

goehle:
This pull request does two relatively minor things
-  It adds a "Assign All Sets to Current User" button to the UserDetail page.  Check by clicking on the "Assigned Sets" link next to a user and checking to see that it does what it is supposed to.  

- It adds an optional (and commented out by default) `$server_apache_version` variable to site.conf.  When this is defined it will be used to determine the server version.  This is primarily for servers which do not have the version number in the banner for whatever reason.  Check by verifying that Authen, SendMail and Feedback still work, then set the server version string to something its not (pre 2.4 if you are post 2.4 and post 2.4 otherwise) and check that it breaks those pages.

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -59,6 +59,13 @@ $server_userID     = 'www-data';
 $server_groupID    = 'wwdata';
 
 
+#  Uncomment out the following line to set your apache version number manually.
+#  WeBWorK will automatically get the apache version directly from the server
+#  banner.  If you remove the version from the server banner you will have to
+#  set it directly here
+
+#$server_apache_version = '';
+
 ################################################################################
 # Paths to external programs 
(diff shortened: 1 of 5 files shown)


goehle:
I used to use `get_server_version` but it was depreciated.  I changed it to `get_server_banner` in commit 5347de777027.

goehle:
I also just tested it and `get_server_version` just returns the banner in any case.  So if you do `SetTokens Prod` then `get_server_version` wont return the version.  

I made the change about the configuration commenting.

mgage:
seems good enough.  We're still going to get complaints about the error message so we should remember to put an entry into "what could possibly go wrong" about errors that emanate from having the wrong Apache version number defined.

PR #549, opened 2015-01-17
Guest User Check Permission

goehle:
Adds a new permission `can_check_and_submit_answers`.  This is set to `ta` by default and controls who can see both the check and submit button.  If a user has the permission to see each individually, but not both, it shows the submit button.  This allows `check_answers_after_open_date_with_attempts` to be set to `guest`.  This way guest users can check answers on open sets, students can submit them, but not check them, and ta and above can do both. Check by assigning a set and logging in as guest.  With the default permissions you should be able to check answers.  Check that students cannot check answers until they run out of attempts or after the due date.  (Note: On gateways the check and grade buttons are never shown together.) 

This also adds the configuration variable `$options{problemGraderScoreDelta} = 5;` which controls the delta for the score selection drop down on the problem grader page.  Check by visiting the problem grader page and seeing if the drop down responds correctly when the config variable is changed.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -693,10 +693,11 @@ $authen{proctor_module} = "WeBWorK::Authen::Proctor";
         # controls if old answers can be shown
 	can_show_old_answers                            => "student",
 	check_answers_before_open_date                  => "ta",
-	check_answers_after_open_date_with_attempts     => "ta",
+	check_answers_after_open_date_with_attempts     => "guest",
 	check_answers_after_open_date_without_attempts  => "guest",
 	check_answers_after_due_date                    => "guest",
 	check_answers_after_answer_date                 => "guest",
+        can_check_and_submit_answers                    => "ta", 
 	create_new_set_version_when_acting_as_student   => undef,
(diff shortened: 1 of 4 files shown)


mgage:
A student can submit problems (between open and due dates) until they have exhausted their attempts.
Instructors see both check and submit buttons (between open and due dates) and if they are acting as someone the submit button says "submit for xxxx " as a warning.  Guest can only check answers both before and after the due date.  The number of attempts by a guest is not recorded since it is only the check answer button.  Pushing the check answer button does not increase the number of attempts. 

Anyone have ideas for writing a unit test for this behavior?  This is a behavior that could really benefit from automatic testing. 

The increment by 5 in the grading section works.  It will not be obvious to most people how to change this. There is no commented out entry in the localOverrides.   This is not a big deal since I think increments of 5 are a good choice and few will want to change it.

goehle:
I don't know about this behavior.  Unit testing the final interface seems more complicated.  The lower hanging fruit for automated testing would be things like the DB interface, or the webworkWebservice interface.  Stuff that has a collection of interface functions which you can call and test the results.

PR #550, opened 2015-01-17
DateTime refactor and Friends

goehle:
This re-factors some of the webwork2 js in particular. 

-  I moved some of the more essential JS files from the legacy folder to proper application folders.  Mostly this is because I could never remember where setmake.js was, but its also because these js files are actively being used and modified, unlike some of the other ones in legacy.  Test by visiting the pages showing js changes and seeing if the js loads. 

-  I removed the Ipad keys feature.  I saw a bunch of people using ipads to access webwork at the JMM mini course and keys was more trouble than good.  It kept floating over things people cared about and they all used their regular keyboards anyway.  Eventually we should make proper keyboards for webwork (or steal them from someone else).  Test by visiting the problem page on an ipad (or emulated ipad) and checking that the keys functionality is gone. 

-  I re-factored the DateTimePicker js.  The perl generated js was getting too long and complicated to be included inline everywhere.  I separated it out into its own js file.  Now the perl script just prints out some initialization js.  I also changed the js so that it will check for time conflicts and add the changed class if the date is edited manually, in addition to using the datepicker.  Check by visiting the pages that use datepicker (user detail, homework sets editor, problem set detail) and seeing that it works.  Check the date validation functionality by changing dates in an invalid way and checking that they are fixed.


mgage:
DatePicker comments.  Mostly it seems to work very well.  It if you increase the due date then the answer  date increases automatically (and silently).  This seems reasonable to me.

Once slightly unusual behavior appears on the user detail page.   -- click on a "due" blank, then click on a "answer" blank just below it.  The blanks will be filled with dates from around 1970.  

If you start by clicking on one of the pop up datepickers it appears you are initialized with the current date. If you click "done" the 1970 date is entered in the date box, but the popup remains displayed. If you click done again all of the boxes in that group are set to the current date. This doesn't happen 
in the problem set detail page, but there we have a "none selected" warning -- which I suspect replaces the 1970 date whenever that appears.  This is much less confusing. It becomes clear that even though you have brought up the date time picker until you click somewhere on it you have not changed the date in the date box. 

Nothing fatal but a little confusing (and I'm not sure I have described the complete pattern). A reasonable expectation might be that clicking on the popup box (or a blank date blank) would initialize the popup box to the date set for the entire class.

goehle:
I looks like the problem is caused by the `update` function, which sets the due date and answer date to be equal to the open date (plus modifications) if they are empty.  The appropriate approach is probably just to check to see that the open date exists before setting the due date and answer date equal to it.  (Which is what I did.)

mgage:
Any chance of making updates on the user detail page behave identically to the updates in problem set detail page just for uniformity? (I'm referring to the "none specified" message.

goehle:
Try it now.

mgage:
OK.  Looks good.  I haven't checked the iPad yet.  I'm also just leaving the patch on to see if some missing javaScript shows up in the next day or so.  I've also pulled the removeTZ patch. That works so far.  When working with old sets you may or may not see the timezone depending on whether you have recently changed the date -- but that seems acceptable to me. It will go away soon enough.

goehle:
Hmmm.  You should notice an immediate change, even with old sets. The TZ is
a display thing.  Times are stored as epoch seconds.

mgage:
I've tested the "key" overlay on ipad.  It does not appear anymore.  
When using the datetime picker in the set detail pages advancing the due date past the answer date will automatically adjust the answer date.  When using the datetime picker to change the date in the homework editor itself the due date can be advanced past the answer date without any readjustment. (Submitting this change in dates gives an error message that the answer date must be after the due date.)  It would be better to have the same behavior in both places. the best thing IMHO  is to update the code for the datetime picker in the homework editor itself.

mgage:
On the ipad in file manager:  the pop up menu which lists the contents of the templates directory reads "0 items".  A cosmetic bug.  I'll put this in bugzilla now.  No need to hold up pulling this request.

mgage:
Except for the failure to adjust dates when changes are made in the homework editor itself (where only dates can be changed, not the problem list) this seems ready to go.  I'm willing to enter that last glitch into bugzilla and pull this request for now if that's what you'd like.  Or you can fix that first and then I'll pull this request -- which ever is easier.

goehle:
-  I can't seem to find the popup you are talking about.  I'm using chrome's ipad emulator, so its possible that its not emulating right, though. 

-  I also can't seem to reproduce the behavior you are talking about with the Homework Sets Editor.  Both the Homework Sets Editor and the Problem Set Detail page use the same JS so there must be something else.  Are there any JS errors on the Homework Sets Editor Page?

mgage:
The pop up is on the FileManager page.  There are two -- one says that we are looking at the template directory and the other lists the files that are in the directory (It's a scrolling list for the desktop view.)

mgage:
I can't get the datepicker behavior to repeat either now.  It's a little weird because I duplicated it several times before I wrote the comment.  I'll pull this request. The "0 items" bug might actually be in the jquery implementation for the ipad.

PR #551, opened 2015-01-18
Remove TZ from Date Entry

goehle:
This pull request should be dealt with *after* #550.  

This pull request removes the TZ's from value strings on the date input fields on the User Detail, Homework Sets Editor, and Problem Set Detail pages.  Now instead of entering "12/20/13 at 8:00PM EST" you just enter "12/20/13 at 8:00PM" and the time zone will automatically be whatever you have your course set to.  (Actually, this happened before.  All this pull request does is prevent the tz from being printed)  All of the student pages still display the time zone, which will automatically be daylight savings or not depending on the date.  This fixes the issue of changing a date with (either with the date picker or manually) and having the saved time be one hour different because your date didn't match your timezone. 

To test, enter dates on each of the pages listed above and check that the behavior is what you expect.  Also check for other places where homework dates can be set.


goehle:
I tweaked this a little bit
-  I changed the order that the reduced credit date is listed on the ProblemSetDetail page.  It comes between open date and due date now. 
-  I removed all of the TZ's from the ProblemSetList2 page and used the space savings to add the reduced credit date.  
  -  Check that you can set and enable reduced credit from the Homework Set Editor and that date checking works (including the date picker). 
  -  Check that turning off reduced scoring removes the correct columns from the Homework Sets Editor page.

mgage:
This will have a conflict with changes made in UserDetail.pm for formatting the date and time in the previous pull request. 

++<<<<<<< HEAD
 +                                                    -value => $userValue ? $self->formatDateTime($userValue) : "None Specified",
++=======
+                                                     -value => $userValue ? $self->formatDateTime($userValue,'','%m/%d/%Y at %I:%M%P') : "",
++>>>>>>> 81588f6a803555db98b4f5d98dde52ed3daea64a

mgage:
on the homework set detail page I am getting an "undefined" entry when I use the datetime picker.
it corresponds to the timezone not being defined.  I get a 0 in the open date entry and undefined in the due date entry and answer date entry.  I am using chrome on a mac

mgage:
I see the same behavior on the ipad.

mgage:
One further feature suggestion.  Print the current course timezone somewhere on the problem set detail page and on the homework date editor page --- just so that the instructor notices if they haven't set the course time zone correctly for their class.   The timezone does appear on the time entries on the home page which students see.

mgage:
You could put the timezone in parentheses after the open date for example.

goehle:
I'll try to track down the undefined entry bug you have above

I'm a little wary about adding adding the timezone on the page because most
of the time its useless information.  Even if it isn't the instructor isn't
going to look for the timezone on those pages or necessarily recognize the
implication.  Removing the time zone to save space was part of why I did
this.

mgage:
I'm not seeing the "undefined" bug anymore.

PR #555, opened 2015-01-21
Client Side Cacheing

goehle:
Right now, even in MathJax mode, accessing the problem page involves around 30 http requests.  Most of the requests are for static files.  This pull tries to improve webwork performance by setting up the server to provide expire information for the browser cache.  This tells the browser not to even request the files until after the expires date, which should reduce the number of requests significantly.  In order for this to work you have to enable mod_expires on your server, in addition to pulling this pull request.  On Debian/Ubuntu this can be done with `a2enmod expires`.   You will also need to copy over the expires information from the apache config dist files to your webwork apache config file.  This pull also reorders css and js so that all css includes comes before js.  This improves page loading by allowing all css to load in parallell.

This is kind of an experimental pull request.  I don't know what the right intervals on the ExpiresByType directives are and its possible this will have unintended side effects.


mgage:
This needs fixes in SetMaker.pm and Problem.pm to avoid conflicts.

mgage:
using https://hosted2.webwork.rochester.edu/webwork2/2014_07_UR_demo/Calculus_I/2/
Optimizing the order of styles and scripts:

profile before:  The following external CSS files were included after an external JavaScript file in the document head. To ensure CSS files are downloaded in parallel, always include external CSS before external JavaScript.
jquery-ui-1.8.18.custom.css
achievements.css
mathview.css
knowlstyle.css
1 inline script block was found in the head between an external CSS file and another resource. To allow parallel downloading, move the inline script before the external CSS file, or after the next resource.

profile after pull request:  The following external CSS files were included after an external JavaScript file in the document head. To ensure CSS files are downloaded in parallel, always include external CSS before external JavaScript.
mathview.css
2 inline script blocks were found in the head between an external CSS file and another resource. To allow parallel downloading, move the inline script before the external CSS file, or after the next resource.

It looks like the mathview.css link still needs to be relocated.

mgage:
Caching -- after pull request:

The following resources are missing a cache expiration. Resources that do not specify an expiration may not be cached by browsers:
bootstrap.css
bootstrap-responsive.css
font-awesome.min.css
math4.css
jquery.js
jquery-ui-1.8.18.custom.css
jquery-ui-1.9.0.js
bootstrap.js
achievements.css
dragmath.js
addOnLoadEvent.js
java_init.js
color.js
Base64.js
mathview.css
mv_locale_us.js
mathview.js
problem.js
math4.js
underscore.js
knowlstyle.css
knowl.js
wz_tooltip.js
webwork_transparent_logo_small.png
maa_logo_small.png
question_mark.png
tagwidget.js
gage-4648-setCalculus_Iprob2image1.png
TeX-MML-AM_HTMLorMML-full.js
MenuArrow-15.png
/webwork2/instructorXMLHandler
fontdata.js
The following resources are explicitly non-cacheable. Consider making them cacheable if possible:
hosted2.webwork.rochester.edu/
MathJax.js
fontawesome-webfont.woff

It looks like these guys are still missing cache expiration dates    This  is not much different than the version of develop before I included the pull request.

mgage:
There are other savings suggested:  combining css files and javaScript files and also using gzip compression but I think those can wait for another pull request. Probably those improvements can come as we continue to reorganize the directory of javaScript files

mgage:
Do you want to fix the mathview.css and/or the expiration dates before I pull this request? or would you rather this be pulled and you can add the remaining features later.  So far I don't see down sides to this pull request.

goehle:
We can't easily combine the css and javascript files because a lot of those are only loaded if certain settings are turned on.  We would need to refactor things to use something like requirejs which would assemble and deliver all the javascript for us.  Similarly there are complains about unused css lines, but that is because we use the same css file for the whole site. 

I can fix the mathview css thing pretty easily, but the expiration dates work for me so I'm not sure what the issue is there.

mgage:
The timeline profiles for the page I looked at above is not dramatically affected by the pull request.

goehle:
That is my experience also.  Browsers are so aggressive about cacheing on their own that it seems like explicit expires information may not have a big impact

goehle:
screenshot from 2015-05-12 10 41 55


Ideally this is what you should get with all the expires information in place.

mgage:
Using requirejs is what I had in mind, but there is no hurry and in fact it might only be necessary to implement it for webwork3

mgage:
I think we've gotten the low hanging fruit on this speedup attempt.  I'm ready to pull this request.  Other efficiency improvements can be in a future pull request.  Is that ok with you?

pstaabp:
In WW3, there is a huge performance hit at the beginning as the page loads and then afterwards, usually there are just small requests.  I have done any performance on ww3 yet, but requirejs is supposed to help with all of the js files (non libraries).  I just haven't looked into that yet.

PR #559, opened 2015-02-22
Scoring Multiple Select Bug

goehle:
In very recent versions of CGI you have to use "scrolling_record_list" and not "popup_menu" if you want a multiple select.  

Changed Scoring.pm to use "scrolling_record_list" and not "popup_menu".


pstaabp:
This worked fine for me for my setup.

goehle:
Great.  I also disabled the git notifier by default.  I've had, and heard of, too many problems with it.  It can work but shouldn't be a default thing.  

Could you (or Mike) actually deal with pulling it?

goehle:
Added a bandaid fix for Bug 3331. This bug causes lots of error popups to appear on pages which have applets in hints.  This comments out the error messages, but a better fix should be put in develop.

goehle:
Mike already fixed the trig problem issues that part of this pull  request was to fix, so I reverted those changes.  The remaining change is small and needs to be done so I'm pulling.

PR #562, opened 2015-03-14
Develop wwsh

mgage:
Modify wwsh so that it takes courseID and the path to a script file.

It no longer uses perl -d to give an interactive shell.  This stopped working (at least for me) around perl 5.16.

To test.  Try to add an administrator from the command line to an existing course
wwsh maa101 addadmin

Then update the password

wwsh maa101 putadmin

Remember to remove the admin user from the maa101 course since the password is generic.  This is a major use case for wwsh.  To add an emergency administrator when you have been locked out of a course.

### DIFF
--- a/bin/putadmin
+++ b/bin/putadmin
@@ -1,0 +1,10 @@
+
+$db->putUser($db->newUser(user_id=>"admin", first_name=>"Prof", last_name=>"Admin",email_address=>"", student_id=>"admin", status=>"C",section=>"", recitation=>"",comment=>"administrator"));
+
+$db->putPassword($db->newPassword(user_id=>"admin", password=>crypt("admin", "dc")));
+
+$db->putPermissionLevel($db->newPermissionLevel(user_id=>"admin", permission=>"10"));
+
+
+print "data changed for user admin\n" ;
+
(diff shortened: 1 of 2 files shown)


goehle:
I've tested this a bit. 
-  If you try to run wwsh with no courseID or script file you do get a usage message, but you also get a couple of "uninitialized value" warnings.  It would be nice if it was just the usage message. 
-  The interactive shell functionality was really powerful.  Is it worth it to try and get it back?  (I like having script support in either case because it makes simple tasks easier.)  I've messed with it a bunch and it seems like something has changed with Safe which is causing the trouble.  When the configuration files are "revaled" they can't see the "include" subroutine which should have been revaled in the compartment earlier.  There is something squirrely going on deep down.

mgage:
I'll fix the uninitialized value warnings.  I played briefly with trying to get the interactive shell back but wasn't able to figure out what was wrong either.  There was an earlier version of wwsh which did not use the perl debug mode -- that is another possibility if we can't figure out how to make the debug mode work.   My main goal for the moment is to able to run scripts -- such as the ones that allow you to insert yourself in a course from the command line.

goehle:
Using a non perl -d solution could be good.  Something like
http://search.cpan.org/~ether/Devel-REPL-1.003026/lib/Devel/REPL.pm

PR #563, opened 2015-03-14
Change "alert" warnings to "console.log" warnings. --develop

mgage:
Convert some JS alerts to console.log calls. This allows applets to be placed in HINTS sections without causing spurious alerts because the applet appears to be missing.  The alerts are still there but are not observed unless one is using developer tools to view the console log.

to test:   View the CSU trig problems as a guest or as a student.  For example at:  
https://devel3.webwork.rochester.edu/webwork2/gage_course/CSU_trig_applets_in_hints

Make sure to refresh the browser cache first.  Check that the hints work as expected.  

Using the developer tools you should be able to see error messages such as "Oops, applet is not ready. X tries left" as well as other messages indicating progress in handling the applet.

It is possible that the console.log messages will not play well with MSIE when the development window
is not open, however there were already a dozen console.log messages in this script and this only
adds two more.


goehle:
Closing this in favor of pulling the master version to both.

PR #569, opened 2015-04-15
Past Answer UTF8 Rendering

goehle:
Right now past answers improperly encodes and renders UTF8 characters.  This fixes that and also has Past Answer use the same encoding scheme as PG.  

Test by pulling PG pull request #195 and entering an answer like ñeumé into an answer blank.  Then view past answers.  Pre patch the accented characters will be mangled.  After applying both the PG patch and this one the accented characters should be fixed.  

Bonus:  I simplified the achievements default setup by removing the "no item" side of things.  Now items should be controlled by the course configuration, not what you import.  Test by creating a new course based off model courses, importing the only available achievement file, and testing you can turn items on and off using the course configuration menu.


mgage:
This fixes the utf8 problems with past answers.

I can turn the achievement features on and off from the course configuration page.

Issue #567, opened 2015-04-06
Library browser fails to update


dlglin:
When I upgraded my dev machine from WW 2.9 to 2.10, the Library Browser stopped updating when I choose a Subject.  It no longer auto-populates the Chapters as soon as a subject is selected.  I've attached a screenshot of the js error console.
![screen shot 2015-04-06 at 3 05 34 pm](https://cloud.githubusercontent.com/assets/1794586/7012489/cf608226-dc6e-11e4-9ec7-243e790612ad.png)

jwj61:
Maybe any relevant entry from the apache error log would shed more light on it.

goehle:
The JS is dying because WeBWorK or Apache or something is returning an error message that doesn't parse as json.  Either looking at the apache error log, or the actual responses sent back by the server.  (You can find the responses by looking in the network tab of most browsers developer tools.)

dlglin:
Here's what I get in the apache error log:

[/webwork2/instructorXMLHandler] Can't locate object method "ssl_opts" via package "SOAP::Transport::HTTP::Client" at /usr/share/perl5/vendor_perl/SOAP/Lite.pm line 483, <DATA> line 522.\n * in SOAP::Transport::__ANON__ called at line 149 of /opt/webwork/webwork2/lib/WebworkClient.pm\n * in WebworkClient::xmlrpcCall called at line 320 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/instructorXMLHandler.pm\n * in WeBWorK::ContentGenerator::instructorXMLHandler::pre_header_initialize called at line 183 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::go called at line 380 of /opt/webwork/webwork2/lib/WeBWorK.pm,

Running SOAP::Lite 0.710.10

mgage:
Hi Danny,

I ran into something like this awhile ago.

Here is a stackoverflow reference to the problem:  http://stackoverflow.com/questions/14839098/perl-cannot-access-web-service-with-ssl <http://stackoverflow.com/questions/14839098/perl-cannot-access-web-service-with-ssl>


What I have added to my webwork.apache2.4-config file is the last line in this group:
-------------------
my $webwork_dir = "/opt/webwork/webwork2";
# Cache this value for use by other scripts not necessarily in the Apache2 hierarchy
# Different scripts use different locations :-)

[...]

PR #570, opened 2015-04-17
Added LWP::Protocol::https to check_modules.pl

goehle:
In response to issue #567


mgage:
This module is already present on many systems.  This should do no harm.

Issue #565, opened 2015-03-25
MathView labels


Alex-Jordan:
Some accessibility issues have been raised by staff here relating to the MathView equation editor. (In prior testing, the editor had usually been disabled.)

1. The button for the editor itself just reads 'link' and the student doesn't know what is the link for. We should give it an informative label.
2. Within the equation editor, when MathJax is outputting MathML, the template buttons (like "a+b", "a-b") just read as "link". This is not an issue when MathJax is outputting HTML-CSS. 

You can see all of this at https://www.youtube.com/watch?v=v3WaNlWy5DM&feature=youtu.be. The video starts using HTML-CSS MathJax output, and at 2:17 switches to MathML.

goehle:
I added pull request #571 which should address these issues.  It adds screen reader only texts to the problem links.

PR #571, opened 2015-04-17
Mathview Accessibility

goehle:
Added screen reader text to mathview.  In particular I added text to the mathview button and to each of the function buttons within mathview.  This is to address #565

Also, Tweaked a couple of places where there were extra `<br>`'s.


mgage:
Should you rebase this pull request so that it doesn't contain my commits from March5?  

I'm not sure what happens if I merge it as is -- does it show these commits being added twice to the tree? or does it magically skip over the commits that are already there?  I'm sure a merge will work properly
but I"m not sure what the resulting tree will look like.

goehle:
I merged develop back into this branch so there are only new changes.

mgage:
Installed this a played around with the MathView popover.  It seemed to work well.  I tried using it with voiceover.  Voiceover doesn't yet interact well with the popover but it didn't before this change either.

PR #572, opened 2015-04-17
Webworkclient2 mods

mgage:
Updates to the webworkClient -- xmlrpc interface.


mgage:
This needs to be reworked.  Can anyone advice me as to the best way to delete a commit from the middle of this branch?  Specifically the modification to Utils.pm needs to be removed. It is incorrect.

mgage:
Some tests you can do:

1. First make sure that you have set your configuration file as described in the configuration
section near the top of the renderProblem.pl file. 

From the command line
renderProblem.pl input.txt 

should render the pg problem in input.txt 

renderProblem.pl bad_input.txt

should report warning messages.


checkProblem.pl and check_problems_in_dir.sh  can be used to check all of the pg problems in a directory to see if they render without creating warnings.  These scripts could be the basis of a task that automatically unit tests pg problems for reliability and accuracy.

goehle:
I don't know about removing the commit, but `git revert` can be used to
make a new commit which undoes the changes in a previous commit.

dpvc:
> This needs to be reworked. Can anyone advice me as to the best way to delete a commit from the middle of this branch? Specifically the modification to Utils.pm needs to be removed. It is incorrect.
> 

You can use "git rebase" to remove a commit.  For your case, 

	git rebase --onto c8b0f59 a85a7bf 93eb92c

should do it.  Here, c8b0f59 is the commit just before the one you want to remove, a85a7bf is the one to be removed, and 93eb92c is the end of the branch (as of the time I wrote this).  You could use "webworkclient2_mods" instead to indicate that.

Once you do this, you will have an unnamed branch plus your original webworkclient2_mods branch (if you use the name in the command,, it might be that the name moves with the new branch, I'm not sure).  So do

	git branch -D webworkclient2_mods
	git checkout -b webworkclient2_mods
	git push --force origin webworkclient2_mods

to remove the old branch and name the new branch "webworclient2_mods", then force the changes to be pushed to GitHub.

The webworkclient2_mods branch is pretty old (it goes back to August 13 last year).  You might want to rebase the whole thing (since you are already doing that for most of the changes).  Once you have done the rebase above, you could use something like

	git rebase --onto develop b0c1b88 webworkclient2_mods

to rebase everything onto the current develop branch.  Be sure you have a current copy of develop first.

Hope that helps.

Davide

mgage:
Thanks Davide and Geoff. I'll try the rebase route -- I've wanted to get more proficient at that anyway.  The recent spate of pull requests is my attempt to get a lot of the remedial development things I've done over the last year out of my own machine and into the main stream.  There is a lot of old code, especially on the PG side of things, that needs to be paid some attention.

mgage:
Another test is to run clients/TestWW.html.  This is the nascent version of embedding webwork problems in ordinary web pages.   The biggest defect in the current iteration is the authentication mechanism which is rather hacked.  I think it would be worth reviewing the entire Authen infrastructure and planning a way to make it more cohesive. This would help ww3 as well.

cmhughes:
Regarding removing a commit: I've had success using

git rebase -i HEAD~<number of commits back in time>

It is described in Greg Hewgill's answer here:
http://stackoverflow.com/questions/1338728/delete-commits-from-a-branch-in-git

I like this approach as it brings up an editor containing commits from the
point you specified above; you can combine, remove and re-order commits by
manipulating the lines in the editor. Probably worth trying on a copy of
the branch before going live :)

mgage:
Replacing this with develop_wwclient3_mods

PR #573, opened 2015-04-19
Replace \1 by  in Utils.pm

mgage:
Replaces \1 by $1 in one grep expression

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -928,7 +928,7 @@ sub cryptPassword($) {
 sub dequote($) {
 	local $_ = shift;
 	my ($white, $leader); # common whitespace and common leading string
-	if (/^\s*(?:([^\w\s]+)(\s*).*\n)(?:\s*\1\2?.*\n)+$/) {
+	if (/^\s*(?:([^\w\s]+)(\s*).*\n)(?:\s*$1$2?.*\n)+$/) {    # changed \1\2 to $1$2  MEG
 		($white, $leader) = ($2, quotemeta($1));
 	} else {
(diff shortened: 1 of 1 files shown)


goehle:
What does this change do, OOC?

mgage:
This and the similar change in PG  converts from an old perl paradigm to a new one

s/(foo.)/\1bar/    would replace  foo3 by foo3bar in a string and it would replace foo7 by foo7bar.     This method is deprecated and creates lots of 
warning messages in the log (for new versions of perl)

The new way to write this is
s/(foo.)/$1bar/       the contents of $1 is replaced by what is captured in the first set of parentheses. 

(My example is overly simple -- normally you'd have a more complicated matching pattern.)

goehle:
I dont think this one should be changed either, since its also matching a back reference.

Now, I'm not sure this function is used anywhere but when I inserted some test code (shown below) it didn't properly dequote the @ unless the \1\2 was used in place of $1$2
```
	my $poem = <<POEM;
@     Now far ahead the Road has gone,
@        And I must follow, if I can, ``
@     Pursuing it with eager feet,
@        Until it joins some larger way...
@           --- Bilbo
POEM
die(dequote($poem));
```

dpvc:
I agree, these should remain `\1\2` or you will get substitution of the *previous* values of `$1` and `$2` into the match string (whatever they were from the last match, if there was one, or empty strings otherwise).

mgage:
You're right.  I'm withdrawing this pull request.  Only the \1 on the right side of a substitution
string should be changed to $1.   Those are the ones causing the warnings.  The back references in
matches and on the left side of substitution strings must remain as \1.    

Sorry about that -- I'd forgotten the difference.

-- Mike

PR #574, opened 2015-04-19
Remove warning that cookie userID and sessionID do not match. 

mgage:
Resolve in favor of the sessionID and then kill the current cookie.

This safety precaution has caused a good deal of trouble, particularly when a user logs in to webwork from moodle
but also from webwork using the same browser.    It's not clear to me that it provides significant safety either.


goehle:
I've been meaning to deal with this one myself.

PR #576, opened 2015-05-05
Update documentation to show that "session_cookies" not "key" is the def...

mgage:
...ault value of $session_managment_via

This is a minimal fix for the moment.

The documentation about how the timestamp is maintained in session_cookies mode is not yet in alignment with the way the code works.  Before fixing that discrepancy (by changing the code and/or the documentation) we need to decide whether the cookie_timestamp or the database timestamp should take priority, or perhaps we should take the later one.

If one starts a session as an instructor for example and then starts a session as a student there are warning messages about conflicts between the cookie data and the hidden variable data embedded in the html page.  It seems to me the latter should take priority -- at least in this  case.  Comments?

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -777,18 +777,27 @@ $gatewayGracePeriod = 120;
 ## cookies is more appropriate when external authentication systems
 ## are used, e.g., LDAP, LTIBasic, etc.
 ##
-## The reason one might want to use cookies for session management
-## is to avoid having to obtain a write lock on the Key database
-## every time a request is received in order to update the timestamp
-## in the Key database.  When cookies are used for session management,
-## one obtains a write lock on the Key database for the original
-## login request in order to write the new session key and its initial
-## timestamp to the Key database, but on subsequent requests,
-## one merely obtains a read lock on the Key database in order
-## to verify that the session key in the session cookie is the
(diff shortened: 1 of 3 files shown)


goehle:
Having a cookie timestamp is actually something of a security issue.  The main thing is that if someone modifies the cookie and changes the timestamp to 99999999 they can unexpire keys and get around the timeout.  If the server speedup is enough it might be worth it.  If we know anyone with a really busy server we can ask them if using the key db is slowing things down. 
(Actually you might comment out https://github.com/openwebwork/webwork2/blob/master/lib/WeBWorK/Authen.pm#L750 as part of this pull just because it doesn't do anything except open up this hole.)

goehle:
I tested this out by manually changing my cookie timeout.  Before the patch I can get around the server timeout and after the patch I can't get around the timeout. 

This also changes the configuration files, however it doesn't change the description at https://github.com/mgage/webwork2/blob/develop_fix_cookies/conf/defaults.config#L770
In particular using "session_cookies" still makes use of the key db.  Should I hold off on merging or do you want to leave it as it is until we figure out what to do with the system?

mgage:
OK.  I've tried to make the documentation in defaults.config clearer.

PR #577, opened 2015-05-05
Develop fix null password

mgage:
Explicitly check whether the password stored in the database is null.  A null password can never be matched -- the only way to access this user account is to be handed a session key

Use case:  Moodle creates users with null passwords and creates a session key which it hands off to the browser.

Test: For many perl's the crypt function simply handled null entries and refused to match them against any entered password.  We were relying on this feature.  In at least one case the local perl did not behave this way and allowed any entered password to match against a null entry in the database.  Test to make sure this change does no harm.


goehle:
I tested this.  I manually set a blank password in the database.  Before the patch I could log in and after the patch I could not.

PR #578, opened 2015-05-07
Develop wwclient3 mods

mgage:
This pull request replaces wwclient2_mods.  It should at least be cleaner since I rebased the commits on top of the current develop branch.


lib/WeBWorK/ContentGenerator/renderViaXMLRPC.pm
```diff
@@ -154,9 +154,12 @@ sub pre_header_initialize {
 	$xmlrpc_client->{form_action_url} = $FORM_ACTION_URL;
 	$xmlrpc_client->{displayMode}     = DISPLAYMODE();
 	$xmlrpc_client->{userID}          = $inputs_ref{userID};
-#	$xmlrpc_client->{password}        = $XML_PASSWORD;
+	$xmlrpc_client->{password}        = $inputs_ref{password};
+	$xmlrpc_client->{site_password}   = $XML_PASSWORD;
```
mgage: Changing the key for the site password to site_password.  This the password that the xmlrpc interface uses to see
if it will even talk to the client.  It used to be called "password".    Now "password" has it's usual meaning of being
the password associated with userID where userID is assumed to be an instructor in some course.

mgage:
To test:
1. Change directory to webwork2/clients
2. Read the top part of renderProblem.pl

goehle:
I've been testing this.  One thing which is kind of a feature request (so feel free to knock it down) is that at least on the library browser the xmlrpc stuff fails mysteriously if you time out.  Would it be possible to add a time out message or redirect?

mgage:
I think so -- but that is a feature request. There is a lot more that I want to do with the authentication procedure to make it more robust.

mgage:
If you can figure out a way to make the "solution" and "hint" popdowns work that would also be nice. Is it as simple as adding a llnk to the javaScript in the right place?

goehle:
I already did this as part of the jitar2 branch.  Its mostly just having the js in the right place, but there were some other variables and things that needed to be passed correctly.

mgage:
OK.  Incidentally notice that we can have a large number of different specialized formats for the final HTML stored in WebworkClients.pm    Eventually we might break the formats out into a separate .pm module if they become too  numerous.

PR #581, opened 2015-05-15
Develop fix undef msgs

mgage:
Minor fixes to define things so as to eliminate "undefined" warnings.  

boolean as a bareword replaced by 'boolean' in defaults.conf

DESTROY() subroutines defined in some database subroutines.

### DIFF
--- a/lib/WeBWorK.pm
+++ b/lib/WeBWorK.pm
@@ -54,7 +54,7 @@ use WeBWorK::URLPath;
 use WeBWorK::CGI;
 use WeBWorK::Utils qw(runtime_use writeTimingLogEntry);
 
-use mod_perl;
+
 use constant MP2 => ( exists $ENV{MOD_PERL_API_VERSION} and $ENV{MOD_PERL_API_VERSION} >= 2 );
 
 # Apache2 needs upload class
@@ -64,6 +64,8 @@ BEGIN {
 		Apache2::Upload->import();
 		require Apache2::RequestUtil;
 		Apache2::RequestUtil->import();
+	} else {
+		require "mod_perl.pm"; # should we still support apache mod_perl1?
 	}
 }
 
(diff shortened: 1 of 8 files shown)


lib/WeBWorK.pm
```diff
@@ -64,6 +64,8 @@ BEGIN {
 		Apache2::Upload->import();
 		require Apache2::RequestUtil;
 		Apache2::RequestUtil->import();
+	} else {
+		require "mod_perl.pm"; # should we still support apache mod_perl1?
```
goehle: My vote on  this would be no.

goehle:
Are there specific testing instructions for this?

mgage:
1. In configuration of checkProblems.pl set the site url for rendering to something that doesn't exist -- you should get a "could not connect to rendering site error message when running checkProblems.pl input.txt from the command line
2. The undefined warnings you won't see unless you watch the error log religiously. I found them compiling some of these scripts directly with perl.  -- the lack of DESTROY() dummy subroutines for example and the unquoted word 'boolean' in the default.config file.  I suggest checking the code for 
obvious errors (it never hurts to have another pair of eyes look at it) and then running through a few
homework sets to make sure they display.  --- The biggest potential problem is that I have used the operator // which will not work on perl 5.8 and earlier.  -- I'm in favor of not supporting perl's earlier than 5.10 and not supporting mod_perl1 either.

Except for checkProblem.pl these fixes should not be very noticeable.  The main thing is to make sure that they don't do any obvious harm.

lib/WeBWorK/PG/Local.pm
```diff
@@ -44,9 +44,13 @@ use WeBWorK::Utils qw(readFile writeTimingLogEntry);
 #use WeBWorK::Utils::RestrictedMailer;
 use WeBWorK::Utils::DelayedMailer;
 
-use mod_perl;
-use constant MP2 => ( exists $ENV{MOD_PERL_API_VERSION} and $ENV{MOD_PERL_API_VERSION} >= 2 );
+BEGIN{
+ 	unless (exists $ENV{MOD_PERL_API_VERSION} and $ENV{MOD_PERL_API_VERSION} >= 2) {
+		require "mod_perl.pm";  # used only for mod_perl1  should we continue to support this?
```
goehle: Should this be 
``` 
$ENV{MOD_PERL_API_VERSION} < 2
```
?

PR #582, opened 2015-05-16
User Detail Bug

goehle:
On User Detail I added an "assign all sets" button.  The button works correctly, in that it assigns the sets.  However, in chrome and chromium for some reason the check boxes for the assigned sets are not checked.  I can't figure out why the existing code doesn't work so I implemented the assign all button in an alternative fashion using jquery. 

To test:  Go to the user detail page for a user which doesn't have all sets assigned.  (Classlist editor, click the number of homework sets) and assign all.  Before the patch on some browsers some of the check boxes will not be checked.  After the patch all the check boxes will be checked.


mgage:
I was able to assign all sets to the user in chrome, safari and firefox (using MacBook Pro).

PR #583, opened 2015-05-19
Develop wwclient4 mods

mgage:
Fixes which make knowl slide down sections (Solutions and Hints) work for WW embedded in HTML

Test using TestWW.html.dist.  You will probably need to copy that to TestWW.html and make some configuration changes if you are using a local server.


lib/WebworkClient.pm
```diff
@@ -588,6 +588,36 @@ $self->{outputformats}->{simple}= <<ENDPROBLEMTEMPLATE;
...
+<script src="/webwork2_files/js/legacy/java_init.js" type="tesxt/javascript"></script>
+<script src="/webwork2_files/js/apps/InputColor/color.js" type="text/javascript"></script>
+<script src="/webwork2_files/js/apps/Base64/Base64.js" type="text/javascript"></script>
+<script src="/webwork2_files/mathjax/MathJax.js?config=TeX-MML-AM_HTMLorMML-full" type="text/javascript"></script>
+<script type="text/javascript">mathView_basepath = "/webwork2_files/images/mathview/";</script>
+<script src="/webwork2_files/js/apps/MathView/mv_locale_us.js" type="text/javascript"></script>
+<script src="/webwork2_files/js/apps/MathView/mathview.js" type="text/javascript"></script>
+<script type="textx/javascript" src="/webwork2_files/js/vendor/underscore/underscore.js"></script>
+<script type="text/javascript" src="/webwork2_files/js/legacy/vendor/knowl.js"></script>
+<script src="/webwork2_files/js/apps/Problem/problem.js" type="text/javascript"></script>
```
goehle: I"m not sure problem.js shoudl be here since its specific to the Problem page in webwork2

lib/WebworkClient.pm
```diff
@@ -588,6 +588,36 @@ $self->{outputformats}->{simple}= <<ENDPROBLEMTEMPLATE;
...
+<script type="text/javascript" src="/webwork2_files/themes/math4/math4.js"></script>	
+
```
goehle: actually, math4.js and bootstrap.js are more for webwork2, unless you like how they format this page.

mgage:
You may be right that there are some extra .js scripts -- remember that at least one use case is when this is going into an iframe  to display the problem -- so these two might be relevant.  It's worth doing more tests. -- I mostly wanted to get knowls working properly and it took more javascripts than I thought. :-)

goehle:
This pull doesn't merge cleanly anymore after the opaque server mods branch was pulled

mgage:
OK.  I cleaned it up.  I pulled develop into my copy and pushed to the github.com/mgage site.  I'm not sure what changed but it will now merge cleanly.

goehle:
Looks good.  I can use knowls now, its inheriting enough bootstrap formatting that it looks kind of nice. Couple small things
-  You have some extra css files at the top of TestWW.html which can be removed. 
-  I would remove the three lines that include mathview, since there isn't any way to turn it off and it doesn't interact particularly nicely with the iframe.  It can always be added back in later with a different view.

PR #585, opened 2015-05-19
Ww3 feature/problem sets manager

pstaabp:
This is a fairly robust bug fix PR.  Hard to believe some it goes back to last summer. 

I guess this needs some general testing especially on the 
1) Calendar
2) Problem Set Manager
3) Problem Set Detail View
4) Classlist Manager. 

There's a list of testing criteria on the webwork2 repository blog for these sections.


goehle:
This no longer automatically merges.

pstaabp:
looks like pulling ww3 back in has it auto merging again.

goehle:
You accidentally pushed `ww3-feature/problem-sets-manager` to the openwebwork repository.  Is it safe for me to delete that?

pstaabp:
Yes.  Oops.

pstaabp:
I just did. 

Peter

goehle:
This pull fixes a lot of the issues I had with ww3.  In particular since the version of ww3 that is currently on openwebwork is a little unusable, I'm going to pull this just on the grounds that its a big improvement.  I do have some things to report, though. 
-  On Student Overrides there doesn't seem to be any date validation.  I can set illegal combinations of due dates/ reduced credit dates/ etc. 
-  I'm not sure what the Save Changes button does on Student Overrides since things seem to save dynamically. 
-  When I try to change my own information on the Classlist Manager I get a "Validation Error".  The associated js error is `Syntax error, unrecognized expression: :nth-child`
-  On problem set detail or the library browser when viewing problems the edit and view buttons don't seem to work. 
-  I can't find any way to set "max attempts" on the Problem Set Detail page.  This is a pretty core feature and needs to be added. 
-  On problem set detail when viewing problems the Value field doesn't validate.  It should only take integers.  
-  On problem sets manager I wasn't able to get the "Change Properties for Multiple Sets" feature to work.  When clicking save changes I get `TypeError: self.problemSets.findWhere(...) is undefined` in the js console. 
-  Add Users from a File is formatted very badly for me.  I wasn't able to test it because the table with the names and fields wasn't usable. 
-  I wasn't able to import sets.  I got `TypeError: config.settings is undefined`
-  There really needs to be a way to import sets from def files already on the server for classes that have been copied via the template path or if the site administrator puts def files in the templates or library directories.  
-  Small Thing:  For some reason text in the Password field persists in the Add Users to Course dialog.  Also, if I click Save Student and Add Another then I *have* to add another student if I want my previously entered student to be added. 
-  On the library browser if I try to view a textbook I get the js error `TypeError: memo is undefined`.  (I did run OPL update to check if that was the problem)
-  I wasn't able to get local problems or SetDefn. Files or Search to work on the library browser.

pstaabp:
Thanks Geoff.  I'll add these bugs to my list.  Some of these I know I haven't done and some (like the library, I have a lot of items to go on).

Issue #479, opened 2014-09-07
WW3 Security Issue


goehle:
A lot of the paths for WW3 just check the permission level of the current dancer session.  However, they don't validate the WeBWorK authentication cookie, that the user has the correct permission level in a given course or even that a user exists. 

For example, log into a test course with a test professor.  In my case it was TestCourse, with "testprof".  Then go to `http://webwork.wcu.edu/webwork3/courses/:course/users` for a course 'different' than the course you logged in with.  You should be shown the users in that course.  This happens even if "testprof" is not a member of the course.  

Looking at the code there are a lot of routes with this problem.  Things with 
```
  if(session->{permission} < 10){send_error($PERMISSION_ERROR,403)}
```
and no other user validation code will be vulnerable.   This includes most of Users.pm.   Things with 
```
    checkPermissions(10,session->{user});
```
do check against the WeBWorK session and all routes should do something like that.

pstaabp:
I agree that this will need a more thorough rethinking moving forward.  It doesn't check the permission level set in the course configuration, which it should do as well.  I'll fix this first and then move to a more sophisticated permission checking.

PR #588, opened 2015-05-26
Changed the way that permissions are checked for almost all routes.  …

pstaabp:
This fixes issue #479.  

To test, log into one course.   Try fetching data from another course, like going to ` http://webwork.wcu.edu/webwork3/courses/:course/users` for :course a different course.  Also try "sets" at the end.  You should now receive an error. 

Also, try going to ww3 for a course and testing various areas that the respond the same way as before.  Changing dates, adding problems, etc.

### DIFF
--- a/webwork3/lib/Routes/Authentication.pm
+++ b/webwork3/lib/Routes/Authentication.pm
@@ -166,11 +166,11 @@ sub setCookie {
 
 		my $hostname = vars->{ce}->{server_root_url};
 		$hostname =~ s/https?:\/\///;
-
+        
 		if ($hostname ne "localhost" && $hostname ne "127.0.0.1") {
-			cookie "WeBWorKCourseAuthen." . params->{course_id} => $cookieValue, domain=>$hostname;
+			cookie "WeBWorKCourseAuthen." . session->{course} => $cookieValue, domain=>$hostname;
 		} else {
-			cookie "WeBWorKCourseAuthen." . params->{course_id} => $cookieValue;
+			cookie "WeBWorKCourseAuthen." . session->{course} => $cookieValue;
 		}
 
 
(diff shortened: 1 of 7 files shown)


goehle:
This pull request tests correctly.  Some generic issues that I found are below.  I did restart the server and did hard refreshes on the pages.  Its possible that I'm forgetting something and these are config issues, or just problems with getting ww3 back on its feet on my machine. 

-  I can't change properties of a set with name "this-has-a-pretty-long-set-name" using the Problem Set Details page.  I get the error
```
Uncaught TypeError: undefined is not a function : config.js: 88
```
-  I get the following errors from the library browser 
```
LibraryTreeView.js:98 Uncaught TypeError: Cannot read property 'subfields' of undefined
backbone.stickit.js:202 Uncaught TypeError: Cannot read property 'label' of undefined
```
and if I try to add problems to a set using the + button I get
```
PUT http://webwork.wcu.edu/webwork3/courses/TestCourse/sets/14545454 500 (Internal Server Error)jquery.js:8623 jQuery.ajaxTransport.sendjquery.js:8152 jQuery.extend.ajaxbackbone.js:1209 Backbone.ajaxbackbone.js:1188 Backbone.syncbackbone.js:286 _.extend.syncbackbone.js:493 _.extend.saveProblemSet.js:118 Backbone.Model.extend.addProblemLibraryView.js:78 TabView.extend.addProblemProblemView.js:168 Backbone.View.extend.addProblemjquery.js:4409 jQuery.event.dispatchjquery.js:4095 jQuery.event.add.elemData.handle
```
- If I try to change a students status I get
```
underscore.js:1074 Uncaught TypeError: Cannot convert undefined or null to object
```

There are others.  I'm reasonably certain that I've got something wrong, but I can't figure out what it is.  So I'm going to merge this.  I'm also going to pull develop into ww3 and we will see where that leaves everything.

pstaabp:
I think some of these are fixed in the other pull request I just submitted.  I tried to just concentrate on the permissions issue in this one. 

Peter

PR #590, opened 2015-05-28
Database Engine type fix

pstaabp:
This sets the default database engine to innodb. for  

Also, not sure why there are 146 commits here?  I only made one commit in one file.


pstaabp:
oops.  Sent this to mast.

Issue #589, opened 2015-05-28
database problem on new install


pstaabp:
I've been upgrading the software on  my development machine, since almost everything on it was 4+ years old.  I have a new installation nearly running, however when getting up and running I'm adding a new course (the admin one) via the command line with `addcourse admin`  (I know I need to add users too), and when doing this I get the following error:
```
DBD::mysql::db do failed: Specified key was too long; max key length is 1000 bytes at /Library/WebServer/Documents/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm line 826.
```
I'm not exactly sure what triggers this (is there a way to get a stack trace on this?) 

I'm installing this on a Mac and just upgraded mysql to version 5.6.24.  This occurs on both the master and develop branches.

pstaabp:
So I went to post this on the webwork forums and saw the error.  Okay I withdraw my post.  

We need to change the line 150 of site.conf.dist
```
$database_storage_engine = 'myisam';
```
to
```
$database_storage_engine = 'innodb';
```

I'm happy to do this.

pstaabp:
Geoff, do you want me to submit this as a pull request?

goehle:
Sure, although I wish I knew more about what was going on.  I've been using myisam for all my installations up to this point and its worked fine.  So its not clear to me what changed.  Which forum post are you talking about, btw?

pstaabp:
Here's the post:

http://webwork.maa.org/moodle/mod/forum/discuss.php?d=3174#p7891

Switching to innodb fixed the problem on the thread and also for me.  This thread is also pointed out in the installation instructions (http://webwork.maa.org/wiki/Installation_Manual_for_2.10_on_Ubuntu_14.04#MySQL_Security_and_Performance_Issuses ) 

It appears that leaving that line in site.conf  is an easier way to set the database type than the mysql settings listed in the documentation.

goehle:
Thanks for the post, that was helpful.  I'm reopening the issue since it seems like there is something to talk about. So its not clear to me that we need to change the default database type in `site.conf.dist`.  This problem appears to be at least two years old and I've done many installations since then which work fine for myisam.  Also, this seems to be a problem only for those people who change the default mysql character encoding to utf8 since utf8 characters take more bytes.  We have a few options. 
-  Change the default database type.  There seem to be performance bonuses to myisam on some systems, and the default character set for mysql is latin1, not utf8 so in general myisam should work.  
[...]

PR #591, opened 2015-05-28
Change the default database engine type

pstaabp:
Changed the default database engine type to InnoDB, which seems to be more robust than mysiam.

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -154,7 +154,7 @@ $externalPrograms{mysqldump} ="/usr/bin/mysqldump";
 # The format is dbi:mysql:[databasename] for databases on the local machine
 # For a remote database the format is dbi:mysql:[databasename]:[hostname]:[port]
 $database_dsn ="dbi:mysql:webwork";
-$database_storage_engine = 'myisam';
+$database_storage_engine = 'innodb';
 
 # The following two variables must match the GRANT statement run on the mysql server as described above.
 $database_username ="webworkWrite";


pstaabp:
This resolves issue #589, which also appeared in the maa.org forums.

mgage:
two years ago there were bad performance issues (see posts by Arnie in the forums and in the installation documents for wiki) particularly when used for the library.  As a test try running the OPL-update on your innodb system and see if the performance is acceptable.  It may well be that the performance of innodb with mysql has improved.  I know that my moodle installation now forces me to use innodb.

I seem to remember that there was another way to resolve the long key issue -- but I haven't found a reference yet. Perhaps Arnie remembers.

pstaabp:
In terms of performance, do you mean just running OPL-update?  If so, I don't see that it is any slower than before.

goehle:
You may have to delete your library tables first so that they are rebuilt as innodb.

pstaabp:
I think OPL-update always deletes the tables.  Anyway, to test, I deleted all of the tables.  I just ran it again and it took about 6 minutes.  I think that's about the same.  I decided to use mariadb 10.0 (to test it out), so maybe that has improved performance too.  I can test mysql as well if you'd like me to.

goehle:
You might ask John Jones about it as well.

jwj61:
There was definitely a huge difference before with OPL-update between database types using mysql.  Since mysql is the standard used by webwork, testing with that seems to make sense.

mgage:
Also make sure that the library tables are being built with innodb and not myisam.  Not all tables in a mysql database are of the same type.

mgage:
that said -- it's possible that the performance problems with innodb have been fixed over the last two years -- that gives us more choices and innodb has some advantages

pstaabp:
It doesn't look like putting the line
```
$database_storage_engine = 'innodb';
```
actually made a innodb engine.  Looking at the actual table, I'm getting "MyISAM" in the OPL_update.  So I'll try to track down what the line:
```
$database_storage_engine = 'myisam';
```
does and why I get the error I was getting.

goehle:
You might check if OPL-update actually deletes the tables or just empties
them.  If it empties them then it retains stuff like the storage engine.

aubreyja:
I'm pretty sure that the setting in site.conf sets the type for webwork2
tables, but OPL-update uses the perl dbi directly and doesn't go through
the (ab)normal webwork DB interface.  So OPL-update probably just uses your
MySQL default engine.

pstaabp:
Actually, it looks like OPL-update uses MyISAM engine and now with the line above, I'm getting other tables, like test_user to be a "innodb" table.   

It's hard-coded as "MYISAM" in defaults.config in line 540.

dlglin:
The storage engine for the OPL tables are handled separately from the WW tables.  $database_storage_engine would cover tables created by the GUI.  The OPL tables are governed by $problemLibrary_db->{storage_engine} which is first set in defaults.config.

pstaabp:
Reading through OPL-update, the tables are dropped (deleted) at the beginning of the script and then rebuilt.

pstaabp:
Quick experiment here.  If the library database engine is "MyISAM", then the speed of OPL-update in nearly 10 times the speed when it is "InnoDB".

pstaabp:
Returning to the original purpose of this pull request, we may need to do something about the line in site.conf.dist or determine another work around.

mgage:
A 10 times speed differential is about what I remembered.  Of course this is without having done any optimization for innodb tables (see some of the references made by Danny in the discussion a few years ago).  If someone has the time to experiment it's possible that, with work, innodb can be made much faster. It's advantage would be with read/write situations under heavy load since it is transactional. For the library browser which is almost all read and no write it probably doesn't have any advantage so unless the speeds are nearly identical it's best to leave the library tables on myisam.

goehle:
What is the character set for your server and could you change it to the default (latin1) ?

If we really want to support utf8 then I suggest we follow Danny's suggestion and change the indexing to only look at the first 100 characters of key fields.

dlglin:
It looks like WW is hard-coded to cut off text/blob key fields at 255 characters, which according to the comments in the code is the limit in mysql.  100 is just as arbitrary of a number, and should solve this issue.  As Geoff says, it is unlikely that these fields will be anywhere near 100 characters anyway, so I prefer this as the fix.

Of course the real solution is to add a numeric key to each table rather than keying on things like usernames and set names, but that would be a significant rewrite.

goehle:
I've made an attempt at a numeric key a couple of times.  However, the gains never seem to justify the pains, as it were.  The main issue currently is that because of gateways the UserSet and UserProblem tables are not actually joins of User and Set and User and Problem, respectively.

goehle:
If you want to try out the solution mentioned above try the dbkeylength from my repo: 
```
git pull https://github.com/goehle/webwork2.git dbkeylength
```
In particular, Peter if you could try it with myisam to see if it fixes your troubles.  I also had to change problem_id to int in the past_answers database.  I don't *think* this impacts existing tables, but there could be a side effect I'm missing.

pstaabp:
It looks like that worked for me.  I guess what we can do is to pull Geoff's dbkeylength into develop instead  (does it need to go in master too?) to eliminate this error.

goehle:
I wouldn't pull it into master.  I want to be sure that changing the key index length, and changing problem_id to int in the past answer table, doesn't have an unforseen concequences.  Since this has been an issue for years there doesn't seem to be a rush.

pstaabp:
Since I just upgraded nearly everything on my machine, maybe the newer version of mysql defaults to UTF8 and since most won't do this often, we probably won't see this much.

PR #592, opened 2015-05-29
Updates of all the standard js libraries used for ww3. 

pstaabp:
In addition, some small changed needed to be made for some incompatibilities with updated library apis.  (For example the _.template function from underscore.js changed and momentjs deprecated one of functions). 

Note:  it appears that the calendar functionality broke with this.  If you drag an assignment to a new day, it changes the date, but doesn't rerender the calendar correctly.  I have reworked this in another branch to do a better job separately code from the views, so didn't feel like tracking down this regression.


goehle:
I'm running through and doing some testing for this.  In theory we want things hosted on openwebwork to be at least reasonably functional.  So I'm hesitant to pull anything broken.  I would prefer to have a "parallel" pull request with code changes that fix some of these issues and then pull them both at the same time.  The big ones are: 
-  The calender is broken, as you noticed above. 
-  I can no longer change dates on student overrides.  When I try to I get the js error `ReferenceError: moment is not defined`
-  The Student Progress view seems to be non functional.  Selecting a type and a for doesn't display a table.  
-  There is a weird "timezone" tab in Settings with no entries.  Timezone is also missing from the General settings tab  where it should be.  

I also ran across some other things which don't have to get fixed right away but should be added to your list.  
-  Some of the messages in "All Messages" are nonsensical.  For example
``` The value of value_changed in problem set test_set has changed from [object Object] to undefined. 
```
-  When exporting a classlist to a csv the exporter does not properly quote entries with commas. So if you have a comment with a comma in it you get a broken csv file.  
-  In the library browser Local Problems and Set Defn. Files don't get past the spinny "loading library" message.  They never worked for me anyway, but now they don't even load the drop down. 
-  Choosing images in "Library Options" doesn't cause problems to be rendered with images.

pstaabp:
* I fixed the student overrides error, the timezone weirdness and the local problems spinny problem.  
* If you can let it slide, I'll get to the calendar in the next pull. 
* I haven't seen the all messages problem.  I'll look for that and let me know if you can reproduce it
* I fixed the comma problem, by wrapping all of the cells in "  ", but now have an issue if there's a "  in a cell.  :(  I think I found a library that might help this. 
* Are there any errors with the images in the library.   Or is it showing the math as mathjax?

goehle:
I would find a library for the csv stuff.  Doing csv sounds like it should
be just split, but there are actually dozens of corner cases.  Its kind of
a pain.

As far as I can tell rendering all the math as mathjax.  Although it could
be showing all the math as text and then mathjax is automatically rendering
that text.

pstaabp:
I added a CSV library which does a read a CSV file, but not export to one.  I altered my code for export and it appears to handle both commas and double quotes.  They say that export to CSV is coming, and if that does happen, I'll switch to that. 

About the the rendering, I'm in the library browser (By Subject) and select some problems.  They render correctly in MathJax.  In the side pane "Library Options" I switch to Images and it rerenders everything with images.   Is this what you did?  I'm assuming images are working in ww2?

What happens in the Problem Set Details view? Can you switch rendering modes?

goehle:
The csv stuff looks better.  I also can't reproduce my images issue, so ignore that.  Student Progress is still broken.  If you can fix that I'll pull this, even though calendar is also broken.

pstaabp:
The StudentProgress view is now working. It obviously needs some work, but it functioning again.

goehle:
Ok.  I'll pull this.  All of the above issues are fixed (except for the calendar as noted).

PR #593, opened 2015-06-01
Add library tag RESOURCES

jwj61:
The RESOURCES tag holds a list of files to be copied around with the current file (e.g., static image files).  When submissions are made to Contrib, they are copied to Pending, and if accepted to the OPL.  This will help keep track of auxilliary files.

### DIFF
--- a/lib/WeBWorK/Utils/Tags.pm
+++ b/lib/WeBWorK/Utils/Tags.pm
@@ -30,11 +30,11 @@ use IO::File;
 
 our @EXPORT    = ();
 our @EXPORT_OK = qw();
-#	list_set_versions
-#);
 
 use constant BASIC => qw( DBsubject DBchapter DBsection Date Institution Author MLT MLTleader Level Language Static MO Status );
 use constant NUMBERED => qw( TitleText AuthorText EditionText Section Problem );
+
+# KEYWORDS and RESOURCES are treated specially since each takes a list of values
 
 my $basics = join('|', BASIC);
 my $numbered = join('|', NUMBERED);
@@ -44,6 +44,7 @@ sub istagline {
   my $line = shift;
   return 1 if($line =~ /$re/);
   return 1 if($line =~ /#\s*\bKEYWORDS?\s*\(\s*'?(.*?)'?\s*\)/);
+  return 1 if($line =~ /#\s*\bRESOURCES?\s*\(\s*'?(.*?)'?\s*\)/);
   return 1 if($line =~ /#\s*\b($numbered)\d+\s*\(\s*'?(.*?)'?\s*\)/);
   return 0;
 }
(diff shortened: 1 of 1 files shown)


goehle:
I'm trying to find a way to test this.  This doesn't seem to add anything to the OPL database and the tagwidget.js pathway doesn't support it.  Neither `dumptags` or `addresourse` seem to be used and I cant find anything that either sets or uses a resource tag.

mgage:
Hi John,

Try adding this to the end of one of your .pg problem files.  It will print out all of the resources which have been found using "alias".  This will include most html and png files but not (yet) javaScript files.  Still it might be helpful in collecting the resources for each file.  You can use what's already available in PGalias to automate at least some of the work.  I hope to work some more on this over the summer.

`pretty_print(  $PG->{PG_alias}->{resource_list} , $displayMode );`

jwj61:
There is nothing in webwork which accesses this (so it can't hurt :)).

Part of the intent of Tags.pm is that it can be used by either command line scripts, or a webwork server.  At the moment, this is a tag only accessed by command line scripts of the OPL managing editor.  In addition to helping there, it officially makes RESOURCES the name of the tag for listing auxilliary files.

In the future, webwork servers might want to use it in the following scenerio: user wants to edit an OPL file, so they have webwork do a "save as".  Currently, webwork doesn't know to copy the auxilliary files into the new directory, so the problem is broken.  If OPL problems are tagged with their resource files, it would be easy to copy them too.  (If I understand your second comment, it seems relevant to figuring out what should be put in the RESOURCES tag, rather than something webwork should be doing to figure this out again and again.)

A simple test is as follows.  Inside webwork-open-problem-library/Pending/utilities we now have some scripts.  If you in a directory with myfile.pg, you can do

 path/to/addresource.pl myfile.pg image1.gif
 path/to/addresource.pl myfile.pg image2.gif
 path/to/addresource.pl myfile.pg image2.gif
 path/to/addresource.pl myfile.pg image2.gif
 path/to/addresource.pl myfile.pg image1.gif

And myfile.pg will have one RESOURCES tag listing the two files (not repeated).

More complicated testing (which I have done if you want to take my word for it) would be to use contrib2pending.pl from the same directory.  For it to do something, you would want a temporary clone of the OPL (which you promise to delete after mucking it up), delete Pending/GRPS/isosceles4.pg and Pending/GRPS/isosceles3.gif, then run

path/to/contrib2pending.pl /absolute/path/to/webwork-open/problem-library

and the two files from Pending should reappear.  If you go back, *and* remove the RESOURCES tag and repeat, you only get the pg file and not the image.

Then delete your temporary copy of the OPL (you promised).

goehle:
I ran through your tests and everything works.  Thanks for walking me through the OPL process.  Its good to at least have a basic understanding of how things are working on that end.

PR #594, opened 2015-06-02
DB Key Prefix Length

goehle:
This pull request does two things.  It changes the index prefix of text/blob fields to 100 characters, instead of the 255 it was at before, and it changes the type of problem_id in the past_answers table to int instead of varchar.  Together these mean that there are at most 300 total characters worth of indexed keys, and with 3 bytes per character with utf8, that puts us below the max 1000 byte width for myisam tables. 

-  To test, change your mysql servers default character set to utf8 and try to add a course.  You should get an error about having too wide of a key.  
-  After the pull request you should be able to create a course.   
-  You should also test that the past answer database still works correctly for both old and new courses.  Try submitting a problem answer and viewing past answers.

### DIFF
--- a/lib/WeBWorK/DB/Record/PastAnswer.pm
+++ b/lib/WeBWorK/DB/Record/PastAnswer.pm
@@ -33,7 +33,7 @@ BEGIN {
 	    course_id         => { type=>"VARCHAR(100) NOT NULL", key=>1},
 	    user_id           => { type=>"VARCHAR(100) NOT NULL", key=>1},	
 	    set_id            => { type=>"VARCHAR(100) NOT NULL", key=>1},
-	    problem_id        => { type=>"VARCHAR(100) NOT NULL", key=>1},
+	    problem_id        => { type=>"INT NOT NULL", key=>1},
 	    source_file       => { type=>"TEXT"},
 	    timestamp         => { type=>"INT" }, 
 	    scores            => { type=>"TINYTEXT"}, 
--- a/lib/WeBWorK/DB/Schema/NewSQL/Std.pm
+++ b/lib/WeBWorK/DB/Schema/NewSQL/Std.pm
@@ -146,12 +146,12 @@ sub _create_table_stmt {
 		foreach my $component (@keyfields[$start .. $#keyfields]) {
 			my $sql_field_name = $self->sql_field_name($component);
 			my $sql_field_type = $self->field_data->{$component}{type};
-			my $length_specifier = $sql_field_type =~ /(text|blob)/i ? "(255)" : "";
+			my $length_specifier = $sql_field_type =~ /(text|blob)/i ? "(100)" : "";
 			if ($start == 0 and $length_specifier and $sql_field_type !~ /tiny/i) {
 				warn "warning: UNIQUE KEY component $sql_field_name is a $sql_field_type, which can"
-					. " hold values longer than 255 characters. However, the maximum key prefix"
-					. " length for text/blob fields is 255. Therefore, uniqueness must occur within"
-					. " the first 255 characters of this field.";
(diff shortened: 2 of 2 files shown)


dpvc:
Just FYI, utf-8 is a variable-width encoding, not a 3-character encoding.  It can use anywhere from 1 to 6 bytes per character.  For ASCII characters (U+0000 to U+007F), it uses one byte, for U+0080 to U+07FF (this includes most of the Latin accented characters) it uses 2 bytes, for U+0800 to U+FFFF (the rest of Unicode's plane 0), it uses 3 bytes.  But for unicode Plane 1 and higher (which includes things like the Math alphabetic characters), it can use 4, 5, or even 6 bytes.  I doubt these will be used much, however, and since most characters will be taken from the ASCII or 2-character range, it is unlikely that an occasional 4-byte character will be a problem.

goehle:
So looking it up, for mysql utf8 uses 3 byte characters and so cannot
support 4 byte utf characters.  If you wanted that you would set your table
character encoding to utf8mb4.  Those support 4 byte characters, but would
cause us to overflow the key limit.  (
https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html)

So I guess the question is if we are changing the prefix length to support
utf8, should we change it so that it also supports utf8mb4?  We would need
to do something like 80 characters per key.

On Tue, Jun 2, 2015 at 12:46 PM, Davide P. Cervone <notifications@github.com

dpvc:
If you look carefully, the link you provided says it is a *maximum* of 3 bytes per character.  Characters in the ASCII range will only take 1, not 3, which (I suspect) is what most characters will be.  Even most European accented characters will only be 2.  Something like Devanagari (used for Indian scripts) would require 3.  The mysql `utf8` implementation limits you to the basic multi-lingual plane (BMP, Unicode plane 0), but I suspect that that is not a problem, unless someone is going to use Imperial Aramaic or Phoenician or Cuneiform, which are in Plane 1.

Using `utf8mb4` would use exactly the same amount of space as `utf8` for anything in the BMP (the first bullet point in the link you gave), so this would not have adverse consequences for any key that doesn't use characters from Plane 1.  You would only need to limit to 80 characters if you expected handle something that is *entirely* in Plane 1.  Switching to `utf8mb4` would mean you could do all the same keys that you can with `utf8`, plus could use ones that includes Plane 1 characters (provided there were also some ASCII characters or ones from below U+07FF to compensate).

Yes, you would run the risk of overflowing if all characters came from Plane 1.  I seriously doubt that will happen.

pstaabp:
The addcourse script now works and I'm able to create problem sets, add problems and view past answers.  This looks fine.

goehle:
I'm not sure we were done with this Peter...

goehle:
I think the issue is that when mysql is checking for the key length its
using the maximum length of each character.  (Thats why it complains at the
time of table creation, before any rows have been entered.)  So if we plan
on supporting utf8mb4 then we would need to set the maximum indexed length
to 80 characters, even if we never expect to have 80 4 byte unicode
characters.   I just did a quick test and I can confirm that if you set the
default character encoding to utf8mb4 you cannot create a course, even with
the patch.

pstaabp:
ok.  oops.  I was a bit hasty.  I can reset develop to the previous commit.

goehle:
I can always submit a new pull that overwrites this one if it needs to be changed.

pstaabp:
That would be easier.

dpvc:
> I think the issue is that when mysql is checking for the key length its using the maximum length of each character.

OK, that makes a difference.  I think you shield use `utf8`, then, as the Plane 1 characters are unlikely to be needed.

goehle:
Sounds good.  I'll close this pull request as already merged then.

PR #595, opened 2015-06-04
Improve tagging menu system

jwj61:
This does several things related to tagging menus:

  1. move loading of tagwidget.js code to the header (i.e. to output_JS) in the two content generators which use it (still conditioned on the user having permission to modify tags)
  2. load underscore.js in Problem.pm if we are loading tagwidget since it uses it
  3. have OPL-update generate a json file with the full taxonomy (but not showing cross-lists)
  4. use this json file in the tagwidget

(1) is mainly aesthetic, but also avoids multiple reloading of the code in the library browser.

(2) fixes a bug (the tagwidget was throwing an error when the tagging menus appeared on a single problem page).

(3) This creates the json file, akin to the ones Peter added.  Together with (4), this makes the updating of tagging menus much faster since the taxonomy is only pulled once instead of lots of ajax/database calls.  

This also fixes a bug.  The tagging menu needs to be different from the search menus in the library browser in two ways: the search menus have empty "sections" or "chapters" removed, but they need to be there when someone is tagging problems; and cross-listed sections should appear when a user is searching, but nothing should be tagged into these "virtual sections".  Before, the tagging menus just used the search menus, and now they get it right.  For example, go to "Complex analysis"/"Complex equations" and view problems.  The search menu has only "Linear" but the tagging menus have "Quadratic" and "Non-linear" as well.

### DIFF
--- a/bin/OPL-update
+++ b/bin/OPL-update
@@ -438,6 +438,9 @@ if(open(IN, "$libraryRoot/Textbooks")) {
 my $clsep = '<<<';
 my $clinner = '__';
 my @cllist = ();
+# Record full taxonomy for tagging menus (does not include cross-lists)
+my $tagtaxo = [];
+my ($chaplist, $seclist) = ([],[]);
 
 my $canopenfile = 0;
 if(open(IN, "$libraryRoot/Taxonomy2")) {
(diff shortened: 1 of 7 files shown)


goehle:
I'm checking this out and it looks like  underscore.js  is loaded on the Problem.pm page twice.  Its loaded for knowls in line 2011.  It should only be loaded once, although it can be moved around if necessary. 

Otherwise things check out.  I messed around and cross listed sections don't appear but blank ones do.   The json file is created and loaded (only once).

goehle:
P.S.  I  figured out the issue.  Instead of adding another line loading underscore.js you should change the "type" in line 2011 of Problem.pm to `text/javascript` instead of `textx/javascript`

jwj61:
Thanks for catching that, and sorting out the problem.  It should now be fixed (textx->text and removing the second loading of underscore).

jwj61:
hold on, there may be an issue with unwanted merges.

goehle:
The graph looks a little wierd.  It looks like you had a local copy of develop, made commit a65d49fd8f634c then forked off of that local copy of develop and then made additional commits.  It would be nice if all the commits were on a completely fresh copy of develop.  I don't know how to fix it, though, and am not sure its super important.

dpvc:
@goehle, the tree is not ideal, but it should be fine.

jwj61:
The part which concerns me are the changes to Std.pm and PastAnswer.pm.  The diff page indicates that there are changes to those files, but looking at the develop versions, it appears that those files will not change after all.  So, I don't object to it being merged.

dpvc:
@jwj61, It looks like you might have cherry-picked or rebased a couple of commits from Geoff onto the end of your branch, and those are duplicate commits, which will be ignored during the merge of this pull request.  So I think that should be OK.

PR #596, opened 2015-06-08
Develop embed2

mgage:
Modifications to allow embeddings of files into html pages.  Including embedding the source.

There is also some cleanup of the files WebworkWebservice/RenderProblem.pm and
WebworkClient.pm.  The CSS and JS code was added to the "standard" format in WebworkClient
to match the CSS and JS code in the "simple" format.


mgage:
This is not quite as scary as it looks.  Most of these changes are new files:

AttemptsTable.pm contains a version of the formatting used in Problem.pm to format the table reporting the results of the attempt to the student.  It is currently used in WebworkClient.pm to return answers to problems embedded in HTML pages.  The substantial changes are in these two files. 

There are some files in webwork2/t which can be used for testing AttemptsTemplate.pm.   To futher test:  Customize `clients/WWTest.html.dist` to point at a webwork installation that has been updated with this pull request.  Make sure that the four problems work and that when answers are submitted the responses are appropriate.

AttemptsTable.pm will be used in returning problems to the moodle quiz and eventually it will be used to replace the code in Problem.pm.  This is is one step in making Problem.pm more modular.

goehle:
The graph for this looks a little wierd.  It looks like you made some commits to your local copy of develop and then decided to branch off that and then make more commits to this branch.  Ideally this develop_embed2 would branch off of an unmodified version of openwebwork/develop.  To be fair, I don't exactly know how to make this happen.

mgage:
OK.  I removed the develop tag from my own repo at github.com/mgage and the graph looks cleaner now.

goehle:
I've done some testing with this and noticed the following.  You are going to have a follow up pull request, so these fixes cane go in this or that one.  
-  The message box is always there, (and colored yellow) even if its blank.  
-  The "Score" label on the answer box should probably be "Result" like it is in Problem.pm.  It also should contain js so that the result text is a link which takes you to the corresponding input field.  
-  Something to think about.  The way this currently works it is logging in each submission.  This means that it is also logging *out* anyone else who is currently logged in.  This may or may not be an issue if it is used in such a way that everyone is using the same "daemon" setup.  
-  There is no preview button and I cant find a way to not show solutions or the correct answer.   (In general it still behaves pretty differently than Problem.pm in some key ways)
-  The display mode is not preserved when you submit answers.  If I set it to MathJax by passing the appropriate parameter, when I hit submit it goes back to images.  (Note:  display mode "tex" does seem to work and could be a good way to get raw latex.)

mgage:
1. I think I can fix the message behavior easily so it behaves like Problem.pm
2. I'll change Score to Result as well. 
3. the behavior with the daemon login is important to keep track of -- we may need to change that if it is causing trouble.  We may need to change the authentication methods anyway depending on how we continue to develop this feature.  For now let's just watch it and see what happens.
4.  The moodle use of attemptsTable has buttons: preview, submit and grade would that be preferable? -- the original use case for html embedding was just for a one shot example in the middle of text so I didn't put in all the other features. -- I could add them -- but I recommend making that a feature request for the next update.
5.  I'll fix the display mode -- it is supposed to default to the display mode set in daemon_course

mgage:
1. Should be fixed now.  The student's simplified answer is not shown (as it usually is in problem)
 ----- that will probably be done eventually but not for this pull request.
2. fixed
5. fixed
4. is a feature request for now -- I'll see how soon I can get to it.

mgage:
4 -- Added Preview, Submit answer, and Show correct answer buttons

goehle:
Oops, I already merged
You'll have to make a new pull

PR #597, opened 2015-06-19
Bugfixes

goehle:
This fixes two small bugs. 

-  If you visit a gateway the log out button is now improperly formatted.  This pull request fixes that. 
-  Warning output when process via Apache/Webwork.pm (not ContentGenerator.pm) has double html encoding.  This fixes that.  If you put something like 
```
warn("<script>alert('hi')</script>);
warn("This has <=>");
die('die!');
```
in, say, Problem.pm then the resulting warning out put will be badly formatted before this patch.

### DIFF
--- a/htdocs/themes/math4/gateway.template
+++ b/htdocs/themes/math4/gateway.template
@@ -73,6 +73,12 @@
 
 <div class="row-fluid">
 
+<!-- Run masthead class js because it gets rendered early -->
+<script type="text/javascript">
+  // Changes links in masthead
+  $('#loginstatus a').addClass('btn btn-small');
+  $('#loginstatus a').append(' <span class="icon icon-signout" data-alt="signout"></span>'); 
+</script>
 
   <div id="content" class="span12">
 		
--- a/lib/Apache/WeBWorK.pm
+++ b/lib/Apache/WeBWorK.pm
@@ -187,7 +187,7 @@ sub htmlMessage($$$@) {
 	$warnings = htmlEscape($warnings);
 	$exception = htmlEscape($exception);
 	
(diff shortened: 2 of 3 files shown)


pstaabp:
The logout button seems to be fixed.  I'm not noticing a difference between formatting between the two branches.  This is how it looks:
```
<script>alert('hi')</script>
``` 
Did you expect it to throw an alert?

Which I put in the pre_header_initialize subroutine of Problem.pm  I tried also in the body subroutine with no warning at all.  Is there somewhere specific I should try?

goehle:
Putting it in pre header initialize should work.  Body wont work because that isn't used anymore.  Make sure you have the die command.  Otherwise the warnings use the ContentGenerator warning pathway and not the WeBWorK warning pathway.  Before the patch those warnings should give you something like: 
```
&lt;script&gt;alert(&#39;hi&#39;)&lt;/script&gt; at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 494.<br />This has &lt;=&gt; at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 495.<br />
```
where there is only one bullet point and things are clearly double encoded.  After the patch you should get 
```
    <script>alert('hi')</script> at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 486.
    This has <=> at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 487.
```
with two bullet points and no double encoding.  The alert should *not* trigger in any case to prevent XSS stuff.

pstaabp:
I see this now.  I didn't have the die statement in it and in both cases they wear outputting both to the red box at the bottom of the problem page. 

I will merge this in.  Doesn't look like there could be ill effects with this.

Issue #472, opened 2014-08-14
WW3 and Open Sets


goehle:
-  WW3 shouldn't allow users to reorder problems in open sets.  This will break their homework systems (at least until we get the id/name thing sorted out.)  
-  Is there a way to "mark all correct" using WW3?  You can delete problems from open sets and it won't reorder them unless you pick up a problem, but I couldn't find "mark all correct". 
-  Is there a way to modify UserProblem objects, i.e. scores/seeds/etc for problems for individual students?

Issue #316, opened 2014-01-27
Homework Manger : Set Details


goehle:
This form is missing the "description" field.  This field is also not being displayed (either as text or via tooltip) anywhere that I can see.

Issue #286, opened 2014-01-14
Homework Manager: Set Details


pstaabp:
The display mode initially is set to "images" even if the default settings is MathJax.

pstaabp:
This is fixed in pull request #598

PR #598, opened 2015-06-19
Ww3 feature/problem set details

pstaabp:
I guess I forgot to put in this pull request.  Overall this updates the Problem Set Details views with many bug fixes.  This includes:

1) the ability for the maxAttempts for a problem to be changed. 
2) the value (weight) of a problem to be validated 
3) A new tab in the ProblemSetDetails view that has a way to add the description and set the headers.  issue #316 
4) Reworking the Student Overrides tab to make it clearer how to change the dates for multiple students at once. 

This clears up issue #286

A couple of questions about issue #472:

1) About mark all correct.  Do you mean marking all problems correct for a given user in a given set?  Or something else?  I'm trying to brainstorm when to add that functionality
2) Also about changing users seeds/score for individual problems;  how would you suggest we add this?  Once I implement "act like a user", then changing these things under a problem set may be a good way to go. 

Thoughts?


goehle:
- Mark all correct is used to mark that problem correct for all users.  Its generally "safer" to do that than delete a problem, since deleting a problem and then renumbering can cause issues.  So it should be something you can do when viewing the problems contained within a set.  

-  Acting as a user could alter the Problem Set Detail view to have user specific data which can be altered (like seeds).  That seems like a good way to do that since there are several ways to easily act as a user (including bringing up the user list in the sidebar) 

-  You should probably add some checks so that gateway and jitar sets are not displayed in the problem set detail view until its able to support them.  For jitar sets specifically the problem tree structure would probably be wiped out by the current Problem Set Detail.

mgage:
If problem instances have a unique id in the database which does not depend on which set they are in or their order of appearance in the set then it should be possible to delete a problem.  You can do that in the current webwork if you do not reorder the problems.  If you reorder the problems then problem 7 would refer to one problem before the deletion and to another problem after the deletion and reordering.  This causes confusion if some students do the problem before the change and others do it afterword.  Not having a separate unique id for each problem in a course was the biggest flaw in the current database scheme.

pstaabp:
So as a follow up
1. The mark all correct can be made in a problem in the ProblemSetDetail view.  This won't be difficult to implement.  Alternatively, we can explore the addition of the separating the problem id from the problem number (this seems like a good idea anyway moving forward). 
2.  My plan was to implement some "Act like a User" action.   This will alter the interface and make it look more like a student view.  I don't want to put that in this PR, but it is on the list of things to do. 
3. I'll add checks to make sure that JITAR and gateways aren't viewable (or better yet) or not destroyed in this PR.

goehle:
I've made a couple of attempts at getting unique id's into the tables.  Its not worth the pain its going to cause with table upgrades unless we also do some reshuffling of the tables to be closer to what we came up with in Ann Arbor.  (Basically it will require a script to go through and completely refactor the tables, moving "id" to "name" and recreating the row.)  This isn't something we want to do just to add ids.  However I hit a stang with the table refactoring because it turns out UserSet isn't a join of User and Set.  (There are versioned gateway sets in UserSet which have no corresponding set in Set.)  I haven't been able to come back to it since then.  

You won't be able to have it deal with JITAR properly until JITAR is in develop (I would assume anyway).  Makeing gateways feature complete should be easier, you mostly just have to add the missing parameters and add support for problem groups.  

Let me know when you have the safety checks in place and are ready for this to be looked at.  (I'm not messaged when you add commits, only when you post something.)

pstaabp:
What Gateway property are getting changed in ww3 that messes it up?  I was thinking about how to limit ww3 until I get Gateways working, but doing so will be more work than I think I want to do (since I'll pull it out later) and I don't think it will be too hard to add it anyway, and I'm thinking I might just do it here. 

(I haven't used gateways much, so I'm learning about them as I go here.)

Here's what's going on with homework sets.  It loads all of the properties from the database into ww3.  Since most of the gateway properties aren't touched, they should be stored on ww3 and anytime sent to the server, the values shouldn't change.  My first few attempts at changing a gateway is only messing with the parameters I changed.

goehle:
-  I don't think anything about gateways is broken by ww3.  (I was thinking specifically of jitar sets and the fact that the problem number for jitar sets is some huge bit mapped thing.)  However, when you visit a gateway it isn't clear that you are on a gateway (the set type is not shown), or that there are gateway properties that aren't being shown.  Also, it doesn't understand what to do with problem groups (.e.g group:set_name).  People should at least be warned that they aren't going to be able to properly configure the set in ww3.  

-  Another issue I randomly noticed while checking this is that the Problems portion of Problem Set Details doesn't deal with repeat problems properly.  If there are 4 copies of the same problem in a homework set it only shows on instance of that problem.

pstaabp:
* I think I can easily add the same form fields for gateways that we see in ww2.  I'll add those for this pull request. 

*  I'm trying to rethink two copies of the same problem.  The way I originally set this up is that you can't add another copy of the same problem source and I realize that this isn't the same way that ww2 handles it, so I know I need to change that.

goehle:
Its not common, but you can include multiple copies of the same problem and rely on seeds to make sure nobody has duplicates.  Its more common to have two copies of the same problem group in Gateways.

pstaabp:
I'm back to thinking about the issue of not allowing duplicate problems in a set.  The way that backbone handles data on the client side is that every thing (User, Set, Problem, etc) has an id with it.  I originally had the source_file (path) property of a problem be the id.  This works great in the Library, but not in Problem sets.  

Because for collection of things (like problems) when you add two things to the same collection, it won't add the second one (because it already has one in the set).  So I need to come up with another item to be the id.  

Logically it makes sense to use the problem_id of the problem for its id, but that comes back to the issue with that changes when reordering.  When you worked on the separation of is that you mentioned above, did you get to problems?  It seems that we could have another column in the course_problem and course_problem_user databases called number.  If it's an existing database, number is set equal to problem_id at first  and then all of the reordering is done on number.  

That would easily solve my problem of not allowing duplicates, and other advantages that Mike mentioned.  If you haven't tackled this, I would be willing to look at it, but could use a lot of help on the ww2 side of things.

goehle:
I did get to problem ID's.  The main issue is that you can't really just
leave existing databases intact.  You want id to be a unique,
auto-increment, int and you can't just change the existing column to that
type.  You have to strip the table apart and rebuild it.  I've made scripts
which do this, but they are a little fragile.  In some ways it takes us
back to the day's where database upgrades were dangerous.  Its worth it in
the long run, especially if it moves us forward in our vision for how the
backend should look.  Its not a small project though and it doesn't make
sense to me to have people go trough this intensive kind of db upgrade more
than once (or twice...).

The quick and dirty solution to your problem may be to use problem_id
anyway and have the same issues webwork 2 does (i.e. deleting and
reordering is dangerous).  Then when we do sort out the id situation it
will automatically be fixed.

Mike, you may have an opinion.  I could pick the database id project back
up, but it would mean cutting back on one of my existing projects.

mgage:
Hi,

I think Peter is running into the same issues that have led me to believe that at the very least every table 
should have a row id that means absolutely nothing else besides which row it describes.
Ideally we’d assign problem ids globally, not just for one site, similarly for other ids.  But I haven’t figured 
out how to do that in a reasonable way.  There are schemes for global ids out their, but the last time I looked
at them they didn’t seem feasible for our purposes. My own rule now for myself is that every unique id should have no meaning whatsoever aside from the fact that it is an id. I don’t know if that is overboard or not but it works for me.

My approach to the upgrade issue is to (for now) add columns to the table rather than to change any additional
columns.  So I would add something like row_id or uniq_id to each table, make it an auto-increment id and use
it where ever possible in new database joins and such like.  Sometimes one might have to have some redundant
mechanisms which use both the old and new id schemes simultaneously and if possible update references to use the new ids.  After a while the references to the new uniq_ids are all in place and one can slowly eliminate the use of the old ids.

A small example of this are the visible and publish columns.  I originally liked the name publish (it’s what the early blogs used) for describing when a homework set went live.  At the time no-one understood what that meant and it was decided that “visible” would be a better name.  The two column were side by side for a while duplicating information until we made sure that all references to published were removed from access codes and all information in “published" columns were also in the “visible” column.  At this point (5 or more years later) I’d be ready to automatically delete all “published” columns in tables for old courses.

The example above is much simpler than the problem you guys are talking about I think something somewhat similar can be done — add a parallel structure.  Use the new structure in new code.  As much as possible update the new data reference fields from old data reference fields as they are being used.  Eventually, after testing for consistency between the new and old data representations, eliminate the old data representation.

I’m driving back to Canton later today but I’ll be back on line tomorrow.

Take care,

Mike

goehle:
I'm not the biggest fan of this approach.  Partly because I don't like having a column named, in essense, `problem->id_for_reals_this_time`, but also because there is some danger that it turns out like `psvn`, which was supposed to eventually play the role of a unique ID for versioned sets, but now just hangs around not doing anything.  However, it is the quickest fix to Peter's problem, and does at least get some forward motion on the whole project.

pstaabp:
Why is the approach I listed above not valid?  That is:
1. create a new column, called number  (which is an int)
2. on upgradeDB, fill it with the current problem_id
3. wherever we used the problem_id as the problem number, use the "number" column instead.

Note: #3 seem like the difficult step (I just search problem_id in all of lib/ and it's around a lot).  Maybe this is the deal breaker. 

This leaves problem_id now as the id as we intended. For existing tables, there may problem_id's like 2,4,8,9  in the database (some integers missing), but I don't think that matters.  For new tables, there shouldn't be problems.

mgage:
Actually PSVN predates gateway quizzes by 5 or 6 years and was in the original Webwork1.0.
It was taken from the original CAPA approach which used a single seed for the entire homework set. (Moodle still does this.). We found it more useful to have a different seed for each problem — that way you could add a problem to a set without changing all of the random numbers generated by the other problems. I’m not sure exactly what use gateway quizzes tried to make of it — it had been used in an ad hoc way as a unique id for problem sets before gateways — but we still hadn’t really appreciated how important unique ids were at that time.

— Mike

goehle:
For one thing, we dont want to change every time we use problem->id, its
only the times when we were using problem->id as a problem number, and not
as an id.  Lots of the problem->id calls are exactly what they should be.
However, at the end of the day, your approach and mine are basically the
same. The extra step I take is to then regenerate the problem->id column so
that it is a proper autoindexed, unique id column.  However, you kind of
have to do this.  You can't leave it as is because there are duplicates and
future code will assume that the id's are unique.  So then we are back in
the situation where you have to rebuild the problem table row by row.
 (Although then you kind of have to do the same thing for user problem,
because its so closely tied with problem, and things kind of spiral
outwards.)


If you just want to add a column with the minimal amount of fuss, Mike's
suggestion is the better one.  It can be added, will appear automatically
on database upgrades, will have its entries autogenerated, and will be the
unique id you want in ww3.  I suppose at some unknown future time someone
would go through webwork2 and change all of the problem->id's to
problem->unique_id's wherever they should be changed.

pstaabp:
First, I realize that a lot of the problem->id (or problem->problem_id) calls are actually to get the problem_id  and not the problem number. 

I had assumed that problem_id is autoincrement, but I guess it isn't, so we will need a column that is autoincrement.  Therefore Mike's suggestion will be the easier way to go in that row_id would be the ID of the problem and problem_id will continue to be the number.   

If this doesn't sound too difficult, I would say we should do this as help moving forward and not worry about too much since thinking about moving to a new DB structure down the road anyway. 

On the ContentGenerator side of things, how much work would we need to do?  On the ww3 side, this will solve the problem on the client-side and the corresponding server (dancer) code is very self contained and not much will need to be done.

goehle:
P.S.  Its also good to note that it is a lot easier to change problem->id
to problem->number than it is to change problem->id to problem->unique_id.
The reason is that problem numbers are visual artifacts.  They are used in
display and ordering.  Nothing terrible happens if you accidentally use a
problem->id there, and its easy to see because you just look for id's where
problem numbers should be, or things that are ordered strangely.   The
other way around is more complicated because bad things can happen if you
use a problem number in place of a problem id, and the effects are not
immediatelly obvious.  (It would end up as a bug where mysteriously
problems in certain sets would get deleted or changed when you modified
other problems.)

goehle:
Either lots or nothing.  Lots if there is an effort made to actually go
through and change problem->problem_id calls to problem->row_id calls,
nothing if we just leave content generator as is and ignore the addition of
row_id entirely.

pstaabp:
My only concern is if I start using problem_id as a number in ww3, will that mess up anything?  I don't think much will else will change and I'm thinking that as long as the problem_id's continue to be unique, there shouldn't be any ill effects... but... just want to be sure.

goehle:
It might.  You will have to write ww3 so that it still treats problem_id as
an ID in the sense of ww2.  (Basically it will have to do all of the ww2
stuff, in addition to whatever happens to row_id) In particular it will
still be indexed and part of the unique row identifier for ww2 so things
like reordering problems will not be as simple as changing a field.  As you
write ww3 you will need to check back and see that it leaves problems
intact for ww2 to use.

I've been thinking about this a fair bit, and maybe this is a good time to
talk about it.  I don't honestly think the Ann Arbor vision of the db is
going to happen.  Its been however many years, the grant is ending, and the
fact is people want to work on other stuff.  I think what we can accomplish
is to redo the existing database so that it has proper unique ids, joins,
etc...  That would really clean up some parts of the code, it would fix
some long standing bugs and feature requests (like students persisting
outside of courses and renaming sets).  Its really not that big a project,
but I feel like we need to just go for it.  It can't take another 5 years,
because in 5 years we are going to be thinking about how to display webwork
with a vr headset or some other such thing.  Since it doesn't change how
webwork currently works we have a good baseline to test against.  It would
basically be adding unique ids, and some other stuff to the db, building a
conversion script, and then going through webwork to see if things still
work and to change the odd id to a name/number.

pstaabp:
Why don't I try to see if I can do this.  I'll make a separate branch only for the database upgrade and test ww2 first.

pstaabp:
I've been playing with changing the COURSE_problems and COURSE_problem_user to add a problem number to the database. If this will work, my plan will be to get it to work for ww2 and merge into develop, then pull into ww3.  

Here's a bit of a plan and wanted to see what you thought since I haven't messed around the DB at all.

1. update the COURSE_problem table in the following way.  It appears that adding commands to the DBupgrade.pm module is how we do this. 
a. rename the column problem_id as number
b. add a new column called problem_id which is an int(11) primary key auto_increment. 

Note:  doing this by hand seems to give a table as we mentioned above where problem_id is now the "row_id" 

2. Do the same to the COURSE_problem_user table. 

3. At this point we should have the same situation as now except the number links the two tables and not the problem_id anymore.  

4. Find throughout the code where a problem_id is really the problem_id and where it is a number.  


Let me know what you think and I can move ahead with such a endevour and where the potential pitfalls are.

goehle:
I thought the plan was to add a new column, called row_id, and then use
that.  Then, at some undefined future date, we would change calls from
problem_id to row_id.  It sounds like you are trying to do my plan of
remaking problem_id to be auto increment and creating a new number column
with values copied out of problem_id.  If we are doing that one then it
gets more complicated and is not a small thing.  I think its worth doing,
but all of the id stuff should be done all at once because the upgrade
process will be just as painful for all of them as one of them.

pstaabp:
ok.  Then I'll just add a row_id.  It didn't seem that bad, but you know from experience.  

If I just add the row_id to the `WeBWorK::DB::Record::Problem` it appears to work without having to add anything to the DBUpdate.pm module, but running the upgrade courses within  Is that true?

goehle:
Yes.  DBUpdate.pm is set up to automatically add new columns if they appear
in the Record files.  You might double check that mysql is automatically
filling out the autoincrement on existing rows.

mgage:
Where is DBUpdate being called from? I thought that was obsolete code -- used only when updates needed to be made externally outside of the admin course.

goehle:
Sorry, I just meant the database checking and updating code in general.
Specifically its the stuff in CouseIntegrityCheck.

mgage:
OK.  That makes sense.  I was afraid you were using an old module.

pstaabp:
Although I thought just adding the row_id would be innocuous, I'm getting a record exists error on reordering problems.  It's happening at line 885 of ProblemSetDetail.  Isn't a putGlobalProblem an update?  As I understand it, the problem_id of the problem is being changed here and the updated, so yes the record does exist.

goehle:
It is difficult to say without seeing your error or how you added row_id.
Just adding row_id should not have affected anything at all, in theory.
However, something to keep in mind is that together the problem_id and the
set_id form the "key" of the problem row.  Keys have to be unique so mysql
will complain if there are ever two problems in the same set with the same
number (even if you are only temporarily allowing them to be the same while
you reorder).

pstaabp:
It's bombing right when the problem_id is being updated to a new value and I think that is the issue.  

I made the row_id the primary key and auto_increment and looking at the table in the db it looks like it was made correctly. Should I keep set_id and problem_id keys as well.  Maybe I can relax that condition now that we have a row_id.

goehle:
You shouldn't change anything about keys.  Those determine what values are
used to identify the row and what values are used to create the table
index.  In this setup row_id isn't even a key, since we aren't writing any
accessors that search up the row using row_id.  Its just an identifier.

pstaabp:
Then row_id should only be an auto_increment int ?

goehle:
Yes, its basically just a hack for stickit, and doesn't do anything other
than sit around and be a number.

pstaabp:
It appears that in mysql an auto_increment number needs to be a key.    It also seems like the row_id ought to be a key (perhaps the primary key).

goehle:
Then I don't know.  I dont think this piecemeal approach is a good idea if
it requires us to start messing with ww2.  We don't have the man power to
make a bunch of changes to ww2 and then go back later on and redo those
changes so they are correct.  Maybe Mike has better advice.  Given time and
permission I would recommend refactoring the database so that all the
tables have unique autoincrement id's which are actually used as id's in
the code, and give the tables a better relational structure.  (E.G.  The
unique row_id of the problem table is what shows up as problem_id in the
user_problem table, etc...)

pstaabp:
First, i guess making the row_id auto increment and therefore a key messes up the keys of the table. I tried to retroactively make row_id problem_id and set_id a collective key (which is what set_id and problem_id is now) but am unsuccessful so far. 

I think this is turning out to be more trouble than it's worth and have another idea about adding copies of the same problem which was the original goal in this discussion.

pstaabp:
Finally have an update for this pull request.   To get multiple problems to be added to a single set there was some backend work that I overhauled.  Therefore I would do some additional testing through the Problem Set Details. (A list of thing is at: https://github.com/openwebwork/webwork2/wiki/Testing-Webwork3).  

The new features and those that need testing:

1. the ability for the maxAttempts for a problem to be changed.  There a validator for this as well, so test some wrong inputs. 
2. the value (weight) of a problem to be validated 
3. A new tab in the ProblemSetDetails view that has a way to add the description and set the headers. issue #316 
4. Reworking the Student Overrides tab to make it clearer how to change the dates for multiple students at once.  In addition, a student override that is not the same value as the global value is displayed in blue.  This clears up issue #286
5. There's now a button (a check) on an individual problem in the "Problems" tab to give everyone full credit on the problem.   (Note: I did not include the ability to change invidual scores and problem seeds in this pull).
6. A set can be changed to a Gateway (I still need to add proctored gateway and eventually JITAR) showing the additional attributes for a Gateway Quiz.
7. There is a calendar in the Details tab of a Problem Set showing only the single problem set, but allowing a user to change the dates this way. 
8. When reordering a problem set, this is disabled if the set is open.

pstaabp:
Hold off for a bit if anyone is looking at this.  After merging in ww3, that introduced some errors.  I'll post a fix soon.

pstaabp:
Made some fixes to the messaging when problems are added/deleted.  This is ready to be looked at now.

pstaabp:
Functionality Question:  I'm working on improving the library.  The Library Options (similar to the Problem Set Options) has a display mode.  Should this change the display_mode of the user settings (i.e. update the database?) or just temporarily change the display_mode?  I think as instructor, it should be a temporary change.

goehle:
I would have it be temporary.

goehle:
- [ ]  You are running dancer directly, which is why you don't see this problem, but in app.pl the inclusion of the WeBWorK packages have to come after the inclusion of the webwork2 and pg paths.  E.G.  https://github.com/openwebwork/webwork2/pull/598/files#diff-7e8d06d30c6237885f658f8c724b1651L9  Otherwise the fcgi path fails when it tries to load CourseEnvironment. 
- [ ]  It looks like you added a couple of perl packages Data::Dump and Data::Compare.  If this is true you should add those to check_modules for everyone's sanity. 
- [ ]  Starting up the server I get:
```
values on reference is experimental at /opt/webwork/webwork2-vanilla/webwork3/lib/Utils/ProblemSets.pm line 78.
keys on reference is experimental at /opt/webwork/webwork2-vanilla/webwork3/lib/Utils/ProblemSets.pm line 102.
Smartmatch is experimental at /opt/webwork/webwork2-vanilla/webwork3/lib/Routes/ProblemSets.pm line 939.
Smartmatch is experimental at /opt/webwork/webwork2-vanilla/webwork3/lib/Routes/ProblemSets.pm line 939.
Smartmatch is experimental at /opt/webwork/webwork2-vanilla/webwork3/lib/Routes/ProblemSets.pm line 943.
```
You should just check and make sure these are things you are ok with.

pstaabp:
1. About the order of loading packages,  I switched these because I was getting an error that wouldn't let me log in.   A little background:

I've been trying to use the ww2 version of packages whenever possible because I don't need to duplicate code and then trying to debug can be a nightmare.  However, there are problems when the ww2 package uses mod_perl stuff.  For example, the Authen.pm package (this is where I ran into a problem).  The call to `maketext` is on the Request object, which isn't part of dancer, so I need to drop that (and we'll try to include it another way).  

I have a version of Authen.pm in webwork3 that drops all such code that gives errors on dancer.  I decided to call it the same:  i.e. package WeBWorK::Authen with that thought that it would grab the ww3 version if it exists and if not use the ww2 version.  

I was running into problems with logging in because it was finding the ww2 version.  I thought that switching the order would fix this.  I guess it does on the standalone dancer version, but not on the fcgi version. 

I guess a fix to this is to call it something different and in the dancer routes specifically call the webwork3 versions.  

2. Data::Compare is necessary now.  Data::Dump is just for debugging, so I can get rid of that and add it in as needed.  I'll add Data::Compare to the check_modules.  

3. I had seen the Smartmatch is experimental  warnings, but not the others.  I was wondering if we're okay with those.  I can rewrite them easily to not use the smartmatching  (I think it was a snippet of code I got from the interweb a while ago).

goehle:
-  It fixes it because in the standalone version you have the paths added to your PerlLib environment variable in such a way that webwork3 comes later (or before?) webwork2.  I.E.  It loads the webwork3 version.  I think a better version would be to *not* have the packages named the same but rather to have your webwork3 authen package be a subclass of the webwork2 authen package, overriding things where necessary.  You don't want to have your server only work if things are loaded in a particular order. 
-  I would not use it if you don't have to.  The warnings make sysadmins nervous.

goehle:
A couple more issues
- [ ]  I'm getting wierd behavior with gateway sets.  If I change one set to gateway, then switch to another and change it to gateway I get too many changed notifications and the original set is changed back
```
 The value of assignment_type in problem set testtable has changed from to .
The value of assignment_type in problem set Orientable has changed from to .
The value of assignment_type in problem set testtable has changed from to .
The value of assignment_type in problem set testtable has changed from to . 
```
Also, if I change gateway values I get wierd messages and it doesn't change the value in the right set.  So if I change the number of tries variable for *testtable* I get
```
 The value of version_time_limit in problem set Orientable has changed from 2 to 2.
The value of version_time_limit in problem set Orientable has changed from 2 to 2. 
```
So its not the right set or the right variable. 
- [ ]  I've got my server set to New Zealand time to test more extreme timezones.  If I set dates in ww3 to 08/10/2014 I get 08/11/2014 at 07:24am.  My guess is there is something going wrong with the timezone conversions.  
- [ ]  Max attempts is rendering as -1 when it should render as unlimited.  Max attempts is rendering on gateway sets when it shouldn't render at all. 
- [ ]  Semi-feature request:  The show me another options aren't shown.  Jitar sets will add even more problem options.  So thinking about how to deal with lots of problem options that may or may not be shown in specific circumstances is something to think about. 
- [ ]  The header files don't respect the defaults.  That is, when my header file parameters are blank ww3 is showing "choose a header file" but what is actually going on is that there is a "defaultHeader" defined in the config files.  It should show that the default file is selected. 
- [ ]  The view and edit buttons dont work for header files. 
- [ ]  Mark problem correct for all users doesn't seem to work. 
- [ ]  If I have a set with assigned users, no reduced scoring date, and I click enable reduced scoring date the reduced scoring date column appears in the student overrides file but doesn't have any editable fields.  This is true even if I start changing the global reduced credit date.  You should also test out that its easyish to enable reduced scoring for a single student since thats a common use case. 
- [ ] Semi feature request: The message for when a source file does not exist is kind of scary.  It could be friendlier because its something thats likely to happen to an average user.  Currently its 
```
WeBWorK::Utils::readFile(//problembanks/gateways/precalentr/topic_substitution/prob2.pg) says: failed to read file //problembanks/gateways/precalentr/topic_substitution/prob2.pg: No such file or directory at /opt/webwork/webwork2/lib/WeBWorK/PG/Local.pm line 305. 
```
- [ ]  Semi feature request: Problem groups (e.g. group:setname in the problem path) do not seem to be supported.  They don't display if they are there and there doesn't seem to be a way to enter them. 
- [ ]  Semi feature request:  Gateways seem to be supported but proctored gateways don't seem to be supported or recognized at all.

pstaabp:
About the view/edit buttons for header files, can you elaborate?  The Edit buttons should show the content of the header files (the pg source), and the view buttons should show the rendered versions of these.  My version is working.

pstaabp:
Also, about the timezone issue.  The ww3 client isn't storing anything about the timezone on a per problem set basis.  It gets the date/time from the server and passes it back and forth as the unix integer time.   When you said above that you get " 08/11/2014 at 07:24am"  does this mean in ww2 or in the database?

goehle:
-  ww2 takes the unix integer time modified by the timezone from the course environment.  See https://github.com/openwebwork/webwork2/blob/develop/lib/WeBWorK/ContentGenerator.pm#L2161  So when I say its "08/11/2014 at 07:24am" I mean that the unix timestamp is getting converted to that time when the NZST timezone is applied.
-  When I try to view a set header or hardcopy header I get a very thin empty box under the buttons.  When I try to edit I get a large empty box.  There aren't any js errors when I click the button.  However when I load the page I get 500 server errors from jquery which have this
```
   "exception" : "Warning caught during route execution: Use of uninitialized value in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/PG/Local.pm line 166.\n",
   "error" : "Warning caught during route execution: Use of uninitialized value in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/PG/Local.pm line 166.\n"
```

goehle:
P.S.  When I try to change either header I get the error
```
POST http://webwork.wcu.edu/webwork3/courses/TestCourse/sets/substitution/setheader 404 (Not Found)
POST http://webwork.wcu.edu/webwork3/courses/TestCourse/sets/substitution/setheader 404 (Not Found)
```

pstaabp:
* About the time zone.  I think this is an interpretation issue.  It seems if we use the time zone as defined on the server (that is, the one defined in course configuration), then the client should automatically be in that same time zone.  If that is true, then is there an issue?

If we are thinking that the client and the server are in different time zones, then I need to rethink this (and maybe we need to have a user setting with a time zone property). 

* I am seeing the error about the "warning during route execution, however I'm getting the rendered version of the header file and not seeing a POST error.  Are you getting the rendered version or the source of the headers?

goehle:
-  That sounds right to me, but I don't think ww3 is displaying the time
according to the time zone set on the server.  Or at least its not
displaying the same time as ww2 and ww2 does display the time according to
the time zone on the server.
-  I'm not seeing either the rendered or the source of the headers.

goehle:
In particular, if I change the time zone in the course configuration and
then refresh ww3 I don't see changed times on due dates in problem set
detail.  (I dont think I noticed before because times are not shown by
default.)   The times should change because the linux time on the server is
the same but the time zone its being displayed in has changed.

pstaabp:
I'm actually not doing anything with the timezone, so good catch.  

On the headers, can you look in the "network" section of the browser information window.  When going to the tab you should see two GETs  (http://localhost/webwork3/courses/test/headers) which load the set header and hardcopy header and then two other GETs (http://localhost/webwork3/courses/test/sets/XXXXX/setheader )   to load the source of the two headers.  Are you getting any of these?

goehle:
Those are the requests that are returning to 500 server error with 
```
"exception" : "Warning caught during route execution: Use of uninitialized value in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/PG/Local.pm line 166.\n",
   "error" : "Warning caught during route execution: Use of uninitialized value in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/PG/Local.pm line 166.\n"
```

pstaabp:
Just posted a fix to those errors for the headers.

pstaabp:
I want to clarify what is happening with datetimes on ww2.  I appears that datetimes in the db are stored in UTC.  Whenever reading from the DB and sending back to the client, we shift to the timezone defined in the course environment.  

Then parsing is the opposite.  That is, parse the time sent from the client in the CE timezone and shift to UTC before writing to the db.  

Is this correct?  

I'm not going to directly use the subroutines in WeBWorK::Utils because I don't need all of the overhead of parsing/writing in human readable strings.  That is all done on the client side in ww3.

goehle:
That is correct.  Before you do everything on the client side you might look and see how robust timezone support is for js.  Time zones are extremely complicated.  Officially they are designated by things like "America/Los_Angeles" or "Pacific/Auckland".  However there are a lot of semi official abbreviations like MST or NZDT which also take daylight savings time into account.  Unfortunatelly you have to use these for human readable stuff because the official designations are too long and things like UTC-8 are too confusing.  Everything in Utils is based off the perl module DateTime which is reasonably robust for this kind of thing.  Last I looked (and it was a year ago so things may have changed) there wasn't anything quite as good on the js side.

pstaabp:
There's a great js library called momentjs (http://momentjs.com/) which handles everything.  It has a timezone component of it, which I haven't yet used mainly because if you download all timezone stuff it's huge.   The other great part of it will be time localization when we get to that and will handle all of the mm/dd/yyyy versus dd/mm/yyyy as well as 24 hour time, etc.  

I also know about the nonstandard EST versus timezone names.  I planned on going back to the server to handle this stuff as needed.

pstaabp:
I'm still wrestling with the timezone stuff.  The past couple of days, I've integrated some tests to help track down errors, however, I think I'm still not sure what ww2 does.  I thought that (see comment above) that the datetime number in the database is the datetime in UTC, however, I don't think that is true now.  It seems like it has to do with the parameter $locale as you mentioned in https://github.com/openwebwork/webwork2/blob/develop/lib/WeBWorK/ContentGenerator.pm#L2161, however, my server doesn't have ce->{siteDefaults}{locale} defined and it doesn't seem to be playing a role. It's seems that in site.conf, there's the line:

```
# If you do not fill this in, the system will default to "en_US"
$siteDefaults{locale}="";
```
From the DateTime help page, this seems to only affect the way that dates/times are printed.  

What ww2 seems to do is take the datetimes in EDT (I guess since that's where I'm sitting). If I change to America/Chicago, I get datetimes one hour earlier.  

Is this what's going on?

goehle:
I've run some tests and I think the time stored in the database is UTC epoch, i.e. unix time.  In particular if I save the date `06/23/2015 at 05:00am EDT` then the value stored in the database is `1435050000`.  If I translate that at here: http://www.epochconverter.com/ I get the expected time. 

If I change my time zone to America/Los_Angeles in the config menu then saving the time `06/23/2015 at 05:00am` gives the database time `1435060800`.  (Note that I did not provide a timezone so its using the webwork server timezone from simple.conf).  When translated at the above website that is coming up as Tue 23 Jun 2015 08:00:00 AM EDT, which is the same as 5:00AM PDT.  

So as far as I can tell the database is getting a valid unix time which corresponds to either the displayed or provided time in the timezone set by WeBWorK.  If your js is going to pass the unix time stamps back and forth it will need to, for instance, take the provided time and create a unix time which corresponds to the provided time in the course timezone.  

I don't think `$locale` enters into it at all.  That just controls how the times and dates are formatted.

pstaabp:
I think I figured out what was going.  Leaving everything as epochs is the way to go and I don't need to do any conversion at all on the server, which is what I just started doing, so I need to roll back some of the changes I just made.  The client side moment.js converts from epochs to local time automatically, except that it uses the time zone of the machine you're on.  

There were two reasons why you didn't see any changes when switching the time zone.  
1. When writing the timezone to the simple.conf file, there was an error such that rereading the simple.conf file didn't accept the changes (and that was read when a server hit took place).  That's fixed now.

2. If the date/times are epochs on the ww3 side and then printed out using momentjs, then it will always take the timezone of the machine to do the conversion and the webwork timezone isn't used.  I think this is what you want to be able to change.  (Although one option is not to and just allow it to work on a local machine.  I would argue for this at least right now, although will not correspond to ww2.).  

If we do go ahead with this, I can use the moment-timezone js package:  http://momentjs.com/timezone

goehle:
I see, so its using the timezone of my browser and not the timezone set in the course.  On the one hand there is a nice elegance in having the time be whatever it is on the server and then it will "just work" when you view it from a browser. 

On the other hand its a bit mysterious and takes control away from the professor.  Its going to be *super* confusing if its different from ww2.  E.G. If you have a timezone on your browser thats not the same as the course timezone. 

I'm torn.  Blackboard displays everything according to the time zone set for the course.  Moodle does that as well, but it also has a user setting (also stored on the server) which can override the course timezone.  If it works well it will seem like magic.  If it doesn't work because of some use case I'm not thinking of then it will seem like broken magic.

pstaabp:
Let me put it on my todo list to add the timezone info to the client, but let it slip for this pull release.

pstaabp:
Couple of things here: 

1. It appears that the only proctored gateway parameters that differ from the gateway ones is that of a password.  Is this correct?   Also, it seems that the password is set as the user "set_id:SETNAME" and the code in the block at https://github.com/openwebwork/webwork2/blob/develop/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm#L1324   is needed.  (I realize it much different on the user side). 

2. I put in a bit of fix for the set headers.  Can you let me know if you're still having problems with this?

3. I guess I've never used problem groups that you mentioned above.  Can you point me to something to help me implement this?

goehle:
Sorry, I've been busy with the installer.

1. I haven't really used proctored gateways, but the main thing is the
password, as you noticed, and the fact that it has to add a "proctor user"
to the class in order for proctoring to work.   You will also need to edit
your classlist editor to exclude these "set proctor users".

2.  I'll check it when I can.

3.  http://webwork.maa.org/wiki/Selecting_Problems_from_Problem_Groups is a
good description.  Basically you make a set and you add group:setname to
the "path" portion of a problem.  Kind of hacky but useful.

goehle:
I've tested the header stuff out.  It still doesn't work for me.  Now I get 
```
exception:"Can't use an undefined value as an ARRAY reference at /opt/webwork/webwork2/webwork3/lib/WeBWorK3/PG/Local.pm line 94.
"
``` 
as a response whenever I try to use it.

pstaabp:
Just updated the code for the set header error.  I hope this fixes it.  

I'm running through a bunch of tests for the Problem Set Details view right now and will let you know when I feel it's ready.  As you can probably see from the # of commits and changes that I overhauled lots of the code.  I was having a hell of a time with reordering and deleting problems again, but the code is much simpler on both the client and server side, which I think is good.

pstaabp:
I've done some more thorough testing and this pull request ready to look at again.

goehle:
- [ ] The edit and view of  the hardcopy and set headers work now, but the Hardcopy header view renders the the header in HTML mode.  This is particularly bad for combined headers which render very differently in HTML and TeX mode. 
- [ ] The proctor password doesn't work right.  If it is set then it should just ask for the proctor password when you generate a gateway quiz.  Right now when it is set it still asks for a proctor user.  It looks like it is creating the proctor user, but if I set the password in ww3 the field still comes up blank in ww2.  
- [ ] I know it works this way in ww2, but maybe the proctor password shouldn't turn into *'s or dots.  Its supposed to be a shared password and masking it doesn't really help anything. 
- [ ]  I still dont know how to set a problem group using ww3.  On a related note, I don't know how to manually set a problem path using ww3.  
- [ ] Mark correct for all students still doesn't seem to work.  
- [ ] The "Select a Header File..." text on the drop down should probably be "Using Default Header".
- [ ] I'm having trouble with true false options again.  If a true false value is false and I change it to true it automatically changes back.  
- [ ] Deleting a problem in ww3 automatically renumbers the problems.  This is very dangerous if done after a set is open because answers and grades will get mixed up between problems.  
- [ ] Changing a global due date doesn't seem to change the corresponding date in the student overrides table.  That is if I load up Student Overrides change due date in Set Details and go back to Student Overrides the due dates have not changed.

pstaabp:
* hardcopy header: maybe we shouldn't have the view option.  Or alternatively, can we generate the PDF of the header. 
* When I create a proctored gateway quiz with a password, if I'm an instructor I get a proctor user/pass.  If I logout and login as a student, I just get the proctored password.   Again, I'm not exactly sure what I should see.  

I do see that going to the problem set detail shows no password.  I'll look into that. 
* How should we display the proctored password?  I was mimicking the ww2 way.  If we want to see the password in plaintext, then we probably just should store it in the DB as  plain text because decrypting the password isn't trivial. 
* To include a problem in a group, if you go to Problem Set details View and the "Problem List Options" sidebar, there is a new way to add a problem from a group. 
* Which true/false options are you having trouble with?  I thought I have them all passing through the same code to handle the javascript true/false versus the perl version, but I may have missed something. 
* About deleting problems, maybe we should not allow deletions if the set is open.  This is similar to the reordering problem.  
* I see the other issues you mentioned and will work on those.

goehle:
- I would not have a view for the hardcopy header, or would have a link to the hardcopy generator for the set (which is what ww2 does.)
-  The behaviour should be the same for professors and students.  When there is a proctor password you should just be asked for the proctor password, not a proctor user login.  
-  Hmm, good question.  For consistency the password should be stored in the same way as all of the other passwords.  I guess I would leave it the way it is.  
-  Thanks for pointing me to the sidebar. 
-  I'm having  trouble with pretty much all true/false type options.  
-  I've thought about that.  I've got similar issues with problem set detail 2.  If you delete something there it doesn't automatically reorder, but if you reorder any problems at all it does make things consecutive and you can't preserve the "gap".  I think the right thing to do would be to allow deletes, and preserve the skipped number.  Maybe you could add a warning that deleting after the open date is dangerous.  If/when we ever get around to redoing the database ID stuff this wouldn't be an issue, and I guess I would prefer to aim for where we are going with the UI.

pstaabp:
* I'm trying to test the problem groups and it appears that if I put in `group:set_id` in the filepath that I'm getting in the ProblemSetDetail page that it is saying: "Problem source is drawn from a grouping set", however if I view the problem set (as a student), I'm getting readFile errors.  

```
WeBWorK::Utils::readFile(/Library/WebServer/Documents/webwork/courses/test/templates/group:HW1) says: failed to read file /Library/WebServer/Documents/webwork/courses/test/templates/group:HW1: No such file or directory at /Library/Webserver/Documents/webwork/webwork2/lib/WeBWorK/PG/Local.pm line 305.
```

* Also, just an FYI, but in ww2, you are allowed to view the Hardcopy header as HTML in the ProblemSetDetail page.   If you change the display mode above the header info, it will show the HTML.  

* Are you getting any errors (js or from dancer) on the saving of the true/false?  I can't replicate that.

pstaabp:
About editing the set headers, in ww2, if you try to edit the default headers, it edits the simpleCombinedHeader.  Is this correct?  I tried to edit it in ww2, but am getting a number of errors and a strange path 
```
Failed to create directory /Library/WebServer/Documents/webwork/courses/test/templates/Library/Webserver/
```
which is nesting the directory twice.  

In ww3, should the default header just automatically edit simpleCombinedHeader?  My thought was once a user wants to edit a header file that he/she should select one to edit and therefore cannot edit the "defaultHeader".  That's what's going on now.

goehle:
The default header behaviour is a little strange.  Right now things like ProblemSetList2 set the headers of newly created sets to the string "defaultHeader".  Then things like ProblemSet check for the string 'defaultHeader' and use files defined in defaults.conf 

https://github.com/openwebwork/webwork2/blob/master/conf/defaults.config#L351
https://github.com/openwebwork/webwork2/blob/master/lib/WeBWorK/ContentGenerator/ProblemSet.pm#L233

In particular administrators can set the default header file for their universities.  This behaviour should be supported, although I dont know how you want to implement things.  Maybe the string is defaultHeader, but if you click view or edit it checks the course environment to see what the file should be.

goehle:
How are you viewing the problem set as a student?  Problem groups are only supported for gateways.  What happens is when a student generates a new version of the gateway the problem group is replaced with a random problem from that group.  You would get an error like the one you are describing if you viewed a problem:group in a default/homework type set.

goehle:
-  You are right about Problem Set Detail and viewing a Hardcopy Header as HTML.  By random chance you cant do it in Problem Set Detail 2. 

-  Whenever I try to turn on any checkbox it immediately turns on and then off.  Checking the checkbox generates two requests.  The first is 
```
_id: "testcurrency"
answer_date: "1413733380"
assigned_users: ["grgoehle@email.wcu.edu",…]
assignment_type: "default"
attempts_per_version: "0"
description: ""
due_date: "1413733380"
enable_reduced_scoring: false
hardcopy_header: "defaultHeader"
hide_hint: false
hide_score: ""
hide_score_by_problem: ""
hide_work: ""
open_date: "1412782980"
pg_password: ""
problem_randorder: false
problems: [{showMeAnother: "-1", source_file: "local/settestcurrency/bsssslankProblem.pg", value: "1",…}]
problems_per_page: "0"
reduced_scoring_date: "1413560580"
relax_restrict_ip: "No"
restrict_ip: "No"
restricted_login_proctor: ""
restricted_release: ""
restricted_status: "1"
set_header: "defaultHeader"
set_id: "testcurrency"
time_interval: "0"
time_limit_cap: false
version_creation_time: ""
version_last_attempt_time: ""
version_time_limit: "0"
versions_per_interval: "0"
visible: false
```
with response
```
   "hide_score_by_problem" : "",
   "attempts_per_version" : "0",
   "hardcopy_header" : "defaultHeader",
   "hide_hint" : false,
   "version_creation_time" : "",
   "set_id" : "testcurrency",
   "version_time_limit" : "0",
   "versions_per_interval" : "0",
   "restricted_status" : "1",
   "restricted_release" : "",
   "pg_password" : "",
   "open_date" : "1412782980",
   "problems_per_page" : "0",
   "answer_date" : "1413733380",
   "hide_work" : "",
   "time_interval" : "0",
   "problem_randorder" : false,
   "visible" : false,
   "version_last_attempt_time" : "",
   "reduced_scoring_date" : "1413560580",
   "assigned_users" : [
      "goehle@gmail.com",
   ],
   "description" : "",
   "problems" : [
      {
         "showMeAnother" : "-1",
         "source_file" : "local/settestcurrency/bsssslankProblem.pg",
         "value" : "1",
         "problem_id" : "1",
         "_id" : "testcurrency:1",
         "flags" : "",
         "max_attempts" : "-1",
         "data" : "",
         "problem_seed" : 1,
         "set_id" : "testcurrency",
         "showMeAnotherCount" : "0"
      }
   ],
   "relax_restrict_ip" : "No",
   "enable_reduced_scoring" : false,
   "due_date" : "1413733380",
   "restricted_login_proctor" : "",
   "restrict_ip" : "No",
   "time_limit_cap" : false,
   "assignment_type" : "default",
   "set_header" : "defaultHeader",
   "hide_score" : "",
   "_id" : "testcurrency"
```
The *second* request is 
``_id: "testcurrency"
answer_date: "1413733380"
assigned_users: ["grgoehle@email.wcu.edu",…]
assignment_type: "default"
attempts_per_version: "0"
description: ""
due_date: "1413733380"
enable_reduced_scoring: false
hardcopy_header: "defaultHeader"
hide_hint: false
hide_score: ""
hide_score_by_problem: ""
hide_work: ""
open_date: "1412782980"
pg_password: ""
problem_randorder: false
problems: [{showMeAnother: "-1", problem_id: "1", flags: "", value: "1", _id: "testcurrency:1",…}]
problems_per_page: "0"
reduced_scoring_date: "1413560580"
relax_restrict_ip: "No"
restrict_ip: "No"
restricted_login_proctor: ""
restricted_release: ""
restricted_status: "1"
set_header: "defaultHeader"
set_id: "testcurrency"
time_interval: "0"
time_limit_cap: false
version_creation_time: ""
version_last_attempt_time: ""
version_time_limit: "0"
versions_per_interval: "0"
visible: false`
```
with response
```
   "due_date" : "1413733380",
   "enable_reduced_scoring" : false,
   "relax_restrict_ip" : "No",
   "restrict_ip" : "No",
   "time_limit_cap" : false,
   "restricted_login_proctor" : "",
   "hide_score" : "",
   "_id" : "testcurrency",
   "set_header" : "defaultHeader",
   "assignment_type" : "default",
   "time_interval" : "0",
   "visible" : false,
   "problem_randorder" : false,
   "assigned_users" : [
      "goehle@gmail.com",
   ],
   "version_last_attempt_time" : "",
   "reduced_scoring_date" : "1413560580",
   "problems" : [
      {
         "showMeAnother" : "-1",
         "problem_id" : "1",
         "flags" : "",
         "value" : "1",
         "_id" : "testcurrency:1",
         "source_file" : "local/settestcurrency/bsssslankProblem.pg",
         "max_attempts" : "-1",
         "showMeAnotherCount" : "0",
         "data" : "",
         "problem_seed" : 1,
         "set_id" : "testcurrency"
      }
   ],
   "description" : "",
   "version_creation_time" : "",
   "restricted_release" : "",
   "set_id" : "testcurrency",
   "restricted_status" : "1",
   "version_time_limit" : "0",
   "versions_per_interval" : "0",
   "problems_per_page" : "0",
   "pg_password" : "",
   "open_date" : "1412782980",
   "answer_date" : "1413733380",
   "hide_work" : "",
   "hide_score_by_problem" : "",
   "attempts_per_version" : "0",
   "hide_hint" : false,
   "hardcopy_header" : "defaultHeader"
```

**Comment by pstaabp** [comment]
I'm trying to come up with a reason why you're installation isn't handling the booleans.  I just did more tests on another machine and it's working there.  A few other ideas:  
1. what version of dancer and perl are you running?
2. What boolean did you change above to get the requests? 

The fields that I have as booleans for a problem set are: visible enable_reduced_scoring hide_hint time_limit_cap problem_randorder It seems that your http request has everything false.  Just to confirm, you clicked one of the checkboxes and these are the requests/responses you received?  And then the checkbox went back to unchecked.

**Comment by pstaabp** [comment]
About the set headers:  I'm reading the field of the ce variable to get the default header for both screen and hardcopy.  I guess the question is on the UI side: if the "Use the default header" and we want to edit the file, should they just go ahead an edit the default one? 

As you can see, I was going to have the user select the one to use, but that may be an extra step.

**Comment by goehle** [comment]
1.  I'm using Dancer 1.3140 and Perl 5.20.2
2.  It happens with all booleans, both on the set detail page and on other pages.  Although something that might be a clue, if I have a set detail page with multiple checkboxes all checked and I uncheck one of the boxes, then they all get unchecked.  

P.S.  If I look at a "jitar" type set I see "Homework" in the drop down but it also has Proctored Gateway/Quiz parameters.  Maybe it should just fail gracefully for jitar sets?

**Comment by goehle** [comment]
Probably not.  The standard default header files are in webwork2/conf/snippet and the webserver doesn't have write permissions there.  They should have to choose a header in the templates folder before they can  edit it.

**Comment by pstaabp** [comment]
So on the set header:  is the current UI sufficient.  It requires the user to pick a set header. Alternatively, If they start to edit, I can switch to the version in the templates directory (if it exists). 

On JITAR, ideas of failing gracefully:  Is preventing reordering of problems sufficient?  I would think the remainder of the problem set details would apply.

**Comment by goehle** [comment]
-  I think "Select a Header File..." is misleading.  It makes it seem like there isn't a header file.  All I wanted was to change that to "Using Default Header" or something similar.  In particular you should be able to be on "defaultHeader", select a different header, and then have a clear way to get back to the defaultHeader. 

-  jitar has additional problem set parameters, as well as additional parameters for each problem.  What's more, the problem parameters should be shown/hidden depending on if the problem is a subproblem/has subproblems.  I think a nice message saying "this homework type isn't supported"  is probably better than something that half works.

**Comment by pstaabp** [comment]
I think I have a fix for the booleans on the problem sets.  Are boolean working in other places?  Settings and UserSettings are the only locations I see?

**Comment by goehle** [comment]
Booleans seem to work now.  

On a side note, how are you supposed to refresh the header view.  When I change set headers it doesn't change the rendered header, and switching back and forth between view/edit or different sets doesn't do anything.  I have similar difficulties with the edit headers views.

**Comment by pstaabp** [comment]
If you edit the set header and leave the textarea, then it automatically saves the file and opens the "View Set Header" tab with the update.  

I've taken out the view hardcopy header as we mentioned above, so there isn't a way to view that right now.

**Comment by pstaabp** [comment]
Whenever you get a chance this is ready to test again unless the set header interface needs more work.

**Comment by goehle** [comment]
I'll test it when I can.  My classes have started so my available time has gone down considerably.

**Comment by pstaabp** [comment]
I understand.  I'm ramping up right now.  I wasn't sure if you were waiting from me.

**Comment by goehle** [comment]
Did you change something about how things set up with ww3?  I'm having trouble getting it running via the fcgi pathway.  The issue seems to be that the WEBWORK_ROOT environment variable is getting lost somehow.

**Comment by goehle** [comment]
I had to add the lines 
```
BEGIN{
  $ENV{WEBWORK_ROOT} = config->{webwork_dir};
}
```
to WeBWorK3.pm before any of the WeBWorK includes in order to get the fcgi path to work.  I dont know why it used to work or what has changed.

**Comment by goehle** [comment]
I'm also getting this error again 
```
Uncaught TypeError: Cannot read property 'label' of undefined
```
when I load up the page.  Last time it was caused by there being nonstandard permission level stuff.  I got this error by changing a user to the TA permission level using the ww2 classlist editor.

**Comment by goehle** [comment]
After getting ww3 up and running: 
- [ ] Switching between different headers (hardcopy or set) doesn't change the content of the view/edit box.  If I switch headers I get the message saying that the header was saved, but the view/edit box only shows the content of the headers from when the page was loaded. 
- [ ] If I go to Set Details, then go to Student Overrides, then go to Set Details and change a global date, when I go back to Student Overrides all of the students have the previous date as an override.  This is almost certainly not the intended behaviour.  If a global date is changed and a student date matches the global date, it should also be changed. 

One subtle issue is when you mark correct for all students it also sets student overrides for every problem parameter for that student.  Its not the worst but it means that future global changes to things like weight will not be reflected in the problem.  The issue may be getting a "mergedUserProblem" from the DB and then putting that into the database with "putUserProblem".  You should really use "getUserProblem" because it will have null fields for things which are not overridden.  The same is true for sets.   Keep an eye out and make sure you aren't using "putUserSet" with a set object generated by "getMergedSet".

**Comment by goehle** [comment]
I'm getting errors that look like 
```
User 'set_id:Orientable' has no PermissionLevel record -- assuming no permission. at /opt/webwork/webwork2/lib/WeBWorK/Authz.pm line 247.
User 'set_id:Algebra_Practice' has no PermissionLevel record -- assuming no permission. at /opt/webwork/webwork2/lib/WeBWorK/Authz.pm line 247.
```
when I visit the problem page.  Those "proctor" users were added by WeBWorK3.  I think its not adding a permission record (with permission set to proctor)

**Comment by pstaabp** [comment]
Getting these fixed, but a couple of questions/comments:
 
* I changed the WeBWorK3.pm (which is new) because of the way that dancer runs tests on get/put/post commands to the server.  I had tested the new setup on a couple of machines, but I guess yours is setup a bit differently. 
* About the set headers:  do you have 2 different set headers (screen and hardcopy) selected?  If they are the same, you'll get the same file in both.  If you make changes in the screen version, is it reflected in the output?  

* I have a fix for the student overrides, but it's a bit strange in that it will say that a student's value has been updated (when it hasn't).   However it generally doesn't mess with the db.  See the next item below. 

* About the getMergedSet/getMergedProblem/getUserSet/getUserSet:  on the client side, there are userSets and userProblems  and when fetching it calls the getMerged version.  I think it would be hard to change this.  On the back end to update a userSet, I update only parameters that differ from the globalSet.  I didn't think to do this for the userProblem but will.

**Comment by goehle** [comment]
-  I do have different screen and hardcopy headers and I get them both rendered correctly.  The issue is if I change either of them the render doesn't change as well.  
-  That is fine on the client side.  In some sense it should maybe be the database layer which handles how overrides are actually stored, but in our setup its the object layer.  The only reason I mentioned it is that there are cases where non-overriden values are showing up in the user tables.

**Comment by pstaabp** [comment]
On the set headers, are you getting errors in the network console of the browser (or in the dancer log)?

**Comment by pstaabp** [comment]
about the gateway quiz.  What did you do to recreate it?  I just did the following:
* created a new problem set in the Problem Set Manager (called it gw)
* went to the Problem Set Details view for gw
* changed the type to proctored gateway
* gave a password 1234
* went to the Problem List Details sidebar and added a problem from a problem group
* checked that the set_id:gw user was created in both COURSE_user and COURSE_permission
* adjusted the dates to make it open and visible
* went to ww2
* clicked on "take the quiz gw"  
* added the proctor id and password (not sure why the proctor id was needed)
* went into the quiz.  (I have a error about something being numeric from gateway.pm, but this is probably because a field in the database should be numeric.)

**Comment by goehle** [comment]
You shouldn't need the proctor id in that case.  Was adding permissions something that wasn't happening earlier and was fixed?

**Comment by pstaabp** [comment]
I had the permission problem fixed a couple of weeks ago.  Any idea why it's asking for the proctor_id?

**Comment by goehle** [comment]
-  Depending on how things are set up, it may not have added permissions to proctor users that were created before the permissions thing was fixed.  
-  I'm trying to test stuff but a lot of features have stopped working.  There are js errors like
```
SyntaxError: unreachable code after return statement jquery.truncate.js:195:4
SyntaxError: unreachable code after return statement underscore.js:15:4
```
I did pull develop (with jitar) into ww3 so that may have caused issues.

pstaabp:
I just noticed those two errors in firefox about unreachable code.  Those are in libraries, so I won't worry about those. 

I'll pull ww3 in again and do more testing to see if I can reproduce your errors.

goehle:
OK.  For more specific things to test
-  When I change the type of homework to proctored gateway quiz it doesn't save the change. 
-  When I change the password it doesn't save the change.  
-  The headers aren't showing up in the view/edit box at all.

goehle:
I'm still having trouble with the set headers, but this pull is too unwieldy.  It fixes enough stuff and has enough improvements that I'm just going to merge it.

PR #601, opened 2015-06-23
WebworkClient Local Storage

goehle:
This adds a local storage engine to the WebworlClient path for rendering problems.  Specifically it has some new js files and code which stores submitted answers locally using browser local storage so that they can persist when the same problem is brought up again.  

To test: 
1.  Get the pull
2.  Change the `outputformat` in `clients/TestWW.html` to `sticky`. 
3.  Load TestWW.html and try submitting some answers to problems.  Close the window and open it again.  When you load the problems up the previous answers should still be there (even after a hard refresh).  
4.  Also check the `simple` output format since I added some bootstrap classes to make things render a bit better.


goehle:
I also added the problem score and messages to simple and sticky and added a persisting overall score to sticky which should be tested.

pstaabp:
When I load the page, they appears to be rendering correctly, however, when submitting, I'm getting the following errors for the first problem:

```
Problem13
ERROR caught by Translator while processing problem file:WebworkClient.pm:: define fileName in environment
****************
 Unrecognized character \xA2; marked by <-- HERE after ;   zw(u�R<-- HERE near column 23 at line 1 of (eval 1654)
```

and for the second problem: 
```
WeBWorK Editor using host: http://devel2.webwork.rochester.edu/mod_xmlrpc, format: standard

Unable to decode problem text Error message for command: renderProblem 
faultcode: Server 
faultstring: Unknown exception when trying to verify authentication. argument 1 must contain a user_id at /opt/webwork/webwork2/lib/WeBWorK/Authen.pm line 443. 
End error message
```

I'm guessing that since they are rendering correctly, that the server is up and running okay.

goehle:
Do they work without this patch?

pstaabp:
After thinking about this, I'm a bit confused on how to test this.  The TestWW.html file is unchanged in this and calls the server at rochester.  It doesn't appear to be using my local test server.  Should it?  If so, do I need to change to the local server?

goehle:
Yes you should change it to your local server.  Right now there is a TestWW.html.dist file that you can use to create your own TestWW.html file which has your machines local information.

pstaabp:
Looks like I need to also create a daemon course with user daemon.

pstaabp:
I've got this working now.  The only difference I notice between sticky and simple is that sticky saves your answers. The output looks the same.  Is it supposed to be different?

One thing I noticed is that there is a semicolon at the top of each problem.  (I'm guessing this is coming from PG).  Looks a bit odd.

goehle:
I'll check for that on mine.  It might be coming from pg, it might be an errant semicolon in the format block.  

Sticky should save your answers, as well as your score from previous attempts.

goehle:
The ; was a mistake on my part.  Its fixed.

pstaabp:
It seems that the semicolon is still there if the output is "simple", and no semicolon with output "sticky".  Other than that, it looks like it's working the way you intended.

goehle:
Fixed more of them.

pstaabp:
Looking at what other effects this may have.  Is the attemptsTable used anywhere else?  That seems to be the only code that might be used elsewhere that may affect something else?

It appears that this is all localized to WeBWorKClient.  I'm not sure what else would call that.

goehle:
attemptsTable will eventually be used in the opaqueServer moodle bridge and in Problem.pm.  The attempts table stuff is more of a bugfix than anything else.  It was putting an extra, empty, table row at the top of the table that was messing up scaffolded pg problems. Overall this is localized to the webworkclient side

pstaabp:
I just wanted to make sure.  I think it looks good now.  Should I test on a scaffolded problem?

goehle:
Sure.  You can replace the pg text in the third TestWW.html problem with 
```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGML.pl",
  "MathObjects.pl",
  "parserMultiAnswer.pl",
  "scaffold.pl",
  "PGcourse.pl",
);

TEXT(beginproblem());
$showPartialCorrectAnswers = 1;

############################################
#
#  Setup
#

Context("Numeric");
Context()->variables->are(x => 'Real');

$a = Compute(random(2,9,1));
$c = Compute(random(1,9,1));
$b = Compute(random(2,6,1));
if ($a*$c > 0) {$b += floor(sqrt(4*$a*$c))}

$quadratic = Formula("$a x^2 + $b x + $c");
$x1 = (-$b + sqrt($b**2-4*$a*$c))/(2*$a);
$x2 = (-$b - sqrt($b**2-4*$a*$c))/(2*$a);


###########################################
#
#  The scaffold
#

Scaffold::Begin();

###########################################
Section::Begin("Part 1: Identify the coefficients");

BEGIN_PGML
Consider the quadratic equation given by [`[$quadratic] = 0`].

First, identify the coefficients for the quadratic using the standard form 
[`ax^2 + bx + c = 0`]:

[`a`] = [____]{$a}, [`b`] = [____]{$b}, [`c`] = [____]{$c}
END_PGML

BEGIN_PGML_SOLUTION
Take the coefficient of [`x^2`] for the value of [`a`], the coefficient
of [`x`] for [`b`], and the constant for [`c`].  In this case, they
are [`a = [$a]`], [`b = [$b]`], and [`c = [$c]`].
END_PGML_SOLUTION

Section::End();

###########################################
Section::Begin("Part 2: Solve using the quadratic formula");

$multians1 = MultiAnswer($x1,$x2)->with(
  singleResult => 0,
  allowBlankAnswers => 1,
  checker => sub {
      my ($correct,$student,$self) = @_;
      my ($s1, $s2) = @{$student};
      my ($c1, $c2) = @{$correct};

      return (1,1) if ($c1 == $s1 and $c2 == $s2) or
                      ($c1 == $s2 and $c2 == $s1);
      return (1,0) if $c1 == $s1 or $c2 == $s1;
      return (0,1) if $c1 == $s2 or $c2 == $s2; 
      return (0,0);
   }
);

BEGIN_PGML
Using the quadratic formula, solve [`[$quadratic] = 0`]

>> [`x`] = [______]{$multians1} or [`x`] = [______]{$multians1} <<
END_PGML

BEGIN_PGML_SOLUTION
Recall that the quadratic equation is

    [``x = {-b \pm \sqrt{b^2 - 4ac} \over 2a}``].

You already identified [`a = [$a]`], [`b = [$b]`], and [`c = [$c]`],
so the results are:

    [``x = {-[$b] + \sqrt{[$b]^2 - 4[$a][$c]} \over 2[$a]} = [$x1]``]

or

    [``x = {-[$b] - \sqrt{[$b]^2 - 4[$a][$c]} \over 2[$a]} = [$x2]``].
END_PGML_SOLUTION

Section::End();

###########################################

Scaffold::End();

ENDDOCUMENT();
```

lib/WeBWorK/Utils/AttemptsTable.pm
```diff
@@ -157,16 +157,17 @@ sub answerTemplate {
 			($self->showCorrectAnswers)?  CGI::th("Correct Answer"):'',
 			($self->showMessages)?        CGI::th("Message"):'',
 		);
+
 	my $answerNumber     = 1;
 	#FIXME need to use answer order variable, not sort.
     foreach my $ans_id (@{ $self->answerOrder() }) {  
         #FIXME -- need something more sophisticated to get the answer order correct
```
mgage: Can delete this comment. -- it's been fixed.

lib/WebworkClient.pm
```diff
@@ -581,8 +588,22 @@ $tbl->imgGen->render(refresh => 1) if $tbl->displayMode eq 'images';
 #warn "answersSubmitted ", $tbl->answersSubmitted;
 # render equation images
 
+if ($submitMode && $problemResult) {
```
mgage: A question of style.  This summary was part of attemptsTables (when it was in Problem.pm)  I was planning to add it to attemptsTable (with an option as to whether or not to print the summary) when I got around to it.  Do you think it is better to add the summary outside the attemptsTable?  It's ok with me, but we'll have to rewrite Problem.pm a little when we factor attemptsTable() back in to Problem.pm.  -- I think we should do that refactoring pretty soon by the way.

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -482,7 +483,7 @@ sub renderProblem {
 		print DEBUGCODE "\n\nStart xml encoding\n";
 	}
 	
-	$out2->{answers} = xml_filter($out2->{answers}); # check this -- it might not be working correctly
+	$out2 = xml_filter($out2); # check this -- it might not be working correctly
```
mgage: Why did you decided to filter everything instead of just the answers?

goehle: There are some problems that have other information which needs to be filtered.  It came up with trying to render problems which used contextComplexJ.  They didn't render correctly if I didn't do this.  I'm afraid I can't recall the exact details.

lib/WebworkClient.pm
goehle: attemptsTable could return a hash with both pieces (or even more) and then the caller could decide what to do with them.  It might be better if they were all together.  Some of that summary information is important and I"m not sure students get past reading the table.

mgage:
Possible bug.

For last problem in TestWW.html (the one using base64 version of problem).  

Enter 10, 20, 30 for the answers -- hit show correct answer
Enter 1, 2, 3, for answers -- hit show correct

Refresh the page

10, 20, 30 fill in the answer blanks.
Hit show correct answers 
10, 20 , 30 file the preview section  but 1 ,2 ,3 fill the answer blanks.
(some confusion about previous_answers ?)

lib/WebworkWebservice/RenderProblem.pm
mgage: np.  I can't recall why I only filtered answers.  I think it's ok to leave it for now.

lib/WebworkClient.pm
mgage: That's what I'd like to do -- there is a little stub of code in Problems.pm that could be added on to attemptsTable to accomplish this.  I just didn't do it on the first pass.   I think they should be together as well.

mgage:
Doesn't quite fix it -- there are still issues.
enter 1, 2, 3   submit
enter 12 23 34 submit
enter 125 235 345 submit
refresh the page (reload)
you get 12 23 34 in the answer blanks
hit submit
you get 12 23 34 in the preview and 
125 235 345 in the answer blanks

after that things go well

goehle:
Did you do a hard refresh to get the new js?

goehle:
P.S.  I double checked on a fresh webpage and got the correct behavior.

mgage:
I'm still seeing bad behavior as described above when using refresh/reload (I do shift reloads).  This is using the chrome browser.    Using the Safari browser the feature doesn't seem to work at all.  If I close the page in safari and reopen it then I get only blank entries.  There were only css errors (from bootstrap) in the console.

mgage:
I get the same behavior in firefox that I get in chrome.  Often on opening a new window I'll get incorrect data in the answer blanks.  If I press submit then the correct data appears in the answer blanks.

goehle:
I"ve tested both firefox and chrome and can't replicate this.

mgage:
This works when I use Geoff's machine.  There seems to be something mysterious on my local machine.  I'll pull this.  If I find out something specific that keeps it from working on my local setup I'll send a followup pull request.

Issue #599, opened 2015-06-22
WW3 and WW2 cookies


goehle:
I've been meaning to post about this for a while.  Ever since you changed the path on the ww2 cookie (something which needed to be done to make it ww3 compatible) in 75d35321bed5c7ecaf whenever I switch between develop and ww3 I can't log in unless I remove any old cookies I had laying around.  

To test
1.  Switch to ww3 and log in.  
2.  Switch to develop and restart your server
3.  Open a different browser and log in to the same course with the same user.  (This will log you out in the server while leaving your cookie in the original browser intact.)  
4.  Go to the original browser.  When you click a webwork link it should redirect you to the log in page and your login should fail without any warnings.  

The issue is that the WW3 cookie is still hanging around and has path '/' and is being checked for authentication.  The authentication should kill the cookie if its no good, but it kills the cookie with ww2 path 'webwork2/' and that doesn't kill the ww3 cookie (which hangs around).  It would be good to fix this before ww3 goes in develop because it can cause really confusing issues for people using ww3 for the first time. 

Its a subtle issue.  One possible solution is to change the name of the cookie from `WeBWorKCourseAuthen.$courseID` to `WeBWorK.CourseAuthen.$courseID` in ww3.  This will cause there to actually be different cookies for ww3 and develop and the collision won't happen.  It needs to be changed everywhere in ww3 as well as ww2 (in the ww3 branch).

pstaabp:
It's been on my mind as well and it will need to be dealt with.  I'll try the cookie rename.  That's a good idea.

PR #602, opened 2015-06-24
Bugfix/cookie

pstaabp:
This fixes an error where the user can't login when switching between the develop and the ww3 branches of webwork.  It will be a serious problem when ww3 gets merged into develop.  This fixes issue #599.   To test:

1. Switch to ww3 and log in.
2. Switch to develop and restart your server
3. Open a different browser and log in to the same course with the same user. (This will log you out in the server while leaving your cookie in the original browser intact.)

4. Go to the original browser. When you click a webwork link it should redirect you to the log in page and you should now be able to login.

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright © 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
+# Copyright ï¿½ 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
 # $CVSHeader: webwork2/lib/WeBWorK/Authen.pm,v 1.63 2012/06/06 22:03:15 wheeler Exp $
 # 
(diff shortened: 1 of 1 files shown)


goehle:
After this pull request I have to relogin whenever I switch between ww3 and ww2.  I think you forgot to also change the cookie name in ww3 files.  A grep shows the old cookie prefix is still present in 
-  ./webwork3/lib/Routes/Course.pm
-  ./webwork3/lib/Routes/Authentication.pm
-  ./webwork3/bin/app.pl
-  ./webwork3/Authen.pm

pstaabp:
I was thinking about this, but should that go in this pull request, since this is off of develop?  Perhaps I should have a second PR off of ww3 that fixes this.

goehle:
I didn't see this was off of develop.  It should be off ww3 to ww3, but should change the cookie prefix in both ww2 and ww3.  Then when you switch back and forth from develop to ww3 you will have different cookies.

pstaabp:
I'll move it to off ww3 and resubmit.

pstaabp:
I just opened a new pull request off ww3 and will close this.

Issue #470, opened 2014-08-13
WW3 Redefines Before And After


goehle:
When I run webwork3/public/dispatch.fcgi I get 
```
Subroutine before redefined at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 708.
Subroutine after redefined at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 717.
```
I thought we took care of this back in Asheville?

pstaabp:
I'm revisiting this error.  I realize that it comes from adding `use Dancer` to the top of the lib/WeBWorK/Utils.pm file.  I think I did this because somehow I ran into an error that this fixed it.  If I remove this line then I get an error related to a missing Apache::Cookie module, but I'm not running apache1, so it shouldn't need this.     

The missing Apache::Cookie error is tracked this down to a missing environmental variable. In lib/WeBWorK/Cookie.pm there's a switch (depending on apache1 vs. apache2) and if I define `MOD_PERL_API_VERSION=2` then this is all switched. 

This is an easy fix, but will need some checking on different configurations.

pstaabp:
Since it was quiet on our server the past few days, I've been playing with this on our production server. 

If you look at the file change sited above (774fe36), then I just commented out the `use Dancer` line from the WeBWorK::Utils file.  I was able to get this to work on my develop machine in that I run dancer on the command line and was able to set the env var: `MOD_PERL_API_VERSION=2`  which is used to branch on in the WeBWorK::Cookie module.   I thought that explicitly setting this environmental variable would then select the apache2 modules instead of the apache1 modules.  

It appears that I can't get the same results on my production machine which uses the fastcgi scripts from the webwork.apache2.4-config.  It doesn't appear that the variable `MOD_PERL_API_VERSION` isn't being recognized.  Here's the error:

```
[Thu Jun 04 14:51:24.786764 2015] [:warn] [pid 4992] FastCGI: (dynamic) server "/opt/webwork/webwork2/webwork3/public/dispatch.fcgi" started (pid 4998)
Use of uninitialized value $WeBWorK::Constants::WEBWORK_DIRECTORY in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/Constants.pm line 41.
[...]

PR #603, opened 2015-06-24
Ww3 feature/before after fix

pstaabp:
This is a fix for issue #470.  

To check, there is a new line in webwork.apache2-config.dist (and webwork.apache2.4-config.dist) that adds the line: `SetEnv MOD_PERL_API_VERSION 2`

You will need to add this line to your non .dist version and restart the server.  You shouldn't get the warning in issue #470 anymore. 

I figure this is safe because if anyone is running apache2.2 or 2.4 then they will be running mod_perl2.  This variable gets picked up in Dancer that wasn't being set another way.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -39,7 +39,7 @@ use Errno;
 use File::Path qw(rmtree);
 use Storable;
 use Carp;
-use Dancer qw(:script);
+#use Dancer qw(:script);
 
 use constant MKDIR_ATTEMPTS => 10;
 
--- a/bin/startDancer
+++ b/bin/startDancer
@@ -1,3 +1,5 @@
 #!/bin/sh
 
-perl -I /opt/webwork/webwork2/lib -I /opt/webwork/pg/lib /opt/webwork/webwork2/webwork3/bin/app.pl
+export MOD_PERL_API_VERSION=2
+
+perl -I $WEBWORK_ROOT/lib -I $WEBWORK_ROOT/../pg/lib $WEBWORK_ROOT/webwork3/bin/app.pl
--- a/conf/webwork.apache2-config.dist
+++ b/conf/webwork.apache2-config.dist
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright © 2000-2010 The WeBWorK Project, http://openwebwork.sf.net/
+# Copyright ï¿½ 2000-2010 The WeBWorK Project, http://openwebwork.sf.net/
 # 
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
@@ -38,6 +38,8 @@
 #ScriptAliasMatch /webwork2_course_files/([^/]*)/show-source.cgi/(.*) /opt/webwork/courses/$1/html/show-source.cgi/$2
 
 PerlModule mod_perl2
+
+SetEnv MOD_PERL_API_VERSION 2
 
 <Perl>
 
--- a/conf/webwork.apache2.4-config.dist
+++ b/conf/webwork.apache2.4-config.dist
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright © 2000-2010 The WeBWorK Project, http://openwebwork.sf.net/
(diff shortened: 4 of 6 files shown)


lib/WeBWorK/Utils.pm
```diff
@@ -39,7 +39,7 @@ use Errno;
 use File::Path qw(rmtree);
 use Storable;
 use Carp;
-use Dancer qw(:script);
+#use Dancer qw(:script);
```
goehle: Commenting this out is causing app.pl to return the error
```
Can't locate mod_perl.pm in @INC (@INC contains: /opt/webwork/pg/lib /opt/webwork/webwork2/lib /opt/webwork/webwork2-vanilla/webwork3/lib /opt/webwork/webwork2/webwork3/lib /etc/perl /usr/local/lib/perl/5.14.2 /usr/local/share/perl/5.14.2 /usr/lib/perl5 /usr/share/perl5 /usr/lib/perl/5.14 /usr/share/perl/5.14 /usr/local/lib/site_perl .) at /opt/webwork/webwork2/lib/WeBWorK/Cookie.pm line 35.
```

pstaabp:
Did you add the line:
```
SetEnv MOD_PERL_API_VERSION 2
```

in your webwork.apache2-config file?

goehle:
Yes, that is the line that causes the error when it is present.

pstaabp:
Where the error is occurring is in WeBWorK::Cookie because it is used via the Authen.pm module, which Dancer uses to login.  However, Dancer doesn't use mod_perl and when it hits line 35 of WeBWorK cookie, then it defaults to Apache1 and tries to load mod_perl.pm (instead of mod_perl2.pm).  What I thought would work is the line:
```
SetEnv MOD_PERL_API_VERSION 2
```
would send the cookie to the correct branch.  Evidently, it is not grabbing that environmental variable correctly.  

As I mentioned before, I stumbled on adding Dancer to the lib/WeBWorK/Utils.pm to not get the error, but it shouldn't be there.  That Utils.pm module has nothing to do with Dancer. 

I also tried adding a line like:
```
$ENV{MOD_PERL_API_VERSION } = 2;
```
inside the <Perl> part of webwork.apache2.4-config, but this didn't work on my production server. 

Any thoughts?

pstaabp:
So this is working on my machine anymore.  I don't know what's going on.  For some reason, it's not picking up the env variable `$ENV{MOD_PERL_API_VERSION}`. 

I'll keep working on this.

pstaabp:
It appears that environmental variables don't pass well into dancer applications.  @goehle: How are you deploying dancer to test ww3?

goehle:
I'm using the fcgi setup.  That is going to be *super* bad at passing environment variables.  You might be able to add them directly to the fcgi script.

goehle:
I still have problems with your latest changes.  Should we set the environment variable in app.pl?  It seem like that would be a catch all way to make sure it gets set by something running dancer.

goehle:
OK.  I think I have something.  First, I set the environment variable in app.pl by adding 
```
BEGIN {$ENV{MOD_PERL_API_VERSION}=2};
```
*before* any of the `use WeBWorK::...` lines.  This sets the environment variable for all of the perl stuff correctly.  Unfortunately for things that call `use mod_perl` the mod_perl stuff is looking for a more deeply defined environment variable to decide if its supposed to be version 2.0 or not and its not picking up this environment variable.  So I removed the `use mod_perl` line from `Cookie.pm` and changed the begin block to 
```
BEGIN {
	if (MP2) {
	    require mod_perl2;
	    require Apache2::Cookie;
	    require APR::Request::Error;
	    Apache2::Cookie->import;
	    push @WeBWorK::Cookie::ISA, "Apache2::Cookie";
	} else {
	    require mod_perl;
	    require Apache::Cookie;
	    Apache::Cookie->import;
	    push @WeBWorK::Cookie::ISA, "Apache::Cookie";
	}
}
```
This explicitly uses the webwork MP2 variable (which is set correctly) to determine if mod_perl or mod_perl2 should be imported.  With this I was able to get ww3 running without including Dancer into Utils.

pstaabp:
I deleted the remote branch to throw away a lot of the attempts. I thought I could reopen this.

pstaabp:
I just opened a new version of this.  Thanks for the fix Geoff.  I almost had the exact code except didn't put the BEGIN block above the use WeBWorK lines and that was it.

PR #604, opened 2015-06-30
Ww3 bugfix/cookie

pstaabp:
This fixes a discrepancy an error when switching branches between ww2 and ww3 due to differences in cookie.  To test:

1. Switch to ww3 and log in.
1. Switch to develop and restart your server
1. Open a different browser and log in to the same course with the same user. (This will log you out in the server while leaving your cookie in the original browser intact.)
1. Go to the original browser. When you click a webwork link it should redirect you to the log in page and you should now be able to login.


goehle:
This works for me.  I'm able to switch back and forth between ww2 and ww3 in the ww3 branch, and when I switch back to the develop branch I can still log in. Both cookies seem to be functioning.

PR #606, opened 2015-06-30
PG IO Cleanup

goehle:
This is a tweak to WeBWorK that goes along with PG pull request 219.  

Basically this replaces a particular object oriented Perl package by its regular "function" alternative.  It should have no effect on WeBWorK2. 

To Test: 
-  Open a problem in a problem editor.  
-  Save it as another problem. 
-  Save it as another problem with a relative path not in the courses directory .e.g. `../../../../../`.  You should see an error.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -32,7 +32,7 @@ use DateTime::TimeZone;
 use Date::Parse;
 use Date::Format;
 use File::Copy;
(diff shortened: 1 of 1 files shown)


mgage:
It checks out.  No problem.  This is the kind of change often needed to get things into Safe compartments.  It forces more compilation of the subroutine at compile time instead of at perl's run time. Sometimes this kind of fix is not enough.

Issue #444, opened 2014-07-11
WW3 and problemRandomize.pl


goehle:
When I try to run a problem that uses problemRandomize.pl in the Library Browser or the Problem Set Detail viewer I get a warning about an uninitialized value in a string ne at line 181, which is referring to
```
$main::inputs_ref->{effectiveUser} ne $main::inputs_ref->{user}
```
These problems work when viewed using the WW2 library browser.  My guess is that WW3 isn't setting effectiveUser when its calling PG, but it could be something else.

pstaabp:
I'm back working on this issue.  It appears that that the line that Geoff cited above: 
```
$main::inputs_ref->{effectiveUser} ne $main::inputs_ref->{user}
```
are the standard parameters passed around using requests.  When ww3 is rendering a problem, I have written a general rendering subroutine at /webwork2/webwork3/lib/Utils/LibraryUtils.pm starting on line 27.  This sub builds up the needed information in the hash renderParams and passes it to a WeBWorK::PG object.  

Is there a way to pass the effective user to the PG renderer this way?

PR #607, opened 2015-06-30
Fix so that the effectiveUser is now passed correctly to the Library …

pstaabp:
This fixes the error in issue #444.  This fixes problems with problemRandomize.pl, but not sage problems (one of which was listed in issue #444).

### DIFF
--- a/webwork3/lib/Utils/LibraryUtils.pm
+++ b/webwork3/lib/Utils/LibraryUtils.pm
@@ -34,6 +34,7 @@ sub render {
 	for my $key (@anskeys){
 		$renderParams->{formFields}->{$key} = params->{$key};
 	}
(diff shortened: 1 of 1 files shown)


goehle:
I'm still getting 
```
 "Warning caught during route execution: Use of uninitialized value in string ne at line 181 of [PG]/macros/problemRandomize.pl\n", "exception" : "Warning caught during route execution: Use of uninitialized value in string ne at line 181 of [PG]/macros/problemRandomize.pl\n"
```
from all problems that use problemRandomize, not just the ones in Sage.  Is there something else I need to do?

pstaabp:
I just committed a fix to this.

goehle:
This fixes the problem Randomize problems.  The sage problems are still broken, however that can wait for another pull

PR #608, opened 2015-07-05
LTI Changes

goehle:
These are some changes I made to get LTI working with the Blackboard server at WCU.  In particular
1.  I added a variable which allows you to use the email address to get the username when the username isn't provided.   Set `$preferred_source_of_username` to `lis_person_contact_email_primary` and `$strip_address_from_email = 1`.  Then when a user is created it will use the username from the email and not the whole email. 

2.  I set up the system so that if `$LTIBasicToThisSiteURL` is not set then the system will try to get the url from the path given in the LTI request.  (It will try both the http and https versions.)  This can be tested by removing the `$LTIBasicToThisSiteURL` variable as long as the path in the LTI request and your server address match. 

It works on our LTI setup but it would be nice if this could be tested on a few others.


goehle:
BTW:  The best way to test this would be with warn statements. 
-  Add `warn($self->{user_id})` to line 253 of LTIBasic.pm  With $preferred_source_of_username set to "lis_person_contact_email_primary" and "$strip_address_from_email" set to 1 you should get only the username.  Otherwise you should get the email. 
-  Add `warn($path)` to line 417 of LTIBasic.pm.  If $LTIBasicToThisSiteURL is set you should get that value for path.  Otherwise it should be automatically constructed from the request.

pstaabp:
I'd be happy to test this.  We use blackboard on campus.  Do you have some setup instructions that you use for this?

goehle:
First you need to set up webwork as an LTI through blackboard.  I'm not entirely sure how this is done, since I had a blackboard administrator do it for me.  I imagine these will help: https://help.blackboard.com/en-us/Learn/9.1_SP_10_and_SP_11/Administrator/120_System_Integration/LTI_Tool_Providers

Next you have to uncomment out the LTI authen line from localOverrides, and copy the LTI authen file from the dist version.  For this branch the only thing you need to set is `$LTIBasicConsumerSecret`, which should be the same as what was set in blackboard. 

Next you create a link in a blackboard course to a particular homework set.  You should also click the "This link is to a Tool Provider" box.  When you click the link it should log you into WeBWorK automatically, and even create student users if they don't exist.

pstaabp:
I'm trying to help our admin person set up blackboard LTI.  I sent along the link to http://webwork.maa.org/wiki/LTI-Basic_Authentication

Does the admin person just need the info in in the bottom (the example connecting to the IU site)?  What is the URL that they need to add in?

goehle:
I dont know about those instructions.  In particular I thought the domain/url was the url of the LTI site, i.e. of your webwork server.  Here is a screen shot of our configuration. 

screen_shot_2015-02-27_at_8 17 32_am

goehle:
Above that there is the url for our webwork site and the consumer secret.

pstaabp:
Is the URL just to /webwork2  or is it to a subdirectory?

goehle:
I'm pretty sure its just the full url to the webwork2 server.  To be fair, I didn't set ours up either.  In particular the official instructions are
You can add new tool providers using Register Provider Domain. 

    On the Administrator Panel, under Building Blocks, click Building Blocks.
    Click LTI Tool Providers.
    Click Register Provider Domain.
    Type the Provider Domain, such as http://www.example.com.
    Select the Provider Domain Status - Approved or Excluded. Approved is the default setting. Selecting Excluded prevents users from adding tool links to that provider.
    Optionally, provide any Secondary Hostnames for the provider.  The provider can supply this information.
    Tool providers can offer configuration for the entire site, or for each individual link. If the provider allows configuration for the entire site, select Set Globally and type the Tool Provider Key and Tool Provider Secret. Leaving the Default Configuration as Set separately for each link requires users to enter the Key and Secret when creating each link.
    Add any custom parameters required by the tool provider in the text box Tool Provider Custom Parameters.
        Each parameter must be on its own line.
        Parameters must be formatted as a "name=value" pair.
        Custom parameters can include template variables that are resolved using data from Learn, in the Learn parameter format @x@ user.full_name @x@. To learn more, see Template Variables.
    The Institution Policies section pulls preferences from the Global Properties. You can manage settings for the individual provider here.  Any choices you make for the individual provider override the Global Properties.

So it looks like its just the webwork2 server domain.  You don't even need the /webwork2.  You can use the global key option if you want your key to be for the whole server, or you can set the key per course using course.conf.

conf/authen_LTI.conf.dist
```diff
@@ -81,18 +81,24 @@ $authen{user_module} = [
 $preferred_source_of_username = "lis_person_sourcedid";
 #$preferred_source_of_username = "lis_person_contact_email_primary";
 
+# Some LMS systems, Blackboard in particular, do not send lis_person_sourcedid
+# if you enable this flag and have $preferred_source_of_username set to
+# lis_person_contact_email_primary, then webwork will strip off the address 
+# portion of the email and just use the username. 
+$strip_address_from_email = 1;
```
mgage: This is odd.  This line and the comment above it don't appear in authen_LTI.conf.dist when I pull this request.  ?????

conf/authen_LTI.conf.dist
```diff
@@ -81,18 +81,25 @@ $authen{user_module} = [
 $preferred_source_of_username = "lis_person_sourcedid";
 #$preferred_source_of_username = "lis_person_contact_email_primary";
 
+# Some LMS systems, Blackboard in particular, do not send lis_person_sourcedid
+# if you enable this flag and have $preferred_source_of_username set to
+# lis_person_contact_email_primary, then webwork will strip off the address 
+# portion of the email and just use the username. 
+
+# $strip_address_from_email = 1;
```
goehle: No idea.  I commented out this variable by default.  (it probably should be).  Maybe the new commit will trigger something.

lib/WeBWorK/Authen/LTIBasic.pm
```diff
@@ -233,15 +233,24 @@ sub get_credentials {
...
-		
 
 		$self -> {email} = uri_unescape($r -> param("lis_person_contact_email_primary"));
 		if (!defined($self->{user_id})
+		    or (defined($self -> {email})  
+			and $ce -> {get_username_from_email})) {
+		    $self->{user_id} = $self -> {email};
+
+		}
+
```
mgage: Geoff.  Did you mean to leave the if clause above in the code? There is no parameter get_username_from_email in any of the config files that I can find. And this causes somewhat confusing logic.  For example if you set
get_username_from_email=1 then you get the full email address for the user_id (no matter what userid LTI provided).
If you set strip_address_from_email =1   then you still get the full address.
If you now set $preferred_source_of_username = "lis_person_contact_email_primary"; you get the stripped address, 

My guess is that you want to leave out the get_username_from_email flag and the above snippet out entirely. 
Comments?

goehle:
Good catch.  I think that was left over from when I was implementing the ability to get the username from the email before I decided to use the existing flags.

goehle:
Mike added some extra debug statements.  I've tested them and they help out.  The various LTI setups are substantially different in terms of what parameters they are passing and what the call the various roles users can have.  This should help individual administrators untangle that mess. 

Peter, could you try setting up the LTI stuff again.  If you add
```
$debug_lti_parameters = 1; 
```
to the autthen_LTI.conf file it will print out a bunch of debug info.

mgage:
I think this is a reasonable pausing point for now.  I can see further improvements that will make installing LTI easier and perhaps increase it's utilities.  I think those major improvements should wait until after the 2.11 release.

PR #609, opened 2015-07-06
Tweaks for 2.11

goehle:
In preparation for 2.11 this pull request contains a collection of tweaks and fixes. 
1. I implemented the suggested fix for issue number #538.  I don't have CAS so I'm assuming the given fix works on their system and is sane.  
2.  Added curl to check_modules since it is now one of the programs in site.conf and is used by AskSage
  -  Run check_modules.pl to test. 
3.  Fixed a bug with sticky answers to work with radio problems. 
  -  To test visit a problem with radio buttons using TestWW.html (e.g. /FortLewis/DiffEq/3-Linear-systems/05-2D-systems-vector-fields/Systems-Classification-06.pg) and set the output to sticky.  After submitting once, the radio buttons should all have the same value before the patch, and the correct values after 
4.  Fixed a bug with the new url validation and unattached problems.  
  -  To test, save as an unattached problem in any of the PG editors.  Before the patch it should fail and after the patch it will work.  
5.  Added "Feedback by Section" to simple.conf in the "Email" options area. 
  -  Visit the configuration page (email options) and check that the new option works and that the description is coherent. 
6.  Added more verbose error reporting to the library browser.  People will still probably need to check there apache logs to figure out what is really wrong, but this will at least let them know there is a server issue and not a js issue. 
  -  To test the error reporting for request errors just change your `#server_root_url` to something nonsensical, restart the server, and try to add a problem or select a subject in the library browser.  After a (pretty lengthy) 10 second delay you should get an error alert. 
  -  To test the error reporting on WeBWorK errors you need to add a die statement to WeBWorK Client somewhere.  Restart the server and try to add a problem or select a subject in the library browser. 
  -  Also test to see that the usual functionality is still present.  
7.  Because editing courses.conf is dangerous (one errant semicolon and you are locked out of your course) there is now a blacklist of files that cannot be edited via the File Manager unless the user has "admin" privileges.  The blacklist is currently course.conf and simple.conf.  This also implements changes so that users cannot elevate their permissions above what they currently have and so that administrators are added as "admin" to courses.  
  -  Test to see that you are unable to edit course.conf in an existing course (with professor permissions).
  -  Test to see that in that course you are unable to give yourself admin privileges. 
  -  Create a new course and check that you have admin privileges, can edit course.conf, and can give other users admin privileges.


lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
```diff
@@ -480,11 +480,30 @@ sub Edit {
 	my $self = shift;
 	my $filename = $self->getFile('edit'); return unless $filename;
 	my $file = "$self->{courseRoot}/$self->{pwd}/$filename";
+	my $r = $self->r;
+	my $userID = $r->param('user');
+	my $ce = $r->ce;
+	my $db = $r->db;
```
dpvc: Not sure why `$db` is needed.  Is the not used in the code you added below.

goehle: It was left over from when this was going to be a coursewide setting.  I'll remove it.

goehle:
Extra Addition:  I added some code so that errors within the WebworkWebserver will correctly get to the library browser.  Now you can add a die command in `RenderProblem.pm` and the error message will appear when you try to refresh a problem in the library browser.

goehle:
Extra Addition:  Two small things
-  I modified addadmin and putadmin to add the admin user with admin privileges.  This is so if a course doesn't have an admin privileged user it can get one.  Test by using addadmin and putadmin with wwsh. 
-  I modified check_modules.pl so that it has the ability to check the version of a package.  Run and see if it works.  Change the version of LWP::Protocol::https to 20.4 to see if it complains.

mgage:
HI Geoff,

This pull request now has some conflicts with the develop branch.  Could you fix those and update?
(there were only two files that needed minor changes I believe.)

mgage:
1.    check_modules.pl  --- checks for curl properly 
   -- there are these errors which are probably spurious
     *   Prototype mismatch: sub main::from_json: none vs ($@) at (eval 194) line 2.
     *  Prototype mismatch: sub main::to_json: none vs ($@) at (eval 194) line 2.
     *   ** Apache2::Request found, but failed to load: Can't load   (this might only be on freeBSD installations) '/usr/local/lib/perl5/site_perl/5.16/mach/auto/APR/Request/Request.so' for module APR::Request:  /usr/local/lib/perl5/site_perl/5.16/mach/auto/APR/Request/Request.so: Undefined symbol "apreq_hook_disable_uploads" at /usr/local/lib/perl5/5.16/mach/DynaLoader.pm line 190.
 at /usr/local/lib/perl5/site_perl/5.16/mach/APR/Request/Param.pm line 27.
2. Radio buttons work
3. unattached problems can be saved from editor
4. Feedback section reporting works
5. Error messages for xmlrpc (Library browser callbacks) works great!!
6. Admin user level works well.  prevention of elevating yourself above current permission level works
    *   When creating a new course for yourself you need to be cautious. If you fill in yourself as the contact professor of the course you will be given professor privileges.  If you leave the contact person blank then you will be added to the new course with admin privileges.  
   * On reflection this works the way it should but we should explicitly mention it in the release notes since some administrators might be surprised by this behavior.

mgage:
This is ready to pull as soon as the conflicts are resolved.

mgage:
It is still possible for a TA to change a professor's or admin's password.

goehle:
Should I add that to this pull request?

mgage:
Let's add it to the next one.  But I think perhaps we should address it before we split off 2.11  It's kind of a long shot security hole unless you don't trust your TA's.  It can however completely circumvent your other security fixes.  A professor or a TA for example, could change the password of the admin, login as the admin and do what they wish.

PR #610, opened 2015-07-06
Throwing a random integer into the set name "Undefined set"

jwj61:
Currently, when you view a problem in the library browser, or in a "try it" window, or viewing from an editor window, webwork pretends the name of the problem set is Undefined_Set.  Since images are links to actual images where the link name is built out of user/seed/set name/etc., there can be collisions when there are multiple places using images in this way.

The prepends random integers to the "fake set name" Undefined_Set, which should reduce the chance of a collision to a very small number.  Most of the changes are spots where webwork has to detect the special set name Undefined_Set to allow for the added digits.

A good place to find images in files is single variable calculus/limits and contin./finding limits using graphs.

### DIFF
--- a/lib/WeBWorK/Authz.pm
+++ b/lib/WeBWorK/Authz.pm
@@ -390,7 +390,7 @@ sub checkSet {
 		     $set->version_id eq $verNum ) {
 			# then we can just use this set and skip the rest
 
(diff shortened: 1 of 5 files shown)


goehle:
We've all agreed that the PG solution is better so I'm closing this.  Thanks for tracking it down and getting us started on fixing it.

PR #613, opened 2015-07-31
OPLUtils Change

goehle:
This is a small change to OPLUtils.  There are calls to `$sth->rows` to fetch the number of rows.  However, in general that is fragile because the select statement may not return the number of rows until all of them have been fetched.  See: http://search.cpan.org/~timb/DBI-1.633/DBI.pm#rows

This is a slight workaround.  Test by running OPL-update and making sure that it generates the subject-tree and textbook-tree correctly.


mgage:
Checked.

PR #614, opened 2015-08-04
Fix bug 3394, display of ans_array matrices in math3 theme

jwj61:
Brackets around matrices with ans_array were way too big.  Try

 Library/Rochester/setLinearAlgebra14TransfOfRn/ur_la_14_31.pg

in math3.

### DIFF
--- a/htdocs/themes/math3/math3.css
+++ b/htdocs/themes/math3/math3.css
@@ -599,6 +599,10 @@ table.attemptResults th {
 }
 
 /* override above settings in tables used to display ans_array results */
+.ans_array_open, .ans_array_close {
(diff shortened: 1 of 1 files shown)


dpvc:
This will fix it for MathJax mode, but I think there may need to be more for image mode.  You might need to add CSS that scales images as well.  Something like

```
.ans_array_open img, .ans_array_close img {
    width: 67%; height: 67%;
}
```

might work.  I haven't tried it in image mode.

jwj61:
Just looked with images and it looks ok to me.

dpvc:
OK, sounds good.

PR #615, opened 2015-08-06
added perioding re-randomization

abasyrov:
the feature could be turned on/off on the course level
the course-level default period for re-randomization could be set
per-problem setting of re-randomization period is also possible

the per-problem re-randomization period is exported into set-def files regardless of the course-wide setting in the format max_attempts:re-randomization-period

the re-randomization period value is also imported from set-def file regardless of the course-wide settings (if it is present in the set-def file)


conf/defaults.config
```diff
@@ -169,6 +169,15 @@ $editor{textTitle} = '';
...
+###############################################################################
+# periodicRandomization
+################################################################################
+# switch to enable periodic re-randomization
+$pg{options}{enablePeriodicRandomization} = 0;
+# course-wide default period for re-randomization, should be an integer
+# the value of 0 disables re-randomization
+$pg{options}{periodicRandomizationPeriod} = 20;
+
 ################################################################################
```
goehle: These options (and the ones below) should also be put in localOverrides.conf.dist.  That is the file where administrators set up server defaults.

goehle:
Thanks for your contribution; this seems like an interesting feature.  It allows professors to have something which is kind of half way between unlimited attempts and limited attempts.  You should know this is a bit of an awkward time because we are trying to get release/2.11 ready to be spun on off and there is a reasonably large pull request for "just-in-time" type homework sets #580 which has a non-zero intersection with this.  That pull request is going to be merged into develop first, so what is likely going to happen is that you will need to either wait until #580 is pulled into develop (or rebase this branch off of #580) and make some changes to this code.  The big ones I see so far are:
- [ ] The changes you made to ProblemSetDetail (and fixes for the issues below) will need to be side ported to ProblemSetDetail2.  
- [ ] The  changes you made to importing/exporting set defs will need to be redone to fit with the changes to ProblemSetList2 in Just-In-Time.  In particular right now there are errors if you import a file created before this patch with the patched code (e.g. SetDemo) and the notation that is used in the export (with the `:`) is too hacky.  This isn't really your fault.  The old def format wasn't set up well for problems to have a lot of potentially optional parameters.  There is a new format in the Just-In-Time branch for specifying problems which is better for this kind of thing (see  http://webwork.maa.org/wiki/Set_Definition_Files ).  Basically you shouldn't mess with the old `problemList` parser in ProblemSetList2 and just add your parameter keys to the `problemListV2` parser so that it will recognize them. 

As I said, those changes will have to wait until #580 is added to develop, or you will have to do some fancy git rebasing.   There are some other issues which can be looked at right now:
- [ ] Your options in defaults.config should be put into localOverrides.conf.dist.  The local overrides file is where site administrators will set site wide defaults. 
- [ ] The "Rerandomize after" box shows up on the Problem Set Detail  page, even if the option is turned off.  To keep the interface as simple as possible all text/options/boxes/whatever related to a feature should be turned off if the feature isn't enabled.
- [ ] The default number of attempts between re-randomization is really high.  It doesn't have to be changed, but according to the data gathered so far in the big data project average number of attempts total on any given problem is less than 5.  
- [ ] The parameter value of 0 for `Rerandomize` after does not show up as "Never" in the box.  This is probably a bug in line 606 of ProblemSetDetail.pm where some or all of those `or`'s should be `defined or`'s.  
 - [ ] One issue, and it may or may not be something you want to deal with, is that a new seed does not guarantee a new problem.  Most problems only have a handful of possible values for their random parameters, and there are quite a few problems with no random parameters at all.  The ShowMeAnother feature has some code which tries to find a seed which gives a different problem text.  Its possible you might want to use that. 
- I would be interested in hearing why you have the "Request New Version" button rather than automatically creating a new version and putting a message on the Problem page saying that you have done so.
There will some more testing and code review in the future.  However this should be enough to work with for now.  Thanks again for your submission to the project.

abasyrov:
>I would be interested in hearing why you have the "Request New Version" button rather than automatically creating a new version and putting a message on the Problem page saying that you have done so.

If we go with automatic creation of new version, then we would be subjecting students to the following confusing scenario at the last attempt before re-randomization:
- student submits their answer to the old version
- webwork grades the answers against the old version (or maybe new version -- it depends on exact placement of code) of the problem
- webwork feedback is shown above the new version of the problem. 

With the 'request new version' button option, the student would submit their answer to the old version, it would be graded against that same version of the problem, and webwork feedback would be shown above the old version of the problem and the 'request new version' button would be shown. After the 'request new version' button is pressed, the new version is generated and presented to the student. 

I'll work on the rest of the issues you've pointed out. I thought about guaranteeing a new version of the problem (if possible) -- I think that the code from 'show me another' could be easily stolen/adapted to re-randomization. However, that would only guarantee that the new version is different from the old one, but it does not prevent the 'version cycle'. If the problem has only two versions in total the student would see alternating sequence of the two versions.

goehle:
The "Request New Version" thing makes sense.

The version cycle is not something you can prevent, I think.  Even more of
an issue are problems with no randomization at all.  But again, you cant do
anything about that.

mgage:
Hi Alex,

I've merged the JITAR pull request ( a really big one) into develop.  When you have time reconcile this commit with the current develop.  I'll be pulling all the requests I can this week preparing to split off release/2.11
-- Mike

goehle:
Closed because a new pull request was created.

PR #618, opened 2015-08-13
Math 4 Javascript upgrades

goehle:
Two changes to Math4 js

- Images in problems (specifically in div.problem-content) will be resized according to their attribute height and width and not according to their native height and width.
 - To test: Create a problem with an image and set the height and width attributes to be something with a different aspect ratio than the images native height and width. Before the patch the image should be rescaled to fit the div.problem-content container according to its native aspect ratio and after the patch it should use the attribute aspect ratio.
 - Also test that images still rescale correctly when there is no specified height/width.
 - Resize your window and test that the images scale up and down correctly when the window is small. The image should never go outside the problem area.
- I also redid the navigation collapsing code and ported over some changes which would collapse and expand it dynamically if the window was change.
 - Check that the sidebar collapses and expands.
 - Check that as you make a window small or large the sidebar collapses and expands automatically
 - Check that the sidebar and windows look correct, including pages where there are no sidebars.


mgage:
This pull request also needs updating.

mgage:
This checks out.

PR #619, opened 2015-08-15
js for iframe-resizer

Alex-Jordan:
With this in place, anyone using the embedding blogged about here (http://michaelgage.blogspot.com/2015/06/whether-writing-full-text-book-or-just.html) can use some additional javascript to get the iframe to have the appropriate height. Also the iframe will resize itself as the solution knowl is clicked, submit buttons are pushed, etc.

The page that has the iframes needs to load
`<script type="text/javascript" src="path to iframeResizer.min.js"></script>`

Then you also need 
`<script type="text/javascript">iFrameResize({log:true,inPageLinks:true,resizeFrom:'child'})</script>`
(possibly with different options; see https://github.com/davidjbradshaw/iframe-resizer)

I found that I need to put this second line of script right after each iframe, but I suspect that is because in Mike's examples, each iframe is inside its own knowl. Maybe without knowls, the second line can go with the first line and that will be enough for the whole page.

To test this, I used Mike's example .html. I modified it as already suggested with the script lines. Then I changed the server to my own server with the courses and logins changed accordingly. I removed each iframe's height attribute, and set each width attribute to 100%.

The modification to WebworkClient.pm only adds a line of script to the standard, sticky, and simple modes. Not debug. I'm not sure if that is the right decision, but debug didn't have other lines of script, so I went with consistency. If anyone thinks putting this into debug too would be good, let me know.


Alex-Jordan:
Looks like the content I added to js/vendor didn't come through. Is such stuff protected with gitignore?

Alex-Jordan:
Never mind. I had copied the iframe-resizer folder up with its own git repo, and that interfered with the webwork2 repo. I think it's fixed now.

Is this too much? I thought that including the entire iframe-resizer project was the right thing to do. But maybe only the specific parts that we use? Including the license?

goehle:
I think this is too much.  I would just include the js file you are actually using in the page and maybe their project description file if you want.

Alex-Jordan:
OK, I took out all but the license, the readme, and the project's js folder. There are six files in that folder. One is the js that the content page within the iframe uses (and that has been added to WebworkClient.pm). One is something that may be necessary in the same place to support ie8 and earlier, although the documentation says "This is not normally needed if you use the jQuery version". And the third is the js that the external page that has an iframe needs. Then there are three corresponding .map files. I don't know too much about js...are these map files needed for the js to work? 

I figured it wouldn't hurt to have these here for reference (the other two js beyond the one that has been put into WebworkClient.pm.) But I'll take your advice about removing them and the map files.

goehle:
Sorry it took a while to get back to this.  I've only now managed to wrangle the early fall semester craziness into something manageable.  

-  I think its fine to have both the "resizer" and "contentWindow" js files.  Its not unreasonable that the content page actually using the iframe resizer would want to get the contentWindow js files from the webwork servers. 
-  As far as I understand it the map files make it so that the js console can still split out relatvent error information, even if the .min.js files are used, so they can be kept around. 
-  You might as well get rid of the ie8 files.  I wouldn't trust most of what we do on IE 8 and its safe to say its not supported.

Alex-Jordan:
OK, I removed the ie8 files. No need to worry about the time frame; I know you've been busy.

PR #623, opened 2015-08-27
Develop rfbug2

mgage:
changes to default.config to point to location of CAPA macro files.  The course must have a link "capaLibrary" in its template directory pointing to the Contrib/CAPA directory.


conf/defaults.config
```diff
@@ -1046,8 +1046,8 @@ $pg{specialPGEnvironmentVars}{use_javascript_for_live3d} = 1;
 
 # Locations of CAPA resources. (Only necessary if you need to use converted CAPA
 # problems.)
-$pg{specialPGEnvironmentVars}{CAPA_Tools}             = "/opt/webwork/libraries/CAPA/CAPA_Tools/",
-$pg{specialPGEnvironmentVars}{CAPA_MCTools}           = "/opt/webwork/libraries/CAPA/CAPA_MCTools/",
+$pg{specialPGEnvironmentVars}{CAPA_Tools}             = "$courseDirs{templates}/capaLibrary/macros/CAPA_Tools/",
+$pg{specialPGEnvironmentVars}{CAPA_MCTools}           = "$courseDirs{templates}/capaLibrary/macros/CAPA_MCTools/",
 $pg{specialPGEnvironmentVars}{CAPA_GraphicsDirectory} = "$webworkDirs{htdocs}/CAPA_Graphics/",
```
goehle: should there be macros here?  In the comment above you say to make a link from capaLibrary to CAPA, but then macros/CAPA_tools wont point to CAPA_tools.

jwj61:
If the new standard is to have a Contrib link in the templates directory, then maybe paths should work off of that since the capa things are there.  Then it would take less, if any, configuration.

conf/defaults.config
mgage: so the current location of CAPA_Tools is Contrib/CAPA/macros/CAPA_Tools.  (An earlier incarnation had them in a different location. )

mgage:
we could do that.  But you already have to create a capaLibrary link in templates pointing to 
Contrib/CAPA in order to get the CAPA button.  And it's possible that some people have CAPA in non-standard places.  We did at UoR until recently.  I think what I've got is reasonable for now.

PR #624, opened 2015-08-27
Editor Tweaks

goehle:
This pull request does three things: 

-  Added some checks to the Homework Sets Editor so that if you are in edit mode with insufficient permissions it doesn't spit out undefined variable errors.  
  -  To test go the homework sets as a ta and click the pencil.  Before the patch you should get errors.  Patch and then reload the page. There should be a nicer message. 
-  Added checks so the pencil icons are not displayed unless you have permissions. 
  -  To test, go to Classlist Editor and Hmwrk Sets Editor as a professor, the pencils should be there.  As a TA they should not.  
-  Added a little tweak to math4 so that the side navigation bar won't be shown in a half formatted state.  
  -  This is hard to test.  Go to the library browser and load up a bunch of problems.  On some setups this will cause the left hand navigation bar to "half load" so that it is there but not formatted correctly.  After this change the bar will be hidden until after its formatting js has run.

### DIFF
--- a/htdocs/themes/math4/system.template
+++ b/htdocs/themes/math4/system.template
@@ -106,7 +106,7 @@ var tabberOptions = {manualStartup:true};
 <div class="row-fluid">
 
 <!-- Navigation -->
-<div id="site-navigation" class="span2" role="navigation" aria-label="main navigation">
+<div id="site-navigation" class="span2 hidden" role="navigation" aria-label="main navigation">
   <!--#if can="links"-->
         <div id="site-links">
 	  <!--#links-->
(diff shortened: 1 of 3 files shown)


mgage:
Checked the pencil behavior -- that is ok now and since you can't get to the classlistEditor easily now as a TA the error message is moot.

I couldn't directly test the last tweak but I notice now that the left side bar does not render until after the library problems in the left half of the page.  This can give the impression that the page is hung if there are a lot of library problems but there is a progress bar in most browsers that gives some visual indication that things are being processed.

PR #627, opened 2015-09-04
Develop tls updates

mgage:
Make it possible to disable tls (the new version of ssl) when communicating with the smtp server.
See bug #3407

Make it possible to set the email address in the default generic Error message to something other than ServerAdmin.  (ServerAdmin, alias $ENV{SERVER_ADMIN}; ) This is set in webwork.apache2(.4)-config.dist

Minor: Fixed the reference of SOAP::Apache2   to SOAP::Apache (since the module SOAP::Apache2 is not being maintained and is no longer needed.) This change is in webwork.apache2-config.dist.

Testing may be a bit difficult unless your smtp server uses TLS by default and the certificates are not correctly set up.  Instead test that no harm has been done by using email the instructor feedback, send mail to students and the automatic email instructor that JITAR sends when review problem has not succeeded.


conf/webwork.apache2.4-config.dist
```diff
@@ -154,7 +154,7 @@ if ($webwork_url eq "") {
 
  # WEBWORK SOAP CONFIGURATION
  <Location /webwork2_rpc>
-         PerlHandler Apache2::SOAP
+         PerlHandler Apache::SOAP
          SetHandler perl-script
```
goehle: When did Apache2::SOAP become Apache::SOAP?

courses.dist/modelCourse/templates/Contrib
```diff
@@ -0,0 +1 @@
+/opt/webwork/libraries/webwork-open-problem-library/Contrib
```
goehle: Should this be here?

conf/webwork.apache2.4-config.dist
mgage: See the discussion at http://webwork.maa.org/moodle/mod/forum/search.php?search=Apache2%3A%3ASOAP&id=3

courses.dist/modelCourse/templates/Contrib
mgage: This is my attempt to make access to CAPA automatic "out of the box".  This will work for 90% of the installations but will fail for non-standard placement. 

Making this absolutely foolproof is tricky, but the other possibility is to define the Contrib link when the LibraryBrowser is first viewed. (this is already done with Library link).  That is also not completely foolproof because sometimes people try to create a homework set before they have visited the Library Broswer page

mgage:
This also fixes a rare bug where a url which just hangs (e.g. because it is behind a firewall) causes the problem to timeout -- leading to misleading error messages.  This can be hidden because check_url() is called in many places (particularly inside alias) to insure that resources actually exist for the problem.

lib/Apache/WeBWorK.pm
```diff
@@ -192,7 +192,7 @@ sub htmlMessage($$$@) {
 	my $backtrace = htmlBacktrace(@backtrace);
 	
 	my $admin = ($ENV{SERVER_ADMIN}
```
goehle: Should this be WEBWORK_SERVER_ADMIN ?

lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
```diff
@@ -837,6 +837,7 @@ sub mail_message_to_recipients {
 			$error_messages .= "There were errors in processing user $recipient, merge file $merge_file. \n$@\n" if $@;
 			#warn "message is ok";
 			my $mailer = eval{ Mail::Sender->new({
+					tls_allowed => $ce->{tls_allowed}//1, # the default for this for  Mail::Sender is 1
```
goehle: Do you also need to copy this line over to https://github.com/openwebwork/webwork2/blob/develop/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm#L756

mgage: Good catch.  I've copied this configuration over to the Mail::Sender call in ProblemUtil.pm as well.

lib/Apache/WeBWorK.pm
mgage: Yes.   I've updated the pull request to include this patch.

mgage:
Give this a try.  I think I've got a more robust way of defining WEBWORK_SERVER_ADMIN.  It turns out that on my laptop I have the same problem you had about SERVER_ADMIN not being defined in webwork.apache-config.  I'm not sure why it worked on math.webwork.rochester.edu

goehle:
The changes work for me.  My guess is that this worked on rochester because the environment variable was set explicitly.  If the SERVER_ADMIN is set by the apache ServerAdmin command it may not become an environment variable until a child process is started.  In any case, this does what its supposed to and preserves the old behavior.

PR #628, opened 2015-09-04
adds a check mode to WebworkClient.pm, applies it to simple format in…

Alex-Jordan:
… lieu of submit mode, and removes preview from simple format

Makes it so that when using the format=simple mode, some of the items from https://groups.google.com/forum/#!topic/ww-mbx/Xt3IGA4m0ME are applied. Specifically, removes preview button, implements a check button in place of the submit button, and makes it so that the check button doesn't provide confusing messages about partial credit being available.

To see the difference, open the first problem at http://spot.pcc.edu/~ajordan/ww-mbx/html/section-1.html, which uses a server on the current develop, and compare to http://spot.pcc.edu/~ajordan/ww-mbx-dev/html/section-1.html which uses a server that has these commits.
(The contents and behavior of these addresses may change if much time passes.)

The code for the submit mode is still there in case there is a future where cookies allow for scores to be recorded. Question: should I change the other modes (standard, sticky) to also swap out submit mode with check? And remove preview?

### DIFF
--- a/lib/WebworkClient.pm
+++ b/lib/WebworkClient.pm
@@ -549,6 +549,7 @@ sub formatRenderedProblem {
 	my $session_key      =  $rh_result->{session_key}//'';
 	my $displayMode      =  $self->{displayMode};
 	my $previewMode      =  defined($self->{inputs_ref}->{preview});
+	my $checkMode        =  defined($self->{inputs_ref}->{WWcheck});
 	my $submitMode       =  defined($self->{inputs_ref}->{WWsubmit});
 	my $showCorrectMode  =  defined($self->{inputs_ref}->{WWgrade});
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
OK, I completely forgot that the WBWK server used in the second link is not open to the world. So to see the difference in behavior, pull to a server where you have an example of the anonymous WeBWorK use, and compare to the first link. Another difference will be the iframe resizing. To see that in your own example, you need to insert the javascript for it into your example.

goehle:
I checked this out and it works for me.  

A comment for future consideration is that we should probably figure out a better way to include these templates.  In particular I see a need for site administrators to add/modify their own templates for this.

PR #629, opened 2015-09-05
Add snippet to authenticate in LTIBasic which sends xmlrpc calls to th…

mgage:
…e SUPER:authenticate

To test:  Without this patch the page

http://michaelgage.blogspot.com/2015/06/whether-writing-full-text-book-or-just.html

will not render the embedded WW problem.  You get an authentication error message.

With the patch the problem renders correctly.


mgage:
You will want to modify that problem to point to your own opaque server.  The blog page points to hosted2 which is running code with the LTI patch installed.

PR #630, opened 2015-09-06
Tell which term is not defined in the taxonomy in library alert

mgage:
at least you'll know why the alert is annoying you. It will tell you the missing term.


mgage:
Make sure library tags are enabled.
Find a problem with a defined subject or create one using the library tags. 
Edit the problem to make the DBsubject entry non-sensical.
If you view the problem you will get a popup alert saying that the value is not in the taxonomy.
With the patch you get the same message but it also tells you the offending value.

PR #631, opened 2015-09-09
Develop rename update

mgage:
This adds the ability to change the courseTitle and the courseInstitution in addition to changing the courseID for the course. Changing the courseID involves renaming several tables in the database, changing the title or the institution simply changes a field in the database. title and institution were set when a course was created and recorded in the database but their was no way to change their values.

To test:  

There are four items to be chosen in order to rename the database.  oldCourseID (from a scrolling menu)
newCourseID, newCourseTitle and newCourseInstitution.  Each of the last three has both a checkbox to indicate that it is to be changed and an input field for the new value. 

You always need to select an old courseID but there are many combinations of checkbox and input field that need to be checked.  If the checkbox is checked and the corresponding input field is empty there should be an error message (after submit).  If the input field has data but the checkbox is not checked then no operation should occur.  If the checkbox is not checked and the input field is empty there should not be an error.  Finally if the checkbox is checked and the input field is not empty then the report should indicate that the title or institution or ID has been changed. There are 64 combinations to check.  

After the initial submit there should be an accurate report of what errors need to be corrected (e.g. a field needs to be filled in) and what changes will be made. 

Cancel buttons should take you back to the first rename page with the checkboxs and input fields in their original state.

If you have direct access to the database make sure that reported changes are actually made.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -1055,8 +1055,8 @@ sub rename_course_confirm {
 	my $dbLayoutName = $ce->{dbLayoutName};
 	my $db = new WeBWorK::DB($ce->{dbLayouts}->{$dbLayoutName});
 	my $oldDB =new WeBWorK::DB($ce2->{dbLayouts}->{$dbLayoutName});
-	my $rename_oldCourseTitle = $oldDB->getSettingValue('courseTitle');
-	my $rename_oldCourseInstitution = $oldDB->getSettingValue('courseInstitution');
+	my $rename_oldCourseTitle = $oldDB->getSettingValue('courseTitle')//'""';
+	my $rename_oldCourseInstitution = $oldDB->getSettingValue('courseInstitution')//'""';
 	
 	my ($change_course_title_str, $change_course_institution_str)=("");
 	if ( $rename_newCourseTitle_checkbox) {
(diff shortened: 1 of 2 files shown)


mgage:
Make sure $rename_oldCourseTitle is defined even if it doesn't exist in the database   (use ____  for default)  

Same thing for course Institution

result will look something like 

Change course institution from ___ to American University

mgage:
Should I comment out the "harmless" error messages?   They alert you to the extra "published" field in your tables -- but most people don't care and the messages look ugly.

goehle:
Your above comment seems to indicate that you took care of having an uninitialized $rename_oldCourseTitle, but I'm still getting uninitialized variable errors if I rename a course which doesn't have an existing course title.  Did those changes make it to github? 

Also,  I would remove error messages about extra columns.  The main reason is that you can't fix the issue in any reasonable way so the error message can't actually be dealt with by an admin.

goehle:
Other things: 
-  I don't know if you care but the behavior is a little strange if you select a course and don't click any of the checkboxes and don't enter any data. 
-  Super nitpicky, but if you are just changing the course name and institution name then there isn't any white space between the confirmation text and the confirm button.   Also the "Rename" text is h4 but the "Change title" and "Change institution" text is just a div.  
These are super tiny and you can ignore them.  The uninitialized variable thing should be dealt with. 

Other than that everything seems to check out.  
-  The changes do get made in the database correctly. 
-  If you click any of the check boxes and don't have data it gives you an error. 
-  If there is data in a field with no check box then it is ignored. 
-  The cancel button does bring you back to the previous page and the intermediate page does give you a preview of the changes to be made.

goehle:
Also, these changes turned out to be reasonably minor and it adds functionality that is not currently available.  So would it be worth it to back port this to 2.11?

mgage:
I've made cosmetic changes in the response.

goehle:
All of the changes here check out so I'm going to pull the 2.11 version of this.  We regularly pull 2.11 into develop so I won't pull this since the changes will make there way into develop naturally.

PR #633, opened 2015-09-10
Upgrade setOrientation

apizer:
Second try. Upgrade setOrientation so that the information presented matches what students see in WeBWorK 2.11

### DIFF
--- a/courses.dist/modelCourse/templates/setOrientation/course_info.txt
+++ b/courses.dist/modelCourse/templates/setOrientation/course_info.txt
@@ -3,9 +3,13 @@
 <p align="justify">
 
 If you haven't already done so, you should change your password to
-something other than your student ID.  To do this, click on the
-Password/Email link in the list at the far left, and follow the
-directions on that page.
+something other than your student ID or whatever your instructor
+used as your initial password.  To do this, click on the
+User Settings link under MAIN MENU in the list at the far left, 
+and follow the directions on that page. Note you can collapse and
+uncollapse the MAIN MENU by clicking on the boxed angle bracket just 
+above the boldface course name so if you don't see the MAIN MENU click 
+on the boxed angle bracket to make MAIN MENU reappear.
 
 <p align="justify">
 
@@ -14,9 +18,4 @@ in the white area at the left.  This will take you to the the first
 homework assignment, which will help you to learn how to use WeBWorK
 effectively.
 
-<p align="justify">
 
-If you wish to get a printed copy of one of your assignments, you can
-do so from this page by pressing the "Download Hardcopy" button.  You
-will be taken to a page where you can select what information to
-include in the set, and which set to download.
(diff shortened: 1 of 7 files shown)


apizer:
git rebase didn't work. I got 
fatal: Cannot update paths and switch to branch 'new_update_Orientation2' at the same time.
but doing things the "long' way was't bad.

goehle:
You dont make a new branch.  It switches the current branch.  In any case this is fine.  

I hate to keep adding to this pull, but while we are messing around with this stuff...

Do we want to replace the java applet problem in setDemo; i.e. problem https://github.com/openwebwork/webwork2/blob/develop/courses.dist/modelCourse/templates/setDemo/prob6b.pg

Java applets are getting phased out and we could probably find a better demo problem to show things off.  My suggestion would be a scaffolding problem and/or a live graphics problem.

apizer:
Maybe I should have run git fetch before git rebase so that my system knew about release-2.11.  I had to do that before my other (successful) attempt.  Oh well it worked.

I really want to keep this problem as is for historical reasons.  As it says in the site info:

9. Java example. This gives an example of incorporating a java applet which can be used experimentally to determine answers for WeBWorK questions. This example is of historical interest since it comes from the first site after Rochester, Johns Hopkins University, to use WeBWorK. It currently gives an example of what happens when a WeBWorK problem called an applet residing on a server that no longer exists. 

Adding a few additional problems would be fine.  It has been years since we did that. Set Demo basically shows what WeBWorK could do in the 90's.  Does anybody have specific problems that would be good?  Is it correct that Just-In-Time requires a separate set so you can not e.g. have a 2 problem Just-In-Time subset of a standard set?

goehle:
Yes, Just-In-Time is a separate homework type, so you cant have two just in time homeworks. 

I think scaffolding and livegraphics are good candidates.  Others I can think of are questions involving sage or geogebra. 

As for the java applet.  I appreciate the historical reasons.  I am slightly wary of having broken things be part of the "demo" set, but I won't make a fuss if its kept in.

goehle:
When testing this I ran across a problem that needs to be fixed.  With the addition of the "Contrib" link to modelCourse the link to "Library" isn't getting created any more.  I don't know what created it, though, or how that should be fixed.  Do you know Mike?

goehle:
Never mind, I misunderstood how that worked. Its created by the library browser the first time you use it and that is still fine.  This pull looks good.

jwj61:
Yes, but the model course should come with a Library link since sometimes people will use a Library problem from a set before using the library browser.

Or, I suppose one could insert a check, if looking for a problem with path starting with Library/... and it is not found, see if the Library link exists and create it if it doesn't.

goehle:
I was thinking the same thing.  In particular I don't really like how the Contrib link is included since it has a hard coded path.  I'll put it on the list to  think about for a future pull.

PR #634, opened 2015-09-11
Fix bug  #3415.  The line 2093 was added (compare with editor2 at the…

mgage:
… same location).  The other changes are simply white space changes to regularize the indentation pattern.  Fixes the error reported in #3415 (that saving and adding the problem to the set doesn't behave correctly when the file being saved already exists on the disk).


goehle:
fixed as advertized.

PR #635, opened 2015-09-13
Clean up a warning about uninitialized  for course_info.txt

Alex-Jordan:
If I go to a course on develop right now, and click to edit the course information text, I get a warning message about $setID not being defined. It arises when the editor is looking to see if the file name is of the form set$setID as part of a check on how to process the file name. This adds a check to ensure $setID is defined before doing this other check, and the warning goes away while editing course_info.txt

Perhaps this should apply to 2.11 too? It's not critical, but it gives the wrong impression to a faculty user who just wants to edit their course info that something is wrong.


goehle:
This should go into release-2.11.  Could you rebase and submit a new pull?

Issue #637, opened 2015-09-22
Conditional Release seems broken in develop


Alex-Jordan:
We are using develop (I'm afraid I won't have time to explore this in 2.11 for a while). 

It appear to us that Conditional Release is broken. If we set up a problem set A to depend on completion of set B, then all appears well until you actually complete set B, and try to work on set A. You can get to the list of problems for set A, but then when you click on Problem 1, you get an error message like this one:

You do not have permission to view unopened sets at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 529.

Turning off conditional release appears to give access to set A, so it seems certain that CR is the culprit.

goehle:
This seems like a regression caused by JITAR. I can't test right now, but try removing the `$set->set_id` from https://github.com/openwebwork/webwork2/blob/develop/lib/WeBWorK/ContentGenerator/Problem.pm#L518

PR #639, opened 2015-09-23
Conditional Release Bug

goehle:
This fixes a regression issue with conditional release. 
To Test:  
-  Enable and set up conditional release for a set. 
-  Make sure you have satisfied the conditional release conditions.  You should be able to access the set problem list, but if you click on a problem you will get an authorization error. 
-  After the patch you should be able to access the problems. 

This pull also tweaks the css on the pg editor pages a bit so the buttons on the top of the editor aren't stacking.  

This fixes issue #637

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -985,7 +985,7 @@ table caption{
 }
 
 #editor{
-	width: 585px;
+	width: 750px;
 }
 
 #editor .pg_editor_input_span{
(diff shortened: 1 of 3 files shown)


goehle:
Tacking another bug fix on this instead of making a new pull.  I've removed mathjax formatting of past answers.  
-  Test by visiting the past answers page.  Formula type answer should no longer be rendered using mathjax.  Essay answers should still be rendered.

mgage:
You had to be a student (not an instructor) to get the authentication error on the conditional release bug.  The fix works as advertised.  The mathjax tweak for past answers is also correct.

PR #641, opened 2015-09-29
Develop gavinfixes

mgage:
Fixes authored by Gavin LaRose.  (There is a typo in the first commit -- the change is to GatewayQuiz not to Hardcopy. The second commit concerns changes to Hardcopy.)

 I'm attaching the
diffs (diff old new) for Hardcopy.pm and GatewayQuiz.pm.  The change in
the latter is to deal with the checkbox multiple choice problems; in the
former to show the "show correct answers" button when a set version is
completed but the regular set is not.  I did not check to be sure that the
logic the prohibits actually showing the correct answers when they
shouldn't be shown is still there.  

Thanks,
Gavin

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -2178,8 +2178,10 @@ sub body {
 				my $curr_prefix = 'Q' . sprintf("%04d", $probOrder[$i]+1) . '_';
 				my @curr_fields = grep /^$curr_prefix/, keys %{$self->{formFields}};
 				foreach my $curr_field ( @curr_fields ) {
-					print CGI::hidden({-name=>$curr_field, 
-							   -value=>$self->{formFields}->{$curr_field}});
+ 					foreach ( split(/\0/, $self->{formFields}->{$curr_field}) ) {
+ 						print CGI::hidden({-name=>$curr_field, 
+ 							   	   -value=>$_});
+ 					}
 				}
 				# finally, store the problem status for 
 				#    continued attempts recording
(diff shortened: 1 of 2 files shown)


goehle:
Could you elaborate on what this is suppose to change and how to test it?  I'm afraid I dont follow.

mgage:
Gavin asked me to submit these changes for him.  I believe this is the issue:

Hi all,

Follow-up: I believe this is in the following lines (c. 2170) in
GatewayQuiz.pm.
 foreach my $curr_field ( @curr_fields ) {
         print CGI::hidden({-name=>$curr_field,
			     -value=>$self->{formFields}->{$currField}});
 }

The value in $self->{formFields}->{$currField} is a "\0" joined string of
answers from the form.  Thus, I think the solution is to take
 foreach my $curr_field ( @curr_fields ) {
 	  foreach ( split(/\0/, $self->{formFields}->{$currField} ) {
                 print CGI::hidden({-name=>$curr_field,
			             -value=>$_});
	  }
 }

If anyone sees something conspicuously wrong about that, please let me
know.

Thanks,
Gavin

On Fri, Sep 11, 2015 at 01:50:46PM -0400, P. Gavin LaRose wrote:
Hi all,

We've discovered an issue with gateway scoring on multiple-page gateways:
with a checkbox_multiple_choice object, if one selects options and moves
to another page, the selections are lost.  This appears to be because of
the encoding of answers on the subsequent page (problem answers are stored
as a hidden field, and the spacing characters are coming as special
characters).

Does this ring a bell for anyone that indicates what is going on?

Thanks,
Gavin

goehle:
I tested the two components of the patch. 
-  Now checkbox problems have the checks saved correctly if the problem calls for more than one check.  
-  Now students can see solutions on tests they have finished.  The permissions respect the various gateway options as well.

PR #645, opened 2015-10-07
Properly Seed Apache Children

goehle:
An apparently known "feature" of how mod_perl and srand interact is that different mod_perl child processes all get the *same* seed for generating random numbers.  http://marc.info/?l=apache-modperl&m=123904225030744&w=1  Because of this random numbers generated by the clients all follow the same pattern.  The most obvious side effect of this is that logging in as a practice user, logging out, waiting 10-15sec for the child process to die, then logging back in as a practice user always gives you the same practice user.  However, there is a very good possibility that this is having other side effects and it certainly has negative security implications.  

The proper solution to this is to add something which calls srand once per child, which is what I have done by adding the line `PerlChildInitHandler "sub { srand }"` to the apache configuration files.  To test:
- Import the demo classlist to enable guest login. 
- Log out, log in as guest, note your practice user number, log out, wait 10-15 sec, and log in as guest again.  You should have same practice user. 
-  Download the patch and be sure to *update* your non dist apache configuration file.  Or just add the PerlChildInitHandler line to your apache config. 
- Repeat the first test.  You should get a different practice user each log in now. 

My guess is that this pull is appropriate for release-2.11 but its quite possible that some feature or another actually depended on different child processes all receiving the same seed.  However, such a feature would have to run early in the life of a child process before the random number generators diverged.  On the other hand this means that up until now there has been significantly less randomness to things like authentication tokens.  I would say this is a change worth making, but there might be some gotchas.  

P.S.  It turns out srand is called before cookie keys are created, so this will have less of an impact than it might have because at least for some children srand was getting run pretty early.

### DIFF
--- a/conf/webwork.apache2-config.dist
+++ b/conf/webwork.apache2-config.dist
@@ -37,7 +37,7 @@
 
 ScriptAliasMatch /webwork2_course_files/([^/]*)/show-source.cgi/(.*) /opt/webwork/courses/$1/html/show-source.cgi/$2
 
-
+PerlChildInitHandler "sub { srand }"
 
 PerlModule mod_perl2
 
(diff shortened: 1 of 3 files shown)


mgage:
Since my devel servers are not heavily used I needed to restart the server, rather than wait 20 seconds for a new child.  With that I was able first to verify that the guest login starts with the same practice user before the patch and with a random choice after the patch.

PR #646, opened 2015-10-20
Develop attempt tables update

mgage:
This pull request is for the AttemptTables.pm which abstracts the code which presents the result of a student's attempt to answer the question.  It replaces (or will replace -- I have a few commits to add still on this project) the attemptResults subroutine in Problem.pm -- hopefully this is a good start in simplifying that file.  I'm submitting this now in order to separate the attemptTables updates from further work being done on WebworkClient, OpaqueServer and the various rendering scripts: renderProblem.pl, checkProblem.pl, renderProblem_rawoutput.pl and standAlonePGRenderer.pl.  Changes to these latter files are in develop_wwclientsup5 -- which was getting rather long. 

To test:

(1) run   webwork2/clients/renderProblem.pl  input.txt   
make sure that it renders the file correctly and that when you submit the answers the answers are checked and the results properly returned.

(2) Do the same test on the TestWW*.html  family of files -- make sure the problem renders and that when the answers are submitted they are graded and the results are correctly returned.


mgage:
replace this with the pull request develop_attemptTables_update2

PR #647, opened 2015-10-20
Wwclientsup5

mgage:
I'm afraid this pull request has gotten a bit out of hand.  It changes 26 files but many of those changes are minor (changing the key "password" to "course_password" for example to avoid future confusion).  

The biggest changes are:
1. Modification so that the path to the source file is relative to the templates directory of the course on the server which is rendering the problem (remotely).  Make sure that the templates directory is described correctly.   This hopefully resolves issues that we were having using includePGfile and includePGproblem.
2.  (a) In webwork2/clients run ./check_problems_in_dir.sh   path_to_Union_subdirectory of OPL -- this should check all of the problems in that subdirectory and list the problems that don't work in webwork2/DATA/bad_problems.txt  Union problems that contain -text in the title are not complete problems and should fail the rest should succeed.  You can run "tail -f bad_problems.txt" in a separate window to watch the progress.
    (b) run renderProblem.pl input.txt and   renderProblem.pl bad_input.txt in webwork2/clients to make sure they work.  Answer the questions and submit to make sure that the answers are graded.


mgage:
replace with develop_wwclientsup5

PR #656, opened 2015-11-22
Develop wwclientsup5

mgage:
I'm afraid this pull request has gotten a bit out of hand. It changes 26 files but many of those changes are minor (changing the key "password" to "course_password" for example to avoid future confusion).

The biggest changes are:
1. Modification so that the path to the source file is relative to the templates directory of the course on the server which is rendering the problem (remotely). Make sure that the templates directory is described correctly. This hopefully resolves issues that we were having using includePGfile and includePGproblem.
2. (a) In webwork2/clients run ./check_problems_in_dir.sh path_to_Union_subdirectory of OPL -- this should check all of the problems in that subdirectory and list the problems that don't work in webwork2/DATA/bad_problems.txt Union problems that contain -text in the title are not complete problems and should fail the rest should succeed. You can run "tail -f bad_problems.txt" in a separate window to watch the progress.
(b) run renderProblem.pl input.txt and renderProblem.pl bad_input.txt in webwork2/clients to make sure they work. Answer the questions and submit to make sure that the answers are graded.


goehle:
This needs to be merged before I can test the wwclientsup series of pulls.

goehle:
Sorry, that should have said "had conflicts resolved".

mgage:
OK.  I fixed the conflict. (minor changes of format).  This now has all the changes made in develop since wwclientsup5 was first posted. 

Check for entries that involve $ or mr_u   or u'.   Those work in develop but stop working by the time you get to wwclientsup7  so somewhere the patch to the HTML entities filter got screwed up.  It might have been in this pull request already.

mgage:
I suggest concentrating on making sure that the attemptsTables works alright with this pull request  when viewing a problem within webwork2.  Also make sure that all of the Library functions (which use instructorXMLRPC ) work correctly.  And make sure that showmeanother works properly. 

Considerably more work was done for the next PR on the renderProblem client (in fact it is replaced by sendXMLRPC ) so I suggest not checking that at all but check sendXMLRPC when you pull wwclientsup7 or wwclientsup8  For now just make sure that nothing is broken standard webwork2 usage (particularly the library and showmeanother).  Then go on to either wwclients7 or wwclients8 -- I think it might be ok to skip to wwclients8 because it is not that far beyond wwclients7.

mgage:
The entry of $  and mr_u  and other HTML entities issues already appear in this PR.  If you can indicate to me where the fix has to be (re) applied I'll do that and update this PR.

goehle:
If you merged develop into this branch then whatever fixes were made in develop (and 2.11) are already reconciled with your changes.  It must be something about this branch specifically which is causing trouble.

mgage:
Where did you put the filter which checks for HTML entities in student input?  Is this in Problem.pm? or somewhere else? If it's a subroutine can you remember what it is called?  I can search for it.

goehle:
We don't filter student input so much as encode student input.  The subroutine which encodes pg and html special characters is in PGcore at https://github.com/openwebwork/pg/blob/master/lib/PGcore.pm#L577

mgage:
OK.  Thanks.  Have to go run the dog.  I’ll fix this when I get back.

Take care,

Mike

conf/webwork.apache2-config.dist
```diff
@@ -78,13 +78,13 @@ my $webwork_htdocs_dir = $ce->{webwork_htdocs_dir};      # e.g.  /opt/webwork/we
 my $webwork_courses_url = $ce->{webwork_courses_url};    # e.g.  /webwork2_course_files
 my $webwork_courses_dir = $ce->{webwork_courses_dir};    # e.g.  /opt/webwork/courses
 
-my $webwork_server_admin_email = $ce->{webwork_server_admin_email};
+my $webwork_server_admin_email;
 if ($webwork_server_admin_email) {
 	$webwork_server_admin_email = $ce->{webwork_server_admin_email};
-	warn "set email from site.conf";
 } else {
```
goehle: Does this if statement still work?  It seems like it would always register as false.  Should it check `$ce->{webwork_server_admin_email}`

conf/webwork.apache2.4-config.dist
```diff
@@ -79,10 +78,9 @@ my $webwork_htdocs_dir = $ce->{webwork_htdocs_dir};      # e.g.  /opt/webwork/we
 my $webwork_courses_url = $ce->{webwork_courses_url};    # e.g.  /webwork2_course_files
 my $webwork_courses_dir = $ce->{webwork_courses_dir};    # e.g.  /opt/webwork/courses
 
-my $webwork_server_admin_email = $ce->{webwork_server_admin_email};
+my $webwork_server_admin_email;
 if ($webwork_server_admin_email) {
 	$webwork_server_admin_email = $ce->{webwork_server_admin_email};
```
goehle: Same as above.

lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -284,182 +271,260 @@ sub attemptResults {
...
-
-		push @tableRows, $row;
-	}
-	
+# Create AttemptsTable object	
+	my $tbl = WeBWorK::Utils::AttemptsTable->new(
+		$answers,
+		answersSubmitted       => 1,
+		answerOrder            => $pg->{flags}->{ANSWER_ENTRY_ORDER},
+		displayMode            => 'MathJax',
```
goehle: Should this be hard coded to MathJax?

lib/WebworkClient.pm
```diff
@@ -178,12 +267,13 @@ sub xmlrpcCall {
 		-> proxy(($self->url).'/'.REQUEST_URI);
 	};
 	print STDERR "WebworkClient: Initiating xmlrpc request to url ",($self->url).'/'.REQUEST_URI, " \n Error: $@\n" if $@;
-	# turn of verification of the ssl cert 
+	# turn off verification of the ssl cert 
 	$transporter->transport->ssl_opts(verify_hostname=>0,
-	    SSL_verify_mode => IO::Socket::SSL::SSL_VERIFY_NONE);
+	    							SSL_verify_mode => 'SSL_VERIFY_NONE'
+	);
```
goehle: This part needs to be reverted.  It should be 
```
 SSL_verify_mode => IO::Socket::SSL::SSL_VERIFY_NONE
```
and `use IO::Socket::SSL;` has to be added back in.  The reason is that new versions of this package produce errors when the string 'SSL_VERIFY_NONE' is used so you have to use the appropriate constant.

goehle:
I'm trying to test this now.  When I use the Library Browser and I change subjects the chapters don't load.  When I look at the details of the xml request I see that the response is 
```
Chapters loaded.
```
If I do that on a regular develop server I get a JSON object as the response with "Chapters loaded" as the server response but also with a result_data field that contains an array.

lib/WeBWorK/ContentGenerator/Problem.pm
goehle: I can confirm that the answer table is always rendered in mathjax irregardless of what display mode is used.

goehle:
The attempts table on the problem page doesn't protect against xss anymore.  Enter
`
<script>alert('hi');</script>
` 
into an answer rule.  You should see a 'hi' alert.  Old versions used to protect against this.  I'm confused since I thought the encoding happened in PG, so there must be something  going on.  It may be related to your $ issue.

conf/webwork.apache2.4-config.dist
mgage: Good catch.  I've fixed these two.  There were some changes in the webwork.apache2-config files that weren't copied to webwork.apache2-config.dist.

Test this to make sure that there are no undefined errors even if the webwork_server_admin_email variable in site.conf is left undefined (commented out) -- in that case you should fall back to using the SERVER_ADMIN address in the messages that ask you to report errors in a problem. 

That's what should happen if someone updates without checking for changes in the site.conf vs site.conf.dist file.

lib/WeBWorK/ContentGenerator/Problem.pm
mgage: You are right.  In addition, once I allowed images mode, I found that the images mode is not yet working correctly.  It prints tex output instead of running it through the image generator.

I'll work on a fix for that.

mgage:
OK -- image mode works now.  (I needed to render the images AFTER the formulas were pulled in to the 
attempts table.  :-(

lib/WebworkClient.pm
mgage: Return to using IO::Socket::SSL::SSL_VERIFY_NONE  has been committed.

mgage: This has now been fixed.

mgage:
Current status.   The behavior of the Library browser has not yet been rechecked.  
The xss exploit has not yet been rechecked. 
The $ _ issues are not present when I use the develop branch of pg only when I use versions with additional commits so I'm leaving that issue for later.

mgage:
I've added in some of the commits from develop_wwclientsup7 which stablized the way in which xmlrpc requests are authenticated.  (Included changing the key "password" to "course_password" because there were to many password variables being used.  (e.g. site_password, pw etc. ) The library works now. This may cause conflicts when we pull develop_wwclientsup7 but I don't think they'll be too bad. 

Only think left now is xss.

mgage:
I can’t reproduce this error.  Can you give me some more detail?  

(You can email — it looks like the github issue editor might have scrubbed out some 
of the command you used. )

lib/WeBWorK/Utils/AttemptsTable.pm
```diff
@@ -108,29 +122,60 @@ sub _init {
...
+
+sub maketext {
+	my $self = shift;
+	return &{$self->{maketext}}(@_);
+}
 sub formatAnswerRow {
 	my $self          = shift;
 	my $rh_answer     = shift;
+	my $ans_id        = shift;
 	my $answerNumber  = shift;
```
goehle: All of the ||'s below should be //'s since 0 is a reasonable answer.

goehle:
The library browser is working better.  There are still a couple issues: 

-  If I use the rerender/rerandomize button I get artifacts like
```
{"server_response":"  /  ","result_data":""}
```
before and after the problem body. 
-   I also see text like 
```
(1 point) Library/UMN/calculusStewartET/s_14_4_prob01.pg
```
which doesn't appear in the initial render.

mgage:
I've fixed the artifacts.  

The appearance of the (1 point) Library/UMN/calculusStewartET/s_14_4_prob01.pg is still present.
John Jones used some trick to silence beginproblem() from issuing that line -- there is probably a more
straightforward way to do it. I'll ask John Jones about it.

I also see that if you view the problem by clicking on the eye that you get the library
path but with (0 points).

conf/webwork.apache2-config.dist
```diff
@@ -81,10 +81,13 @@ my $webwork_courses_dir = $ce->{webwork_courses_dir};    # e.g.  /opt/webwork/co
 my $webwork_server_admin_email = $ce->{webwork_server_admin_email};
 if ($webwork_server_admin_email) {
 	$webwork_server_admin_email = $ce->{webwork_server_admin_email};
```
goehle: This line doesn't seem necessary.

conf/webwork.apache2-config.dist
```diff
@@ -96,9 +99,11 @@ print "webwork.apache2-config:  WeBWorK server userID is ", $ce->{server_userID}
 print "webwork.apache2-config:  WeBWorK server groupID is ", $ce->{server_groupID}, "\n";
 print "webwork.apache2-config:  The webwork url on this site is ", $ce->{server_root_url},"$webwork_url\n";
 
-
 # link to PG code libraries
 eval "use lib '$pg_dir/lib'"; die $@ if $@;
+$WeBWorK::Constants::PG_DIRECTORY = $pg_dir;
```
goehle: Is there a reason this line is in this file twice?  Also do we need both PG_DIRECTORY and PG_ROOT as environment variables?

lib/WeBWorK/Constants.pm
```diff
@@ -46,7 +46,7 @@ $WeBWorK::Debug::Logfile = $WeBWorK::Constants::WEBWORK_DIRECTORY . "/logs/debug
 # For example, this pattern prevents the dispatch() function from logging:
 #     $WeBWorK::Debug::DenySubroutineOutput = qr/^WeBWorK::dispatch$/;
 #
-$WeBWorK::Debug::DenySubroutineOutput = undef;
+$WeBWorK::Debug::DenySubroutineOutput = qr/^WeBWorK::dispatch$/;
```
goehle: Is this something you wanted to be a permanent change?

lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -284,182 +271,264 @@ sub attemptResults {
...
+		showAnswerNumbers      => 0,
+		showAttemptAnswers     => $showAttemptAnswers && $showEvaluatedAnswers,
+		showAttemptPreviews    => $showAttemptPreview,
+		showAttemptResults     => $showAttemptResults,
+		showCorrectAnswers     => $showCorrectAnswers,
+		showMessages           => $showAttemptAnswers, # internally checks for messages
+		showSummary            => $showSummary,
+		imgGen                 => $imgGen, # not needed if ce is present ,
+		ce                     => '',	   # not needed if $imgGen is present
+		maketext               => WeBWorK::Localize::getLoc("en"),
```
goehle: Should this localization be hardcoded to english?  Right now I'm getting english text in the attempts table even if I set the locality to french in course config.

goehle:
You fixed the re-rendering on the LibraryBrowser but I'm getting those same artifacts on Problem Set Detail 2.  It seems like this change here

mgage:
The beginproblem() path can be fixed for now by pulling develop_beginproblem_bugfix on the pg side.  The problem is that JJ sets the points value to a negative number when he renders the problem the first time, but not when he rerandomizes using the javascript.

conf/webwork.apache2-config.dist
mgage: ok

lib/WeBWorK/Utils/AttemptsTable.pm
```diff
@@ -108,29 +122,61 @@ sub _init {
...
+		# push @correct_ids,   $name if $answerScore == 1;
+	} elsif (($rh_answer->{type}//'') eq 'essay') {
+		$resultString =  $self->maketext("Ungraded"); 
+		# $self->{essayFlag} = 1;
+	} elsif ( defined($answerScore) and $answerScore == 0) { # MEG: I think $answerScore ==0 is clearer than "not $answerScore"
+		# push @incorrect_ids, $name if $answerScore < 1;
+		$resultStringClass = "ResultsWithError";
+		$resultString = CGI::span({class=>"ResultsWithError ResultsWithErrorInResultsTable"}, $self->maketext("incorrect")); # If the latter class is defined, override the older red-on-white 
+	} else {
+		$resultString =  $self->maketext("[_1]% correct", int($answerScore*100));
```
goehle: Could you change this to use wwRound
```
https://github.com/openwebwork/webwork2/blob/master/lib/WeBWorK/ContentGenerator/Problem.pm#L367
```
That way its consistent with other ww answer display.  Right now if you have something that has partial credit of say 12.5 then this line says 13% but the problem score is 12%

lib/WeBWorK/Utils/AttemptsTable.pm
```diff
@@ -108,29 +122,61 @@ sub _init {
...
 	my $row = join('',
-			  CGI::td({},$answerNumber) ,
-			  CGI::td({},$answerString) ,   # student original answer
-			  ($self->showAttemptPreviews)?  CGI::td({},$answerPreview):"" ,
-			  ($self->showAttemptResults)?  $score :"" ,
-			  ($self->showCorrectAnswers)?  CGI::td({},$correctAnswer):"" ,
+			  ($self->showAnswerNumbers) ? CGI::td({},$answerNumber):'',
+			  ($self->showAttemptAnswers) ? CGI::td({},$self->nbsp($answerString)):'' ,   # student original answer
+			  ($self->showAttemptPreviews)?  $self->formatToolTip($answerString, $answerPreview):"" ,
+			  ($self->showAttemptResults)?   CGI::td({class=>$resultStringClass},
```
goehle: Is there a reason you arent using `$attemptResults` here?

conf/webwork.apache2-config.dist
mgage: I see your point -- no reason to have it twice  and ENV{PG_ROOT}  should be enough.

lib/WeBWorK/Constants.pm
mgage: I guess not -- although this is the dominant use case since dispatch is very noisy

lib/WeBWorK/ContentGenerator/Problem.pm
mgage: I'll fix this.

lib/WeBWorK/Utils/AttemptsTable.pm
mgage: Well it was used in the earlier version.  In particular it allows you to turn off the score column when the preview button is clicked.

goehle: Sorry, I should have been more specific.  In line 168 you define `$attemptResults` which is very similar to what you are doing in line 176.

goehle:
Ok.  This checks out.  I'l pull it.  You should check wwclientsup7 and wwclientsup8 since they may have conflicts.

PR #657, opened 2015-11-22
Develop wwclientsup7

mgage:
Pull this after develop_wwclients5.   

You will need develop_clients7 to test the new features in the pg
develop_translator_upd PR.  Those features include various debugging checkboxes available to administrators.


conf/defaults.config
```diff
@@ -1253,26 +1279,24 @@ $webservices = {
 # in course.conf file to tune the behavior for individual courses
 # The items below can be placed in localOverrides.conf override the settings in this file
 # One can also use $showeditors{classlisteditor1}=0; to override a setting.
```
goehle: Why these changes to `%showeditors`? 
-  No `ww3`
-  No `librarybrowsernojs`
-  `pgproblemeditor1` is on by default
-  `simplepgeditor` is on by default
-  `pgproblemeditor3` is on by default

lib/WeBWorK/Constants.pm
```diff
@@ -34,7 +34,7 @@ $WeBWorK::Constants::WEBWORK_DIRECTORY = $ENV{WEBWORK_ROOT} unless defined($WeBW
 
 # If true, WeBWorK::Debug will print debugging output.
 #
```
goehle: You should probably turn off debug in the pull request.

mgage:
Starting at the top the first commits finish the process of separating the presentation of attemptsTables into a separate module.  You can use the file webwork2/t/testattemptsTables.pl ot check that it works -- but you still need to make sure it is hooked up right. (We finished most of this in debugging wwclients5)

There is good (ok pretty good) POD documentation for attemptsTable. 

There is a struggle to determine where one should look for displayMode and for problemSeed (also some other problem variables but these two are the worst).  Is it envir->{displayMode}, envir->{inputs_ref}->{displayMode} or just $displayMode?  Each way of accessing the render (renderViaXMLRPC, instructorXMLRPC, the html files, renderProblem. had a different way of addressing this. Often they defined the value in several variable but then they were not always in sync.  After some 
false starts (some of the commits reverse previous ones) I'm pretty sure we want $envir->{displayMode} to be the single central value.  When you call the renderer from an html page you have no choice but to define displayMode in an inputs_ref -- but you quickly store it in $envir->{displayMode} and delete the inputs_ref version. (Particularly with problemSeed most use cases do not want to allow the user to change the problemSeed from the webpage they are submitting. ) Inside a PG problem you can safely
assume that displayMode and problemSeed have been safely derived from $envir->{displayMode}, etc.
but elsewhere is a grey area. 

this takes us up to about commit 9e84b8d 

The main thing to test is that none of these changes affect the way that problems are evaluated.  There will be good testing mechanisms for this in a bit.  I think we already accomplished this part.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
```diff
@@ -1704,7 +1704,7 @@ sub importSetsFromDef {
 
 		debug("$set_definition_file: reading set definition file");
 		# read data in set definition file
```
goehle: All of the changes to this file are the result of a bad merge.  This removes the work done to make reduced credit dates imported and exported by to set def files (and other various bugs).   As far as I can tell they should all be reverted to develop in your pull.

goehle:
Could you post this on the github for reference?

lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -1862,25 +1776,40 @@ sub output_comments{
 		    my $comment = $userPastAnswer->comment_string;
```
goehle: The changes to output_comments should be reverted back to their develop state.  Its better just to use mathjax to render them than to try and create some sort of wonky pg problem.  (which is why I changed it)

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
mgage: I didn't work much, if at all on this file, so merging it back to the develop version shouldn't be too hard.  I can do it by hand using the bbedit merge tool -- is there a slicker way?

mgage:
Next we start building sendXMLRPC. Originally I was just going to enhance renderProblem.pl which takes a file, renders it and returns the output either through less or as HTML through a browser.  

I eventually figured out that what I was doing for this and for the embedded html iframes and for opaque  moodle clients had a lot of overlap.  So what has evolved at this point is sendXMLRPC which has problems rendered at a (possibly remote) server and standalonePGrenderer (separate project) which is very similar to sendXMLRPC but ties into PG directly to do it's rendering.  standalonePGrenderer works but is not finished so you can ignore it for now if you wish -- but explains some of the design choices for sendXMLRPC.

The pod docs for this seem to have been added in wwclient8 -- but they are essentially the same as the readme for standalonePGproblemrenderer https://github.com/mgage/standaloneProblemRenderer

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
mgage: probably a result of merge at af84d97   -- it did not go easily.

goehle: If you are certain that none of these changes were intended you can just open the develop version in a text editor, switch to this branch, and then save it again.  But you should double check that nothing here was intentional.

mgage:
Next --

there are some additions of some switches that appear on a problem page if you have admin privileges (==20).   These allow you to see debugging info for the problem easily -- e.g. answerHash table,  the resourceTable constructed by alias, and so forth.  These won't doing anything yet until you update PG --that's where the real work is going on.

I've pulled the format templates for webwork-in-iframes, sendXMLRPC et. al. into a separate directory WebworkClient/  which makes it easier to add new formats.  

By the way you need to construction the "credentials" file to use sendXMLRPC  it usually has password information so you don't want it in the main source code except for demo sites which you are willing to allow to be spammed.

lib/WebworkClient.pm
```diff
@@ -759,10 +764,58 @@ sub formatRenderedProblem {
...
 	my ($function, $details, $beginEnd) = @_;
 	$beginEnd = ($beginEnd eq "begin") ? ">" : ($beginEnd eq "end") ? "<" : "-";
 	WeBWorK::Utils::writeLog($seed_ce, "render_timing", "$$ ".time." $beginEnd $function [$details]");
 }
 
+=item pretty_print_self
+
+=cut
+
+
```
goehle: I don't suppose we can just use the PG pretty print?  Its not a huge deal either way.

lib/WebworkWebservice.pm
```diff
@@ -270,9 +268,14 @@ my $authz  = $fake_r->authz;
 	}
 	
 	die "Please use 'course_password' instead of 'password' as the key for submitting
+<<<<<<< HEAD
```
goehle: This looks like the result of a failed merge with develop and should be fixed.

lib/WebworkClient.pm
mgage: maybe, but I’d like to wait on that.  My long range hope is that we can get a version of this 
that uses only pg and as few other files as possible and can package for CPAN
as an editor for writing PG problems.  I’ve spent some time disentangling things
that were included just because they were there and it was easy to grab it from another file.
Often that’s a good thing, but it can also cause problems.

By the way I haven’t yet quite figured out how to refactor this writeRenderLogEntry
function — but it seemed like something that could wait.

lib/WebworkWebservice.pm
mgage: No.  I think I did this one on purpose and I think it works as I intended.  

I did a global change of password to course_password in all of the WebworkClient stuff so 
that it was clear whether we were talking about the course_password or the site_password.

(I left password alone in Authen because it only knows about course passwords — the concept of a site password didn’t exist for it. )

There are however still clients in the wild (e.g in html pages ) that were wired to use the 
keyword password and I wanted to warn them that the key has been changed to course_password.  That’s what this does.

You can test it with some of the TestWW.html pages in clients.


Does this make sense?
Take care,

Mike

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
mgage: ok.  I can do it essentially that way but a bit more safely.
Take care,

Mike

lib/WebworkWebservice.pm
mgage: Fixed.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
mgage: This file ProblemSetList2.pm has been reverted to the develop version

conf/defaults.config
mgage: I've reset this.  As an aside even turning $showEditors{ww3}=1 in localOverrides didn't display a link to ww3.  But I don't think I have the right things set up for ww3 anyway.  I can try to get to that after JMM. Although Bill Wheelers stuff comes first.

goehle:
-  [ ] There seems to be double encoding of some characters (like " or \).  Try an enter \ into an answer blank. 
- [ ]  Using "Show Auxiliary Resources" creates a knowls link in the warning, but the link opens up way at the bottom of the page below all of the other debugging info.  
- [ ] The PGinfo option is a little strange in that it requires you to manually add a macro and prints out the info in a different place than the other options.

mgage:
The problems with the Answer Hash are because the pretty_print is being called outside of the safe compartment -- we'll need to stringify everything and test it.  This will probably show up in PGinfo also when we move it to be printed like the others.  

This needs to be fixed but I'd rather it didn't hold up everything else.  The basic delivery of the information is there and we just need to adjust the interaction with mathobjects. I'll have to think about it a bit -- we could just stringify everything in PGcore when it is exported but that might affect the answer processing so we want to look carefully to find the time when it is safe to stiringigy PGcore and its contents.

conf/defaults.config
```diff
@@ -1043,9 +1054,24 @@ $pg{directories}{appletPath} = [    # paths to search for applets (requires full
...
 	"$webworkURLs{htdocs}/applets/Image_and_Cursor_All",
-	
 ];
 
+
+$pg{directories}{htmlPath} = [    # paths to search for applets (requires full url)
+    ".",
+	"$courseURLs{html}",  
+	"$webworkURLs{htdocs}",
+];
```
goehle: Lines 1065 and 1070 say `# paths to search for applets`.  I wouldn't make a fuss but its in defaults.config so it might actually get read by someone.

lib/WeBWorK/ContentGenerator/Problem.pm
mgage: I believe I've done this.  But please double check since I'm not sure which changes to output_comments you are referring to.

goehle: Hmm, could you double check things on your end.  I can see the file changes below that need to be reverted.  (All the things in red should stay and the things in green shouldn't be added.)

conf/defaults.config
mgage: OK.  this is fixed.  These are now default paths to search for html files, image files and pdf files

lib/WeBWorK/ContentGenerator/Problem.pm
mgage: OK. That's the change I made in Problem.pm.  Just double checking.

goehle: Sorry to keep harping on this, but if you made that change in Problem.pm then its not showing up below. Did you make it in a different branch?  Or did you not push it to your github repo.  Lines 1865-1883 need to be added back in and 1779-1812 removed.

goehle:
New list of things to do: 

-  [ ]  There is an issue with `pretty_print_tex`.  You can debug it with the following problem 
  ```
Context()->texStrings;
BEGIN_TEXT
$BR
Enter pi \{ans_rule(20)\}
$BR $BR
Test ${DOLLAR}PG->pretty_print (${DOLLAR}pi->cmp, "TeX", level=3);
$BR$BR
\{$PG->pretty_print($pi->cmp, "TeX", 3)\}
$BR $BR
Test ${DOLLAR}PG->pretty_print (${DOLLAR}pi->cmp, "TeX", level=4);
$BR $BR
\{$PG->pretty_print($pi->cmp, "TeX", 4)\}
$BR $BR
Test pretty_print(${DOLLAR}pi->cmp) -- this should give current display
\{pretty_print($pi->cmp)\}
END_TEXT
 ```
 - [ ]  I want to test the sage additions you made.  Could you post a pg file which uses the functions defined at https://github.com/openwebwork/pg/pull/255/files#diff-ec688a0529a3964cd5ab6e67145adba3R82 and at https://github.com/openwebwork/pg/pull/255/files#diff-ec688a0529a3964cd5ab6e67145adba3R102

mgage:
I've reverted the parts of Problem.pm that needed it.

clients/check2_problems_in_dir.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+#usage 
+#   check_problems_in_dir.s  directory_name
+
+echo "Checking problems in directory $1"
+echo '' > "$WEBWORK_ROOT/DATA/bad_problems.txt"
+echo "Results sent to" 
+echo "$WEBWORK_ROOT/DATA/bad_problems.txt"
+time find $1 -name "*.pg" -exec /Volumes/WW_test/opt/local/bin/perl $WEBWORK_ROOT/clients/sendXMLRPC.pl -C -B {} ';'  #tail -f /Volumes/WW_test/opt/webwork/webwork2/DATA/bad_problems.txt;
```
goehle: This is a hard coded path to perl and should probably just rely on the path or a config variable.

mgage: ok -- although this .sh file becomes redundant once the new sendXMLRPC is stable. It's functionality is embedded in the new sendXMLRPC.

PR #658, opened 2015-12-01
Bugfixes for 2.11

goehle:
There are really 2 things to check.  I'm trying to keep these small rather than pile up one huge pull. 

1.   Newer versions of XMLRPC::Lite don't like having a string for SSL_verify_mode.  So I imported the correct constant from the package.  
  -  To test check to see that you can still render problems using Problem Set Detail 2.  If you are using Ubuntu 15.10 you should have gotten an error before this patch.  Older distros should see no change. 

2.  Changed how the datetimepicker conflict resolution works.  I tried to get to fancy before.  Now there is a consistent rule.  Given two dates if date A should come before date B but doesn't then date B is set to date A (irregardless of if you changed A or B to get into this situation).  I also changed it so that it won't fill in "None Specified" fields.  
  -  To test, enable reduced scoring and check that when you change dates it has the behavior described above. 
  -  Also test by visiting the User Detail page or the Problem Set Detail page for a user and overriding an open date.  Before the patch it will fill in all of the dates, after the patch it should fill in only the one date.  
  -  Note:  It should really validate your override date against the global dates, but that will have to be a different release.  

3.  Refactored how showeditors is used in localOverrides.  It currently redefines the hash which means that the hash set up in defaults.config is never used.  This changes redoes the settings in defaults.config to reflect what the "default" localOverrides version looked like and then sets up lines in localOverrides to override individual entries in the hash. 
  -  To test, make sure that the expected editors are shown and that you can enable and disable things from localOverrides.  

4.  Added some js so that if you change a student override on ProblemSetDetail, ProblemSetDetail2, or UserDetail then the corresponding checkbox is automatically ticked.  
  -  To test, go to those three pages and change some fields, the override checkbox should become checked.  Make sure that you can still uncheck things and that the usual functionality works. 
  -  Also point out anywhere else this behavior should be tweaked.  

The rest of the changes are just documentation stuff.


mgage:
All the tests pass.  When using the UserList you can confuse yourself if you change some dates but not others.  If there are conflicts between specific and general dates errors are flagged when you commit not when you enter.  This is as expected and is good enough.  I've also seen spurious error warnings regarding the reduced scoring if that entry is left blank when changing others.

PR #659, opened 2015-12-02
More 2.11 Bugfixes

goehle:
This is another bugfix pull for Release 2.11.  I'm trying to limit them to 3 or 4 things per pull to keep it manageable.  This time around: 

1.  Fixed a bug where if a value was not defined for a problem in a set def file then the value of the previous problem would be used.  
  -  To test.  Import the following def file. Before the patch the showMeAnother value for problems 3 and 4 will be 3.  After the patch it should be -1/Never/whatever the default is on your server.  
  ```
assignmentType      = default
openDate          = 11/20/2014 at 11:59pm EST
dueDate           = 11/27/2016 at 11:59pm EST
answerDate        = 11/27/2016 at 11:59pm EST
paperHeaderFile   = defaultHeader
screenHeaderFile  = defaultHeader
description       = 
restrictProbProgression = 0
emailInstructor   = 0
enableReducedScoring = Y

problemListV2 
problem_start
source_file = Library/UMN/algebraKaufmannSchwitters/ks_5_2_79.pg
value = 1
max_attempts = -1
showMeAnother = -1
problem_id = 1
counts_parent_grade = 0
att_to_open_children = 0 
problem_end
problem_start
source_file = Library/ma117DB/set1/srw1_1_51a.pg
value = 1
max_attempts = -1
showMeAnother = 3
problem_id = 2
counts_parent_grade = 0
att_to_open_children = 0 
problem_end
problem_start
source_file = Library/ma117DB/set1/srw1_1_51b.pg
value = 1
max_attempts = -1
problem_id = 3
counts_parent_grade = 0
att_to_open_children = 0 
problem_end
problem_start
source_file = Library/ma117DB/set1/srw1_1_53b.pg
value = 1
max_attempts = -1
problem_id = 4
counts_parent_grade = 0
att_to_open_children = 0 
problem_end
  ```
2.  Added the new reduced scoring parameters to the import/export functionality of Homework Set Editor.  Previously this data was not saved on import/export. 
  -  To test import the above def file.  Before the patch it will have a reduced scoring date far in the past (1970).  After the patch the reduced scoring date should be the due date minus the default reduced scoring period (maybe 0).  
  -  Also add the lines 
`reducedScoringDate = 11/25/2016 at 11:59pm EST` and `enableReducedScoring = Y` to the def file and re-import it.  It should have the appropriate date and reduced scoring should be enabled. 
  -  Export a set and make sure the reduced scoring date and enable reduced scoring variables show up in the def file.  
  -  Note:  I'm not keeping the Old Homework Sets editor up to date any more so these changes will not appear there.  

3.  I rebuilt the def files for the default sets in modelCourse.  These new def files show off the newer def format and some of the more recent options.  If people are using these def files as a basis to write their own then these will be much better. 
  -  To test:  Create a course, import the default sets and make sure they look ok.  

4.  I fixed a bug where using "New Version" or "Append" in PG Problem Editor 3 would open the new editor in the modal window.  
  -  To test:  Edit a problem  using PG problem editor 3.  Before the patch if you use New Version or Append it will always open the editor in the modal window.  After the patch it will either replace your current window or open in a new window, depending on if you have the check box ticked.

### DIFF
--- a/courses.dist/modelCourse/templates/set0.def
+++ b/courses.dist/modelCourse/templates/set0.def
@@ -7,13 +7,13 @@ enableReducedScoring = N
 paperHeaderFile   = set0/paperHeaderFile0.pg
 screenHeaderFile  = set0/screenHeaderFile0.pg
 description       = 
-restrictProbProgression = 
-emailInstructor   = 
+restrictProbProgression = 0
+emailInstructor   = 0
 
 problemListV2 
 problem_start
 source_file = set0/prob1.pg
-value = 0
+value = 1
 max_attempts = -1
 showMeAnother = -1
 problem_id = 1
@@ -22,7 +22,7 @@ att_to_open_children = 1
 problem_end
 problem_start
 source_file = set0/prob1a.pg
-value = 0
+value = 1
 max_attempts = -1
 showMeAnother = -1
 problem_id = 2
@@ -31,7 +31,7 @@ att_to_open_children = 1
 problem_end
 problem_start
 source_file = set0/prob1b.pg
-value = 0
+value = 1
 max_attempts = -1
 showMeAnother = -1
 problem_id = 3
@@ -40,7 +40,7 @@ att_to_open_children = 1
 problem_end
 problem_start
 source_file = set0/prob2.pg
-value = 0
+value = 1
 max_attempts = -1
 showMeAnother = -1
 problem_id = 4
@@ -49,7 +49,7 @@ att_to_open_children = 1
 problem_end
 problem_start
 source_file = set0/prob3.pg
-value = 0
+value = 1
 max_attempts = -1
 showMeAnother = -1
 problem_id = 5
@@ -58,7 +58,7 @@ att_to_open_children = 1
 problem_end
 problem_start
 source_file = set0/prob4/prob4.pg
-value = 0
+value = 1
 max_attempts = -1
 showMeAnother = -1
 problem_id = 6
@@ -67,7 +67,7 @@ att_to_open_children = 1
 problem_end
 problem_start
 source_file = set0/prob5.pg
-value = 0
+value = 1
 max_attempts = -1
 showMeAnother = -1
 problem_id = 7
--- a/courses.dist/modelCourse/templates/set0.def
+++ b/courses.dist/modelCourse/templates/set0.def
@@ -12,65 +12,65 @@ emailInstructor   = 0
 
 problemListV2 
 problem_start
+problem_id = 1
 source_file = set0/prob1.pg
 value = 1
 max_attempts = -1
 showMeAnother = -1
-problem_id = 1
 counts_parent_grade = 0
 att_to_open_children = 1 
 problem_end
 problem_start
+problem_id = 2
 source_file = set0/prob1a.pg
 value = 1
 max_attempts = -1
 showMeAnother = -1
-problem_id = 2
 counts_parent_grade = 0
 att_to_open_children = 1 
 problem_end
 problem_start
+problem_id = 3
 source_file = set0/prob1b.pg
 value = 1
 max_attempts = -1
 showMeAnother = -1
-problem_id = 3
 counts_parent_grade = 0
 att_to_open_children = 1 
 problem_end
 problem_start
+problem_id = 4
 source_file = set0/prob2.pg
 value = 1
 max_attempts = -1
 showMeAnother = -1
-problem_id = 4
 counts_parent_grade = 0
 att_to_open_children = 1 
 problem_end
 problem_start
+problem_id = 5
 source_file = set0/prob3.pg
 value = 1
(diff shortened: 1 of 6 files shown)


mgage:
Pretty good but a few comments.  The first is that the changes in localOverrides.conf.dist in bugfix didn't seem to make it over to localOverrides.conf.dist in bugfix2.   

2. Did you mean for this to be omitted from localOverrides.conf.dist?

# The default prPeriod value (re-randomization period) to use for the newly created problem.
# It is suggested to use the value of -1, which means that the course-wide setting would be used
# Setting this to -1 defaults to the use of course-wide settings (suggested)
# Setting this to 0 disables periodic randomization regardless of the course-wide setting
# Setting this to a positive value will override the course-wide setting
$problemDefaults{prPeriod} = -1;


# Periodic re-randomization
# switch to enable periodic re-randomization
$pg{options}{enablePeriodicRandomization} = 0;
# course-wide default period for re-randomization, should be an integer
# the value of 0 disables re-randomization
$pg{options}{periodicRandomizationPeriod} = 5;

3. importing the set definition file (I was on branch bugfix ) I get this error:

Getting error 

readSetDef error, can't read the line: ||enableReducedScoring = Y|| at /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm line 1947, line 10.
on bugfix branch



but the reducedScoring date is set to the due date (which seems reasonable).

I named the set settest_setdeffile.def and this was imported as deffile  :-)
It's a bug -- but it serves me right for choosing a weird name.

With the bugfix2 branch everything works as expected except that the settest_setdeffile.def name is still mangled.

MOST SERIOUSLY.  I can't view problems at all -- and get this error:


[Thu Dec 03 20:57:02 2015] [error] [client ::1] [/webwork2/gage_demo/Calculus_II/9/] no type matches module WeBWorK::ContentGenerator::Instructor::SimplePGEditor with args courseID=>gage_demo setID=>Calculus_II problemID=>9 at /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm line 636\n * in Carp::croak called at line 911 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/URLPath.pm\n * in WeBWorK::URLPath::newFromModule called at line 636 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::__ANON__ called at line 807 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::links called at line 155 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/Template.pm\n * in WeBWorK::Template::template called at line 531 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::content called at line 470 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm\n * in WeBWorK::ContentGenerator::Problem::content called at line 203 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::go called at line 382 of /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK.pm, referer: http://localhost/webwork2/gage_demo/Calculus_II/?effectiveUser=gage&user=gage&key=qs947hBrPcSPcplFGEs8oSfEa0Kt491t

in the error log file.

the problem is that in defaults.config  

all of the problem editors are turned on including simplePGEditor which doesn't exist (at least on my machine)

So defaults.config needs to be fixed -- I recommend turning all of the pgproblemeditors off except for 3 (or if you prefer 2 -- I don't care that much between these two.).

goehle:
1.  Yes.  Periodic Rerandomization is a develop feature and isn't in this branch. 

2.  The branches weren't compatible.  You should have gotten errors about it not understanding "enableReducedScoring" unless you were on bugfixes2.  As long as it works correctly on this branch.   

3.  I have changed the regexp that pulls of the set name to 
```
$fileName =~ m|^(.*/)?set([.\w-]+)\.def$|
$setName = $2;
```
This should pull off the *first* set but also allow for folder names.  You should test it out since I could potential break some case I'm not thinking of. 

4.  I think this might have been a bad interaction between switching between bugfix2 and bugfix branches.  Could you test again?  I'm not seeing simplePGEditor and in particular defaults.conf only has one pg editor turned on.

mgage:
Thanks for updating the regex -- it wasn't a major issue but it's nice to have it fixed.  

The defaults.config works for me now. (pulling a fresh copy directly from bugfix2).

Looks like it's good to go.

PR #660, opened 2015-12-11
Release 2.11 Bugfixes 3

goehle:
More bugfixes for release 2.11

1.  Fixed an undefined variable bug in show answers. 
  -  To test:  Create a gateway, create a version, enter some answers, then click show past answers.  Before the fix you should get an error and after it should work. 
2.  Fixed a small bug with the rename feature. 
  -  To test:  Rename a course but only change the institution.  Before the fix you should get error messages and after you should be allowed to change only the institution.  
3.  Fixed a bug with restrict to IP and apache 2.4  For apache 2.4 the api for getting the request ip address changed.  To test.  
   -  Set up some remote addresses in the admin course, and set up a homework to use the restrict ip feature. 
   -  Log in as a student and try to access a homework with a restricted ip.  Before the patch if you are using apache 2.4 you should get an error about an undefined method.  After the patch it should work normally.  Ideally this test should also be repeated on a pre 2.4 server to check and make sure there aren't regressions.

### DIFF
--- a/conf/localOverrides.conf.dist
+++ b/conf/localOverrides.conf.dist
@@ -257,24 +257,6 @@ $problemDefaults{max_attempts} = -1;
 # setting this to -1 disables the showMeAnother button
 $problemDefaults{showMeAnother} = -1;   
 
-# The default prPeriod value (re-randomization period) to use for the newly created problem.
-# It is suggested to use the value of -1, which means that the course-wide setting would be used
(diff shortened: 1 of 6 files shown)


mgage:
I've verified 1. and 2.  and verified that 3. works as expected on apache2.2  Since Geoff tested apache2.4 I'll pull this request.

PR #661, opened 2015-12-14
Even More 2.11 Bugfixes

goehle:
- Fixed an issue where answers of 0 were not being preserved after timeout. 
 -  To test set your timeout value to be low, enter a 0 for any problem answer, wait, and hit check answers.  Before the patch after logging back in the submitted answer will be blank, after the patch it should have a zero.  
- Made a better error when there is no set name when creating a set via the library browser. 
  -  To test make a set via the library browser but don't input a set name.  
-  Fixed an issue with the success message when appending a problem via PGProblemEditor3
  -  To test append a problem with PGProblemEditor3.  
-  Fixed an issue with columns on the Problem Set page.  
  -  To test, make a set with an essay answer.  Check the formatting of Problem Set.  Submit an answer and check the formatting of Problem Set with the grader link.  Change the set type to Jitar and check the formatting again.  
-  Changed how essay answer comments were rendered on  the Problem page. It should be faster and more generic.  Test by assigning an essay answer, grading it with a comment, and checking that it renders on the problem page correctly.  
-  Added css to make problem set detail 2 work in math3.  
  -  To test, change to math3 and visit the problem set detail 2 page.  Everything should function, although it doesn't look quite as good.

### DIFF
--- a/htdocs/js/apps/MathView/mathview.js
+++ b/htdocs/js/apps/MathView/mathview.js
@@ -122,7 +122,7 @@ function MathViewer(field,button,container,userOptions) {
     if (this.options.decoratedTextBoxAsInput) {
 	this.inputTextBox = $(field);
     } else {
-	this.inputTextBox = $('<input>',{type : 'text', class : 'mv-input', size:'40'});
+	this.inputTextBox = $('<input>',{type : 'text', class : 'mv-input', size:'32'});
     }
 
     MathJax.Hub.Config({
--- a/htdocs/js/apps/PGProblemEditor3/pgproblemeditor3.js
+++ b/htdocs/js/apps/PGProblemEditor3/pgproblemeditor3.js
@@ -12,6 +12,8 @@ $(function(){
 	    .addClass('hidden-desktop hidden-tablet');
 	$('#pg_editor_frame_id').contents().find('#content')
 	    .removeClass('span10').addClass('span12');
+	$("#pg_editor_frame_id").contents().find('#toggle-sidebar')
+	    .addClass('hidden');
     });
 });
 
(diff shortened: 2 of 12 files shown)


mgage:
I did not see a change in behavior when no set name was defined in the Library Browser and a problem was added to the set.  (Pop up message urges you to define a target set).

I did not see a change in the columns for essay questions.  They looked ok to me before and after the upgrade.  What was the problem?

goehle:
Huh, I didn't get that popup,  I wonder why.  Well I improved the message
that happens if you get past the popup somehow.

If you look closely (or use the html inspector) before the change to the
"grading link" column for essay questions the grading column header was not
lined up with the actual column.  This made it look slightly offset and
causes issues with accessibility.

mgage:
If I go to the library browser, choose subject, chapter and section and "view problems" and then I choose "add" for the first problem I get the following pop-up message:  "You need to pick a target set above so we know what set to which we should add this problem." (The target set has not been chosen and no new set was created either.).  I don't see additional messages about not having chosen a set.  

However if I push "create a new set for this course"  without filling in the blank for the set name I get a message (in red) "You did not specify a new set name."  This is an improvement over the old message "Problem creating set 
argument 1 contains empty 'set_id' field at /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm line 1439 at /Volumes/WW_test/opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm line 1439"  

so I assume this is the message you are talking about.

mgage:
I see the mismatch between the header and the text in the "grade problem" column now.  I just missed it before.

This pull request is good to go.

goehle:
Yup, thats it.

PR #664, opened 2015-12-17
Hardcopyfix

jwj61:
Put  a problem like

Library/WHFreeman/Holt_linear_algebra/Chaps_1-4/4.3.5.pg

in a problem set and get hardcopy showing correct answers.  The student would get correct answers which are in latex, but they put in verbatim, so they see the raw latex which is very hard to read, even if you are well versed in latex.  This change makes it so the hardcopy module detects this and usually puts correct answers in verbatim, unless they are already in math mode.

### DIFF
--- a/htdocs/js/apps/TagWidget/tagwidget.js
+++ b/htdocs/js/apps/TagWidget/tagwidget.js
@@ -18,6 +18,7 @@ var basicRequestObject = {
 var taxo=[];  // Global variable to hold it
 var loadtaxo = $.ajax({
   dataType: "json",
+  async: false,
   url: "/webwork2_files/DATA/tagging-taxonomy.json", 
   success: function(data) {
     taxo = data;
--- a/htdocs/js/apps/TagWidget/tagwidget.js
+++ b/htdocs/js/apps/TagWidget/tagwidget.js
@@ -18,7 +18,6 @@ var basicRequestObject = {
 var taxo=[];  // Global variable to hold it
 var loadtaxo = $.ajax({
   dataType: "json",
-  async: false,
   url: "/webwork2_files/DATA/tagging-taxonomy.json", 
   success: function(data) {
     taxo = data;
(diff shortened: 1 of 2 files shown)


dpvc:
I would be against doing this change.  The `correct_ans` field should never contain LaTeX.  It is supposed to be what the student should type in the answer blank to get the correct answer, and should always be a plain text string.  (Currently, it sometimes contains HTML formatting, but even that is a misuse of the field).  The `correct_ans_latex_string` is the one that should contain a LaTeX version of the correct answer.  If you want to show formatted answers int eh hardcopy, that is what should be used (when available), not `correct_ans`. 

In my opinion, doing this sort of content-based decision makes it harder to write new answer checkers.  For example, if someone wanted to use `\` for set subtraction rather than `-` (since it is a common notation for that), then an answer of `R\(1,3]`, which should be put in verbatim mode, would end up inserted as-is, and would likely cause a TeX error.  We would need to make the MathObject parser aware of the heuristics that the hardcopy module is using in order to work around them.  There is no reason it should have to be aware of such things.  This sort of hack just leads to maintenance nightmares.

One of the sources of the problem is that the correct use of `correct_ans`, `student_ans`, and the other related fields is not well specified anywhere, and has evolved over time.  The `correct_ans has been overloaded to include HTML in some cases, and now LaTeX, and it is not clear which ones have had what type of escaping applied to them.  There really need to be separate fields for each format that is needed (e.g., `correct_ans_text`, `correct_ans_latex`, `correct_ans_html`, and the same for `student_ans`).  These should contain valid strings for each of those formats, and the correct one should be used in the correct place (i.e., `correct_ans_html` in the results table, not `correct_ans_text`).

The problem that you cite uses `basis_cmp`, and it seems to me that if it is setting `correct_ans` to a LaTeX string, then that answer checker needs to be fixed rather than adding some fragile heuristic into the hardcopy generator.  I looked at the code briefly, and it appears to use `display_matrix` to generate the `correct_ans`, and that `display_matrix` makes its output dependent on the `$displayMode`.  That is not the correct behavior for `correct_ans`, so my suggestion is to fix `basis_com` to produce proper content for `correct_ans`.

jwj61:
The potential error you propose with R\(1,3] would not be a problem since the string has to start with (possible whitespace followed immediately by) \(.  It would not detect an internal \(.

That said, I think the current situation should be fixed one way or another.  I don't care which.

dpvc:
OK, I agree that it has a lower probability of failure that I originally thought, but I still believe it is the answer checker that is at fault, not the hardcopy code.

jwj61:
Closing this.  There is now a pull request on the pg side.

PR #665, opened 2015-12-18
OPL Statistics

goehle:
This is a new feature which adds some basic statistical information to the Library Browser.  You can read about the details here: http://webwork.maa.org/wiki/OPL_Problem_Statistics but the short version is that problems in the library browser will look like
opl_statistics

To get this running: 
1) Switch over to this branch. 
2) Download the latest version of the OPL
3) Run `update-OPL-statistics`. 

Things to test: 
- Test that you get both global and local statistics in the Library Browser. 
- Test that you can turn the feature on and off by using the two stats variables in `localOverrides.conf.dist`
- Check that formatting and help links are sane. 
- Check that if you have an open set, submissions do not change the local stats (after running `update-OPL-statistics` again) but if you change the due date to be closed then the local stats are updated.  

Note:  This also contains the `dump_past_answers` script which is being used in the other big data project.  It can be ignored for now.


mgage:
Got this error while running the OPL script:  

Updating Library Statistics.
sh: mysql: command not found

goehle:
Hmm, what are you using for your databse?  I check with MariaDB and it
still provides a "mysql" command.  Does the database program you are using
support runing .sql scripts?

mgage:
OK.  Everything checks out now.

PR #666, opened 2015-12-20
Yet Still Even More Release 2.11 Bugfixes

goehle:
-  No longer show progression message on subproblems which don't count towards the parent grade. 
  -  To test: Create a just in time set with restricted progression and have a problem with two child problems, one which counts for the parent grade and one which doesn't.  When viewing the problems you should get a "you cannot proceed" message on the problem page for only the one which counts for the parent grade. 

-  Fixed a bug where restricted progression didnt work properly. 
  -  To test:  Have problems 1, 1.1 and 1.2 with all three problems having a max attempts limit and *both* child problems must count toward the parent.  Run out the attempts on the parent problem and a subproblem, but get the other subproblem right while still having attempts remaining.  Before the patch you will not be able to proceed to the next problem and after you will. 

-  Changed how scores are displayed on Problem Set page for just in time sets.  Now status of all problems is shown, as well as adjusted status for top level problems and  a flag if child problems count to the parent grad.e 
  -  To test:  create a just in time set and look at the problem set page.   Check that columns all work. 

-  Cleaned up single set csvs in "Scoring Tools".  
  -  To test:  after the patch you should see a due date and due time in single set scoring csvs.  Also, for jitar sets you should see the problem number, not some big integer.  

-  Added some code to improve formatting of the Problem Set page when the window is smallish but not tablet sized (about 1000px wide).  For just in time sets the Problems table was getting obscured by the Set Info page.  I reduced the padding on that table.  This doesn't get rid of the clipping but it minimizes it.  
  -  To test:  Create a just in time set and view the ProblemSet page and shrink your window.  You should see the table get compressed, then there will be about a small interval where the set info box clips over the table, then the window should go into "phone mode" and the Set Info box should drop below the table. 

-  Changed code so that User Settings is always shown, since it pretty much always has something the user can do.  

- Small CSS Tweaks:  
  -  Tweaked the masthead css so that it no longer overflows the container.  This should get rid of the bottom scrollbar that appears no matter how wide the window is.  Check that the login button renders nicely and there is no scrollbar. 
  -  On IE the vertical scrollbar actually covers up page content.  This is annoying in the library browser where the scrollbar can cover up the "X"'s and make them hard to use.  Added some css to inset the library browser icons a bit to avoid this.  Check to see that it renders nicely.  
 -  The "skip to main content" link now skips to the page title.  Before it skipped to before the breadcrumbs.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -369,9 +369,8 @@ table.classlist-table .tbody {
     margin-top: 0px;
 }
 
-.problem_set_table {
-    display:block;
-    overflow:auto;
+.problem_set_table td, .problem_set_table th {
+    padding: 8px 3px 8px 3px;
 }
 
 .problem_set_table td a {
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -83,8 +83,7 @@ legend {
 #masthead {
 	background-color: #038;
 	margin: 0;
-	padding: 0;
-	padding: 5px;
+	padding: 5px 0px 5px 0px;
 }
 
 .webwork_logo {
@@ -118,7 +117,8 @@ legend {
 }
 
 #loginstatus {
-    	padding-top:.5ex;
+        padding-top:.5ex;
+        padding-right: .5em;
 	color: white;
 	text-align: right;
 	font-size: 0.85em;
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -369,6 +369,11 @@ table.classlist-table .tbody {
     margin-top: 0px;
 }
 
+.problem_set_table {
+    display:block;
+    overflow:auto;
+}
+
 .problem_set_table td a {
     font-weight:bold;
     font-size:13px;
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -62,7 +62,7 @@ $(function(){
 
     // Fix bug with skip to main content link in chrome
     $('#stmc-link').click(function() {
-	$('#content').attr('tabIndex', -1).focus();
+	$('#page-title').attr('tabIndex', -1).focus();
     });
 
     // Turn submit inputs into buttons
@@ -88,7 +88,7 @@ $(function(){
     });
     $('a.help-popup').popover({trigger : 'hover'}).click(function (event) {
 	event.preventDefault();
-    }).html('<i class="icon-question-sign"/>');
+    }).html('<i class="icon-question-sign"/><span class="sr-only">Help Icon</span>');
 
     // Sets login form input to bigger size
     $('#login_form input').addClass('input-large');    
--- a/htdocs/themes/math4/system.template
+++ b/htdocs/themes/math4/system.template
@@ -76,7 +76,7 @@ var tabberOptions = {manualStartup:true};
 </head>
 
 <body>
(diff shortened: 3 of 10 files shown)


mgage:
Use the set definition file to create a homework set with problems 1, 1.1, 1.2 and 2.  1.1 counts toward parent, 1.2 does not.  Problems are "restricted" in the set. (you'll need to replace .txt with .def

[setjitar_2.11_tests.txt](https://github.com/openwebwork/webwork2/files/69640/setjitar_2.11_tests.txt)

Initially 1 and 1.1 are available but 2 is  not. 
Do 1.1 -- progress bar moves over and problem 1 (in total) is effectively 100% and problem 2 is available. Neither problem 1 nor 1.2 has been correctly answered.
Do 1.2 -- progress bar moves again even though it doesn't count. 
Do 1 -- progress bar doesn't move -- problem 1 was already correct.
Do 2 -- progress bar moves to perfect

Is this the behavior you expected?

mgage:
It looks like it is.  I also tried this with 1.1 not counting toward parent but 1.2 counting toward parent.
I didn't see a progress message but did see changes in the progress bar when I completed a problem that didn't count toward the parent.

goehle:
Yes, although to see the bug I was fixing you would need to change max
attempts for 1.1 to something other than -1.


- Initially 1 and 1.1 are available, but 1.2 and 2 are not.  If child
problems are open then the first child problem is available.  You would
need to do something with "attempts to open children" otherwise.

-  After you answer 1.1 correctly, the score for 1 is 100% so you can move
on to problem 2.  Unless 1.2 counts for the grade of 1 it won't impact
restricted progression.

-  The progress bar stuff sounds correct.  I'll admit it can be a little
unexpected but I don't think there is anything obviously wrong.


Geoff.

goehle:
I toyed with not showing status of problems that don't count for the parent
grade and leaving those off the progress bar, but I got the sneaking
feeling that it would encourage students to skip them.

mgage:
I agree.  The current behavior seems ok to me.  I haven't used JITAR much and the logic is complicated because eventually you want to do complicated things.  I expect to be confused at the beginning for a while.

goehle:
Eventually I should write up some use cases that are simpler.  For example,
restrict problem progression can be used on a set with no child problems at
all to force students to go through one problem at a time, which is a
feature I've heard requested before.

mgage:
OK -- point two works as expected -- but both children (1.1 and 1.2) must count toward parent.

mgage:
Display columns for scores of JITAR work fine on problem set page --- really quite clear what the status is.  Adjusted status message is good -- I hadn't realized that you could either get the children right or the parent problem right -- I thought you had to get both right.

mgage:
I have an issue with showing the User settings all of the time. Set the permissions so that students can change email addresses and passwords (but guests can't) -- this is the default permission setting. With the new system guests can change email (and perhaps their password if they guess their current one).

If you are going to expose the User settings page then the individual entries on that page should be controlled by permission settings -- and preferably have some visual sign as to whether or not they can be changed.

goehle:
That should be how it works already.  If you don't have permission to change your password/email then it disables those on the User Settings page.  I tested that, but I'll test it again.

mgage:
I was able to change my email as guest.

The CSS tweaks work.  Including the ones for iphone display

mgage:
If it's not too hard --- grey out and disable the password change section and the email change section when the user doesn't have permission.  Once that is done this PR is good to go.

PR #669, opened 2015-12-24
Problem Add Bug

goehle:
Some values in `addProblemToSet` should be defined or and not or, otherwise perfectly reasonable values of zero are replaced with a potentially non-zero default. 

To test:  Import a set with a showMeAnother/attToOpenChildren as zero (with nonzero default values) and check that it works.


mgage:
This works for me.

PR #671, opened 2015-12-27
Goehle smabug

mgage:
reconciled version of Goehle's smabug.


mgage:
This merges Goehle's smabug into the develop branch

PR #672, opened 2016-01-01
Develop wwclientsup8

mgage:
Replaces the client "renderProblem.pl" and various derivatives with a single script "sendXMLRPC.pl".

Using switches this script has all the capabilities of the previous collection of clients.  

It should also be compared to the script "standalonePGproblemRenderer.pl" in the github.com/mgage/standaloneProblemRenderer repo.  That script behaves in exactly the same way but instead of sending the problem via xmlrpc to be rendered remotely it connects directly to a local pg repo and renderers it that way.  standaloneProblemRenderer is currently more experimental. It is faster, but it also reveals a memory leak in PG, so having it process 100's of files in a directory will probably cause the  process to exceed RAM memory bounds

Command line commands connect the clients to browsers and to editors.  Some adjustments may need
to be made at the top of the file to switch the commands from Mac specific commands for BBEdit and Chrome to unix commands for other browsers and editors.

After setting up a credentials file the beginning tests 
(the tests for standaloneProblemRenderer are essentially the same.)

sendXMLRPC.pl -bB input.pg   --- displays problem in the browser, then answers the questions and displays again

sendXMLRPC.pl -C input.pg    -- enters an entry in the log file webwork2/DATA/bad_problems.txt -- indicates whether the problem would accept its own correct answers.

sendXMLRPC.pl -C /opt/webwork/libraries/webwork-open-problem-library/OpenProblemLibrary/Utah
will check all of the files in the Utah section of the OPL.  

sendXMLRPC.pl       --help   will print a copy of the POD documentation describing the switches. 
(Check that these two bits of documentation remain in sync and that they give adequate instructions.)

sendXMLRPC.pl  -v input.pg     ----- gives more verbose output useful for debugging

__________________________________________________________________
NAME
    webwork2/clients/sendXMLRPC.pl

DESCRIPTION
    This module provides functions for rendering html from files outside the
    normal context of providing a webwork homework set user an existing
    problem set.

    It can be used to create a live version of a single problem, one that is
    not part of any set, and can facilitate editing these problems outside
    of the context of WeBWorK2.

    This script will take a list of files or directories and send it to a
    WeBWorK daemon webservice to have it rendered. For directories each .pg
    file under that directory is rendered.

    The results can be displayed in a browser (use -b or -B switches) as was
    done with renderProblem.pl, on the command line (Use -h or -H switches)
    as was done with renderProblem_rawoutput.pl or summary information about
    whether the problem was correctly rendered can be sent to a log file
    (use -c or C switches).

    The capital letter switches, -B, -H, and -C render the question twice.
    The first time returns an answer hash which contains the correct
    answers. The question is then resubmitted to the renderer with the
    correct answers filled in and displayed.

    IMPORTANT: Remember to configure the local output file and display
    command near the top of this script. !!!!!!!!

    IMPORTANT: Create a valid credentials file.

SYNOPSIS
            sendXMLRPC -vcCbB input.pg

DETAILS
  credentials file
        These locations are searched, in order,  for the credentials file.
        ("$ENV{HOME}/.ww_credentials", "$ENV{HOME}/ww_session_credentials", 'ww_credentials')
    
        Place a credential file containing the following information at one of the locations above 
        or create a file with this information and specify it with the --credentials option.
    
            %credentials = (
                            userID                 => "my login name for the webwork course",
                            course_password        => "my password ",
                            courseID               => "the name of the webwork course",
                    site_url                   => "url of rendering site
                    site_password          => "site password" # preliminary access to site
                  form_action_url          =>  'http://localhost:80/webwork2/html2xml'; #action url for form
                  WWdisplayMode            =>  'MathJax', # optional 
                                              # --  'MathJax' and 'images' are the possible values           
                  HTMLdisplayCommand       =>  "open -a 'Google Chrome' "  # optional
                                                          # for Mac: have Chrome read html output file
                                                          # modify this for other platforms or browsers  
            );

  Options
    -a
                Displays the answer hashes returned by the question on the command line.

    -A
                Same as -a but renders the question with the correct answers submitted.

    -b
                Display the rendered question in a browser (specified by the DISPLAY_HTML_COMMAND variable).

    -B
                Same as -b but renders the question with the correct answers submitted. 
                The evaluation of the answer submitted is displayed as well as the correct
                answer.

    -h
                Prints to the command line the entire object returned by 
            the webwork_client xmlrpc request.
            This includes the answer information displayed by -a and -A and much more.

    -H
                Same as -h but renders the question with the correct answers submitted

    -c
                "check" -- Record success or failure of rendering the question to a log file.

    -C
                Same as -c but the question is rendered with the correct answers submitted. 
            This succeeds only if the correct answers, as determined from the answer hash, all succeed.

    f=s
                Specify the format used by the browser in displaying the question. 
                 Choices for s are
                 standard
                 sticky
                 debug 
                 simple

    -v
                Verbose output. Used mostly for debugging. 
            In particular it displays explicitly the correct answers 
            which are (will be)  submitted to the question.

    -e Open the source file in an editor.

                The single letter options can be "bundled" e.g.  -vcCbB

    --list pg_list Read and process a list of .pg files contained in the
    file "pg_list". "pg_list" consists of a sequence of lines each of which
    contains the full path to a pg file that should be processed. (For
    example this might be the output from an earlier run of sendXMLRPC using
    the -c flag. )
    --pg
                Triggers the printing of the all of the variables available to the PG question. 
            The table appears within the question content. Use in conjunction with -b or -B.

    --anshash
                Prints the answer hash for each answer in the PG_debug output which appears below
            the question content. Use in conjunction with -b or -B. 
            Similar to -a or -A but the output appears in the browser and 
            not on the command line.

    --ansgrp
                Prints the PGanswergroup for each answer evaluator. The information appears in 
            the PG_debug output which follows the question content.  Use in conjunction with -b or -B.
            This contains more information than printing the answer hash. (perhaps too much).

    --credentials=s
                Specifies a file s where the  credential information can be found.

    --help
               Prints help information.


mgage:
This set of commits should look a lot shorter once wwclientsup7 is pulled.

This pull should finish the job of making it possible to check all problems in a given directory.

mgage:
At that point you can do more massive checking of the webwork2 part of this change.

The PG part of this change seems more straightforward but it changes some basic structure of PG and it will take a while to make sure that subtle gremlins haven't been created.  The automatic tester helps with this.

mgage:
Check the merge at 6f9fc08.    Make sure that the choices I made with respect to essay questions in Problem.pm are correct. (I didn't work on those files and it wasn't always clear which was the desired code in the conflict.)

goehle:
I dont see any changes in Problem.pm that impact essay questions.  What
lines are you talking about?  Did you test the changes?

goehle:
What is going on with that commit.  It has an awful lot of jitar changes
that should have been in place a while ago.

mgage:
What I was worried about came from the merge.  I expect I had built on an earlier version of your
handling of essay questions and in the merge with develop I couldn’t tell for sure which was the 
code you intended. I did do basic checks with essays and didn’t run into problems.
But I’m not super familiar with them yet.

goehle:
I still dont know what code your talking about.  But essay answers havent
changed in a long time so I suspect its OK.

mgage:
It’s referencing devel from a long time ago at the moment.  I was expecting
many of those commit to disappear once wwclients5 is pulled.

goehle:
I did pull wwclient 5

goehle:
Here is a list of things from wwclients7 that were pushed forward: 
-  Using "Show Auxiliary Resources" creates a knowls link in the warning, but the link opens up way at the bottom of the page below all of the other debugging info.
-  The PGinfo option is a little strange in that it requires you to manually add a macro and prints out the info in a different place than the other options.
-  There are some server specific configuration settings (e.g. the command to open a webpage) that should be moved out of the files in clients.conf to either a config file or ww_credentials.  Basically we want to limit the amount of stuff that *has* to be configured for a specific server but is also managed by git.  
-  The utilities (like `renderProblem_rawoutput` ) that open up `less` should just print to stdout.  Users can then pipe to less if they want.

clients/sendxmlrpc_bbedit.pl
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/perl -w
```
goehle: This whole tool is too mac specific.  It has hardcoded paths and is centered around BB edit.  I would just keep this on your computer as a personal tool, or find some way to make it configurable to work on linux with emacs/vim/nano/whatever.

lib/FormatRenderedProblem.pm
```diff
@@ -0,0 +1,260 @@
...
+}
+
+sub new {
+    my $invocant = shift;
+    my $class = ref $invocant || $invocant;
+	$self = {
+		return_object => {},
+		encoded_source => {},
+		sourceFilePath => '',
+		url            => 'https://hosted2.webwork.rochester.edu',
```
goehle: Should this hardcoded path be here?

htdocs/themes/math4/system2.template
```diff
@@ -0,0 +1,291 @@
+<!DOCTYPE html>
```
goehle: What is this file for, out of curiosity.  I can't find anything in this pull request which uses it.  Also, is there a better name than system2.template.  That makes it sound like its intended to replace system.template.  (If it is intended to replace system.template then what are the differences and why do we need both?)

goehle:
Is it possible there were merge issues when you fixed conflicts with develop?  There are only 4 changed files, two basically trivial ones and two new additions.  The newly added files are FormatRenderedProblem.pm and system2.template, however I can't find any references to either of these two files being used in develop or this branch.

clients/sendxmlrpc_bbedit.pl
goehle: P.S.  A good place for this (and other scripts you find useful) would be http://webwork.maa.org/wiki/Contributed_Admin_Scripts_(Large_Installations)#.VqVM3N9vGEI  

I think it would be great if this page could be cleaned up and brought into the light a little bit more.  We need a place for tools which sys admins may find useful but which are not polished enough, or are too specific, for git.

mgage: What I was envisioning was having a number of tools like this -- one for each OS.  The tool is so small it's hardly worth configuring.   Which editor do you use?  Could you write a matching tool for that?  It's convenient to write in your favorite editor and then just hit a button to see what you have rendered so far.

It could go in admin scripts (it has the same flavor in that it almost demands being customized to the users environment) but it's really more of an author tool -- as are the various flavors of clients.

mgage: I'd really like to see some versions of this file for emacs, or vim or nano or sublime, etc.  Then authors will have some models to start with to build connecting links that suit their own work flow.

lib/FormatRenderedProblem.pm
mgage: Probably not.  Also this file will eventually be a super category for WebworkClient.   It will inherit it's formatting abilities from this.  (the standalone project will also inherit formatting from this).

clients/sendxmlrpc_bbedit.pl
goehle: I think having multiple versions is the wrong way to go.  IMO it should be a single tool configurable by things like environment variables, config files, etc...  For example most linux distributions have an `EDITOR` environment variable that people will set (https://en.wikibooks.org/wiki/Guide_to_Unix/Environment_Variables).  If you used that you could probably get rid of the BB edit specific stuff.  

More generally, I'm not a fan of having files that you need to configure to use be managed by git.  It really interferes with the git workflow since those files are tracked and your changes will have a tendency to revert, get overwritten, or cause conflicts.

mgage:
The merge appeared to be very easy.  There was only one minor conflict in instructorXMLRPC.  There were a lot of automatic merges and those are not guaranteed to be error free. The two files you mention are still being developed.  The system2.template can already be used. (see a small change in ContentGenerator that allows you to choose templates ).  The FormatRenderedProblem abstracts part of WebworkClient which is doing two duties 1. communicating between client and server over xmlrpc (or eventually json) and 2. formatting the returned result for presentation in the browser.  FormatRenderedProblem splits off the latter function and -- once it is tested -- WebworkClient will inherit from it

clients/sendxmlrpc_bbedit.pl
mgage: It wouldn't work for BBedit -- this file is an extension of BBedit and needs to be moved to a specific directory in order to work.  It's like writing a small extension for emacs.

mgage: By the way — its ok if the bbedit connection file is stored somewhere else as long as it is highly visible to people who want to use it as a model to 
write their own connecting tool.

goehle:
I can test system2 just by changing content generator, then.  What is different about it, so that I know what I am looking for.  

Do you have code which I can use to test FormatRenderedProblem?  Or is there a change I can make in WebworkClient to test it?

mgage:
I've removed FormatRenderedProblem for now since it is not yet used.  (If you want to see what it does look at my standalonerenderer project on github.com/mgage. )  This is a version of the sendXMLRP client which connects directly to pg instead of via the webservice.  I'm working toward refactoring these two clients as much as possible.

mgage:
I've removed system2.templates --- it was an earlier version of simple.templates.  I've also changed the templateName() method in Problem.pm, ProblemSet.pm and ProblemSets.pm so that those pages can access other templates. (I also removed a restriction that the template had to be named system, simple or gateway as being overly cautious).  There are files in webwork2/t  for testing these templates:
simpleTemplateTest.html.dist and alternate_templates.html.dist.  I've added the .dist extension since they will probably require some modification to test a local webwork site.  At the moment they are pointed at hosted2 which is running wwclients8.

goehle:
Should the the 2 test .pg problems be in pg/t and not webwork2/t ?

goehle:
I'll make a list of things below to check up on: 
-  [ ]  Move test .pg files to the pg repo? 
-  [ ]  In TestWW.html if I use .dist version I get
```
 Error in alias: PGalias.pm: unrecognizable displayMode =  at [PG]/lib/PGalias.pm line 276
```
as an error when trying to render the first problem.  Adding `displayMode=MathJax` makes the error go away.  I think this means that the "default" display mode is not getting set properly.  It could be a "defined or" being used where a regular or should be used.  
- [ ] In `alternate_templates.html` you have some templates being set as `templateName=plain` and I think it should be `templateName=simple`.  In addition the last to iframes point to the same place.  Do you mean for there to be a `/Demo` in the very last url?

mgage:
alternate_templates.html has been cleaned up and points to hosted2.  Check that two templates work acceptably when viewing a problem (Problem.pm), a homework set homepage (ProblemSet.pm) or a course homepage (ProblemSets.pm).   The links on the homework set homepage and course homepage revert back to the "system" template for now.  The "simple" template is not sticky in those cases.  It should be sticky for a single problem. 

simpleTemplateTest.html.dist also points to hosted2 -- compares appearance of the WebworkClient "simple" template (used for xmlrpc) with the simple template used through ContentGenerator (the webwork2 route)

If displayMode, courseID, userID, or problemSeed is missing then the problem is not rendered and the server returns an error message for the request.  Check that this works properly on hosted2 (when it is running the most recent develop). Check that it is not possible to use this error message for XSS.

mgage:
Fixed display of Show.

Test by viewing a problem containing a hint as a student.  If you have patience disable showing knowls for hints and for solutions and test again.  

Also view the same problem after the due date when both show solutions and show correct answer check boxes are available to make sure that the presentation is acceptable.

goehle:
Great, I'll try to check this stuff out later today.  Two things. 

1.  Could you merge my pull to this pull?  I fixed a bunch of `sticky` format stuff that was badly broken. 
2.  After having experimented with adding lti stuff to xmlrpc I was wondering if some of the options could be defaulted. 
  -  `displayMode` could pretty easily default to MathJax if  there isn't one specified.  
  -  `courseID` could default to a specific course specified in the servers .conf file if it exists or give an error message if it doesn't exist.  
  -  If `userID` and `course_password` are required parameters then you can't use LTI based authentication.   This may or may not matter to you in terms of fixing for this pull.  
  -  If `problemSeed` is missing then it could default to a static seed.  You could also have a flag to generate a random seed.  In particular with LTI stuff there isn't really a way to generate a random number when creating the LTI link so you either give all students the same seed or have a way to ask WeBWorK to pick a random one. (More generally with the LTI stuff it might be worth it put seeds and lti_source_did's in the key database so you can randomly generate a seed and then re-use it if the same user accesses the problem.)  
  The reason I'm asking for this stuff is that passing parameters through LTI can be a bit of a pain so reducing the number required has some good payoff.  Outside of the simple defaults this could also be put off to another pull, though.

mgage:
which branch do you want pulled?  ltigrade?  That hasn't yet been pull requested to openwebwork, but I can pull directly from your github.

There is a trade-off between error messages when important information hasn't been provided and supplying defaults (which in some cases can become confusing if the guessed defaults are not appropriate).  

Needing the defaults for LTI requests seems like a good reason to provide them. I'll make those changes in the next pull request wwclientsup9 so there is a clear record of which changes were made to accommodate LTI.

goehle:
I made a pull request directly to your github account.
https://github.com/mgage/webwork2/pull/16

mgage:
Move more configuration to the ww_credentials.dist file. Allow log file path to be specified using --log.

Test by changing the various configurations in ww_creditentials (or your private copy of same).  Make sure that the configurations take effect and that if they are omitted either an appropriate warning or an appropriate default is chosen.

Check that the --log=path_to_log_file switch works correctly. (The full path is required.)

Check that the directions in --help and the directions given in the POD documentation agree.

mgage:
Merged myclients8 from goehle.  

Tests: ??

mgage:
Remaining items:

- [ ] Move test .pg files to the pg repo? In any case the test problems in commandline_test_problems as well as the test problems in clients should all be consolidated, I think.
- [ ] Fix the resource object knowls (probably by adding &lt;p&gt;'s) so that when you open a resource knowls it doesn't open at the very bottom of the warn area.
- [ ]  Maybe change how PGInfo works so you don't have to add the macro file to your problems to use the PGInfo checkbox.

mgage:
PGinfo now prints without using the macro file PGinfo.pl.
I don't know if I understand the comment about fixing the resource object knowls. I've added a little space at the bottom of the hash when the knowl expands.

Both of these fixes involve changes only to pg files -- they appear in the pull request develop_showinfo

goehle:
I tested the following and everything checked out, so I will merge the result.  
-  Tested new sendXMLRPC options. 
-  Tested new ww_credentials configurations. 
-  I tested that the PGinfo and Resource options work and that the "Show:" test is displayed properly.  
-  I tested that the templateName parameter works
-  I tested the various .html files in `t` and the test pg problems as well.

PR #673, opened 2016-01-03
Develop Tweaks

goehle:
This addresses several tweaks/bugs for develop. 

1) I cleaned up the progress bars for OPL-update (they looked weird now that we have 100+ textbooks) and added a progress bar to update-OPL-statistics.  
  -  To test:  Run OPL-update (with the OPL statistics feature enabled).  You should see better progress counters for OPL-update and a new progress counter for statistics. 

2)  I made an attempt at getting the shortcut icon working.  As part of it I had to replace the icon by a very similar (but square) one. 
  -  To test:  Check to see if the favicon shows up next to the title of a WeBWorK page in your tab. 

3)  I made it so that die error messages are scrubbed and not encoded.  These messages have html in them so the encoding made them look bad.  I still want to have some protection to make the automated xss checkers happy, though.  
  -  To test:  Add a die statement to the initialize portion of a page with text like `die("<b>Thing</b>")`.  The html should be correctly rendered.  If you use `die("<script>alert('hi');</script>")` then the script should be scrubbed. 

4)  I removed some of the inline formatting from SetMaker.pm and added classes instead. Then I added the appropriate formatting to math3 and math4.  This fixes an issue in math4 with tiny little scroll bars appearing in the problem headers in the Library Browser and an issue in math3 where the icons in the problem headers looked bad. 
  -  To test:  Load up some problems in the library browser in math3 and math4 and check to see that they render nicely.  

5)  I fixed a bug where the raw problem number was used in SMA for just-in-time sets.  
  -  To test:  Create a just in time set and enable show me another.  Visit the show me another.  Before the patch the problem number will be some huge integer and after it should be "correct".


mgage:
What enables the OPL statistics update feature when running OPL-update?  

There is/was a separate script for updating statistics?

goehle:
If either of the stats types are enabled in localOverrides.conf then OPL-update will automatically run update-OPL-stats.  Or you can just run it directly.

mgage:
The favicon isn't showing up in chrome.  Firefox (developer version)  and Safari no longer appear to use favicons

mgage:
The progress report is nicely spaced out

goehle:
Try a hard refresh.  The caching and favicon can be annoying.  I have had
trouble getting it to work consistently though.

mgage:
The icons look fine and the spacing looks fine in the library although I didn't notice that much of a problem before.  In general math3 looks good.  I can see why some people (e.g. JJ) prefer it. It's bolder. 
The yellow pencil icon sticks out in that one.  the blue pencil in math4 blends in a bit more. Both look fine to me.

 The favicon is showing up now.  It looks good to

mgage:
The OPL-updated also reported importing statistics automatically.

mgage:
The die statement works as advertised.

mgage:
The problem number looks good now.  And the weights will be set properly when adding a library problem to a set once wwclients7 is pulled.  

I'm ready to pull this request.

PR #675, opened 2016-01-15
Grades and Scoring

goehle:
This adds a few tweaks to grades and scoring. 
-  Fixed a bug with just-in-time sets making too many grading columns. 
  -  Test by having a user only assigned a just in time set with subproblems.  If you visit the grades page it should have the correct number of problems. 
-  Replaced C by 100 as per Mikes suggestion. 
  -  Check that all C's are now 100's
-  Changed ? help popups in math4 to click instead of hover. 
  -  Check that they are click instead of hover.  You can find them on the set detail page when looking at a just-in-time set or with show me another enabled.   
-  Moved the more like this box (in the library browser) formatting to css and thickened it in math4. 
  -  Make sure more like this box is good in math3 and math4
-  Added maxlength to course_id fields in course admin.  
  -  Go to the various course_id fields in the admin course and check that they won't take more than 40 characters. 
-  Added subproblem grades to individual set grades when scoring jitar set. 
  -  Score a jitar sets and produce individual set scores.  The individual set scores should have both scores and adjusted scores for problems and sub problems.  
  -  Scoring of the totals table should not change, and individual scoring of other types of sets (gateway and default) should not change.  
- One extra change. I reordered stuff on the Config page a little. I tried to put the big optional modules all under one "Modules" list. 
  -  Take a look at the Config page and see how it looks.


goehle:
One extra change.  I reordered stuff on the Config page a little. I tried to put the big optional modules all under one "Modules" list.  Take a look at the Config page and see how it looks.

mgage:
1.Correct number of columns for jitar -- works (before the problem columns were printed without regard to the jitar subsetting)
2. C replaced by 100 -- check
3. The pop-ups work on clicks not hover.  But they seemed to do that in my version of develop also, before this pull --check
4. More like this box looks fine in math3 and math4 themes.  the difference seemed subtle to me.
5. Max length course names works.
6.Subproblems of jitar are reported in the scoring function
7.  The modules addition to the config page looks like a good improvement.  It's a little clearer where to look when you want to turn on one of the major features.

PR #676, opened 2016-01-17
Add message at bottom of Library browser searches explaining how more like this works.

jwj61:
Since more-like-this was introduced, several people have been confused by the fact that the number of problems shown does not match the number of database matches.  This adds an explanation at the bottom of the page just below the note "4 of 10 shown", so it is next to the offending comment, but is out of the way (bottom of page) of people who already know how it works.

To test, do a search in the library browser and look at the bottom of the page.


goehle:
Message received.

PR #677, opened 2016-01-18
Past Answer Refactor (and CSV)

goehle:
The intention was to add csv export to past answers and replace the regexp input fields with scrolling lists.  This ended up meaning that `ShowAnswer.pm` had to be almost completely refactored.  So extra testing is in order here to look for regressions. 

-  New Thing:  Test that you can generate a csv of a collection of past answers by using the "Create CSV" option.  Test that the csv is put in scoring and that backups are made. 

-  New Thing:  Test that you can select students, sets and problems in a reasonable way.  Test that selected, users, sets and problems are displayed when selected. 
  -  Something to look out for is that I have to do a reasonable amount of processing to figure out how many versions of a gateway to show.   So if you have some way to test this against a large class it would help check and see if things are too slow. 

-  Old Things to test:
  -  Test that answers from gateways, just-in-time sets and regular sets are all displayed correctly. 
  -  Test that answers are formatted reasonably.  Essay questions should be rendered.  Everything else shouldn't be rendered.  Check that it works with complicated problems.  Check that answers are colored appropriately.  
  -  Test that the links in Problem and Gateway work like they should.
  -  Test that student restrictions are in place (can't view other peoples answers, answers aren't colored).


mgage:
Feature request.   Since collating past answers for a large class over several homework sets takes a fair amount of time it would be nice if the creating of the csv could be deferred as it is with mass mailings.

goehle:
How would you get the download link in that case?

mgage:
You would get an email about completion of the event (as for mass emailings) and could then download the result using file manager.  Printing the result to the browser would not be useful in that case.  I'm doing 300 or so students over three homework sets on the MacBook.  It's taking more than 10 minutes to process.

goehle:
Is it faster when you are not creating the CSV?  Is that how long it takes if you try on develop?  If things have really slowed down then I should look into that.

mgage:
After modifications:

Rendering past answers for about 300 students, 14 sets, total 32790 lines in the .csv file -- took a little over a minute.  -- nice speed up.

mgage:
On gateway quizzes it looks like there is an entry for everytime the preview button is pushed.  Is this a feature or a bug?  Often the entries are blank because an empty answer is recorded when a previous answer in the test is previewed.  Being able to view all of the previewed entries might be useful for research purposes but it might just get in the way for other use cases.

mgage:
Previews are not recorded in past answers for ordinary homework.

mgage:
This is off topic but I'm seeing a vertical bar artifact at the end of numbers when using MathJax (but not when using images.  

image

goehle:
- The gateway [preview] / [submit] thing has been around forever.  You would have to ask Gavin about it. 

-  I'm seeing the vertical bar too.  Let me dig in and see if I broke something with MathJax.

mgage:
Student restrictions:  with my configurations students can't view past answers at all.

goehle:
That is by design.  You can change a permission to fix it.  When its enabled they should be able to view past answers, but only their own and it shouldn't color the entries.

mgage:
If I allow students to review past answers it appears that they can only view their own (in fact only the past answers for the problem they are currently looking at. )

mgage:
I've covered all the points above at least to some extent.  Were there any particular links in Problems and Gateways that concerned you?  I have run in to no issues clicking through homework and gateway quizzes.  

The only outstanding issue that should at least be thought about is whether to report every preview of a gateway quiz. (I believe that the previews are recorded to prevent data from being lossed if something happens in the middle of a quiz -- this is different behavior than for homeworks, I believe, where preview requests are not recorded.)

goehle:
-  I think you are right about previews.  I wouldn't mess with it since its expected behavior. The answers are tagged as preview so its not the worst. I could be convinced otherwise, though. 

- You checked all the links that matter. 

-  Could you try switching to "images mode" then viewing a bunch of past answers problems?  My only thought about the MathJax error is that PastAnswers is doing something wrong when it loads up its backup copy of MathJax.  I wasn't able to reproduce it, however.

dpvc:
>  I'm seeing a vertical bar artifact at the end of numbers when using MathJax

I'm assuming you are using Chrome.  There was a bug introduced recently in Chrome that causes this vertical line in MathJax version 2.5 and below.  MathJax 2.6 works around it, and I think the latest Chrome (or maybe it was the beta for the next version) also fixes it.

goehle:
Good to know.  Thanks for chiming in Davide.

On Fri, Jan 29, 2016 at 6:15 PM, Davide P. Cervone <notifications@github.com

mgage:
I had no troubles in image mode.  I'll update MathJax (and Chrome) later.  I'm ready to pull this unless there are objections.

goehle:
I"m ok with it.

PR #679, opened 2016-01-19
Date Shift Bug (master)

goehle:
Fixed date shift so it shifts reduced scoring.  To test, import a set with a reduced credit date and the date shift option and check that it also shifts the reduced scoring date.


mgage:
This works as advertised.  I think this patch is already incorporated into the develop version.

PR #680, opened 2016-01-19
Date Shift Bug (develop)

goehle:
This is the develop version of #679


mgage:
As far as I can tell develop already shifts the scoring date properly.

goehle:
Yeah.  Thats my fault.  I did some fancy stuff.  That change is in there, but it has a different commit number.  I don't know if it matters if we pull this or not, but I wanted to have a record.

PR #681, opened 2016-01-30
Automated Testing

goehle:
This is a pull request to add some automated testing elements. There isn't any change to WeBWorK other than adding some id's to the system.template file, so this shouldn't affect users at all.  What it does add is a framework for automated browser based testing as well as some basic tests which can be used to check that the major components of a fresh installation are working.  

To run the tests you will need to do: 
1.  The Test::WWW::Selenium package
```
sudo cpanm Test::WWW::Selenium
```
2.  The selenium server jar file from http://docs.seleniumhq.org/download/
```
wget http://goo.gl/76CpPA -O selenium-server.jar
```
3.  Start the selenium server.  (Note:  This starts a server listening on port 4444 and should not be left exposed to the open internet.)  
```
java -jar selenium-server.jar -log selenium.log
```
4.  Run tests either by using perl directly on the `.t` files in `webwork2/t/Selenium/Testing` or by using the `run_tests.pl` utility.  You should see a Firefox window pop up (you will need to have Firefox installed) and can actually watch the tests being run.  The results are reported in the same way that all Test::Harness results are reported. 
  -  Note:  The tests assume that you have an admin user with username admin and password admin.  You can override this by using the --uname and --pwd options on run tests. 
```
./run_tests --uname=admin --pwd=12345 --all-tests
```
Alternatively you can set the `WW_TEST_UNAME` and `WW_TEST_PWD` environment variables and then you can run the tests directly using perl with an alternate username and pwd.  
5.  You can run the tests "headlessly" (i.e. without the Firefox window or even a GUI desktop environment) if you use Xvfb.  
```
sudo apt-get install xvfb (or whatever works on your machine)
Xvfb :1 -screen 5 1024x768x8 &
DISPLAY=:1.5 java -jar selenium-server.jar -log selenium.log
```
This will be necessary if you are trying to use this on a server with no desktop interface.  

You can create your own tests by using the Selenium IDE tools at 
-  Selenium IDE Plugin: https://addons.mozilla.org/en-US/firefox/addon/selenium-ide/
-  Selenium IDE Perl Plugin: https://addons.mozilla.org/en-US/firefox/addon/selenium-ide-perl-formatter/
Once these plugins are installed you open the IDE from the Tools menu in Firefox.  You can then record your actions and export the resulting script as perl commands.  These can be used in the `skeleton.t` file to make new tests.  Importantly there are commands in the `Utilities` module which will automatically create a course, create a set, create a user, and delete the course when you are done, so these steps do not have to be part of your IDE script. 

Its possible that these tools shouldn't be part of the main WeBWorK repository and should be kept elsewhere.  The main question is if the system is robust enough, and easy enough to use, that we should encourage developers to use these tests on their own code and/or write their own tests when they submit new features.


mgage:
Could these installation instructions be added as a README to the Selenium directory?  They should probably stay with the test code.  Not everyone will install the test applications right away upon downloading WW.

mgage:
After getting the selenium driver set up and a fresh version of firefox running I get the results in the attached file.  I am using an existing site with my login name and password and many existing courses so some errors may result from that. 
[results.txt](https://github.com/openwebwork/webwork2/files/129480/results.txt)

goehle:
Are you providing an admin username and password on the command line?

mgage:
Yes.

Sent from my iPad

On Feb 13, 2016, at 10:05 PM, Geoff Goehle <notifications@github.com<mailto:notifications@github.com>> wrote:

Are you providing an admin username and password on the command line?


On Sat, Feb 13, 2016 at 8:17 PM, Michael Gage <notifications@github.com<mailto:notifications@github.com>>
wrote:

> After getting the selenium driver set up and a fresh version of firefox
> running I get the results in the attached file. I am using an existing site
> with my login name and password and many existing courses so some errors
> may result from that.
> results.txt
> <https://github.com/openwebwork/webwork2/files/129480/results.txt>
>
> -
> <https://github.com/openwebwork/webwork2/pull/681#issuecomment-183785267>.
>

-

goehle:
I was kind of worried about this.  Since Selenium is using the relative position of html elements to do its testing it can be very fragile.  I hoped the ID's I added would help but it doesn't seem to be enough.  I don't think its a good idea to include these test in the webwork2 repository because they won't be robust enough to run on existing installations.  I would only be confident in them running on completely fresh installed.  I'm going to remove all of the tests from this pull and create a separate repository.  I'll leave the ID changes to system.template since they have to be part of webwork2 and won't cause any harm.

goehle:
Since I removed most of this pull I decided to add a couple of things to make it meatier. 

1.  I added a flag to turn of the course total on the Grades page.  The reason is that for people who don't just use the total of all sets assigned the course total is misleading. 
  -  To test:  Use the Course Configuration page to enable and disable the flag and see that it does what it should. 

2.  I fixed a bug where every copy of a gateway was added to the course totals on the Grades page.  
  -  To test:  Create several  gateways and get different scores on multiple versions of each.   Then visit the grades page.  Before the patch each version is added as a separate set.  After the patch only the best score from each gateway should contribute to the course totals.

mgage:
Checked items 1 and 2.  Inspected code changes -- adding id's seems fine.

PR #682, opened 2016-02-04
Hardcopy Themes

goehle:
This pull creates the ability to have Hardcopy "Themes" and adds a new "oneColumn" theme which prints out homework with one (i.e no) column.   This pull should be tested alongside PG pull:  https://github.com/openwebwork/pg/pull/270

The main things to watch out for: 
-  There were duplicates of the "snippets" directory, so I consolidated everything into webwork2/snippets.  
-  I left links to the old snippet files available to support backwards compatibility.  I also left the option to override the snippet files using conf variables.  (These variables will override any theme choices.) 
-  I added some TeX to PG so that it will check to see if a "nocolumns" flag/method is set before it does anything with multicols.  I may not have done that in the cleanest possible way.  

To test: 
-  Test that you can print out hardcopies in one and two column mode.  Test all three homework types.  
  -  In particular for Gateways I was getting text like "Problem 2. 2." in hardcopies.  I commented out the code which caused it, but I'm not sure why it was there.  My suspicion is that it was added by someone who didn't use ` TEXT(beginproblem());` in their pg files.  
-  Test that you can switch and set the mode using the course configuration menu. 
-  Check that the various instructor features still work (multiple users/multiple sets) and that features are restricted for students.


mgage:
I'm pretty sure that the main snippets directory was conf/snippets and that the webwork/snippets directory was a temporary storage directory that never got removed.  Not absolutely positive.  In any case I agree that we only need one snippets directory and it's ok with me if its add webwork2/snippets.  

Does anyone have principled reasons why the directory should be located in any particular place?

mgage:
As long as we are modifying these files I suggest that we remove all of the copyright material in all the .tex files. This will be replaced with the text of the OPL_LICENSE https://github.com/openwebwork/webwork-open-problem-library/blob/master/OPL_LICENSE
placed only in the hardcopyPreamble.tex file. This will dramatically reduce the clutter in the TeX output of homework sets. The OPL_LICENSE is more appropriate than the WeBWorK GPL license since it is really the content of the homework sets that we are protecting.

Does anyone else have thoughts on this?

mgage:
I can print one column versions of JITAR, gateway and regular homework problems.

Feature request-- if it's not too much trouble.   At least on the instructors hardcopy download page I'd like to have an option to select the hardcopyTheme to override for that one printing the course default.  I suggest a pop-down menu which allows you to select one or two column.  This is because I think that more themes may be added --- for example a "worksheet" or "exam" version of the hardcopy download that puts space between each  problem.

Other than the feature request (and removing the extraneous copyright notices) this pull request passes.

goehle:
- Removed copyright stuff in TeX Snippets
- Moved all snippets to conf/snippets.  Removed snippets
- Added ability to choose theme on Hardcopy page.  Also added ability to have "nice" names for the themes.

mgage:
It looks very nice.  I'd be willing to keep one copy of the OPL_LIcense embedded in the tex preamble file, but I'm not sure it's really that useful.  I'm satisfied with the way things are.  

I'm actually going to find this ability quite useful.  It's almost at the point where you could easily print an exam from a webwork homework setby printing the TeX output and then doctoring slightly.  The only iffy part would be downloading the graphs and other auxiliary files properly.  At the moment that would have to be done by hand.  

I'll pull this.

PR #683, opened 2016-02-07
Reduced Scoring Validation Bug

goehle:
Because of missing ()'s the reduced scoring date was being validated even if the feature wasn't enabled.  

To test:  On each of ProblemSetDetail, ProblemSetDetail2 and ProblemSetList2 do the following: 
1.  Enable reduced scoring for the course and for the set. 
2.  Try to set the reduced scoring date to something not between the open and due date.  You should get an error. 
3.  Turn off reduced scoring for the set.  You should be able to then save the reduced scoring date to something outside the open/due range.  
4.  Turn on reduced scoring for the set and fix the date. 
5.  Turn off reduced scoring for the course and change all of the dates the open date is past the old reduced scoring date.  There shouldn't be any errors.  

In addition I made a first attempt at dealing with the confusion surrounding reduced credit dates.  A number of people have said that to them it makes more sense for the "Reduced Scoring Date" to come after the "Due Date".  The problem with making this change is that Due Date has been synonymous with "Set Close Date" as long as WeBWorK has been around.  Changing it so that sets close either on the due date or the reduced scoring date is a project. 

However, I did make changes to how reduced scoring dates are reported to students to try and reinforce how the reduced scoring date works.  In general the reduced scoring date is now given as much prominence as the due date.  In particular I changed it so that: 
-  For homework and gateways if there is am enabled reduced scoring date, and it is between the reduced scoring date then the reduced scoring date is shown on the Problem Sets page.  I chose **not** to display the due date in this case so that there is no possibility that students confuse the two dates.  
-  If there is an enabled reduced scoring date and it is between the reduced scoring date and the due date then the due date is shown, along with a message that the reduced scoring period started.  
-  If there is an enabled reduced scoring date and it is between the open date and reduced scoring date then the Problem Set page now displays the reduced scoring date in the main page title.  There is a message about what the reduced scoring date means, as well as the actual due date printed below.  If students want to know the actual final due date of the set then this is where they would find it. 
-  If there is an enabled reduced scoring date and it is between the reduced scoring date and the due date then the due date is displayed in the title.  There is also a message that the reduced scoring period has started. 
In addition, I  ended up also redoing a lot of the code that generated the due date status text in ProblemSets since there was a lot of duplication.  

To Test: 
1.  Look at the status column entry on the Problem Sets Page (and the title and message on the Problem Set page when appropriate) in a variety of situations.  Technically any combination of the following is possible: 
-  The set is either a regular homework, a gateway, or a specific version of a gateway. 
-  The current time is between the open date and reduced scoring date, between the reduced scoring date and the due date, between the due date and the answer date, and after the answer date. 
-  Reduced scoring is enabled or disabled. 
-  Restrict Set Progression is enabled or disabled and if its enabled then there is one or more than one prerequisite set.

### DIFF
--- a/lib/WeBWorK/AchievementItems.pm
+++ b/lib/WeBWorK/AchievementItems.pm
@@ -332,7 +332,7 @@ sub new {
 	name => "Ring of Reduction",
 	#Reduced credit needs to be set up in course configuration for this
 	# item to work,
-	description => "Enable reduced scoring for a homework set.  This will allow you to submit answers for partial credit for limited time after the due date.",
+	description => "Enable reduced scoring for a homework set.  This will allow you to submit answers for partial credit for 24 hours after the due date.",
 	%options,
     };
     
@@ -394,9 +394,8 @@ sub use_item {
 	($set);
 
 
-    # enable reduced scoring on the set and add the reduced scoring period 
-    # to the due date.  
-    my $additionalTime = 60*$ce->{pg}{ansEvalDefaults}{reducedScoringPeriod};
+    # enable reduced scoring on the set add 24 hours to the due date.
+    my $additionalTime = 86400;
     $set->enable_reduced_scoring(1);
     $set->reduced_scoring_date($set->due_date());
     $set->due_date($set->due_date()+$additionalTime);
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -1058,9 +1058,9 @@ sub initialize {
 
 		my $enable_reduced_scoring = 
 		    $ce->{pg}{ansEvalDefaults}{enableReducedScoring} && 
-		    defined($r->param("set.$setID.enable_reduced_scoring")) ? 
+		    (defined($r->param("set.$setID.enable_reduced_scoring")) ? 
 		    $r->param("set.$setID.enable_reduced_scoring") : 
-		    $setRecord->enable_reduced_scoring;
+		    $setRecord->enable_reduced_scoring);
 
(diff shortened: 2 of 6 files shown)


mgage:
Changing dates works as advertised.

mgage:
I have a suggestion for the labeling on the homework set.   
Currently it says:
After the reduced scoring period begins all work counts for 75% of its value.  The set is due 08/09/2016 at 11:31am EDT

Better might be:
After the reduced scoring period begins on 08/07/16 all works counts for 75% of its value.

The due date could be placed in the headline after the course:
Calculus II -- Due 08/09/2016 at 11:31am EDT

(this is in line with how the due date is placed if we are in the middle of the reduced scoring period. or if reduced scoring is not enabled at all.)

Everything else seems to be working fine.

goehle:
So I'm actually avoiding showing student the "due date" when the reduced
scoring period is enabled on purpose. In fact, I should probably remove the
portion "The set is due [_1]".  The reason is I want to emphasize the
reduced scoring date.  I think once you have both the reduced scoring date
and the due date on the same page at the same time you set yourself up for
confusion.  That is why I replaced the due date in the title with the
reduced scoring date.  Maybe it wont matter once we change to "close date"
but my guess is students will still get confused.

mgage:
I’m ok with that.  

If I remember right there is one weird case.  If you enable
reduced scoring and set the reduced scoring date and the due date equal to each other
then you get no headline or text at all on the Problem Set page.    I think.   I’ll double
check that when I check everything else but if so perhaps you’d like to make special case for that.

Take care,

Mike

goehle:
I just checked and if the reduced scoring date is equal to the due date then it just shows the due date with no reduced scoring message.  This is by design since in this case there is no functional reduced scoring period.

mgage:
Everything checks.

PR #684, opened 2016-02-08
LTI Bug

goehle:
Fixes the bug at http://bugs.webwork.maa.org/show_bug.cgi?id=3588

To test: 
1.  Import a homework set but date shift so that the open date is far in the future.  Only assign the set to yourself. 
2.  Log into WeBWorK via LTI as a new student (so that sets are automatically assigned to you.)  
3.  Log in as the instructor again and see how many students are assigned to the set from 1.  Before the patch the newly created student should be assigned to it (but will be missing problem data).  After the patch you should still be the only one assigned to the set. 

P.S.  Yes there are two commits to add 41 characters of code.  I should have tested before committing the first time!

### DIFF
--- a/lib/WeBWorK/Authen/LTIBasic.pm
+++ b/lib/WeBWorK/Authen/LTIBasic.pm
@@ -637,7 +637,7 @@ sub authenticate
 								$db -> addUserSet($userSet);
 								}
 							}
-						else
+						elsif ( $globalSet -> open_date < $open_cut and $globalSet -> due_date < $due_cut )
 							{
 							if (not $db -> existsUserSet($userID, $globalSet -> set_id ) ) {
 								$userSet = $db -> newUserSet();
--- a/lib/WeBWorK/Authen/LTIBasic.pm
+++ b/lib/WeBWorK/Authen/LTIBasic.pm
@@ -637,7 +637,7 @@ sub authenticate
 								$db -> addUserSet($userSet);
 								}
 							}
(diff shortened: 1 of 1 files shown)


paulvojta:
It's not clear to me why the block of code in lines 641-651 exists at all.  It seems to me that whenever this code is reached, it is creating the bug.

goehle:
Its for when someone doesn't set alterations to the due date and answer
date when students join late.  It does work if its a set that has problem
data generated by the for loop at 601.  Really all of that should be
removed and this should be refactored to use the standard AssignSetsToUser
methods.

Geoff.

mgage:
"Really all of that should be
removed and this should be refactored to use the standard AssignSetsToUser
methods."

I second the above but I decided that it didn't have to be done right away.  

Nice catch on finding the line that was causing the trouble.  I was having difficulties getting the thing to fail consistently.

mgage:
Before the patch, using Blackboard for LTI.  The problem set is assigned with adding problems.  it's interesting that the student is registered for the course (and the problem) as soon as the Blackboard page with the link to the course is displayed even BEFORE the link to the course is clicked.

mgage:
After the patch:

1. Reset an existing problem set (with 7 fake class members) so that its open date is far in the future. 
2. Make sure that the LTI student (MEG) is deleted from the course.
3. Open the Blackboard page linking to the course -- 
4. MEG is assigned to all problem sets which have open dates in the past (including sets that are now closed)  and is not assigned to sets with open dates far in the future.


All this looks correct.  I'll pull this.

PR #686, opened 2016-02-11
Achievement Upgrades

goehle:
This pull requests gives several upgrades to achievements. 
 1.  Achievements can now be used with gateway and just-in-time sets.  There is an option when you edit an achievement to select which sets that particular achievement is valid for. 
 2.  Achievements are now ordered according to a "Number" field.  Its not quite as nice as drag and drop numbering, but its better than being ordered according to the achievement_id field which couldn't be edited.  
 3.  There are now some achievement items for Gateway Tests. 
 4.  Other minor changes including: 
  -  Using Text::CSV to deal with importing and exporting achievements. 
  -  Adding maketext calls to achievement items. 

To test:  
 1.  Set up a course with achievements already enabled, imported and assigned.  
 2.  Pull this branch and then run a db update on the course.  (The branch adds two new columns to the Achievements table.)  
 3.  Go to the course which had achievements.  The achievements will be missing the number column, but they should still have the same order and if you edit them they should all have the "homework" set type enabled.  
 3.  Delete the achievements and import the default achievements.  Check that achievements are imported from the older style file correctly.  
 4.  Set up a new course copying from the latest version of modelCourse
 5.  Import the default achievements from the new course.  They should all have numbers and should  be assigned to appropriate set types if you use "edit all sets".  
 6.  Check that you can edit achievements and that all of the fields work correctly.  
 6.  Create an achievement with an evaluator that just contains `return 1;`  Test that you can earn this achievement in all three homework types when that type is enabled for the achievement.  
  -  You can unassign achievements you have earned by clicking in the Edit Users column of Achievement Editors. 
 7.  Create an achievement with the evaluator 

  ```
$globalData->{AddNewTestGW} = 1;
$globalData->{ExtendDueDateGW} = 1;
$globalData->{RessurectGW} = 1;
return 1;
  ```
  And earn it.  Check that the three new items work.  (Make sure you have items enabled.)
  -  One item gives you an extra version of a gateway per time period. 
  -  One item extends the due date of the gateway by 24 hours. 
  -  One item reopens a gateway.


mgage:
Goeff:  #686 on achievement upgrades needs reconciling of conflicts before it can be pulled.

goehle:
Fixed conflicts with develop.

mgage:
3.  When I edit the achievement by clicking on the pencil none of the check boxes (homework, gateway, just in time) are checked.

mgage:
(3,4) When I delete and import new achievements the homework box is checked.

mgage:
When I tried to test the imported achievements on a gateway test I got this message:

```
WeBWorK Warnings

WeBWorK has encountered warnings while processing your request. If this occured when viewing a problem, it was likely caused by an error or ambiguity in that problem. Otherwise, it may indicate a problem with the WeBWorK system itself. If you are a student, report these warnings to your professor to have them corrected. If you are a professor, please consult the warning output below for more information.

Warning messages

There were errors in achievement 01_complete_one_problem
Can't call method "status" without a package or object reference at (eval 10784) line 6.
Re
```
(this is with the old achievements package)

mgage:
The first step achievement worked correctly when triggered from a homework set.

mgage:
I don't seem to be able to trigger achievements from jitar problem sets in the original course.  Now copying and setting up new course.

mgage:
In the newly created courses the achievements now have numbers in the 100's -- before they were consecutive integers starting from 0

mgage:
And I was able to trigger achievements from homework, gate-way and from jitar sets.

goehle:
-  I added some extra code to help with transition.  Now if you have achievements in place when you switch over to the new code the achievements should be set to regular homework only when you either submit a problem for the first time or when you view the achievement editor page for the first time.  
-  You turned on the achievement 01_complete_one_problem for gateways but that achievement is not valid for gateways, which is why you are getting the error. 
-  You shouldn't be able to get any achievements on just-in-time sets when using an existing set of achievements since they are only enabled for regular homeworks be default.  
-  The newly created achievements are supposed to have larger numbers.  If it is numbering existing unnumbered achievements then it numbers them sequentially.  However, in general its better to use numbers which are spread out so that re-ordering things is easier.

mgage:
(9)  This achievement works.

In general everything works, except that there are a few questionable things during the change over -- e.g. if you use this new system on an old course.  Fixing the "status" error message might be desirable -- the questionable numbering of old achievements is not really a problem.

I'd like to suggest that we relabel Achievement Items to "Awards"  or to "Prizes"  or even "Gold stars" or "Super Powers" so that they are not so easily confused with the MathAchievement badges themselves. 

That doesn't have to be done for this pull request however.

mgage:
Ok.  I take back the comment about "status"  in that case.

mgage:
I think the pull request is ready to go.  Relabeling "achievement items" is something I'd like to consider but it's a completely separate issue from all of the others.

Any other fixes you want to add?

goehle:
No, I am fine with things as they are.  The change over will be a little rough for people who re-use the same course semester after semester.  For people who create new courses each semester it will be a little easier since they will be importing achievements into the new system (and not using in place ones).

PR #687, opened 2016-02-25
LTI Grade Passback and Refactor

goehle:
This pull request contains a complete rewrite of the LTI authentication module as well as upgrades to allow grades to be passed back from WeBWorK to the LMS (e.g. Blackboard, Moodle, etc...)  The major features are: 
-  A complete rewrite of LTIBasic.pm.  Acutally, LTIBasic.pm is left as it is so that people who want to keep using it can use it.  The new authentication module is LTIAdvanced.pm.  The authen_LTI.conf.dist file has been rewritten to reflect this. Features of the rewrite include
  -  Debug mode for testing parameters (old). 
  -  The ability for the system to add new users and keep users up to date (old). 
  -  New admin defined functions which can modify a user and their sets after they are added (new).  In particular some of the more specific Indiana stuff was removed.  The theory is that that level of modification could be done with these user functions.
  -  Assigning sets now uses the standard Instructor utility functions (new). 
  -  Nonces are now managed entirely via apache/webwork.  There is no need to run a cron job anymore.  (Most people probably forgot this step anyway).  
-  The ability for WeBWorK to pass back grades.  There are currently two and a half grading modes. 
  -  Course Grading Mode:  In this mode you create a single Link/Assignment in your LMS which points to your WeBWorK Course address.  E.G. In Blackboard
  ```
Name: WeBWorK
URL: https://webwork.university.edu/webwork2/CourseName
Check Yes:  This link is to a Tool Provider
Enable Evaluation: Yes
Points Possible: 100
  ```
    Students will receive a LMS grade associated to the Link/Assignment which is determined by their percentage total homework grade in WeBWorK. This total homework grade is the same as the grade that is reported on the grades page.  You can change the weight of sets by altering the weight of  the problems in the set. 
  -  Set Grading Mode:  In this mode you create a single Link/Assignment for *every* homework set in  the course.  The address should be the address of that set in the Course. E.G. in Blackboard
  ```
Name: Set Name
URL: https://webwork.university.edu/webwork2/CourseName/SetName
Check Yes:  This link is to a Tool Provider
Enable Evaluation: Yes
Points Possible: 100
  ```
    Students will receive a grade for each Link/Assignment which is determined by their percentage homework grade on the Set which the Link/Assignment points to.  Students need to use the Link/Assignment in the LMS at least once to enable grade passback.  In particular when working in this mode it is recommended that you only allow students to log in via the LMS.  However there are navigation features in place to make sure that students can't access a set that doesn't have the information it needs to send grades to the LMS. 
  -  Problem Grade Mode:  This is more of a proof of concept using the XMLRPC interface.  Basically you create a link to a single problem with all the parameters that you would find in `TestWW.html`.  This will use xmlrpc to render the problem and as long as the LMS is providing the necessary information to pass the grade back the xmlrpc system will attempt to do so.  However, its not ready for prime time for a number of reasons.  
  -  Note:  For the first two methods there are some caveats:  
    -  First, only the grades are passed back.  In particular nothing else about the Link/Assignment in the LMS and the homework set in WeBWorK is syncronized.  For example , the total number of  points/problems, the due dates, the open dates are not kept in sync automatically.  
    -  Students need to log into the course/set at least once (depending on the grading mode) but after that they don't need to go through the LMS.  In particular if you are using course grades as long as their accounts are created by the LMS they can log in however they want.  
    -  There are two grading flags that can be set in the configuration file.  One enables passing grades back every time a student hits submit.  This keeps student grades up to date, but could be a drain on the server.  I don't know how well it holds up at larger institutions, for example.  
    -  The second important flag controls how often WeBWorK attempts to do a "mass grade update".  Because grades can change when instructors add sets, add problems to sets, delete problems from sets, change weights of problems, mark problems correct, etc... once per 24 hours the system spins off a background process which updates all of the grades.  
-  Because I couldn't help myself there are also a couple of extra additions. 
  -  I made some minor formatting changes to single column mode for hardcopies to make it look better. 
  -  I added maketext calls to a couple of modules.


goehle:
Installation instructions:  
1.  Download the branch and restart the server.  Then do a database upgrade.  You will need to run `upgrade_admin_db.pl` since a column has been added to the `user` table.   
2.  Set up your WeBWorK server in an LMS.  I've done a reasonable amount of testing with Canvas and Blackboard so it would be good if someone works with Moodle a bit.  I have included a screenshot below with the configuration information I used in Canvas.  There should be similar fields in the other LMS's.  
screenshot from 2016-02-25 17-19-42
  Note that in some LMS's (like Blackboard) there is an additional setting you have to set if you want your "external LTI tools" to also provide grades.  
3.  Set up the `authen_LTI.conf` file.  The one thing you have to set is the "consumer secret".  This is the "password" that authenticates *all* LTI transactions.  Other things you should mess with.
  -  You can set `debug_lti_parameters` to get a lot more information about what is going on if you run into trouble.  The default error messages are vague on purpose.  
  -  You may need to make the nonce lifetime longer if two two servers are not well synced timewise.  
  -  The grading feature is off by default.  Its a good idea to test it out once with the `debug_lti_parameters` flag on.  *However* many LMS's don't keep track of "instructor" grades which means you will need a test student account to actually make the passback work.

goehle:
Testing the authentication capabilities:
-  Test that you can log in. 
-  Test that new users are created as long as their permission level is below `$LTIAccountCreationCutoff`
-  Test that if you change the user (i.e. first name/last name) and `$LMSManageUserData` is on then the user data is reset next time the user logs in.  (This only works for users with permission level less than `$LTIAccountCreationCutoff`. 
-  Test that `$preferred_source_of_username` and `$strip_address_from_email` work.  
-  Test that when users are added they get visible sets correctly assigned to them. 
-  Test that both sample "local routines" work.  The first gives users a random student id (which can be sent out and used as a password) and the second changes due dates on newly assigned sets.  
-  Check that if the nonce lifetime is too short or the secret is wrong then authentication fails. 
-  Check that if you log in via the LMS and then refresh the login fails (because the same nonce is submitted and you can only use a nonce once).  

Testing the Grading Mechanism (course version): 
-  Enable "course" grading and log into WeBWorK via the LMS as a student.  (Grades don't work for instructors because many LMS's don't actually keep grades for instructors.) 
-  Get a problem right on a homework set and check that the grade is automatically updated in the LMS. 
-  Check that turning off `LTIGradeOnSubmit` means that the grade doesn't update right away but that making the `$LTIMassUpdateInterval` short and then visiting any page does update the grade.  (The mass update is triggered when ContentGenerator is activated.)  
-  Create a just-in-time set (with graded sub-problems) and a gateway set.  Get some problems right on each and make sure that the grade in the LMS matches the "Total Homework" grade on the Grades page.  (In particular Gateways use the highest score on a version and just-in-time sets use adjusted status' so you should make sure both are being done correctly.) 
-  Create a user via WeBWorK, log in via WeBWorK, and submit a problem.  Check that there is a message about the LMS grade not being updated correctly.  Log in via an LMS and check that it does update correctly.  Log out and log in via WeBWorK and check that it does update correctly.  

Test Grading Mechanism (homework version): 
 -  Set the grading mode to "homework" and have a just-in-time set, a homework set and a gateway set.  Create a different link in the LMS for each of them. 
  -  Note:  For the gateway you will need to use the appropriate gateway address.  E.G. 
    ```
http://webwork.wcu.edu/webwork2/TestCourse2/quiz_mode/Gateway/
    ```
-  Visit each set via the LMS link and get some problems right.  Check that all three grades are reported back to the LMS correctly (including gateway best versions and adjusted status').  
-  Create a new set in WeBWorK.  Now navigate to the Problem Sets page.  For students there should be a message that they have to log in to the set via the LMS.  If you put the set name in the html and try to access the set that way you should get a similar error.  
-  Check that instructors see the messages but can access the set anyway.  
-  Check that the messages go away once you have access the set once via the LMS and that if you switch between sets in WeBWorK the grades still get updated in the right places on the LMS.  
-  Recheck the `LTIGradeOnSubmit`  and the `$LTIMassUpdateInterval` options.  All three set grades should be updated after the mass update.

Test Grading Mechanism (problem version): 
-  Create a link in the LMS similar to one from `TestWW.html` 
  ```http://webwork.wcu.edu/webwork2/html2xml?                                 
        &answersSubmitted=0                                                    
        &sourceFilePath=Library/LoyolaChicago/Precalc/Chap1Sec2/Q03.pg         
        &problemSeed=123567                                                    
        &courseID=TestCourse                                                   
        &userID=<userid>                                                          
        &displayMode=MathJax                                                   
        &password=<password>                                                
        &outputformat=simple"
  ```
  You may need to set the parameters somewhere other than the URL.  For example see the "section" and "recitation" parameters in the Canvas screenshot in my previous post.  
-  Add the following hash to your .conf file.  
  ```
LISConsumerKeyHash = {<consumerKey> => '<consumerSecret>'};
  ```
  This hash contains key/secret pairs for the xmlrpc server.  In theory you could have one WeBWorK server serving a number of different LMS instances. 
-  Click the link in the LMS and check that the problem loads.  Check that correct grades are sent back.  Note:  This is feature incomplete.  Currently Solutions are always shown, the seed is always the same, and there are conflicting messages about scores being recorded.

Other things to test: 
-  Test that after 24 hours if you log in via the LMS and then check the `key` database using mysql all of the nonces from the previous day have vanished. 
-  Test that generating a single column hardcopy looks good.  (I changed the margins a bit.) 
-  Trigger as many of the newly "maketext"-ed messages in `Authz.pm` as you are able. 
-  Trigger as many of the newly "maketext"-ed messages in `GatewayQuiz.pm` as you are able. 
-  Doublecheck my spelling on anything that is being printed!

mgage:
May need to add Net/OAuth to your CPAN collection (it's already listed as being necessary in check_modules.pl)

These are the results for the single sign on.  Everything works with moodle except that
1. Moodle doesn't fill in the $lis_person_sourcedid variable (as far as I can tell).
2. I didn't find a local routine for assigning a random user id but the other local routines worked.

* Results
	* Won't enroll instructor
	* Will enroll student -- with email address entered into Moodle
		* Assigns all visible sets
		* Asks whether one can trust the site to use Java (eventually clicked NEVER)
	* Looks like grade_set is called on every problem set when I look at grades in WeBWorK.		
	* Changing name works -- can I change my email?
	* can change section and recitation name but the global setting overrides the setting for a specific link 
	* stripping address from email works
	* Moodle does not fill out $lis_person_sourcedid so can't test that
	* local subroutine modifying section works
	* local subroutine modifying due dates works
	* I didn't find a local routine for assigning a random user id. 

Next up -- changing the grade return.

mgage:
I don’t see how students who are registered via an LMS can log in directly to webwork — they don’t know their password — or do they?

I don’t see how students who are registered via an LMS can log in directly to webwork — they don’t know their password — or do they?  this seems fine to me — but I don’t understand the comments about checking if these students get certain messages if they log in directly

I also can’t change the password for a student — I get “can’t call method “password” on an undefined student.  I get similar message if I try to change password as a student (probably should not even see the “change password” option in the user settings panel.  it also appears that a student is allowed to change their email address — but it doesn’t seem to register in the database.

goehle:
-  The routine for modifying a user before adding it is `$LTI_modify_user` and is in the conf file.  At some point I thought you might generate a random string and use it as the student id, but that doesn't set the password so you can ignore that.  (I dont see it in the conf file documentation, if it is somewhere else let me know.) 
-  There are a couple ways that students can log into WeBWorK directly.  The first is if you add them to WeBWorK and not via the LMS.  The second is if your LMS uses something like LDAP.  Then you can have your WeBWorK server also use the LDAP authentication module and students will be able to log in directly to WeBWorK with the same password they use for Blackboard.  Third, I'm pretty sure you could use the `LTI_modify_user` subroutine to add a random student id to a student and *also* set it to be their password.  (You get the $self object so you have access to the database handler.)  Then it would generate random intitial passwords which you could send out sing mail merge.  It all seems like a lot of effort.  
-  As for "checking certain messages".  Create two links pointing to different Homework sets and set things to "homework set grading mode".  Log in via one link and then in WeBWorK navigate to the main Problem Sets page.  It should prevent you from accessing the second homework link until you log in via the appropriate LTI link.  
-  Students can change their email address, but if you have it set to keep things up to date with the LMS then the student changes will get overwritten once they log in again.  You could set the permissions so that doesnt happen.  I'm a little wary about automatically trying to adjust the permissions though. 
-  The problem is that things aren't set up to change passwords if there is no password entry in the database.  I made some changes so that students with no password can now set a password if they so choose (and instructors can also set passwords, even if the students password is not set initially).  You can test by creating a student via log in and changing their password.  Of course, this will be confusing if external authentication is turned off.  If people are using turning off external authentication they should probably also turn off password changing for students.

mgage:
OK.  That's clear enough.  For wwassignment the login assigned the student a random password (which it never uses since all successive authentication uses the session key).  This is when a login is created -- if the user already exists then the existing password is left unchanged.  This effectively locks students out of accessing WW directly without going through moodle.  In some cases I found that to be a feature in terms of making sure students read certain messages and so forth although it could also be annoying for students used to logging in to WeBWorK directly.

goehle:
With the current setup, and feel free to test this, if you log in for the
first time using the LMS then you won't have a password and you will not be
able to log in via WeBWorK.  If you set a password, then you will be able
to log in via WeBWorK, but a professor could set permissions so that you
can't set a password.  So you should have both kinds of functionality.

mgage:
* Results -- problem set grade mode
	* homework, JITAR and GatewayQuiz all report grades back correctly
	* LITGradeOnSubmit and LTIMassUpdateInterval work
	* Messages about which sets have been assigned work for both student and instructor
* Results --- course grade mode
	* the course grades seem to work
	* Need to redo this with a simpler course to make sure the percentages are right.

Left to do -- check the xmlrpc problem grading mode 
-- Check some more random methods of trying to login to LMS and/or webwork


Question:  I'm assuming that if you want one WW course to do course grading and another WW course to do set grading then you enable the switches in course.conf.  Is that what is expected?

mgage:
This seems to do everything that wwassignment does -- or very nearly so --

goehle:
I had intended people to use course.conf to set course specific settings.  If it turns out that this is easy enough to us and popular enough we could add a tab to the config page.  (If we add it in the auten_LTI.conf file it would only appear if the conf file was being used.)

mgage:
LISConsumerKeyHash  should be $LISConsumerKeyHash in the instructions

mgage:
* Results problem display via xmlrpc
	*  LISConsumerKeyHash  should be $LISConsumerKeyHash in the instructions
	* and password = <password>  should be
course_password =<password>
	* Problem path should be Library/LoyolaChicago/Precalc/Chap1Sec2/Q03.pg
	*   there were two typos in the problem path
	* Homework problem grade updates properly
Things look good.

mgage:
The same setup, without any changes in the WeBWorK course, works for Canvas on problem set access

mgage:
I'm ready to pull this.  It's pretty complicated so there might be some combination of settings and LMS that hasn't been tested but the results so far show no problems.

goehle:
I"ve fixed the typos in the instructions above.

mgage:
Great.  Do you have any objections to my pulling this request?

Mike

goehle:
Nope.  In theory this shouldn't affect users not using the LTI or using the old LTIBasic so it should be pretty safe.

PR #688, opened 2016-03-08
handle corrupt previous answers in decodeAnswers

bldewolf:
We got a report that a user was unable to open a problem they had previously
attempted and they were receiving this error instead:

	Can't use an undefined value as an ARRAY reference at
	/opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 871.

The error refers to decodeAnswers, which is trying to use the result of a
Storable::thaw call without verifying that it succeeded (it returns undef on
some errors).

This change makes decodeAnswers catch those errors.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -877,7 +877,7 @@ sub decodeAnswers($) {
 	my $serialized = shift;
 	return unless defined $serialized and $serialized;
 	my $array_ref = eval{ Storable::thaw($serialized) };
-	if ($@) {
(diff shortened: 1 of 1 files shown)


goehle:
Thanks for submitting this fix to WeBWorK.  I tested this by forcing `$array_ref` to be undefined.  Without the patch there is a fatal error.  With the patch the problem loads (without the previous answers.) 

As a side note, we generally only accept pull requests into the develop branch.  This change is small and so I will merge it directly into master.  In the future if you could make your changes on a branch of develop and submit the pull request back into develop it would fit our workflow better. 

Thanks again.

PR #690, opened 2016-03-12
Add russian localization files from Ruslan Sharipov

mgage:
Localization files for the Russian language contributed by Ruslan Sharipov.
and some changes to ContentGenerator.pm to aid localization.

Question -- should we make decode('utf8',$r->maketext("Library Browser"))  part of the maketext macro so that it can be used everywhere?  Is there ever a case where one does NOT want to use decode?


lib/WebworkClient/test_formats.pl
```diff
@@ -1,33 +0,0 @@
-#!/usr/bin/perl -w
-use strict;
-use 5.10.0;
```
goehle: Did you want to delete this file?

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -701,7 +703,7 @@ sub links {
 	print CGI::h2($r->maketext("Main Menu"));
 	print CGI::end_li();
```
goehle: Its a little wierd that we are making these `decode` changes here, but everywhere we are using make text.  You mentioned this in your email chain, but would it make more sense to overload maketext and have the change be universal?  

Another option would be to move over to something like https://metacpan.org/pod/Locale::Maketext::Gettext This package merges `maketext` and `gettext`.  In particular `gettext`is multibyte safe which should fix some of the utf8 stuff.

mgage: Hi Goeff,

You’ll see a recent email which was forwarded to you in which Ruslan makes some observations that would
encourage us to overload maketext to include decode.  

I hadn’t see the Maketext::Gettext  module.  That might be an even better solution to this problem.
It’s worth checking out before we decide on a definite strategy.  Any conjecture on whether this 
new package is likely to be the successor to the previous packages?

Take care,

Mike

goehle: I think Maketext::Gettext is more of a "transitional" package to help older code.  Newer applications probably just use gettext directly.  In either case, I would remove the changes from ContentGenerator.  Then, either in this pull or a future one I would either overload `maktext` or add this package to Locale (which will effectively do the same thing.)

goehle:
Could you pull my changes into this branch using your github?  You should see testing instructions there.  https://github.com/mgage/webwork2/pull/17

goehle:
I'm pulling this.  It checks out, and is also somewhat overshadowed by the localization pull request.

Issue #685, opened 2016-02-08
Change "Due Date" to "Close Date"


goehle:
One problem I have been thinking about is that there has been a lot of confusion about how reduced credit works because people expect the "partial credit due date" to be after the "due date".    They don't like that the current reduced credit date comes before the "due date" of the set. However, for WeBWorK we have always used due date to mean "close date".  This makes it tough to implement a reduced credit system where the set stays open after the due date.   One alternative is to try to make the current implementation easier to understand.  I was thinking about going through and changing the term "Due Date" to "Close Date" in all ContentGenerator pages.  The up sides are that this is a better description of what the date does and that its fairly safe.  (If I miss something it just means the wrong text is displayed on screen.)  The down side is its wide reaching change (influencing documentation, not just code) and that it may not solve the underlying problem.  Any thoughts?

pstaabp:
Geoff,

Yes.  I get tons of confusing from both students and faculty.  I think this terminology sounds good.

PR #691, opened 2016-03-31
Localization Upgrades

goehle:
-  Created a script `update_localization_files` which will update `webwork2.pot` and all of the po files.  In theory we can ask any developer making a pull request to run this.  That will keep our language files a little more up to date.  
-  Changed all mentions of "due date" to "close date" and mentions of sets being "due" to sets "closing" as per #685 
-  Added `maketext` calls to many new things.   So many maketext calls... 
-  Added a dummy utility function `x` which can be wrapped around strings and leaves them unchanged.  This is for when the string is defined as a constant (and `$r->maketext`) isnt present.  The constant can be fed to `maketext` for translation and the `x` function around the string definition allows `xgettext.pl` to scrape it for the pot file.   
-  Minor Changes: 
  -  Added "xXx" to the `maketext` definitions to help with debugging.  This is a temporary change and should be reverted before the pull is merged. 
  -  Deleted some blocks of code that had been commented out for a long time.  

To test:  Check out this branch, as well as the corresponding PG branch (https://github.com/openwebwork/pg/pull/273)  The entire interface basically needs a once over.   Each page (that we still support) should be visited and interacted with.  Things to look for: 
1)  The functionality of the page should be unchanged. 
2)  All non-user generated strings should be wrapped in "xXx" to indicate that they have been run through maketext. (Including transient strings like notifications and messages.)
3)  All mentions of due date should be close date

If you find missing maketext strings you can post about it here and I will fix them, or you can make a pull request to the `localize` branch on my github and I will pull those changes there.  

In some ways this pull request is not completely done.  The template pot file has been updated to include a ton of new strings and all of the po files are very out of date.  Keeping the translations current, though is likely to be an ongoing project.  So if you know someone who is bilingual and likes WeBWorK let them know.   

Edit:  We are managing the actual translation of the pot file at https://www.transifex.com/webwork/webwork2/  You should direct people here if they are interested in doing translation for WeBWorK.


goehle:
Fixed conflicts with develop.

mgage:
Hi Goeff.  The localization pull request now has conflicts probably because there were some maketext changes in the earlier achievements PR.  It's probably best if you compare the 5 files that have simultaneous changes and make choices.  I'd probably get the choices right but there is a chance that I wouldn't understand which is the right one.

I'll check and pull the localization PR as soon as you have the conflicts resolved.

goehle:
Conflicts resolved.  I also updated the po files to include strings added by #686

mgage:
OK. It’s all good so far.  This review may take awhile.

Take care,

Mike

mgage:
The edit button on the course info and problem set info (problem header ) pages blends in with the background instead of standing out in white.  --- this might be because the xXx brackets are interfering with the style sheets or the javaScript

goehle:
It is because of the xXx.  The xXx will be removed before this is pulled
and the edit button should go back to normal.

mgage:
what else needs to be done here?  Is this ready to pull?

goehle:
I would need to remove the xXx brackets before we pull.  Another possibility is to solicit feedback from various translators.  It would be a good time for them to take a look at the pot file (I put the latest one on Transifex) to see if any of the strings need to be reworked.

goehle:
xXx debugs removed.

goehle:
On second thought lets pull this.  There are changes in a lot of files and there is more chance that we will generate conflicts with feature branches the longer this sits around.

mgage:
I agree.

PR #692, opened 2016-04-03
maketext call for Course Administration

heiderich:


heiderich:
Sorry, I did not notice the pull request #691 before. I will try to remember to follow the instructions there the next time.

goehle:
Thanks again for helping with localization.  The changes in this pull are actually already included in #691, along with many other maketext calls.  Since you have been active in localization you may be interested in participating in that pull request.  In particular we need help on the following: 

1.  The `webwork2.pot` has seen a huge increase in the number of strings, making our existing `po` files out of date.  If you have been doing translation work (or working with a translator) you might consider translating the new webwork2.pot file and contributing the results. 
2.  There are a lot more strings passed through maketext now, but some are probably still missing.  You could help with that too. 

Thanks again for your contribution.  However, since the contents of this pull are replicated elsewhere I'm going to close it.

Issue #693, opened 2016-04-06
Can't locate object method "mk_ro_accessors"


dlglin:
Just installed the develop branch, and every time I submit an answer I get:
Can't locate object method "mk_ro_accessors" via package "WeBWorK::Utils::AttemptsTable" at /opt/webwork/webwork2/lib/WeBWorK/Utils/AttemptsTable.pm line 187.

A quick googling says that this is provided by Class::Accessor.  Should I be installing that package?  If so, it should be added to check_modules.pl.

goehle:
Yes, you should install that package.  We did add `Clas::Accessor` to `check_modules.pl`, but only for the develop branch (https://github.com/openwebwork/webwork2/blob/develop/bin/check_modules.pl#L45).  Let me know if it doesn't work for some reason. 

Geoff.

dlglin:
I didn't read the output of check_modules.pl very carefully.  I do have it installed, so that's not the problem.  Here's the whole stack trace:
Error messages

Can't locate object method "mk_ro_accessors" via package "WeBWorK::Utils::AttemptsTable" at /opt/webwork/webwork2/lib/WeBWorK/Utils/AttemptsTable.pm line 187.
Call stack

The information below can help locate the source of the problem.

in WeBWorK::Utils::AttemptsTable::new called at line 315 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm
in WeBWorK::ContentGenerator::Problem::attemptResults called at line 1834 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm
in WeBWorK::ContentGenerator::Problem::output_summary called at line 155 of /opt/webwork/webwork2/lib/WeBWorK/Template.pm
in WeBWorK::Template::template called at line 560 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::content called at line 356 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm
in WeBWorK::ContentGenerator::Problem::content called at line 232 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::go called at line 382 of /opt/webwork/webwork2/lib/WeBWorK.pm

goehle:
[...]

PR #694, opened 2016-04-06
AttemptsTable Fix

goehle:
Added "base" to AttemptsTabe.  This is to address issue #693.  I'm not entirely sure why it works on some installations and not others.  This addition fixes the issue for Danny and is based off this example:
```
  package Foo;
  use base qw(Class::Accessor);
  Foo->follow_best_practice;
  Foo->mk_accessors(qw(name role salary));
```
from http://search.cpan.org/~kasei/Class-Accessor-0.34/lib/Class/Accessor.pm

To test:  Check out the branch, restart, and try to answer a problem.  Make sure that things still work with the attempt table.

### DIFF
--- a/lib/WeBWorK/Utils/AttemptsTable.pm
+++ b/lib/WeBWorK/Utils/AttemptsTable.pm
@@ -139,14 +139,14 @@ of the default summaries are created:
 
 =cut
 
+package WeBWorK::Utils::AttemptsTable;
+use base qw(Class::Accessor);
+
 use strict;
 use warnings;
 use Scalar::Util 'blessed';
 use WeBWorK::Utils 'wwRound';
 use CGI;
-
-package WeBWorK::Utils::AttemptsTable;
-use base qw(Class::Accessor);
 
 # Object contains hash of answer results
 # Object contains display mode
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/Utils/AttemptsTable.pm
```diff
@@ -141,12 +141,14 @@ of the default summaries are created:
 
 use strict;
 use warnings;
-package WeBWorK::Utils::AttemptsTable;
 use Class::Accessor 'antlers';
```
mgage: This line use Class::Accessor 'antlers' needs to be moved below the "use base " line in order for the file to compile on my machine. 
After that it appears to work fine.

goehle:
Made the suggested changes.  

@dlglin: Do you want to try this out and make sure it works on your machine?  It seems like this package stuff is sensitive to different versions of perl.

dlglin:
Works for me (on CentOS 6.6, perl 5.10.1)

Danny
On Apr 13, 2016, at 7:16 AM, Geoff Goehle <notifications@github.com<mailto:notifications@github.com>> wrote:


Made the suggested changes.

@dlglin<https://github.com/dlglin>: Do you want to try this out and make sure it works on your machine? It seems like this package stuff is sensitive to different versions of perl.

PR #698, opened 2016-04-28
Increase space allotted to keywords in the database.

jwj61:
There was a report of OPL-updating because a pg file had keywords which were too long.  This would fix it.

### DIFF
--- a/bin/OPL-update
+++ b/bin/OPL-update
@@ -186,7 +186,7 @@ if($libraryVersion eq '2.5') {
 '],
 [$tables{keyword}, '
 	keyword_id int(15) NOT NULL auto_increment,
-	keyword varchar(65) NOT NULL,
+	keyword varchar(256) NOT NULL,
 	KEY (keyword),
 	PRIMARY KEY (keyword_id)
 '],


goehle:
I just ran across this today, along with some other OPL-stuff.  (In Ubuntu 16.10 mysql is more strict, so there are errors when OPL-update tries to process a problem with an empty "edition" field.)

In any case, this pull request should really be targeted at 2.12.  You have two options,  you can redo the changes off of a fork of 2.12 and resubmit or you can just do nothing and wait for me to submit a pull for  my 2.12 bugfixes branch since it also includes a fix.

jwj61:
Waiting is fine with me.  Does that mean I should close this?

goehle:
Yes.  Also, as an FYI, there are also Library Browser issues in Ubuntu 16.04.  `ListingsDB.pm` has some sql searches that have an `ORDER BY` on a field that is not being selected and this causes an error.  This is because of a configuration change which adds `ONLY_FULL_GROUP_BY` mode to sql.  There is a workaround here:  http://www.sitepoint.com/quick-tip-how-to-permanently-change-sql-mode-in-mysql/  but I am also putting fixes directly in ListingsDB.

PR #702, opened 2016-06-07
Comment out pretty_print command that is causing a segmentation 11 fa…

mgage:
…ult (on macbook at least)

### DIFF
--- a/lib/WeBWorK/Utils/AttemptsTable.pm
+++ b/lib/WeBWorK/Utils/AttemptsTable.pm
@@ -269,7 +269,7 @@ sub formatAnswerRow {
 		# push @correct_ids,   $name if $answerScore == 1;
 	} elsif (($rh_answer->{type}//'') eq 'essay') {
 		$resultString =  $self->maketext("Ungraded"); 
-		# $self->{essayFlag} = 1;
+		$self->{essayFlag} = 1;
 	} elsif ( defined($answerScore) and $answerScore == 0) { # MEG: I think $answerScore ==0 is clearer than "not $answerScore"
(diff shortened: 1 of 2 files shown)


mgage:
As far as I can tell this variable is never used so commenting it out should not have any effect.
If pretty print is called (i.e. it's not commented out) then when using clients/sendXMLPRC.pl or the standalonePGrenderer  or the opaque_server (see my repo on standalone renderer and the openwebwork repo on opaque_server)  a segment 11 fault is generated as the WebworkClient is used to format output.

goehle:
Note:  This flag used to be used to indicate if a lone essay answer would be graded later. 

https://github.com/openwebwork/webwork2/blob/WeBWorK-2.11/lib/WeBWorK/ContentGenerator/Problem.pm

but the new messaging is different and no longer uses the flag.  However, the flag is stilled used for gateways

https://github.com/goehle/webwork2/blob/develop/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm

so if this is ever ported over to gateways then the essayFlag will have to be removed there.

goehle:
As an aside, this hotfix to master was not done correctly.  There should have been a single branch off of master and two pull requests, one to master on openwebwork and the other two develop on openwebwork.  By having two feature branches with the same code changes but different commit numbers there is a greater chance that conflicts will be created down the road.  This is the case here because in each branch you commented out the pretty print line in WebworkClient in a different way.  Its my fault for not noticing until after I merged, so I will fix the conflicts by pulling the "master_hotfix" branch into develop.

mgage:
ok -- thanks for fixing this for me.

PR #703, opened 2016-06-07
Comment out a pretty print of $self  which was causing a segmentation…

mgage:
… fault when WebClient's formatting routine was called.

As far as I can tell this variable is never used so commenting it out should not have any effect.
If pretty print is called (i.e. it's not commented out) then when using clients/sendXMLPRC.pl or the standalonePGrenderer or the opaque_server (see my repo on standalone renderer and the openwebwork repo on opaque_server) a segment 11 fault is generated as the WebworkClient is used to format output.


mgage:
Fix error in Attempts table that incorrectly set the essayFlag indicator.
This fixes the bug reported by Andras Balogh at
http://webwork.maa.org/moodle/mod/forum/discuss.php?d=3912

Test by making sure that Library/Rochester/setAlgebra15Functions/p3.pg
responds correctly to an incorrect answer -- and that the essay questions also still respond correctly

PR #704, opened 2016-06-07
Develop localize

mgage:
Works with change in pg to pass the localization subroutine into Safe directly


mgage:
Withdrawn because there were some changes of variables from dueDate to closeDate

PR #705, opened 2016-06-07
Change that sends localization subroutine directly to Safe compartment.

mgage:
companion to pg pull request develop_localize2


goehle:
I tested this and the localization subroutine is still getting to pg.

PR #707, opened 2016-06-14
Hardwire taxo

jwj61:
This solves the following problem with the tagging widget (used to set OPL tags for a problem).  In the library browser, there are many problems, so many instances of the widget, and they all need the taxonomy.  We had been trying to load it asynchronously, but in some browsers, particular on windows computers, it would fail.  There was a timing issue/problem with making the ajax call since the widget needs the taxonomy before it initializes.

This fixes it by just putting the taxonomy directly in the web page (in a script variable).  Ultimately, the same amount of data gets sent to the browser.

In both versions (old and new), we relied on the json file with the taxonomy, generated by OPL-update.    We now do a javascript alert if we cannot read that file.

This affects to places - the library browser and the display of an individual problem, but only when the user authorization to modify tags.


goehle:
This pull request is to master, when it should be to develop.  I will close it and make the develop version.

PR #709, opened 2016-06-17
Hardwire taxo

goehle:
This solves the following problem with the tagging widget (used to set OPL tags for a problem). In the library browser, there are many problems, so many instances of the widget, and they all need the taxonomy. We had been trying to load it asynchronously, but in some browsers, particular on windows computers, it would fail. There was a timing issue/problem with making the ajax call since the widget needs the taxonomy before it initializes.

This fixes it by just putting the taxonomy directly in the web page (in a script variable). Ultimately, the same amount of data gets sent to the browser.

In both versions (old and new), we relied on the json file with the taxonomy, generated by OPL-update. We now do a javascript alert if we cannot read that file.

This affects to places - the library browser and the display of an individual problem, but only when the user authorization to modify tags.

### DIFF
--- a/bin/OPL-update
+++ b/bin/OPL-update
@@ -186,7 +186,7 @@ if($libraryVersion eq '2.5') {
 '],
 [$tables{keyword}, '
 	keyword_id int(15) NOT NULL auto_increment,
-	keyword varchar(255) NOT NULL,
+	keyword varchar(256) NOT NULL,
 	KEY (keyword),
 	PRIMARY KEY (keyword_id)
 '],
--- a/htdocs/js/apps/TagWidget/tagwidget.js
+++ b/htdocs/js/apps/TagWidget/tagwidget.js
@@ -13,13 +13,14 @@ var basicRequestObject = {
 };
 
 function readfromtaxo(who, valarray) {
+  var mytaxo = taxo;
   if(who == 'subjects') {
-	return(taxo.map(function(z) {return(z['name']);} ));
+	return(mytaxo.map(function(z) {return(z['name']);} ));
   }
   var failed = true;
-  for(var i=0; i<taxo.length; i++) {
-    if(taxo[i]['name'] == valarray[0]) {
-	  taxo = taxo[i]['subfields'];
+  for(var i=0; i<mytaxo.length; i++) {
+    if(mytaxo[i]['name'] == valarray[0]) {
+	  mytaxo = mytaxo[i]['subfields'];
 	  failed=false;
 	  break;
 	}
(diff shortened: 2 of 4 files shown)


goehle:
@jwj61 Could you add a check so that the taxonomy information is only printed if 
```
($r->authz->hasPermissions($r->param('user'), "modify_tags")
```
(I.E. Put the code that prints the taxo variable in the if block a couple lines up where the `tagwidget.js` file is getting printed.  Otherwise the taxonomy file is getting opened and included every time the problem page is rendered, even if its only a student checking their answer.

jwj61:
Moved the taxonomy code as suggested

mgage:
This often works but not always.  If I skim through several problems in a row in a homework set I get this:
tagwidget.js:30 Uncaught RangeError: Maximum call stack size exceeded
fetch_taxo @ tagwidget.js:30
fetch_taxo @ tagwidget.js:34
fetch_taxo @ tagwidget.js:34
fetch_taxo @ tagwidget.js:34
fetch_taxo @ tagwidget.js:34
fetch_taxo @ tagwidget.js:34
fetch_taxo @ tagwidget.js:34
fetch_taxo @ tagwidget.js:34
fetch_taxo @ tagwidget.js:34f
etch_taxo @ tagwidget.js:34

I was using the course https://hosted2.webwork.rochester.edu/webwork2/2014_07_UR_demo/Calculus_I/
(I can give you admin access to that.)
On my mac laptop I get timeout errors regularly, but I think there might be a configuration error on the laptop -- I'll try to track it down.

mgage:
In the library browser things seem to work ok.  There are several zillion calls to instructorXMLHandler
but only one to tagging-taxonomy.json.

mgage:
Fixed the timeout errors on the laptop by increasing the number of MaxClients from 6 to 10. (I was keeping the number of clients low to check for errors.) I do not use lighttpd on the laptop and at least when using the chrome browser I think there were too many simultaneous requests to the server for a single page.  Things got blocked. 

I can still trigger the tagwidget error on occasion:  
``` 
// If needed, wait until asynchronous load is done
function fetch_taxo() {
  if(taxo.length>0) {
    return(taxo);
  } else {
    loadtaxo.done(fetch_taxo());
  }
}
```
I'm not sure this is serious. I only notice it because I'm watching the console.

mgage:
As best I can tell the tagwidget report above is erroneous.  I don't think that hosted2 was updated properly to the latest software (or there was some caching in the browser).  Those lines of code exist in the old tagwidget but not the new.  I've rebuilt everything on hosted2 and can no longer get the error.

mgage:
I'll pull this with the comment that at least with the chrome there need to be enough clients available to handle the extra ajax calls.

jwj61:
Incidentally, the code you quoted was the source of the problem I was trying to fix, and this pull request deleted it.

mgage:
Yeah -- I finally figured that out.  Something went wrong when I pulled your PR and the fix didn't take on the one machine (hosted2) so I was testing the old code.

PR #710, opened 2016-06-17
Pretty Print (develop version)

goehle:
This is the correct version of #702.  It gets the commits that were part of the master hofix in #703 into develop and fixes conflicts.


goehle:
Conflicts fixed and merged via command line.

Issue #708, opened 2016-06-15
Archiving queue not working


Alex-Jordan:
I'm trying to archive a lot of courses, and I encountered one that had already been archived. When I click "Skip archiving this course" it just keeps trying the same course instead of moving on to the next one.

goehle:
Fixed in #711

PR #711, opened 2016-06-17
Skip Archive Bug

goehle:
This fixes a bug where using the "Skip Archiving This Course" button didn't actually do anything.  See #708. 

To test:  Check the archive functionality of the admin page.  You should be able to: 
-  Archive courses. 
-  Delete courses as part of archiving them. 
-  Archive multiple courses in a row.  
-  Skip archiving a course, unless it is the last course in the list in which case the skip button should not appear.


mgage:
Works fine.  One remark -- possibly a feature/bug?? -- When archiving a list of courses you either archive and delete all of them (except for the ones you skip) or you delete none of them.  I was surprised by this at first but on reflection the current behavior might be for the best.

Issue #715, opened 2016-06-28
update-OPL-statistics doesn't escape special characters in passwords


dlglin:
When I run update-OPL-statistics (either on its own or through OPL-update), everything works until the last statement, where I get
ERROR 1045 (28000): Access denied for user 'webworkWrite'@'localhost' (using password: YES)
This comes from [https://github.com/openwebwork/webwork2/blob/master/bin/update-OPL-statistics#L202](url), and I believe the problem is that my DB password contains an ampersand, which is treated differently at the command line.  Is it possible to either escape the special characters in the password used in this command, or quote the password verbatim?

goehle:
In `upload-OPL-statistics` I use shell quote to quote the database password.  I must have just forgot to add that to `update-OPL-statistics`.  In any case I think it would solve the problem. 

https://github.com/openwebwork/webwork2/blob/master/bin/upload-OPL-statistics#L46

PR #716, opened 2016-06-28
Quote DB Passwd

goehle:
Add quoting for mysql command in `update-OPL-statistics`.  To test, make sure that the `update-OPL-statistics` command works. Bonus points for testing with a special character (like &) in your db password. 

Addresses Issue #715


dlglin:
This works for me.  I think that this should be backported to 2.12.  Without it, those with special characters in their passwords won't be able to use the Library Browser.

goehle:
Closed so a hotfix branch can be created.

PR #719, opened 2016-07-13
Quote DB Passwd (master)

goehle:
This is the master hotfix version of #716


goehle:
Pulling since there is confirmation that #716 works.

PR #721, opened 2016-07-15
converted file encodings to utf8

heiderich:


goehle:
Thanks for the contribution.  

I'm going to close this and instead pull it into my UTF8 feature branch.  That will keep things a little more organized.

PR #724, opened 2016-07-22
PGProblemEditor and CodeMirror

goehle:
This pull requests adds the option to use a CodeMIrror based code editor in PGProblemEditor2 and PGProblemEditor3 instead of a vanilla text area.  You can read about CodeMirror here https://codemirror.net/ but it is basically a javascript text editor for code.  New features include: 
-   Line numbers. 
-  Tab inserts spaces and doesn't change focus. 
-  Colors (with extra modifications made for PG)

Potential additional features include
-  Autocompletion 
-  Code Folding
-  Bracket Matching
-  Trailing Space Highlighting
-  Lots of other addons listed here:  https://codemirror.net/doc/manual.html#addons

I don't actually write that many problems so I am open to suggestions about what features would be useful.  (I don't want to add them all because I prefer to keep things simple as possible.)  In particular if people have been editing PG problems using more advanced editors like vim or emacs, what features do you appreciate.  

To Test: 
-  Checkout the branch and edit a problem using the editor 2 and editor 3. 
-  Check that the new editor works and that you can go back to the old one by setting
``` 
$options{PGCodeMirror} = 1;
```
-  Look at the coloring on problems and see if it is sane.
-  Edit some problems and see if its an improvement and if any of the existing plugins would be helpful.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -918,6 +918,17 @@ $options{useOPLdefFiles} = 0;
 # gap betweens cores in score dropdown - e.g. 5 gives 5, 10, 15, 20
 $options{problemGraderScoreDelta} = 5;
 
+##########################################################################################
+#### Default settings for the problem editor pages
+##########################################################################################
+
+# This sets if the PG editor should use a js based "codemirror" editor or
+# just a textarea
+$options{PGCodeMirror} = 1;
+
+# This sets if mathview is available on the PG editor for use as a minimal latex equation editor
+$options{PGMathView} = 0;
+
 ###########################################################################################
 #### Default settings for the PG translator
 #### This section controls the display of equations, HINTS, answers, SOLUTIONS, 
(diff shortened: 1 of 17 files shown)


goehle:
I added three things based off of local suggestions at WCU: 
1.  I added search functionality to the editor.  (The browser search doesn't work with codemirror).  Use Ctl-S to search for something. 
2.  I added bracket matching. This highlights matching brackets when you are next to them. 
3.  I changed the behavior of the "New Version" tab so that it only puts "local" in front of the file name if you can't write to the directory.   Since the Library directories are not writable (or at least they shouldn't be) this means that you will get local versions of those, but not of files already in your templates directory.  This could be controversial for people who like to have custom writable directories that are also shared between classes.

goehle:
Added basic coloring for stuff inside EV3 and PGML blocks so its not all one big red blur.   I'm willing to take suggestions on what should be colored what.  Should the blocks even be colored red or should they be black?  What about latex?  

Added trailing whitespace marking for PGML blocks.  This should help you find out if you have one, two or three trailing spaces.

mgage:
I think this snippet should be added (commented out) to localOverrides.conf.dist

```
# This sets if the PG editor should use a js based "codemirror" editor or
# just a textarea
$options{PGCodeMirror} = 1;

# This sets if mathview is available on the PG editor for use as a minimal latex equation editor
$options{PGMathView} = 0;

```

goehle:
Added config text.

mgage:
I would like to pull this so that it is easier for others to test. Since it can be turned off 
it won't get in the way if it still has some minor bugs.  Is this ok with you Goeff?

goehle:
No problem with me.  Just being able to tab means its probably enough of an upgrade already.

Issue #725, opened 2016-07-28
LTI Grade passback - Desire2Learn issue with newline


dlglin:
Desire2Learn won't accept the grade passback because the XML block has a newline at the end.  Adding
`chomp($replaceResultXML);`
to line 212 of SubmitGrade.pm fixes the issue in D2L.  I've tested the altered code in Moodle, and it doesn't break anything.  I don't have access to any of the other LMSs to test.

goehle:
Thanks for tracking down this fix.  I"ll test it out in Blackboard and
Canvas and make a pull.

Geoff.

goehle:
Submitted fix.

goehle:
P.S.  If you could find time to add instructions for Direct2Learn to the wiki that would be amazing. 

http://webwork.maa.org/wiki/LTI-Advanced_Authentication

http://webwork.maa.org/wiki/LTI-Advanced_Grading

PR #726, opened 2016-07-28
Fix for Desire2Learn LTI passback

goehle:
This fixes issue #725 as suggested.  Danny has tested this with Desire2Learn and Moodle.  I have tested this with Blackboard and Canvas.  Grade passback works as expected in all cases.  

To Test:  Check that LTI grade passback works with your LMS set up.  

P.S.  I also added a more complete error message for troubleshooting the LTI authentication process.  This is already in master, but you can the code change in the diff for #727.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -580,10 +580,12 @@ sub maybe_update_user {
     # Create a temp user and run it through the create process
     my $tempUser = $db->newUser();
     $tempUser->user_id($userID);
-    $self->{last_name} =~ s/\+/ /g;
-    $tempUser->last_name($self->{last_name});
-    $self->{first_name} =~ s/\+/ /g;
-    $tempUser->first_name($self->{first_name});
+    my $last_name = $self->{last_name} // '';
+    $last_name =~ s/\+/ /g;
+    $tempUser->last_name($last_name);
+    my $first_name = $self->{first_name} // '';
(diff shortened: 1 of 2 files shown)


mgage:
This works fine with blackboard.  

With canvas I get this message.

Warning messages

    Use of uninitialized value in substitution (s///) at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 581.
    Use of uninitialized value in substitution (s///) at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 583.
    Use of uninitialized value in string ne at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 602.
    Use of uninitialized value in string ne at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 602.

This is because Canvas, at least with the settings I have, doesn't pass back 
 'lis_person_name_family'  'lis_person_name_given'
so that 'first_name' and 'last_name' aren't defined
	

I am using hosted2 as the webserver and the link UR105 Homeworkset Demo hosted2 
in my course Test WeBWorK -- with UR105 on hosted2

goehle:
I've added some checks for this, but I would argue that fundamentally your canvas settings are wrong.  It should be set up to give webwork "full" access.

mgage:
Your checks work.  And you are right I had the convas configured for email only. Configuring it for "public" also solved the problem of warning messages.  I'll pull this.  I'll let you reconcile this hotfix with develop and pull it into develop.

PR #727, opened 2016-07-28
Fix for Desire2Learn LTI passback Develop

goehle:
Develop version of #726.


mgage:
@Goehle   This has conflicts at the moment.  They will need to be resolved.

goehle:
Because this is a hotfix pull request, the way this will have to work is that #726 will have to be pulled first.  Then develop can be pulled into the d2lfix and the conflicts fixed.  Then this can be pulled.

PR #730, opened 2016-08-03
Maketext in achievement user editor

heiderich:
I could not test this, because I do not see how to reach the corresponding page. Though, I believe that the syntax used previously was not correct (there should be no variables directly in the strings).


heiderich:
Sorry, I should have created this against develop. I will open a new pull request and close this one.

PR #737, opened 2016-08-04
Changed the form of a maketext construction containing two dropdown lists

heiderich:
I would suggest to use this kind of construction instead of something like $r->maketext("foo").<dropdown stuff>.$r->maketext("bar") whenever possible.
The reason is that languages have different word order and sometimes it is difficult to translate parts of an English sentence into another language while preserving the order.
Using the construction I adopted here, the translator is free to change the position of the dropdown lists in the sentence.


goehle:
These forms are relics from before we really did a lot of localization.  I've slowly been trying to change things to be a bit more grammar neutral.

goehle:
Also, if you want you can group these maketext pull requests together.  It might save on creating lots of tiny pulls.

PR #739, opened 2016-08-05
XSS Fixes

goehle:
This fixes some XSS stuff.  At the request of a security person I also changed the response code from 200 (OK) to 403 (FORBIDDEN) when `Apache/WeBWorK.pm` throws an error.  Note, the response code is still 200 (OK) if `ContentGenerator.pm` is the module that throws the error. 

To test
1.  Try the following URLS and check that the `<script>` tag is encoded (i.e. visible)
 - `http://yourserver.edu/webwork2/<script>`
 - `http://yourserver.edu/webwork2/admin/<script>`
2.  Visit the above links with a webpage inspector open to the Network tab and check that the error response is 403 and not 200.

### DIFF
--- a/lib/Apache/WeBWorK.pm
+++ b/lib/Apache/WeBWorK.pm
@@ -30,6 +30,7 @@ details.
 
 use strict;
 use warnings;
+use Apache2::Const qw(:common);
 use HTML::Entities;
 use HTML::Scrubber;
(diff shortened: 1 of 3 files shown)


mgage:
This works as described.  Even before the patch the <script> generated an illegal character error, but this is safer.

Issue #742, opened 2016-09-03
develop with code mirror has bug when codemirror is not enabled. 


mgage:
In develop, with the codemirror patch installed, set the value 

`$options{PGCodeMirror} = 0;`
in `localOverrides.conf`

The tabs for the PGProblemEditor will loose formatting -- both the ones above the textarea and the tabs for choosing the action in saving the edit.  The tabs at the bottom also loose the ability to respond to clicks.  

Restore 
`$options{PGCodeMirror} = 1;`

and everything works fine.

-- also reported as bug #3726

goehle:
Could you check and see if a hard refresh fixes the issue?  I can't
reproduce the bug and I've seen similar issues when math4.js is out of
date.

Geoff.

goehle:
I figured out the issue and fixed it in a pr.

mgage:
Thanks.  I did do a hard refresh, but thanks for the suggestion. Sometimes I overlook that possibility.  I'll go look at the pull request now.

PR #743, opened 2016-09-03
CodeMirror Bug

goehle:
Fixes issue #742.  Test by setting `$options{PGCodeMirror} = 0;`  Both PGProblemEditor2 and PGProblemEditor3 should be properly formatted.


mgage:
Works for me.  Thanks Goeff.

PR #744, opened 2016-09-07
Codemirror Updates

goehle:
Two small changes:
1.  I enabled line wrapping for the editor.  There are lots of long lines in PG problems and this helps with that. 
2.  I removed some debugging code from PG.js which was left in by accident.

### DIFF
--- a/htdocs/js/apps/PGProblemEditor2/pgproblemeditor2.js
+++ b/htdocs/js/apps/PGProblemEditor2/pgproblemeditor2.js
@@ -18,6 +18,7 @@ $(function () {
 	     indentUnit: 4,
 	     tabMode: "spaces",
              lineNumbers: true,
+	     lineWrapping: true,
              extraKeys:
              {Tab: function(cm) {cm.execCommand('insertSoftTab')}},
 	     highlightSelectionMatches: true,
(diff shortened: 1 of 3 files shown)


pstaabp:
This seems like a straightforward PR that will by default line wrap in the editor.  Perhaps this could be made a course/user configurable parameter, but I'm fine pulling this through.  Any other thoughts before I do so?

mgage:
Let's make this a configurable parameter and put it in default.conf.  (to be overridden in localOverrides.conf or in course.conf. ).  Can you figure out how to pass this parameter into the .js script?
I haven't looked.

pstaabp:
I can do this.  Should I just take this branch as one of my own and resubmit as a new PR?

mgage:
Actually let’s pull the current PR and make a new one.

I would like to add a few more changes in the configuration of the editors.
Specifically — make all of the url’s for the help buttons that appear above the textarea
box  defined in defaults.conf instead of being hard coded in the PG editor.
(at the moment the configuration for PGML is out of date since MAA changed courses.webwork.maa to 
courses1.webwork.maa).  

I think for release ww2.13 I’d make the editor3 the only one that is shown by default — unless there are objections — there are few differences between editor2 and editor3.

Any other minor changes needed for the PGProblemEditor page?

Take care,

Mike

PR #748, opened 2016-11-04
Sanitize ScoringDownload  getFile parameter.  Fixes bug #3793

mgage:
This fixes at least part of bug #3793.  Now the file name cannot contain / which prevents the exploit described in #3793


mgage:
To test:  Enter the following url as per bug #3793:

https://webwork.swarthmore.edu/webwork2/coursename/instructor/scoringDownload?getFile=../../../../../etc/passwd

You should get an error complaining about the path components.

pstaabp:
Looks like anything with a / generates an error.  I also tried a ~ such as `?getFile=~pstaab` and we get a 404 error.  Should we say the same error for that as well?

mgage:
OK.  I added a check for ~ anywhere in the "filename".  Anything else?

PR #752, opened 2016-12-24
Path to Rserve.pm in defaults.config

daverosoff:
This eliminated some "Failed to evaluate module" errors that appeared in every problem whether it used R or not. I'll be happy to open another PR to the `develop` branch if that's appropriate.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1219,7 +1219,7 @@ ${pg}{modules} = [
 	[qw(Locale::Maketext)],
 	[qw(WeBWorK::Localize)],
 	[qw(JSON)],
-        [qw(Rserve Class::Tiny IO::Handle)],
+        [qw(Statistics::R::IO::Rserve Class::Tiny IO::Handle)],
 ];
 
 ##### Problem creation defaults


mgage:
Danny, could you take a look at this? I haven't used this branch but I can't imagine that this change would do any harm.

dlglin:
I believe that this is pointing at the wrong Rserve package.  This should load Rserve from /opt/webwork/pg/lib/Rserve.pm, which in turn loads Statistics::R::IO::Rserve.
If you check out the 2.12-r branch of both webwork2 and pg, this should not be necessary.

mgage:
Thanks Danny.  Dave, can you confirm that this works? If so we can close this PR. 
Do we need some kind of warning so that others don't fall into this trap of not updating
pg as well as webwork2 to 2.12-r?

daverosoff:
Hi Mike and Danny,

Thanks Danny and Mike. I changed both the `webwork2` and `pg` repos on our server to point to `2.12-r`, which seems to have done the trick. I just didn't think to change `pg` before. Danny is quite right. I just didn't put all this together from the forums posts. I think a warning of the kind you suggest would be helpful. I don't know about other schools but we've quickly come to rely on the WeBWorK-R bridge. I'll leave it to you to close the PR. Thanks for looking into this.

dlglin:
The instructions on the wiki show updating both webwork2 and pg.  Is there somewhere else where we could put the warning where people would notice it?

Danny

On Jan 30, 2017, at 10:44 PM, Dave Rosoff <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi Mike and Danny,

Thanks Danny and Mike. I changed both the webwork2 and pg repos on our server to point to 2.12-r, which seems to have done the trick. I just didn't think to change pg before. Danny is quite right. I just didn't put all this together from the forums posts. I think a warning of the kind you suggest would be helpful. I don't know about other schools but we've quickly come to rely on the WeBWorK-R bridge. I'll leave it to you to close the PR. Thanks for looking into this.

daverosoff:
You're quite right, I think the wiki directions are completely adequate. I was trying to connect the dots from forums posts and didn't think to check it. I'll certainly be more thorough next time. Thanks again.

PR #760, opened 2017-02-28
add maketext calls

heiderich:


mgage:
This is a small change and works fine.

PR #761, opened 2017-02-28
add maketext calls to SetMaker.pm

heiderich:


mgage:
These are small changes and work fine.

PR #768, opened 2017-06-05
Search bpl

tknayak75:


mgage:
Hi,

Thanks very much for the submission.  There is a lot of useful translation work here.

We merge new submissions into the develop branch first so it would be helpful if you could close this pull request and resubmit it to the develop branch.  There will probably be fewer changes and it will be easier to evaluate and check the new files.  (see pull request #761 for an example  of a submission to the develop branch  https://github.com/openwebwork/webwork2/pull/761 ) 

I want to pull some of the previous pull requests correcting bugs in the mechanism for saving past answers first.  Then I'll split off a candidate for ww2.13 which we'll try to get ready for release toward the end of June.  With luck I can do the split with in two weeks or so 

At that point we can start folding in the various localization pull requests to the new develop branch after the split.  If we have enough help testing the localization results we might be able to have code that is well localized by the end of the summer.  That's a little ambitious but we'll see how it goes.

Take care,

Mike

mgage:
Never mind about closing the commit and submitting to develop.  I figured out how to do that from this end.  That reduces the files changed from 130+ to 54 so that looks a little less daunting. :-)

dpvc:
@mgage, technically, this is still a pull from master to develop, which is not allowed by the release model that we are using.  Although this branch has had develop merged into it, it originated from a branch off master originally.  That means that it should really be rebased off develop.  I don't know how much of a stickler you want to be for that, but we've ben pretty good about that since we adopted the modified git-flow model.  It was hard to get things straightened out and working that way, so I'd hate to see it revert to the earlier more chaotic situation.

tknayak75:
Please disregard. It was by mistake

PR #769, opened 2017-06-14
Fix bug by initializing variable text

jwj61:
One line fix to the mentioned bug.  Without this, the variable $text might be tested without initializing it.


pstaabp:
Looks like this does what is intended.

PR #773, opened 2017-06-21
Develop scoring download bug fix2

mgage:
A more robust check of the fileName using the FileManager  checkName function.  This insures
that the fileName for the scoring file conforms precisely to the same rules used by the FileManager for files.

This pull request is not quite ready yet.  I have to implement a similar check on the scoring tools page so that you cannot enter an incorrect file name there. I'll do that shortly.

Sorry to put you through the extra work -- I caught this problem this morning around 8 and I thought I had closed the pull request, but I guess I didn't push "commit and close" on the github web page.


mgage:
Ooops.  This is supposed to go to develop.

PR #774, opened 2017-06-21
Improve file name check for scoring download exploit bug.  

mgage:
A more robust check of the fileName using the FileManager checkName function. This insures
that the fileName for the scoring file conforms precisely to the same rules used by the FileManager for files.

This pull request is not quite ready yet. I have to implement a similar check on the scoring tools page so that you cannot enter an incorrect file name there. I'll do that shortly.

Sorry to put you through the extra work -- I caught this problem this morning around 8 and I thought I had closed the pull request, but I guess I didn't push "commit and close" on the github web page.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
@@ -122,8 +122,8 @@ sub initialize {
 		if (!$scoringFileNameOK) { # fileName is not properly formed
 			$self->addbadmessage($r->maketext("Your file name is not valid! "));
 		    $self->addbadmessage($r->maketext("A file name cannot begin with a dot, it cannot be empty, it cannot contain a " .
-				 "directory path component and only the characters ~[-_.a-zA-Z0-9 ~]  are allowed.")
-			); # ~ is needed to escape [ and ] within maketext
+				 "directory path component and only the characters -_.a-zA-Z0-9 and space  are allowed.")
+			); 
 		}
 	}
 	
--- a/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
@@ -29,6 +29,7 @@ use warnings;
 use WeBWorK::CGI;
 use WeBWorK::Debug;
 use WeBWorK::Utils qw(readFile seq_to_jitar_id jitar_id_to_seq jitar_problem_adjusted_status wwRound x);
+use WeBWorK::ContentGenerator::Instructor::FileManager;
 
 our @userInfoColumnHeadings = (x("STUDENT ID"), x("login ID"), x("LAST NAME"), x("FIRST NAME"), x("SECTION"), x("RECITATION"));
 our @userInfoFields = ("student_id", "user_id","last_name", "first_name", "section", "recitation");
@@ -52,11 +53,14 @@ sub initialize {
 	my $scoreSelected = $r->param('scoreSelected');
 	my $scoringFileName = $r->param('scoringFileName') || "${courseName}_totals";
 	$scoringFileName =~ s/\.csv\s*$//; $scoringFileName .='.csv';  # must end in .csv
+	my $scoringFileNameOK = (
+		$scoringFileName eq  WeBWorK::ContentGenerator::Instructor::FileManager::checkName($scoringFileName)
+	);
 	$self->{scoringFileName}=$scoringFileName;
 	
 	$self->{padFields}  = defined($r->param('padFields') ) ? 1 : 0; 
 	
-	if (defined $scoreSelected && @selected) {
+	if (defined $scoreSelected && @selected && $scoringFileNameOK) {
 
 		my @totals                 = ();
 		my $recordSingleSetScores  = $r->param('recordSingleSetScores');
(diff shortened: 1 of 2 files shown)


pstaabp:
No problem.  Is to be tested the same way as the previous PR?

mgage:
Yes.  But also check bad entries for file names on the scoring tool page.  I still have some last minute things to add to clean up the ui on the scoring tool page


Also I will do configuration stuff for pgproblen editor page unless you have already started it



On Jun 21, 2017 1:48 PM, Peter Staab <notifications@github.com> wrote:

No problem. Is to be tested the same way as the previous PR?

lib/WeBWorK/ContentGenerator/Instructor/ScoringDownload.pm
```diff
@@ -34,19 +35,21 @@ sub pre_header_initialize {
...
-	} else {
+	
+
+ # the parameter 'getFile" needs to be sanitized. (see bug #3793 )
+ # See checkName in FileManager.pm for a more complete sanitization.
+  	if ($authz->hasPermissions($user, "score_sets")) {
+ 		unless ( $file eq  WeBWorK::ContentGenerator::Instructor::FileManager::checkName($file) ) {  #
+ 			$self->addbadmessage("Your file name is not valid. ".
+ 			 "A file name cannot begin with a dot, it cannot be empty, it cannot contain a " .
+ 			 "directory path component and only the characters [^-_.a-zA-Z0-9 ]  are allowed."
```
dpvc: I think you want to remove the `^` from `[^-_.a-zA-Z0-9 ]` (that negates the character set to set for NOT those characters in the `checkName()` function).

mgage: Good catch.  I fixed the typo in the message.

lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
```diff
@@ -111,9 +115,17 @@ sub initialize {
...
-		$self->addbadmessage($r->maketext("You must select one or more sets for scoring"));
-	} 
+	} else {
+		if (!@selected) {  # nothing selected for scoring
+			$self->addbadmessage($r->maketext("You must select one or more sets for scoring!"));
+		}
+		if (!$scoringFileNameOK) { # fileName is not properly formed
+			$self->addbadmessage($r->maketext("Your file name is not valid! ")).
+		    $self->addbadmessage($r->maketext("A file name cannot begin with a dot, it cannot be empty, it cannot contain a " .
+				 "directory path component and only the characters ~[-_.a-zA-Z0-9 ~]  are allowed.")
```
dpvc: What does the `~` before the `[...]` mean?

dpvc: Oops, sorry, now I see it.  (D'oh!)

mgage:
I think this is ready now.  

To test:

1: Enter the following url as per bug #3793:

https://your.site.here/webwork2/coursename/instructor/scoringDownload?getFile=../../../../../etc/passwd

You should get an error complaining about the path components.
You can try various paths trying to reach any file outside the `scoring` directory.

2. On the Scoring Tools page try to enter an illegal file name such as ../templates/local/set1/problem1.pg

then press the action button.

You should get an error message and that fileName is incorrect. (previously
such a file would have been created, probably overwriting what was already there.)
Try simpler illegal file paths as well. 

There is also an error message if you forget to pick the sets to be graded.  These two error messages should work independently of each other.

lib/WeBWorK/ContentGenerator/Instructor/ScoringDownload.pm
Alex-Jordan: It looks like you removed the `^` from a similar line in `Scoring.pm`, but this instance that Davide commented on in `ScoringDownload.pm` is still here. Also it is not clear to me if escape tildes are needed here as they are used in the similar message in `Scoring.pm`.

pstaabp: @Alex-Jordan that's true.  This is just a message to the user however.  Perhaps it should say:
```
...only the characters a-z A-Z 0-9 - . and _  are allowed.
```
 to be clearly for someone who doesn't speak regular expressions.  
```

dpvc: Space is also allowed.

pstaabp: Yes.  I guess I missed that empty space in the RE.

mgage: I've updated the error message so that it is the same in both places.  I also mention space as a legal character explicitly instead of being cute and using the regular expression notation.

mgage:
I improved the error messages when the file name was bad and made the error messages the same in both the Scoring.pm and ScoringDownload.pm files.

Alex-Jordan:
OK, and just to be clear, I can merge this even though I haven't tested it myself? Not a good habit in general, but just getting me to run through a merge?

mgage:
Yes I think it's ok.  Look quickly at the changes in the files.  But they are very localized and not extensive


On Jun 24, 2017 3:06 PM, Alex Jordan <notifications@github.com> wrote:

OK, and just to be clear, I can merge this even though I haven't tested it myself? Not a good habit in general, but just getting me to run through a merge?

PR #775, opened 2017-06-21
Configure URLs for PGProblemEditor help buttons in default.conf

mgage:
This allows the urls for the PGProblemEditor help buttons to be configured in defaults.config (and overridden in localOverrides.conf or course.conf if desired.)

You should test that all of these buttons now reach correct targets for both ProblemEditor2 and ProblemEditor3.   You can also change the urls in defaults.conf and see if the buttons then fail.

I will comment out displaying ProblemEditor2 as a default in defaults.config unless there are objections.

Does anyone know the best way of inserting a value from the course configuration ($ce... )
into the javaScript file htdocs/js/apps/PGProblemEditor2/pgproblemeditor2.js?  I want to 
be able to set whether linewrapping is "true" or "false" in defaults.config and in localOverrides.conf.


pstaabp:
A reasonably easy way, but pretty hacky is to make a hidden input in the PGProblemEditor2.pm like:
```
<input id="cm-line-wrap" value="$ce->{...}" type="hidden">
```
using CGI::Input.

And then change the pgproblemeditor2.js to pulll the value from the hidden as

```
lineWrapping: $("#cm-line-wrap").val()
```

where this is using [jquery's val method](http://api.jquery.com/val/).

mgage:
I think that's what I'll try.  I think I'll also look at the way that Goeff used local storage for remembering whether to use MathJax or dvipng and see if I can use that method.  I'll also
put up a radio button/checkbox or pull down menu to control the change from line wrapping to no line wrapping.   This might take a little time since it is not the most urgent need, so if the current configuration of the help buttons is ok we should pull this now -- I'll submit a new PR for the line wrapping issue.

pstaabp:
As a follow-up, you should probably do:
```
lineWrapping: $("#cm-line-wrap").val() === "true"
```

which will return a native boolean true if the value is the string "true".  And the === checks equality without any type conversion, which can happen.  Unfortunately, any string that is not "true" will be false, like "TRUE".

pstaabp:
Also, do you want this PR to be only the links?  They look good and are easy to override in `localOverrides.conf`.

mgage:
I think this PR can be accepted as is.  Just changing the links to the buttons.

PR #778, opened 2017-06-22
Limit the length of the author KEY in the author table. 

mgage:
Limit the length of the author KEY in the author table. 

This will become more important as more people use utf8 in databases but
it doesn't interfere with any of the current databases that use only the latin1 encodings.


mgage:
This partially addresses the discussion in PR #745.  It resolves the problem for the moment for those who are already using utf8 or longer character sets in their database.   The larger matter of 
converting all of WeBWorK to use utf8 consistently -- and how to transfer existing installations over to using utf8 is something that will take a while to address fully.  I'm hoping that we can make that one focus of release 2.14 --targeted for release in late August or early September.

PR #783, opened 2017-06-29
Wiriseditoranswer

joaquimwiris:
We integrate the Wiris Editor for the student answer. Course manager can select this option in the course configuration. Answers written with the Editor are translated automatically to webwork assertion engine.


pstaabp:
If we're going to pull into ww2.13, then I'm switching the importing branch mainly to make sure there are no conflicts.

pstaabp:
On the handwriting mode, I'm getting the following javascript error (in the console):

```
com.wiris.system.Exception.$hxClasses.com.wiris.system.Exception
message: "Syntax error from line 7 column 7 to line 18 column 18.
```

To a user, it's not clear there is an error.  However, if I hit cancel, it seems to put the correct expression in the box and then is accepted. 

I tried on both chrome and safari (latest versions), so doesn't seem to be browser dependent.

mgage:
OK.  Thanks for checking Peter.  I wasn't seeing this  error before, but I didn't work with the handwriting very much.  I have found that the sticky answer feature stops working (the answer blank doesn't remember it's contents after it is submitted) so I've asked the WIRIS folks to work on that. I think we agreed that it was not necessary to pull this into release 2.13 -- we can pull it into develop once release 2.13 is complete and spend a little more time getting the WIRIS handwriting app to work just the way we want it to.

pstaabp:
Sounds good.  I wanted to make sure there were no conflicts with the current 2.13 branch, which I why I switched it.  We can easily switch it back and see if they can fix the errors and pull into develop later.

pstaabp:
After the last commit, I see no issues and this looks like a nice feature.  I tried it on my iPad with good success.   It would be nicer with some interface where I can see both the question and the input panel, but I don't think that's a deal breaker.  

I think this can be pulled into 2.13 and add another feature.

mgage:
Hi Peter,

I’m back from Oregon now.  We had a great stay at hotel on Driftwood Beach just north of 
Florence, OR.  Our flight from Portland was delayed sufficiently coming back that we missed our
connection in Chicago and got to spend most of the day yesterday in O’Hare.  Feeling much 
better this morning.  

If you have time lets consult about the last moves before merging 2.13 to master.

Take care,

Mike

mgage:
I've used this with my bamboo tablet attached to the macbook laptop. (My ipad2 is too old to allow easy drawing ).  It doesn't interfere with anything else.  I can still find examples where it has trouble with hand writing recognition.  For example on problem 2 of set 0 entering (1+sin(3))/(2+tan(4)) it would fail to recognize that sin was one word and would complain about s not being defined. (I think the wiris app had inserted some spaces.  )  Entering the word false would often create fa1se, etc.  I expect there is a learning curve for this app for both the handwriting recognition and the user.  

I think we should merge this into develop so that those who would like to experiment with it or to provide feedback to the WIRIS people will be able to use it easily.  It is easily turned off if it is in the way.

PR #788, opened 2017-07-15
Develop now

mgage:
These changes may already have been made or be identical to develop_uptds


mgage:
I think this was submitted by accident.  Most of this is code very stale. There are some changes to sendXMLRPC.pl which I'll incorporate into the pull request rearranging the client directory.

PR #789, opened 2017-07-15
Develop updated checkmodules

mgage:
Update check_modules.  Check that the network lines look ok.


mgage:
This pr is not needed.  Changes will be added when 2.13 is merged back in to develop

PR #793, opened 2017-07-17
Add template statements for overriding permission levels for PG info …

mgage:
…buttons and for library tags.

It turns out the version numbers were already updated. 

This replaces the pull request for reorganizing the client directory.  
It turns out that I had updated sendXMLRPC.pl in more than one branch on my computer
and these updates need to be combined and reconciled before I can commit. I don't want to do that 
just before a release.  I'll save this feature and reorganizing the client directory for 2.14 -- it is not urgently needed by most people.


pstaabp:
This looks fine.

PR #796, opened 2017-07-27
Hot fix sendmail notification bug

mgage:
This fixes a bug in which the notification that emails were successfully sent to a class was not properly sent to the instructor. 

Test by going the "Email" page and mail a message to yourself. Assuming you are an instructor you should receive two email messages:  the first is the message that you sent, and a second email saying that all of the emails had been successfully sent.

### DIFF
--- a/conf/localOverrides.conf.dist
+++ b/conf/localOverrides.conf.dist
@@ -412,7 +412,7 @@ $showeditors{simplepgeditor}   = 0;
 ## Key database, see the   Session Managment section
 ## of defaults.config.dist
 
-## For session management using the key databasse table, uncomment the following line,
+## For session management using the key database table, uncomment the following line,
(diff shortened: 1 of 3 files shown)


pstaabp:
I finally got this working.  I was having trouble emailing anyone outside of our fitchburgstate.edu domain and had set up a student with my gmail account.  Again, I think this will be the problems that arise, but they are independent of the Email::Sender fix.

mgage:
Go ahead and pull it.  It was a fairly straightforward bug and bugfix.  The restrictions imposed by the smtp on who can be emailed are not something we can do much about.

pstaabp:
Do you want to have this merge into develop as well?

mgage:
I’ll take care of that.  Thanks for the help.

Take care,

Mike

PR #797, opened 2017-07-27
Hot fix sendmail bug devel

mgage:


mgage:
This seems ok, despite being a hot_fix starting from master.  The only unexpected change is in the Tasks.pm file and that seems ok.  Might be a fix that was in danger of being lost.

PR #799, opened 2017-08-02
add maketext calls in lib/WeBWorK/ContentGenerator/Feedback.pm

heiderich:


Alex-Jordan:
I don't fully understand how maketext works, but I think I understand its purpose for localization.

Aside from checking for typos, is there anything else I should look at for this and #806?  For instance, should I be looking to see if these exact same strings already have a maketext around them somewhere else?

heiderich:
Maketext is used to translate strings in different languages. Currently the strings that are translated are mostly (parts) of English sentences, or simply English words. Maketext would look up in the .po file corresponding to the language a course uses to translate these strings.

I a string has occurs with maketext around it in two separate places, it will be translated in the same waty in both places. It can happen that a given string may need different translations depending on the context it occurs in. For full sentences this is usually not an issue, but for single words or parts of sentences it could be. For the moment this problem is not dealt with in WeBWorK very well (in my opinion). I am not an expert in localization, but I guess one would sometimes need to change the strings to something that specifies the context. For instance a string "foo" might need to be changed to maketext("foo_CONTEXT1") in one context and to maketext("foo_CONTEXT2") in another context. This will allow different translations in these contexts. But then you would also need an entry in en.po that makes sure that both are displayed as "foo" if "en" is chosen as a language in WeBWorK. Otherwise I believe they will be displayed as "foo_CONTEXT1"  and "foo_CONTEXT2".

In the case of this PR and #806 I see no reason to specify the context.

PR #798, opened 2017-07-31
Develop uft8 ver2

mgage:
This is experimental -- do not merge it yet.  This is a summary of the the utf8 related changes I have made to get international characters to work. It  has been successful to some extent and no longer fails on PGML material.  The companion request for pg is #319.

It may well fail other test cases in which case I encourage you to report them to this pull request. 

Here is a pg problem that can be used for initial testing.  All of the bulgarian characters should
now appear properly.

```
##DESCRIPTION
##  Test problem 
##ENDDESCRIPTION


## DBsubject(WeBWorK)
## Date(7/30/2017)
## TitleText1('')
## AuthorText1('')
## EditionText1('')
## Section1('')
## Problem1('')
## KEYWORDS('test')

########################################################################

DOCUMENT();      

loadMacros(
   "PGstandard.pl",     # Standard macros for PG language
   "MathObjects.pl",
   "PGML.pl",
   #"source.pl",        # allows code to be displayed on certain sites.
   #"PGcourse.pl",      # Customization file for the course
);

# Print problem number and point value (weight) for the problem



TEXT($PAR, "bulgarian printed from within a TEXT() function:   ьеижъанч",$PAR);

DEBUG_MESSAGE("this is bulgarian in the debug messages  еияащнвЧ -- it works");

BEGIN_TEXT
This is bulgarian printed from within BEGIN_TEXT/END_TEXT block
$PAR $HR
ьеижъан   it works
$HR $PAR
END_TEXT

DEBUG_MESSAGE("this is bulgarian in the debug messages  еияащнвЧ -- it works");

BEGIN_PGML_SOLUTION
---
This is Bulgarian printed from within a pgml solution:  

ьеижъанч

It fails.
---
END_PGML_SOLUTION


BEGIN_PGML
---
This is bulgarian printed from within BEGIN_PGML/END_PGML block

pgml text ьеижъан

It also fails. 

---
And after PGML mode the TEXT doesn't work anymore either. 

END_PGML


TEXT($PAR, "from within a TEXT() function.  bulgarian:   ьеижъанч",$PAR);

DEBUG_MESSAGE("this is bulgarian in the debug messages  еияащнвЧ and it no longer works.");

BEGIN_TEXT
This is bulgarian printed from within BEGIN_TEXT/END_TEXT block
$PAR $HR
ьеижъан   it still works
$HR $PAR
END_TEXT
ENDDOCUMENT();        
```

Issue #781, opened 2017-06-26
Localization in .js files


heiderich:
The line that informs the instructor about the number of problems found in the SetMaker is constructed in two places:

In htdocs/js/apps/SetMaker/setmaker.js and in lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm.

The strings in the latter are (partially) localizable thanks to the maketext calls.

How can we localize the strings in the .js file? In any case it would be good if these strings would only be stored in one place and not in two places.

Furthermore it would be good if the formating would be done in the same way in both places, which is currently not the case. In the .js file the special case of 1 found problem is treated specially, while in the .pm file the case of no found problem is treated specially.

pstaabp:
I'm working with webwork3 and have on my todo to think about localization in js files, because webwork3 is nearly all js on the user interface side.  What my plan was a couple of years ago when I first thought about this is:
1. put all user-facing strings in some templating files.  I've generally adopted this anyway because it is the correct way do do UI. 
2. If the language is switched, all of the templates will be needed to be translated once (or at code-update/server reboot) and create new template files.   This will have the routine to call makeText, so we're using all of the work currently being done. 
3. When any files are loaded, pull in the appropriate language-version templates.  

I haven't implemented any of this yet, but that is my plan for webwork3.  I don't know when I will get to that, but will probably do some other thinking about things before implementation.

taniwallach:
As I remarked in https://github.com/openwebwork/webwork2/issues/914 the CCMD fork has some work on localization for JavaScript code.

In any case, I don't think this is mission critical to the extent that it should defer releasing WW 2.15 with the current level of support for UTF-8.

Alex-Jordan:
I don't know if this would be helpful, but in PreTeXt we put localized versions of certain things as data attributes on some HTML element up in the head. Then for any js that needs to say something to the user, get the text from the appropriate data attribute. In this case, one or two data attributes would need to carry the localized version of `There is 1 matching WeBWorK problem` and also the localized version of `There are ${num} matching WeBWorK problems.`

drgrice1:
A much more versatile approach would be to use a JavaScript localization library for this that has interpolation capabilities (like `i18next`).  In fact, I already have this mostly implemented.  It does require quite the change though because it requires switching to using JavaScript modules.  I need to finish that off.  I have had it sitting around since March.

PR #800, opened 2017-08-04
Develop uft8 ver3

heiderich:
This pull request is a follow up to pull request #798 by @mgage.

This is experimental -- do not merge it yet.

Remaining problems (this list is not exhaustive):

As mentioned by @jutrembBDEB:

> I added a extra library button in localOverrides file, it's call "Banque de problèmes libres". the characters doesn't show correctly in the SetMaker page.

Entries from exiting MySQL tables do not display correctly. Probably they need to be reencoded.
@jutrembBDEB: How did you convert your MySQL database?

As observed by @heiderich:

- Non-ASCII characters in the metadata of problems (and the Taxonomy2 file) cause problems in the library browser. Entries are shown correctly, but once selected no problems are found.


jutrembBDEB:
I also found a solution to my problem about the extra library button. It's the exact same solution I did with Login.pm.

In the lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm file, you add the line `use Encode;`

and you replace the line 406 : `my $name = ($lib eq '')? $r->maketext('Local') : $problib{$lib};` by
`my $name = ($lib eq '')? $r->maketext('Local') : Encode::decode_utf8($problib{$lib});`

and also the line 837 : `$libs .= ' '. CGI::submit(-name=>"browse_$lib", -value=>$problib{$lib},` by

`$libs .= ' '. CGI::submit(-name=>"browse_$lib", -value=>Encode::decode_utf8($problib{$lib}),`

 @heiderich , can you make the change to your branch. I don't want to create another pullrequest just for that. Thank you!

jutrembBDEB:
I didn't convert my SQL database yet, i just change the charset to utf8 for new created courses, but the existing table have not been converted. Maybe I found a solution, but I don't want to test it yet until I showed it to someone.

heiderich:
@jutrembBDEB Let me just mention that @goehle posted an SQL command in his original pull request #712 to convert the database to utf8. I remember that I used a variant of it. The main difference was that I used utf8mb4 instead of utf8 for full utf-8 support (I think the names are a little misleading). I would also propose utf8mb4 as a default for WeBWorK. For more detail see
https://stackoverflow.com/questions/30074492/what-is-the-difference-between-utf8mb4-and-utf8-charsets-in-mysql
and
https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html

I remember that I needed some more SQL commands to change encodings on different levels, but I do not recall them right now.

Use the mentioned SQL commands at your own risk and better create a backup beforehand. Please let us know about your experiences.

heiderich:
I made the following observation concerning the problem with non-ASCII characters in the metadata of problems / the Taxonomy2 file:

- With the current version of this PR the following happens:

  When selecting a section containing a non-ASCII character the correct number of problems found would be shown. Once you click on "View problems" the page reloads and the string "There are no matching WeBWorK problems" is shown.

- When adding the line

  `mysql_enable_utf8 => 1,`

  to the options of the DBI->connect call (around line 112) in the file `lib/WeBWorK/Utils/ListingDB.pm`, then the opposite happens:

  When selecting a section as above, it is shown that there are no problems, but if you click on "View problems" the problems are actually shown (together with the right number of problems). 

Note that the two strings "The are XXX/no matching WeBWorK problems" come from different places. The first one (shown before clicking on "View problems") is defined in `htdocs/js/apps/SetMaker/setmaker.js`. The second one (shown after clicking on "View problems") is defined in `lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm`.

Another issue here is the translation of this string in the .js file. I opened the issue #781 for this.

heiderich:
I still experience a problem when I try to create a homework set from the library browser with a name that contains non-ASCII characters. The error is:

```
Problem creating set ü
argument 1 contains invalid characters in 'set_id field: 'ü' (valid characters are [0-9]) at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm line 1483. at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm line 1483. 
```

mgage:
Florian:

Thanks for the report -- this message is from webwork -- in
lib/WeBWorK/DB.pm subroutine validateKeyfieldValue() around 2316.  We'll
need to modify the allowed characters (or perhaps make it a configuration choice) once
we are using mysql with extended characters.

heiderich:
I think that commit d24b860 and ac90685 should fix the problem in the library browser that no problems were shown when subject, chapter or section containe non-ASCII characters. It seems that this was caused by the following: Some strings were internally not stored in utf8 in perl. This probably causes problems when they were passed to MySQL (which I think expected utf8, at least after commit  d24b860). Applying utf8::upgrade to these strings seems to solve this problem.

heiderich:
@mgage: I think the problem with non-ASCII characters in names of homework sets I mentioned above is not necessarily related to this pull request and I think it could be addressed later on.

mgage:
Here we go!!!!

mgage:
See also PR #798 for comments

PR #802, opened 2017-08-09
theme based on math4, with right to left text direction and 

heiderich:
lang="ar" attribute for <html>

Purpose: Arabic environments

The pull request https://github.com/openwebwork/pg/pull/323 for pg provides BLTR and ELTR for left-to-right environments within a global right-to-left environment (math is usually still written from left to right).

### DIFF
--- a/htdocs/themes/math4-ar/README
+++ b/htdocs/themes/math4-ar/README
@@ -1,0 +1,18 @@
+This folder contains the files necessary for the math4 theme.  These files are
+tracked by git and any changes made to them will be overwritten when you 
+upgrade.  The two exceptions are math4-overrides.css and math4-overrides.js.  
+
+These files do not need to be present, but if they are they will be included 
+in system.conf and will override the math4.css and math4.js theme.  They can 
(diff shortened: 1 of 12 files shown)


mgage:
This adds math4-ar to the list of possible templates.  If math4-ar is not selected there is no effect on existing problems.  The most obvious change of selecting math4-ar is right justification for all of the pages. Since we don't have arabic translations of most webwork2 pages yet it's hard to tell whether this is the desired effect.  

We will also need to merge the PR related to utf8 before we will be able to view the arabic script. These files can serve as good tests for the utf8 changes as we start to review and merge them.

@Alex-Jordan has pointed out in #323 (left to right environments) that in the long run cloning math4 might require more work as each change in math4 would need to be duplicated. This is true, but in the short term it's useful to have this file separate so that it can be experimented with and updated without fear of affecting math4 and existing courses.   Once it's determined what a math4_ar (or math4_RL) template should look like we can try to overlay the css of math4 to accommodate right to left languages.

In the meantime this these files do no harm and give us something to experiment with.

PR #804, opened 2017-08-29
Add LTI role of Faculty to support new version of D2L Brightspace

dlglin:
Not sure if it's worth backporting this to master.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -208,6 +208,7 @@ $LTIMassUpdateInterval = 86400; #in seconds
 			   "Designer" => "professor",
 			   "instructor" => "professor",
 			   "Instructor" => "professor",
+			   "Faculty" => "professor",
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
If I understand right, an instructor at a school with a newer version of D2L might be classified in the "Faculty" role. And then suppose they follow an External Learning Tool link into a WeBWorK course where they do not already have an account. Then, prior to this commit, WeBWorK does not know what permission level to give them as it creates their account. Do I have that right?

We use D2L at my school and just had an upgrade over summer. It doesn't appear that anyone is having trouble, so maybe the "Faculty" role is not being used here. But then again, when we initialize WeBWorK courses, we add the "professor" at that point. So maybe there has been no trouble simply because they all already had accounts in the WeBWorK courses. 

Anyway, it seems like there's nothing to test here. Mike, can I just merge a PR without testing anything if it looks so straightforward as this one?

PR #805, opened 2017-09-02
add support for the UTF-8 euro symbol in pg files

heiderich:
This makes it possible to use the UTF-8 symbol U+20AC (€) directly in pg files.


mgage:
we need a modification for the 1 column as well

mgage:
Should \euro render in math mode as well as in straight text?  It doesn't seem to.  

In addition the associative directive in contextCurrency.pl doesn't seem to making the values display properly (the euro is always on the left) .  
I am using pdflatex as a tex renderer and the standard develop package with no extra PR added except for this one.

Here is what 
I'm using as a test problem.  Implementing the euro symbol may be trickier than I thought.


########################################################################

DOCUMENT();      

loadMacros(
   "PGstandard.pl",     # Standard macros for PG language
   "MathObjects.pl",
   "contextCurrency.pl",
   #"source.pl",        # allows code to be displayed on certain sites.
   #"PGcourse.pl",      # Customization file for the course
);

# Print problem number and point value (weight) for the problem
TEXT(beginproblem());

# Show which answers are correct and which ones are incorrect
$showPartialCorrectAnswers = 1;

##############################################################
#
#  Setup
#
#
Context("Currency");
Context()->currency->set(symbol=>'euro');
Context()->currency->setSymbol(euro=>{associativity=>"left"});
$amount = Compute(" euro 4");
Context()->currency->setSymbol(euro=>{associativity=>"right"});
# accepts '12 euro'
$amount2 = Compute("4 euro");
##############################################################
#
#  Text
#
#

Context()->texStrings;
BEGIN_TEXT
The euro symbol in mathmode  \(\euro \) and outside of mathmode \euro and using 
a PG value $EURO and a mac symbol  â¬  $PAR
Enter $amount. \{ans_rule()\}
$PAR
Enter $amount2.  \{ans_rule() \}
END_TEXT
Context()->normalStrings;

##############################################################
#
#  Answers
#
#

ANS($amount->cmp);
ANS($amount2->cmp);
# relative tolerance --3.1412 is incorrect but 3.1413 is correct
# default tolerance is .01 or one percent.

heiderich:
> Should \euro render in math mode as well as in straight text? It doesn't seem to.

The purpose of this pull request was to allow to use the UTF-8 symbol € directly in .pg files (instead of `\euro` or something else). Sometimes one may just want to use the euro symbol within the description of a problem and it may not necessarily appear in the answer(s). I think the HTML output does not require any alteration. I tested this PR with Firefox 57.0.1 and Chrome Version 66.0.3359.181. Both render the euro symbol correctly using this PR (rebased onto PR #800). Probably this PR does not make much sense without #800. I should have mentioned this.

Here is a test problem that illustrates its use.

```
DOCUMENT();      

loadMacros(
   "PGstandard.pl",     # Standard macros for PG language
   "MathObjects.pl",
);

TEXT(beginproblem());
$showPartialCorrectAnswers = 1;

$amount = Compute("4");

Context()->texStrings;

BEGIN_TEXT
The euro symbol in mathmode  \( € \) and outside of mathmode €.
Enter $amount €. \{ans_rule()\} € 
END_TEXT

Context()->normalStrings;

ANS($amount->cmp);

ENDDOCUMENT();

```
I did not use `contextCurrency.pl` in my tests. Introducing the euro symbol there would probably be another issue. This was not my aim here.

I added support in the oneColumn hardcopy theme in commit 8e7a07a.

PR #807, opened 2017-09-29
PreTeXt support

Alex-Jordan:
This is the companion PR for [a PR to pg](https://github.com/openwebwork/pg/pull/327). That PR has a full description. 

Beyond what is there, here we introduce a "ptx" format for an XMLRPC call. PreTeXt would use this simplified format to easily access WeBWorK-generated PTX code.


mgage:
I've been able to verify that this does no harm to the standard use of webwork.  
I haven't done extensive testing with PreTeXt but if there are problems or if improvements
need to be made that can be done in later PR.

PR #813, opened 2017-10-05
Allow student_id to be retrieved from a (configurable) LTI parameter

dsteinmo:
The default LTI parameter is custom_student_id, otherwise it can be set in authen_LTI.conf.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -86,6 +86,18 @@ $preferred_source_of_username = "lis_person_contact_email_primary";
 #$strip_address_from_email = 1;
 
 ################################################################################
+# LTI Preferred source of Student Id
+################################################################################
+
+# If preferred_source_of_student_id is not set, WeBWorK will look in the LTI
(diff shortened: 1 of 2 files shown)


mgage:
For the test with Canvas used: $preferred_source_of_student_id = "ext_d2l_orgdefinedid"
as the name of the LTI parameter describing the student and set it to "mgage".
This is configured in canvas as a custom parameter -- I wasn't sure how to retrieve the student name from canvas.  

The new parameter  is transmitted as custom_ext_d2l_orgdefinedid
New parameters are added to the debug list:
student id:
preferred_source_of_student_id:
custom_ext_d2l_orgdefinedid: 

Turn on $LMSManageUserData=1;

This seems to do the right thing.  It updates the student_id properly.  

The interaction with canvas and LTI seems a little flakey (or perhaps I don't remember exactly
what it is supposed to do) but this modification doesn't affect the behavior other than in the 
way it's supposed to.

LTI integration deserves another thorough review. (sigh)

PR #815, opened 2017-10-17
Wiriseditoranswer

joaquimwiris:
Corrected some bugs in Wiris Editor integration for WeBWorK. Also, disable syntax checking and prevent submit before recognition in Hand.


Alex-Jordan:
Just checked this out, and am doing some testing. I played with all the palette tools I see (from Basic, Functions, and Intervals tabs) and it all works except for some comments below. I did not do anything close to exhaustive testing with the handwriting tool, but I did find one issue noted below.

- Suppose a problem has two answer blanks; I click the radical symbol to open a popup window; for whatever reason, then I click the radical symbol on the _other_ answer blank. A popup window appears, but nothing loads inside it. This scenario could happen if a user is in a Wiris window and then goes back to the main browser window for some reason (making the popup hidden beneath the browser window) and doesn't remember that a Wiris window is already open.
- There are some accessibility issues with Wiris. But a screen reader user can be quite effective just typing in expressions in the traditional way, using correct order of operations syntax. A palette and handwriting tool seems like it would not be something a visually or mobility disabled person would want to use in the first place. I'll ask some people with more expertise to comment. The one thing I'm more worried about is if more-abled users get an advantage with entering things like nth radicals, where the plain keyboard syntax is not always known to the user.
- I recommend adding an option to User Settings so that an individual can turn Wiris off. Unless the accessibility issues are addressed, it's bad if a blind user navigates into the popup window. But with a User Settings switch, that user could turn this off permanently for them, the same was they can switch between MathJax and images.
- Once I make something in the palette that has nontrivial typesetting and click OK, there is an image in the answer field. A feature I would like to have is the ability to clear this away and type something different directly. (As it is now, I must reopen the palette to clear things away.) From the UX perspective, it would be simplest to just be able to click in the answer field and start typing to do this.
- If I use the palette to make \sqrt[3]{x} and submit an answer, I can see that it is actually passing `x^(1/3)`. From my perspective, teaching at a level where the relationship between x^(1/3) and \sqrt[3]{x} is still something to be taught, it's bad to use one of them symbolically but really pass the other.  And there is an actual difference: in WW, `x^(1/3)` is undefined for negative x, but `root(3,x)` and friends handles negative x. But then again `root(3,x)` is only recognized if `parserRoot.pl` has been loaded. My preference would be that if the palette is used to make \sqrt[3]{x}, that it passes `root(3,x)`. I recognize the down side: people would make \sqrt[3]{x} in problems where `parserRoot.pl` has not been loaded. But those problems were never planning to allow literal \sqrt[3]{x} as an answer in the first place. And _some_ of those problems (like ones at lower levels) may be intentionally assuming that  \sqrt[3]{x} should not be entered, and so then the palette here would be cheating and letting them use it anyway.
- The last time I tested, there was a matrix palette tool. I can't find it now. (I see Basic, Functions, and Intervals tabs.) When entering a matrix into a single answer blank, WeBWorK syntax is like `[[1,2],[3,4]]`. When I make a matrix using the handwriting tool, it does make one, and I see a matrix image in the answer blank. But then when I hit submit, it is as if there had been nothing in the answer blank, and the image disappears from the reloaded version of the page. If a matrix template is intentionally being withheld, could we make it so that even with the handwriting tool, the student won't be able to see a matrix in the answer blank before hitting Submit? It could be an issue if they think they are submitting a correct matrix answer, but it just disappears.
- There is a palette tool for things like \log_{2}(x). This ends up passing `log(x)/log(2)` for answer checking. This raises an issue similar to the one I raised for things like cube root, since I have to teach the relationship that \log_{2}(x) equals \log(x)/\log(2). And this palette tool would break certain problems that rely on there being no way to type \log_{2}(x).  My short term preference would be to remove this one from the palette. But my long term preference is to have something like `parserLogarithm.pl` that would recognize things like `log(b,x)`. And then have this palette tool pass that syntax. For such a `parserLogarithm.pl`, we would have to work out how to make it not clash with the regular `log(x)`.

Overall, this is an amazing tool. I will help more with getting this finalized, if you can give me guidance on anything I might do for it. Also, I don't yet know what the plan is for licensing Wiris in WeBWorK. Has that been worked out?

manuelwiris:
Hi @Alex-Jordan 

Regarding your issues: 

> Suppose a problem has two answer blanks; I click the radical symbol to open a popup window; for whatever reason, then I click the radical symbol on the other answer blank. A popup window appears, but nothing loads inside it. This scenario could happen if a user is in a Wiris window and then goes back to the main browser window for some reason (making the popup hidden beneath the browser window) and doesn't remember that a Wiris window is already open.

We add this bug to our tracker.

> There are some accessibility issues with WIRIS. But a screen reader user can be quite effective just typing in expressions in the traditional way, using correct order of operations syntax. A palette and handwriting tool seems like it would not be something a visually or mobility disabled person would want to use in the first place. I'll ask some people with more expertise to comment. The one thing I'm more worried about is if more-abled users get an advantage with entering things like nth radicals, where the plain keyboard syntax is not always known to the user.

> I recommend adding an option to User Settings so that an individual can turn Wiris off. Unless the accessibility issues are addressed, it's bad if a blind user navigates into the popup window. But with a User Settings switch, that user could turn this off permanently for them, the same was they can switch between MathJax and images.

WIRIS EDITOR is prepared for persons with visual disability and is compatible with screen readers. Please read the following documents: http://www.wiris.com/en/editor/docs/keyboardonly and http://www.wiris.com/en/editor/docs/accessibility

> Once I make something in the palette that has nontrivial typesetting and click OK, there is an image in the answer field. A feature I would like to have is the ability to clear this away and type something different directly. (As it is now, I must reopen the palette to clear things away.) From the UX perspective, it would be simplest to just be able to click in the answer field and start typing to do this.

We add this bug to our tracker.

> If I use the palette to make \sqrt[3]{x} and submit an answer, I can see that it is actually passing x^(1/3). From my perspective, teaching at a level where the relationship between x^(1/3) and \sqrt[3]{x} is still something to be taught, it's bad to use one of them symbolically but really pass the other. And there is an actual difference: in WW, x^(1/3) is undefined for negative x, but root(3,x) and friends handles negative x. But then again root(3,x) is only recognized if parserRoot.pl has been loaded. My preference would be that if the palette is used to make \sqrt[3]{x}, that it passes root(3,x). I recognize the down side: people would make \sqrt[3]{x} in problems where parserRoot.pl has not been loaded. But those problems were never planning to allow literal \sqrt[3]{x} as an answer in the first place. And some of those problems (like ones at lower levels) may be intentionally assuming that \sqrt[3]{x} should not be entered, and so then the palette here would be cheating and letting them use it anyway.

We add this bug to our tracker.

> The last time I tested, there was a matrix palette tool. I can't find it now. (I see Basic, Functions, and Intervals tabs.) When entering a matrix into a single answer blank, WeBWorK syntax is like [[1,2],[3,4]]. When I make a matrix using the handwriting tool, it does make one, and I see a matrix image in the answer blank. But then when I hit submit, it is as if there had been nothing in the answer blank, and the image disappears from the reloaded version of the page. If a matrix template is intentionally being withheld, could we make it so that even with the handwriting tool, the student won't be able to see a matrix in the answer blank before hitting Submit? It could be an issue if they think they are submitting a correct matrix answer, but it just disappears.

We add this bug to our tracker. We expect have it solved in our February release.

> There is a palette tool for things like \log_{2}(x). This ends up passing log(x)/log(2) for answer checking. This raises an issue similar to the one I raised for things like cube root, since I have to teach the relationship that \log_{2}(x) equals \log(x)/\log(2). And this palette tool would break certain problems that rely on there being no way to type \log_{2}(x). My short term preference would be to remove this one from the palette. But my long term preference is to have something like parserLogarithm.pl that would recognize things like log(b,x). And then have this palette tool pass that syntax. For such a parserLogarithm.pl, we would have to work out how to make it not clash with the regular log(x).

We can remove this option from the palette until a parserLogarithm.pl is finished. Do you have any issue in your tracker? It would very useful to rollback the changes,

Regarding the license: I'm not the responsible about the license but , right now, we are having conversations about the license terms. As soon as we know something I will back to you.

In the other hand: are all this issues blocking to accept the pull request? It would be useful figure out which of the issues are blocking the pull request. It will allow us prioritize them into our tracker.

joaquimwiris:
I just resolved the issue with matrices. We added the matrix icon to the toolbar, and made sure that they can be translated from MathML to webwork syntax and the other way around.

mgage:
Hi @joaquimwiris 

This appears to work fine.  Do you want this pull request included in release 2.14?  That would mean the choice of configuring to use wiris would be available to all instructors starting this summer.  You would start getting some traffic from many WeBWorK users.  Are you ready for the logistics of handling that?  If you want to wait we can put it in develop which will become release 2.15 towards the end of the summer -- that means that only the adventurous early adopters would be using wiris until probably the semester break in January 2019.    Send me an email gage@math.rochester.edu and we can discuss logistics.  The WeBWorK Project has a bank account now. :-)

mgage:
@joaquimwiris I cannot view the little checkmark symbol that opens the mathequation editor pallet.  Originally I could see  the checkmark on the Chrome browser but after a shift reload that stopped
working as well.  

If I click on a space near the blank I get the error message
no display module found for path:

/gage_test/CollegeAlgebraTests/1/quizzes/resources/popup.html/   

That's the wrong path -- 
You probably want something like
/webwork2_files/js/apps/WirisEditor/popup.html

mgage:
One more thing -- could you pull the localOverrides.conf.dist file into your local version -- there have been some changes made (mostly just comments)

joaquimwiris:
Hi, @mgage I solved the problem with the icons. I had updated the Wiris Quizzes javascript to the last version, but messed up the configuration for WebWork. It is fixed now.

mgage:
Thanks Manuel.  It works now.  I still have issues making this work on my Fire 8 tablet (the only new tablet I have easily available).  Sometimes when I press ok it doesn't enter the result in the answer blank. Almost always when I try to reopen the  answer blank in the popup window I get a blank screen in the popup window.

I'm using https://demo.webwork.rochester.edu/webwork2/2014_gage_demo/CollegeAlgebra for testing.  It has guest login.

mgage:
@joaquimwiris , @manuelwiris could you resubmit the wiriseditoranswer PR again this time to the branch  rel-ww2.14?      I accidentally merged it into the develop branch.  I reverted that merge but now I can't create a new PR to branch rel-ww2.14 using github tools.   Thanks.

mgage:
Hi @joaquimwiris,

I tested it and it works pretty well in general.  it doesn’t work on my fire 8 tablet — after the first time the popup is used 
the popup comes up blank whenever you try to enter a new answer.  You need to login to the course again? in order to fix it.

Could you resubmit  wirisEditorAnswer branch to rel-ww2.14 instead of to develop?  I mistakenly pulled it into develop and now I can’t redirect
the PR easily from github.

Thanks.

Take care,

Mike

PR #816, opened 2017-10-25
Fix keyword searching

jwj61:
Fix bug in library browser where the user enters keywords for their search.  Previously, it was always returning nothing.

### DIFF
--- a/lib/WeBWorK/Utils/ListingDB.pm
+++ b/lib/WeBWorK/Utils/ListingDB.pm
@@ -182,8 +182,8 @@ sub makeKeywordWhere {
 	my $kwstring = shift;
 	my @kwlist = keywordCleaner($kwstring);
 #	@kwlist = map { "kw.keyword = \"$_\"" } @kwlist;
-	@kwlist = map { "kw.keyword = ? " } @kwlist;
-	my $where = join(" OR ", @kwlist);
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Thanks John. Should we also do a hotfix pull request into master? And if so, how does that affect the git flow of the fix? I'm guessing the fix would need to branch from master, and then after merging into master, we merge master into develop?

dpvc:
@Alex-Jordan, you are right, a hot fix should be a branch from master merged into master.  But you don' t pull master into develop (that is not allowed by git-flow).  Instead, you merge the hot fix branch into both master *and* develop.  The merges to master are supposed to be one way (only into master).

jwj61:
So, I should have a single pull request to master?

dpvc:
You should rebase your branch to be off of master and make a pull request to master.  When that is approved, make another pull request to develop to get your patch merged there as well.  You should not have to wait for additional review for the one to develop, once it has been approved for master.

dpvc:
Then I think there needs to be a new tag for 2.13.1 or whatever.

mgage:
Arnie and Alex say this is ok, and it seems right so here goes.

Alex-Jordan:
Mike, you merged this into develop. What do we do with the corresponding #817 into master?

PR #817, opened 2017-10-25
Fix keyword searching

jwj61:
Keyword search fix, based off of master.


Alex-Jordan:
Tested on my little laptop installation and this works there. Haven't yet tested on my production server. I started too, but then I saw that I'm on some 2.13 testing branch instead of the final 2.13 master and there are more changes than I may want to mess with mid term.

apizer:
This worked on my WW 2.13 master branch installation.

mgage:
this hotfix matches pull requests #816

PR #818, opened 2017-11-04
Modernize 'isText' subroutine

dsteinmo:
Check for control characters instead of non-ASCII characters to determine if a file is a text file or a binary file.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -1274,7 +1274,7 @@ sub showHTML {
 #
 sub isText {
 	my $string = shift;
-	return $string !~ m/[\x00-\x1F\x7F]{2}/;
+	return $text !~ m/(?=[^\r\n\t])[\x00-\x1F\x7F]{2}/;
 }
 
 ##################################################
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -1274,7 +1274,7 @@ sub showHTML {
 #
 sub isText {
 	my $string = shift;
-	return $text !~ m/(?=[^\r\n\t])[\x00-\x1F\x7F]{2}/;
+	return $string !~ m/(?=[^\r\n\t])[\x00-\x1F\x7F]{2}/;
 }
 
 ##################################################
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -1274,7 +1274,7 @@ sub showHTML {
 #
 sub isText {
 	my $string = shift;
-	return $string !~ m/(?=[^\r\n\t])[\x00-\x1F\x7F]{2}/;
+	return $string !~ m/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]{2}/;
 }
 
 ##################################################
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -1274,7 +1274,7 @@ sub showHTML {
 #
 sub isText {
 	my $string = shift;
-	return $string !~ m/[^\s\x20-\x7E]{4}/;
+	return $string !~ m/[\x00-\x1F\x7F]{2}/;
 }
 
 ##################################################
--- a/t/test_fileManager.pl
+++ b/t/test_fileManager.pl
@@ -8,7 +8,7 @@ BEGIN {
         eval "use lib '$WebworkBase::RootWebwork2Dir/lib'"; die $@ if $@;
 }
 
-use Test::More tests => 12;
+use Test::More tests => 14;
 use WeBWorK::ContentGenerator::Instructor::FileManager qw/isText/;
 
 ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abcdghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"), 'ordinary string is text';
@@ -23,3 +23,5 @@ ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abc\r\n\r\ndef\r\
 ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("                 "), 'string of all spaces is text';
 ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("      \r\n\t\r\n\t        "), 'string of all spaces, tabs, CRs and newlines is text';
 ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abcdef\t\t\t\tasdfjka"), 'string with multiple consecutive tabs is text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abc\x00\ndef"), 'control character before newline surrounded by ASCII chars is text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abc\n\x00def"), 'newline before control character surrounded by ASCII chars is text';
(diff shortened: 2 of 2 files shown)


Alex-Jordan:
Do I understand right? It's currently looking for 4 characters in a row that are not printable ASCII, and the commit makes it explicitly look for 2 adjacent ASCII control characters instead?

dsteinmo:
Hi. Yes, it's currently looking for 4 characters in a row that are not printable ascii. The problem (that I see) with that is it will match printable unicode characters as well. 

I switched it to match 2 adjacent control characters so that it wouldn't match the unicode characters. But, come to think of it, we probably need to be a bit more careful than this because carriage returns and new-lines fall into the category of ascii control characters, so a file with 2 or more blank lines in a row would be considered binary by this logic.  

I'll probably fix this by capturing carriage return and new-line in the regex, and being careful that there aren't other important control characters that could be in a text file. I could also write up some unit tests for 'isText' so we can add all the pathological cases and so on.

dsteinmo:
This looks to be good now. The unit tests are all passing and testing through the UI seems to detect text files as text and .tgz archives as binary. Let me know if you have any thoughts or suggestions.

dsteinmo:
@mgage I'd be happy to help with reviewing in my spare time if you're looking for help. Let me know what is involved in the training process. I can be reached at dsteinmoeller@wlu.ca

mgage:
Hi Derek,

Thanks for your offer. Sorry for the delay in getting back to you. My classes are concentrated
on Monday—Wednesday and it takes me a day to recover. :-)

As usual we can use all the help we can get.  One of our release managers has had 
to back off a bit to concentrate on academic work so at present I’m the only active maintainer.

I’ve added you as a reviewer and invited you to join the openwebwork organization so that your name
will appear on lists of possible reviewers, maintainers, etc.

We do not have an established training procedure but I believe we need one 
so it looks like you and I will be guinea pigs in  getting this process started.  Let me know if 
I’m explaining too much or too little. 

From your earlier correspondence I’m assuming that you know your way around git and github at least to some 
extent. (I can get by but I’m not a power user by any means.)  That will help a lot.  
Do you have a develop machine that you can safely use to download modifications to webwork and test the new version?

Our “flow” for handling pull requests is approximately 

https://github.com/openwebwork/webwork2/wiki <https://github.com/openwebwork/webwork2/wiki>
and https://github.com/openwebwork/webwork2/wiki/Merging-Pull-Requests <https://github.com/openwebwork/webwork2/wiki/Merging-Pull-Requests>

These instructions have not been widely advertised and aren’t always followed to the letter, but it seems a good start.
We will vow to do better. :-)

One of the few things that we have been pretty consistent with is that the maintainers can’t pull their own submissions.
(Or at the very least the submissions must be reviewed by someone else.)  That’s part of why some of my pull requests
are still outstanding, even the very simple ones. 

Once you are in the organization I’ll point you at some pull requests.  I’ll start with my simpler pull requests — which need review
before I can merge them — and the LTI patches.  I gather that you have connections with D2L and notice that you have 
submitted patches for the LTI interface.  Unfortunately my university uses Blackboard (and I run a Moodle instance as well) so 
I mostly work on keeping those interfaces working.  Do you know anything about LTI Advantage by any chance?  We’ll need
to make sure that our LTI interface is compatible with that.

I hope that’s enough to get started. Let me know if you have any questions.

Take care,

Mike

dsteinmo:
Hi Mike,

Thanks for the details. It seems like a standard git-flow type workflow where reviewers are also expected to test code before it gets merged, which is fair enough, and I'm happy to follow along with this. Feel free to assign me to a few simple reviews to start, and we'll take things from there.

I am a software developer at D2L with at least 4 years experience in git and github. I've also taken up lecturing math part-time at a nearby university (Wilfrid Laurier) this year, and our department has adopted WeBWorK as its online homework system and uses Brightspace as its LMS. My interest in integrating the two web apps with LTI thus happened quite naturally.

I have heard of LTI Advantage, but don't know the implementation details yet. I can help WeBWorK in supporting it, though, as time allows me to. That's all for now!

Cheers,
Derek.

lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
```diff
@@ -1274,7 +1274,7 @@ sub showHTML {
 #
 sub isText {
 	my $string = shift;
-	return $string !~ m/[^\s\x20-\x7E]{4}/;
+	return $string !~ m/(?=[^\r\n\t])[\x00-\x1F\x7F]{2}/;
```
dpvc: Is there a reason you aren't just leaving `\r`, `\n`, and `\t` out of the list of characters in brackets rather than doing a lookahead?  It seems to me that your expression would match something like `"\x00\n"`, since the the lookahead only prevents `\r`, `\n`, and `\t` in the first character.  So this is effectively matching against *one* non-standard character.

Also, is there a reason for reducing the match from 4 to 2?

t/test_fileManager.pl
```diff
@@ -0,0 +1,25 @@
...
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abcdef\x0Aasdfd"), 'text with 1 control character is still considered text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("\x00abcdef\x1Fghijklm"), 'text with multiple instances of single isolated control characters is still considered text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abc\n\ndef\n\n\n\nghijklm"), 'text with multiple consecutive newlines is text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abc\r\n\r\ndef\r\n\r\n\r\n\r\nghijklm"), 'text with multiple consecutive CRLF\'s (windows style) is text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("                 "), 'string of all spaces is text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("      \r\n\t\r\n\t        "), 'string of all spaces, tabs, CRs and newlines is text';
+ok WeBWorK::ContentGenerator::Instructor::FileManager::isText("abcdef\t\t\t\tasdfjka"), 'string with multiple consecutive tabs is text';
```
dpvc: I'd like to see a test for the edge cases of `abc\x00\ndef` and `abc\n\x00def`.

dsteinmo: Sounds good, I'll add these tests and confirm I have a regex that makes them pass!

lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
dsteinmo: Thanks for the feedback! I'm not aware of a way (in perl) of excluding characters from a range without a look-ahead. (Aside from listing every single character I want). If you do know a way, please let me know. In any case, I will try to tweak the regex until I can pass your edge cases.

As for the 4 becoming 2, the 4 seemed kind of arbitrary, so I went for a stricter condition here. I guess since I don't really have any other good reason, I can put it back to 4.

dpvc: > I'm not aware of a way (in perl) of excluding characters from a range without a look-ahead

Just leave them out of the brackets:  `[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]`  This leaves out `\x09` (tab, `\t`), `\x0A` (linefeed, `\n`), and `\x0D` (carriage return, `\r`).

> the 4 seemed kind of arbitrary, so I went for a stricter condition 

If by stricter, you mean it is harder to be called a text file, then I agree that this 2 is stricter than 4.  (That is, to be a binary file, you only need to match two non-standard characters in a row rather than 4, so it is easier to be called a binary file with 2 than with 4, so harder to be called a text file.)  I suspect that, for WeBWorK files, even one non-standard control character would be enough to make a file non-text (though there used to be text files that contained bells (`\x07`), backspaces (`\x08`), and form-feeds (`\x0C`), but that was the old days, and I doubt they will appear in files for WeBWorK).  I was just curious about your reason for the change.  And yes, 4 was pretty arbitrary, as well.

dpvc:
I'm OK with the changes as they now stand.

mgage:
Thank you @dsteinmo, @dpvc and @alex-jordan for your help with reviewing this pull request.  It seems ready to go.

-- Mike

Issue #811, opened 2017-10-02
Error updating grades via LTI


aruethe2:
I had a problem with grades not updating in our Moodle LMS via the LTI interface.  I saw the following error in ssl_error_log:

`An error occurred while trying to update grades via LTI: Can't call method "getSettingValue" on an undefined value at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm line 295.\n\n`

I noticed that in the `mass_update` function of SubmitGrade.pm, the first line is `my $self = shift;`, but in the calling function from /lib/WeBWorK/ContentGenerator.pm, there aren't any arguments passed to `mass_update`.  

I found that I was able to eliminate the error and update the grade via LTI by updating line 180 in /lib/WeBWorK/ContentGenerator.pm.

From:
`-           my $result_message = eval { $grader->mass_update() };`

To:
`+           my $result_message = eval { $grader->mass_update($self) };`

Was this the right thing to do?  We're using WeBWorK 2.12.

dsteinmo:
This sounds like a pretty serious bug, perhaps it is even hotfix-worthy for 2.13? I'm not a developer on this project, so I'm not sure. Maybe create a pull-request direct into the 2.13 release branch and see what the developers think.

mgage:
This may be at least partially resolved by PR #856 and #857

drgrice1:
This issue is obsolete.

PR #819, opened 2017-11-04
Pass missing param into grader->mass_update().

dsteinmo:
This should fix https://github.com/openwebwork/webwork2/issues/811 but requires verification.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -177,7 +177,7 @@ sub go {
 	    my $grader = shift;
 	    
 	    # catch exceptions generated during the sending process
-	    my $result_message = eval { $grader->mass_update() };
+	    my $result_message = eval { $grader->mass_update($self) };
 	    if ($@) {
 	      # add the die message to the result message
 	      $result_message .= "An error occurred while trying to update grades via LTI.\n"


Alex-Jordan:
I think this might be making the logged error message go away, but not really doing what this code block should be doing. With `$grader->mass_update()`, it _is_ passing an argument; it applies `mass_update()` to `$grader`. So I think the original issue is that `$grader` is not defined right. Then you are feeding it `$self` instead, and that appeases the error message issue, but (I think) it is the wrong thing for `mass_update` to receive.

Tracing `$grader` back up through a few levels to find outwhere it all goes wrong, I wonder if the current line 165 `my ($self) = @_;` should really be `my $self = shift;`. Could you try making this change instead and see if you still get the error described in #811? And also see if you can confirm that the mass update is working with or without this change?

dsteinmo:
Sure, I'll take a look at this when I get a moment.

rdalhoff:
I tried the suggested change of lib/WeBWorK/ContentGenerator.pm current line 165 from my ($self) = @_; to my $self = shift;  but it still gives me an error.
I assume my error is the same in trying to update grades in Canvas as this original problem here was with Moodle.

Grades are not being updated in Canvas using the Advanced LTI 
We are running version $WW_VERSION = 'WeBWorK-2.13'; 

The apache error_log has this entry every 5 minutes:
[Fri Feb 02 14:57:31 2018] [error] [client 129.186.6.15] An error occurred while trying to update grades via LTI: Can't call method "getSettingValue" on an undefined value at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm line 297.\n\n

Line 297 in /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
is
my $lastUpdate = $db->getSettingValue('LTILastUpdate') // 0; 

I am using this setting in the authen_LTI.conf file:
$LTIGradeMode = "homework"; 

It appears that the $db variable is not set or defined in this routine:
sub mass_update {
my $self = shift;
my $r = $self->{r};
my $ce = $r->{ce};
my $db = $self->{r}->{db};
my $lastUpdate = $db->getSettingValue('LTILastUpdate') // 0;

goehle:
So my perl is rusty, but unless things have changed a lot I think `my $self = shift` and `my ($self) = @_;` do the same thing, namely set `$self` to be a pointer to the current object, which is always set as the first value in `@_` in method calls.  In particular unless I'm way off the solution in #811 isn't going to work.  Any method call like `$grader->mass_update()` will automatically put a reference to `$grader` as the first argument in `@_` for the call to `mass_update`.  Its a funky unintuitive thing about how perl objects work.   (I'm not trying to insult anyone by explaining the basics, I'm literally trying to remember how it operates, so if things have changed so that what I said above isn't true anymore feel free to ignore this.)  The only reason it might matter is because `shift` actually shifts the value off of `@_` while the second one doesn't.  However `@_` isn't referenced again, so that won't affect anything in this case.  

The problem is most likely more subtle.  Again, those familiar with the code base will have to excuse my delving back into basics.   The problem is that the database object `$db` is not defined at 297 of Grader.pm.  This is either because the `$db` variable is not defined in the `$r` object or because the `$r` variable is not defined in the `$self` \ `$grader` object.  (You would have to debug to find out which).  The `$r` object is a reference to the apache request object generated by mod_perl by the actual apache web request.  Its constructed deep in the ContentGenerator code and a reference to it is passed to the `$grader` object upon construction (line 174 of ContentGenrator.pm).  It seems unlikely to me that `$r` is not defined since that contains all of the parameters of the request and I feel like things would fail much earlier if it weren't around.  On the other hand I can imagine situations where `$r` gets defined but for one reason or another `$db` isn't defined.   This could be because the request doesn't have enough information to get the right database (i.e. its missing `courseID` for some reason) or because there is some other error which is preventing `$db` from getting created.  In this case the `$r` object would be created, but the `$db` object wouldn't and since `mass_update()` is pretty much the first thing that happens in `go` that's where an error would come from.    This is all speculation on my part.  I can imagine that there are random web requests which trigger `mass_update()` call without having a defined `$db` variable but I don't know what they would be.  Its a place to start looking anyway.

mgage:
Thanks Geoff. That makes sense to me.  It sounds like there aren't any unusual quirks to look for --  I knew that you had done quite a bit of work with this code at one point which is why asked.  LTI calls aren't configured in exactly the same way as calls coming in through the webpage interace so I'll look for the problem there.  Thanks again.

mgage:
I'm closing this to be replaced by PR #856 and PR #857

PR #820, opened 2017-11-15
Remove Mail::Sender and clean up spaces.

apizer:
Edited check_modules.pl to remove Mail::Sender and also cleaned up some white space issues.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -43,7 +43,7 @@ my @modulesList = qw(
 	Carp
 	CGI
 	Class::Accessor
-        Crypt::SSLeay
+	Crypt::SSLeay
 	Dancer
 	Dancer::Plugin::Database
 	Data::Dump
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I inspected the code/text changes, and it looks right. I was able to run checkModules with no issues.

I grepped for "Mail::Sender", and it appears in the following files, but commented out. Should we just delete these lines referencing Mail::Sender instead, if we are committed to something else now?

```
lib/WeBWorK/ContentGenerator/Feedback.pm
lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
lib/WeBWorK/Utils/DelayedMailer.pm
lib/WeBWorK/Utils.pm
```

Lastly, these files have "Mail::Sender", and it is _not_ commented out. Do these need to change?

```
lib/WeBWorK/Utils/RestrictedMailer.pm
lib/WebworkWebservice/CourseActions.pm
```

mgage:
As far as I can tell RestrictedMailer.pm is not being used.  This and RestrictedClosureClass.pm were introduced by Sam Hathaway @sh002i with the plan of using it to block some possible security holes still open with the Safe compartment. They were to be used with Safe::Hole which is a CPAN module.

I'll merge this PR  now.  Soon I'll issue a new PR fixing the Mail::Sender reference in Webworkwebservice/CourseActions.pm.  There is note there from Peter Staab @pstaabp indicating that the webservice function sendMail() has not been working for some time. It's main use case at the moment is for WW3.

PR #823, opened 2017-12-04
show number of active students in Statistics

heiderich:
Show the number of active students for each problem on the Statistics page. As far as I am aware this number this not easily available elsewhere.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
@@ -680,7 +680,7 @@ print  CGI::p($r->maketext('The percentage of active students with correct answe
 			]
 			)),
 		CGI::Tr(CGI::td(
-			[ $r->maketext('number of active students'), map {($number_of_students_attempting_problem{$_})
+			[ $r->maketext('&#35; of active students'), map {($number_of_students_attempting_problem{$_})
 			                      ? $number_of_students_attempting_problem{$_}
 			                      : '-'}
 			                       @problemIDs
--- a/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
@@ -680,7 +680,7 @@ print  CGI::p($r->maketext('The percentage of active students with correct answe
 			]
 			)),
 		CGI::Tr(CGI::td(
-			[ $r->maketext('&#35; of active students'), map {($number_of_students_attempting_problem{$_})
+			[ $r->maketext('# of active students'), map {($number_of_students_attempting_problem{$_})
 			                      ? $number_of_students_attempting_problem{$_}
 			                      : '-'}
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Looking at this, I see several things about these tables that should change. I started to catalog them, but now it seems irrelevant to this PR. But I think we should revamp the structure/labeling of these tables at some point.

As for this PR, could we shorten 'number of active students' to '# active students'? Doing so will make it so the header column on the table doesn't get much wider from this.

heiderich:
I agree that these tables could be improved. But for the moment being I hope this new line will be an improvement. It will hopefully also make the person who revamps these tables to include these numbers. I replaced "number" by "#" in commit 942d43b .

Alex-Jordan:
I tested this and it works as advertised.

The`maketext` argument is `'&#35; of active students'`, using `&#35;` instead of `#`. Earlier in this file at line 662, there is a `maketext` call with `'Problem #'` as the argument, using `#`, which then shows up in the localization files also using `#`. So what I am wondering is does using `&#35;` here mean the localization file must also use `&#35;`? Or will it not matter, and the localization file can use either `&#35;` or `#`?

Aside from making a deliberate decision about `&#35;` versus `#`, this can be merged. Despite my earlier suggestion, I think that `'active students'` would suffice, keep the first column even more narrow, and avoid a deliberate decision about `&#35;` versus `#`.

(As noted earlier, I think the language surrounding these tables and how things are styled could use some attention. A project for some future day.)

heiderich:
I wrote `&#35;` instead of `#`, because it is the correct HTML encoding for # in HTML. Then my tests show that you need `&#35; of active students` in the `.po` file (`# of active students` would not work). I consider this not to be very elegant, especially because it is HTML specific.

We could probably keep the WeBWorK source code as well as the localization files more readable when using the ASCII `#` character instead of `&#35;`. I did not observe any rendering problems using it. It is also used in other places, such as line 662 in `lib/WeBWorK/ContentGenerator/Instructor/Stats.pm`:

                        [ $r->maketext('Problem #'), 

What do you think?

Considering the phrase `The percentage of active students with correct answers for each problem` above the table, I think it would be good to indicate that this line shows *numbers* of students. That phrase itself seems not very suitable to me, because the average attempts is neither a percentage.

Alex-Jordan:
Given that `#` is used elsewhere and appears to not cause any rendering issues, I like the human readability of using it in the WeBWorK source files (and the `.po` files).

> Considering the phrase The percentage of active students with correct answers for each problem above the table, I think it would be good to indicate that this line shows numbers of students. That phrase itself seems not very suitable to me, because the average attempts is neither a percentage.

I agree, and whatever happens now, cleaning up the descriptive language should be part of cleaning up the statistics tables later. 

It sounds like making it be `# of active students` in both paces is what we are settling on for now. Does that sound right? If so, please make that change and I will merge this.

heiderich:
I changed `&#35;` for `#`.

PR #825, opened 2017-12-08
Fix spelling of maketext in achievements assignment module

mgage:
Fixed the spelling of maketext (was maktext) in file.  This bug was blocking the ability to update the achievements for an individual student. 

To test:  

before patch trying to unassign an achievement for a single student gave an error message that maktext was not defined. 
(I believe assigning a single student would give the same message.) 

after the patch the achievement is assigned or unassigned properly.


mgage:
This should be an easy and quick pull request to test for those interested in becoming comfortable working with github.

PR #830, opened 2017-12-25
add maketext calls

heiderich:
maketext calls in PGProblemEditor.pm and SetMaker.pm


mgage:
Thanks Florian and Alex for all the activity on pull requests.

PR #839, opened 2018-01-12
Change statements to students with regard to reduced scoring.

jwj61:
This changes messages students see when reduced scoring is in effect.  It refers to the start of reduced scoring as the due date.  This is because students found the original way confusing -- they felt cheated for getting reduced credit when they submitted their work before the due date, in spite of messages in webwork.


mgage:
Looks good.  Translations will have to be updated.

PR #840, opened 2018-01-13
modelCourse problem setOrientation/prob07.pg domain error.

gajennings:
Bug 4050 reports a "can't generate enough valid points for comparison" error in modelCourse/templates/setOrientation/prob07.pg

Fixed by imposing limits on the domains of the three displayed formulas.  The three domains are different so I changed one variable from "x" to "u" to make it easier to specify the limits.

### DIFF
--- a/courses.dist/modelCourse/templates/setOrientation/prob07.pg
+++ b/courses.dist/modelCourse/templates/setOrientation/prob07.pg
@@ -111,16 +111,24 @@ $al = new_aligned_list(
     spacing => 10,
 );
 
-Context("Numeric")->variables->are(t=>'Real'); $t = Formula('t');
-Context("Numeric")->variables->are(y=>'Real'); $y = Formula('y');
-Context("Numeric")->variables->are(x=>'Real'); $x = Formula('x');
-Context()->flags->set(limits=>[-2,10]);
+Context("Numeric")->variables->are(
+  u => ['Real',limits=>[0.1,1.5]],
+  t => ['Real',limits=>[-1.9,-0.1]],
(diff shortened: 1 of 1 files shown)


jwj61:
Is there a reason this is a pull request instead of just pushing the changes?

mgage:
I don't believe George has commit privileges on this repos.

gajennings:
I probably do have push permission to webwork2.  Git says:

$ git remote -v
origin    https://github.com/openwebwork/webwork2.git (fetch)
origin    https://github.com/openwebwork/webwork2.git (push)

but I haven't pushed to webwork2 before so I don't know what the review process is.  I figured I'd let you guys sort that out, even though the changes in this file were trivial.

George

On 01/13/2018 06:29 AM, John Jones wrote:

Is there a reason this is a pull request instead of just pushing the changes?

mgage:
I prefer the pull request, merge method for webwork2 and pg because it leaves a clear and public message of how the change was made.  For trivial changes this doesn’t matter much but for consistency I’ve gotten in the habit of submitting all my changes as PR and not pulling them myself. (You can see a couple of pull requests sitting there.)

I’m not sure how much of a message is left when you push directly.  Is it just the commit message? Sometimes those are a bit brief.

For the library I think push would be ok since the majority of the changes are in one problem and orthogonal to everything else that is going on.

Take care,

Mike

Alex-Jordan:
I think it would be good practice to never push something or merge your own PR. I have caught trivial commits with some kind of typo before. Sometimes something that the author couldn't have caught, having to do with a text encoding issue.

PR #842, opened 2018-01-22
bugfix: ShowMeAnother "problem-level SMA = course default" fix

drdrew42:
problem-level "ShowMeAnother = -2 (course default)" setting breaks functionality, see webwork.maa.org/moodle/mod/forum/discuss.php?d=4218 for details.


jwj61:
I don't know anything about this particular issue, but pull requests should be made against the develop branch.

drdrew42:
I get that, but this is a simple one line edit, not a new feature - it has no chance of breaking anything else, and it fixes Show Me Another to work the way it's already stated to work in the existing release.

Bug replication:

1. Enable SMA in Course Config/Optional Modules, and set default attempts to 0 at the Course Config level;
2. Create/populate a problem set or use existing set, set one or more of the problems' SMA to 0 in Hmwk Sets Editor;
3. Visit problem page, click Show Me Another to confirm that a new version _can_ be generated.
4. Return to Hmwk Sets Editor, change problem from 0 to -2 (supposedly to match the course config setting - which we also set to zero, so nothing should change)
5. Re-visit the problem page, click Show Me Another to confirm that a new version now _cannot_ be generated. (Although we know it can, because we just confirmed it when SMA=0 at the problem level.)
6. Go back to Hmwk Sets Editor, change from "Default" back to 0 - then return to problem page and see that Show Me Another now works again.

The rest of the logic is good in ShowMeAnother.pm, except for the line that I changed - where %showMeAnother{active} is initialized. When the "-2 (course default)" option was added at the problem level, this line must have been overlooked. All other conditionals allow for this setting and handle it appropriately. This single edit resolves the error described above.

If you still think it's preferable, I'll re-submit this against develop - I just thought this was the best place for it.

Alex-Jordan:
I haven't tried but I think this would not do what is desired if the value is -2 but the course default is set to -1. I have not dived into the surrounding code, but is there something like an "effective" SMA value, where if the SMA value is -2, then the effective value is assigned to be the course default? The test should be if such an "effective" value is nonnegative.

There are questions to settle about what branch(es) to merge this into, but first I think the above point may be an issue.

drdrew42:
Line 111 of ShowMeAnother.pm sets %showMeAnother{TriesNeeded} to match the course default prior to the logic of whether or not to continue with SMA. As a result, if the course default is set to -1, the call to can_showMeAnother in line 119 will cause Show Me Another to die with $showMeAnother{TriesNeeded} = -1. (Problem.pm line 253)

Confirmed on my end: 

- enableShowMeAnother=TRUE, 
- course default attempts = -1, 
- problem SMA = -2 

results in no SMA button appearing on the problem page.

mgage:
develop will be released as 2.14 very soon at this point so I'm committing this to develop.  It's a small change, but it took awhile to test.  I'm also ok with the changes to setOrientation/prob07.pg.

PR #844, opened 2018-03-22
WeBWorK  client dir cleanup

mgage:
It looks like these cleanup commits didn't get included in  WeBWorK 2.13 (I meant to include them.)

The main thing that needs testing is sendXMLRPC.pl.   

There are instructions for testing it in the POD documentation in the file


Alex-Jordan:
It appears that all three files are identical, except that clients/TestWW.html.dist has `outputformat=debug` for the third problem where the other two still have `outputformat=simple`.

Alex-Jordan:
OK, sorry for my confusion. I see now that clients/Second_semester_calculus_topics.html.dist and clients/TestWW.html.dist are different. I had accidentally copied one as the other. 

But that still leaves clients/TestWW.html, which I suspect is not supposed to be here.

Alex-Jordan:
There are files here:  clients/t/input.pdf and a clients/t/input.txt. Are these intended to be committed to the repo? Or are they just the output from applying sendXMLRPC to  clients/t/input.pg?

In clients/ww_credentials.dist, it uses a rochester webwork server. Is that your intention? It could be webwork.myschool.edu or something like that instead.

This PR has some deletions and file renaming in `conf/` and `lib/`. Are you certain those are supposed to be here? I'm unsure how to test these are safe.

OK, these plus the question about clients/TestWW.html are my first run questions. Next I will try using the command line sendXMLRPC tool. It may be as late as the weekend.

Alex-Jordan:
With the various .dist files here, should the non-.dist versions be listed in .gitignore?

In the script, there is a place where "Command for editing and viewing the **tex** output of the pg question" should say "Command for editing and viewing the **pdf** output of the pg question".

At lines 413 and 414 of the script, some slashes are missing.

At line 521, the variable `$formatter` is used, but it doesn't know what that is. I get the error `Global symbol "$formatter" requires explicit package name at ./sendXMLRPC.pl line 521.`

Starting at line 528, I get many similar errors about `$xmlrpc_client`: `Global symbol "$xmlrpc_client" requires explicit package name at ./sendXMLRPC.pl line 528.`

I don't know how to patch these last kinds of errors, so I'll wait to continue testing.

mgage:
* There are some big file changes -- but they mostly consist of moving files around and removing files that haven't been used in years.

1. clients/README is replaced with `clients/readme.md`
2. references to hosted2 have been replaced by references to demo
3. clients/sendXMLRPC.pl is the file that needs to be checked closely
4. *.html files have been removed by *.html.dist files are kept


There are some changes in the `webwork2/conf` folder

1. Improved description in `conf/README` (site.conf is no longer "new")
2. defaults.config has new address for MathJax which will eventually replace the old one.
3. The devel.apache-config files have all been removed 


`lib/FormatRenderedProblem.pm` has been added to support WebWorkClient.pm and sendXMLRPC.pl

`lib/RQP.pm` is not used and has been removed.

mgage:
I apologize for bugs in the sendXMLRPC.pl file.  It turns out that I did work on this file in two different local branches.  I had to go back and reconcile the changes in the two branches which took some time.

mgage:
This pull request is closed in favor of PR #859 and a later pull request updating sendXMLRPC.pl

PR #845, opened 2018-04-02
Prefer whitelisting over blacklisting to prevent XSS

catatonicprime:
Warning: I haven't I actually tested this (I don't have a deployment of my own & I am looking for a non-production environment)

Anyway this PR prevents the below XSS by preferring a whitelisting approach for theme selection over a blacklisting approach for data validation of the 'theme' querystring parameter.

https://[site]/webwork2/?user=&effectiveUser=&key=&theme=1%3Cimg+src=xyz+onerror=alert(150)%3E%3Cxss_/%3E

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -536,7 +536,7 @@ sub content {
 	
 	my $themesDir = $ce->{webworkDirs}{themes};
 	my $theme = $r->param("theme") || $ce->{defaultTheme};
-	$theme = $ce->{defaultTheme} unless $theme =~ m!^[a-zA-Z0-9_.-]+$!;
+	$theme = $ce->{defaultTheme} if $theme =~ m!(?:^|/)\.\.(?:/|$)!;
 	#$ce->{webworkURLs}->{stylesheet} = ($ce->{webworkURLs}->{htdocs})."/css/$theme.css";   # reset the style sheet
 	# the line above is clever -- but I think it is better to link directly to the style sheet from the system.template
 	# then the link between template and css is made in .template file instead of hard coded as above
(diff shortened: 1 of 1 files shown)


dpvc:
I think your regex isn't sufficient.  If I read the expression correctly, you are using the default theme unless the theme name contains one of the special characters.  It only has to match one for that to be true.  You are not requiring that *all* the characters are in the special set. For that, you need something like `m!^[a-z0-9_.-]+$!i`.  This assumes you only want to use ASCII characters, and not Unicode characters needed for other languages.

catatonicprime:
omg... yeah I totally missed the ^$ This is fixed now.

And you're right I have no accounting for Unicode - which may be totally acceptable for filenames in other languages. This is... not fixed yet. Any thoughts on how to white list acceptable Unicode, in the meantime I might go and look around on owasp...

catatonicprime:
Actually on further thought it'd probably be better to do output encoding instead. Like that's _really_ the right way to solve an XSS.

catatonicprime:
There we go - that should fix the XSS without changing any other constraints.

dpvc:
Yes, I like that better.  Approved.

mgage:
Is it ok to just merge this into develop? or is it urgent enough that you want it merged into both master and develop?

mgage:
(and thanks @catatonicprime  for your contribution.  :-) )

dpvc:
@mgage, I don' think it requires a hot fix to master.  This avoids an XSS code insertion based on the `theme` parameter passed in the URL, and if a student changes that, it can insert javascript into the page he or she sees.  But it only affects his or her own page, not anyone else's, so I consider it a low-value XSS vector.  I suppose it would be possible for a man-in-the-middle attack to modify the theme without the student's knowledge, but that seems a pretty slim possibility, and since it would result in the page not being properly rendered (no theme file) and the error message about the incorrect theme, I think it would be pretty apparent.  I am unaware of this happening in the 15 years that webwork2 has been available.  I'm also not sure of using https would prevent that (I don't know if the URL or any of its parameters are in the secure channel).   

In any case, I think it can wait until the next release.

dpvc:
> Https does not mitigate XSS

I meant in terms of the man-in-the-middle attack, not in general.  I looked it up, and the query string is sent encoded, so it should prevent man-in-the-middle XSS.  

As for the phishing scenario, I thought of that (after posting), and it seems to me that since such a link can be to anywhere, they don't need to highjack the webwork server for that, but just link to their own site that imitates the login page (as you suggest).  Although there is an advantage to the URL looking like a legitimate one to a casual glance, I don't think it increases the effectiveness of the phishing attack by all that much.

I didn't mean to say that there is no reason to be concerned, only that I think the threat does not rise to the level of requiring an immediate hot-fix to master.  I was trying to answer Mike's question about that.  I have to admit that I didn't look at the branches (as I should have) and didn't realize this was from your master to webwork's master branch.  That makes it a hot-fix, and I would have recommended the you close this PR and make one from develop to develop instead.

catatonicprime:
I can do that.

PR #851, opened 2018-05-10
preliminary code - Add lang and dir attributes to HTML div elements containing problems

taniwallach:
This is NOT ready to merge yet.

This code sets the HTML LANGuage attribute and a DIRection attribute on the DIV elements which contain problems. It depends on PG code added in [PG pull request 356](https://github.com/openwebwork/pg/pull/356).

Note: It is somewhat related to [PR 850](https://github.com/openwebwork/webwork2/pull/850) which uses the course language setting to set the same attributes on the main HTML element of the general web pages (for math4 theme). 

A new WeBWorK::Utils::DetermineProblemLangAndDirection module was added with a function which processes the PG flags, the course language setting, and the value of a new configuration variable
perProblemLangAndDirSettingMode  to determine the attributes to add. This keeps the decision making code in a single place.

The output_problem_lang_and_dir() function was added to Problem.pm to do the actual addition of the attributes when it is used to generate the displayed problem, and is called in the math4 and math4-ar templates. In problems put into the HTML pages using the various "SetManager" type ContentGenerator, code was written to add the HTML attributes to a hash which is passed into CGI::div instead of the default hash {class=>"RenderSolo"}. 

Warning: This has not yet been done for  lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm.

The new configuration setting can be used to force specific values for the attributes, to turn off the addition of these attributes, and to set the fallback default values used when there are no PG flag values set (as is the case for all existing content). This should allow sufficient flexibility so that the with a proper setting, the new code will not break any existing courses. The default value of the variable was set to "force::ltr" which leaves the language attribute out, and forces the direction attribute to "ltr" which it would be expected to have in all courses, except those in Hebrew or Arabic.

Testing:

1. See how viewing problems (as a student), in the OPL library browser, and other place behaves with various settings of the configuration setting. I recommend trying force::rtl, force::ltr, auto:en:ltr.
2. Check how it behaves under math4 and math4-ar (a RTL theme).
3. Checks can also be made to see the language attribute is set (show source).
4. Someone using non-English problems would want to test auto:LANG:ltr (LANG here is a setting to be chosen) which would set the language to the language LANG even when the problem language is not set in the problem PG code instead of the global fallback default of English. The choices for LANG are the same ones which can be chosen as a WW course language, except we use the modern standard "he" for Hebrew and not "heb" as was chosen once for the course option.

To be done in the future:

- Do testing with the themes other than system.theme with which the initial testing was done.
- Figure out how to add similar code to ProblemSetDetail.pm where RenderSolo is used.
- Consider where else problems can be rendered and how to add suitable code to the relevant places.
- Consider whether changes are needed to the default behavior in automatic mode.
- Consider if changes should also be made to the math3 theme.


heiderich:
With this branch and pull request https://github.com/openwebwork/pg/pull/356 for pg, I obtain the following error after changing the language of a course to "ar". Am I doing something wrong?

```
Error messages

    Unknown encoding 'CHARSET' at /usr/share/perl5/Locale/Maketext/Lexicon/Gettext.pm line 142. 

Call stack

    in Carp::croak called at line 198 of /usr/lib/x86_64-linux-gnu/perl/5.26/Encode.pm
    in Encode::decode called at line 142 of /usr/share/perl5/Locale/Maketext/Lexicon/Gettext.pm
    in Locale::Maketext::Lexicon::Gettext::transform called at line 38 of /usr/share/perl5/Locale/Maketext/Lexicon/Gettext.pm
    in Locale::Maketext::Lexicon::Gettext::__ANON__ called at line 86 of /usr/share/perl5/Locale/Maketext/Lexicon/Gettext.pm
    in Locale::Maketext::Lexicon::Gettext::parse called at line 213 of /usr/share/perl5/Locale/Maketext/Lexicon.pm
    in Locale::Maketext::Lexicon::_force called at line 222 of /usr/share/perl5/Locale/Maketext/Lexicon.pm
    in Locale::Maketext::Lexicon::SCALAR called at line 492 of /usr/share/perl/5.26/Locale/Maketext.pm
    in Locale::Maketext::_try_use called at line 381 of /usr/share/perl/5.26/Locale/Maketext.pm
    in Locale::Maketext::get_handle called at line 49 of /opt/webwork/webwork2/lib/WeBWorK/Localize.pm
    in WeBWorK::Localize::getLoc called at line 248 of /opt/webwork/webwork2/lib/WeBWorK.pm


```

PR #852, opened 2018-05-16
Gateway and Grade Bugfix

goehle:
This pull request contains a couple of bug fixes. 

1.  The Hmwk Sets Editor now properly processes the "hide_score_by_problem" parameter for gateways.  To test: 
  - [ ] Import a Gateway set and check in the database that the hide_score_by_problem column is not null for the set
  - [ ] Export and Import a Gateway set and make sure that the hide_score_by_problem field is being exported and imported correctly. 
  - [ ] Check that students are shown scores on problems when they should be. 

2.  The Course Total percentage in the Grades page (and for the LTI course total grade) has been changed so that it no longer includes grades from sets that are not yet open. To test: 
  - [ ] Make a set that is not open and check that it does not change the course totals on the Grades page. 
  - [ ] Make a set that is not open and check that it does not change the course total for the LTI grade passback feature.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Grades.pm
+++ b/lib/WeBWorK/ContentGenerator/Grades.pm
@@ -29,7 +29,7 @@ use warnings;
 #use CGI qw(-nosticky );
 use WeBWorK::CGI;
 use WeBWorK::Debug;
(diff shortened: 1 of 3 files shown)


mgage:
Works fine.  Thanks Goeff.

PR #853, opened 2018-05-30
Adding R support

xcompass:
This PR add R support to docker image and docker-compose. After running docker-compose up, an additional container Rserve will be running to process requests from webwork.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,7 +73,7 @@ RUN apt-get update \
        git \
        mysql-client \
     && curl -Lk https://cpanmin.us | perl - App::cpanminus \
-    && cpanm install XML::Parser::EasyTree Iterator Iterator::Util Pod::WSDL Array::Utils HTML::Template XMLRPC::Lite Mail::Sender Email::Sender::Simple Data::Dump \
+    && cpanm install XML::Parser::EasyTree Iterator Iterator::Util Pod::WSDL Array::Utils HTML::Template XMLRPC::Lite Mail::Sender Email::Sender::Simple Data::Dump Statistics::R::IO \
     && rm -fr /var/lib/apt/lists/* ./cpanm /root/.cpanm /tmp/*
 
 RUN mkdir -p $APP_ROOT/courses $APP_ROOT/libraries $APP_ROOT/webwork2
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,10 @@ services:
     volumes:
       - ".:/opt/webwork/webwork2"
       - "./.data/courses:/opt/webwork/courses"
+      # Uncomment the line below to use local OPL for development
       #- "../opl:/opt/webwork/libraries/webwork-open-problem-library"
+      # Uncomment the line below to use local PG for development
+      #- "../pg:/opt/webwork/pg"
     ports:
       - "8080:80"
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,3 +23,7 @@ services:
       - "8080:80"
     environment:
       DEV: 0
+  r:
+    image: ubcctlt/rserve
+    ports:
+      - "6311:6311"


mgage:
I'm having trouble connecting manually to the database.  The docker-compose operation emits this fact:
```
 2018-07-01 20:22:42 140023753058240 [Note] mysqld: ready for connections.
db_1   | Version: '10.1.30-MariaDB-1~jessie'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306  mariadb.org binary distribution
```
But when I type mysql (either alone or with a user name and password) into a bash terminal
opened within the container I get:

root@eb2fd0ba2da4:/usr/bin# mysql
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)

mgage:
Is there a command I can use from inside the webwork2 container to test that the Rserver is running?

It appears to be running, from the printout from docker-compose up

I've uncommented the appropriate line in the localOverrides.conf file to define the Rserve host
as 'localhost'.

I get this error when trying to run a problem that uses RserveClient.pl

```
 connect: Connection refused at line 129 of [PG]/macros/RserveClient.pl
   Died within Carp::croak called at line 27 of /usr/local/share/perl/5.22.1/Statistics/R/IO/Rserve.pm
   from within Statistics::R::IO::Rserve::fh called at line 323 of /usr/local/share/perl/5.22.1/Statistics/R/IO/Rserve.pm
   from within Statistics::R::IO::Rserve::_send_command called at line 234 of /usr/local/share/perl/5.22.1/Statistics/R/IO/Rserve.pm
   from within Statistics::R::IO::Rserve::eval called at line 129 of [PG]/macros/RserveClient.pl
   from within main::rserve_start called at line 102 of (eval 1559)
```

telnet localhost 6311  

doesn't work because telnet doesn't seem to be available from within the container.

Is there some other tool I can use to test the connection to the R server?

taniwallach:
About DB access - mysql seems to be available also in both docker containers, but the startup command is a bit different in the app container, as it needs to give the "-h" option. The following work for me:
 * DB container:
   o docker container exec -it webwork2_db_1 bash
   o mysql -u webworkWrite -p webwork
 * DB container:
   o docker container exec -it webwork2_app_1 bash
   o mysql -u webworkWrite -p webwork -h db

mgage:
Thanks for helping out Nelson.  

This page is a helpful reference:    http://webwork.maa.org/wiki/R_in_WeBWorK

Library/UBC/STAT/STAT300/hw07/stat300_hw07_q02.pg  is one problem

(found under Statistics -> Simple linear regression -> Hypothesis tests).   UBC has created many more.

nmoller:
To make it work you have to modify the docker-compose.yml file to add in the app service:
```
depends_on:
  - db
  - r
```
(you have to had the r server)
Once you modifies `localOverrides.conf`:

`$pg{specialPGEnvironmentVars}{Rserve} = {host => "r"};`

Everything works as expected:
image

mgage:
Making the changes to the docker file and to localOverrides.conf make it work perfectly.  Beautiful.

Thanks very much.

taniwallach:
I have been working with Docker for a while, and had an existing "app" container, which apparently was not updated to include Perl module Statistics::R::IO automatically.

1. I personally did not need to have the app container depend on the r container, although that is certainly a logical requirement, and maybe that change triggers an update to the app container which leads it to execute the container setup code and installs the Perl module Statistics::R::IO. That might explain why this change solved the issue for @mgage  and @nmoller .
2. In my case, even after setting 
`$pg{specialPGEnvironmentVars}{Rserve} = {host => "r"}; `
in conf/localOverrides.conf I still did not get data from R displayed, and saw an error message:

> 1. ERROR caught by Translator while processing problem file:tmpEdit/Library/UBC/STAT/STAT300/hw07/stat300_hw07_q02.pg.tani.tmp
> ****************
> ERRORS from evaluating PG file:
>  Statistics::R::IO::Rserve could not be loaded. Have you installed the module? at [PG]/lib/Rserve.pm line 12
3. I overcame the problem (on my machine) by entering the app container and manually installing the Perl module Statistics::R::IO.
   * docker container exec -it webwork2_r_1 bash
   * cpan Statistics::R::IO
after which the sample problem using R executed properly and the plot from R was displayed (after a container shutdown/restart).
4. Note: In earlier attempts to solve the problem I decided to several other things, which in retrospect are apparently not necessary:
  a. Add the list of RserveClient modules to conf/localOverrides.conf as described in the [UBC install instructions](http://wiki.ubc.ca/Documentation:WeBWork/The_WeBWorKiR_Project:_Integrating_WeBWorK_with_R/Installation_guide). I reverted this change and things still work.
  b. Adding a file /etc/Rserv.conf in the r container, as described in the [UBC install instructions](http://wiki.ubc.ca/Documentation:WeBWork/The_WeBWorKiR_Project:_Integrating_WeBWorK_with_R/Installation_guide). I reverted this change and things still work.
 5. I did find it helpful at one point to install telnet in the app container, to test the connection to the R server. Since @mgage  had mentioned the lack of telnet as hampering his testing, here is the procedure I used to get telnet installed.
    * docker container exec -it webwork2_r_1 bash
    * apt-get update
    * apt-get install telnet
    * telnet r 6311

@xcompass - Thanks for creating the docker support in the webwork2 devel tree, it has been of immense help to me in starting to use WW and to work on the WW codebase.

mgage:
Thanks.  I'm just getting started using Docker containers so these examples help.

I found out one other thing about using internal vs. external links for opl and for pg.  
(You can use an external -- i.e. outside the container --directory for the opl and for pg
by uncommenting the lines in docker-compose.yml for local opl and local pg.   When using the local
opl and pg I found that switching to the Homework set editor or switching LibraryBrowser to use "OPL directory" was a  _very_  slow process, my macBook Pro laptop fan started running constantly and the CPU use by the hypervisor went over 100%. (I've never trusted the exact numbers on the Activity Monitor anyway.)  Recommenting those lines and restarting the container 
restored the normal behavior for using HomeworkSetsEditor and the OPL directory in the library.

My guess is that  directory searches became expensive operations with external directory links.

taniwallach:
Some comments on using external pg and opl directories.

I rebuilt my container images, with the settings in the `docker-compose.yml` to use external pg and opl directories. That was needed to make a long-term fix for the missing Perl module problem I encountered. 

1. The container ran into permission problems in compiling color.c from `../pg/lib/chromatic/` which I hacked around (making it world writeable until the compiled file was created) to allow the file to be compiled and saved to disk. This is clearly a permissions problem, but is solved by a one time workaround. @mgage : Maybe some note about various docker hacks can be collected in the GitHub Wiki pages.
2. Although there was a pre-downloaded copy of the OPL in `../opl` the build process downloads and install the OPL from a tar.gz file as that is part of the build process in `Dockerfile`. I'm not sure if this make sense if the feature of using an external OPL directory turns out to work well enough, as that would save on the need to download/install the OPL on image updates.
3. @mgage wrote about heavy CPU usage when using external opl and pg trees. I tested on my machine (Debian Linux) and did not see any such behavior. I did not notice any difference between the time needed to access the OPL by directory tree in the Library browser when using the external OPL tree and using an OPL tree inside the container image. In both cases it took about 3 seconds from when I clicked the `OPL Directory` button until the browser displayed the relevant chooser on the updated web-page. Maybe this is OS related? Can some other people test whether they see any performance issues like those Mike reported.

mgage:
I'm going to pull this.  It passes preliminary tests. At the very least it doesn't interfere with other uses of webwork when you don't run docker.  The comments on this pull request are quite helpful and should be copied over to the "docker newbies" document on the webwork2 github wiki page

xcompass:
Sorry, I'm a little late to the party. I've been off the grid for a few days.

@taniwallach Sometime, it might be useful to use `docker-compose down`, since it is reset the container and network. I found it is cleaner restart when using it.

> I found out one other thing about using internal vs. external links for opl and for pg.
(You can use an external -- i.e. outside the container --directory for the opl and for pg
by uncommenting the lines in docker-compose.yml for local opl and local pg. When using the local
opl and pg I found that switching to the Homework set editor or switching LibraryBrowser to use "OPL directory" was a very slow process, my macBook Pro laptop fan started running constantly and the CPU use by the hypervisor went over 100%. (I've never trusted the exact numbers on the Activity Monitor anyway.) Recommenting those lines and restarting the container
restored the normal behavior for using HomeworkSetsEditor and the OPL directory in the library.

@mgage re: slowness of the external directories. There is an issue for docker on mac to mount directory with large number of files. The workaround is to only mount the subdirectories that you are working on. E.g. opl/UBC/STAT. Also you might try this: https://docs.docker.com/docker-for-mac/osxfs-caching/

> Although there was a pre-downloaded copy of the OPL in ../opl the build process downloads and install the OPL from a tar.gz file as that is part of the build process in Dockerfile. I'm not sure if this make sense if the feature of using an external OPL directory turns out to work well enough, as that would save on the need to download/install the OPL on image updates.

@taniwallach Some of our instructors using external directory to work on building PG problems. But for other WW core developers, they don't need to touch OPL, so they can just use the build in one. Also if someone just wants to checkout webwork, they don't need to download a copy of OPL.

mgage:
Thanks Pan -- that is very helpful.  By the way I have merged this pull request into the develop branch so if you wish to do further tests on your docker addition you can do it from the develop branch on this site. I ran many problems with Rserve.pl using docker since I don't yet have an Rserver on Rochester's unix site.

PR #856, opened 2018-06-15
Commit additions to warnings to make setting up an LTI connection a l…

mgage:
…ittle easier.

Don't pull this request yet.  It is meant to solve some of the issues mentioned in PR #819. 
I've added a lot of debugging code and then prettied it up so that it is somewhat usable for 
configuring LTI grade passback.  

What it shows at the moment is that the mass_updating routine which runs in post processing will update some user sets but not others. In particular it always seems to update the most recent set submitted but fails with a "tool invalid" error message for the other sets.  So far I haven't been able to detect any difference between the requests made from the submit button action and the requests made from the post processing phase.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -10,6 +10,9 @@
 # debugging.  This is useful when setting things up for the first time because
 # different LMS systems have different parameters
 $debug_lti_parameters = 0; 
+$debug_lti_grade_passback = 0;
+#FIXME -- these parameters should be merged with the debug mechanism in 
+# WeBWorK::Debug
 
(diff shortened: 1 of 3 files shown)


mgage:
I've brought this 'hot fix' pull request up to date with the pull request #857 submitted to develop.
Since develop will be released fairly soon as 2.14 it may not be necessary to execute this pull request to backport the documentation changes to 2.13.  The develop pull request has had more testing.

mgage:
Since develop will be released soon as ww-2.14 I don't think this hotfix is necessary any longer.

mgage:
I've closed this in favor of PR #857

PR #857, opened 2018-06-21
Adjust warning for testing grade passback to the LMS.

mgage:
This is a parallel PR to #856. This provides an improved messaging system for determining whether grades are being updated. This is particularly tricky in the post processing mass update phase where all of the sets in a course are checked against the grades recorded by the LMS. It is easy for the mass update process to fail silently.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -6,10 +6,10 @@
 # To enable this file, uncomment the appropriate lines in localOverrides.conf
 #############################################################################
 
-# set debug_lti_parameters  to 1 to have LTI calling parameters printed to HTML page for
+# Set debug_lti_parameters  to 1 to have LTI calling parameters printed to HTML page for
 # debugging.  This is useful when setting things up for the first time because
 # different LMS systems have different parameters
-$debug_lti_parameters = 0; 
+$debug_lti_parameters = 1; 
 
 # To get more information on passing grades back to the LMS enmass set debug_lti_grade_passback
 # to one.  And set the LTIMassUpdateInterval to 60 (seconds).
@@ -48,8 +48,9 @@ $authen{user_module} = [
 # Basic_TheLastOption, then WeBWorK::Login
 # will display a screen directing the user
 # back to an external authentication system.
+# This prevents you from attempting to login in to WeBWorK directly.
 
-$external_auth=1;
+#$external_auth=1;
 
 # NOTE:  If external authentication is disabled
 # then you should probably also prevent students
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -6,10 +6,28 @@
 # To enable this file, uncomment the appropriate lines in localOverrides.conf
 #############################################################################
 
-# set this to 1 to have LTI calling parameters printed to HTML page for
+# set debug_lti_parameters  to 1 to have LTI calling parameters printed to HTML page for
 # debugging.  This is useful when setting things up for the first time because
 # different LMS systems have different parameters
 $debug_lti_parameters = 0; 
+
+# To get more information on passing grades back to the LMS enmass set debug_lti_grade_passback
+# to one.  And set the LTIMassUpdateInterval to 60 (seconds).
+$debug_lti_grade_passback = 0;
+
+# This will print into the apache log the success or failure of updating each user/set.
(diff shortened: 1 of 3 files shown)


mgage:
I've added a lot of debugging code and then prettied it up so that it is  usable for
configuring LTI grade passback.

The current state is that the mass_updating routine, which runs in post processing, will update some user sets but not others. In particular it always seems to update the most recent set submitted but fails with a "tool invalid" error message for the other sets. So far I haven't been able to detect any difference between the requests made from the submit button action and the requests made from the post processing phase so it must be some behavior due to Canvas's state.

The last commits in this PR insert options for setting the debug switches in the various conf files along with explanations of how they work.  I've tested this develop version fairly extensively with Canvas. 

The "tool invalid" error message seems only to occur when the score on Canvas does not need updating, so perhaps the behavior is what is expected.  Despite the worrisome error message the scores on Canvas and the scores in WeBWorK seem to stay in sync.

@dsteinmo Derek could you test whether this grade passback works with D2L, Brightspace?

Testing by anyone else using LTI grade passback will be appreciated.

The mode I was working with was to assign individual homework sets ("homework mode") and enable grade passback.  To allow you to view the mass_update phenomenon easily set the LTIMassUpdateInterval to something short, such as 60 seconds and watch for messages in the server error logs. 

This PR provides the instrumentation to help resolve the issue addressed in PR #819 and first raised in issue #811.  I don't believe that the fix proposed in #819 resolves the real issue.  The analysis of Goehle seems right to me.  I did find that the mass_update process was silently failing when any warning messages were printed. (The CGI::escapeHTML() subroutine was missing the $r parameter which was no longer available.) Since the mass_update runs after the HTML page is returned and prints only to the log file this extra escape doesn't seem necessary so I have removed it, at least for now. 

I'm closing the PR #819.  If further testing reveals that we still have a problem with grade_passback we should have additional information provided by this testing package which will make it easier to find a permanent fix.

dsteinmo:
@mgage will try to set up a test scenario with D2L this week.

mgage:
At least some of it was deliberate. 
1. I think that LTI debugging should be on by default.  It is pretty essential to the initial set up of LTI.  Perhaps add a message that this is just useful setup information and say where to turn it off.
2. I agree that the suggestions in localOverrrides.conf.dist such as Email WeBWorK TA should all be commented out. 
3. I see your point about having the debug and other LTI switches in just one place. On the other hand it's also good to reduce the number of .conf files you need to adjust and the authentication_LTI_conf file often gets overlooked, further the localOverrides.conf file is the place to override other configuration settings -- such as the ones in authen_LTI.conf.  I believe my intent was to alert people to configurations in authen_LTI.conf that are most likely to
need changes.  Many of the configurations in authen_LTI.conf rarely need to be touched.

dlglin:
I don’t feel strongly about the default LTI debugging status.  On one hand, someone who is setting up LTI for the first time will almost certainly benefit from the debugging. On the other hand, someone who is porting LTI configuration from another system may not expect the debugging info to show up.  We do have to choose one place for this setting, as it is currently set in both localOverrides.conf and authen_LTI.conf.

The original philosophy behind having a separate config file for LTI was so that people who weren’t using LTI wouldn’t see all of the additional stuff in the main config file.  There is a comment in localOverrides.conf to look at authen_LTI.conf when enabling LTI.  The one thing that I have seen cause problems being in that file is setting $authen{user_module}, particularly if allowing multiple authentication methods, as it is also set in authen_ldap.conf for example.

Based on this conversation, we need to look at a minor refactor of the configuration files around authentication.  Anything that rarely needs to be touched could go into defaults.config.  Then we would have to decide how to best handle the remaining options.

On Aug 30, 2019, at 11:20 AM, Michael Gage <notifications@github.com<mailto:notifications@github.com>> wrote:


At least some of it was deliberate.

  1.  I think that LTI debugging should be on by default. It is pretty essential to the initial set up of LTI. Perhaps add a message that this is just useful setup information and say where to turn it off.
  2.  I agree that the suggestions in localOverrrides.conf.dist such as Email WeBWorK TA should all be commented out.
  3.  I see your point about having the debug and other LTI switches in just one place. On the other hand it's also good to reduce the number of .conf files you need to adjust and the authentication_LTI_conf file often gets overlooked, further the localOverrides.conf file is the place to override other configuration settings -- such as the ones in authen_LTI.conf. I believe my intent was to alert people to configurations in authen_LTI.conf that are most likely to
need changes. Many of the configurations in authen_LTI.conf rarely need to be touched.

PR #858, opened 2018-07-04
Minor modifications to config files

mgage:
These modifications make it easier (or at least more obvious) how to set up buttons in the library browser that directly access the directory trees -- e.g. "Contrib" in addition to "OPL directory"
and "CAPA".  Direct links to local libraries can be added in this way.


mgage:
This also adds a commented out line that shows how to run R within a docker container. 
(The usual url pointing to localhost won't work in that case.)

PR #859, opened 2018-07-04
Move test files into appropriate directories

mgage:
This pull request replaces part of PR #844.  It moves some of the files in cllent to subdirectories  and removes some other files that do not seem to have been used in some time.


mgage:
This only consists of moving files around and deleting RQP.pm

To test:
Check that the files have been moved to appropriate places. Make sure that you don't need RQP.pm for anything.

mgage:
This only involves changes in the webwork2/clients subdirectory -- so it seems fairly safe to pull.

PR #860, opened 2018-07-04
Develop send xmlrpc patch

mgage:
Test with 

cd webwork2/clients

sendXMLRPC.pl -b input.pg    (or t/input.pg)
sendXMLRPC.pl --tex input.pg
sendXMLRPC.pl --pdf input.pg 
and
sendXMLRPC.pl -bB input.pg

If you are not using a mac you may need to adjust the credentials file and some display constants
near the top of the file first. 

sendXMLRPC.pl -vb input.pg

The -v triggers the "verbose" mode which will give you extra error messages, including a trace of where it is looking for the credentials file. (It is sometimes a hurdle to get sendXMLRPC connected properly.)


mgage:
Thanks George. I'll look in to this.  I notice that the directories searched (in @inc) are  /home/gjennings/webwork/webwork2/webwork2  etc.  What directory were you in when you 
ran sendXMLRPC.pl?  The webwork2/webwork2 segments look suspicious or do you have the directory structure set up with one webwork2 directory inside another?  

I'll go check that section where it tries to find the credentials file.  I'll be back shortly.

mgage:
It looks like you may be using an earlier version of sendXMLRPC.pl, the one included with develop, but not the one that is updated in this pull requests. (The error at line 335 makes sense for the earlier file, but not for the most recent version in this pull request.)  However the old file should also work (it just has fewer bells and whistles -- for example it won't process to tex and to pdf outputs).  

Here are two things I'd like you to try: 
(1)
sendXMLRPC -b --credentials=/home/gjennings/webwork/webwork2/conf/ww_credentials input.pg
(adjust the credentials path so that it points an actual file -- you can use the ww_credentials.dist file if you want but it's better to copy it over to ww_credentials first in case you want to modify it)

This technique gives an explicit path to the credentials file.

(2) 
sendXMLRPC.pl -bv input.pg

This adds more verbose error output. 

The list of directories normally searched for credentials is at line
290 in the old file and at line
        in the new version of sendXMLRPC.pl
It includes $ENV{HOME} which is supposed to point to an individual users home directory
which is where I create a .ww_credentials  file with the passwords i use for WeBWorK sites. I don't
usually want them to be in the public webwork directory. It's possible that $ENV{HOME} is not defined on your machine.

mgage:
The new version of sendXMLRPC.pl has better error messages while searching for the credentials file if you use the -v switch.  It should help determine what is going wrong.

gajennings:
I found the new version -- please ignore the message I sent earlier today.  With the new version I get these outputs on a Ubuntu server:
$ ./sendXMLRPC.pl -b input.pg
./sendXMLRPC.pl: line 7: syntax error near unexpected token `newline'
./sendXMLRPC.pl: line 7: `<!DOCTYPE html>'

The other commands "sendXMLRPC.pl --tex input.pg", "sendXMLRPC.pl --pdf input.pg", "sendXMLRPC.pl -bB input.pg"  produce the same output

mgage:
Is there some hidden character on line 7 of sendXMLRPC.pl ?  For the file in the PR there are only comment lines anywhere near line 7 of sendXMLRPC.pl.  It's possible this is referring to line 7 of the credentials file, but we first need to figure out which one. 

Try ./sendXMLRPC -vb input.pg which should  tell you which credential file is being used.

mgage:
I see the following if I type command into the docker container running webwork on ubuntu:

 ./sendXMLRPC.pl -vb input.pg
Use of uninitialized value in concatenation (.) or string at ./sendXMLRPC.pl line 390.
looking for credentials file /root/.ww_credentials. Found =
Use of uninitialized value in concatenation (.) or string at ./sendXMLRPC.pl line 390.
looking for credentials file /root/ww_session_credentials. Found =
Use of uninitialized value in concatenation (.) or string at ./sendXMLRPC.pl line 390.
looking for credentials file ww_credentials. Found =
looking for credentials file ww_credentials.dist. Found =1
Credentials taken from file ww_credentials.dist
ww_display_mode =>images
site_url =>https://demo.webwork.rochester.edu
html_display_command =>open -a 'Google Chrome'
userID =>daemon
course_password =>daemon
courseID =>daemon_course
site_password =>xmlrpc
form_action_url =>https://demo.webwork.rochester.edu/webwork2/html2xml
sh: 1: open: not found
DONE --  --

The last comment about "open: not found" is because the html_display_command is set for a mac, not for ubuntu.  You'll need to replace it with some kind of pipe command which takes the output of  sendXMLRPC.pl and pipes it into chrome or firefox. Unfortunately I haven't set up firefox in my docker box so I can't test this immediately.

./sendXMLRPC -h input.pg   will output the raw html data (and much more) which can be piped to the browser.

gajennings:
Adding the -v switch doesn't change the output:

$ ./sendXMLRPC.pl -vb input.pg
./sendXMLRPC.pl: line 7: syntax error near unexpected token `newline'
./sendXMLRPC.pl: line 7: `<!DOCTYPE html>'

Running perl explicitly produces different error messages:

$ perl sendXMLRPC.pl -vb input.pg
syntax error at sendXMLRPC.pl line 10, near "<"
Excessively long <> operator at sendXMLRPC.pl line 21.

I get the same errors on my Ubuntu server and on my Debian server.
Perhaps I'm using the wrong sendXMLRPC.pl file?  I obtained the one I'm using by running
wget https://github.com/mgage/webwork2/blob/develop_sendXMLRPC_patch/clients/sendXMLRPC.pl

mgage:
Hi George,
To download the single file without HTML markup you can use the command:
wget https://github.com/mgage/webwork2/raw/develop_sendXMLRPC_patch/clients/sendXMLRPC.pl

notice that "blob" has been replaced by "raw".  If you are using a browser instead of wget there is a "raw" button that allows you to download the pure file.  
(I consulted these pages for references: https://stackoverflow.com/questions/4604663/download-single-files-from-github
and https://webapps.stackexchange.com/questions/87954/how-do-i-download-a-particular-file-from-github )

For most pull requests it is better to create a new branch using git to insure that you get all of the changes in all of the files.  Something similar to:
```
cd webwork2/clients
git branch
   — master and develop should be shown — one of them highlighted
git checkout -b develop_test_sendXMLRPC origin/develop    (get a fresh copy of develop)
git branch
   — switched to develop_test_sendXMLRPC  (just a duplicate of develop at the moment)

git pull https://github.com/mgage/webwork2 develop_sendXMLRPC_patch (this adds all the changes from the new pull request) 

Restart webserver just to be sure everything is updated.

Test.

When you are done you can do this:

git checkout develop   (puts you back on the develop branch)
git branch -D develop_test_sendXMLRPC  (deletes the test branch you created)
```

gajennings:
Thanks for the github instructions.  The script seems to work now.  In the credentials file I changed the userID, course_password, and courseID to correspond one of my webwork courses, changed the html_display_command to "chromium" (my browser), moved the file to my home directory and renamed it .ww_credentials.  Then running
$ ./sendXMLRPC.pl -vb t/input.pg
opened a web page with a webwork tutorial problem on it.  It grades the answers correctly.

clients/sendXMLRPC.pl
```diff
@@ -444,15 +550,35 @@ sub process_pg_file {
...
+	    	process_problem($file_path, $default_input, $form_data2);
+	    	
+	    # create tex file for both and tex and pdf output
+	    my $tex_file_name = create_tex_output($file_path, $formatter);
+	    # display tex file if --tex option is set 
+	    if ($display_tex_output) {	    
+	    	system($TEX_DISPLAY_COMMAND." ".TEMPOUTPUTDIR().$tex_file_name);
+	    } elsif($display_pdf_output) { # process tex file to create pdf file and display if --pdf option
+	    	my $pdf_path = create_pdf_output($tex_file_name); 	    
+	    	system("open -a Preview ".$pdf_path );
```
heiderich: This seems to be quite specific for Mac OS X.

heiderich:
I made a copy `ww_credentials` of `ww_credentials.dists` with the following changes:

```
$ diff ww_credentials.dist ww_credentials
22c22
<         site_url        => 'http://localhost:80',
---
>         site_url        => 'https://demo.webwork.rochester.edu',
24c24
<         form_action_url => 'http://localhost:80/webwork2/html2xml',
---
>         form_action_url => 'https://demo.webwork.rochester.edu/webwork2/html2xml',
26c26
<         html_display_command => "open -a 'Google Chrome' ", # "open -a Safari ", "lynx "
---
>         html_display_command => "firefox", # "open -a Safari ", "lynx "

```

Then by executing `$ ./sendXMLRPC.pl -b input.pg --credentials ww_credentials` I obtain the following output:

```
Credentials taken from file ww_credentials
courseID =>daemon_course 
course_password =>daemon 
form_action_url =>https://demo.webwork.rochester.edu/webwork2/html2xml 
html_display_command =>firefox 
site_password =>xmlrpc 
site_url =>https://demo.webwork.rochester.edu 
userID =>daemon 
ww_display_mode =>images 

```

Then the program seems to hang. Debugging it, it seems that the problem is around

https://github.com/mgage/webwork2/blob/421c91e363cf617d9935c2dc19259104daa33818/clients/sendXMLRPC.pl#L573-L574

Inserting a debug print statement right before this line is outputted, but a debug print statement inside the subroutine `process_problem` is not executed. I do not know why.

clients/sendXMLRPC.pl
mgage: Good catch.  There is already a $PDF_DISPLAY_COMMAND -- just forgot to use it here.

mgage: For the first problem.  The most likely place for a hang would be at line 750 -- the xmlrpcCall
to server to have the problem rendered.  There might not be enough error messages if the server didn't respond. (Perhaps I should add a timeout if it doesn't respond in twenty seconds or so.)

The credentials you are using seem correct.  I'll update the ww_credentials.dist file to point to demo.webwork.rochester.edu in order to provide an "out of the box" experience.  Not everyone will be running a local server. 

There is a variable $UNIT_TESTS_ON =0 set near the top of the sendXMLRPC.pl file.
Setting $UNIT_TESTS_ON=1;  will provide additional error messages which should help.

For a sanity check -- I don't expect a printout of the credentials file unless you used 
./sendXMLRPC.pl -vb input.pg --credentials ww_credentials

it's the -v flag that produces verbose output about credentials.  
Probably that is what you typed.

mgage: One last thing.   I notice that demo.webwork.rochester.edu  has a performance issue. The first few times I sent a request it took  up to 20 seconds or so to respond.  After about 5 requests it responded almost immediately.  I suspect it is because demo is being hosted for me by our IT on a low priority virtual machine.  Once all of the apache server children are awake the response improves, but it is slow while the first few child processes are being created.

heiderich: Currently I cannot reproduce the "hanging" error anymore. Now I experience a different problem, which I have encountered earlier too. The script seems to try to connect to a MySQL server, which is not running on that host.

```
DBI connect('webwork','webworkWrite',...) failed: Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2 "No such file or directory") at /usr/lib/x86_64-linux-gnu/perl5/5.26/DBI.pm line 691.
```

Is it necessary for the script to access a MySQL server (with a WeBWorK database)? If so, why?

mgage:
The DBI connect error must be from the server you are connecting to.  Are you still pointing the 
client script to https://demo.webwork.rochester.edu ?  (It doesn't look like it since demo.webwork... doesn't have a /usr/lib/x86_64-linux-gnu/perl5/5.26/ directory.)  I usually get this kind of error when I forget to start the mysqld  daemon.  but there could be other reasons.  If you login to that webwork site from the web I would expect you to also get a database connection error message of some kind.

mgage:
Change the ww_display_mode to  ww_display_mode=>MathJax. 
The images mode uses the local webwork2 software and database to display images.

heiderich:
This solves the problem. I would recommend MathJax as a default then (not only for this reason).

mgage:
This looks good.  A very thorough job of reviewing. Thanks very much Tani

mgage:
I appreciate the extra comments and the reformatting of the credentials file.

taniwallach:
My "Docker install" still gave warnings with the final version of the code from this PR. 
 * Docker install still need an extra environment variable `WEBWORK_ROOT_URL` as the Docker version of `site.conf` needs it.
 * However, the approach of using `$credentials{webwork_url}` to handle the other warning did not work, as the credentials file was read too late in the startup process.

https://github.com/openwebwork/webwork2/pull/869 reorders the code of `sendXMLRPC.pl` to overcome that issue.

PR #861, opened 2018-07-09
Add snippet to localOverrides.conf.dist to check whether moodle has a…

mgage:
…lready installed MathJax (thanks Davide)

New MathJax address possibilities have been added.

The CAPA library links have been enabled, now that the libraries themselves are available on 
nearly every site.


mgage:
This should fix the problem with MathJax being called twice in some cases using the opaque client problems.

PR #866, opened 2018-07-21
Rel ww2.14 trim white space

mgage:
Trim leading and trailing white space from user_id's and passwords.


mgage:
Andrew Parker checked this on mobile devices where it is most useful. (He caught one typo -- thanks Andrew)

PR #867, opened 2018-07-23
Develop international

mgage:
After it stabilizes develop_international will be pulled into develop.  Don't pull this yet.


mgage:
This is incorporated into #893

PR #872, opened 2018-08-01
Wiriseditoranswer

joaquimwiris:


mgage:
This worked for me on my iMac desktop.   I found some issues with using the handwriting feature on an Amazon Fire tablet.  The html popup template would load the first time an answer was entered, but would come up blank if one tried to change the entry or enter another answer on the same page. 

This feature is independent of the rest of webwork so even if further glitches appear the feature can  be turned off.  Including the updates in this release 2.14 will provide an opportunity for further testing for those who wish to try it.

mgage:
Thanks @manuelwiris and @joaquimwiris . This feature is now merged into rel-ww2.14

PR #876, opened 2018-08-16
Tani multi library v01 - improved beta version

taniwallach:
This is an improved beta version of multi-library support for the library search tools (`SetMaker.pm`) replacing https://github.com/openwebwork/webwork2/pull/874.

 * The library configuration is more complicated than before, so that multiple libraries can be supported. 
 * Each library requires several settings. Several files which read library settings were modified to use the new library config structure.
 * Sample settings are in `conf/localOverrides.conf.dist` 
 * The modified OPL-update script expects the library "code name" as in the config settings, and the second library and on should set the `noDrop` option (to prevent deletion of the database tables shared by all the libraries).
 * The sample config file assumes that the `course/templates` directory will have symlinks to the libraries.
 * Sample test libraries are available at https://github.com/taniwallach/ww-test-libs

Example setup using the test libraries (including OPL update on 3 libraries):
```
# cd into the course directory
cd templates
ln -s /opt/webwork/libraries/tl-01  TestLibrary1
ln -s /opt/webwork/libraries/tl-02  TestLibrary2
ln -s /opt/webwork/libraries/webwork-open-problem-library/OpenProblemLibrary   Library
/opt/webwork/webwork2/bin/OPL-update local OPL
/opt/webwork/webwork2/bin/OPL-update local TSTL1 noDrop
/opt/webwork/webwork2/bin/OPL-update local TSTL2 noDrop
```

Known issue:
 * Do a search on `All Libraries` and it should show a correct count of problems.
 * Then change back to `All Subjects` (the count should update properly).
 * The click on `View Problems`
 * The correct set of problems should be displayed. 
 * *but* the count will be incorrect and the selections will have changes.
 * I currently do not understand how this is happening.

### DIFF
--- a/bin/OPL-update
+++ b/bin/OPL-update
@@ -22,6 +22,9 @@ use Cwd;
 use DBI;
 
 
+my $myLib = "OPL" ; # default value
+
+
  #(maximum varchar length is 255 for mysql version < 5.0.3.  
  #You can increase path length to  4096 for mysql > 5.0.3)
 
@@ -96,9 +99,9 @@ my $dbh = DBI->connect(
         },
 );
 
-my $libraryRoot = $ce->{problemLibrary}->{root};
+my $libraryRoot = $ce->{problemLibrary}->{$myLib}->{root};
 $libraryRoot =~ s|/+$||;
-my $libraryVersion = $ce->{problemLibrary}->{version};
+my $libraryVersion = $ce->{problemLibrary}->{$myLib}->{version};
 my $db_storage_engine = $ce->{problemLibrary_db}->{storage_engine};
 
 my $verbose = 0;
@@ -431,7 +434,7 @@ if(open(IN, "$libraryRoot/Textbooks")) {
 	close(IN);
 } else {
 	print "Textbooks file was not found in library $libraryRoot. If the path to the problem library doesn't seem
-	correct, make modifications in webwork2/conf/site.conf (\$problemLibrary{root}).  If that is correct then
+	correct, make modifications in webwork2/conf/localOverrides.conf (\$problemLibrary{OPL}{root}).  If that is correct then
 	updating from git should download the Textbooks file.\n";
 }
 #### End of textbooks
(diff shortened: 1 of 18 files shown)


taniwallach:
This code will apparently need some revision **before** it can be merged into a tree with UTF-8 support, or the merge conflicts will need to be carefully handled.

Once https://github.com/openwebwork/webwork2/pull/934 is merged into develop, I will try (as soon as I can) to prepare create a new PR which gets this code to cleanly merge onto the revised develop branch.

From what I recall, it basically works, and searching against several libraries which were processed by the revised `OPL-update` as explained above works. 

The code is **not complete**.

There are several things which are marked with **FIXME** in the code, and in particular there are places where things are still hardwired to use default values (ex. a sub-path starting with `Library/`) appropriate to the OPL, and which may need to be adjusted to work with different libraries.

In particular more work is needed on:
  * `get_library_sets()` in `lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm`
  * `getDBextras()`  in `lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm`
  * `deleteListing()` in `lib/WeBWorK/Utils/ListingDB.pm`
  * `getProblemDirectories()` in `lib/WebworkWebservice/LibraryActions.pm`
but there are almost certainly other parts of the code which require additional work.

Testing of this code by additional people would help.

bin/OPL-update
```diff
@@ -21,6 +21,32 @@ use File::Basename;
 use Cwd;
 use DBI;
 
+# Command line arguments
+# for some reason the first command line argument is disappearing before it gets here
```
eltenedor, line 25: this is caused as `use OPLUtils` seems to cause a shift in the argument, see this excerpt from `OPLUtils.pm` and my comment there

image

bin/OPLUtils.pm
```diff
@@ -25,6 +27,10 @@ use JSON;
 our @EXPORT    = ();
 our @EXPORT_OK = qw(build_library_directory_tree build_library_subject_tree build_library_textbook_tree);
 
+# what library are we handling, now is first argument of the function
+my $myLib = shift;
```
eltenedor, line 31: I suspect, that this line is not needed.

dlglin:
Closed due to inactivity.  Can be reopened if activity resumes.

PR #865, opened 2018-07-20
Rel ww2.14

mgage:
Do not pull this into master yet.  It gives a list of changes between rel-ww2.13 and rel-ww2.14

PR #877, opened 2018-08-16
Hardwire the url for the WeBWorK and PG branches being used.

mgage:
When calling MathJax include the full site address. (Appears to be needed if the port is 8080 instead of 80). (in defaults.config)

The method of guessing the PG branch from reading the webwork VERSION file is fragile.
Until we establish a firm naming scheme for branches
it is safer to require that the WeBWorK and PG branches
being used are explicitly specified in the docker file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,10 @@
 FROM ubuntu:16.04
 
+
+ENV WEBWORK_BRANCH WeBWorK-2.14
+ENV PG_BRANCH rel-PG-2.14
 ENV WEBWORK_URL /webwork2
-ENV WEBWORK_ROOT_URL http://localhost
+ENV WEBWORK_ROOT_URL http://localhost:8080
 ENV WEBWORK_DB_HOST db
 ENV WEBWORK_DB_PORT 3306
 ENV WEBWORK_DB_NAME webwork
@@ -80,11 +83,11 @@ RUN mkdir -p $APP_ROOT/courses $APP_ROOT/libraries $APP_ROOT/webwork2
 
 COPY VERSION /tmp
 
-RUN WEBWORK_VERSION=`cat /tmp/VERSION|sed -n 's/.*\(develop\)'\'';/\1/p' && cat /tmp/VERSION|sed -n 's/.*\([0-9]\.[0-9]*\)'\'';/PG\-\1/p'` \
-    && curl -fSL https://github.com/openwebwork/pg/archive/${WEBWORK_VERSION}.tar.gz -o /tmp/${WEBWORK_VERSION}.tar.gz \
-    && tar xzf /tmp/${WEBWORK_VERSION}.tar.gz \
-    && mv pg-${WEBWORK_VERSION} $APP_ROOT/pg \
-    && rm /tmp/${WEBWORK_VERSION}.tar.gz \
+    
+RUN curl -fSL https://github.com/openwebwork/pg/archive/${PG_BRANCH}.tar.gz -o /tmp/${PG_BRANCH}.tar.gz \
+    && tar xzf /tmp/${PG_BRANCH}.tar.gz \
+    && mv pg-${PG_BRANCH} $APP_ROOT/pg \
+    && rm /tmp/${PG_BRANCH}.tar.gz \
     && curl -fSL https://github.com/openwebwork/webwork-open-problem-library/archive/master.tar.gz -o /tmp/opl.tar.gz \
     && tar xzf /tmp/opl.tar.gz \
     && mv webwork-open-problem-library-master $APP_ROOT/libraries/webwork-open-problem-library \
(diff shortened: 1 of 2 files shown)


mgage:
I can fix the url, for now at least, by being less clever about constructing the webwork and pg branches being pulled in by docker.  In this pull request I specify them directly.  I also add 8080
to the WEBWORK_ROOT_URL and modify the call to mathjax in default.config so that the 8080 port is explicitly used when mathjax calls back to the server. 

This may not be the most elegant fix but it does allow docker to get up and running.  The problem with trying to deduce that webwork and pg branch from VERSION is that up to this point at least the branch names and the version names have not been tightly controlled.  In particular I used rel-PG-2.14 instead of PG-2.14 for the branch name for PG.

I also rearranged the Dockerfile slightly splitting off the compiling of color.c into a separate
RUN line since it generates some errors.

I'm still seeing one difficulty however.  The modelCourse directory is not being copied from webwork2/courses.dist into courses/  .  


Perhaps @xcompass you can see what is going wrong.

mgage:
@taniwallach @nmoller   -- this was my version of the fix that @nmoller submitted in #875 and then closed.  Why did you close it Nelson?  I think it is a reasonable fix for the 2.14 version problem.

In my version of the fix above I think I was already having trouble with my docker build because I seemed to have to hardwire the port number in order to get mathjax to work.  That hasn't appeared necessary when I followed Tani's instructions as in #865

The other problem that remains is that this segment:
```
RUN cd $APP_ROOT/webwork2/courses.dist \
    && cp *.lst $APP_ROOT/courses/ \
    && cp -R modelCourse $APP_ROOT/courses/ \
    && cd $APP_ROOT/pg/lib/chromatic \
    && gcc color.c -o color
```
Doesn't appear to copy a modelCourse directory into courses.

does anyone understand why not?

taniwallach:
@mgage  - I'm making an semi-educated guess, but did not have a chance to test my idea to overcome the problem.

 * The courses directory is "bind mounted" when the machines are run using `docker-compose up` but as far as I can tell, this is not done when the images are created. 
 * Not only is there is no line in the Dockerfile to do this bind mount during the image building process, it is apparently not possible. See:  [this thread on StackOverFlow](https://stackoverflow.com/questions/47942016/add-bind-mount-to-dockerfile-just-like-volume)
 * It seems the solution would be to move these steps into `docker-entrypoint.sh` which runs when the container is started with these directories available.

On Windows, I needed to move the data directory to a named volume to get Docker to startup the MariaDB container properly. I think that it may also be a reasonable idea to move the courses data into a named volume, rather than using a "bind mount". That would achieve data persistence, but avoid the issues of Linux vs. Windows file-systems which seem to make trouble with Docker on Windows. However, I think 

To allow developers to use their code in the Docker container, we still need the webwork2 directory to be bind mounted into the containers.

@xcompass - what do you think about SQL data, courses data being in named volumes?

References:
 * [Official docs on volumes](https://docs.docker.com/storage/volumes/)
 * [Named volumes - brief explanation](https://madcoda.com/2016/03/docker-named-volume-explained/
)
 * [Named volumes - longer explanation](https://boxboat.com/2016/06/18/docker-data-containers-and-named-volumes/
)
 * [more technical material](https://container-solutions.com/understanding-volumes-docker/)
 * [more](https://sandro-keil.de/blog/2016/05/02/docker-compose-with-named-volumes-and-multiple-networks/)
 * [GitHub on named volumes in Dockerfile](https://github.com/moby/moby/issues/30647)
 * [Stackoverflow - no named volumes in Dockerfile](https://stackoverflow.com/questions/38298473/how-to-declare-a-named-volume-in-a-dockerfile)

taniwallach:
## branch names:
* I adopted the hard-coded release names into the Dockerfile available from the `WW-2.14-ubuntu_16.04` subdirectory of <https://github.com/taniwallach/ww_docker_config_sets>
* That set also replaces `PG_VERSION` with a static file, to avoid issues when it is a symbolic link.

## modelCourse and OPL-update for Docker containers:

* I found a solution to the courses/modelCourse directory issue, by doing the copy command in docker-entrypoint.sh instead of in the Dockerfile, as at run-time the bind mount courses directory is available to Docker. 
* The same sort of approach was also used to trigger an OPL-update at container start (`docker-compose up`) when needed. 

## The MathJax issues and issues with the proposed fix:
* When I tried with the Dockerfile setting you suggested to overcome the MathJax problem, I ran into other problems. When
```
ENV WEBWORK_ROOT_URL http://localhost:8080
```
was set, I ran into problems with the Ajax queries from the library browser. Removing the ":8080" port setting fixed that problem, 
* As a result, I believe an alternate fix to the MathJax issue is needed. 
* I think the necessary change is in the MathJax setting in `conf/localOverrides.conf` which is derived from `conf/localOverrides.conf.dist`.  The change I recommend testing is **removing** the `$server_root_url/` from the start of the value set, so that the value will start with `$webworkURLs{htdocs}/mathjax/` just as it does in `conf/defaults.conf` **before** the change proposed in this pull request.

mgage:
This pull request is being closed in favor of #882 .  The comments here may still be useful in considering #882

PR #879, opened 2018-08-30
2.14 lti hotfix

mgage:
checks to prevent webwork from crashing while accessing the homepage when LTI is enabled at the site level.

The homepage does not provide a course or a database since no course has been selected yet.


mgage:
submitted to wrong branch

Issue #878, opened 2018-08-22
Errors relating to LTI after merging 2.14


Alex-Jordan:
I merged the 2.14 release candidate branch into my production branch, and trying to visit any WeBWorK page gives this warning and error:

```
Warning messages

    database reference is not defined at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm line 50.

Error messages

    Could not create WeBWorK::Authen::LTIAdvanced::SubmitGrade object, missing items from request at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm line 174. 
```

Is there anyone with a more firm understanding of the LTI modules (@dsteinmo ?) that could shed light on this? At this point I am unsure if it is an issue that has anything to do with the particulars of my setup (some customization done to my webwork2 and pg repos, and we are using Desire2Learn).

dlglin:
The error is happening in the code that handles the scheduled LTI pass back of grades to the LMS.  You should be able to temporarily get past this by turning off grade pass back: comment out $LTIGradeMode in authentication_LTI.conf.

I’m working on getting 2.14 on my production server in the next day.  I’ll see if I can reproduce the issue.

On Aug 22, 2018, at 3:17 PM, Alex Jordan <notifications@github.com<mailto:notifications@github.com>> wrote:


I merged the 2.14 release candidate branch into my production branch, and trying to visit any WeBWorK page gives this warning and error:

Warning messages

    database reference is not defined at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm line 50.

Error messages

    Could not create WeBWorK::Authen::LTIAdvanced::SubmitGrade object, missing items from request at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm line 174.


Is there anyone with a more firm understanding of the LTI modules (@dsteinmo<https://github.com/dsteinmo> ?) that could shed light on this? At this point I am unsure if it is an issue that has anything to do with the particulars of my setup (some customization done to my webwork2 and pg repos, and we are using Desire2Learn).

mgage:
The database object $db is not being passed in to the SubmitGrade object constructor.  
Perhaps a configuration problem?  I'd follow Danny's suggestion, perhaps even comment out 
the include() statement in localOverides.conf
[...]

PR #880, opened 2018-08-30
Hot fix to prevent  webwork from crashing when landing on the homepag…

mgage:
checks to prevent webwork from crashing while accessing the homepage when LTI is enabled at the site level.

The homepage does not provide a course or a database since no course has been selected yet.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
@@ -45,7 +45,7 @@ sub new {
   # sanity check
   my $ce = $r->{ce};
   my $db = $self->{r}->{db};
-  unless (ref($ce//'') and ref($db)//'') {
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
The only thing about this is I'm not sure how to engineer a test that this is still allowing grade passback at all. That is, can we test to be sure that 
```
if ($ce->{LTIGradeMode} and ref($r->{db}//'')  )
```
isn't simply always evaluating to false? Always evaluating to false would also fix the croak issue, since the code block that causes the croak would just never get executed.

Alex-Jordan:
Well, I _can_ imagine how to engineer such a test. I just need to put aside some time since I don't normally use our LMS for my own courses, and it may take some doing.

mgage:
This appears to fix issue #878. At least it prevents the error message.  We'll have to wait and see if 
the grade passback is working correctly in all cases.

PR #881, opened 2018-09-03
Modify MathJax config setting in conf/localOverrides.conf.dist

taniwallach:
Modify MathJax config setting in conf/localOverrides.conf.dist to drop $server_root_url from the start of the value, as that makes problems when used inside a Docker container running on port 8080.


mgage:
The $site_root_url prefix was added when trying to fix a MathJax issue involving the OpaqueServer which supplies individual questions to Moodle quizzes.  There were some problems with wwassignment and mathjax as well.  I'm not sure that adding the $site_root_url prefix was actually required in those fixes.  If it was we'll need to find a more robust method of accommodating the case when the default site supplied by the browser doesn't match the site where the MathJax repository is found.

PR #882, opened 2018-09-03
Updated docker config files for rel-ww2.14

taniwallach:
Updated docker config files from <https://github.com/taniwallach/ww_docker_config_sets/tree/master/WW-2.14-ubuntu_16.04>  which overcome several issues. 
Summary of changes:
  * docker-compose.yml: MariaDB on Dockers CE for Windows cannot function with bind mounted /var/lib/mysql at least with the default settings. Using a named volume for the persistent storage instead.
  * `Dockerfile`: 
    * Hardwired the PG branch based on proposal from @mgage in PR https://github.com/openwebwork/webwork2/pull/877
    * Merge `ENV` lines and various other fixes proposed by @nmoller in PR https://github.com/openwebwork/webwork2/pull/865
    * Some additional reorganization
  * `docker-entrypoint.sh`: 
    * Create modelCourse directory
    * Automatically run OPL-update on initial container startup
    * Misc additional start-up time actions
  * `.dockerignore`: Add some additional files to 
  * `PG_VERSION` 
    * As a file instead of a symbolic link.
    * Using the symbolic link could cause trouble/confusion in Docker installs.


mgage:
This has worked for me on macos.  

My first attempts ran into some errors where it couldn't chown  some files in htdocs/tmp  but I believe this was because I didn't start the build completely from scratch.  

In addition to persistent files in .data which might cause trouble with a rebuild there are persistent files at least in webwork2/htdocs/tmp and possibly in DATA or some other subdirectories of webwork2.

Rebuilding from a fresh install of rel-ww2.14 plus merging request #882 worked fine.

taniwallach:
Arnie had some issues related to leftovers from prior installs. 
I am going to try (later this week) to see it moving the persisted courses out of the webwork2 directory can make it easy to keep courses from a prior install even if a clean webwork2 tree is installed.
If someone has a chance to experiment with this sooner, just make a data directory parallel to webwork2/ and pg/ and move webwork2/.data/courses to there, and edit the docker-compose.yml file to mount from the new location. (I'm not sure if a "docker-compose build" will be needed to facilitate this change.)

Alex-Jordan:
I got some warnings partway through. Reporting here:

```
Step 9/16 : RUN cd $APP_ROOT/pg/lib/chromatic     && gcc color.c -o color
 ---> Running in d433e8080bdf
color.c:271:1: warning: return type defaults to 'int' [-Wimplicit-int]
 AssignColor(node,color)
 ^
color.c:294:1: warning: return type defaults to 'int' [-Wimplicit-int]
 RemoveColor(node,color)
 ^
color.c: In function 'color':
color.c:381:5: warning: type of 'current_color' defaults to 'int' [-Wimplicit-int]
 int color(i,current_color)
     ^
color.c:433:6: warning: implicit declaration of function 'print_colors' [-Wimplicit-function-declaration]
      print_colors();
      ^
color.c: At top level:
color.c:458:1: warning: return type defaults to 'int' [-Wimplicit-int]
 print_colors() 
 ^
color.c:478:1: warning: return type defaults to 'int' [-Wimplicit-int]
 main(argc,argv)
 ^
```

Alex-Jordan:
Another warning later on:

```
Step 16/16 : CMD ["apache2", "-DFOREGROUND"]
 ---> Running in c9ee76bfee28
Removing intermediate container c9ee76bfee28
 ---> 98fb83ee3b84
Successfully built 98fb83ee3b84
Successfully tagged webwork:latest
WARNING: Image for service app was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
```

Alex-Jordan:
After the OPL update (part of Tani's additons here) I get this:

```
app_1  | mysql: [Warning] Using a password on the command line interface can be insecure.
app_1  | 
app_1  | Done.
app_1  | chown: changing ownership of '../courses/modelCourse/templates/CAPA': No such file or directory
app_1  | chown: changing ownership of '../courses/modelCourse/templates/Contrib': No such file or directory
webwork2_app_1 exited with code 1
```

and it seems to be just hanging. In another terminal I looked and there is no courses folder. (I'm not sure if there is supposed to be one...not sure if the complaint here is in some ephemeral place that gets printed to my drive later or if it's supposed to be looking for a courses folder within my `webwork_docker` folder.)

mgage:
HI Alex,

Look in the directory  webwork2/.data  (use ls)  You should see pg, libraries, courses and inside courses
there should be admin and modelCourses.

The warning is because it is trying to change an alias in modelCourse/templates to CAPA and to Contrib 
and those aliases might not have been set up yet. 

As far as I can tell changing the owner of an alias never changes anything anyway — only the 
ownership and permissions of the target file/directory matter.

Does your docker work?  Can you read it from http://localhost:8080/webwork2? <http://localhost:8080/webwork2?>

Take care,

Mike

Alex-Jordan:
>Does your docker work? Can you read it from http://localhost:8080/webwork2? ><http://localhost:8080/webwork2?>

No, that's not working. Web browser shows "Unable to connect".

I don't feel like the the `docker-compose up` finished. That terminal is still just sitting there at "webwork2_app_1 exited with code 1" and not providing me with a prompt.

I would think that the local web server bit should be working prior to the OPL Update step though, so that may be orthogonal to me not connecting to localhost.

I'll try starting over later tonight.

Alex-Jordan:
Despite that nice-looking output from `docker-compose up`, when I do `docker ps` I am only seeing two containers:
```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                    NAMES
839f84535cf5        mariadb:10.1        "docker-entrypoint.s…"   2 minutes ago       Up 2 minutes        3306/tcp                 webwork2_db_1
5c1f37bfde5f        ubcctlt/rserve      "R -e 'Rserve::run.R…"   2 minutes ago       Up 2 minutes        0.0.0.0:6311->6311/tcp   webwork2_r_1
syst203m4138:webwork2 alex.jordan$
```
This makes me wonder again if something went wrong with spinning up the WeBWorK app. I see now that it exited with an error code 1, where 0 is what you hope for.

Alex-Jordan:
I started over from scratch, this time with tani's branch right away. Same outcome ("unable to connect" in all web browsers when going to http://localhost:8080/webwork2). 

Since I don't really know how Docker operates, I wondered if maybe in the past I had done something to this laptop that might be a problem. So I tried from scratch using my wife's laptop. It's the same as mine, except it's never been used for WBWK development, etc. Anyway, same outcome there too.

We are on El Capitan, and I think Mike tested with (High?) Sierra. I wonder if that could be a problem.

Is it normal that after I do `docker-compose up` and then do `docker ps` that I only see two containers?

taniwallach:
Additional comments:
* The OPL-update and all of the rest of what `docker-entrypoint.sh` does is run before the Apache server is started.
* I am assuming that the startup of `webwork2_app_1` hit the same error, so the container is not really in a running state - so it was "created" but is not "running" what it should be running.  
* `docker ps -a` will also show containers which are not properly running. 
* Regarding the contents of the `.data` directory:
    * I think it should **only** have a `courses` subdirectory, and maybe also a `db` subdirectory, if using the older `docker-compose.yml` which uses a bind mount for `/var/lib/mysql` for the database container.
    * I don't think there should be pg or library directories directly in `.data/`.
* The second warning is meant to tell you that (a) the image was built and (b) that future calls to `docker-compose up` will not automatically rebuild the image, so if changes are made a build must be forced manually.
    * WARNING: Image for service app was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.

Alex-Jordan:
Success, thanks Nathan! For the record here, I followed the suggestion to add the symlink and also commented out the chown lines for good measure. Then after down, build, up, I can connect to localhost.

Earlier, when I followed the procedure from Mike's wiki post and without Nathan's additional commit, I saw more in the `.data` folder, as Mike described. Later, after following the procedure from Nathan with the extra commit included, `.data` only has a courses folder, as Nathan describes.

taniwallach:
@Alex-Jordan - I would like to determine if just fixing the missing symbolic link solves the problem without dropping the `chown` line from `docker-entrypoint.sh`. 

Would you mind testing as follows:
  1. `docker-compose down` (if Docker is running)
  2. Reenable the `chown` line in `docker-entrypoint.sh` again, and the do a new `docker-compose build` (so the updated file gets into the image).
  3. Change into the `webwork2/.data/courses/` directory (on the Mac from outside the Docker container) and run `mv admin admin.old` (this will force Docker to recreate the admin course on the next startup).
  4. Change into the `webwork2/courses.dist/modelCourse/templates` directory (on the Mac from outside the Docker container) and run 
```
ln -s /opt/webwork/libraries/webwork-open-problem-library/OpenProblemLibrary Library
```
to create the expected symbolic link in the source location. 
  4. `docker-compose up`

If this solves the problem, then all we really need to do is add this symbolic link into the main WW repository (for now in the rel2.14 branch) and then the Docker startup should go smoothly for everyone. If not, then we need to look into how to avoid having Docker treat a failure to `chown` an non-existing symbolic link target as a fatal start-up error.

Thanks

Alex-Jordan:
I followed steps 1-4, then did `docker-compose up`. But it's failing in the same way it failed earlier.

```
app_1  | Admin course is created.
app_1  | chown: changing ownership of '../courses/modelCourse/templates/CAPA': No such file or directory
app_1  | chown: changing ownership of '../courses/modelCourse/templates/Contrib': No such file or directory
webwork2_app_1 exited with code 1
```

Now when I look inside `webwork2/courses.dist/modelCourse/templates` and among other things I see:

```
lrwxr-xr-x   1 alex.jordan  staff    20 Sep 10 16:59 CAPA -> Library/Contrib/CAPA
lrwxr-xr-x   1 alex.jordan  staff    15 Sep 10 16:59 Contrib -> Library/Contrib
lrwxr-xr-x   1 alex.jordan  staff    70 Sep 14 14:30 Library -> /opt/webwork/libraries/webwork-open-problem-library/OpenProblemLibrary
```

Is that as it should be?

mgage:
Tani and Alex have worked on this pull request and seem to have it in decent shape. There is discussion (Arnie, Tani and others) about moving the docker code to a separate repos so that it can be updated more quickly. This seems like a good idea for the next release but for now I'll pull this update to the current configuration. 

-------------------------------------------------------------------------------------------
summary of note from Alex jordan:

Hi Alex,

I think this is probably good enough.  I’ll wait and see if Tani has come up with any issues and then 
pull the changes into rel-ww2.14.  Thanks for checking this.

Take care,

mike


On Oct 5, 2018, at 2:40 PM, Alex Jordan <alex.jordan@pcc.edu> wrote:

Now confirming that the change to the data directory worked out fine as well.

At this point, is there anything to gain from me setting the whole thing up from scratch on a "clean" system? There is a Mac desktop here where I could do that. (Well, probably. Unsure if I would run into permissions issues there.) And of course, there is my wife's laptop.

On Fri, Oct 5, 2018 at 11:33 AM Alex Jordan <alex.jordan@pcc.edu> wrote:
Phase 1 complete, and checks out. That is, I followed Nathan's first test:

Change into your webwork2/ directory where the Docker files are.
ls -l docker*
cp -a docker-entrypoint.sh docker-entrypoint.sh.BAK
Put in the attached file as docker-entrypoint.sh and make sure it has the correct permissions (like the old file had).
docker-compose build (will put the new file into the container image)
docker-compose up
If all goes well you should see:
app_1  | Fixing ownership and permissions (just in case it is needed)
without it being followed by an error message similar to
app_1  | chown: changing ownership of '../courses/modelCourse/templates/CAPA': No such file or directory
Make sure you can access http://localhost:8080/webwork2
docker-compose down

And it all works. To be clear, I did this on my laptop, not a fresh one. You two eased my concerns about interfering with the non-Docker WeBWorK.

I will move on to try more now. I may be called to a meeting in a bit, so I wanted to report.


-- 
Alex Jordan
Mathematics Instructor 
Portland Community College

taniwallach:
Pull request https://github.com/openwebwork/webwork2/pull/887 has the fix Alex Jordan tested which should solve the problem he encountered with "chown" failing and preventing the startup of the webwork docker container due to broken symbolic links.

Pull request https://github.com/openwebwork/webwork2/pull/888 does the relocation of the location of the persistent storage of the courses data.

PR #869, opened 2018-07-23
Reorder code in sendXMLRPC.pl to avoid warnings on systems using a "Docker install"

taniwallach:
On a "pure Docker" install, there are differences in `site.conf` which require setting an environment variable `WEBWORK_URL`, or providing an alternate. @mgage added a setting into the credentials file late in the changes for https://github.com/openwebwork/webwork2/pull/860 to handle this issue, but it did not help on my "pure Docker" install tree on my laptop, and warning messages were still issued.

This commit fixes the issue by moving the processing of the credentials file to earlier in startup of `sendXMLRPC.pl` so that the `WEBWORK_URL` will be set early enough to avoid the warning
	"Use of uninitialized value $webwork_url in string at (eval 329) line 266."
from when `conf/defaults.config` is processed.

It is still necessary to set an environment variable `WEBWORK_ROOT_URL` as that is used by the Docker version of `site.cont` to set `$server_root_url`. Ex:
```
export WEBWORK_ROOT_URL=https://demo.webwork.rochester.edu
```

Some minor reformatting was also done.

Below is a log of testing before and after various changes:
```
cd /data2/webwork_docker_devel/webwork2/clients
export PERL5LIB=/home/tani/perl5/lib/perl5/
export WEBWORK_ROOT=/data2/webwork_docker_devel/webwork2

# Before patches

./sendXMLRPC.pl -b  t/input.pg

Use of uninitialized value $webwork_url in string at (eval 329) line 266.
Use of uninitialized value $server_root_url in concatenation (.) or string at /data2/webwork_docker_devel/webwork2/conf/localOverrides.conf line 241.
Use of uninitialized value $webwork_url in string at (eval 353) line 266.
Use of uninitialized value $server_root_url in concatenation (.) or string at /data2/webwork_docker_devel/webwork2/conf/localOverrides.conf line 241.
Use of uninitialized value $WebworkClient::site_url in concatenation (.) or string at /data2/webwork_docker_devel/webwork2/lib/WebworkClient.pm line 165.
Use of uninitialized value $server_root_url in concatenation (.) or string at /data2/webwork_docker_devel/webwork2/conf/localOverrides.conf line 241.

# Just additional export, as site.conf for Docker installs needs WEBWORK_ROOT_URL

export WEBWORK_ROOT_URL=https://demo.webwork.rochester.edu

./sendXMLRPC.pl -b  t/input.pg
Use of uninitialized value $webwork_url in string at (eval 329) line 266.
[...]

PR #883, opened 2018-09-03
Rel ww2.14 send xmlrpc update

mgage:
Fixes a problem with -v (verbose) and adds two features.
1. You can specify the problemSeed using  the option -seed=s  where s is a number
2. You can take input from STDIN  using   sendXMLRPC.pl -h -    (- stands for STDIN)

To test:  in the directory  `webwork2/clients`
1.

sendXMLRPC.pl -v      (before patch this will do nothing)
                                      after patch this will print out information telling you which credential file is being used.

2. 
sendXMLRPC.pl --seed=123 -b t/input.pg     
inspect the line at the top of the problem to make sure that the problemSeed printed there agrees with the seed you entered.

3. cat t/input.pg |sendXMLRPC --seed=432 -b -
should work the same as the command above (except the problemSeed is different)

4. If you have this set up in a docker box and have vim installed in the docker box you can try
vim t/input.pg
inside vim type
:w ! ./pr
This should send an html version of the vim buffer to /var/www/html/xmlrpc_out.html
Using the browser view the rendered version at http://localhost:8080/xmlrpc_out.html
-----------
By editing the script pr you could also use this to print the vim buffer outside the docker environment -- just direct the output of sendXMLRPC.pl to the appropriate location or application

### DIFF
--- a/clients/pr
+++ b/clients/pr
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# inside the docker container
+# this can be called from vim with :w ! ./pr
+
+# it sends the contents of the buffer to /var/www/html/xmlrpc_out.html aka localhost:8080/xmlrpc_output.html
+
+./sendXMLRPC.pl -h - >/var/www/html/xmlrpc_out.html
+
+# process STDIN with sendXMLRPC.p and send it to /var/www/html/xmlrpc_out.html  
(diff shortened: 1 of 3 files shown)


mgage:
This developed conflicts once request #869 was pulled.  It has been replaced by #883

PR #884, opened 2018-09-04
Restore sendXMLRPC.pl changes

mgage:
Fixes a problem with -v (verbose) and adds two features.

    You can specify the problemSeed using the option -seed=s where s is a number
    You can take input from STDIN using sendXMLRPC.pl -h - (- stands for STDIN)

To test: in the directory webwork2/clients
1.

sendXMLRPC.pl -v (before patch this will do nothing)
after patch this will print out information telling you which credential file is being used.

sendXMLRPC.pl --seed=123 -b t/input.pg
inspect the line at the top of the problem to make sure that the problemSeed printed there agrees with the seed you entered.

    cat t/input.pg |sendXMLRPC --seed=432 -b -
    should work the same as the command above (except the problemSeed is different)

    If you have this set up in a docker box and have vim installed in the docker box you can try
    vim t/input.pg
    inside vim type
    :w ! ./pr
    This should send an html version of the vim buffer to /var/www/html/xmlrpc_out.html
    Using the browser view the rendered version at http://localhost:8080/xmlrpc_out.html

By editing the script pr you could also use this to print the vim buffer outside the docker environment -- just direct the output of sendXMLRPC.pl to the appropriate location or application


mgage:
By editing the script pr you could also use this to print the vim buffer outside the docker environment -- just direct the output of sendXMLRPC.pl to the appropriate location or application

This replaces pull request #883 which had conflicts.

Alex-Jordan:
I am in a Docker situation, and checked out a branch there with this PR.

`./sendXMLRPC.pl -v` behaves as advertised.

But then:

```
root@921e10f729b1:/opt/webwork/webwork2/clients# ./sendXMLRPC.pl --seed=123 -b t/input.pg
sh: 1: open: not found
```

So I guess "open" isn't a thing in this Docker box?

Is there an easy way to get `vim` (or `open`) installed in the Docker box? Is it as simple as `apt-get vim` from the terminal within the webwork bash? (I would just try and see, except I still feel on shaky footing with the whole Docker setup.)

mgage:
You can install vim inside the docker box with   apt-get 
just as you would with any unix system. 

I believe you have to do     apt-get update 
to get a fresh list of the apps that are available.

Instead of using -b use -h  or -H which produces raw html output and you can direct that to a file and then view it with a browser.

In my newest patch for 2.14 I have added a little file that allows you to print directly from vim to a file xmloutput.html in the webwork2/htdocs directory
and then you can view that with the browser using localhost:8080/webwork2_fiies/xmloutput.html. (I’m doing this from memory so the names may not 
be quite right). This is relatively new and I’m not yet a native vim user so there may be a better setup than I have created.

This is the best way I have figured out to get html generated inside Docker out to the host machine.  There is probably a way to pipe information
from inside Docker to something outside docker, but I haven’t figured it out yet.


Mac’s use ….. -o firefox  because they are not set up to use pipes     …. |firefox  like a normal unix system does.  

Before I merge PR #884 I think I will change the default configuration in ww_credentials so that it works out of the box for Docker instead of working
right out of the box for macs.

Thanks for looking at this stuff.  I think this sendXMLRPC.pl has great potential for speeding up editing but it still needs some polishing. 

Take care,

Mike

taniwallach:
@Alex-Jordan - I think the intention is to run `sendXMLRPC.pl` from the webwork2 directory from outside the Docker image, in which case it would use your system (Mac) vim. However, then you need to set up a valid credentials file with settings relevant to your PC, and if wanted have it contact the local Docker WW instead of the demo site at U of Rochester.

If you want to get vim installed inside the Docker image in a **persistent** way, it needs to be done by editing the Dockerfile (adding an apt-get to install vim):
```
RUN apt-get update && apt-get install vim
```
and then running `docker-compose build` when the container is not running. See: https://github.com/openwebwork/webwork2/wiki/Docker---things-to-get-used-to-(and-avoid-surprises)

**Note:** If you add the extra install near the very end of the file, it will not need to do much real work to rebuild the image. If you add vim into the main apt-get line, the rebuild will take a very long time.

Alex-Jordan:
@mgage Is `mgage/rel-ww2.14_september_update` important to have, or is it what you refer to as "newest patch for 2.14"? There was a trivial conflict when I merged that into `rel-ww2.14-sendXMLRPC-update2` (within Docker).

Is `mgage/rel-ww2.14_september_update` meant to replace `rel-ww2.14-sendXMLRPC-update2` instead of augment it?

Alex-Jordan:
I've done some testing now as @mgage outlined, and it's working as far as I can tell except for maybe something in item 4 below. I have to admit, I don't know what I'm doing :) I have some newb questions.

1. I'm in a Docker box. Could I really be doing this outside of Docker, with just a clone of the webwork2 repo that isn't configured at all except for the ww_credentials file? (As long as the credentials file points to, say, Mike's site?) What I mean is, could someone simply clone webwork2, edit the ww_credentials file, and start using this if they have the problem files they need?

2. I think I understand that the ww_credentials file (in general) points to some WW site out on the web. And I run `sendXMLRPC` from my local `webwork2/clients` folder. When I provide a PG file for `sendXMLRPC`, I am providing some file that I have locally, correct? (That is, I am not providing a path to a PG file on the remote WW site.)

3. It looks like Mike's site password is `xmlrpc`. If I were to set this up for the PCC or AIM WeBWorK server, where do I set the site_password? 

4. Actually, there doesn't appear to be any requirement to have a "correct" `site_password`. If I use `ww_credentials.dist` and I change `site_password` to something random, I still get output. (Whereas if I change the `course_password`, I get error messages.) Has `site_password` not been implemented?

Alex-Jordan:
I don't see an option to feed `sendXMLRPC` a base64 encoded PG problem. Is that doable?

taniwallach:
Going back to using the local Docker container as the server all of:

  * `./sendXMLRPC.pl --tex  t/input.pg` 
  * `./sendXMLRPC.pl --seed=999  --tex  t/input.pg`
  * `./sendXMLRPC.pl --seed=999  --pdf  t/input.pg`
  * `./sendXMLRPC.pl --seed=999  -h  t/input.pg`
  * `./sendXMLRPC.pl --seed=999  -b  t/input.pg`    (works, but has issue when Preview or Submit Answer is used)
  * `./sendXMLRPC.pl --seed=999  -bB  t/input.pg`    (goes straight to showing correct answers, and the results table looks fine, including MathJax)

seem to work fine.

It is interesting to note that when I used `./sendXMLRPC.pl --seed=999  -h  t/input.pg` I did see the port number in the output (and MathJax also works with the initial HTML display from `./sendXMLRPC.pl --seed=123 -b t/input.pg` and `./sendXMLRPC.pl --seed=123 -bB t/input.pg`) so things only go wrong with the port number **after** using the Preview or Submit Answer buttons. This seems to indicate to me that some sort of hidden form field may be needed to carry a special port number over into the server for such use cases.

taniwallach:
About using it with vim, I don't see any `pr` script in my webwork2 tree, so did not even try this yet.

In any case, it seems to me to be more useful to see if vim from **outside** the Docker container can trigger a call to send the current file to `sendXMLRPC.pl`. 

It seems to me that this should be possible using the option to have `sendXMLRPC.pl` process the standard input as was done using `cat t/input.pg | ./sendXMLRPC --seed=432 -b -` (which should work on any system which has a `cat` type commmand).

taniwallach:
Additional testing was recommended in the earlier PR https://github.com/openwebwork/webwork2/pull/860

Here is a more complete list of features to test (updated to use the relocated input file location and different seeds each time):

```
./sendXMLRPC.pl  --seed=123 -b t/input.pg 
./sendXMLRPC.pl  --seed=234 -h t/input.pg 
./sendXMLRPC.pl  --seed=345 --tex t/input.pg
./sendXMLRPC.pl  --seed=456 --pdf t/input.pg
./sendXMLRPC.pl  --seed=567 -bB t/input.pg
```

taniwallach:
### Remaining warning - fix:
Running
```
export WEBWORK_ROOT_URL=http://localhost:8080
```
overcomes the remaining warning about 
```Use of uninitialized value $WebworkClient::site_url in concatenation (.) or string at .../webwork2/lib/WebworkClient.pm line 165.```

### Missing pr script:

I found that the missing `pr` script is in PR https://github.com/openwebwork/webwork2/pull/883 and can be downloaded from there.

mgage:
Working on getting sendXMLRPC.pl and the docker config files correctly.  I think this pull request will capture the correct changes to sendXMLRPC.pl.  After this is in rel ww2.14 I'll go back and make sure the docker config files contain the latest information.

PR #887, opened 2018-10-13
docker-entrypoint.sh fix

taniwallach:
This pull request modifies how "chown" is run on the courses directory tree to avoid problems encountered on Mac OS X when there was a broken symbolic link.

The change only effects Docker users, and in a very minimal way in a part of the startup code which exists in order to try to prevent permissions problems from interfering with the use of the webwork2 environment running inside Docker.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -69,9 +69,15 @@ if [ "$1" = 'apache2' ]; then
         rm -f /etc/apache2/conf-enabled/apache2-reload.conf
     fi
     # Fix possible permission issues
+    echo "Fixing ownership and permissions (just in case it is needed)"
     cd $APP_ROOT/webwork2
-    chown -R www-data logs tmp DATA htdocs/tmp ../courses
+    chown -R www-data logs tmp DATA htdocs/tmp
     chmod -R u+w logs tmp DATA htdocs/tmp ../courses
+    cd $APP_ROOT
+    find courses -type f -exec chown www-data:root {} \;
+    find courses -type d -exec chown www-data:root {} \;
+    # OLD: chown www-data -R $APP_ROOT/courses
+    #    but that sometimes caused errors in Docker on Mac OS X when there was a broken symbolic link somewhere in the directory tree being processed
 fi
 
 exec "$@"


mgage:
Two things that need to be fixed in the documentation:

If I put a broken link in the webwork2/htdocs/tmp directory then docker-compose up exits with a code 1 error:
For example
`cd webwork-docker/webwork2/htdocs/tmp`
`ln -s foobar2 foobar`

will cause an error
app_1  | chown: changing ownership of 'htdocs/tmp/foobar': No such file or directory

followed eventually by a code 1 error exit

I found this by accident because I had used an earlier version of the webwork2 code in docker
and even though I pulled in rel-ww2.14 and this pull request this doesn't clear out the `htdocs/tmp`
directory. 

I tried using `chown -f -R ....` which suppresses the error message
but doesn't affect the code 1 error exit. 

I was able to add a directive     r`m -rf htdocs/tmp/*`  which fixes this problem, but I don't 
feel comfortable erasing the `htdocs/tmp` directory with every build even though in principle
anything in the tmp directory can be recreated (which costs a little bit of time, but not much)

I suggest for now giving clear instructions that if there are problems with changing ownership of the 
contents of the `htdocs/tmp` directory then the offending files should be removed from 
the directory `webwork2/htdocs/tmp`  (from outside the docker environment). 


(2) The second documentation notice:
If users want to use Rserv in their problems they need to uncomment the appropriate
`Rserv` configuration setting in localOverrides.conf. (there is one for running inside the docker
box and another for running webwork2 outside a docker box). localOverrides.conf can be edited either from within docker or from outside.

mgage:
I also found that this change in Docker still allowed docker-compose up to complete successfully.

```
#RUN cd $APP_ROOT/webwork2/ \
#    && chown -R www-data DATA ../courses htdocs/tmp htdocs/applets logs tmp $APP_ROOT/pg/lib/chromatic \
#    && chmod -R u+w DATA ../courses htdocs/tmp htdocs/applets logs tmp $APP_ROOT/pg/lib/chromatic
```
Since we are also performing this change of ownership in docker-entrypoint-fix it seems confusing to have it in both places.

taniwallach:
@mgage : 
  * I agree that some of the `chown` and `chmod` commands can almost certainly be removed from `Dockerfile`. 
    * The handling of `$APP_ROOT/pg/lib/chromatic` should probably remain in `Dockerfile`.
    * Whatever is handled in `docker-entrypoint.sh` can be omitted he `chown` and `chmod` in `Dockefile`.
  * About the broken symbolic links in `htdocs/tmp`, I have 2 ideas, but am not working on either yet:
    * Handle all the changes in this tree using `find` (as was done in `courses`) to avoid trying to `chown` trying to process any (good or broken) symbolic links.
    * Adding the `symlinks` package to the Docker image, to make the `/usr/bin/symlinks` program available during Docker startup, and adding the following somewhere suitable (shortly before the `chown` line) in `docker-entrypoint.sh` to delete any broken symbolic links before the `chown` is run:
```
/usr/bin/symlinks -r -d $APP_ROOT/webwork2/htdocs/tmp
```

PR #900, opened 2018-11-19
Timezone issue fix for creating assignment

taniwallach:
## Explanation of changes

1. Add a "safe" time format `"%m/%d/%Y at %I:%M%P"`  to the `my $fDueDate = $self->formatDateTime(...)` call to avoid issues with the short time-zone alias used by `%Z` in the default time format which causes the following `$dueDate = $self->parseDateTime()` call to fail in many time-zones. 
   * Ex: Asia/Jerusalem Africa/Johannesburg America/Santiago Asia/Bangkok Australia/Perth

2. In all 4 files use the same error message: 
   * `The set name '[_1]' is already in use.  Pick a different name if you would like to start a new set.`
  *using* `$r->maketext` when the requested set name is already in use.
   * `lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm` 
     * had this maketext string which was copied to the other files.
   * `lib/WeBWorK/ContentGenerator/Instructor/SetMakernojs.pm` and
   * `lib/WeBWorK/ContentGenerator/Instructor/GetLibrarySetProblems.pm` 
     * had no maketext but had essentially the same text.  
     * added $r->maketext()
   * `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm`
     * had slightly different text in maketext.
     * changed to the "standardized" message.

## Explanation of why the changes are needed
 * **Before the change**, setting the course timezone to `Asia/Jerusalem` or many other timezones **would trigger an error whenever the user tried to create a new homework set** using:
    * the "Create a New Set in This Course" button in the library browser (Calls `SetMaker.pm`)
    * the "Create" new set button from "Instructor Tools".  (This method also calls `SetMaker.pm`)
    * the "Create" part of "Hmwk Sets Editor" (This method calls `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm`)
 * **After this patch, the error no longer occurs.**
 * The cause of the error is the fact that the default time format (`"%m/%d/%Y at %I:%M%P %Z"`)
 used by `lib/WeBWorK/Utils.pm` includes `%Z` which generates a "short" time-zone string, which may not be able to be read back in during the parsing stage as representing a unique time-zone.
 * The error messages like

```
WeBWorK error
An error occured while processing your request. For help, please send mail
to this site's webmaster (techdesk@mathnet.technion.ac.il), including all of
the following information as well as what what you were doing when the error
occured.

DATE

Warning messages

Error messages
        Time zone 'IST' not recognized.

Call stack

in WeBWorK::Utils::parseDateTime called at line 2255 of         /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::parseDateTime called at line 1490 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
in WeBWorK::ContentGenerator::Instructor::SetMaker::pre_header_initialize called at line 214 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::go called at line 384 of /opt/webwork/webwork2/lib/WeBWorK.pm
```

```
WeBWorK error
An error occured while processing your request. For help, please send mail
to this site's webmaster (techdesk@mathnet.technion.ac.il), including all of
the following information as well as what what you were doing when the error
occured.

DATE

Warning messages

Error messages
    Time zone 'IST' not recognized.

Call stack

in WeBWorK::Utils::parseDateTime called at line 2255 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::parseDateTime called at line 1150 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
in WeBWorK::ContentGenerator::Instructor::ProblemSetList2::create_handler
called at line 404 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
in WeBWorK::ContentGenerator::Instructor::ProblemSetList2::initialize called at line 231 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::go called at line 384 of /opt/webwork/webwork2/lib/WeBWorK.pm
```
## Additional information
 * It does not seem necessary to include **any** timezone in this process, as the parse phase will be done for the course time-zone regardless of what had been included in the originally generated due date string.
 * The Perl documentation at https://metacpan.org/pod/DateTime about `$dt->time_zone_short_name()` states:
   * This method returns the time zone abbreviation for the current time zone, such as "PST" or "GMT". These names are not definitive, and should **not** be used in any application intended for general use by users around the world.
 * http://webwork.maa.org/wiki/Dates,_Times,_and_Time_Zones#Detailed_Information  suggests not listing an abbreviated time zone when entering dates, but in this setting WW was adding it automatically.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/GetLibrarySetProblems.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/GetLibrarySetProblems.pm
@@ -593,8 +593,7 @@ sub pre_header_initialize {
 			debug("new value of local_sets is ", $r->param('local_sets'));
 			my $newSetRecord	 = $db->getGlobalSet($newSetName);
 			if (defined($newSetRecord)) {
-	            $self->addbadmessage("The set name $newSetName is already in use.  
-	            Pick a different name if you would like to start a new set.");
+		            $self->addbadmessage($r->maketext("The set name '[_1]' is already in use.  Pick a different name if you would like to start a new set.",$newSetName));
 			} else {			# Do it!
 				# DBFIXME use $db->newGlobalSet
 				$newSetRecord = $db->{set}->{record}->new();
@@ -605,7 +604,7 @@ sub pre_header_initialize {
 				
 				my $dueDate = time+2*60*60*24*7;
 				my $display_tz = $ce->{siteDefaults}{timezone};
-				my $fDueDate = $self->formatDateTime($dueDate, $display_tz);
+				my $fDueDate = $self->formatDateTime($dueDate, $display_tz, "%m/%d/%Y at %I:%M%P");
 				my $dueTime = $ce->{pg}{timeAssignDue};
 				
 				# We replace the due time by the one from the config variable
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
@@ -1130,7 +1130,11 @@ sub create_handler {
 
 	my $newSetID = $actionParams->{"action.create.name"}->[0];
 	return CGI::div({class => "ResultsWithError"}, $r->maketext("Failed to create new set: no set name specified!")) unless $newSetID =~ /\S/;
-	return CGI::div({class => "ResultsWithError"}, $r->maketext("Set [_1] exists.  No set created", $newSetID)) if $db->existsGlobalSet($newSetID);
+	return CGI::div({class => "ResultsWithError"},
+			$r->maketext("The set name '[_1]' is already in use.  Pick a different name if you would like to start a new set.",$newSetID)
+			. " " . $r->maketext("No set created.")
+                       ) if $db->existsGlobalSet($newSetID);
+
 	my $newSetRecord = $db->newGlobalSet;
 	my $oldSetID = $self->{selectedSetIDs}->[0];
 
@@ -1140,7 +1144,7 @@ sub create_handler {
 
 	my $dueDate = time+2*ONE_WEEK();
 	my $display_tz = $ce->{siteDefaults}{timezone};
-	my $fDueDate = $self->formatDateTime($dueDate, $display_tz);
+	my $fDueDate = $self->formatDateTime($dueDate, $display_tz, "%m/%d/%Y at %I:%M%P");
 	my $dueTime = $ce->{pg}{timeAssignDue};
 
 	# We replace the due time by the one from the config variable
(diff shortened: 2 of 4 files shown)


taniwallach:
##Various time-zones and the errors hit:

1. Asia/Jerusalem
```
            Time zone 'IST' not recognized.
```

2. Africa/Johannesburg
```
            Time zone 'SAST' not recognized.
```

3. America/Santiago
```
            Time zone 'CLT' not recognized.
```

4. Asia/Bangkok
```
            Time zone 'ICT' not recognized.
```

5. Australia/Perth
```
            Time zone 'AWST' not recognized.
```

##About problematic time-zones
 * I ran some code to check how many time-zones seem to hit the problem this patch fixes.
    * 189 time-zones were reported as **OK** by the test code.
    * 170 time-zones were reported as **bad** by the test code.
 * The test code (based on calling the routines from `webwork2/lib/WeBWorK/Utils.pm` with a small change to replace a `die` with an "error" return value) is included in the file: [TestCode.zip](https://github.com/openwebwork/webwork2/files/2597039/TestCode.zip)

taniwallach:
@mgage This patch will also need to be applied to the development branches and the pull requests and/or branches which have the current internationalization support. It might also merit a hotfix to prior releases which are in use.

taniwallach:
## Testing
 * I tested the modification to the new code which uses the time formatting string in
   * `lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm`
   * `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm`
 * I did not test (but the code is essentially identical to that in `SetMaker.pm`):
   * `lib/WeBWorK/ContentGenerator/Instructor/SetMakernojs.pm`
   * Testing probably requires making a configuration change to use this no-JavaScript version of the Set Maker code.
 * I did not test (but the code is essentially identical to that in `SetMaker.pm`):
   * `lib/WeBWorK/ContentGenerator/Instructor/GetLibrarySetProblems.pm`
   * I'm not sure when this code gets called, but the bug appears in the old code and the change will do the fix.

mgage:
Works for me so far.  
Should https://github.com/openwebwork/webwork2/blob/master/lib/WeBWorK/Utils.pm#L55
use constant DATE_FORMAT => "%m/%d/%Y at %I:%M%P %Z";
in Utils.pm be modified as well?

mgage:
Is there an issue with gateway quizzes?  They use a different formatDateTime() function from ContentGenerator.pm.

mgage:
my experience with the two questions above.  I was able to create gateway tests without a problem. 
by not modifying the DATE_FORMAT i get a report that the gateway test closes at xxxxxx  IST (i.e. the colloquial local abbreviation) which might be desirable for output as long as we don't input it that way.

I recommend not changing the DATE_FORMAT for the time being at least.

PR #902, opened 2018-12-20
Give PreTeXt access to answer hash

Alex-Jordan:
This makes it so that when PreTeXt calls for a problem rendered in PTX, the answer hash(es) is/are included. So for example when writing a book and you include some WeBWorK problem with some specific seed, PTX can print the question and perhaps also print the answer in the back of the book.


mgage:
branched from before rel 2.14 was merged, but changes prior the current branch have been put into both master and develop. I think this is ok.

PR #905, opened 2018-12-31
Rel ww2.14

mgage:
This moves the updates put into master into the develop branch as well

PR #906, opened 2018-12-31
remove commands from docker that change ownership of webwork2/htdocs/tmp files 

mgage:
these are often pointers to files in an OPL that hasn't been created yet.


mgage:
merge the PR #905 first

PR #907, opened 2019-01-01
Develop international

mgage:
This PR contains a number of changes to UTF8 code and should be checked carefully for "do no harm" to existing sites and problems. It's less important for now that it resolve "all" problems with using extended character sets. Further tweaks will slowly extended capabilities for handling all languages.

The plan is to change the code to handle UTF8 first but try to keep it compatible with existing database configurations.  When the code is stable we'll devise a plan to change the configuration on existing databases and convert webwork problems where necessary.


mgage:
Closing this for now.  I'll submit it to github.com/mgage/develop_candidate which has some of the same commits and resolve differences there.

mgage:
It turns out that all of the commits in the branch develop_international have been incorporated into mgage/develop_candidate which has issued a pull request PR#893.  I will close the branch develop_candidate.

PR #908, opened 2019-01-02
New develop candidate 01 01 2019

mgage:
This includes some fixes which have been sitting on my laptop for a while. I think this is an improved version of develop_candidate and I'm withdrawing that pull request.


taniwallach:
I think we need to reconsider the difference between `utf8` and `utf8mb4` and whether the WW database code and tables should use the restricted `utf8` rather than the more complete `utf8mb4` charset.

However, this mainly applies to places where real UTF8 may be stored. Whatever is being base64 encoded before being put into the database need not have `utf8` or `utf8mb4` support in the relevant columns.

It seems to me that we should use `utf8mb4` to avoid the issues with the restricted `utf8` mySQL/MariaDB charset. See for example: https://mathiasbynens.be/notes/mysql-utf8mb4 where this is recommended. 

@heiderich recommended using `utf8mb4` in the discussion for PR 800.

@dvpc and @goehle settled for the more restricted `utf8` due to the key-length issues when PR 594 was discussed. It seems that the allowed key prefix lengths supported by the standard databases at that time may have been shorter than what is supported now. However, @dvpc was working under the assumption that no-one would try to use a character outside of the range of the simpler `utf8` (the essentially the Basic Multilingual plane). I tend to think that this is a dangerous assumption when users/staff may be able to use other UTF8 characters in strings they enter, if the web-browser / WW interface is allowing full UTF8 input.

Modern database options allow longer keys, and up to about varchar(190) is apparently widely supported in keys even when the 767 byte limit of some reasonably recent mySQL/MariaDB InnoDB table configurations applies. Thus, supporting `utf8mb4` seems pretty reasonable to me.

See:
  * https://github.com/openwebwork/webwork2/pull/800#issuecomment-320333459
  * http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4534
  * https://github.com/openwebwork/webwork2/pull/594

Text pulled from my forum post and edited:

The reasons to use "utf8mb4" and not "plain utf8" is explained at ttps://mathiasbynens.be/notes/mysql-utf8mb4 and in various other web pages. PR 594 also had some discussion of the differences.

In short SQL's "plain utf8" is a subset with various limitations and issues, and only supports the haracters from the "real utf8" which require 1-3 bytes to encode. That means that the character to byte conversion factor is only 3, which allows longer keys than a "real UTF8" does.

"utf8mb4" (a full UTF8 for SQL) supports all of Unicode, but as it includes the 4 byte encodedc haracters, requires a character to byte conversion factor of 4, and this reduced what 1000 bytes will allow from about 333 to about 250.  Some versions of InnoDB tables allowed only 767 bytes, while MyIASM allowed 1000 bytes. The 767 byte limit seems to require cutting back to 191 characters with the older settings and utf8mb4. Newer DB versions/table types can apparently allow longer keys, as explained in some of the links below.

The additional byte reserved per character when using "utf8mb4" as the SQL charset means that some WW table keys end up being over the length limit (1000 bypes), while they are OK for the more restricted "utf8" charset.

The problem in changing to utf8mb4 is that various tables include keys which require over 1000 bytes (or 767 bytes) when utf8mb4 is used. This will be one of the challenges in getting full UTF8 support in WW.

There are some discussions of using INNODB tables and "innodb_large_prefix = on" with a COMPRESSED or DYNAMIC row formats and related settings to help avoid bypass issues. 

Some references: 

- https://github.com/openwebwork/webwork2/pull/800 and in particular https://github.com/openwebwork/webwork2/pull/800#issuecomment-320333459
- https://github.com/openwebwork/webwork2/pull/594
- https://dba.stackexchange.com/questions/76469/mysql-varchar-length-and-performance/76470#76470
- https://dba.stackexchange.com/questions/49913/specified-key-was-too-long-max-key-length-is-1000-bytes-in-mysql-5-6
- https://github.com/laravel/framework/issues/17337
- https://stackoverflow.com/questions/43379717/how-to-enable-large-index-in-mariadb-10
 
I mainly paid attention to this issue several months ago when looking at the OPL tables, in which several varchar(255) fields are included in database keys. Jason Aubrey reported that with uft8mb4 as the default charset for MariaDB 10.1 in Debian 9 he could not even create the admin course due t keys longer than the limit.

I had a look at the contents of the OPL tables a while back due to this issue, and the actual data does not currently use such long fields. The two fields which were close to 200 characters of real usage were:
Field: source_file in OPL_global_statistics order which had a max length of 198 characters when I looked at it, and that is a PRImary key field.
Field: path in OPL_path order which had a max length of 171 characters when I looked at it.

mgage:
Hi Tani,

Without having researched as much as you have I also think we should go to using the full utf8mb.  What I have been doing is to change as much of the perl code
to allow utf8mb.  As far as I can tell in perl utf8 and utf8mb are synonymous.  I think there is a reasonable chance that develop would work well with
mysql once the limits on certain field sizes have been adjusted.

I’m worried about what it would take to convert data (e.g. problems) that had been stored in utf8 or smaller format to behave properly whn the format is changed to utf8mb.
For now I’ve been trying to put that problem off. So far, without changing the database the develop branch is doing a good job of handling
french. http://demo.webwork.rochester.edu/webwork2/french_language_webwork/M1L1/4/
arabic: http://demo.webwork.rochester.edu/webwork2/arabic_language_webwork/test_homework_set/1/  (someone who understands arabic will have to help edit that problem)
hebrew is set up: http://demo.webwork.rochester.edu/webwork2/hebrew_language_webwork   but needs a sample problem in Hebrew

all of these courses need more translating done, and also need sample WW problems
spanish:  http://demo.webwork.rochester.edu/webwork2/spanish_language_webwork/
hungarian: http://demo.webwork.rochester.edu/webwork2/hungarian_language_course/
chinese: http://demo.webwork.rochester.edu/webwork2/chinese_language_course/
german: http://demo.webwork.rochester.edu/webwork2/german_language_webwork/
korean:  http://demo.webwork.rochester.edu/webwork2/korean_language_webwork/
turkish: http://demo.webwork.rochester.edu/webwork2/turkish_language_webwork

(email me if you would like instructor permission for one of these language courses so that you can add test webwork problems).


It might be worth seeing what happens if we convert to innodb format. Moodle converted some time ago and the prevailing wisdom was that that was the way to go.

We did some experiments on this early on (when the push to innodb was just getting started) and found that using innodb slowed things down — in particular the library
script OPL-update was very slow.  I think it’s worth doing that experiment again before we go much further with the utf8/utf8mb database question.

Take care,

Mike



On Jan 8, 2019, at 6:21 PM, Nathan Wallach <notifications@github.com<mailto:notifications@github.com>> wrote:


I think we need to reconsider the difference between utf8 and utf8mb4 and whether the WW database code and tables should use the restricted utf8 rather than the more complete utf8mb4 charset.

However, this mainly applies to places where real UTF8 may be stored. Whatever is being base64 encoded before being put into the database need not have utf8 or utf8mb4 support in the relevant columns.

It seems to me that we should use utf8mb4 to avoid the issues with the restricted utf8 mySQL/MariaDB charset. See for example: https://mathiasbynens.be/notes/mysql-utf8mb4<https://urldefense.proofpoint.com/v2/url?u=https-3A__mathiasbynens.be_notes_mysql-2Dutf8mb4&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=sA23zbzB-6nIQlkGrLj6KZwcQxw8J6Otcr1xCa4qngQ&e=> where this is recommended.

@heiderich<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_heiderich&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=3kwmqY17CLXpcDZrlJU5r8VrXJyhzTe8eZPWaFkl0kE&e=> recommended using utf8mb4 in the discussion for PR 800.

@dvpc and @goehle<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_goehle&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=8r9lRJreeN3N-yw_jH06rtCKz3k6pPPXJnZM4RJNRvA&e=> settled for the more restricted utf8 due to the key-length issues when PR 594 was discussed. It seems that the allowed key prefix lengths supported by the standard databases at that time may have been shorter than what is supported now. However, @dvpc was working under the assumption that no-one would try to use a character outside of the range of the simpler utf8 (the essentially the Basic Multilingual plane). I tend to think that this is a dangerous assumption when users/staff may be able to use other UTF8 characters in strings they enter, if the web-browser / WW interface is allowing full UTF8 input.

Modern database options allow longer keys, and up to about varchar(190) is apparently widely supported in keys even when the 767 byte limit of some reasonably recent mySQL/MariaDB InnoDB table configurations applies. Thus, supporting utf8mb4 seems pretty reasonable to me.

See:

  *   #800 (comment)<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_openwebwork_webwork2_pull_800-23issuecomment-2D320333459&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=vuWa3IjA9_GqORYZWimfgjABhF9taEU23QhuX5CsG6U&e=>
  *   http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4534<https://urldefense.proofpoint.com/v2/url?u=http-3A__webwork.maa.org_moodle_mod_forum_discuss.php-3Fd-3D4534&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=aOpO2ZFDakfz-5lbSl-lqy0gv6JcIIl30j4TjViI1xw&e=>
  *   #594<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_openwebwork_webwork2_pull_594&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=OPk53zUHHCM3Ckrm3FEM9yCjqYMYJaVAfZy1jb3Ksms&e=>

Text pulled from my forum post and edited:

The reasons to use "utf8mb4" and not "plain utf8" is explained at ttps://mathiasbynens.be/notes/mysql-utf8mb4 and in various other web pages. PR 594 also had some discussion of the differences.

In short SQL's "plain utf8" is a subset with various limitations and issues, and only supports the haracters from the "real utf8" which require 1-3 bytes to encode. That means that the character to byte conversion factor is only 3, which allows longer keys than a "real UTF8" does.

"utf8mb4" (a full UTF8 for SQL) supports all of Unicode, but as it includes the 4 byte encodedc haracters, requires a character to byte conversion factor of 4, and this reduced what 1000 bytes will allow from about 333 to about 250. Some versions of InnoDB tables allowed only 767 bytes, while MyIASM allowed 1000 bytes. The 767 byte limit seems to require cutting back to 191 characters with the older settings and utf8mb4. Newer DB versions/table types can apparently allow longer keys, as explained in some of the links below.

The additional byte reserved per character when using "utf8mb4" as the SQL charset means that some WW table keys end up being over the length limit (1000 bypes), while they are OK for the more restricted "utf8" charset.

The problem in changing to utf8mb4 is that various tables include keys which require over 1000 bytes (or 767 bytes) when utf8mb4 is used. This will be one of the challenges in getting full UTF8 support in WW.

There are some discussions of using INNODB tables and "innodb_large_prefix = on" with a COMPRESSED or DYNAMIC row formats and related settings to help avoid bypass issues.

Some references:

  *   #800<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_openwebwork_webwork2_pull_800&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=Ktul3nRRDwosjWLHTFWHJfwRMQdb5cXch0RH9OErLD0&e=> and in particular #800 (comment)<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_openwebwork_webwork2_pull_800-23issuecomment-2D320333459&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=vuWa3IjA9_GqORYZWimfgjABhF9taEU23QhuX5CsG6U&e=>
  *   #594<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_openwebwork_webwork2_pull_594&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=OPk53zUHHCM3Ckrm3FEM9yCjqYMYJaVAfZy1jb3Ksms&e=>
  *   https://dba.stackexchange.com/questions/76469/mysql-varchar-length-and-performance/76470#76470<https://urldefense.proofpoint.com/v2/url?u=https-3A__dba.stackexchange.com_questions_76469_mysql-2Dvarchar-2Dlength-2Dand-2Dperformance_76470-2376470&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=0z43wHeqwT9ica2CtpC_yFsxfv3r066hvr8y82Qho4s&e=>
  *   https://dba.stackexchange.com/questions/49913/specified-key-was-too-long-max-key-length-is-1000-bytes-in-mysql-5-6<https://urldefense.proofpoint.com/v2/url?u=https-3A__dba.stackexchange.com_questions_49913_specified-2Dkey-2Dwas-2Dtoo-2Dlong-2Dmax-2Dkey-2Dlength-2Dis-2D1000-2Dbytes-2Din-2Dmysql-2D5-2D6&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=7hLcY1skLszdDQFnhtchKtxa9TIvtxsmSB0PKFY0vms&e=>
  *   laravel/framework#17337<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_laravel_framework_issues_17337&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=RfSIChB2m51XketHuIJOMeSf6b9qxLREGmpfY-1C-jI&e=>
  *   https://stackoverflow.com/questions/43379717/how-to-enable-large-index-in-mariadb-10<https://urldefense.proofpoint.com/v2/url?u=https-3A__stackoverflow.com_questions_43379717_how-2Dto-2Denable-2Dlarge-2Dindex-2Din-2Dmariadb-2D10&d=DwMFaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=fmQKsXRbeQJOgozHvpGNdTsDtRGvNDnTVuSBYvFcAMc&s=jqTxglItv9F0lB82t_uHeY548HQScV3rWOnsC4Ndw5c&e=>

I mainly paid attention to this issue several months ago when looking at the OPL tables, in which several varchar(255) fields are included in database keys. Jason Aubrey reported that with uft8mb4 as the default charset for MariaDB 10.1 in Debian 9 he could not even create the admin course due t keys longer than the limit.

I had a look at the contents of the OPL tables a while back due to this issue, and the actual data does not currently use such long fields. The two fields which were close to 200 characters of real usage were:
Field: source_file in OPL_global_statistics order which had a max length of 198 characters when I looked at it, and that is a PRImary key field.
Field: path in OPL_path order which had a max length of 171 characters when I looked at it.

taniwallach:
> Hi Tani, Without having researched as much as you have I also think we should go to using the full utf8mb. What I have been doing is to change as much of the perl code to allow utf8mb. As far as I can tell in perl utf8 and utf8mb are synonymous. I think there is a reasonable chance that develop would work well with mysql once the limits on certain field sizes have been adjusted. I’m worried about what it would take to convert data (e.g. problems) that had been stored in utf8 or smaller format to behave properly when the format is changed to utf8mb4. 

As far as I understand it the difference between `utf8mb4` and `utf8` is **only** in mySQL and MariaDB, as mySQL decided to create a charset in mySQL called `utf8` which is restricted to Unicode characters which need 3 or fewer bytes in the "real" UTF-8 encoding. Thus there are no Perl side issues except for those triggered by trying to push disallowed UTF-8 characters into mySQL tables which do not permit them (where `utf8mb4` would allow them) or expecting such characters to come back out of the database which could not handle them.

Thus changing from mySQL `utf8` to `utf8mb4` cannot lose data. The challenge is the change from the current `latin1` encoding to `utf8mb4`.

Apparently using `utf8` in mySQL can also trigger problems with encrypted data pushed into VARCHAR columns, as certain byte sequences are disallowed. (But that is bad practice as binary data should not be stored in VARCHAR columns.)

The recommendation seems to be to open connections with `SET NAMES utf8mb4 COLLATE 
utf8mb4_unicode_520_ci` instead of `SET NAMES utf8` and to explicitly set the charset and collation when a database is first created.

References:
  * https://stackoverflow.com/questions/30074492/what-is-the-difference-between-utf8mb4-and-utf8-charsets-in-mysql
  * https://stackoverflow.com/questions/47566730/force-mariadb-clients-to-use-utf8mb4
  * https://mariadb.com/kb/en/library/setting-character-sets-and-collations/
  * https://mariadb.com/kb/en/library/supported-character-sets-and-collations/

References on the conversion process (various options):
   * https://codex.wordpress.org/Converting_Database_Character_Sets
  * https://dba.stackexchange.com/questions/123572/convert-mysql-database-from-latin1-to-utf8mb4-and-take-care-of-german-umlauts
  * A walkthrough via a SQL dump / modify / restore: https://gist.github.com/mdemblani/34d30015f24e1927736f189b88be143a
  * Another walkthrough: https://www.percona.com/blog/2018/04/10/migrating-database-charsets-to-utf8mb4/  which has some comments on how to use `ALTER TABLE` if necessary.
  * That one referred to an explanation of why it uses VARBINARY as an intermediate stage in `ALTER TABLES` in https://www.tusacentral.net/joomla/index.php/mysql-blogs/162-understand-what-happens-in-mysql-when-using-utf-string-with-latin1-encoding.html
  * https://nixmash.com/post/i-never-did-like-latin-from-mysql-latin-1-to-utf8mb4
  * https://medium.com/@manish_demblani/breaking-out-from-the-mysql-character-set-hell-24c6a306e1e5

mgage:
We are getting "The file does not appear to be a text file" when trying to
read pg files using FileManager on demo.webwork.rochester.edu which is using 
new develop candidate 01 01 2019 (#908)

The code being triggered seems to be line 511 in FileManager
and isText() is defined at line 1275 by:
```
##################################################
#
# Check if a string is plain text
# (i.e., doesn't contain four non-regular
# characters in a row.)
#
##################################################
#
# Check if a string is plain text
#
sub isText {
	my $string = shift;
	#	return $string !~ m/[^\s\x20-\x7E]{4}/;
	return utf8::is_utf8($string);
	# return $string !~ m/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]{2}/;
}
```
The read file branch has been changed substantially in this PR but it hasn't caused trouble anywhere else.  isText is only used in this file, however attempting to just bypass the check searching for characters out of range 
doesn't immediately stop the symptoms.

The symptoms: On demo.webwork.rochester.edu/webwork2/2019_JMM_Baltimore (and any other course including UR101 which you can access with profa/profa ) go to the file manager and try to read or view any .pg file. You will get the "is not a text file" error.

dpvc:
I don't think that `utf8::is_utf8($string);` is the right thing to use, here.  The `isText()` function is supposed to tell if the contents of a string is text rather than raw binary data, but that is not what `utf8::is_utf8()` does.  That function is only about how the string is stored internally, not about the actual contents of the string, so is **not** the appropriate test for this.

Of course, there is no actual way to tell if a string is text versus binary, and so `isText()` was just a heuristic to try to work it out.  Of course, it derives from the olden days, when text files were ASCII files, and so would not contain many non-ASCII characters, so the test was to look for four non-ASCII characters in a row as an indication of a binary file.

Now, with Unicode files, that is no longer a sufficient test, since they can contain lots of non-ASCII characters.  But whether the string is stored internally in UTF8 encoding or not has nothing to do with its contents, so is definitely the wrong test.

Perhaps a better test would be to look for a high percentage of spaces and newlines (or carriage returns)?  Text files like .pg files and such should have a higher concentration of those than a binary file.  That might do the trick.

The main reason for the test is to tell if you need to convert line-endings from Mac or DOS line-endings (`\r` or `\r\n`) to unix line-endings (`\n`).   Perhaps that is no longer really necessary (though I suspect it is).  The other reason is to preview viewing/editing non-text files, but I suppose that is not all that critical.  So it might be that you can do without the check entirely.

mgage:
Thanks Davide.  I thought it might be something like that.  I'm finally on the plane ready to fly back to Rochester.   I got snowed out on Sunday and this was the first flight.




On Jan 23, 2019 10:43 AM, "Davide P. Cervone" <notifications@github.com> wrote:

I don't think that utf8::is_utf8($string); is the right thing to use, here. The isText() function is supposed to tell if the contents of a string is text rather than raw binary data, but that is not what utf8::is_utf8() does. That function is only about how the string is stored internally, not about the actual contents of the string, so is not the appropriate test for this.

Of course, there is no actual way to tell if a string is text versus binary, and so isText() was just a heuristic to try to work it out. Of course, it derives from the olden days, when text files were ASCII files, and so would not contain many non-ASCII characters, so the test was to look for four non-ASCII characters in a row as an indication of a binary file.

Now, with Unicode files, that is no longer a sufficient test, since they can contain lots of non-ASCII characters. But whether the string is stored internally in UTF8 encoding or not has nothing to do with its contents, so is definitely the wrong test.

Perhaps a better test would be to look for a high percentage of spaces and newlines (or carriage returns)? Text files like .pg files and such should have a higher concentration of those than a binary file. That might do the trick.

The main reason for the test is to tell if you need to convert line-endings from Mac or DOS line-endings (\r or \r\n) to unix line-endings (\n). Perhaps that is no longer really necessary (though I suspect it is). The other reason is to preview viewing/editing non-text files, but I suppose that is not all that critical. So it might be that you can do without the check entirely.

Alex-Jordan:
I've run into WW telling me the answer log (or was it the login log?) is not a text file, and refusing to show it to me. If a new test is devised, these should be looked at along with .pg.

Speaking of that, why not pass judgment by file extension? Whitelist .pg, .pl, .txt, .html, .log, .csv, ...

taniwallach:
I think Alex is right - the first round of decisions should based on file extensions. That would be fast and accurate for most files in a WW course. File extensions are already how `lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm` is deciding what are image files (~line 469 : `$file =~ m/\.(gif|jpg|png)/i`). (A config option could be included to allow editing the recognized extensions for text and image files.) 

As a fallback, we could use the output from the Unix `file` utility to decide what other file types to treat as text, if there is any real need for it. The `file` utility tends to do a very good jobs (in my experience) of classifying files.  Some output classifications from a quick run on some misc WW related directories:
  * ASCII text
  * ISO-8859 text
  * UTF-8 Unicode text
  * PNG image data ...
  * GIF image data ...
  * JPEG image data ...
  * SVG Scalable Vector Graphics image
  * Perl script text executable
  * zlib compressed data
  * PDF document ...
  * Zip archive data
  * compiled Java class data ...
  * TrueType font data

dpvc:
> we could use the output from the Unix `file` utility to decide

Note that the main usage for `isText()` is for knowing whether to convert line breaks when a file is uploaded to the server.  That is happening *before* the file is created on the unix system, so you can't run `file` on it yet, unless you first save it verbatim as a temporary file, then test it, then delete the temporary file and save the modified one (if it needs to be modified).

For the other situations, yes, the would work, but you now have to decode the `file` output.  I think file extensions are probably sufficient, though hardly perfect.

taniwallach:
> Note that the main usage for `isText()` is for knowing whether to convert line breaks when a file is uploaded to the server. That is happening _before_ the file is created on the unix system, so you can't run `file` on it yet, unless you first save it verbatim as a temporary file, then test it, then delete the temporary file and save the modified one (if it needs to be modified).

Thanks - sorry, your comment that this is the most important need for isText() did not catch my attention. I also think it likely that things will work better if text files on the server do have Unix line-endings. `file` also (at least the current version I use on Debian) reports when there is a text file with non-Unix line endings, and it seems pretty easy to determine from the output that a file is a "text" file with `with CRLF line terminators` (dos) or with `with CR line terminator` (mac) . (See below.)

For the upload context, I am very skeptical that a simple test to properly determine if a file is a text file can be written once UTF-8 is allowed which is less complicated than storing a temp file and checking it with the unix `file` command. Why recreate the wheel... why not just upload the raw file as is, test file file type, and when necessary (detected as a text file with non-Unix line endings) either trigger an automatic conversion (maybe using the utilities `dos2unix` and/or `mac2unix` to convert the file in place) or ask the user for an OK to run such a conversion. We could save on checking with the user for the "known" text file extensions. 

Once that is possible, it should not be difficult to also add a simple feature to allow converting files in place between the different line-endings as needed by the user. (Maybe a user wants to download a PG file with DOS line endings, and has fewer tools to do this on the local PC side of things?)

Sample `file` output reporting non-Unix line-endings:

```
tani@lxtani:/tmp$ file course.conf 
course.conf: ASCII text

tani@lxtani:/tmp$ unix2dos course.conf
unix2dos: converting file course.conf to DOS format...

tani@lxtani:/tmp$ file course.conf 
course.conf: ASCII text, with CRLF line terminators

tani@lxtani:/tmp$ dos2unix course.conf ; unix2mac course.conf
dos2unix: converting file course.conf to Unix format...
unix2mac: converting file course.conf to Mac format...

tani@lxtani:/tmp$ file course.conf 
course.conf: ASCII text, with CR line terminators

tani@lxtani:/tmp$ file HebrewSample1.pg
HebrewSample1.pg: UTF-8 Unicode text

tani@lxtani:/tmp$ unix2dos HebrewSample1.pg
unix2dos: converting file HebrewSample1.pg to DOS format...

tani@lxtani:/tmp$ file HebrewSample1.pg
HebrewSample1.pg: UTF-8 Unicode text, with CRLF line terminators

tani@lxtani:/tmp$ dos2unix HebrewSample1.pg ; unix2mac HebrewSample1.pg
dos2unix: converting file HebrewSample1.pg to Unix format...
unix2mac: converting file HebrewSample1.pg to Mac format...

tani@lxtani:/tmp$ file HebrewSample1.pg
HebrewSample1.pg: UTF-8 Unicode text, with CR line terminators
```

taniwallach:
If the file suffix + `file` test approach seems reasonable, I can try in the coming week to write the code, do some rudimentary tests, and make a PR into 
https://github.com/mgage/webwork2/tree/new_develop_candidate_01_01_2019

However, if someone has a better idea - we can try it.  @dpvc - do you want to try out the white-space/line-end statistics first?

mgage:
This PR has been replaced by PR #927 where the conflicts have already been resolved. 
I'll close this commit, but it should be remembered for the discussion items it contains.  We 
may need them as we vet the changes for the localization commits.

PR #910, opened 2019-01-08
File '' uploaded successfully

goehle:
This is an extremely small change that fixes the file uploaded successfully message.  To test upload a file using the file manager.  After the patch the file name should be displayed correctly.


mgage:
Thanks Goeff

PR #911, opened 2019-01-09
2 related fixes to Authen.pm: a. Deny authentication in checkPassword() of lib/WeBWorK/Authen.pm when the submitted password is an empty string or an all white-space string. b. Fix the new trim() function and override an override so session timeouts will behave again. 

taniwallach:
**Revised and ready to test. See also the following comments**

Modify the main checkPassword() in lib/WeBWorK/Authen.pm to forbid accepting empty or all white-space password even if they happen to match the crypted password in the database. (In certain settings such as LTI authentication and LDAP authentication, and when wwassignment is being used, some users may have such a null password.)

Since the function is called as a fallback by some of the other authentication modules which do not "trim" the possibleClearPassword being sent for testing (unlike what is done in get_credentials() in lib/WeBWorK/Authen.pm), we explicitly test for a non-space character and don't only reject the possibleClearPassword when it is an empty string.

This is intended to address the issue reported in the forums http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4529 .

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -232,8 +232,10 @@ sub verify {
 			$self->write_log_entry("LOGIN FAILED $log_error");
 		}
 		if (defined($error) and $error=~/\S/) { # if error message has a least one non-space character. 
-
-			if (defined($r->param("user")) or defined($r->param("user_id"))) {
+			if ( defined( $log_error ) and $log_error eq "inactivity timeout" ) {
+				# We don't want to override the localized inactivity timeout message.
+				# so do not check next "if" in this case.
+			} elsif (defined($r->param("user")) or defined($r->param("user_id"))) {
 				$error = $r->maketext("Your authentication failed.  Please try again. Please speak with your instructor if you need help.")
 			}
 
@@ -309,8 +311,14 @@ sub do_verify {
 
 sub trim {  # used to trim leading and trailing white space from user_id and password
             # in get_credentials
-  my $s = shift//'';
-  $s =~ s/(^\s+|\s+$)//g;
+  my $s = shift;
+  # If the value was NOT defined, we want to leave it undefined, so
+  # we can still catch session-timeouts and report them properly.
+  # Thus we only do the following substitution if $s is defined.
+  # Otherwise return the undefined value so a non-defined password
+  # can be caught later by authenticate() for the case of a
+  # session-timeout.
+  $s =~ s/(^\s+|\s+$)//g    if ( defined($s) );
   return $s;
 }
 sub get_credentials {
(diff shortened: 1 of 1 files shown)


taniwallach:
The use of the trim() function added to 2.14 means that a password with either leading or trailing white-space will be cleaned up before being tested using the plain "local" authentication. However, it is possible to set passwords which have leading or trailing white-space as part of the password using Classlist editor. The same issue may apply to other methods for creating/editing passwords. This is something to consider detecting and handling in the future.

mgage:
Thanks Tani.

taniwallach:
I think I now have a fully functional fix.

1. It turns out that the new `trim()` function added in release 2.14 to clean up the value of `$self->{password}` would replace an undefined value with an empty string. **Doing so prevented the `authenticate()` function from being able to detect session-timeouts** (as the trigger is when `$self->{password}` has the Perl `undef` value). The `trim()` function has been modified not to change `undef` but only to clean up strings which are defined.
2. Once that was working, the course's `login.log` file was reporting `LOGIN FAILED inactivity timeout` messages on session timeout.
3. However, I was still not getting the expected error message 
    * `Your session has timed out due to inactivity. Please log in again.` after a timeout, 
    * on the login screen which opened, but instead the generic 
    * `Your authentication failed.  Please try again. Please speak with your instructor if you need help.` 
    * error message. (Both are localized by `maketext`.)
    * This was fixed by bypassing the code which overrode the "timeout" error message in the `verify()` function when it was set. The actual bypass is triggered when `$log_error eq "inactivity timeout"` as that log message is set at the same time that the localized session timeout error message is set, in which case we don't want to override it.

I apologize for making the PR before handling these additional aspects. The realization of how the new `trim()` function was triggering the "bad" behavior happened only after I handled the empty/all-white-space issue.

### To test:
1. You may want to set a short timeout value in `Course Configuration` for `Inactivity time before a user is required to login again` in some non-production course. If so, you should probably change it back after you finish testing.

2. Before applying the patches:
  a. Create a temporary local user with an empty string as the password and one with an all white-space password.
  b. Try logging in to those test accounts. The empty string password is expected to work, but not the all white-space password (due to the trimming).
  c. Test timeout behavior (even with local users), both the messages in the  course's `login.log` file and on the displayed login page. 
    * Login, wait for the timeout to pass, and click something.
    * Expect the login page on session timeout to have the generic authentication error message and **not** the timeout message.
3. Apply the patches and test again.
  * The empty password should now be rejected. 
  * The all white-space password should still be rejected.
  * Expect the login page to have the timeout error message.
4. If someone can test if this handles the LDAP session timeout issue from http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4529 it would be appreciated.
5. Clean up any settings/accounts which may be of trouble later.

mgage:
Hi Nathan,
 
Everything works perfectly with the revised Authen.pm file.
Local, LDAP, LTI accounts all time out with the correct message, and empty passwords are not accepted.
 
Thank you very much!
 
-------------
Andras

mgage:
Thanks for the testing and the report Andras.   Thanks very much Tani for the detective work in figuring out the problem.  I'll merge this PR.

PR #918, opened 2019-01-21
Leaderboard for comparing achievement points

drdrew42:
Based on an externally-funded student intern research project, this branch adds a leaderboard page when Achievements are enabled within a given course. (the sponsor message is necessary for display at our institution, but can be easily configured out if this feature is included in release)

Students select their own display names in "user settings" - which is stored in the comment field of the user table.

Student "progress" is currently measured relative to the highest score present in the course. Future updates will make the final column of the leaderboard table relative to the maximum possible score for the course (once I figure out how to raid the DB for all the values I need: total number of problems present in the course, even unassigned; and the sum of achievement points available for all enabled achievements).

One small pre-config is necessary: htdocs/js/apps/Leaderboard/leaderboard.php must be updated with a username and password for a mySQL account with read access to webwork. This php script checks that the request is coming from a user who is logged into WeBWorK before replying. I guess you should probably have php installed as well ;)


Alex-Jordan:
This sounds very cool!

The pull request includes changes to 2573 files. Almost all of these are not changes to file content, but rather changes of file permissions from 100644 → 100755. Could you make it so the pull request leaves out these permissions changes? I'm not knowledgeable enough with git and GitHub to suggest the best way to do that. Someone else could chime in, or maybe Google has an answer.

drdrew42:
closed. see rebased `[#919](https://github.com/openwebwork/webwork2/pull/919)`

PR #927, opened 2019-02-25
Develop candidate 02 24 2019

mgage:
This should replace PR #908.  It changes fewer files since I have already committed the bulk of the changes in to the copyright symbol.


taniwallach:
I made a PR to @mgage tree, to get things to work (at least in Docker) when a uft8mb4 database is used.
See https://github.com/mgage/webwork2/pull/24

There is a special `docker-compose.yml.specialUTF8MB4-db-storage` included which can be used to avoid the need to delete a current named Docker volume of SQL data by using a different container/volume name for the utf8mb4 testing.

With those changes (and the hacks to bypass the missing XML::Simple module in the current Docker image) I could (in a Docker test system):
  * creating and using the admin course in a new MariaDB 10.1 database which was set to use utf8mb4 as the default charset,
  * running OPL_update
    * Additional hacks were needed to get the global statistics to work.
    * See https://github.com/mgage/webwork2/pull/24 for details
  * creating a test course and creating a homework set with the question `setMAAtutorial/hello.pg` in it (selected as it takes a sting answer)
  * creating a user whose name needs UTF8 characters, and
  * submit a UTF8 encoded string as an answer to `setMAAtutorial/hello.pg` and later see the stored UTF8 answer in the course's past_answer table.

The testing was done when PG was using the branch from https://github.com/openwebwork/pg/pull/390

Overall, this seems a pretty good indication that things basically work with this set of UTF8 support for both webwork2 and pg.

I have been running a development server using the older (early 2019) branches with good success.

In my opinion, we won't find many more of the issues with the UTF8 support until more people are testing the code on servers which get a reasonable amount of usage.

mgage:
This is replaced by PR #930

mgage:
replaced by pr #930

mgage:
this is replaced by PR #930

PR #931, opened 2019-03-14
LTI authentication patch - to handle multi-byte characters

taniwallach:
This is a proposed patch to fix https://github.com/openwebwork/webwork2/issues/915 but has not yet been tested on a real system, as I am not using LTI in practice yet.

Testing instructions are given at https://github.com/openwebwork/webwork2/issues/915 .

**Pay attention** that the special setting `$allow_WW_LTI_code_to_fix_what_appear_as_multi_byte_characters = 1;`
needs to be enabled in `conf/authen_LTI.conf` to allow the patch to run. Without that variable defined and turned out the code is inactive. A commented out settings line is in the included version of `conf/authen_LTI.conf.dist`.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -116,6 +116,27 @@ $preferred_source_of_username = "lis_person_contact_email_primary";
 
 #$preferred_source_of_student_id = "ext_d2l_orgdefinedid";
 
+
+################################################################################
+# LTI - Should an attempt be made to "fix" multi-byte characters in the
+# $r->{paramcache} which would otherwise trigger errors in the Net:OAuth code.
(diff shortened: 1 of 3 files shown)


taniwallach:
From my testing with WW and the multilingual (UTF-8) support, this patch is **not** needed once the mail UTF-8 support is in use. 

I have a production server running with the multilingual (UTF-8) support, and LTI authentication works properly with UTF-8 characters in student names, and those names are added to the database properly.

If I understand correctly, it seems that the reason is that the main UTF-8 support assumes all parameters arrive in UTF-8 and applies Encode::decode_utf8() to all the parameters in `lib/WeBWorK/Request.pm` including those used by LTI authentication.

As a result, I am closing this PR.

PR #932, opened 2019-03-16
Develop multi ling 03 14 2019 mb4

mgage:
This is pretty similar to #930 but assuming multibyte capabilities of mysql.

There is a switch in site.conf.dist which turns this capability on (also an entry in localOverrides.conf).
If the switch is turned off this is compatible with older versions of mysql (before 5.3). If the switch is turned on it will build OPL databases that allow all characters.  It will allow all characters in new courses that are created after the switch is turned on.  It will not change courses build earlier. 

As far as I can tell the earlier courses will run exactly the same as before the switch to utf8mb4.  If one finds mangling of some characters with accents (mojibake)
then report this on issues.


lib/WeBWorK/Utils/ListingDB.pm
```diff
@@ -516,7 +516,8 @@ sub getDBListings {
 #               $kw2";
 
 	my $pg_id_ref;
-	$dbh->do(qq{SET NAMES 'utf8mb4';});
+	
+	$dbh->do(qq{SET NAMES 'utf8mb4';}) if $ce->{ENABLE_UTF8MB};
```
taniwallach: @mgage I think that when ENABLE_UTF8MB is not set, it should use 
```
$dbh->do(qq{SET NAMES 'utf8mb4';}) 
```
so this is probably best handled in an if/else stucture.

bin/OPL-update
```diff
@@ -85,19 +85,26 @@ use WeBWorK::Utils::Tags;
...
+            RaiseError => 1,
+			mysql_enable_utf8mb4 => $ENABLE_UTF8MB4,
         },
 );
 
-$dbh->prepare("SET NAMES 'utf8mb4'")->execute();
-
+if ($ENABLE_UTF8MB4) {
+	$dbh->prepare("SET NAMES 'utf8mb4'")->execute();
+}
```
taniwallach: I think then when the ENABLE_UTF8MB4 is not set, you probably want to run
```
$dbh->prepare("SET NAMES 'utf8'")->execute();
```

bin/update-OPL-statistics
```diff
@@ -79,9 +79,9 @@ CREATE TABLE IF NOT EXISTS `OPL_problem_user` (
   `attempted` int(11) DEFAULT NULL,
   `num_correct` int(11) DEFAULT NULL,
   `num_incorrect` int(11) DEFAULT NULL,
-  UNIQUE KEY `unique_key_idx` (`course_id`(100),`user_id`(100),`set_id`(100),`due_date`,`problem_id`),
-  KEY `source_file_idx` (`source_file`(255)))
-  CHARACTER SET ascii
+  UNIQUE KEY `unique_key_idx` (`course_id`(80),`user_id`(80),`set_id`(80),`due_date`,`problem_id`),
+  KEY `source_file_idx` (`source_file`(245)))
+  CHARACTER SET utf8mb4
```
taniwallach: As I recall, this branch is intended to have an override switch to avoid using uf8mb4 on systems whose mySQL version does not support it.
There is no override here to avoid utf8mb4 on a system where it was turned off. Maybe one was added in a later commit.

lib/WeBWorK/ContentGenerator/Grades.pm
```diff
@@ -109,7 +109,7 @@ sub scoring_info {
 	my $header = '';
 	local(*FILE);
 	if (-e "$filePath" and -r "$filePath") {
-		open FILE, "$filePath" || return("Can't open $filePath"); 
+		open FILE, "<:utf8", "$filePath" || return("Can't open $filePath");
```
taniwallach: Maybe we want `<:encoding(UTF-8)` here?

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -692,7 +692,7 @@ sub generate_hardcopy {
 	# create TeX file  (callback  write_multiuser_tex,  or ??)
 	#######################################
 	
-	my $open_result = open my $FH, ">", $tex_file_path;
+	my $open_result = open my $FH, ">:utf8", $tex_file_path;
```
taniwallach: Maybe we want `>:encoding(UTF-8)` here?

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -859,7 +859,7 @@ sub generate_hardcopy_pdf {
 		# read hardcopy.log and report first error
 		my $hardcopy_log = "$temp_dir_path/hardcopy.log";
 		if (-e $hardcopy_log) {
-			if (open my $LOG, "<", $hardcopy_log) {
+			if (open my $LOG, "<:utf8", $hardcopy_log) {
```
taniwallach: Maybe we want `<:encoding(UTF-8)` here?

lib/WeBWorK/ContentGenerator/Instructor/Config.pm
```diff
@@ -486,7 +487,7 @@ sub writeFile {
 	my $writeFileErrors;
 	eval {                                                          
 		local *OUTPUTFILE;
-		if( open OUTPUTFILE, ">", $outputFilePath) {
+		if( open OUTPUTFILE, ">utf8:", $outputFilePath) {
```
taniwallach: Maybe we want `>:encoding(UTF-8)` here?

lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
```diff
@@ -538,7 +539,7 @@ sub Save {
 	if (defined($data)) {
 		$data =~ s/\r\n?/\n/g;  # convert DOS and Mac line ends to unix
 		local (*OUTFILE);
-		if (open(OUTFILE,">$file")) {
+		if (open(OUTFILE,">:utf8",$file)) {
```
taniwallach: Maybe we want `>:encoding(UTF-8)` here and later on in this file?

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -1079,7 +1079,7 @@ sub saveFileChanges {
 
 		eval {
 			local *OUTPUTFILE;
-			open OUTPUTFILE,  ">$outputFilePath"
+			open OUTPUTFILE,  ">:utf8", $outputFilePath
```
taniwallach: Maybe we want `>:encoding(UTF-8)` here?

lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
```diff
@@ -775,7 +775,7 @@ sub readStandardCSV {
 
 sub writeStandardCSV {
 	my ($self, $filename, @csv) = @_;
-	open my $fh, ">", $filename;
+	open my $fh, ">:utf8", $filename;
```
taniwallach: Maybe we want `>:encoding(UTF-8)` here?

lib/WeBWorK/DB/Driver/SQL.pm
```diff
@@ -69,6 +69,7 @@ sub new($$$) {
 		{
 			PrintError => 0,
 			RaiseError => 1,
+			mysql_enable_utf8mb4 => 1,
```
taniwallach: The override switch should modity this to use `mysql_enable_utf8` instead of `mysql_enable_utf8mb4` when necessary. May be done in a later commit.

bin/update-OPL-statistics
```diff
@@ -72,7 +72,9 @@ my $dbh = DBI->connect(
 # get course list
 my @courses = listCourses($ce);
 
-# create tables.  We always redo the statistics table.  
+# create tables.  We always redo the statistics table. 
+
+if ($ENABLE_UTF8MB4){
```
taniwallach: I think that it might be simpler to set a variable to the desired character set (with the if/else), and just use that variable in the line which sets the table charset - something like  `CHARACTER SET $desired_charset` so that the table creation code does not need to appear twice.

bin/OPL-update
mgage: ok

bin/update-OPL-statistics
mgage: ok

lib/WeBWorK/ContentGenerator/Grades.pm
mgage: done

lib/WeBWorK/ContentGenerator/Hardcopy.pm
mgage: done

lib/WeBWorK/ContentGenerator/Hardcopy.pm
mgage:
```suggestion
			if (open my $LOG, "<:utf8", $hardcopy_log) {
```

ok

lib/WeBWorK/ContentGenerator/Instructor/Config.pm
mgage: done

lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
mgage: ok

lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
mgage: done

lib/WeBWorK/DB/Driver/SQL.pm
mgage: Done.  Thanks Tani.

bin/update-OPL-statistics
mgage: Done

lib/WeBWorK/Utils/ListingDB.pm
mgage: I tried something cleverer.  Maybe too clever -- we can take it out and put in an if/else statement.

mgage:
I've found that encoding things in base64 and then decoding them causes a known problem.  I'll post more in issue #933.  It causes things like knowls to fail (which means that Solutions also fails).  Most likely requires changes in both perl encode/decode and javascript encode/decode

mgage:
In docker these commits make such things as using wide characters for user names behave correctly, even with the original utf8 database (not the one set to utf8mb4).    Pulling this PR  to demo.webwork.rochester.edu does not fix the problem of entering user names with wide characters :-(.  even though the tables are correctly created to allow this (for the chinese2 course at least) -- Not sure what to check next -- there is probably some difference in the database configuration.

mgage:
on the docker version of the mysql database I see
mysql> SHOW VARIABLES WHERE Variable_name LIKE 'character\_set\_%' OR Variable_name LIKE 'collation%';
+--------------------------+--------------------+
| Variable_name            | Value              |
+--------------------------+--------------------+
| character_set_client     | utf8mb4            |
| character_set_connection | utf8mb4            |
| character_set_database   | latin1             |
| character_set_filesystem | binary             |
| character_set_results    | utf8mb4            |
| character_set_server     | utf8mb4            |
| character_set_system     | utf8               |
| collation_connection     | utf8mb4_general_ci |
| collation_database       | latin1_swedish_ci  |
| collation_server         | utf8mb4_general_ci |
+--------------------------+--------------------+
10 rows in set (0.01 sec)

on the demo .webwork version I get

mysql> SHOW VARIABLES WHERE Variable_name LIKE 'character\_set\_%' OR Variable_name LIKE 'collation%';
+--------------------------+-------------------+
| Variable_name            | Value             |
+--------------------------+-------------------+
| character_set_client     | utf8              |
| character_set_connection | utf8              |
| character_set_database   | utf8              |
| character_set_filesystem | binary            |
| character_set_results    | utf8              |
| character_set_server     | latin1            |
| character_set_system     | utf8              |
| collation_connection     | utf8_general_ci   |
| collation_database       | utf8_general_ci   |
| collation_server         | latin1_swedish_ci |
+--------------------------+-------------------+
10 rows in set (0.00 sec)


This would account for the different behavior.  Tani has suggested that I could override
the server system variables with something like SET NAMES 'utf8mb4' but I haven't been able to get
that to work yet.

Issue #924, opened 2019-02-05
Multi-part answers are not properly stored in Gateway Quizzes


dlglin:
Steps to reproduce:
Create a gateway quiz which includes a question that uses ans_array on a MathObject.  Enter an answer, then either preview, submit, or change pages (for multipage quizzes) in the quiz.  Both the last_answer field and the past_answers table will only contain the first element of the matrix.
See http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4549 for a MWE

It looks like this was fixed in homework mode (Problem.pm) with the creation of the process_and_log_answer function in ProblemUtil.pm, but the Gateway code was never updated to use this function.  Gateway code should probably be refactored to use the same function, but Gateways also store the type of submission (e.g. [submit], [preview] or [newpage]), so this would have to be considered, as well as testing the loading of past answers into answer blanks.

glennricster:
This also causes a problem if the Gateway Quiz is set to show some of the problems on the page at a time.  If you change pages the additional answers for a multi answer problem are lost.  If you switch back to the page with that multi answer problem, then the additional parts are blank, and when the exam is graded those parts are of course grades wrong.

mgage:
I believe that http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4598 (matrix answers disappearing) is probably related as well.

mgage:
Pull requests #935 and #936 resolve this issue.

PR #935, opened 2019-04-25
Fix extended response group questions in multi page gateway quizzes.

drgrice1:
If a problem uses an extended response group for its answers (like
multianswers with the single result option, answer arrays, or mathquill
answers) and a quiz has multiple pages, then the extended answers are
lost every time the page is changed.

The intended result is that the quiz would create a hidden input for
every given answer that is not on the current page.  It detects these
answers by looking for the quiz prefix in the input names.  However, it
was previously assumed that the quiz prefix would be at the beginning of
the input name, which is incorrect.  Extended response group answers add
a prefix in front of the the quiz prefix.

This fixes the problem by finding all inputs that contain the quiz
prefix, but excludes the previous answer inputs.

This fixes issue #924.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1538,7 +1538,7 @@ sub body {
 					# ref($pg) eq "WeBWorK::PG::Local";
 			} else {
 				my $prefix = sprintf('Q%04d_',$i+1);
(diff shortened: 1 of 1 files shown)


mgage:
I changed my mind.  This is a small enough change to be worthy of a hotfix. 
I checked it on develop and it resolves this problem.  I'll submit a pull request
to develop as well.

mgage:
Thanks very much Glen.

drgrice1:
Thank you!

PR #936, opened 2019-05-10
Fix extended response group questions in multi page gateway quizzes.

mgage:
If a problem uses an extended response group for its answers (like
multianswers with the single result option, answer arrays, or mathquill
answers) and a quiz has multiple pages, then the extended answers are
lost every time the page is changed.

The intended result is that the quiz would create a hidden input for
every given answer that is not on the current page.  It detects these
answers by looking for the quiz prefix in the input names.  However, it
was previously assumed that the quiz prefix would be at the beginning of
the input name, which is incorrect.  Extended response group answers add
a prefix in front of the the quiz prefix.

This fixes the problem by finding all inputs that contain the quiz
prefix, but excludes the previous answer inputs.

This fixes issue #924.

This is a submission from @drgrice1 originally made as a hotfix to master.  This PR is 
a duplicate fix cherry-picked from the original and rebased onto develop.


mgage:
This is the develop version of PR #935

Issue #939, opened 2019-05-16
Problem installing XMLRPC::Lite from CPAN (in docker)


heiderich:
I have problems building the webwork docker image, which seem to be caused by the CPAN package [XMLRPC::Lite](https://metacpan.org/pod/XMLRPC::Lite). Its installation sometimes fails. This seems to occur randomly; at least I have not figured out the exact circumstances yet. The following bug reports confirm this:

https://rt.cpan.org/Public/Bug/Display.html?id=127761
https://rt.cpan.org/Public/Bug/Display.html?id=92742

There is also a related bug report in Debian:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=924391

It is fixed in Debian:

https://salsa.debian.org/perl-team/modules/packages/libxmlrpc-lite-perl/commit/62dc58b3d8e07c9ab79fa96cc4d8198df534febe

However it seems that it is not fixed in CPAN.

A workaround might be to install the package not from CPAN, but using apt-get.

This could have been the cause for #903.

taniwallach:
@mgage  and I have also discussed this - it seems that the install from CPAN runs some tests which can fail when the servers being used for the test do not respond quickly enough...

We also came to conclusion that using the Debian package to install this package should hopefully avoid the annoying problem which sometimes breaks the Docker builds.

However, I think that so far the changes to Dockerfile may either be in one one of the internationalization development branches, on some private branch, or maybe only on one of our machines.

Sorry that I have not had time to test this and make a PR for develop and/or master.

heiderich:
@taniwallach Thanks for the confirmation.

mgage:
I had some similar problems.  My fix is in the branch multilingual on github.com/mgage/webwork2<http://github.com/mgage/webwork2> and is pull request #934.

Unless there are objections I hope to merge this #934 into develop fairly soon so that we don’t have too many development branches lying around.

I’m pretty confident that multilingual will not interfere with existing installations that use latin-1.  It also works well with fresh installations
using utf8mb4.  There remains the question of what measures (if any) need to be taken to switch existing courses using extended character
sets over to using the new multilingual branch.  Florian and Tani and others using extended character sets do you have any comments on this?

Take care,
Mike

On May 16, 2019, at 9:37 AM, Florian Heiderich <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #940, opened 2019-05-16
install XMLRPC::Lite from apt-get instead of CPAN

heiderich:
This pull request aims to solve #939 in installing XMLRPC::Lite using apt-get instead of CPAN. In addition I install `libxml-simple-perl` using apt-get, as it seems to be required.


mgage:
looks good

mgage:
I believe this resolves issue #939

PR #941, opened 2019-05-23
Use a fallback timezone instead of a die in lib/WeBWorK/Utils.pm

taniwallach:
Have `parseDateTime()` in `lib/WeBWorK/Utils.pm` fall back to using the value of `$display_tz` (possibly the local time-zone) as a fallback when the time-zone in the string being parsed would not be recognized by `tz_offset()`. This is meant to allow set definition files with an invalid `%Z` generated short time-zone code to be read in using the "display" time-zone, rather than the import being cancelled by the `die` which used to be called when this issue was encountered.

Note: The `reducedScoringDate` may be misinterpreted as a result, as when the included time-zone is dropped the time given in the file may not be the "correct" special value(the zero value of Unix time): `01/01/1970 at 12:00am`. I think this is a relatively small price to pay for getting the files to import, and manual corrections can be made to the various dates as needed.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
@@ -1986,10 +1986,29 @@ sub readSetDef {
 		}
 
 		# validate reduced credit date
-		$reducedScoringDate = $self->parseDateTime($reducedScoringDate) if ($reducedScoringDate);
+
+		# Special handling for values which seem to roughly correspond to epoch 0.
+		#    namely if the date string contains 12/31/1969 or 01/01/1970
+		if ($reducedScoringDate) {
+			if ( ( $reducedScoringDate =~ m+12/31/1969+ ) || ( $reducedScoringDate =~ m+01/01/1970+ ) ) {
+				my $origReducedScoringDate = $reducedScoringDate;
+				$reducedScoringDate = $self->parseDateTime($reducedScoringDate);
+				if ( $reducedScoringDate != 0 ) {
+					# In this case we want to treat it BY FORCE as if the value did correspond to epoch 0.
+					warn $r->maketext("The reduced credit date [_1] in the file probably was generated from the Unix epoch 0 value and is being treated as if it was Unix epoch 0.", $origReducedScoringDate );
+					$reducedScoringDate = 0;
+				}
+			} else {
+				# Original behavior, which may cause problems for some time-zones when epoch 0 was set and does not parse back to 0
+				$reducedScoringDate = $self->parseDateTime($reducedScoringDate);
+			}
+		}
 
 		if ($reducedScoringDate && ($reducedScoringDate < $time1 || $reducedScoringDate > $time2)) {
 		    warn $r->maketext("The reduced credit date should be between the open date [_1] and close date [_2]", $openDate, $dueDate);
+		} elsif ( $reducedScoringDate == 0 && $enableReducedScoring ne 'Y' ) {
+			# In this case - the date in the file was Unix epoch 0 (or treated as such),
+			# and unless $enableReducedScoring eq 'Y' we will leave it as 0.
 		} elsif (!$reducedScoringDate) {
 		    $reducedScoringDate = $time2 - 60*$r->{ce}->{pg}{ansEvalDefaults}{reducedScoringPeriod};
 		}
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -573,8 +573,26 @@ sub parseDateTime($;$) {
 			#warn "\t\$utc_epoch = $utc_epoch\n";
 
 			# get offset for supplied timezone and utc_epoch
-			my $offset = tz_offset($zone, $utc_epoch) or die "Time zone '$zone' not recognized.\n";
-			#warn "\t\$zone is valid according to Time::Zone (\$offset = $offset)\n";
+			# fall back to $display_tz if that fails
+			#my $offset = tz_offset($zone, $utc_epoch) or die "Time zone '$zone' not recognized.\n";
+
+			my $offset;
+			if( $offset = tz_offset($zone, $utc_epoch) ) {
+			  #warn "\t\$zone is valid according to Time::Zone (\$offset = $offset)\n";
+			} else {
+			  warn "Time zone '$zone' not recognized, falling back to using $display_tz.\n";
+			  $dt = new DateTime(
+			      year      => $year,
+			      month     => $month,
+			      day       => $day,
+			      hour      => $hour,
+			      minute    => $minute,
+			      second    => $second,
+			      time_zone => $display_tz,
+			      );
+			  $epoch = $dt->epoch;
+			  return $epoch;
+			}
 
 			#$epoch = $utc_epoch + $offset;
 			##warn "\t\$epoch = \$utc_epoch + \$offset = $epoch\n";
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -533,92 +533,75 @@ sub parseDateTime($;$) {
 		}
 	}
 
-	my $epoch;
+	my $epoch; # The value we need to calculate and return
 
+	# Determine the best possible time-zone string to use in the (first) call to DateTime()
+	my $tz_to_use = $display_tz;
+	my $is_valid_zone_name = 1; # when later set to 0, we will try the "offset" approach
 	if (defined $zone and $zone ne "") {
-		if (DateTime::TimeZone->is_valid_name($zone)) {
-			#warn "\t\$zone is valid according to DateTime::TimeZone\n";
-
-			my $dt = new DateTime(
-				year      => $year,
-				month     => $month,
-				day       => $day,
-				hour      => $hour,
-				minute    => $minute,
-				second    => $second,
-				time_zone => $zone,
-			);
-			#warn "\t\$dt = ", $dt->strftime(DATE_FORMAT), "\n";
-
-			$epoch = $dt->epoch;
-			#warn "\t\$dt->epoch = $epoch\n";
+		$is_valid_zone_name = DateTime::TimeZone->is_valid_name($zone);
+		if ( $is_valid_zone_name ) {
+			#warn "\t\$zone=$zone is valid according to DateTime::TimeZone\n";
+			$tz_to_use = $zone;
 		} else {
-			#warn "\t\$zone is invalid according to DateTime::TimeZone, so we ask Time::Zone\n";
-
-			# treat the date/time as UTC
-			my $dt = new DateTime(
-				year      => $year,
-				month     => $month,
-				day       => $day,
-				hour      => $hour,
-				minute    => $minute,
-				second    => $second,
-				time_zone => "UTC",
-			);
-			#warn "\t\$dt = ", $dt->strftime(DATE_FORMAT), "\n";
-
-			# convert to an epoch value
-			my $utc_epoch = $dt->epoch
-				or die "Date/time '$string' not representable as an epoch. Get more bits!\n";
-			#warn "\t\$utc_epoch = $utc_epoch\n";
-
-			# get offset for supplied timezone and utc_epoch
-			# fall back to $display_tz if that fails
-			#my $offset = tz_offset($zone, $utc_epoch) or die "Time zone '$zone' not recognized.\n";
-
-			my $offset;
-			if( $offset = tz_offset($zone, $utc_epoch) ) {
-				#warn "\t\$zone is valid according to Time::Zone (\$offset = $offset)\n";
-			} else {
-				warn "Time zone '$zone' not recognized, falling back to using $display_tz.\n";
-				$dt = new DateTime(
-					year      => $year,
-					month     => $month,
-					day       => $day,
-					hour      => $hour,
-					minute    => $minute,
-					second    => $second,
-					time_zone => $display_tz,
-				);
-				$epoch = $dt->epoch;
-				return $epoch;
-			}
-
-			#$epoch = $utc_epoch + $offset;
-			##warn "\t\$epoch = \$utc_epoch + \$offset = $epoch\n";
-
-			$dt->subtract(seconds => $offset);
-			#warn "\t\$dt - \$offset = ", $dt->strftime(DATE_FORMAT), "\n";
-
-			$epoch = $dt->epoch;
-			#warn "\t\$epoch = $epoch\n";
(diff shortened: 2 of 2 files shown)


mgage:
Let's get this reviewed and committed to develop soon so that I can use develop to update WeBWorK-2.15 and get these time issues fixed in the new multilingual version of WeBWorK.

drgrice1:
This looks good to me.

A minor suggested improvement though.  The indentation should be fixed.  Tab characters should be used for indentation in the white space at the beginning of a line.

taniwallach:
@drgrice1 - indentation fixed as requested.
BTW I don't think there is much of a standard about coding style in WW, but most of this file does use tabs for the line-start indentation, but not all.

taniwallach:
From what I can see, since the changes made in https://github.com/openwebwork/webwork2/pull/900 (which was merged into WW 2.14), it seems that when `parseDateTime()` is called it is very rare that the "date+time" string sent to be parsed by `parseDateTime()` will contain a time zone string.

It seems to me that this will **only** occur when set definitions files are being read, which is exactly the case where a `die` would cause an import failure when the `%Z` generated "short time zone string" could not be parsed. The fallback code in this PR avoids the fatal `die` error by using a fall-back value for the time-zone.

(A long term fix might be to use a "full" time-zone string in set definition files, but old files with "bad" short time-zone strings would still need the fix in this PR.)

Since this is still under review, below is an overview of all the WW code I can find which uses the function being modified, and how I came to the conclusion that the only expected behavior expected to change is the parsing of set definition files.

  1. `lib/WeBWorK/Utils.pm` has a function called `intDateTime()` which will call `parseDateTime()` when the input is "not an integer" (as string containing only digits or an empty string). When this is done it does **not** set a `$display_tz` but I could not find any current use of this function.
  2. `lib/WeBWorK/ContentGenerator.pm` defines a local method `parseDateTime()` which sets `$display_tz` using `$ce->{siteDefaults}{timezone}` if it was not sent as an argument in the method call and the calls the function of the same name from `lib/WeBWorK/Utils.pm` with both argument set. 
  3. All the following are calling the "higher level" `parseDateTime()` method from `lib/WeBWorK/ContentGenerator.pm` and using that to call the function in `lib/WeBWorK/Utils.pm`.
     - Code which uses $self->parseDateTime() with only a single argument (not explicitly sending a value for `$display_tz`):
        - `lib/WeBWorK/ContentGenerator/Instructor/UserDetail.pm` (1 call)
          - in the internal `checkDates()` function, used before updating database with override dates for a specific user. Those dates are coming from a web form and do not have a timezone string.
        - `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm`(2 calls)
          - in `saveEdit_handler()` on `_date` fields sent from a web form (which do not have time zone strings)
          - in `readSetDef()` which parses the set definition files.
            - **This is the case where problems occur and need to be fixed.**
            - (this version of the code does not handle an input value for `$reducedScoringDate`)
        - `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm` (5 calls, 1 explicit sets `$display_tz` only on call on line 1154)
          - in `create_handler()` to set an initial (proposed) due date for a new local homework set  (this one has an explicit inclusion of `$display_tz`) and the string being parsed does not have a time zone string (that was the change made in https://github.com/openwebwork/webwork2/pull/900)
          - in `import_handler()` to parse `action.import.start.date` sent from a web form (which does not have a time zone string)
          - in `saveEdit_handler()` on `_date` fields sent from a web form (which do not have time zone strings)
          - in `readSetDef()` which parses the set definition files. (twice, the second one to parse `$reducedScoringDate`)
            - **This is the case where problems occur and need to be fixed.**
        - `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm` (3 calls)
          - when parsing date fields which all come from a web form (and thus have no time zone string)
        - `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail2.pm` (3 calls)
          - when parsing date fields which all come from a web form (and thus have no time zone string)
      - Code which uses `$self->parseDateTime()` with `$display_tz` explicity set from `$ce->{siteDefaults}{timezone}`:  
        - `lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm` (1 call)
          - when setting the initial (proposed) due date for a new local homework set (handling `new_local_set`) and the string being parsed does not have a time zone string (that was the change made in https://github.com/openwebwork/webwork2/pull/900)
        - `lib/WeBWorK/ContentGenerator/Instructor/SetMakernojs.pm` (1 call)
          - when setting the initial (proposed) due date for a new local homework set  (handling `new_local_set`) and the string being parsed does not have a time zone string (that was the change made in https://github.com/openwebwork/webwork2/pull/900)
        - `lib/WeBWorK/ContentGenerator/Instructor/GetLibrarySetProblems.pm`
          - when setting the initial (proposed) due date for a new local homework set  (handling `new_local_set`) and the string being parsed does not have a time zone string (that was the change made in https://github.com/openwebwork/webwork2/pull/900)
        - `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm`
           - See above. The call in `create_handler()` sets `$display_tz` explicitly.

taniwallach:
I am trying to see if I can add a change to better handle the potential issue with the "reduced scoring date" when there are time-zone parsing issues. My intention is to modify the handling when the string in a set definition file has either `12/31/1969` or `01/01/1970` as part of the string. I should have an extra commit added to the PR soon to handle this.

taniwallach:
Sorry to make changes after a review, but having come back to this, I decide to try to fix the "small issue" with how the original PR could potentially mishandle "Unix epoch 0" dates from a set definition file (SDF). The expected location is as the value of `reducedScoringDate` in these files.

One part of the third commit attempts to detect `reducedScoringDate` values which are "roughly" Unix epoch 0 (has either 12/31/1969 or 01/01/1970 as part of string value) and forcibly treats them as Unix epoch 0. As it is quite unlikely and unreasonable that a SDF will contain some other time on those dates as the value of `reducedScoringDate` this seems quite reasonable.

While testing the change, I noticed that when the `reducedScoringDate` parses to 0 (Unix epoch 0), the code would always redefined the value based on the deadline and the course value of `reducedScoringPeriod`. I do not think this automatic change should be made unless the SDF sets `enableReducedScoring` to "Y". (A regular assignment in a course without reduced scoring will have Unix epoch 0 as the value of the `reducedScoringDate` while the current behavior for importing SDF will set it differently.)

As a result, the third commit also adds a change to the handling to leave Unix epoch 0 as the value of the `reducedScoringDate` unless the SDF sets `enableReducedScoring` to "Y".

This change may surprise people who are used to the old behavior which automatically set a new date/time for `reducedScoringDate` and could then only modify the set to turn on reduced scoring without needed to set the `reducedScoringDate`. However, just editing the assignment and marking the box to enable reduced scoring and leaving the "Unix epoch 0" time value and saving the change ("Take action") will trigger an automatic calculation of the `reducedScoringDate` using the same approach (shifting the deadline by `reducedScoringPeriod`).

If the removal of automatic override of the `reducedScoringDate` unless the SDF sets `enableReducedScoring` to "Y" is objectionable, that change can easily be removed.

taniwallach:
@drgrice1 - If you don't mind, please have a look at the changes in the additional commit, and give your feedback about the change to how the `reducedScoringDate` is handled before/after the new changes.

lib/WeBWorK/Utils.pm
```diff
@@ -573,8 +573,26 @@ sub parseDateTime($;$) {
...
+					year      => $year,
+					month     => $month,
+					day       => $day,
+					hour      => $hour,
+					minute    => $minute,
+					second    => $second,
+					time_zone => $display_tz,
+				);
+				$epoch = $dt->epoch;
+				return $epoch;
```
drgrice1: Another comment I didn't make before.  It would be a good idea while editing this section to take the opportunity to clean things up a bit.  The "$epoch = $dt->epoch" (and commented out warn) are common to each case.  Move that to the end of the parseDateTime method with the return statement to reduce repetition.  You will need to bump up the scope of the $dt variable and move lines 600 and 601 into the "if( $offset = tz_offset...)" part.

This is non-essential and others may not agree with this, but it is little clean up like this that could greatly improve the readability of the WeBWorK code base.

lib/WeBWorK/Utils.pm
```diff
@@ -573,8 +573,26 @@ sub parseDateTime($;$) {
 			#warn "\t\$utc_epoch = $utc_epoch\n";
 
 			# get offset for supplied timezone and utc_epoch
-			my $offset = tz_offset($zone, $utc_epoch) or die "Time zone '$zone' not recognized.\n";
-			#warn "\t\$zone is valid according to Time::Zone (\$offset = $offset)\n";
+			# fall back to $display_tz if that fails
+			#my $offset = tz_offset($zone, $utc_epoch) or die "Time zone '$zone' not recognized.\n";
```
drgrice1: This is something I should have mentioned before.  I think that the old code that you are replacing should be deleted, not commented out.  This is the point of a versioning control system like git.  The old code is all still there.  This reduces clutter in the current code.

While you are at it delete the obsolete (and commented out) lines 597 and 598.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
```diff
@@ -1986,10 +1986,29 @@ sub readSetDef {
...
+				if ( $reducedScoringDate != 0 ) {
+					# In this case we want to treat it BY FORCE as if the value did correspond to epoch 0.
+					warn $r->maketext("The reduced credit date [_1] in the file probably was generated from the Unix epoch 0 value and is being treated as if it was Unix epoch 0.", $origReducedScoringDate );
+					$reducedScoringDate = 0;
+				}
+			} else {
+				# Original behavior, which may cause problems for some time-zones when epoch 0 was set and does not parse back to 0
+				$reducedScoringDate = $self->parseDateTime($reducedScoringDate);
+			}
+		}
```
drgrice1: One thing I am not clear on is the exact situation where this is really needed.  I have tested several "malformed" date/time strings in set definition import files both with and without this pull request, and don't really see the difference.  In most cases the import still fails with a "die" from the unformatDateTime call that precedes your changes in the Utils.pm file.  Can you give a precise example that demonstrates what this is really getting at?

taniwallach:
Below is a sample SDF which triggers the `die` on import, but the key to this are the date lines which use `IDT` and `IST` as short versions of Israel Daylight/Standard time, so just setting those date strings in any SDF will cause the problem.

Based on a test I ran and mentioned at  https://github.com/openwebwork/webwork2/pull/900#issuecomment-440041828 about half of the time zones Perl knows about will have similar problems with their "%Z" short time-zone code.

---

```
assignmentType      = default
openDate          = 04/25/2019 at 11:59pm IDT
reducedScoringDate = 01/01/1970 at 02:01am IST
dueDate           = 05/22/2019 at 11:59pm IDT
answerDate        = 05/24/2019 at 11:59pm IDT
enableReducedScoring = N
paperHeaderFile   = defaultHeader
screenHeaderFile  = defaultHeader
description       = 
restrictProbProgression = 0
emailInstructor   = 0

problemListV2 
problem_start
problem_id = 1
source_file = setDemo/c4s5p2.pg
value = 1
max_attempts = -1
showMeAnother = -1
prPeriod = -1
counts_parent_grade = 0
att_to_open_children = 0 
problem_end
```

taniwallach:
The use of
```
reducedScoringDate = 01/01/1970 at 02:01am IST
```
above will also trigger the special behavior about Unix epoch 0 (even were you to set Asia/Jerusalem as the course time zone) as it is slightly off from the epoch. A SDF with the correct epoch for a timezone whose short form is invalid imported using the course time zone will also trigger it.

Changing the value of `enableReducedScoring` from `N` to `Y` will allow the modified import code to auto-set the `reducedScoringDate` while leaving it on `N` will have the code leave it set to epoch 0.

taniwallach:
@drgrice1  - I'll try to clean up the code a bit later... Much commented out code exists in the code-base and for people not using Git can help understand what was changed. But I think you are correct, this sort of change probably warrants just overwriting the old code, and this is a chance to clean out some other leftovers.

taniwallach:
Regarding testing of how the `reducedScoringDate` is being handled - turn on reduced scoring (`Enable Reduced Scoring`)and set a non-zero value for `reducedScoringPeriod` (`Default Length of Reduced Scoring Period in minutes`) in the `Optional Modules` part of `Course Configuration`.

Pay attention to the value set for the `Reduced Scoring Date` of the assignment after import.

drgrice1:
I see.  Thanks for the clarification.  I have been testing with American time zones (or with no time zone).  Now I see what is going on.

taniwallach:
@drgrice1 - Code was cleaned up a bit and seems to still work for me, both when the fallback is needed and when it is not... but now it certainly needs another round of review and testing. :smile:

And, yes, `$tz_to_use = $display_tz;` need not be set again in the case where no zone was supplied, since it was set as a default value when the variable was declared (`my`) but it seems to me more readable if it is there (or there and commented out).

drgrice1:
You certainly went more aggressive on code clean up than I was getting at.  I just meant to move the last two repetitive lines (lines 553-554, 585-586, and 602-603 in the master branch) out of the if/then/else statement.  I wasn't looking for the code to be restructured so much.

In any case, it looks like what you have done is okay and works.  Although I don't like the excessively long lines of code (most notably lines 547 and 587).  I know there are a few other long lines in the file, but there shouldn't be.  It is just plain ugly.  I have a wide screen monitor, but I certainly don't want my text editor to actually use the kind of real estate needed to prevent line wrapping with 160 or more characters.  A good rule of thumb is lines should not exceed 100 characters (although others might say less than that).

mgage:
Thanks Glenn and Tani for all the work you've done on this.  I support the idea of cleaning out the cruft in the code.  The practice of leaving commented out old code was ingrained during the pre git/github days and for the most part it's time to abandon it.  

About style.  Sam Hathaway (at the time an undergrad at U. of R) tried to set down some coding standards but for the most part we've been sloppy about following them.  
http://webwork.maa.org/wiki/Coding_Standards

I attended the North American Perl Conference last week in Pittsburgh.  I learned about a number of new tools which together with the commit hooks provided by github would make some of the unit testing and style enforcement automatic or at least semi-automatic.  I wouldn't want to make changes for this release but I'm planning to investigate the modules Perl::Tidy and Perl::Critic and either encourage their use or use them automatically in github commit hooks or both.  

I also want to design and set up a Test::Harness commit hook that will run at least some automatic tests.  We have been slowly collecting useful unit test files but they are run somewhat erratically and having a standard test harness will encourage more unit tests and help us to keep them in order.

I still invite further tests of the time-zone and DateTime fixes to make sure we haven't missed anything.  Thanks again Glenn and Tani for your  work on this.

mgage:
I'm ready to merge this PR.  I don't want it dangling outside the webwork-2.15 track for too long. If there are remaining issues I don't believe they will be deal stoppers and we can patch them later.

PR #954, opened 2019-06-11
Revert to base class title behaviour in Login.pm

jjrsylvestre:
Previously login page would display course title instead of just URL path. This change reverts to that behaviour.


mgage:
This is the right way to handle title -- don't override SUPER unless necessary.  However, according to my ad hoc tests the base version of title() is broken in ContentManager.pm since the advent of the `maketext()` conversion of languages.  

I think `title()` is a good pilot project to start creating unit tests. We haven't gotten into the habit of creating unit tests.

mgage:
Thanks.

PR #965, opened 2019-06-26
make sure all the minor commits to develop are incorporated into WW-2.15

mgage:
This should clean things up before any major features are added to develop and while WW-2.15 is still in beta.


taniwallach:
Keeping two branches (develop and WeBWorK-2.15) under active development and in sync is going to create extra overhead.

I think it may help to carefully decide which branch "new" changes get merged into in the coming few weeks, and maybe not to try to keep them in sync.

I tend to think that anything targeted for inclusion in WW 2.15 should be put in that branch at present, while develop at present should get changes which are not currently planned for inclusion in WW 2.15. Then when WW 2.15 stabilizes and "freezes", it can be merged into develop and master.

What do people think of this idea?

PR #956, opened 2019-06-20
Changes to the XMLRPC system to support UTF-8 encoded problems

taniwallach:
Changes to the XMLRPC system to allow it to handle UTF-8 encoded problems.

Rendering works already. However, submitting UTF-8 encoded text which is not English (Latin1) as form values (in particular as answer) is not working yet. Due to this, the values set on the submit buttons are currently forced back into English.

`clients/sendXMLRPC.pl` was extended to take a language setting as an argument.

A sample Hebrew UTF-8 problem was added as `clients/t/test-utf8-hebrew.pg` to allow testing. On my system, testing using
  - `./sendXMLRPC.pl -b -f simple t/test-utf8-hebrew.pg`
  - `./sendXMLRPC.pl -b -l heb -f simple t/test-utf8-hebrew.pg`
both give reasonable behavior, so long as no special characters are used in the answers.

`lib/WebworkClient.pm` contains code to set the HTML language and text direction both at the level of the "page" and at the level of the DIV element enclosing a problem (based on data set in the PG file).

`lib/WeBWorK/Localize.pm` received a new method `getLangHandle()` used by `lib/WebworkClient.pm ` to get access to maketext.

`lib/WeBWorK/Utils/DetermineProblemLangAndDirection.pm` was extended to allow forcing options not available via a request hash when the XMLRPC system is rendering problems.

Small related changes and some cleanup work was done to several of the formats in `lib/WebworkClient/` including changes to support UTF-8, to standardize the title setting, to the copyright dates. Additional settings were added to the <form> element and to the "problem-content"
DIV element to be more similar to what is generated by the regular ContentGenerator system. `forcePortNumber` was added as a new hidden field. The value settings of the submit buttons are set by variables so they can get localized text. At present, this is not in active use as UTF-8 values render properly but then cause problems when the buttons are clicked and the non-English values are sent as form data.

This PR also includes changes to allow forcing a port number to be used by the XMLRPC system. This is intended to help when using the XMLRPC system with a Docker container or a host which requires using a port number for the site_url and form_action_url. This is managed by setting the value of `forcePortNumber` in the credentials and/or in the address used to access `html2xml`. 
Ex: `http://localhost:8080/webwork2/html2xml?&answersSubmitted=0&language=en&sourceFilePath=some_pg_file.pg&problemSeed=123567890&displayMode=MathJax&courseID=daemon_course&userID=daemon&course_password=daemon&outputformat=simple&forcePortNumber=8080`

PR #966, opened 2019-06-26
Add json xmlrpc output format

taniwallach:
This is an early version of a new JSON output format for the XMLRPC system, including code to allow `clients/sendXMLRPC.pl` to request the json format output be downloaded and saved to a file.

The long term goal is to provide the data in a format which allows different systems to embed WeBWorK problems inside their pages in a "natural" manner without use of an iframe, by acting as a "man in the middle" between the student using the other system and a "back-end" WW server rendering problems and grading answers.

The first target platform is Open edX, and someone else has started to design an "XBlock" plugin for Open edX to interface to WeBWorK using the XMLRPC system and the JSON output format. Most likely, a more focused subset of the current JSON output will be added later, as well as a method to handle the XMLRPC required additional fields being submitted in a manner which keeps them separate from the "regular" form fields required by a WW problem.

Essentially the code builds a JSON structure which the (majority of the) data which would be in an HTML page from the simple format broken down into small pieces:
  - the hidden input fields usually added to the HTML form to support the use on the XMLRPC system are instead given in key-value pairs in the JSON object whose key name starts with `hidden_input_`,
  - `internal_problem_lang_and_dir` contains the tags usually added to a DIV element enclosing a problem which allows PG problems to set their text language and direction. This saved the other system from needing to parse and extract these settings.
  - `real_webwork_FORM_ACTION_URL` and `real_webwork_XML_URL` are the values for the relevant variables which are usually set in the HTML output. However, as the intermediate system will be handing the form as a "man in the middle" - these addresses are needed by the intermediate system and not the student. Placeholders exist instead of the regular values in the HTML code in the relevant JSON fields.
  - `score` is an integer version of the score (0-100 unless a PG problem bug gives a value outside this range).

Due to the need to construct the JSON structure to be output after variable interpolation, the handling in  `lib/WebworkClient.pm` is somewhat more complex than for the other formats.

A few additional output file handles were fixed to use UTF-8, which should really have been done to them as part of https://github.com/openwebwork/webwork2/pull/956

### DIFF
--- a/clients/sendXMLRPC.pl
+++ b/clients/sendXMLRPC.pl
@@ -205,6 +205,9 @@ on the same computer but does require an internet connection to a remote WeBWorK
 
 	Process question in TeX mode, convert to PDF and display.
 	
+=item   --json
+
+	Process question in JSON mode and save to file
 =item   
 
 	The single letter options can be "bundled" e.g.  -vcCbB
@@ -340,6 +343,7 @@ my $lang = 'en';
 my $edit_source_file = '';
 my $display_tex_output='';
 my $display_pdf_output='';
+my $display_json_output='';
 my $print_answer_hash;
 my $print_answer_group;
 my $print_pg_hash;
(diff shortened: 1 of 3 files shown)


mgage:
OK.  This works for me.  sendXMLRPC is not widely used yet except by experts so even if there are minor bugs I don't think it will be a problem.

PR #968, opened 2019-06-27
Add psvn to xmlrpc

taniwallach:
Add `psvn` to `clients/sendXMLRPC.pl` and `lib/WebworkClient.pm` and related files.

This allows these interfaces to control the value of `psvn` which is sometimes used to reseed the random number generator inside a PG file. The use case is to handle groups of problems which need to have the same seed / random numbers.

Along the way a minor issue which would trigger the warning message:
```
Use of uninitialized value $WebworkClient::site_url in concatenation (.) or string at PATH_TO_WW2/webwork2/lib/WebworkClient.pm line 167.
```
when using `clients/sendXMLRPC.pl` was fixed.

A test file was added.

Sample test from inside `webwork2/clients` with credential set up:
```
./sendXMLRPC.pl -b --psvn=123987 t/show-psvn.pg
```

### DIFF
--- a/clients/sendXMLRPC.pl
+++ b/clients/sendXMLRPC.pl
@@ -258,6 +258,10 @@ on the same computer but does require an internet connection to a remote WeBWorK
                  
        Sets problemSeed to the number contained in string s
 
+=item  --psvn=s
+
+       Sets psvn to the number contained in string s
+
 
 
 =back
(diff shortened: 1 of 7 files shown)


mgage:
Looks ok to me and it worked on the tests that I tried or at least it didn't cause any trouble.  Aside from the test problem I haven't tried to use psvn for a meaningful problem sequence.

PR #971, opened 2019-07-12
Add Caliper

### LINKED ISSUE #970: Add Caliper Analytics support
I’m been working on instrumenting [Caliper](https://www.imsglobal.org/activity/caliper), which is useful for capturing learning analytics events for later analysis, into WeBWorK for UBC (see here: https://github.com/ubc/webwork2/pull/22). I wanted to see if there was interest in adding it upstream to WeBWorK core since it might be useful for other institutions' Learning Analytic needs. 

I've Completed hooking up events for sign in/out ([Session Profile](https://www.imsglobal.org/sites/default/files/caliper/v1p1/caliper-spec-v1p1/caliper-spec-v1p1.html#sessionProfile)) and answer submission for normal/gateway problem sets ([Assessment Profile](https://www.imsglobal.org/sites/default/files/caliper/v1p1/caliper-spec-v1p1/caliper-spec-v1p1.html#assessmentProfile)). 

I'll put up a PR shortly for your consideration (see #971)

andrew-gardener:
- Capture answer submissions (Assessment Submitted & AssessmentItem Completed events) for normal and gateway problem sets
- Attempt durations for gateway problem sets are only stored in the overall Assessment (not individual assessment item completed events) due to multiple problems being on a page at a time
- Additionally capture Assessment Paused event for gateway problems sets (sent when changing pages to help track overall time spend on task)
- Capture login/logout events

Added new `HTTP::Async`  (libnet-https-nb-perl & libhttp-async-perl) dependency so that events can be emitted asynchronous (very important for gateway problems sets since many AssessmentItem Completed events are created at once)

Closes #970

### DIFF
--- a/lib/Caliper/Actor.pm
+++ b/lib/Caliper/Actor.pm
@@ -13,7 +13,7 @@ use Caliper::ResourceIri;
 sub generate_anonymous_actor
 {
 	return {
-		'id' => 'http://purl.imsglobal.org/ctx/caliper/v1p1/Person',
+		'id' => 'http://purl.imsglobal.org/caliper/Person',
 		'type' => 'Person',
 	};
 }
--- a/lib/Caliper/Actor.pm
+++ b/lib/Caliper/Actor.pm
@@ -1,0 +1,50 @@
+package Caliper::Actor;
+
+##### Library Imports #####
+use strict;
+use warnings;
+use WeBWorK::CourseEnvironment;
+use WeBWorK::DB;
+use WeBWorK::Debug;
+use Data::Dumper;
+
+use Caliper::ResourceIri;
+
+sub generate_anonymous_actor
+{
+	return {
+		'id' => 'http://purl.imsglobal.org/ctx/caliper/v1p1/Person',
+		'type' => 'Person',
+	};
+}
+
+sub generate_default_actor
+{
+	my ($ce, $db, $user) = @_;
+	my $resource_iri = Caliper::ResourseIri->new($ce);
+
+	return {
+		'id' => $resource_iri->actor_homepage($user->user_id()),
+		'type' => 'Person',
+		'name' => $user->first_name() . " " . $user->last_name(),
+	};
+}
+
+sub generate_actor
+{
+	my ($ce, $db, $user_id) = @_;
+
+	if (!defined($user_id)) {
+		return Caliper::Entity::generate_anonymous_actor();
+	} else {
+		my $user = $db->getUser($user_id);
+
+		if (defined($ce->{caliper}{custom_actor_generator})) {
+			return $ce->{caliper}{custom_actor_generator}($ce, $db, $user);
+		} else {
+			return Caliper::Entity::generate_default_actor($ce, $db, $user);
+		}
+	}
+}
+
+1;
--- a/Dockerfile
+++ b/Dockerfile
@@ -145,6 +145,8 @@ RUN apt-get update \
 	libpath-class-perl \
 	libphp-serialization-perl \
 	libxml-simple-perl \
+	libnet-https-nb-perl \
+	libhttp-async-perl \
 	libsoap-lite-perl \
 	libsql-abstract-perl \
 	libstring-shellquote-perl \
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -122,6 +122,7 @@ my @modulesList = qw(
 	XML::Writer
 	XMLRPC::Lite
 	YAML
+    HTTP::Async
 );
 
 my %moduleVersion = (
--- a/conf/localOverrides.conf.dist
+++ b/conf/localOverrides.conf.dist
@@ -545,4 +545,32 @@ $hardcopyThemeNames = {
 
 ################################################################################
 
+
+################################################################################
+# Webwork Caliper
+################################################################################
+
+# enable/disable Caliper for install
+$caliper{enabled} = 0;
+# base_url should ideally be hard coded to a persistent url pointing to the webwork root
+# (important to keep it consistent over time)
+$caliper{base_url} = 'https://webwork.elearning.ubc.ca/webwork2/';
+# LRS endpoint
+$caliper{host} = 'http://caliper.example.host.org/api/endpoint';
+# LRS endpoint Bearer API key
+$caliper{api_key} = '1234567890abcdefg';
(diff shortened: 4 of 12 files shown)


taniwallach:
@andrew-gardener 

  1. Would it be reasonable/possible to avoid including plain-text "personal identity" information in what is sent to Caliper, and instead use some sort of UUID or a one-way hash of the PII? 
    - That would be better in terms of protecting privacy. 
    - It could be that WW could keep a site-wide database of UUID codes for each "student" based on some standard fields (Last name, first name, student ID, and maybe email address)
  2. I think that the Ubuntu package `libhttp-async-perl` provide the Perl package, and adding it via the Ubuntu package seems better than using `cpanm` to install it. See https://ubuntu.pkgs.org/16.04/ubuntu-universe-amd64/libhttp-async-perl_0.30-1_all.deb.html  and  https://ubuntu.pkgs.org/18.04/ubuntu-universe-amd64/libhttp-async-perl_0.33-1_all.deb.html

andrew-gardener:
> Would it be reasonable/possible to avoid including plain-text "personal identity" information in what is sent to Caliper, and instead use some sort of UUID or a one-way hash of the PII?

I've added the ability to fully customize the actor piece
see: https://github.com/openwebwork/webwork2/blob/98e59087e5f902f7cad12bb086583dc637f67897/lib/Caliper/Actor.pm#L42 and https://github.com/openwebwork/webwork2/blob/98e59087e5f902f7cad12bb086583dc637f67897/conf/localOverrides.conf.dist#L531

In UBC's case, we use this to set the actor's id to something like `http://www.ubc.ca/PUID` so it's easier for query in the LRS later on but you can set it to whatever your situation requires. 
You can also do the anonymous user with id `http://purl.imsglobal.org/ctx/caliper/v1p1/Person` and type `Person` (with no other information) but keep in mind that with access to WW's database it would be possible to link certain events to students (ex answer_ids from answers). 

> I think that the Ubuntu package libhttp-async-perl provide the Perl package, and adding it via the Ubuntu package seems better than using cpanm to install it.

I'll take a look at it

andrew-gardener:
@taniwallach I've added both libnet-https-nb-perl & libhttp-async-perl instead of installing through cpanm (Net::HTTPS::NB is a requirement for the async library)

andrew-gardener:
Note: updated this to work with the upcoming Caliper 1.2

mgage:
I don't have a set up that uses caliper so I can't test that.  I have verified that it does no harm to existing sites. (Verified both by inspecting the code changes and also by running it locally without using any of the caliper features.)

mgage:
@andrew-gardener   

Line 45 in Caliper::Actor  throws an error:
return Caliper::Entity::generate_default_actor($ce, $db, $user);

(unknown subroutine).  Replacing this with either Caliper::Actor::generate_default_actor or 
just generate_default_actor removes the error message -- and is probably what was meant.

I'm doing more testing with Caliper turned on but not configured.  If you can give me some suggestions for a default configuration I can do more tests. (I have moodle on demo.webwork.rochester.edu if that is a place to post caliper results.) Let me know.

lib/Caliper/Actor.pm
```diff
@@ -0,0 +1,50 @@
...
+
+	if (!defined($user_id)) {
+		return Caliper::Entity::generate_anonymous_actor();
+	} else {
+		my $user = $db->getUser($user_id);
+
+		if (defined($ce->{caliper}{custom_actor_generator})) {
+			return $ce->{caliper}{custom_actor_generator}($ce, $db, $user);
+		} else {
+			return Caliper::Entity::generate_default_actor($ce, $db, $user);
```
mgage: This throws an error.  It should probably be just generate_default_actor() since that subroutine is in the same package.

mgage: I'm going to pull this and then make the obvious fix to generate_default_actor()

andrew-gardener:
You can test it out using IMS Global's Caliper test server: https://caliper.imsglobal.org/sec/#test/server . It will give you a endpoint url and a token for sending Caliper events too. It does require an IMS Global account. If you don't have one, it will be a bit harder to test since you won't be able to validate the events but you could set up a generic server to catch all incoming messages to see what is being sent (ex: [httpbin](https://httpbin.org/))

here's an example of the `conf/localOverrides.conf` settings required to get it running locally (assuming you are using the IMS Global caliper test server)
```
$caliper{enabled} = 1;
$caliper{host} = 'https://caliper.imsglobal.org/caliper/IMS_TEST_SERVER_TOKEN_VALUE/message';
$caliper{api_key} = 'IMS_TEST_SERVER_TOKEN_VALUE';
$caliper{errorlog} = $webworkDirs{logs} . "/caliper_errors.log";
```

In addition you can set a base urls used for the IRI's that identify the resources. This is optional but allows you to use persistent urls in case the domain name ever changes or something along those lines (so your Caliper event history is more consistent). 
```
$caliper{base_url} = 'https://webwork.elearning.ubc.ca/webwork2/';
```

Finally, there is another optional setting for the actor object for the Caliper event. The default Caliper actor will use Webwork account information to identify the user but it could be easier for processing later on to use some other information to uniquely identify the user across applicaitons (In UBC's case, we use a puid that is stored in the `user_id` and set the IRI to `http://www.ubc.ca/PUID`

```
$caliper{custom_actor_generator} = sub {
	my ($ce, $db, $user) = @_;
	my $caliper_id = 'http://www.ubc.ca/' . $user->user_id();
	return {
		'id' => $caliper_id,
		'type' => 'Person',
		'name' => $user->first_name() . " " . $user->last_name(),
	};
};
```

andrew-gardener:
Also just a note that a newer version of Caliper (v1.2) will be released sometime in the near future. I'll make another pull request with updates when it's fully released.

taniwallach:
> @taniwallach I've added both libnet-https-nb-perl & libhttp-async-perl instead of installing through cpanm (Net::HTTPS::NB is a requirement for the async library)

For anyone who runs into trouble using Docker and is not ready to rebuild their Docker image yet after updated webwork2 to `develop` and mounting it onto an older Docker image at present - you can add
```
      ADD_APT_PACKAGES: libnet-https-nb-perl libhttp-async-perl 
```
into the appropriate section of `docker-compose.yml` so the old image can install these packages and start up.

PR #973, opened 2019-07-17
Patch to support email to users with Unicode characters in their name. Also fix mail header to report the body being encoded using UTF-8.

taniwallach:
The first 2 commits were focused on fixes to messages generated by `lib/WeBWorK/ContentGenerator/Feedback.pm` and the third commit addresses code in 2 other places which generate email messages.

The changes are to allow emails to be sent when the `$user->full_name` is in Unicode and to support Unicode characters in the message text (which was already added to `lib/WeBWorK/ContentGenerator/Feedback.pm` but not to the other 2 email generating files.)

We also set the email header
```
	Content-Type: text/plain; charset=UTF-8
```
as the message body is already set to be encoded in UTF-8, and it helps some email clients to see that set properly in the headers.

Most of the chnages have been tested successfully on my servers which are running the WW 2.15 branch, but I did not test the changes to `jitar_send_warning_email` from `lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm`.

---

Before this change, on a site which has UTF-8 support in the database, and the student's name is in Unicode, "Feedback" email could not be sent. An error message:
```
    Failed to send message: Wide character in syswrite at /usr/local/share/perl/5.22.1/Net/Cmd.pm line 210.
```
would be given. However, mail to users with only 7-bit characters in the name would work fine, but some older mail clients (ex. Alpine) would show gibberish instead of the characters encoded in UTF-8 in the message body.

With this change: the email should be properly sent (if the system is able to send emails), and simple clients should understand that the message body is UTF-8 encoded.

---

Prior changes  https://github.com/mgage/webwork2/commit/f9c5c5f9e841418e49546acb0670f26e23fd085f#diff-d7212ed759de7d3a30aa6dadc29f25f7https://github.com/mgage/webwork2/commit/f9c5c5f9e841418e49546acb0670f26e23fd085f#diff-d7212ed759de7d3a30aa6dadc29f25f7 already allowed the use of UTF-8 text in the message body, but the correct `Content-Type` mail header was not being set.

---

References:
  - https://2rfc.net/2047
  - https://juerd.nl/site.plp/perluniadvice
  - https://ncona.com/2011/06/using-utf-8-characters-on-an-e-mail-subject/
  - https://stackoverflow.com/questions/7266935/how-to-send-utf-8-email
  - https://stackoverflow.com/questions/30734516/send-utf-8-encoded-mail-with-emailsender
 
About the future Unicode support for international email addresses where the address itself can be UTF-8 encoded:
 - https://en.wikipedia.org/wiki/International_email#UTF-8_headers

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
@@ -27,14 +27,14 @@ use strict;
 use warnings;
 #use CGI qw(-nosticky );
 use WeBWorK::CGI;
-use Email::Address;
+use Email::Address::XS;
 use HTML::Entities;
 #use Mail::Sender;
 use Email::Simple;
 use Email::Sender::Simple qw(sendmail);
 use Email::Sender::Transport::SMTP qw();
 use Try::Tiny;
-
+use Encode qw(encode_utf8 encode);
 use Data::Dump qw/dump/;
 use WeBWorK::Debug;
 
@@ -118,6 +118,7 @@ sub initialize {
 	my $ur = $self->{db}->getUser($user);
 
 	# store data
+	# rfc822_mailbox was modified to use RFC 2047 "MIME-Header" encoding.
 	$self->{defaultFrom}            =   $ur->rfc822_mailbox;
 	$self->{defaultReply}           =   $ur->rfc822_mailbox;
 	$self->{defaultSubject}         =   $self->r->urlpath->arg("courseID") . " notice";
@@ -388,11 +389,14 @@ sub initialize {
 		#################################################################
 		my $temp_body = ${ $r_text };
 		$temp_body =~ s/\r\n/\n/g;
-		$temp_body = join("",
-				  $r->maketext("From:")," $from \n",
-				  $r->maketext("Reply-To:")," $replyTo\n" ,
-				  $r->maketext("Subject:")," $subject\n" ,
-				  $r->maketext("Message:")," \n    $temp_body");
+		$temp_body = join("\n",
+				  "From: $from",
+				  "Reply-To: $replyTo",
+				  "Subject: $subject",
+				  "Content-Type: text/plain; charset=UTF-8",
+				  "Message:",
+				# Do NOT encode to UTF-8 here.
+				  $temp_body);
 		#warn "FIXME from $from | subject $subject |reply $replyTo|msg $temp_body";
 		#################################################################
 		# overwrite protection
@@ -425,16 +429,16 @@ sub initialize {
 		$self->{response}         = 'preview';
 
 	} elsif ($action eq 'sendEmail') {
-		# verify format of From address (one valid rfc2822 address)
-		my @parsed_from_addrs = Email::Address->parse($self->{from});
+		# verify format of From address (one valid rfc2822/rfc5322 address)
+		my @parsed_from_addrs = Email::Address::XS->parse($self->{from});
 		unless (@parsed_from_addrs == 1) {
 			$self->addbadmessage($r->maketext("From field must contain one valid email address."));
 			return;
 		}
 
-		# verify format of Reply-to address (zero or more valid rfc2822 addresses)
+		# verify format of Reply-to address (zero or more valid rfc2822/ref5322 addresses)
 		if (defined $self->{replyTo} and $self->{replyTo} ne "") {
-			my @parsed_replyto_addrs = Email::Address->parse($self->{replyTo});
+			my @parsed_replyto_addrs = Email::Address::XS->parse($self->{replyTo});
 			unless (@parsed_replyto_addrs > 0) {
 				$self->addbadmessage($r->maketext("Invalid Reply-to address."));
 				return;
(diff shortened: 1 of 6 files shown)


lib/WeBWorK/DB/Record/User.pm
```diff
@@ -87,13 +88,12 @@ sub rfc822_mailbox {
...
-			if ($full_name =~ /[\0-\037\177 ()<>@,;:\\".\[\]]/) {
-				$full_name =~ s/(["\\\r])/\\$1/g; # escape <">, "\", or CR
-				$full_name = "\"$full_name\"";
-			}
-			return "$full_name <$address>";
+			# Encode the user name using "MIME-Header" encoding,
+			# which allows UTF-8 encoded names. Given this change,
+			# the code to escape backslash, double quote, and \r
+			# was removed. Any such characters will be handled by
+			# the MIME-Header encoding.
```
heiderich: Thanks for the explanation. I would however suggest to put it in the commit message rather than as a comment in the code as it may confuse people, who have not seen the old code, in the future.

taniwallach:
> Does `"MIME-Header" encoding` refer to the encoding specified by RFC 2047? If so, maybe explicitly say so in the comments.
> 
> You are encoding the sender that way. What happens to the receivers?

Receivers are handled as they are added using the rfc822_mailbox() subroutine, but a comment about this was added in the second commit.

The first commit message was amended using a forced push, and contains some of what had been in the comment in `lib/WeBWorK/DB/Record/User.pm`. RFC 2017 is now also mentioned in several of the comments in the code (second commit).

A comment about the possible future need for RFC 2047 handling of the subject line (with a commented out line of code) was added in `lib/WeBWorK/ContentGenerator/Feedback.pm`.

A comment about the future support for Unicode email addresses: RFC 6531 and RFC 6532 was added to `lib/WeBWorK/DB/Record/User.pm`.

References:
  - https://tools.ietf.org/html/rfc2047
  - https://tools.ietf.org/html/rfc6531
  - https://tools.ietf.org/html/rfc6531

Apologies if anyone runs into problems due to the history rewrite inside my form and this PR, which should hopefully not effect anyone who did not clone my fork. https://help.github.com/en/articles/changing-a-commit-message has a link to a reference how to overcome such problems.

taniwallach:
A third commit is on the way. Mail is also being generated in:
  - `lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm`
  - `lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm`
and they also need adjustments for UTF-8 support. Those seem to be the only other two files using `Email::Simple` so should finish off these issues.

Also `lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm` was using the deprecated and insecure (https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-7686) `Email::Address` and I am replacing it with `Email::Address::XS` as recommended by https://metacpan.org/pod/release/RJBS/Email-Address-1.909/lib/Email/Address.pm

For the current Dockerfile based on Ubuntu 16.04, this is added via `cpanm`, but when the Dockerfile is updated to use Ubuntu 18.04 it can be pulled in from Ubuntu universe as libemail-address-xs-perl.

@heiderich - I will submit a separate PR to move Docker from Ubuntu 16.04 to 18.04, and that also gets most of the Perl modules from Ubuntu repositories, and cuts down on the use of `cpanm`.

---

Testing needed of `jitar_send_warning_email()` from `lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm`

lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
```diff
@@ -558,18 +562,29 @@ sub print_preview {
 	$msg = wrap("","",$msg);
 
 	$msg = join("",
-		    $errorMessage,
-		    $preview_header,
```
taniwallach: The two lines adding `$errorMessage` and `$preview_header` were moved further down, until after the HTML entities from the "sample message" are encoded.

taniwallach:
@mgage - If you have this installed on a machine:
  1. using `utf8` or `utf8mb` database tables, and
  2. which can send emails via the SMTP settings,
then you should be able to test this PR. Set a student name to have non-English characters in the name, and try to generate a feedback email as that use (name in the "From" field). After the PR merged code, it should work.

Testing the other parts of the code which create emails would also help, in particular the `jitar_send_warning_email()` if you know how to get it to be used. I did testing of the other uses already.

You remarked in the other thread https://github.com/openwebwork/webwork2/pull/972 about an issue installing `Email::Address::XS`. For Docker installs, the newer Docker config files in that PR (https://github.com/openwebwork/webwork2/pull/972) install `Email::Address::XS`  from an Ubuntu packages, which may help for some settings.

taniwallach:
@heiderich @nmoller - Can you please test? 

Note: A UTF8 database is needed for testing, so the student name can have non-Latin1 UTF-8 characters so that the issue/fix can be seen.

nmoller:
@taniwallach : I'm sorry... because you asked for a functional review but I'm going through the details. If you believe it does not help ... please let me know and I'll go straight to the functionality.

Duplication in the dockerfile:
- libxml-simple-perl
- libxmlrpc-lite-perl
installed twice.
Modification of the copyright at the same time change a lot of files (to simplify the diff of branches...i think it would be easier if done in a different PR. 

I did not finish my
`git diff master`
in tani's branch fix-email-utf8-in-username.
By the way.... I test the email in my projects inspired by https://github.com/moodlehq/moodle-docker .... using an SMTPserver like that. @taniwallach : do you think thats ok?

taniwallach:
> @taniwallach : I'm sorry... because you asked for a functional review but I'm going through the details. If you believe it does not help 
> 
> I did not finish my
> `git diff master`
> in tani's branch fix-email-utf8-in-username.

@nmoller - You want to compare my branch to the WeBWorK-2.15 branch.

https://github.com/openwebwork/webwork2/tree/WeBWorK-2.15

> ... please let me know and I'll go straight to the functionality.
> 
> Duplication in the dockerfile:
> 
> * libxml-simple-perl
> * libxmlrpc-lite-perl
>   installed twice.

I think I cleaned up some of that. 

>   Modification of the copyright at the same time change a lot of files (to simplify the diff of branches...i think it would be easier if done in a different PR.

That's part of WeBWorK-2.15.

taniwallach:
> By the way.... I test the email in my projects inspired by https://github.com/moodlehq/moodle-docker .... using an SMTPserver like that. @taniwallach : do you think thats ok?

It sound quite reasonable to use something like that. I've just tested message mainly to/from users/accounts I control via a real SMTP server. Some of the testing "spammed" someone else defined as a lecture in the test course (sorry @mgage).

taniwallach:
> Duplication in the dockerfile:
> 
> * libxml-simple-perl
> * libxmlrpc-lite-perl
>   installed twice.

Sorry - the fixes for that are in https://github.com/openwebwork/webwork2/pull/972 as that is the PR whch needed ubuntu 18.04... This PR should only be modifying 6 files relative to the WW 2.15 branch.

nmoller:
@taniwallach : In the file... you are going to see what I've tested so far.
[ww2-test001.pdf](https://github.com/openwebwork/webwork2/files/3453295/ww2-test001.pdf)

It seems that the testing of xelatex .... modifying the config file is not longer needed?

taniwallach:
> @taniwallach : In the file... you are going to see what I've tested so far.
> [ww2-test001.pdf](https://github.com/openwebwork/webwork2/files/3453295/ww2-test001.pdf)
> 
> It seems that the testing of xelatex .... modifying the config file is not longer needed?

The error message about "ZZZ-Inserted Text" not being in the taxonomy is triggered when using an admin level account with "re-tagging" privileges, when the problem has this placeholder in one of the main DB fields. It can be ignored.

From the last page of the PDF file you sent - I seem to understand that a mail sent in the name of the peter101 user whose name has Unicode characters was accepted for mailing. Before this PR such messages would not make it into the SMTP process. So I think it is a proof that it works properly, as it does on my systems. There was also a sample with UTF-8 encoded text in the message body, which had already worked based on prior commits by @mgage. 

The hardcopy support code is in https://github.com/openwebwork/webwork2/pull/972 and needs to be tested separately from the UTF-8 email support fixes in this PR. You can certainly try xelatex instead of pdflatex with the regular hardcopy theme - but on my machines, I think that ran into problems, as some of the packages in the regular hardcopy theme are not supported by xelatex, which is why there is an English only xelatex hardcopy theme in PR https://github.com/openwebwork/webwork2/pull/972. The changes to the config files are needed to enable using the extra  hardcopy themes.

mgage:
I plan to merge this request dealing with utf-8 since it seems largely ready and that will direct attention to the other important pull request #972 . There are a few duplications in the Dockerfile but this will be fixed once #973 is merged. 

Any comments or suggestions?

PR #976, opened 2019-07-28
Webwork 2.15+problem UUID

mgage:
This, and the companion pull request for PG, add the value of the variable `problemUUID` to the list of items used to create a unique UUID for a problem.  Its use case is for submissions via the webservice where the problem is not part of a homework set framework which help identify the problem and its auxiliary files uniquely.  Specifying  `problemUUID=>3245` (for example) in the form or iframe request for the problem will do this.  `problemUUID` replaces `problemIdentifierPref` as the name of the identifying variable being submitted although to insure backward compatibility the latter will still work.


lib/WebworkClient.pm
```diff
@@ -757,7 +759,8 @@ sub formatRenderedProblem {
 	my $userID           =  $self->{userID};
 	my $course_password  =  $self->{course_password};
 	my $problemSeed      =  $self->{inputs_ref}->{problemSeed}//4444;
-        my $psvn             =  $self->{inputs_ref}->{psvn}//54321;
+	my $problemUUID      =  $self->{inputs_ref}->{problemUUID}//0;
+    my $psvn             =  $self->{inputs_ref}->{psvn}//54321;
```
heiderich: Is the change of indentation intended?

clients/sendXMLRPC.pl
```diff
@@ -636,9 +636,9 @@ BEGIN
 		displayMode	=> $DISPLAYMODE,
 		outputformat 	=> $format//'standard',
 		problemSeed     => $problemSeed//PROBLEMSEED(),
-		psvn		=> $psvn//'23456',
+		psvn		    => $psvn//'23456',
```
heiderich: Is this change intentional?

clients/sendXMLRPC.pl
```diff
@@ -636,9 +636,9 @@ BEGIN
...
 		forcePortNumber => $credentials{forcePortNumber}//'',
-		language	=> $lang//'en',
+		language	    => $lang//'en',
```
heiderich: Is this change intentional?

clients/sendXMLRPC.pl
```diff
@@ -895,16 +895,13 @@ sub process_problem {
 	my $problemSeed = $form_data->{problemSeed};
 	die "problem seed not defined in sendXMLRPC::process_problem" unless $problemSeed;
 
-	my $local_psvn = '34567';
-	if ( defined( $form_data->{psvn} ) ) {
-	  $local_psvn = $form_data->{psvn};
-	}
-
+	
+    $local_psvn = $form_data_>{psvn}//34567;
```
heiderich: What is the role of 34567?
Is `$form_data_>` intended or should it be `$form_data->`?

clients/sendXMLRPC.pl
```diff
@@ -895,16 +895,13 @@ sub process_problem {
...
 	my $updated_input = {%$input, 
 					  envir => $xmlrpc_client->environment(
 							   fileName       => $adj_file_path,
 							   sourceFilePath => $adj_file_path,
-							   psvn		  => $local_psvn,
+							   psvn		      => $local_psvn,
```
heiderich: Is this change intentional?

clients/t/test_two_images_localhost.html
```diff
@@ -60,14 +60,14 @@ <h2>Problem 2</h2>
...
-src="http://localhost/webwork2/html2xml?
+src="http://127.0.0.1/webwork2/html2xml?
 	&answersSubmitted=0
 	&sourceFilePath=setDemo/limits.pg
 	&problemSeed=225
 	&displayMode=MathJax
 	&courseID=daemon_course
 	&userID=daemon
-	&problemIdentifierPrefix=321
+	&problemIdentifierPrefix=231
```
heiderich: What is the meaning of these special values? Why is it changed?

lib/FormatRenderedProblem.pm
```diff
@@ -143,6 +143,7 @@ sub formatRenderedProblem {
 	my $userID           =  $self->{userID}//'';
 	my $course_password  =  $self->{course_password}//'';
 	my $problemSeed      =  $self->{inputs_ref}->{problemSeed}//6666;
+	my $problemUUID      =  $self->{inputs_ref}->{problemUUID}//0;
```
heiderich: What is the meaning of 0?

lib/WeBWorK/ContentGenerator/instructorXMLHandler.pm
```diff
@@ -192,6 +192,7 @@ sub pre_header_initialize {
 		    subcommand		        => $r->param("subcommand") ||undef,
 		    maxdepth		        => $r->param("maxdepth") || 0,
 		    problemSeed	            => $r->param("problemSeed") || 0,
+		    problemUUID             => $r->param("problemUUID") // 0,
```
heiderich: What is the meaning of 0?

lib/WeBWorK/DB/Record/UserSet.pm
```diff
@@ -40,7 +40,7 @@ BEGIN {
 		visible                   => { type=>"INT" },
 		enable_reduced_scoring    => { type=>"INT" },
 		assignment_type           => { type=>"TEXT" },
-	        description               => { type=>"TEXT" },
+	    description               => { type=>"TEXT" },
```
heiderich: Is this change intentional?

lib/WeBWorK/PG.pm
```diff
@@ -72,7 +72,7 @@ sub defineProblemEnvir {
 		$key,
 		$set,
 		$problem,
-		$psvn,  #FIXME  -- not used
+		$psvn,  #FIXME  -- not used  ??
```
heiderich: Is it used or not?

mgage:
Thanks Florian.  I've added some test files:

test_embedded_problems.html
test_formats_WWproblems.html
test_simpleFormat_WWproblem.html
test_two_embedded_images_localhost.html

These have embedded problems that are pointed at demo.webwork.rochester.edu  which is 
running with this PR added.  They should work well just loading them into a browser.

To test whether the pull request works on your machine you will
have to replace the domain in the urls in each of the iframes with your own domain.

For the course/user_name/user_password you can change those as well, or you can create a course named daemon_course with a user daemon (with ta privileges) and a password daemon.

lib/WebworkClient.pm
mgage: maybe not -- I was evening up the line up in my editor but apparently I had extra tabs. I'll go make sure the tabs are gone.

clients/sendXMLRPC.pl
mgage: maybe not -- I was evening up the line up but apparently I had extra tabs.

clients/sendXMLRPC.pl
mgage: ditto

clients/sendXMLRPC.pl
mgage: form_data->  

 thanks.

lib/FormatRenderedProblem.pm
mgage: the problemUUID has no meaning except as an identifier.  0 is a convenient value and
an empty string would probably be ok, but since the problemUUID is expected to be a number, by convention, 0 seemed safer.

lib/WeBWorK/PG.pm
mgage: It turns out the psvn, problem set version number, which is a legacy from 20 years ago and was being phased out, BUT, it has some uses after all. It is essentially the unique id for the user_set row specifying the specific problem for a specific user. 
    https://github.com/openwebwork/webwork2/blob/master/lib/WeBWorK/DB/Record/UserSet.pm

in the next release (2.16) I'd like to add uniq_id and timestamp fields to _every_ database table.

mgage:
I think I caught the cosmetic problems (mostly tabs vs spaces) that were causing the immediate
problem. 

I added a connectivity check that will also allow us to keep better track of which sites are using webwork.  Let me know if that works on your machine. It uses a system call to `ping` so there is a chance that it won't always work. It pings the MAA server once when the perl module starts up.

clients/t/test_two_images_localhost.html
mgage: This differentiates the two problems so that associated auxiliary files (or geogebra applets) have distinct urls.  This allows you to include a problem twice, with the same seed, and create two geogebra apps that operate independently.  problemUUID should be used instead of problemIdentifierPrefix but the latter still works.

taniwallach:
PSVN can be used to seed the RNG for a set of questions which need the same set of random parameters, as by default each question gets its own seed. See http://webwork.maa.org/wiki/LinkingProblems#.XT3wzB5RU0N

Thus, comments in the code which suggest dropping it should be revised to remark that it remains in use.

conf/webwork.apache2-config.dist
```diff
@@ -123,6 +123,10 @@ eval "use lib '$pg_dir/lib'"; die $@ if $@;
 require Apache::WeBWorK; # force compilation of pretty much everything
 
 
+print "Check in with MAA site\n";
+system("ping -c1 -s100 webwork.maa.org "); # check MAA site
```
taniwallach: I think there should be a comment here and in the installation instructions that this is here both to test connectivity and to collect some data on hosts running webwork.

conf/webwork.apache2.4-config.dist
```diff
@@ -123,6 +123,9 @@ if (exists(($ce->{mail}->{smtpPort})) )  { # port not chosen automatically
 
 require Apache::WeBWorK; # force compilation of pretty much everything
 
+print "Check in with MAA site\n";
+system("ping -c1 -s100 webwork.maa.org "); # check MAA site
```
taniwallach: I think there should be a comment here and in the installation instructions that this is here both to test connectivity and to collect some data on hosts running webwork.

conf/webwork.apache2-config.dist
heiderich: It is not clear to me why the connectivity test is necessary. WeBWorK used to be executable without internet connection (at least some parts). Why do we need this test now?

Concerning the collection of usage information, I think that there should at least be a comment in the installation instructions about this. But I doubt whether this is a reliable approach to collect such information. Ping packets can originate anywhere. There is no guarantee that they are sent by WeBWorK servers. How about adding a possibility for a user to opt in and then possibly send more useful information about the system, such as the versions of webwork2 and pg used.

Possibly such an opt in could be combined with the collection of usage information of the OPL. There is a script for this, but I remember having read that the server collecting this data is not available anymore. In principal these two are different things, but they have in common that data is sent back to a central server.

mgage: The connectivity test isn't necessary -- nor are the rest of the printouts that occur in this file -- but it can speed up debugging.  However the main use for this is to find out which webwork sites are active.  We are constantly asked how many sites are set up (and how many students are using webwork) and we essentially have no idea.  We've long asked people to register (this creates the list at http://webwork.maa.org/wiki/WeBWorK_Sites ) but (1) many sites don't register and (2) many of the sites registered are no longer in use (or don't exist at all).  

I'm sensitive to privacy aspects of collecting but pings seem pretty innocuous and seem the minimal way to collect data about active sites. 
As I said above the site administrator has already had the opportunity to opt in (for the last 10 years) but that really hasn't been sufficient to answer questions about number of sites using webwork.  Perhaps we could enforce registration in some way. 

Notice that I have -s100 set in the call to ping.  This means that every ping comes in with a packet length of 108.  I have tcpdump() set up to only record ping packets of that length. 

I'm fine with putting a longer explanation of what the ping does and what it is used for in the code and in the documentation.  I'm open to suggestions for wording.  We can also leave this line out altogether but that means we can't answer the frequent question about number of WeBWorK users.

conf/webwork.apache2.4-config.dist
mgage: see my reply to @heiderich above.  If we add these comments to you think this is too intrusive?

conf/webwork.apache2-config.dist
taniwallach: I fell much better with the extra comments in place.
I still think that the install documentation (both regular and Docker) should mention this, and explain how to turn it off.

Could the system call to run ping be set to run in the background so that it does not delay the startup time?

About the use of tcpdump, I think you can probably add a size based filter to the `tcpdump` call and save the need to use grep.

lib/WeBWorK/PG.pm
heiderich: I removed this comment (along with some others) in https://github.com/mgage/webwork2/pull/29, because I consider it misleading.

clients/sendXMLRPC.pl
mgage: The psvn (Problem Set Version Number) has a historical origin.  Currently it is 
defined as the uniq_id for each row of the user-problemSet table. So it identifies a 
specific homework set assigned to a specific user.  As Tani describes below the psvn=problemUUID  has been used to provide a set-wide seed for the random number generators of problems so that the values of the problems can be coordinated. It's not super resilient, but it has worked for those special cases that require that kind of coordination.

lib/WeBWorK/PG.pm
taniwallach: Please drop the "FIXE -- not user" as `$psvn` does have uses.

clients/t/test_formats_WWproblems.html
```diff
@@ -0,0 +1,106 @@
...
+src="https://demo.webwork.rochester.edu/webwork2/html2xml?  
+    &answersSubmitted=0&  
+    &sourceFilePath=Library/ma122DB/set12/s5_4_14.pg&  
+    &problemSeed=123567&  
+    &courseID=daemon_course&  
+    &userID=daemon&  
+    &course_password=daemon&  
+    &showSummary=1&  
+    &displayMode=MathJax&  
+    &problemIdentifierPrefix=102&
```
taniwallach: Maybe change the `problemIdentifierPrefix` to ` problemUUID` in the multiple occurrences in this file.

clients/t/test_toggled_embedded_WWproblems.html
```diff
@@ -166,8 +179,24 @@ <h1>Embedded WeBWorK Problems</h1>
...
+  Click here to toggle problem display: (PGinfo example)
+</div>
+<div class="wwproblem" style="display: none;">
+<hr>
+<iframe  width="800" height="600" 
+src="http://demo.webwork.rochester.edu/webwork2/html2xml?
+	&answersSubmitted=0&
+	&sourceFilePath=settesting/pgInfo.pg&
+	&problemSeed=123567890&
+	&problemIdentifierPrefix=5555&
```
taniwallach: Maybe change the `problemIdentifierPrefix` to ` problemUUID` in the occurrences in this file.

lib/WeBWorK/PG.pm
mgage: Done

clients/t/test_formats_WWproblems.html
mgage: ok

mgage:
Tani -- take a quick look at this and make sure that I haven't overwritten any of your more recent changes.  I don't think that I did.

mgage:
I'm about to merge this PR involving problem UUID into WeBWorK-2.15 (and the corresponding PR into PG-2.15) unless there are additional comments. -- Mike

lib/WebworkClient.pm
```diff
@@ -757,19 +759,20 @@ sub formatRenderedProblem {
...
 	my $checkMode        =  defined($self->{inputs_ref}->{WWcheck});
 	my $submitMode       =  defined($self->{inputs_ref}->{WWsubmit});
 	my $showCorrectMode  =  defined($self->{inputs_ref}->{WWcorrectAns});
-        # problemIdentifierPrefix can be added to the request as a parameter.  
+        # problemUUID can be added to the request as a parameter.  
         # It adds a prefix to the 
         # identifier used by the  format so that several different problems
         # can appear on the same page.   
-	my $problemIdentifierPrefix = $self->{inputs_ref}->{problemIdentifierPrefix} //'';
+	my $problemUUID = $self->{inputs_ref}->{problemUUID} //'';
```
taniwallach: Second my declaration, different fallback value (empty string) which differs from the standard fallback used elsewhere.
Since the comment explains the purpose - the first declaration should be removed, and the fallback of the second one fixed.
Indentation of following 3 lines is non-standard. Now is a chance to fix it.

clients/t/test_two_embedded_images_localhost.html
```diff
@@ -39,35 +39,42 @@ <h2>Problem 1</h2>
...
 <h2>Problem 2</h2>
 <p>
-This is for debugging and shows all the details of the hash returned by PG.
+By loading the graphs into separate tabs you can see that the two
+graphs have different urls. This allows separation of 
+graphs, auxiliary html, and geogebra applets which might have the same 
+name and are loaded from problems with the same name but are supposed to be
+two different exercises. (Usually the fact that the problems have different
+seeds is enough to separate them but just in case you can set 
+different problemUUID/problemUUID values. problemUUID is the preferred
```
taniwallach: Text is mangled. Probably intended to refer to `problemIdentifierPrefix` the prior name in beta code for this value.

lib/FormatRenderedProblem.pm
```diff
@@ -143,18 +143,19 @@ sub formatRenderedProblem {
...
 	my $checkMode        =  defined($self->{inputs_ref}->{WWcheck})||0;
 	my $submitMode       =  defined($self->{inputs_ref}->{WWsubmit})||0;
 	my $showCorrectMode  =  defined($self->{inputs_ref}->{WWcorrectAns})||0;
-        # problemIdentifierPrefix can be added to the request as a parameter.  
+        # problemUUID can be added to the request as a parameter.  
         # It adds a prefix to the 
         # identifier used by the  format so that several different problems
         # can appear on the same page.   
-	my $problemIdentifierPrefix = $self->{inputs_ref}->{problemIdentifierPrefix} //'';
+	my $problemUUID = $self->{inputs_ref}->{problemUUID} //'';
```
taniwallach: Second my declaration, different fallback value (empty string) which differs from the standard fallback used elsewhere.
Since the comment explains the purpose - the first declaration should be removed, and the fallback of the second one fixed.
Indentation of following 3 lines is non-standard. Now is a chance to fix it.

lib/WeBWorK/PG.pm
```diff
@@ -92,10 +92,13 @@ sub defineProblemEnvir {
 	# ADDED: displayModeFailover, displayHintsQ, displaySolutionsQ,
 	#        refreshMath2img, texDisposition
 
-	# pstaab: changed the next line from
-	#$envir{psvn}                = $set->psvn;
-	# to
-	$envir{psvn}                = $psvn;
+	$envir{psvn}                = $psvn;  #'problem set version number' (associated with homework set)
+	$envir{psvn}                = $envir{psvn}//$set->psvn; # use set value of psvn unless there is an explicit override. 
+	# update problemUUID from submitted form
```
taniwallach: Clarify fallback to old name `problemIdentifierPrefix`

taniwallach:
I tested a modified version of the included `test_two_embedded_images_localhost.html` which pointed to a local Docker server with a running daemon_course which was running the code from this PR and the companion PR for pg.

The two problems loaded by that file are both instances of the same PG file with the same seed and the generated images get different UUID identifiers in the path as a result of the different values of `problemUUID`. That proves that the changes in this PR does what it was intended to do.

---

When testing `test_toggled_embedded_WWproblems.html` I noticed:
    1. That when problems 1,2,3 are first opened using the "Click here to toggle problem display" links the formatting, and in particular the MathJax display was manged. As soon as the frame was reloaded using "Reload frame" from the Chrome context menu (or the similar feature in FireFox) - the layout would be fixed.
    2. That the problems 4,6,7 in the example **all** fail to load, both on my local server and from `https://demo.webwork.rochester.edu` as the problems files are apparently not available where expected even on demo.webwork.rochester.edu.

---

When testing `test_simpleFormat_WWproblem.html` the second problem ("Direct access using RESTful url. Sign in is necessary and score is recorded.") can **only** be tested against the `demo.webwork.rochester.edu` server, where the relevant course/problem is available, but the "Guest login" fails there.

---

When testing  `test_formats_WWproblems.html` the last sample "Format: JSON" fails on `demo.webwork.rochester.edu`  but works on my local system. It seems that `demo.webwork.rochester.edu` is not running a sufficiently recent version of WW to support that new output mode (added for WW 2.15). The error message given includes 
```
    Undefined subroutine &FormatRenderedProblem::to_json called at /opt/webwork/webwork2/lib/FormatRenderedProblem.pm line 480. 
```

---

`test_embedded_problems.html` seems to work fully.

---

Notes on local changes to the test files for testing against a local Docker server (on port 80):

  - OLD: `https://demo.webwork.rochester.edu/webwork2/html2xml?`  
  - NEW: `http://localhost/webwork2/html2xml?`

  - OLD: `http://127.0.0.1/webwork2/html2xml?`
  - NEW: `http://localhost/webwork2/html2xml?`

  - OLD: `https://cdn.mathjax.org/mathjax/latest/MathJax.js` 
  - NEW: `http://localhost/webwork2_files/mathjax/MathJax.js?`

  - OLD: `https://demo.webwork.rochester.edu/webwork2_files`
  - NEW: `http://localhost/webwork2_files`

taniwallach:
@mgage - Do you want to fix the minor issues in the prior comment with the test*.html files?
If not (and it is not really necessary) - then I think this can be merged.
I already merged to code change on the pg side.

mgage:
Go ahead and merge I'll fix things later either later or in devel


sent from my mobile device

On Sep 18, 2019 5:46 PM, Nathan Wallach <notifications@github.com> wrote:

@mgage<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_mgage&d=DwMCaQ&c=kbmfwr1Yojg42sGEpaQh5ofMHBeTl9EI2eaqQZhHbOU&r=C6Pt5AGtImanmAdcooarL-JZO8M5dSFPfs3VweYXYkE&m=CkOGXTHyMffbe9BEnKGa8Wlkq4uhbaCfwZfAwnk9_Eg&s=x912DpqPl_aVBKsQVX48kkpZRAvLq6i40KHHnAne5cA&e=> - Do you want to fix the minor issues in the prior comment with the test*.html files?
If not (and it is not really necessary) - then I think this can be merged.
I already merged to code change on the pg side.

PR #983, opened 2019-08-09
removed use mod_perl lines from code

pstaabp:
This removes all of the `use mod_perl` lines from the webwork code.  Since everyone should be using apache2, therefore mod_perl2 instead and mod_perl2 is loaded from the `webwork2/conf/webwork.apache2.4-config` configuration file.  This is mainly to remove some old code that shouldn't be necessary. 

I was getting errors along the lines of `mod_perl.pm not found`, which shouldn't be necessary. I stopgapped this by making a link to `mod_perl2.pm`, not a good fix. 

I'm not sure how to best test this.  I removed the link from `mod_perl.pm` to `mod_perl2.pm`. There are no errors on startup and visiting all links.

### DIFF
--- a/bin/dump_past_answers
+++ b/bin/dump_past_answers
@@ -17,7 +17,7 @@
 ##############################################################################
 
 # This script dumps the course information from all unarchived courses into
-# a single csv file.  The csv file has the following columns. 
+# a single csv file.  The csv file has the following columns.
 #
 # ID Info
 # 0 - Answer ID hash
@@ -44,7 +44,7 @@
 # 18 - Final Incorrect Attempts
 # 19 - Final Correct Attempts
 # 20 - Final Status
-# OPL Info 
+# OPL Info
 # 21 - Subject
 # 22 - Chapter
 # 23 - Section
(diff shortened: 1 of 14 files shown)


taniwallach:
@pstaabp would you mind trying to make this a PR for the WW 2.15 branch instead.

If it does not break anything (we will need to test it) - then it can be merged in for WW 2.15.

I think I saw a copyright symbol in one of the place with a change. We are being careful to avoid the latin-1 copyright symbol which makes trouble when a file is read in as UTF-8. @mgage changed many to &copy; for now. **Please check.**

Maybe - split into 2 commits, one which only removes the `use mod_perl` lines, and the second with all the white-space cleanup. Then we can look at the "real" changes and the cosmetic ones separately by focusing on one commit at a time.

If that is hard for your to do, then please post a "diff" without white-space changes. [One site ](https://coderwall.com/p/crj69a/from-a-useless-git-diff-to-a-useful-one) suggests (for Git version >= 1.8.4):
```
git diff --ignore-space-at-eol -b -w --ignore-blank-lines [commit] ...
```
which will help us review the real changes.

pstaabp:
@taniwallach: I can definitely do this.  I know the white-space issue is a problem.  We should probably have some kind of policy about white space in the code.  I'm currently doing a lot of javascript and typescript right now and using a linter--it seems that [Perl::Critic](https://metacpan.org/pod/Perl::Critic) may be something of use for this. 

I'll skip messing with any white space right now. 

About the copyright symbol: I see the bad symbol in a lot of the code.  For example, `/lib/WeBWork/ContentGenerator.pm` on line 3.  Do you want me to replace all of those?

mgage:
you can fix those copyright signs if you wish.  I didn't bother because the latin-1 copyright 
only causes problems if something is read as text (as for example is the case with course.conf).

For future reference, in addition to Perl::Critic, we should look at Code::TidyAll.  Both of them were prominent in the talks I went  to at the perl conference in June.

pstaabp:
I just opened a PR for the webwork-2.15 branch.  Are the plans to pull all of webwork-2.15 into develop or should I keep this as is?  

If we are pulling in webwork-2.15 into develop after finalizing everything, then we can delete this.

taniwallach:
Since https://github.com/openwebwork/webwork2/pull/984 was merged into WeBWorK-2.15, and that will later be merged into both develop and master, this can be closed.

mgage:
we are planning to pull all of WeBWorK-2.15 into develop.

On Aug 12, 2019, at 2:39 PM, Peter Staab <notifications@github.com<mailto:notifications@github.com>> wrote:


I just opened a PR for the webwork-2.15 branch. Are the plans to pull all of webwork-2.15 into develop or should I keep this as is?

If we are pulling in webwork-2.15 into develop after finalizing everything, then we can delete this.

taniwallach:
We should pull the full WW 2.15 branch into develop, probably when it merges into master.
Thus this PR should not be needed any longer.

PR #984, opened 2019-08-12
Remove mod perl ww2.15

pstaabp:
This removes all of the use mod_perl lines from the webwork code and replaces #983.  Since everyone should be using apache2, therefore mod_perl2 instead and mod_perl2 is loaded from the webwork2/conf/webwork.apache2.4-config configuration file. This is mainly to remove some old code that shouldn't be necessary.

I was getting errors along the lines of mod_perl.pm not found, which shouldn't be necessary. I stopgapped this by making a link to mod_perl2.pm, not a good fix.

I'm not sure how to best test this. I removed the link from mod_perl.pm to mod_perl2.pm. There are no errors on startup and visiting all links.

### DIFF
--- a/lib/Apache/WeBWorK.pm
+++ b/lib/Apache/WeBWorK.pm
@@ -38,7 +38,7 @@ use WeBWorK;
 use Encode;
 use utf8;
 
-use mod_perl;
+
 use constant MP2 => ( exists $ENV{MOD_PERL_API_VERSION} and $ENV{MOD_PERL_API_VERSION} >= 2 );
 
 # load correct modules
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -61,7 +61,7 @@ use URI::Escape;
 use Carp;
 use Scalar::Util qw(weaken);
 
-use mod_perl;
+
 use constant MP2 => ( exists $ENV{MOD_PERL_API_VERSION} and $ENV{MOD_PERL_API_VERSION} >= 2 );
 
 #####################
--- a/lib/WeBWorK/Authen/CAS.pm
+++ b/lib/WeBWorK/Authen/CAS.pm
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright © 2012 The WeBWorK Project, http://openwebwork.sf.net/
+# Copyright &copy; 2012 The WeBWorK Project, http://openwebwork.sf.net/
 # $CVSHeader: $
 # 
 # This program is free software; you can redistribute it and/or modify it under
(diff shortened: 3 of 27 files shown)


taniwallach:
@heiderich @nmoller @mgage - I recommend we merge this into WW 2.15.I tested this patch via Docker and everything I tried works just fine without the `use mod_perl` lines removed by the PR.

However, there is a catch. Some very old installations using old Linux distributions may still be using the old `mod_perl` (version 1) with Apache 1.x. It seems that someone would need to be running an ancient version of Linux to do this. (See below.) Given that, I think that anyone who still wants to run WeBWorK on such an old system, can be expected to handle some minor inconveniences. If the issue comes up, we could somewhat help such users by providing a tgz file with the old versions of the files that had a `use mod_perl` line, and users who need it can manually patch the files based on those fixed samples, but I don't think we need to continue to maintain support for such sites in the future.

---

Based on what I checked, RHEL 5 and RHEL 6 used `mod_perl` (version 2.0.2/2.0.4)
  - RHEL 5 had a Red Hat package for 2.0.2 and 2.0.4.
  - RHEL 6 had a Red Hat package for 2.0.4.
  - https://access.redhat.com/downloads/content/mod_perl/2.0.4-12.el6_10/x86_64/fd431d51/package-changelog    (behind Red Hat paywall...)
  - RHEL 4 ended "End of Extended Life-cycle Support" on March 31, 2017.
  - RHEL 5 is still in "End of Extended Life-cycle Support" until November 30, 2020	
  - If someone is still running something based on RHEL 4 - they are way behind the times and no longer have any support.
  - https://access.redhat.com/support/policy/updates/errata
  - https://en.m.wikipedia.org/wiki/Red_Hat_Enterprise_Linux

Debian packaged `libapache2-mod-perl2`since 2007 (and 1.99 since 2005) and Apache 2 and `mod_perl2` seem to have been standard by Debian 5 = Lenny which was released in 2009. However, Debian 4 Etch (released in 2007) still used Apache 1.3 and the old version 1 of mod_perl, but that version of Debian had "End of Life" in early 2010:
  - http://archive.debian.org/debian/pool/main/liba/libapache2-mod-perl2/
  - https://www.linode.com/docs/web-servers/apache/apache-2-web-server-on-debian-5-lenny/
  - https://wiki.debian.org/DebianReleases
  - http://alexpeters.net/tag/mod_perl/
  - https://www.debian.org/News/2010/20100121

---

Note: RHEL 7 and derivatives need to use an external source for `mod_perl` such as EPEL (as Red Hat is not packaging `mod_perl` for RHEL 7):
  - http://rpm.pbone.net/index.php3/stat/4/idpl/49981069/dir/redhat_el_7/com/mod_perl-2.0.10-3.el7.x86_64.rpm.html
  - https://access.redhat.com/solutions/3027371 (if you have a RedHat account). 
  > Starting with Red Hat Enterprise Linux 7, mod_perl is no longer available. To use perl, mod_fcgid should be used and is installed with the httpd package.
> If mod_perl is required, it's available in the EPEL repository, but this will be unsupported.
  - https://centos.pkgs.org/7/epel-x86_64/mod_perl-2.0.10-3.el7.x86_64.rpm.html

Issue #981, opened 2019-08-06
Small feature request: auto-check/uncheck checkboxes on edit of field


taniwallach:
When editing the dates for an individual student - if the checkbox is not marked, then the date does not get saved. Tabbing to a different field automatically checks the checkbox, but editing the field does not.

Teaching staff has complained that sometimes they forget to mark the checkbox to to tab to get it marked, and then fail to pay attention that their change did not get saved.

Could we make these type of checkboxes mark themselves as soon as the relevant field has a key-press, unless the field became empty?

mgage:
I think that could be done at this point.  There are javascript commands for updating a field on “key up” which might be
appropriate.  At the time this was written I believe that updating on “leaveFocus” was the best you could do — hence
the current behavior.

Take care,
Mike


On Aug 6, 2019, at 4:16 AM, Nathan Wallach <notifications@github.com<mailto:notifications@github.com>> wrote:


When editing the dates for an individual student - if the checkbox is not marked, then the date does not get saved. Tabbing to a different field automatically checks the checkbox, but editing the field does not.

Teaching staff has complained that sometimes they forget to mark the checkbox to to tab to get it marked, and then fail to pay attention that their change did not get saved.

Could we make these type of checkboxes mark themselves as soon as the relevant field has a key-press, unless the field became empty?

taniwallach:
> I think that could be done at this point. There are javascript commands for updating a field on “key up” which might be appropriate. At the time this was written I believe that updating on “leaveFocus” was the best you could do — hence the current behavior. 

I don't have time to work in this one now, but it annoyed someone enough to prompt me to file the issue.

If you have a volunteer to do it - we can get it in for WW 2.15 :smiley:  If not, it can wait.

mgage:
I’m pretty sure this one will have to wait.  It’s good to have it filed as an issue however.

Perhaps someone will be inspired to start thinking about this and other javaScript/AJAX
tweaks for release 2.16 :-)

Mike

On Aug 6, 2019, at 1:17 PM, Nathan Wallach <notifications@github.com<mailto:notifications@github.com>> wrote:


I think that could be done at this point. There are javascript commands for updating a field on “key up” which might be appropriate. At the time this was written I believe that updating on “leaveFocus” was the best you could do — hence the current behavior.

I don't have time to work in this one now, but it annoyed someone enough to prompt me to file the issue.

If you have a volunteer to do it - we can get it in for WW 2.15 😃 If not, it can wait.

[...]

PR #988, opened 2019-08-19
Auto checkboxes 981

drdrew42:
# Overview

This PR causes the override checkboxes to automatically be checked when changes occur in their corresponding date/time fields ( #981 ). Additionally, the default text (usually "None Specified") is automatically removed when the field receives focus. And finally, if the date/time is removed from a field, the override box is automatically de-checked and the default text is returned to the field.

Changes have been implemented for the Set Detail 2 page, where dates can be customized for a specific user on a specific set -- as well as on the User Detail page, where dates can be customized for a specific user across ALL assigned sets.

# Testing

Using a non-empty course with problem sets assigned to at least one user: 

## What pages to test:
1. From the Homework Sets Editor page, edit assigned users for some assigned set, then edit data for some user -- this will take you to the Set Detail 2 page: interact with the open/close/answer dates.
1. From the Classlist Editor page, edit assigned sets for some user -- this will take you to the User Detail page: interact with the open/close/answer dates for multiple sets.

## Testing Interaction with the dates:
1. "None Specified" disappears when you click into a field
1. "None Specified" returns to the field when you click out of that field, leaving it empty.
1. The checkbox is not checked when you leave a field empty.
1. The checkbox becomes checked as soon as you type anything into a blank field.
1. The checkbox becomes checked as soon as you select a date from the 'Datepicker' (calendar)
1. Deleting the contents of a field and then leaving the field causes "None Specified" to reappear in the field AND the checkbox becomes un-checked.


lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail2.pm
```diff
@@ -718,6 +724,9 @@ sub FieldHTML {
 				value => $r->param("$recordType.$recordID.$field") || ($forUsers ? $userValue : $globalValue),
 				size => $properties{size} || 5,
 				onChange => $onChange,
+				onkeyup => $onKeyUp,
+				onfocus => "if (this.value=='None Specified') { this.value = ''; }",
+				onblur => "if (this.value == '') { this.value = 'None Specified'; $uncheckBox; $resetClass; }",
 		}));
```
drgrice1: There are several things here that should be addressed.

First, I realize that you copied the selector that was used for the onChange handler, but that selector has problems and should be fixed.  The dots that are explicitly added here are escaped, but what about the possibility of dots in the variable $recordID (or other characters in that variable that might need to be escaped in the selector).  I discovered this issue as I use homework names that do have dots in them, and this selector fails.  It would be better to use a selector that does not need escaping.  A selector that does so is
``$onChange = "\$('input[id=\"$recordType.$recordID.$field.override_id\"]').prop('checked',true)"``
The selectors in UserDetail.pm need to also be fixed in this way.

Also note that technicially this should be using prop and not attr for the 'checked' propery.  See https://api.jquery.com/prop/.  Although attr works here since it effectively also sets the property.  It would be more serious if we were reading the value.  The attribute is really the initial setting, and the property is the current state.

Next, I see is that you are essentially hacking in with javascript what a "placeholder" attribute does naturally.  That is the "None Specified" behaviour you have implemented.  If you remove the onfocus handler and the part in the onblur handler setting the value, delete "labels" on lines 119 and beyond, and add  `placeholder => x("None Specified")` here you get nicer behavior in my opinion.  This also should be changed in UserDetail.pm.   There line 393 needs to be modified to compare to the empty string, rather than "None Specified".  This also allows for the string to be translated as it is not then used in that comparison on line 393.

drgrice1:
It seems there are some problems with my pull request to this one that need to be worked out.  Namely the placeholder should only be set for date fields.

mgage:
I'd like to suggest that we point this at develop.  It's going to make a slightly tricky change 
to a feature people are currently using.  I'm also hoping to make a 2.16 release around January -- it would probably not be a big change -- more like a collection of small changes of this type.

Does that seem reasonable?

drgrice1:
I agree that this definitely should point to develop.  The milestone is set for 2.16, but it is still set to merge into the WeBWorK-2.15 branch.

mgage:
Some changes need to be made before this can be tested.

mgage:
@drdrew42   Andrew, could you merge the pull request that Glenn made to github.com/drdrew42/webwork2  -- then I'll test this and merge it into develop.

drdrew42:
Sorry for this delay - I've been having issues with my VM, and I've had other projects preventing me from spending the necessary time to fix/reinstall my VM. And without it, I've been unable to test any changes or do any further development.

mgage:
no problem.  I finally have some free time to do pull requests and I didn’t want to overwrite or redo your work.  Thanks for pushing your initial changes.
Hope you had a good Christmas and have a happy new year. — Mike


On Dec 28, 2019, at 11:01 AM, drdrew42 <notifications@github.com<mailto:notifications@github.com>> wrote:


Sorry for this delay - I've been having issues with my VM, and I've had other projects preventing me from spending the necessary time to fix/reinstall my VM. And without it, I've been unable to test any changes or do any further development.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail2.pm
mgage: the changes suggested by @drgrice1 have been merged

mgage:
1.    "None Specified" disappears when you click into a field -- once you type one character
2.  "None Specified" returns to the field when you click out of that field, leaving it empty. -- non-empty appears when the box is empty, even while still in focus.
3.    The checkbox is not checked when you leave a field empty. -- unchecks when you remove focus.
4.     The checkbox becomes checked as soon as you type anything into a blank field.
5.  The checkbox becomes checked as soon as you select a date from the 'Datepicker' (calendar) -- works
6.    Deleting the contents of a field and then leaving the field causes "None Specified" to reappear in the field AND the checkbox becomes un-checked. -- this works.  

!!!!!
However you can delete the field (which does not turn off the check box until you defocus) then hit return (instead of tab) and the page will refresh and the older default date will appear. If you make a change and remove focus by some means other than return then things work as desired. 

This is getting pretty fussy but you could submit a pull request later which forces the save button or return to defocus entry items. 

I'll pull this as is.

drgrice1:
>     1. "None Specified" disappears when you click into a field -- once you type one character

This is the expected behaviour of the html placeholder.  Yes, the behaviour is not quite the same as what Andrew had described.

>     2. "None Specified" returns to the field when you click out of that field, leaving it empty. -- non-empty appears when the box is empty, even while still in focus.

Again, the html placeholder does not behave exactly as Andrew described.  I believe that this behaviour is intuitive though.

>     3. The checkbox is not checked when you leave a field empty. -- unchecks when you remove focus.

> However you can delete the field (which does not turn off the check box until you defocus) then hit return (instead of tab) and the page will refresh and the older default date will appear. If you make a change and remove focus by some means other than return then things work as desired.

I see what you are saying, and I will work to fix this.

PR #989, opened 2019-08-22
Need MooX::Options for webwork2 PR #959

taniwallach:
Add installation (in Docker)  and module test for MooX::Options which is needed by new code in https://github.com/openwebwork/webwork2/pull/959

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -187,6 +187,7 @@ RUN apt-get update \
 	libuniversal-isa-perl \
 	libtest-fatal-perl \
 	libjson-xs-perl \
+	libmoox-options-perl \
 	make \
 	netpbm \
 	preview-latex-style \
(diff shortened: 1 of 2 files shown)


mgage:
Despite my tentative concerns about unnecessary bloat listed in comments for PR #959 I plan to merge this pull request.  It simply makes #959, which is already merged, complete. It may be that Moo is sufficiently useful (above and beyond Class::Accessor) and light weight that we want to make it a permanent part of WW.  In any case that decision can wait until we make a comprehensive review of WeBWorK and engage in some basic refactoring. I'm personally looking forward to the idea of basic refactoring as we bring webwork3 on line. Any 20 year old project needs to be rejuvenated from time to time.

drgrice1:
This pull request should not have been merged, at least not until pull request #959 is merged.  At this point, I am not sure if that pull request will be merged or not for ww 2.16.  If it is not, then this should be reverted.

PR #990, opened 2019-08-22
Set $ENABLE_UTF8MB4=1 by default.

taniwallach:
Modern systems have versions of mySQL or MariaDB with uft8mb4 support, and should default to using utf8mb4 as the character set of the tables.

Older systems with mySQL too old to have this support (which was added to mySQL version 5.5.3 in early 2010) **must** change to `$ENABLE_UTF8MB4=0` in `con/site.conf`.

See:
  - https://dev.mysql.com/doc/relnotes/mysql/5.5/en/news-5-5-3.html
  - https://mathiasbynens.be/notes/mysql-utf8mb4
  - https://www.eversql.com/mysql-utf8-vs-utf8mb4-whats-the-difference-between-utf8-and-utf8mb4/

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -184,7 +184,7 @@ $database_storage_engine = 'myisam';
 # encoding utf8mb4 which allows the encoding of characters from many languages
 # including chinese, arabic and hebrew.
 
-$ENABLE_UTF8MB4 =0;    # setting this to 1 enables utf8mb4 encoding, setting this to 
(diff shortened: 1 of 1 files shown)


mgage:
I think you are right that most sites won't run into any problem with this. Some sites may forget
to transfer this from site.conf.dist to site.conf but they should read the instructions. :-)  I've entered a stub in the release notes on the wiki about this but it should be rewritten for clarity. 

I'll merge this PR #990

PR #991, opened 2019-08-23
install zip

heiderich:
Motivation: It will be helpful to create compressed release files of the OPL.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -209,6 +209,7 @@ RUN apt-get update \
 	culmus \
 	fonts-linuxlibertine \
 	lmodern \
+	zip \
     && apt-get clean \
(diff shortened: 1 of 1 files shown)


taniwallach:
Would it not suffice to install it as an extra Ubuntu package on systems which need it (only those which create such release files)?

mgage:
It's pretty clear this will do no harm -- and can easily be reversed if we decide it is superfluous.

heiderich:
Tani is right. It would suffice to install this on systems that create the release files.

I guess it boils down to the question of how many of the potentially needed packages we want to include in the standard image. What is not included may need to be installed in containers (for instance using `docker-entrypoint.sh`). Maybe we will need several configurations for different types of containers in the long term.

Thanks for merging it anyway.

PR #992, opened 2019-08-26
add ARG WEBWORK2_GIT_URL and ARG PG_GIT_URL

heiderich:
This allows to configure the URLs of the cloned repositories.


heiderich:
There are two reasons here:

1) avoid code duplication (here mostly the URLs)
2) make is easier to automate builds with different parameters

Without parameters one would have to change the `Dockerfile` using tools such as `sed`, which I believe to be more error prone.

At the other side: If we really want to use this for example for CI in the long term, I think it would actually be better to use a local copy of these repositories in order to make sure that test results are predictable.
Therefore I once suggested to create a GIT repository with webwork2 and pg as submodules.

PR #997, opened 2019-08-28
rename ADD_PACKAGES to ADD_APT_PACKAGES

heiderich:
Reason: I hope the change of the name adds clarity, also in view of a possible addition of a similar variable controlling the installation of additional CPAN packages.


mgage:
This is simple enough.  There should be no problem with adding this to WW-2.15

PR #998, opened 2019-08-28
Fix an issue with the text block handlers of MathQuill.

drgrice1:
If a text block was created inside another node (like a sqrt) then the textBlockExit handler was not called because the cursor parent did not have the controller.  So we need to navigate up the tree to find it in the same way that we do for the textBlockEnter handler.
To see this bug in action (without this fix) type sqrt or use the square root button on the toolbar, and then use the text button.  Then use back space to delete the text block that is created.  The toolbar remains grayed out, and of course it should become active again.

I also added a minor tweak to the mqeditor.js javascript for the toolbar that eliminates the need for the Object.entries polyfill for browsers that don't have that method.


mgage:
This PR should be reviewed and added to WW-2.15. It's basically a hot fix for a new feature we've already included in WW-2.15

mgage:
This changes only mathquill files.  I will pull it into WeBWorK-2.15.

PR #1002, opened 2019-09-03
Scripts to release versions of webwork problem libraries and to download metadata from such releases

heiderich:
This adds the script `bin/OPL_releases/release.sh` that creates releases of webwork problem libraries such as the Open Problem Library on GitHub. Those releases contain archives of metadata of the library.

In addition, scripts are added to download the metadata from the latest such release.
They are used in `docker-config/docker-entrypoint.sh` to avoid rebuilding the OPL tables in the sql database, which takes a considerable amount of time.
This should therefore significantly speed up the time to get a docker container up and running for the first time.

As the docker configuration in its current form uses a hardcoded branch of webwork2, in order to test this you have to include some new scripts in your container (as long as this is not merged, afterwards this will not be needed anymore). This can be achieved using bind mounts. Just add the following lines to the `volumes` section of the service `app` in your `docker-compose.yml` file:

      - "./bin/OPL_releases:/opt/webwork/webwork2/bin/OPL_releases"
      - "./bin/restore-OPL-tables.pl:/opt/webwork/webwork2/bin/restore-OPL-tables.pl"
      - "./bin/load-OPL-global-statistics.pl:/opt/webwork/webwork2/bin/load-OPL-global-statistics.pl"
      - "./bin/update-OPL-statistics.pl:/opt/webwork/webwork2/bin/update-OPL-statistics.pl"

Note that currently metadata is retrieved from the repository https://github.com/heiderich/webwork-open-problem-library.git. This should be changed once https://github.com/heiderich/webwork-open-problem-library receives regular updates.

### DIFF
--- a/bin/OPL_releases/Makefile
+++ b/bin/OPL_releases/Makefile
@@ -1,0 +1,21 @@
+RELEASE_FILENAME=webwork-open-problem-library-METADATA.tar.gz
+GITHUB_USERNAME=heiderich
+GITHUB_REPO=webwork-open-problem-library
+JSON_FILENAME=latest_release.json
+RELEASE_TAG=latest_release.tag
+
+${JSON_FILENAME}:
+	@echo "Download a JSON containing information on the latest release"
+	curl -s https://api.github.com/repos/${GITHUB_USERNAME}/${GITHUB_REPO}/releases/latest > ${JSON_FILENAME}
+
+${RELEASE_TAG}: ${JSON_FILENAME}
+	jq -r '.tag_name' ${JSON_FILENAME} > ${RELEASE_TAG}
+
+${RELEASE_FILENAME}: ${JSON_FILENAME}
+	curl -L `jq -r '[.assets | .[] | select(.name == "${RELEASE_FILENAME}" ) | .browser_download_url][0]' ${JSON_FILENAME}` -o ${RELEASE_FILENAME}
+
+extract: ${RELEASE_FILENAME}
+	tar xfz ${RELEASE_FILENAME}
+
+clean:
+	rm -f ${RELEASE_FILENAME} ${RELEASE_TAG} ${JSON_FILENAME}
--- a/docker-config/docker-entrypoint.sh
+++ b/docker-config/docker-entrypoint.sh
@@ -56,10 +56,6 @@ if [ ! -d "$APP_ROOT/libraries/webwork-open-problem-library/OpenProblemLibrary" 
   cd $APP_ROOT/libraries/
   /usr/bin/git clone -v --progress --single-branch --branch master https://github.com/openwebwork/webwork-open-problem-library.git
 
-  # FIXME / TO-DO : Download a saved version of the OPL sql table data to be loaded and extract
-  #    it in the appropriate location.This would avoid the need for a length run of OPL-update.
-  # At present, a distribution point has not been set up for such data.
-
   # The next line forces the system to run OPL-update or load saved OPL tables below, as we just installed it
   touch "$APP_ROOT/libraries/Restore_or_build_OPL_tables"
 fi
@@ -135,6 +131,22 @@ if [ "$1" = 'apache2' ]; then
       touch "$APP_ROOT/libraries/Restore_or_build_OPL_tables"
     fi
     if [ -f "$APP_ROOT/libraries/Restore_or_build_OPL_tables" ]; then
+      if [ ! -f "$APP_ROOT/libraries/webwork-open-problem-library/TABLE-DUMP/OPL-tables.sql" ]; then
+        # Download a saved version of the OPL sql table data to be loaded and extract
+        cd $WEBWORK_ROOT/bin/OPL_releases/
+        make extract
+        cp -r webwork-open-problem-library $APP_ROOT/libraries/
+        # check out commit corresponding to the release
+        make latest_release.tag
+        OPL_TAG=`cat $WEBWORK_ROOT/bin/OPL_releases/latest_release.tag`
+        cd $APP_ROOT/libraries/webwork-open-problem-library/
+        git remote rm heiderich || true
+        git remote add heiderich https://github.com/heiderich/webwork-open-problem-library.git
+        git fetch --tags heiderich
+        echo "Checking out tag $OPL_TAG"
+        git checkout $OPL_TAG
+      fi
+
       if [ -f "$APP_ROOT/libraries/webwork-open-problem-library/TABLE-DUMP/OPL-tables.sql" ]; then
         echo "Restoring OPL tables from the TABLE-DUMP/OPL-tables.sql file"
         wait_for_db
--- a/Dockerfile
+++ b/Dockerfile
@@ -211,6 +211,7 @@ RUN apt-get update \
 	fonts-linuxlibertine \
 	lmodern \
 	zip \
+	jq \
     && apt-get clean \
     && rm -fr /var/lib/apt/lists/* /tmp/*
 
--- a/bin/OPL_releases/release.sh
+++ b/bin/OPL_releases/release.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Shell script to make releases of webwork problem libraries, heavily inspired
 # by https://github.com/gap-system/ReleaseTools/blob/master/release
@@ -299,6 +299,26 @@ if [ x"$RELEASE_ID" = x ] ; then
     error "creating release on GitHub failed: no release id"
 fi
 
+
+######################################################################
+#
+# Test whether ARCHIVE_FORMATS contains at least one valid format
+#
+
+FOUND_VALID_FORMAT=0
+VALID_ARCHIVE_FORMATS='.tar.gz .tar.bz2 .zip'
+for VALID_FORMAT in $VALID_ARCHIVE_FORMATS
+do
+	if [[ $ARCHIVE_FORMATS =~ $VALID_FORMAT ]]
+	then
+		FOUND_VALID_FORMAT=1;
+	fi;
+done;
+if [[ $FOUND_VALID_FORMAT == "0" ]]
+then
+	warning "No valid archive format specified." ;
+	exit 1;
+fi;
 
 ######################################################################
 #
--- a/bin/OPL_releases/release.sh
+++ b/bin/OPL_releases/release.sh
@@ -1,0 +1,345 @@
+#!/bin/sh
+#
+# Shell script to make releases of webwork problem libraries, heavily inspired
+# by https://github.com/gap-system/ReleaseTools/blob/master/release
+
+set -e
+
+######################################################################
+#
+# Usage information
+#
+help() {
+cat <<EOF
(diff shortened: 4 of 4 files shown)


mgage:
Florian can this wait for WeBWorK-2.16?  It could be included in 2.15 if it's important to you since it consists mostly of additions that won't get in the way if you don't use them.

docker-config/docker-entrypoint.sh
```diff
@@ -135,6 +131,21 @@ if [ "$1" = 'apache2' ]; then
...
     fi
     if [ -f "$APP_ROOT/libraries/Restore_or_build_OPL_tables" ]; then
+      if [ ! -f "$APP_ROOT/libraries/webwork-open-problem-library/TABLE-DUMP/OPL-tables.sql" ]; then
+        # Download a saved version of the OPL sql table data to be loaded and extract
+        cd $WEBWORK_ROOT/bin/OPL_releases/
+        make extract
+        cp -r webwork-open-problem-library $APP_ROOT/libraries/
+        # check out commit corresponding to the release
+        make latest_release.tag
+        OPL_TAG=`cat $WEBWORK_ROOT/bin/OPL_releases/latest-release_tag`
```
drgrice1: Line 141 doesn't work since the file name is latest_release.tag and not latest-release_tag.  So the OPL_TAG variable does not get set to the correct value.  This results in the following terminal output:
`app_1  | cat: /opt/webwork/webwork2/bin/OPL_releases/latest-release_tag: No such file or directory
webwork2_app_1 exited with code 1`

bin/OPL_releases/get_release_tag.sh
```diff
@@ -0,0 +1 @@
+jq -r '.tag_name' $JSON_FILENAME > $1
```
drgrice1: Is there really a need for the three script files download_metadata.sh, get_latest_release.sh, and get_release_tag.sh?  Those almost single line files should be incorporated directly into the Makefile.

bin/OPL_releases/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	export JSON_FILENAME=${JSON_FILENAME}; ./get_release_tag.sh ${RELEASE_TAG}
+
+${RELEASE_FILENAME}: download_metadata.sh ${JSON_FILENAME}
+	RELEASE_FILENAME=${RELEASE_FILENAME} JSON_FILENAME=${JSON_FILENAME} ./download_metadata.sh
+
+extract: ${RELEASE_FILENAME}
+	tar xfz ${RELEASE_FILENAME}
+
+clean:
+	rm ${RELEASE_FILENAME} ${RELEASE_TAG} ${JSON_FILENAME}
```
drgrice1: Usually make file targets do not depend on files that are not generated by the make file itself and are not themselves targets in the make file.  For example you have the ${JSON_FILENAME} target depend on the script get_latest_release_json.sh.

taniwallach:
> This certainly is needed to speed up getting a docker build up and running.

I'm also for getting this into WW 2.15, if we can.

docker-config/docker-entrypoint.sh
heiderich: Thanks. Fixed in 2fc5d10.

bin/OPL_releases/get_release_tag.sh
heiderich: There is no real need for those files now. I created them just in case one may need them later for something else. But in 0675134 I incorporated them into the `Makefile`.

heiderich:
@mgage There is no strict need. But this could speed up testing significantly.

bin/OPL_releases/Makefile
```diff
@@ -0,0 +1,21 @@
+RELEASE_FILENAME=webwork-open-problem-library-METADATA.tar.gz
+GITHUB_USERNAME=heiderich
+GITHUB_REPO=webwork-open-problem-library
+JSON_FILENAME=latest_release.json
+RELEASE_TAG=latest_release.tag
+
+${JSON_FILENAME}:
+	echo "Download a JSON containing information on the latest release"
+	GITHUB_USERNAME=${GITHUB_USERNAME} GITHUB_REPO=${GITHUB_REPO} curl -s https://api.github.com/repos/${GITHUB_USERNAME}/${GITHUB_REPO}/releases/latest > ${JSON_FILENAME}
```
drgrice1: There is no need to re-define the make variables GITHUB_USERNAME and GITHUB_REPO to be themselves before the curl command.  The same is true for JSON_FILENAME in the ${RELEASE_TAG} target below.

bin/OPL_releases/Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+${RELEASE_TAG}: ${JSON_FILENAME}
+	JSON_FILENAME=${JSON_FILENAME} jq -r '.tag_name' ${JSON_FILENAME} > ${RELEASE_TAG}
+
+${RELEASE_FILENAME}: ${JSON_FILENAME}
+	LATEST_RELEASE_FILENAME_URL=`jq -r '[.assets | .[] | select(.name == "${RELEASE_FILENAME}" ) | .browser_download_url][0]' ${JSON_FILENAME}` curl -L $$LATEST_RELEASE_FILENAME_URL -o ${RELEASE_FILENAME}
```
drgrice1: This isn't working.  Instead try
``curl -L `jq -r '[.assets | .[] | select(.name == "${RELEASE_FILENAME}" ) | .browser_download_url][0]' ${JSON_FILENAME}` -o ${RELEASE_FILENAME}``

bin/OPL_releases/Makefile
```diff
@@ -0,0 +1,21 @@
+RELEASE_FILENAME=webwork-open-problem-library-METADATA.tar.gz
+GITHUB_USERNAME=heiderich
+GITHUB_REPO=webwork-open-problem-library
+JSON_FILENAME=latest_release.json
+RELEASE_TAG=latest_release.tag
+
+${JSON_FILENAME}:
+	echo "Download a JSON containing information on the latest release"
```
drgrice1: You can clean up the make output by putting an @ before "echo" commands.  Without the @ you get
`echo "Download a JSON containing information on the latest release"`
`Download a JSON containing information on the latest release`
in the output.
With the @ you only get the second line.

bin/OPL_releases/Makefile
```diff
@@ -0,0 +1,21 @@
...
+	JSON_FILENAME=${JSON_FILENAME} jq -r '.tag_name' ${JSON_FILENAME} > ${RELEASE_TAG}
+
+${RELEASE_FILENAME}: ${JSON_FILENAME}
+	LATEST_RELEASE_FILENAME_URL=`jq -r '[.assets | .[] | select(.name == "${RELEASE_FILENAME}" ) | .browser_download_url][0]' ${JSON_FILENAME}` curl -L $$LATEST_RELEASE_FILENAME_URL -o ${RELEASE_FILENAME}
+
+extract: ${RELEASE_FILENAME}
+	tar xfz ${RELEASE_FILENAME}
+
+clean:
+	rm ${RELEASE_FILENAME} ${RELEASE_TAG} ${JSON_FILENAME}
```
drgrice1: Add the -f flag to rm to suppress "No such file or directory" warnings for files that don't exist.

heiderich:
Thanks for your review @drgrice1. I addressed all your points in 1933501.

drgrice1:
This looks good now!

taniwallach:
Overall it basically works, but I ran into some small issues when testing, so think some small additional changes are probably needed.

Notes on testing the code to make an OPL meta-data release:
1. For the OPL.
2. Create a GitHub token
3. Locally modify `bin/OPL_releases/release.sh` and comment out the line `rm -f .git* .hg* .cvs*
`
4. Run:

```
# Get into bash in the Docker container
docker container exec -it webwork2_app_1 bash

# Test making a release:
cd /opt/webwork/libraries/webwork-open-problem-library

# Needed to avoid the mkdir bug
rm -rf  tmp

# Needed to comment out the "rm" line which wanted to delete .git to get the following to work:
/opt/webwork/webwork2/bin/OPL_releases/release.sh -t "my-tag-name" -r taniwallach/webwork-open-problem-library --token my-secret-token  --libraryname webwork-open-problem-library
```

taniwallach:
> @taniwallach I merged your PR with the suggested changes. Thank you.

Thanks! I'll test again later before I merge the PR, unless you think there is any reason to wait longer.

heiderich:
Just modified a error message. I would be pleased if you could test it again.

Before merging I would suggest to squash the commits of this PR into one if you agree (there is one commit of you involved, @taniwallach). I think this would keep the history cleaner.

taniwallach:
Tested after the changes, and everything ran smoothly, even when I tried to trigger the problems encountered before.

@heiderich Additional comments:
1. Maybe use a default value for ARCHIVE_FORMATS (say at least ".tar.gz") if nothing was set? My first releases did not create the METADATA files as I had not set any value. Alternative: Display a warning when the script is run without a (valid) value for ARCHIVE_FORMATS set or in the case that no METADATA archive file is created.
2. I have no problems with you doing a "squash" and revising/replacing the PR with one with just one commit.

---

**Updated notes** on testing the code to make an OPL meta-data release:
```
# Get into bash in the Docker container
docker container exec -it webwork2_app_1 bash

# Test making a release:
cd /opt/webwork/libraries/webwork-open-problem-library

# Need to set an environment variable for ARCHIVE_FORMATS to be created (otherwise the release is missing them)
export ARCHIVE_FORMATS=".tar.gz .tar.bz2"

/opt/webwork/webwork2/bin/OPL_releases/release.sh -t "my-tag-name" -r GitHubUserName/webwork-open-problem-library --token my-secret-token  --libraryname webwork-open-problem-library
```

heiderich:
I squashed the previous commits into only two.
In c500a115e84321ae8d3ed235d1ae473b50f36e05 I added a sanity check for ARCHIVE_FORMATS.
In 35e6d3207e5033b2753550fe10c7a33ffba77086 I added a call of `OPL-update` before dumping the OPL tables. Otherwise we may create a release with old metadata. (I guess this happened to me in the past.) I could not yet definitively test whether calling OPL-update may cause problems. But I think we should do this.

taniwallach:
> In [35e6d32](https://github.com/openwebwork/webwork2/commit/35e6d3207e5033b2753550fe10c7a33ffba77086) I added a call of `OPL-update` before dumping the OPL tables. Otherwise we may create a release with old metadata. (I guess this happened to me in the past.) I could not yet definitively test whether calling OPL-update may cause problems. But I think we should do this.

Really you would want a `git fetch origin; git pull` in the OPL directory before the `OPL-update` but maybe that should be handled manually or outside the `OPL_releases/release.sh` script.

heiderich:
I agree that this would be nice. However I would rather consider this the responsibility of the user of `release.sh`. In the long term, maybe we can create another script that handles everything (along with creation of a meaningful release name and so on). I already have a draft for it, but it is too early to include it in this PR.

In the meantime I tested 35e6d32. It seems to work fine for me.

taniwallach:
So I think we should merge this in.

Issue #901, opened 2018-11-21
Refreshing page resubmits answers


glennricster:
If a student submits an answer on a problem, and then refreshes the page after the problem reloads with the results and confirms form resubmission, then the browser will resubmit the form from its cache. The server then grades the problem again and adds an attempt on the problem. So one of the students' attempts on the problem is wasted.

Of course that is only a problem if the problem is set up with a limited number of attempts. However, things get worse if the course is set up to request a new version of the problem after a certain number of attempts. In this case, if the student submits answers until the "Request New Version" button appears and then refreshes the page and confirms form resubmission, the server will still grade the problem again and increment the number of attempts and reload that same problem again with 2 more attempts left. A new seed and problem is not generated. Using this the student can effectively work the same problem an unlimited number of times.

taniwallach:
@drgrice1 - I think this issue can be closed, as https://github.com/openwebwork/webwork2/pull/1003 fixes the problem for WW 2.15.

drgrice1:
Yep. Closing.

PR #1003, opened 2019-09-03
Restructure problem randomization and fix issue #901

drgrice1:
Restructure the way that problem randomization works and add a check to test for a browser refresh to prevent a cached form resubmission from circumventing a new seed being generated when it should be.

This fixes issue #901 (which I submitted).

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Problem.pm
+++ b/lib/WeBWorK/ContentGenerator/Problem.pm
@@ -566,8 +566,10 @@ sub pre_header_initialize {
 
 	# Check for a page refresh which causes a cached form resubmission.  In that case this is
 	# not a valid submission of answers.
-	$submitAnswers = 0 if ($submitAnswers && defined($formFields->{num_attempts}) &&
-		$formFields->{num_attempts} != $problem->num_correct + $problem->num_incorrect);
+	$submitAnswers = 0, $self->{resubmitDetected} = 1
+	if ($submitAnswers && (!defined($formFields->{num_attempts}) ||
+			(defined($formFields->{num_attempts}) &&
+				$formFields->{num_attempts} != $problem->num_correct + $problem->num_incorrect)));
 
 	$self->{displayMode}    = $displayMode;
 	$self->{redisplay}      = $redisplay;
@@ -749,6 +751,7 @@ sub pre_header_initialize {
 		# If this happens, it means that the page was refreshed.  So prevent the answers from
 		# being recorded and the number of attempts from being increased.
 		if ($problem->{prCount} > $rerandomizePeriod) {
+			$self->{resubmitDetected} = 1;
 			$must{recordAnswers} = 0;
 			$can{recordAnswers} = 0;
 			$want{recordAnswers} = 0;
@@ -1945,7 +1948,11 @@ sub output_summary{
 	    # don't show attempt results (correctness)
 	    # show attempt previews
 	  }
-	
+
+	  print CGI::div({class=>'ResultsWithError'},
+		  $r->maketext("ATTEMPT NOT ACCEPTED -- Please submit answers again (or request new version if neccessary).")),
+	  CGI::br() if ($self->{resubmitDetected});
+
 	if ($set->set_id ne 'Undefined_Set' && $set->assignment_type() eq 'jitar') {
 	my $hasChildren = 0;
 	my @problemIDs = $db->listUserProblems($effectiveUser, $set->set_id);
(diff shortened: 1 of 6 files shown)


Alex-Jordan:
I changed this PR to go to develop instead of master. Assuming that is what you intend. But is this something to consider for version 2.15?

drgrice1:
I did intend it for consideration for 2.15.  It was an issue that was marked to be fixed for the next release.  However, if you think it should be delayed then move it to develop.

Alex-Jordan:
OK, now it points to the 2.15 branch. @mgage can decide if it should also be merged into develop. It's just that merging into master is not the right thing at this stage.

drgrice1:
Is 2.15 considered a release at this point?  I didn't realize that it was.  If that is the case, then you are correct, and this should be moved to develop.

Alex-Jordan:
@mgage will know the right thing to do. I could tell that it should not go to master, which is still version 2.14. The telltale sign something was wrong was when I got a message from GitHub indicating a PR with several hundred commits and file changes.

drgrice1:
I thought I had it set to go to 2.15, and not to master.  I guess I accidentally switched it again.

Alex-Jordan:
Ah right. I find in GH that when I set the target repository and branch, then set the source repository and branch, that it reverts the target to the defaults. That must have happened here. For whatever reason (maybe a GH bug), the source has to be set first.

mgage:
It's true that I'm trying to wrap up webwork-2.15 and merge it in to master.  Things can get confusing when there are too many active branches at play.   I think this bugfix is sufficiently urgent that we should try to include it in webwork-2.15 as soon as we can get it reviewed by a few people. 

I have been marking PR that I thought should wait for WW-2.16 by setting that label in "milestones".
WeBWorK-2.15 already has major changes to allow for 4 byte characters and all languages. and improvements to the docker downloads. I don't want to overload it with too many more changes.

mgage:
Are there any additional comments on this PR fixing the problem randomization issue?  I'm about to merge it into WeBWorK-2.15

lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -563,14 +563,18 @@ sub pre_header_initialize {
 	my $requestNewSeed            = $r->param("requestNewSeed") // 0;
 
 	my $formFields = { WeBWorK::Form->new_from_paramable($r)->Vars };
-	
+
+	# Check for a page refresh which causes a cached form resubmission.  In that case this is
+	# not a valid submission of answers.
+	$submitAnswers = 0 if ($submitAnswers && defined($formFields->{num_attempts}) &&
```
taniwallach: What would happen if somewhat tampered with the form submission  and deleted the `num_attempts` field?
I think that if this field is missing, then maybe the submission should be rejected.

drgrice1: Yeah, that makes sense.  I will delete the check to see if the num_attempts key of the $formFields hash is defined.  However, it is also possible for someone to change the value of the hidden num_attempts field in the form submission and increment it.  The possibility of that form field being hacked is taken into account in this PR.  Basically, if a student hacks the form field for a regular submission (when it is not time to create a new version), then the student will only be throwing an attempt away.  So there is no benefit.  If a student does hacks the form field when it is time to reseed for a new version, then the check on line 751 catches that.  The point is the prCount for the current problem will still be incremented in that case, and the check on line 751 will catch it and stop the problem from being recorded and from skipping over reseeding which is what happens without this PR.

taniwallach: I think you may need to handle the not-defined case separately, as Perl may complain (or `warn`) if you try to use the value when it it not defined. It might be easier to revert the last commit and just add another line:
```
	$submitAnswers = 0 if ($submitAnswers && !defined($formFields->{num_attempts}));
```
rather than handling it by a more complicated set of chained conditions.

taniwallach:
Question: Could you trigger the display of some error message (`warn`???) in the case that the `num_attempts` test fails and use the error message to inform that student it will be necessary to reload the problem and then resubmit the answer, if the submission was intentional?

lib/WeBWorK/ContentGenerator/Problem.pm
drgrice1: Perl won't complain (or `warn`) if you try to compare a hash key that is not defined to another value. The != comparison will be false.  So this will work the way it is intended.

drgrice1:
In that scenario there is no need to reload the problem.  You are correct that the "regular" submission will not work in the second browser the first time, but it will the second.

I will add a warn message in this case.  I personally don't really like a warn here though.  What do you think the message should say?

drgrice1:
Nathan:  What do you think of the approach in the last commit, instead of a `warn`.

taniwallach:
> Nathan: What do you think of the approach in the last commit, instead of a `warn`.

Great!
@mgage was almost ready to merge before these last minor changes, so I'm going to go ahead and do the merge.

PR #1005, opened 2019-09-08
Stop search engines indexing individual courses

drdrew42:
I have multiple _hidden_ model courses on my WeBWorK server.

These courses are not intended for students or professors, but they are indexed by Google and seem **very** appealing in a search... 

result of search

Students' logins fail (because they're logging into the model course, rather than their professor's section) and confusion ensues.

This PR adds a meta tag to the header of all sections' login pages, thereby preventing individual courses from being indexed by (most) search engines. (I'm already using this code, I don't know if anyone else has run into a similar problem...)

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Login.pm
+++ b/lib/WeBWorK/ContentGenerator/Login.pm
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright © 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
+# Copyright Â© 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
 # $CVSHeader: webwork2/lib/WeBWorK/ContentGenerator/Login.pm,v 1.47 2012/06/08 22:59:55 wheeler Exp $
 # 
 # This program is free software; you can redistribute it and/or modify it under
@@ -147,6 +147,10 @@ sub pre_header_initialize {
 	}
 }
 
+sub head {
+        print '<meta name="robots" content="noindex" />';
+        return "";
+}
 
 sub body {
 	my ($self) = @_;


dlglin:
Is this the desired default behaviour?  I could see some institutions preferring to have google index their course login pages to make it easier for students to find them.

On Sep 8, 2019, at 11:58 AM, drdrew42 <notifications@github.com<mailto:notifications@github.com>> wrote:


I have multiple hidden model courses on my WeBWorK server.

These courses are not intended for students or professors, but they are indexed by Google and seem very appealing in a search...

[result of search]<https://user-images.githubusercontent.com/3912882/64492273-4e9a8480-d240-11e9-9d1d-b25c57a96069.png>

Students' logins fail (because they're logging into the model course, rather than their professor's section) and confusion ensues.

This PR adds a meta tag to the header of all sections' login pages, thereby preventing individual courses from being indexed by (most) search engines. (I'm already using this code, I don't know if anyone else has run into a similar problem...)

You can view, comment on, or merge this pull request online at:

  https://github.com/openwebwork/webwork2/pull/1005

Commit Summary

  *   Stop search engines indexing individual courses

File Changes

  *   M lib/WeBWorK/ContentGenerator/Login.pm<https://github.com/openwebwork/webwork2/pull/1005/files#diff-0> (6)

Patch Links:

  *   https://github.com/openwebwork/webwork2/pull/1005.patch
  *   https://github.com/openwebwork/webwork2/pull/1005.diff

drdrew42:
Would it be preferable to add a site-level configuration to handle this?

drgrice1:
I think having this as a site-level configuration option is a good idea.  Is there a way that it can be done per course even?  I use LTI authentication and disable direct login for my courses, and so I don't want those courses to appear in search engines.  However, if other faculty are using the same server but don't use LTI authentication they could benefit from having their courses appear.

taniwallach:
I recommend trying to make this change depended on the value of a configuration setting.

I think (based on other `head` subroutines in `lib/WeBWorK/ContentGenerator/`) that it should be possible to add
```
        my ($self) = @_;
        my $ce = $self->r->ce;
``` 
to the start of the new subroutine, and then make this behavior conditional on a configuration variable set in the `CourseEnvironment` which could come from either:
  - `conf/defaults.config`,
  - `conf/localOverrides.conf` (I think), 
  - or a course specific `course.conf`.
A fallback behavior can be set in case the expected configuration variable is not defined in the `CourseEnvironment`.

taniwallach:
@drdrew42 Can you rebase this off the WeBWorK-2.15 branch? I think it can merge in to master when WW 2.15 releases rather than be be treated as a hotfix to 2.14. (@mgage - do you agree?)

@dlglin @mgage - Would you recommend the default from `conf/defaults.config` be to keep the current behavior (no header added to attempt to prevent search engine indexing) and the sample line in `conf/localOverrides.conf` to be a sample flipping the value, or the other way around?
I take it that you

drdrew42:
#1017 adds configuration support and rebases onto 2.15, closing this PR in favor of that one.

Issue #1007, opened 2019-09-11
"Specified key was too long" error with MySQL 5.7


jerrac:
When trying to run Webwork 2.15 via Docker, the app container exits after throwing this error:

```
DBD::mysql::db do failed: Specified key was too long; max key length is 1000 bytes at /opt/webwork/webwork2/bin/OPL-update line 281.
```
When searching for an answer, I found http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4341 From that, it looks to like the issue was fixed months ago. 

I think it has to have something to do with MySQL 5.7 since I have been able to get Webwork up and running with the MariaDB container in the example docker-compose file.

Our database server runs MySQL 5.7. I can't run MariaDB as a container right now outside of my local development environment.

I have tried making sure my database is using utf8mb4 and utf8mb4_general_ci. As well as the non-mb4 versions. I've tried deleting and recreating the database.

This is pretty much what I see every single time.

```bash
docker-compose -f dev.docker-compose.yml up --force-recreate
Creating network "lanewebwork_webwknet" with driver "bridge"
Creating webwork_rserve ... done
Creating webwork_frontend ... done
Attaching to webwork_rserve, webwork_frontend
webwork_rserve | 
webwork_rserve | R version 3.3.3 (2017-03-06) -- "Another Canoe"
webwork_rserve | Copyright (C) 2017 The R Foundation for Statistical Computing
webwork_rserve | Platform: x86_64-pc-linux-gnu (64-bit)
webwork_rserve | 
webwork_rserve | R is free software and comes with ABSOLUTELY NO WARRANTY.
webwork_frontend | Setting system timezone to America/Los_Angeles
webwork_rserve | You are welcome to redistribute it under certain conditions.
webwork_rserve | Type 'license()' or 'licence()' for distribution details.
webwork_rserve | 
webwork_rserve | R is a collaborative project with many contributors.
webwork_rserve | Type 'contributors()' for more information and
webwork_rserve | 'citation()' on how to cite R or R packages in publications.
webwork_rserve | 
webwork_rserve | Type 'demo()' for some demos, 'help()' for on-line help, or
webwork_rserve | 'help.start()' for an HTML browser interface to help.
webwork_rserve | Type 'q()' to quit R.
webwork_rserve | 
webwork_rserve | > Rserve::run.Rserve(remote=TRUE, auth=FALSE, daemon=FALSE)
[...]

PR #1009, opened 2019-09-13
In bin/OPL-update, change keyword table's keyword column to varchar(2…

jerrac:
…45) so that mysql 5.7 doesn't complain about keys being to long.

Fix for #1007 .


taniwallach:
Merged - Thanks!

taniwallach:
Merging the change.

PR #1010, opened 2019-09-14
Fix gateway correct/partialy correct/incorrect message.

drgrice1:
This is a continual source of confusion to students.  They take a gateway quiz and it shows that they have missed some points, but on reviewing the quiz they see at the bottom "All of the answers are correct" on all problems.  This is an obvious bug that needs to be fixed.  A score greater than zero is not the correct condition for "completely" correct.

There is really more that needs to be done here.  Line 471 doesn't work right.  It is using the score on the last question in the problem only to determine "complete" correctness of all questions in the problem.


taniwallach:
I won't have a chance to test this soon, but the proposed change cannot be worse than the current code which totally ignores the possibility of partial credit. I think we should merge the change.

mgage:
I agree.

PR #1011, opened 2019-09-15
We b wor k 2.15+fixed error message

mgage:
Fix/improve error messages in WeBWorK.pm (the error message confused students -- hopefully the new one is better).

and DB.pm (the error message was simply incorrect)


drgrice1:
This looks good to me.  I am all for fixing error messages that are confusing or incorrect.

PR #1017, opened 2019-09-20
Meta Robots tags for hiding individual courses from search engines

drdrew42:
The course environment will support a new option: `$ce{options}{metaRobotsContent}`, which allows individual courses to control whether or not they are indexed by search engines.

The value of `$ce{options}{metaRobotsContent}` should be a comma-separated list of valid options for the meta robots tag. This list of widely-supported options includes: index, noindex, follow, nofollow, and noarchive. Google additionally supports the "unavailable_after:" option with a date in RFC850 format. For example, "index, unavailable_after: 20-Sep-2019 18:00:00 EST"

This PR replaces #1005 and rebases onto WW-2.15

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -317,4 +317,13 @@ $siteDefaults{timezone} = "America/New_York";
 # If you do not fill this in, the system will default to "en_US"
 $siteDefaults{locale}="";
 
+################################################################################
+# Search Engine Indexing Enable/Disable
+################################################################################
+# sets the default meta robots content for individual course pages
+# this will not stop your main course listing page from being indexed
+# valid contents: index, noindex, follow, nofollow, noarchive, and
+# unavailable_after (example: "index, unavailable_after: 23-Jul-2007 18:00:00 EST")
+$options{metaRobotsContent}='noindex, nofollow';
+
 1; #final line of the file to reassure perl that it was read properly.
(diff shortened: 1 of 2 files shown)


mgage:
Looks ok just reading it.  What happens if they leave the metaoptions blank?  Will that still work ok? 
I'm assuming that it would.

drdrew42:
Yes, if the hash contains a blank string or if the key-value does not exists, WeBWorK defaults to “none” which is interpreted as ‘noindex, nofollow.’

If the consensus is to default to allowing search engine indexing, this can easily be swapped for ‘index, follow.’

mgage:
I think a blank or a space will be passed through by //.
 only undefined triggers 'none'
|| Maybe what you want


sent from my phone, Mike

drdrew42:
My understanding of the meta robots tag is that invalid content in the tag would simply be ignored by the robots. So a `<meta name="robots" content=" ">` tag would use the robots default, which is "index, follow."

The key-value being undefined indicates that the current installation hasn't updated their `site.conf`, in which case - default to block search engine indexing of individual courses. If the key-value exists, that means the choice has been made to deliberately use an empty (or blank) string. This is the same as someone using invalid options... robots will ignore and revert to their default behavior in the absence of valid instructions indicating otherwise.

PR #1018, opened 2019-09-25
Code to support adding additional CSS files requested by problems

taniwallach:
The webwork2 code added to provides mechanisms to add CSS files request by a specific PG file (by use of the new `ADD_CSS_FILE()` method in PG) or "globally" requested via `course.conf` (or even via `conf/localOverrides.conf`) by setting a value of `$pg{specialPGEnvironmentVars}{extra_css_files}`. The value should be an anonymous array. Ex:
```
$pg{specialPGEnvironmentVars}{extra_css_files} = ['file1.css', 'file2.css'];
```

The requested CSS files need to be stored in `webwork2/htdocs/css`.

The pg side is in https://github.com/openwebwork/pg/pull/428 and depends on this PR. However, this PR does not depend on the PR for pg.

Motivation: http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4624

A sample `htdocs/css/rtl.css` is included with a very preliminary bit of CSS code for RTL courses, in this case to move radio buttons to the right of the text.

---

A small amount of cleanup work was done to existing nearby code.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Problem.pm
+++ b/lib/WeBWorK/ContentGenerator/Problem.pm
@@ -2268,20 +2268,46 @@ sub output_CSS {
 	my $site_url = $ce->{webworkURLs}->{htdocs};
 
         # Javascript and style for knowls
-        print qq{
-           <link href="$site_url/css/knowlstyle.css" rel="stylesheet" type="text/css" />};
+        print "<link href=\"$site_url/css/knowlstyle.css\" rel=\"stylesheet\" type=\"text/css\" />\n";
 
 	#style for mathview
 	if ($self->{will}->{useMathView}) {
-	    print "<link href=\"$site_url/js/apps/MathView/mathview.css\" rel=\"stylesheet\" />";
+	    print "<link href=\"$site_url/js/apps/MathView/mathview.css\" rel=\"stylesheet\" />\n";
 	}
 	
 	#style for mathquill
 	if ($self->{will}->{useMathQuill}) {
-		print "<link href=\"$site_url/js/apps/MathQuill/mathquill.css\" rel=\"stylesheet\" />";
-		print "<link href=\"$site_url/js/apps/MathQuill/mqeditor.css\" rel=\"stylesheet\" />";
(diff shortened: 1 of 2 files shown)


lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -2268,20 +2268,46 @@ sub output_CSS {
 	my $site_url = $ce->{webworkURLs}->{htdocs};
 
         # Javascript and style for knowls
-        print qq{
-           <link href="$site_url/css/knowlstyle.css" rel="stylesheet" type="text/css" />};
+        print "<link href=\"$site_url/css/knowlstyle.css\" rel=\"stylesheet\" type=\"text/css\" />\n";
```
drgrice1: Why switch from `qq{}` to double quotes?  The neatness of `qq` is that you don't need to escape double quotes that are part of the string which makes things more readable.  Note that `\n` will still work in `qq`.  Although, I don't think that the newline should be added.  I know that it makes the generated html more readable, but readability should not be the goal with html.  Minimization should be the goal, and newlines are simply not necessary, and so should be dropped.  Furthermore, in some cases newlines can even make things render incorrectly by causing text nodes to be created.

taniwallach: The `qq` was followed by a line with leading spaces... 

I think there should be a reasonable balance between human readability and minimization in anything a human might reasonably want to debug. I have spent lots of time in recent years debugging issues by examining and editing HTML code from either "view source", in browser developer tools, or via downloaded HTML files. Far too often I end up trying to pare down a problem file to a minimal example triggering the issue. As such, I wanted to clean up the layout of the generated HTML code to a small extent.

Probably the `qq{` could have been just moved to the following line to avoid that, but the other lines all just printed without use of `qq` so I figured sticking to one approach in all the prints in the subroutine was simpler.

New lines in the `<head>` between such `<link>` lines should not cause any problems, to the best of my knowledge.

drgrice1: I spend a lot of time utilizing "view source" as well, and I agree it is a pain dealing with the long lines.  This is fine.

PR #1024, opened 2019-09-26
Add tests of unique ID tools , remove obsolete develop configurations

mgage:


mgage:
submitted to wrong branch

Issue #1020, opened 2019-09-26
Applet problems are broken in Gateway Quizzes


drgrice1:
This was reported in http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4655.

There are several problems that prevent these sort of problems from working in gateway quizzes.

First, HEADER_TEXT is broken for gateway quizzes in general.

Second, the javascript for the addOnLoadEvent is not loaded for gateway quizzes, and that is used by the ww_applet_support.js javascript loaded in AppletObjects.pl.

Third, is some issue with named answers that I do not fully understand yet.

drgrice1:
Actually, I think the real problem is not the named answers, but the submit action.  Problem.pm calls the submitAction javascript function anytime the problemMainForm is submitted.  The gateway quiz form does not.  The ww_applet_support.js redefines that function, and that is where the javascript values are put into the hidden inputs.

drgrice1:
There seem to be a lot of problems here.  Even if the submitAction javascript is made to work it is not enough.  The results table is still empty.  I am not sure if this has to do with the named answers.  In any case, the named answers not having the quiz prefix will be a problem if these types of problems are used on multipage quizzes.  I am moving on from this for now.  I admit I have a bit of a lack of interest as I do not use these types of problems.

drgrice1:
I have a fix for this now for GeogebraApplet problems.  However, there is no way to make this work for the existing problems.  Those problems will have to be changed to make them work in gateway quizzes.

So there are two ways to proceed.

The first is to make a new macro that newly written Geogebra problems should use with the way that they implement the Geogebra applet slightly modified.  The existing problems will continue to work in homework with the old macro, but will not work in gateway quizzes.

The other way is to modify the existing macro (and Applet.pm pg lib module), and require that existing problems make the neccesary changes to function with the modified macro.  This is the way that I think it should be done.  There are only a few changes that will need to be made to fix the problems.  Basically, the problems will have to start giving the applet for the problem a unique appletName (same as the appletId) and stop using names like "ggbfoo" in every problem.  No problem should ever define a global javascript ggbOnInit method.  That will be handled by code provided to the onInit option which will be called in a global ggbOnInit method defined in the ww_applet_support.js file.  Also there will be a need to take some care with the answer names.

[...]

PR #1026, opened 2019-09-26
Fix HEADER_TEXT in gateway quizzes

drgrice1:
Make the HEADER_TEXT method work for gateway quizzes.  This is one piece of the puzzle needed to fix issue #1020.

The head method defined on line 1302 of GatewayQuiz.pm was apparently copied from Problem.pm and not modified to actually work.

### DIFF
--- a/htdocs/themes/math4/gateway.template
+++ b/htdocs/themes/math4/gateway.template
@@ -52,8 +52,8 @@
 <!-- [gateway] since the left-side menus are gone, don't indent the main content area -->
 
 <title><!--#path style="text" text=" : " textonly="1"--></title>
-<!--#head-->
 <!--#output_JS-->
+<!--#head-->
 </head>
 <body>
 <a href="#content" id="stmc-link" class="sr-only sr-only-focusable">Skip to main content</a>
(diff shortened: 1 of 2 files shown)


drgrice1:
> I think this is fine. My distant memory is that there was some question of whether loading the header text from all problems sequentially would result in either duplicate or conflicting data or code.

I suspected that may have been the case.  There certainly can be duplication and possible conflicts that could arise.  If it is decided that allowing the header to be set from a problem in gateway quizzes is a bad idea, then this function should be made a stub (or even remove the "head" call from the template) rather than broken code.

mgage:
This works ok.  You get dozen or so calls to MathJax in the header section but they don't seem to be interfering with each other. 

A long term project: 
1. there must be javaScript call or framework that keeps you from loading a javaScript function or file twice.  That would be a useful addition, but doesn't seem to be immediately necessary.
2. we should also create a means of encapsulating problems when necessary (iframes?) Even for ordinary problems the CSS files associated with downloading Sage applets conflict with the CSS that decorates the Next/Previous buttons and some other checkbox items that instructors see.

Issue #1022, opened 2019-09-26
FileManager - Give better response if a download of more than one file is requested


taniwallach:
If I select multiple files and click Download in the FileManager, the displayed page looks unusual (layout messed up) and the warnings are not particularly helpful. Giving a nicer error message and a regularly formatted page would be nice.

```
WeBWorK Warnings
WeBWorK has encountered warnings while processing your request. If this occured when viewing a problem, it was likely caused by an error or ambiguity in that problem. Otherwise, it may indicate a problem with the WeBWorK system itself. If you are a student, report these warnings to your professor to have them corrected. If you are a professor, please consult the warning output below for more information.

Warning messages
Use of uninitialized value $pwd in string eq at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm line 246.
Use of uninitialized value $pwd in string eq at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm line 246.
Use of uninitialized value $dir in substitution (s///) at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm line 1032.
Use of uninitialized value $dir in split at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm line 1033.
Use of uninitialized value $root in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm line 1052.
Use of uninitialized value $pwd in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm line 1052.
Request information
Time	Thu Sep 26 13:15:35 2019
Method	POST
URI	/webwork2/spring17mth164/instructor/file_manager/
```
followed by
```
You can only download one file at a time.
```

mgage:
These are only "undefined" warning messages (the last message is the real one).   It's probably safe to use  the construction $foobar//'';        to uniformly make sure that every value is defined.

drgrice1:
It seems that it would be nice to update this to create a zip file (or tar ball) and offer that for download in the event that multiple files are selected.  Of course that takes some work to implement.

In any case the warnings and the garbled result should be fixed for now!

dpvc:
[...]

PR #1027, opened 2019-09-27
Check for proper action (non-capitalized) in getFile().

dpvc:
This undoes incorrect localization from commit 2a4ceac7d. The `$action` variable here is not from the UI, but is a hard-coded constant passed to `getFile()` at each invocation, e.g., `$self->getFile("download')`.  The change to capitalization caused the `Refresh()` method to be called when it should not be.  Also, there is no need to localize the action, since its value is never localized initially.  (It may want to be, since it is inserted into a message, but that is a different question; it is usually a bad idea to localize a single word inserted into a sentence like this, so the whole thing may need to be adjusted, for example, taking localized complete sentences from a hash indexed by $action, for example.)

Resolves issue #1022.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -239,6 +239,7 @@ sub HiddenFlags {
 #
 # Display the directory listing and associated buttons
 #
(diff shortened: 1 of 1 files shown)


taniwallach:
@dpvc - Thanks!

PR #1029, opened 2019-09-29
Add handling for a new LTI setting: $block_lti_create_user

taniwallach:
Add handling for a new LTI setting: `$block_lti_create_user` intended for use in an LTI course where automatic LTI account creation is not desired, ex. after the end of a semester if new accounts may potentially be added to the LMS. The following setting can be added to the `course.conf` file:
```
	$block_lti_create_user = 1;
```
By default, accounts will be created, such as when this variable is not defined.

The `return 0;` is used to trigger a general authentication error rather than a page full of strange warning that a student would not understand.

Some formatting cleanup was done, but only to the inner blocks, so that the indentation of the outer blocks remain in sync with the existing code.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -421,23 +421,39 @@ sub authenticate {
       debug("OAuth verification SUCCEEDED !!");
       
       my $userID = $self->{user_id};
+
+	# Indentation of the internal blocks below was modified to follow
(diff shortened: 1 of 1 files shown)


taniwallach:
Since the new code is only triggered is someone adds a new setting to their control files, I think that it is reasonable to make this a last minute addition to the WW 2.15 code.

I have tested it on a production server using LTI, and the code does prevent new account creation when the special setting is added to `course.conf` and things work as before if that setting is not included.

PR #1030, opened 2019-09-29
Implement TikZ images for problems

drgrice1:
See https://github.com/openwebwork/pg/pull/432

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -112,6 +112,11 @@ $externalPrograms{pdflatex} ="/usr/bin/pdflatex --shell-escape";
 
 $externalPrograms{dvipng}   ="/usr/bin/dvipng";
 
+# In order to use imagemagick convert you need to change the rights for PDF files from
+# "none" to "read" in the policy file /etc/ImageMagick-6/policy.xml.  This has possible
+# security implications for the server.
+$externalPrograms{convert}  ="/usr/bin/convert";
+
 ####################################################
 # NetPBM - basic image manipulation utilities
 # Most sites only need to configure $netpbm_prefix.
@@ -128,7 +133,7 @@ $externalPrograms{pngtopnm} = "$netpbm_prefix/pngtopnm";
 ####################################################
 # set timeout time (-t 40 sec) to be less than timeout for problem (usually 60 seconds)
 $externalPrograms{checkurl}     = "/usr/bin/lwp-request -d -t 40 -mHEAD "; # or "/usr/local/bin/w3c -head "
-$externalPrograms{curlCommand}  = "/usr/bin/curl";
+$externalPrograms{curl}  = "/usr/bin/curl";
 
 ####################################################
 # image conversions utiltiies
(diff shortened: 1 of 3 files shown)


conf/site.conf.dist
```diff
@@ -128,7 +133,7 @@ $externalPrograms{pngtopnm} = "$netpbm_prefix/pngtopnm";
 ####################################################
 # set timeout time (-t 40 sec) to be less than timeout for problem (usually 60 seconds)
 $externalPrograms{checkurl}     = "/usr/bin/lwp-request -d -t 40 -mHEAD "; # or "/usr/local/bin/w3c -head "
-$externalPrograms{curlCommand}  = "/usr/bin/curl";
```
mgage: the curlCommand is used in a number of places

mgage: `/Volumes/WW_test/opt/webwork/webwork2_2018/lib/WeBWorK/PG.pm:211: 	$envir{externalCurlCommand}  = $ce->{externalPrograms}->{curlCommand};`
and 
`/Volumes/WW_test/opt/webwork/pg_2018/lib/PGcore.pm:769: 	$options->{curlCommand} = WeBWorK::PG::IO::curlCommand();  #FIXME just changed from alternate source`

Also curlCommand occurs in several places in PG/IO.pm where it looks like the code is in the midst of a change over from curlCommand to curl.   Using curlCommand within IO.pm and askSage() might
be ok -- it makes some of the commands to askSage clearer. I could go either way. The other places, outside of those subroutines it seems to me that using {curl} as you have done is clearer.   I no longer remember my exact thinking when I was last working on this code.  What do you think? should we change all occurrences of curlCommand to curl?

drgrice1: I could put it back if you think that would be safer. I was just trying to complete what was done in you tikz pull request. It does seem the naming was inconsistent relative to the other external program variables. I had thought I changed all instances of {curl command} to {curl}.  Perhaps I missed some.

mgage: On closer inspection I think you only missed one.   The line 
211 in PG.pm .  This defines a variable in the PG environment called
externalCurlCommand.  I don't see where that is ever used. 

I think I'm in favor of commenting out line 211 in PG.pm (see whether that ever causes problems).  

I've decided I'm ok with continuing to use curlCommand inside askSage and IO.pm since I believe my idea there was that we might want to add additional switches to the bare curl command and therefore created a front end subroutine to make it easier to modify the bare curl command if we needed to. I'm willing to change my mind on that if you think that is over complicating things. 

I _was_ trying to make the assignments in site.conf more uniform -- and assigning {curlCommand} stuck out -- so you were following my initial intentions there.

drgrice1: I see.  I did miss line 211 of PG.pm.  We can comment that out, or just adjust it to be {curl}, or put everything back the way it was.  That isn't part of the tikz image work, and I don't want it to interfere with getting that in.  Just let me know what you prefer.

mgage:
Let's adjust it to be curl (for compatibility) but then comment it out.  I
don't see why we should expose curl directly as a command within a PG
problem or .pl macro.   It seems like it could be asking for security
trouble and I can't imagine a common use case.  I could be wrong in which
case it will be easy enough to uncomment  the line.

taniwallach:
A comment about the need to make the change to `/etc/ImageMagick-6/policy.xml` to use the feature added (to allow use of `/usr/bin/convert` to process PDF files) should be include in the install instructions and release notes.

Issue #1031, opened 2019-10-02
Answers that exceed 127 characters can not be saved to the database


drgrice1:
This is discussed in http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4662#p13763.
This can happen with the latex form of a student answer saved in the sticky answers for MathQuill answer boxes.  This also happens for essay type answers.  This did not happen with lengthy answers with WeBWorK 2.14, and is perhaps related to the database switch to utf8mb4.

mgage:
Thanks very much Glenn.  We'll need to track this down.

taniwallach:
Based on the discussion in the forum - it seems that the underlying problem is that `Storable::nfreeze` adds a byte count as a "byte" and when that is over (2^7)-1 it often is not a valid UTF-8 value, so the database rejects the string. 

This is handled done in `lib/WeBWorK/Utils.pm` by `decodeAnswers()` and `decodeAnswers()`.

Instead of fixing this by changing the database field to a BLOB, I would like to propose that we look into using JSON to serialize the data for this database field, but try to do so in a backwards compatible way by having `decodeAnswers()` in `lib/WeBWorK/Utils.pm` decide whether to use the "old" (thaw) approach or the JSON approach based on some regexp check of the contents of the field. All new data put into this field would be serialized/deserialized using standard Perl JSON tools, and only old style `nfreeze` data would be deserialized using the old code.

What do people think of this idea?

taniwallach:
A related issue, which I have not have yet had a chance to investigate is that the current code to store data in the `last_answer` database field does not seem to properly handle any answers which use UTF-8 characters which are not in Latin-1. I tend to suspect that the problem is somehow related either to `Storable::nfreeze` not working well with real Unicode data, or an issue with the character count. In any case, I suspect that using JSON to store the serialized form is likely to work and overcome that issue also.

drgrice1:
> Instead of fixing this by changing the database field to a BLOB, I would like to propose that we look into using JSON to serialize the data for this database field, but try to do so in a backwards compatible way by having `decodeAnswers()` in `lib/WeBWorK/Utils.pm` decide whether to use the "old" (thaw) approach or the JSON approach based on some regexp check of the contents of the field. All new data put into this field would be serialized/deserialized using standard Perl JSON tools, and only old style `nfreeze` data would be deserialized using the old code.

[...]

PR #1033, opened 2019-10-03
Change the serialization/deserialization code for the last_answer database field

taniwallach:
Motivation is to address the issue discussed at: 
  - https://github.com/openwebwork/webwork2/issues/1031
  - http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4662#p13763

Change the serialization/deserialization code of `encodeAnswers()` and `decodeAnswers()` of `lib/WeBWorK/Utils.pm` which are used to encode/decode the last_answer field in the COURSEID_problem_user table to use JSON instead of `Storable::nfreeze()` and `Storable::thaw()` which failed to properly handle UTF-8 data in answers, or many answers with over 127 characters (due to the binary length code appearing to be an invalid UTF-8 character).

The `decodeAnswers()` falls back to the **old** `Storable::thaw()` based code for backwards compatibility when the value returned from the database does not seem to be a JSON array (currently detected by checking for a leading `[` and a trailing `]`) while the `Storable::nfreeze()` data seems to always start with several control characters.

The new code has been tested with homework problems, and stores long answer strings, answer strings with UTF-8 characters, and prior last_answer data from the old format seems to load properly in all cases tested.

Testing was **not** done with gateway problems.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -42,7 +42,7 @@ use Storable;
 use Carp;
 #use Mail::Sender;
 use Storable qw(nfreeze thaw);
-
+use JSON;
 
 use open IO => ':encoding(UTF-8)';
 
@@ -914,7 +914,7 @@ our $BASE64_ENCODED = 'base64_encoded:';
 #  statements
 
 
-sub decodeAnswers($) {
+sub OLDdecodeAnswers($) {
 	my $serialized = shift;
 	return unless defined $serialized and $serialized;
 	my $array_ref = eval{ Storable::thaw($serialized) };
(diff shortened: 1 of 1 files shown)


taniwallach:
@drdrew42  - It would be good if someone tested the change with MathQuill enabled. Could you possible test it?

drgrice1:
I have tested this with MathQuill enabled.  In particular I tested this with the problem that Sean posted on the forums that led us to this issue.  It works fine.

mgage:
it worked for me in a fresh download of webwork-2.15 in a docker container

sean-fitzpatrick:
Thanks everyone for your prompt attention on this. I will go update my local WeBWorK install right now, and then re-enable MathQuill. If anything doesn't work, releasing 280 calculus students on it should find it. Do I need to upgrade databases after doing a git pull?

mgage:
I think you can get by without doing a database update.  It's been pointed out that with the change to webwork-2.15 the upgrade course procedure does not completely update things.  It adds new fields and tables as necessary but if the size of a field has changed it doesn't notice.  Potentially this could cause problems but so far it's not clear if it will.   That should be a project for the next release however --- to review how the database is managed from the admin webpage and anticipate any problems.

PR #1034, opened 2019-10-06
Patch for PR 1018

taniwallach:
Fix a small bug in code from PR https://github.com/openwebwork/webwork2/pull/1018 , which prevented course/system wide changes from working if not problem specific css file load was in use.


taniwallach:
I noticed that the main if conditional was only check for per problem special CSS load requests, and not for system-wide/course-wide requests, so if no problem specific request was made the system-wide/course-wide requests would not be processed. This fixed the problem.

I'm merging immediately, so the bug is fixed before WW 2.15 gets released.

mgage:
Thanks.

PR #1037, opened 2019-10-11
Fix feedback warnings with hidden answer blanks

drgrice1:
This is discussed on http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4668.

If a student attempts to email the instructor/ta via the feedback button and the problem has parts that are hidden and unanswered (like the latter parts of a scaffolding problem), then the value for that key in the database is null and that causes an error to be displayed for the student.  This is an easy fix.

If it is desired that all answer labels are shown in the email even if there is not answer, this could be slightly modified to achieve that.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -503,7 +503,7 @@ sub format_userproblem {
 	if (%last_answer) {
 		$result .= "Last answer:\n";
 		foreach my $key (sort keys %last_answer) {
-			$result .= "\t$key: $last_answer{$key}\n";
+			$result .= "\t$key: $last_answer{$key}\n" if $last_answer{$key};
 		}
 	} else {
 		$result .= "Last answer:                  none\n";


mgage:
This is ok with me. (the alternative is to make undefined answers blank using \\).  The change is small so it's safe to add it to master. 

The behavior enabled by this fix should suffice until someone finds a need for an alternative.  Glenn, could you also issue a pull request from your fix-feedback-warnings branch to develop and we'll pull both hot fixes at once.   Thanks.

drgrice1:
> Glenn, could you also issue a pull request from your fix-feedback-warnings branch to develop and we'll pull both hot fixes at once. Thanks.

I will set it up.  It looks like I will have to rebase off develop after merging this, as it wants to pull in the version changes from mater.

drgrice1:
Why is the merge not appearing in master?  Did I do something wrong?

sean-fitzpatrick:
Seems it's in master as commit 6df4473
from October 11? A bit odd, but it's there...

PR #1040, opened 2019-10-19
typo in array name

Alex-Jordan:
I think this typo is harmless since it is repeated the only time the array is used. But it's distracting when reviewing this file.


mgage:
I've verified that this is the only place it is used.

PR #1041, opened 2019-10-20
Gateway resultstable update

drgrice1:
It is time that gateway quizzes got a bit of a facelift.  It looks like AttemptsTable.pm was created in 2015, and it usage implemented in Problem.pm at that time.  However, it was not implemented in GatewayQuiz.pm.  It works there and so it should be used there also.  That table is much nicer, and aligns homework gateway quiz experience.  This pull request does that.


mgage:
This looks good. Much improved.

PR #1042, opened 2019-10-20
Problem randomization clear answers

drgrice1:
When a new version of a problem is presented with problem randomization enabled, clear out the answers that were submitted for the previous problem.  This could be made an option if the old behaviour is desired, but it seems a bit silly to me to show the answers to a previous version.  

Showing old answers can cause some undesirable effects particularly when scaffolding is used in a problem.  If there are parts that have radio answers, pop up boxes, or generally few possible answers, then the odds are pretty high that the last problem's answer is correct for the new version.  Scaffolding then marks the part green to indicate it is already correct.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Problem.pm
+++ b/lib/WeBWorK/ContentGenerator/Problem.pm
@@ -708,10 +708,15 @@ sub pre_header_initialize {
 
 	##### sticky answers #####
 
-	if (not ($submitAnswers or $previewAnswers or $checkAnswers) and $will{showOldAnswers}) {
-		# do this only if new answers are NOT being submitted
+	if (!($submitAnswers || $previewAnswers || $checkAnswers) && $will{showOldAnswers}) {
 		my %oldAnswers = decodeAnswers($problem->last_answer);
-		$formFields->{$_} = $oldAnswers{$_} foreach keys %oldAnswers;
+		# Do this only if new answers are NOT being submitted
+		if ($prEnabled && !$problem->{prCount}) {
+			# Clear answers if this is a new problem version
+			delete $formFields->{$_} foreach keys %oldAnswers;
+		} else {
+			$formFields->{$_} = $oldAnswers{$_} foreach keys %oldAnswers;
+		}
 	}
 
 	##### translation #####


mgage:
I want to check something first.  Your patch clears out the last answer.  However I can't get it to respect the number of times to allow a wrong answer before presenting the "Request new version" button. It seems to present that immediately after a wrong answer even if the number of wrong answers was set to 5 or 25.  This happens before your patch is applied as well.  Is this your experience as well?

drgrice1:
No.  In my experience the "Request new version" button does not appear until after the number of attempts has expired.

drgrice1:
You might check to see if the settings for the particular homework set are overriding the course configuration settings for problem randomization.  This is just a guess.

drgrice1:
Another thing to watch for is that if you have already attempted a homework problem, then you will have to change the individual homework settings for that problem for the setting to take effect.

mgage:
Aah.  I think this is it. I did change the global settings for the homework set but not the individual ones.

On Dec 30, 2019, at 4:01 PM, Glenn Rice <notifications@github.com<mailto:notifications@github.com>> wrote:


Another thing to watch for is that if you have already attempted a homework problem, then you will have to change the individual homework settings for that problem for the setting to take effect.

mgage:
OK.  I got everything to work.  I don't use this feature so it took a while to work my way through the options.  It can be more complicated when you are testing it than when you are using it since there are hysteresis effects.

mgage:
By the way does anyone use the other version of problemRandomize?  Should it be looked at again?

drgrice1:
I assume you mean the macro version.  There are a few problems in the library that seem to use it.

mgage:
There are quite a few actually.  At first I was starting to use those to test this pull request and then realized that I had the wrong version. I remember the macro version was around first and I don't mind having two versions around, particularly if they have different strengths, but it would be worth checking to see if the macro version is now unnecessary -- not right now perhaps but eventually.

drgrice1:
Yeah.  Once my file search completed it came up with more than 250 files that load the macro in the library.  We will probably need to keep the macro version around for a while for backwards compatibility.

dlglin:
At one point the OPL editorial board removed the macro version from a large number of contributed problems.  The argument against the macro version is that it forces anyone using that problem to allow rerandomization, whereas the WW-based rerandomization can be toggled for any problem.

I suggest checking with John Jones, but I think the better solution is to convert the OPL problems that use the macro.  That being said, how many of those problems actually use rerandomization vs. just loading the macro?

On Dec 30, 2019, at 2:40 PM, Glenn Rice <notifications@github.com<mailto:notifications@github.com>> wrote:


Yeah. Once my file search completed it came up with more than 250 files that load the macro in the library. We will probably need to keep the macro version around for a while for backwards compatibility.

drgrice1:
I agree that the macro version has the limitation of being forced onto anyone that uses it, and the WW-based re-randomization is the better way to go.

I don't know how many problems in the OPL actually use the macro.  I looked at a few, and saw that they loaded the macro, but didn't actually use the macro.

PR #1043, opened 2019-10-20
Problem randomization: Show correct answers after the last attempt before a new version must be requested.

drgrice1:
This is a new option to use when problem randomization is enabled.  If the option is enabled, the students will be shown the correct answer for a problem version after the last attempt and before a new version is shown.

There are most likely those that will not want this feature, so this is a course option.


mgage:
This works nicely.

PR #1044, opened 2019-10-21
Correct way to access ENV variable

SumitLubal:


mgage:
Thanks Sumeet.  I've resubmitted this patch to the develop branch #1056 and I'm closing this PR for now. 
I can't easily check shibboleth connections but your changes look right -- and I don't really see how the original code worked.  If someone who is using shibboleth tells me that your patch is urgently needed I'll reopen this pull request

PR #1046, opened 2019-11-02
make achievement items have effect on reduced scoring date

Alex-Jordan:
Currently if I have a "full credit" date (aka "reduced credit" date) in the middle of the term, with the "close" date at the end of the term, then using any item that extends a due date just pushes back the "close" date. The student feels cheated that the "full credit" date has not also been moved. These edits change that.

### DIFF
--- a/lib/WeBWorK/AchievementItems.pm
+++ b/lib/WeBWorK/AchievementItems.pm
@@ -155,6 +155,7 @@ sub use_item {
     # Set a new close date and answer time for the student and remove the item
     $set->due_date(time()+86400);
     $set->answer_date(time()+86400);
+    $set->reduced_scoring_date(time()+86400);
 
     $db->putUserSet($set);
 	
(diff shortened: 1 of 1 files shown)


drgrice1:
It is not really the proper github flow of things to put your pull request branches directly into the main github repository for the project.  That is what your individual fork is for.  The webwork github repository is getting cluttered.  We now have Alex-Jordan-patch-1, goehly-uploaded-message, and this branch that I don't think belong there.

Alex-Jordan:
Sorry, I thought I was editing in my own GitHub fork. I find it easy to get mixed up, and GitHub has moved me to a different fork (probably because of my own user error.)

drgrice1:
I see.  It isn't a big deal.  It just adds clutter.

drdrew42:
I am not able to test anything for a bit, but is there any possible conflict here for courses that do not have reduced scoring turned on? 

Courses with reduced scoring _disabled_ will have NULL values for `reduced_scoring_date`. What will perl think of adding NULL + 86400?

Alex-Jordan:
I added conditionals to check if `$set->reduced_scoring_date()` is already defined before trying to add time to it.

mgage:
I have trouble testing this because I don't immediately know how to artificially trigger an achievement.  However the code looks sound and any difficulties that are caused will be cosmetic and fairly easy to track down.

Alex-Jordan:
Thanks Mike, and I'll delete the branch this was on. I'll be more careful to create these GitHub patch branches within my own fork.

PR #1050, opened 2019-11-26
Fix trying library problems when problem randomization is enabled.

drgrice1:
I discovered that an error that occurs if problem randomization is enabled when attempting to view a problem that does not exist in the course (e.g. from the library).  This fixes that problem.

I introduced this bug when I restructured problem randomization.


mgage:
Did you mean for this PR to go to master? Is/will a similar PR be needed for develop?

mgage:
This works nicely.

drgrice1:
I did intend for this to go into master.  It is a rather visible bug for those that use problem randomization.

drgrice1:
Should I recreate this pull request to develop?

PR #1057, opened 2019-12-22
correct the links to CAPA and to Contrib in the modelCourses/template…

mgage:
… directory

### DIFF
--- a/courses.dist/modelCourse/templates/CAPA
+++ b/courses.dist/modelCourse/templates/CAPA
@@ -1,1 +1,0 @@
-Library/Contrib/CAPA
--- a/courses.dist/modelCourse/templates/Contrib
+++ b/courses.dist/modelCourse/templates/Contrib
@@ -1,1 +1,0 @@
-Library/Contrib


mgage:
Changed link pointers.

drgrice1:
It looks to me like this just deletes the links.  Did you mean to do that, or to change the links to point to the correct location?

mgage:
I meant to have this point links to the correct location.  But for some reason this isn't be 
transmitted.

ln -s /opt/webwork/webwork-open-problem-library/Contrib  Contrib
ln -s /opt/webwork/webwork-open-problem-library/CAPA  CAPA

but then these appear to get erased when I commit.   ???? 

It could be that we should make a more permanent and flexible fix in any case.

Perhaps define in site.conf 
webwork_toplevel  to be /opt/webwork
and then Contrib is webwork_toplevel/webwork-open-problem-library/Contrib
and CAPA is webwork_toplevel/webwork-open-problem-library/CAPA 

is that overkill?

drgrice1:
To add the links you will need to use the "-f" flag.  For example, "git add -f courses.dist/modelCourse/templates/CAPA".  This is due to the fact the the courses.dist directory is in the .gitignore file.

You might also note that the CAPA directory is a subdirectory of the Contrib directory.  So you should adjust that link appropriately.

I am not sure what we want the directory structure structure to look like.  Isn't the current default set up with "/opt/webwork/libraries/webwork-open-problem-library"?  Are you saying to drop the extra layer of the "libraries" directory?

If we define these locations in site.conf, then would they also need to be in localOverrides.conf if a different location is desired?  Also, are you suggesting some approach that removes the need for the links?

taniwallach:
Maybe relative links could be used to avoid the assumption that the path starts `/opt/webwork`? Just "../" enough times to get down from `courses/NAME/templates` to the "main" level and then follow the standard path to the expected location of CAPA and Contrib.

There would still be an assumption about the expected layout under the "main" level, but that seems unavoidable, and I assume it would be overrideable via `localOverrides.conf` and/or by local modifications of the symbolic links.

drgrice1:
Nathan's suggestion sounds like a good way to go.  Choose what the default directory structure should look like, and use relative links.  The update the $problemLibrary{root} and $contribLibrary{root} variables in localOverrides.conf.dist appropriately.

drgrice1:
By the way, is there a reason that the courses.dist directory is in the .gitignore file?  It doesn't seem like that should be the case.

mgage:
I'm pretty sure relative links won't work.  We disabled being able to reach
outside the template directory for security reasons.  Can't use ..

That's also the reason for using explicit links in the library.

I left out the libraries segment by accident. I've been away from this for
amonth and I'm already forgetting things :-(.   Time to eat now but I'll
rework the pull request.  I'm trying to fix the problem that Arnie noticed
that Library/Contrib just wouldn't work.



sent from my phone, Mike

mgage:
I agree I think.  The location of my entire library is non standard and
perhaps I hid courses.conf.dist to avoid uploading configs.   I think it's
better if it is not in gitignore





sent from my phone, Mike

drgrice1:
In my tests the relative links seem to work fine.  We disable leaving the template directory inside webwork, but webwork doesn't care how the links are on the system.  As long as they can be followed by the operating system, they should work.

drgrice1:
Note that since the CAPA directory is a subdirectory of the Contrib directory, I think that the CAPA link should simply point to the Contrib link (create with "ln -s Contrib/CAPA CAPA").  Then the CAPA link will work as long as the Contrib link does.

mgage:
Sounds good


sent from my phone, Mike

mgage:
The new attempt.  #1058 attempts to fix some of the problems with this pull request (which I merged accidentally) Fortunately not much was changed and corrections will be applied with #1058

PR #1058, opened 2019-12-23
Hotfix correct model course initial values

mgage:
* Change ASimpleCombinedHeaderFile.pg
	* Add PGML.pl 
	* loadMacros(
        "PG.pl",
        "PGbasicmacros.pl",
        "PGML.pl",
        \# "source.pl",
        "PGcourse.pl",
	);
	* Also change `BEGIN_TEXT/END_TEXT` to `BEGIN_PGML/END_PGML`
* Move CAPA_GRAPHICS
	* Moved to Contrib/CAPA/graphics/CAPA_Graphics
		* This hides the individual graphics so that they don't appear in the LibraryBrowser listing of CAPA problems.
* Updated paths to CAPA problems in default.config to correspond to default values in localOverrides
* Added link to CAPA_Graphics in webwork2/htdocs
#### This fix started with an observation of Arnie Pizer:

Somebody tried to set up links to the Contrib and CAPA libraries but messed up.
In the templates directory of the model course we have:
```
lrwxrwxrwx 1 wwadmin wwdata   20 Nov 30 16:26 CAPA -> Library/Contrib/CAPA
lrwxrwxrwx 1 wwadmin wwdata   15 Nov 30 16:26 Contrib -> Library/Contrib
```
but when a course is set up we have
```
lrwxrwxrwx  1 www-data wwdata   70 Dec  9 16:45 Library -> /opt/webwork/libraries/webwork-open-problem-library/OpenProblemLibrary/
```
as it should be.

But Contrib is under  webwork-open-problem-library/   NOT UNDER   OpenProblemLibrary/

#### To test

From CourseAdmin 

1. Add a new course.
2. Using LibraryBrowser inspect that CAPA and Contrib buttons are visible
3. Check that these button each lead to lists of problems in the Contrib and Contrib/CAPA directories
4. Check that in the CAPA directory the contents of the graphics subdirectory (Gtype... ) are not listed.
5. create a capa_problems homework in the LibraryBrowser
6. From the CAPA directory (type05)list add some problems that include graphics
7. Make sure that the graphics appear in 
	1. the library browser
	2. in the homework editor for 	capa_problems	 when "render all" button is pushed
	3. look at the capa_problems homework set directly and make sure the graphics appear in each of the pg problems.

### DIFF
--- a/conf/snippets/ASimpleCombinedHeaderFile.pg
+++ b/conf/snippets/ASimpleCombinedHeaderFile.pg
@@ -7,6 +7,8 @@ DOCUMENT();
 loadMacros(
 	"PG.pl",
 	"PGbasicmacros.pl",
+	"PGML.pl",
+	#"source.pl", #uncommenting this will display "show source" button if that is permitted
 	"PGcourse.pl",
 );
 
@@ -78,9 +80,9 @@ EOT
 #
 ####################################################
 
-BEGIN_TEXT
+BEGIN_PGML
 
-END_TEXT
+END_PGML
 
 
 TEXT($END_ONE_COLUMN);
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 applets
 tmp
 logs
-courses.dist
 library-directory-tree.json
 library-subject-tree.json
 textbook-tree.json
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1174,13 +1174,15 @@ $pg{specialPGEnvironmentVars}{use_javascript_for_live3d} = 1;
  $pg{specialPGEnvironmentVars}{DragMath} = 0;
  $pg{specialPGEnvironmentVars}{CAPA_Tools}             = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_Tools/",
  $pg{specialPGEnvironmentVars}{CAPA_MCTools}           = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_MCTools/",
- $pg{specialPGEnvironmentVars}{CAPA_GraphicsDirectory} = "$courseDirs{templates}/Contrib/CAPA/CAPA_Graphics/",
+ $pg{specialPGEnvironmentVars}{CAPA_GraphicsDirectory} = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_Graphics/",
+ $pg{specialPGEnvironmentVars}{CAPA_Graphics_URL}      = "$webworkURLs{htdocs}/CAPA_Graphics/",
+
  push @{$pg{directories}{macrosPath}},  
    "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_Tools",
    "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_MCTools";
 
  # The link Contrib in the course templates directory should point to ../webwork-open-problem-library/Contrib 
-
+ # The link webwork2/htdocs/CAPA_Graphics should point to ../webwork-open-problem-library/Contrib/CAPA/macros/CAPA_graphics
 # Size in pixels of dynamically-generated images, i.e. graphs.
 $pg{specialPGEnvironmentVars}{onTheFlyImageSize}      = 400,
  
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1167,10 +1167,19 @@ $pg{specialPGEnvironmentVars}{use_javascript_for_live3d} = 1;
 
 # Locations of CAPA resources. (Only necessary if you need to use converted CAPA
 # problems.)
-$pg{specialPGEnvironmentVars}{CAPA_Tools}             = "$courseDirs{templates}/capaLibrary/macros/CAPA_Tools/",
-$pg{specialPGEnvironmentVars}{CAPA_MCTools}           = "$courseDirs{templates}/capaLibrary/macros/CAPA_MCTools/",
-$pg{specialPGEnvironmentVars}{CAPA_GraphicsDirectory} = "$webworkDirs{htdocs}/CAPA_Graphics/",
-$pg{specialPGEnvironmentVars}{CAPA_Graphics_URL}      = "$webworkURLs{htdocs}/CAPA_Graphics/",
+################################################################################
+# "Special" PG environment variables. (Stuff that doesn't fit in anywhere else.)
+################################################################################
+
+ $pg{specialPGEnvironmentVars}{DragMath} = 0;
+ $pg{specialPGEnvironmentVars}{CAPA_Tools}             = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_Tools/",
+ $pg{specialPGEnvironmentVars}{CAPA_MCTools}           = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_MCTools/",
+ $pg{specialPGEnvironmentVars}{CAPA_GraphicsDirectory} = "$courseDirs{templates}/Contrib/CAPA/CAPA_Graphics/",
+ push @{$pg{directories}{macrosPath}},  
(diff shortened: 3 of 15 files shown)


mgage:
one other change:    courses.dir directory was removed from .gitignore

drgrice1:
This looks good.  I like that you also added the Library link.

mgage:
@drgrice1 thanks.  and thanks for your help and Tani's help in getting this right. 
 
You were right that using ../ syntax in the links works just fine.  I was remembering that we disabled that construction in all (I hope or at least almost all) paths that can be set anywhere from the web. This includes the conf files since course.conf can override any of those configurations.

I hope that using these alias links won't be too fragile.  I haven't yet tried them on a fresh download and setup within Docker.

conf/snippets/ASimpleCombinedHeaderFile.pg
```diff
@@ -78,9 +80,9 @@ EOT
 #
 ####################################################
 
-BEGIN_TEXT
+BEGIN_PGML
 
-END_TEXT
+END_PGML
```
drgrice1: Switching to PGML here seems to cause problems with hardcopy generation for problem sets that use this header file.  The problem is that this causes the \pgmlSetup command to be used early in the tex file before it is defined.

mgage: Good catch.  I suspect that adding beginproblem() to header file will fix the hardcopy generation problem. It isn't completely natural since beginproblem() also outputs the
number of points a problem is worth.  It's time to refactor beginproblem().  It's a hold over from WW1.9 when it was the only place we could initialize problems.  It's performing too many functions.

mgage: it appears to be complicated. it would be worth asking @dpvc for advice.

This works fine:
```
loadMacros(
	"PG.pl",
	"PGbasicmacros.pl",
    "PGML.pl",
    "unionLists.pl",
	"PGcourse.pl",
);
#TEXT($BEGIN_ONE_COLUMN);
TEXT(beginproblem());
#
# The item below printed out only when a hardcopy is made.
#
```
If I comment out `TEXT(beginproblem())`  it breaks.
If I uncomment `TEXT($BEGIN_ONE_COLUMN)` and the matching $END_ONE_COLUMN
at the end of the file it also breaks.  I think this is because braces are introduced
into the TeX code which undefines \pgmlSetup where it is needed.  I seem to recall that this was a problem with scaffolding sections as well. It was important to enter certain header items in exactly the right order.

drgrice1:
@Alex-Jordan: The issue with this in hardcopy is known about, and needs to be fixed before the next release.  See issue #1066.

taniwallach:
My 2 cents:

1. I do not see why each PG file needs to add the `$PGML::preamble` into the output file. It seems to me that most of what is in the preamble defined in `pg/macros/PGML.pl` should be permanently moved into to all the `hardcopyPreamble.tex` files, so it is all defined once at the very top of the LaTeX file. I suspect that such a change would allow `\pgmlSetup` to work in the header file.
    - If this does work, a comment about it needs to be in the release notes, so locally customized `hardcopyPreamble.tex`  files can be suitable fixed also.
2. If that does not work, then for the sample header files in `conf` I recommend changing back to  `BEGIN_TEXT` and `END_TEXT` and not putting in more effort.

drgrice1:
One relatively easy solution is to add `TEXT(MODES(%$problemPreamble));` on the line before `BEGIN_PGML` in ASimpleCombinedHeaderFile.pg.  Although @taniwallach is correct that it is redundant for each PG file to add the PGML preamble to the generated tex file.  Moving the PGML preamble into the hardcopyPreamble.tex files would only add these definitions once.  I am not sure that is the right way to go though.  Those definitions aren't needed if PGML is not used.  They shouldn't cause a problem in that case though.

I am not sure what the motivation for switching from `BEGIN_TEXT`/`END_TEXT` to `BEGIN_PGML`/`END_PGML` is.  Perhaps for now just reverting back to the `BEGIN_TEXT`/`END_TEXT` would be the easy way to go.

drgrice1:
It seems that what is really needed for many things is a way for a PG file to "really" only do something once.  The PGML.pl file tries to only add its preamble once by checking the preamble with
```$problemPreamble->{TeX} .= $PGML::preamble unless $problemPreamble->{TeX} =~ m/definitions for PGML/;```
This of course fails because each PG file has its own local copy of the $problemPreamble hash.

There are other places where this occurs also.  For example, parserRadioButtons.pl attempts to only add the "uncheck" javascript by using a local $jsPrinted variable.  That of course fails in the same way.

There is a way to add include a CSS file once (added in https://github.com/openwebwork/webwork2/pull/1018), and there will be a way to include a JS file once (assuming https://github.com/openwebwork/webwork2/pull/1121 and https://github.com/openwebwork/pg/pull/489 are merged).  These are situation specific hacks really.  I think there needs to be a better communication system between each translator instance and what is outside to properly fix this.

taniwallach:
@drgrice1 wrote:
> Those definitions aren't needed if PGML is not used. They shouldn't cause a problem in that case though.

Almost any latex file loads some packages, or at least the "core" of LaTeX, and such things already include a significant amount of macro definitions which will be processed on LaTeX startup. Adding some small number of additional lines will not significantly change the processing time when they are not needed, and is very unlikely to have any adverse effect on assignments which do not use PGML.

> This of course fails because each PG file has its own local copy of the $problemPreamble hash.

It might apply when recursive loading of problems is done using `includePGproblem()` but such approaches from inside the "renderer" process/name-space will certainly not coordinate between problems parsed in different calls to the parser.

So, any solution will certainly need to depend on the higher-level processing which is making render request being able to "process" request for "single inclusion" similar to the special cases you mentioned.

>  I think there needs to be a better communication system between each translator instance and what is outside to properly fix this.

This is something which certainly merits careful consideration. However, it might be somewhat complicated to do, especially given that problems could in the future be rendered by calls to a "standalone" renderer and in potential parallel calls to different renderer processes could be run in parallel.

For the specific issue of LaTeX "boilerplate" - I really thing it belongs either in a header file whose contents are just part of the generated LaTeX file or a local style file imported by the header section of the generated LaTeX file.

drgrice1:
> This is something which certainly merits careful consideration. However, it might be somewhat complicated to do, especially given that problems could in the future be rendered by calls to a "standalone" renderer and in potential parallel calls to different renderer processes could be run in parallel.

The point is that PG needs a general communication layer that any renderer, including the "standalone" render, would use.  If done correctly each PG process would return information as to what needs to be added, and the renderer would then handle ensuring it is added, and only once.

conf/snippets/ASimpleCombinedHeaderFile.pg
taniwallach: I don't think it is worth the effort to get PGML to work in the header file.
Being reverted by @drgrice1 to non-PGML by https://github.com/openwebwork/webwork2/pull/1178.

mgage:
Just for the record, my reason for wanting PGML enabled in header files is to allow the easy construction of lists.  The markup syntax is particularly convenient for creating lists which, for example, give a quick description of each problem in the set, something that I do frequently. Header files also tend to be more text oriented and the handling of paragraph breaks is particularly natural in the mark down language.  None of these are deal breakers, one can use $PAR in TEXT mode, and BeginList (from unionLists.pl) or even OL(@array) from PGbasicmacros.pl  but the syntax doesn't seem as clean to me. 

The more important underlying problem that this discussion highlights is a general framework for handling initialization problems, in this case for TeX, in a general situation where PG might appear in several environments: homework problem, gateway quiz, standalone, possibly within PreTeXt. Similar issues affect javaScript initialization.

taniwallach:
I created an issue to track the "smaller" issue. We should better define what we want in terms of back-end rendered to front-end system communications/coordination to address the deeper issues. I don't think we can solve this for WW 2.16, but can target the following release.

PR #1059, opened 2019-12-29
Show me another use MathQuill answer boxes if enabled

drgrice1:
Use MathQuill answer boxes on "Show me another" problems if the option to use MathQuill answer boxes is enabled. Currently the input is hidden and no MathQuill answer box is injected.


mgage:
Seems to work ok.

PR #1060, opened 2019-12-29
Autouncheck empty date field on the set detail pages

drgrice1:
See the discussion on https://github.com/openwebwork/webwork2/pull/988

Mike, does this fix the issue you described?

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail2.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail2.pm
@@ -689,8 +689,8 @@ sub FieldHTML {
 	# if we are creating override feilds we should add the js to automatically check the
 	# override box.
 	if ($forUsers && $check) {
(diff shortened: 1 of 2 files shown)


mgage:
Not quite.  Go to set detail 2 for individuals.
Enter dates for closes and answers available.
Delete the date for closes (but don't click anywhere else) -- you get "none specified" and the box is still clicked.
Now press "save changes" -- you will get an error message about a badly formed or missing date
If then reload the page everything will be fine.
Instead of "save changes" you can hit return and get the same behavior.
Thanks for debugging this.

I haven't checked the User detail list page but I expect it is the same.

drgrice1:
Are you sure that you restarted apache2 after checking out this pull request?  I just want to make sure, because for me if I do what you describe with this pull request it works as expected with no errors and the check box is unchecked.

mgage:
Cancel the previous comment.  I hadn't restarted apache2 and flushed all of the css.

mgage:
Everything looks good now.

PR #1063, opened 2019-12-31
Fix trying library problems when problem randomization is enabled

drgrice1:
I discovered that an error that occurs if problem randomization is enabled when attempting to view a problem that does not exist in the course (e.g. from the library). This fixes that problem.

I introduced this bug when I restructured problem randomization.

This is the recreation of pull request #1050 but targeting the develop branch.  Although, I think that this is a good candidate for a hot fix into master.


mgage:
OK.  I'll pull this to develop.  If you want to submit a similar request to master I'll pull it to that as well.  I'm sorry I messed up your first pull request.  I got confused between two pull requests I was working on .

drgrice1:
Sorry about that.  I guess I should specify the intent in the title for a hot fix, or make it more clear in some way.

PR #1065, opened 2019-12-31
add provisional fix to Caliper::Actor

mgage:


mgage:
@andrew-gardener   please check this patch over to see if it makes sense.

PR #1071, opened 2020-01-28
Hotfix to mqeditor.css

drdrew42:
The mqeditor.css hides all `<INPUT type='text'>` fields, but this is problematic for problems that use javascript with interactive text fields that are not submitted to WeBWorK for grading... (see Library/Rochester/setMAAtutorial/javascriptexample1.pg for one example)

This PR instead identifies the input fields to be hidden by their 'codeshard' class instead of hiding all text input fields.

Is anyone aware of other objects that might share the 'codeshard' class - causing conflict with this approach?

(This is submitted as a hot fix, as I have multiple 'oracle'-style problems in my current curriculum)

### DIFF
--- a/htdocs/js/apps/MathQuill/mqeditor.css
+++ b/htdocs/js/apps/MathQuill/mqeditor.css
@@ -31,7 +31,7 @@ span[id^=mq-answer]
     margin-left: 0
 }
 
-input[type=text]
+.codeshard
 {
     display: none !important;
 }
--- a/htdocs/js/apps/MathQuill/mqeditor.css
+++ b/htdocs/js/apps/MathQuill/mqeditor.css
@@ -31,7 +31,7 @@ span[id^=mq-answer]
     margin-left: 0
 }
 
-.codeshard
+input[type=text].codeshard
 {
     display: none !important;
 }


htdocs/js/apps/MathQuill/mqeditor.css
```diff
@@ -31,7 +31,7 @@ span[id^=mq-answer]
     margin-left: 0
 }
 
-input[type=text]
+.codeshard
```
drgrice1: I would recommend that you combine these to make this even more precise.  So use
`input[type=text].codeshard`

drdrew42: I'm all for the extra safety here.

taniwallach:
I'll  revisit an old suggestion: Can we create a mechanism where a problem author can mark an entire question or a specific input field to be omitted from the MathQuill handling?

Even just the second idea with the use of CSS `:not()` should help solve the problem of an overly broad initial selector. (https://developer.mozilla.org/en-US/docs/Web/CSS/:not)

For the general "skip MQ" for this problem, I'm not sure what would be the simplest mechanism, maybe setting a CSS class or a JavaScript variable which when true would "turn off" the MathQuill processing for the current page (or better yet only for DOM descendants of where the class was set).

For the specific input fields exception - let's assume that we will soon be able (in PG) to add additional class names (and style attributes, etc.) to input fields via some sort of extension of the `ans_rule()` method, so we could have a skipMQ class for this purpose. (See https://github.com/openwebwork/pg/issues/444 ) I intend to work on implementing that within the next few months (and sooner would be better for me also).

Note: The `openDiv()` support already supports creating a DIV with a given class name. See: https://github.com/openwebwork/pg/commit/da0f007e17bd25d220328f3dba2f5f8d3a2585d7#diff-2c3e0f1c9ff2f13866441f52d449f855

drgrice1:
> I'll revisit an old suggestion: Can we create a mechanism where a problem author can mark an entire question or a specific input field to be omitted from the MathQuill handling?

This is easy to implement per input.  I have an implementation that modifies about 4 lines of code.  Basically you add a check to see if the mathQuillOpts flag is equal to 'disabled' in ProblemUtil.pm, and then don't add the hidden MaThQuIlL inputs and have javascript hide the original input instead of css. 

> Even just the second idea with the use of CSS `:not()` should help solve the problem of an overly broad initial selector. (https://developer.mozilla.org/en-US/docs/Web/CSS/:not)

This approach would be harder to implement at this point, as there isn't an easy way to add a class to the original input with the current implementation.  We have easy access to the associated hidden MaThQuIlL input, but not the original text input.  We would need to make the MathQuill implementation intrude further into the PG code to do that.

drgrice1:
Sure.  I have attached a patch that does this.  This patch applies cleanly to WW 2.15, and applies with minor offsets to the develop branch.

The first thing that it does is removes the css code that hides all text inputs with the class codeshard, and instead hides those text inputs in javascript.  Next a check is added in ProblemUtil.pm where the hidden MathQuill responses are inserted for "disabled" in the (largely undocumented) mathQuillOpts flag that you can already pass to the answer hash via a MathObject's cmp method, and in that case it doesn't add the hidden MathQuill response, in which case the javascript ignores that answer.

Of course you need admin access to the server to apply this patch.  It is created with `git diff`, and so applies at a patch level of 1.  (Apply with `patch -p1 < per-response-disable-mq.patch` run from the directory /opt/webwork/webwork2.)

[per-response-disable-mq.zip](https://github.com/openwebwork/webwork2/files/5893651/per-response-disable-mq.zip)

PR #1076, opened 2020-02-13
Add Caliper tool use profile

andrew-gardener:
- Also add missing profile property for log in/out events

Related to #970 and #971

### DIFF
--- a/lib/Caliper/Entity.pm
+++ b/lib/Caliper/Entity.pm
@@ -8,6 +8,7 @@ use WeBWorK::DB;
 use WeBWorK::Debug;
 use Data::Dumper;
 use WeBWorK::Utils::Tags;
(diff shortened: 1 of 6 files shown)


andrew-gardener:
Add a few minor improvements (just using this PR since its still open):

- session ids are now sha1 hashed (its important to report session consistently but it doesn't need to use the actual session id)
- try checking HTTP_X_FORWARDED_FOR, REMOTE_ADDR, and HTTP_CLIENT_IP for ip addresses (instead of just REMOTE_ADDR)
- async now has a timeout of 5 seconds and a max request time of 10 seconds (to help reduce potential Caliper endpoint timeout issues)

mgage:
I agree with @taniwallach 's review.  I don't yet have anything set up that uses Caliper so I can't test it directly yet.

PR #1081, opened 2020-02-25
Try to sync changes made to master into develop

taniwallach:
There are some 21 commits in master which are not in develop.
  - 8 files were modified, and almost all the changes are quite minor fixes.
  - Other than a few `ARG` settings in Dockerfile which should probably be edited for develop, all these small changes should be in develop IMHO.

**Warning:** GitHub reports that this cannot be automatically merged. We need to find and fix the merge conflict.


taniwallach:
The conflict is only in `Dockerfile`.

drgrice1:
I see what is going on here now.  You just merged master into develop.  That is why a diff from the origin/develop branch to this pull request shows everything from develop being reverted.  Of course that won't happen with a merge as it will only pull in the missing commits.  The thing is that most of those aren't actually missing.  They just have different git treeish hashes.  Most of the things in master are actually in the develop branch already.  The hot fixes to the master branch have corresponding pull requests to the develop branch.  The change to the README.md file is the one case where this was not done.

(The same comment applies to your pull request to the pg code, except there is nothing actually missing there like the README.md file here.)

taniwallach:
> I see what is going on here now. You just merged master into develop. That is why a diff from the origin/develop branch to this pull request shows everything from develop being reverted. Of course that won't happen with a merge as it will only pull in the missing commits. The thing is that most of those aren't actually missing. They just have different git treeish hashes. Most of the things in master are actually in the develop branch already. The hot fixes to the master branch have corresponding pull requests to the develop branch. The change to the README.md file is the one case where this was not done.
> 
> (The same comment applies to your pull request to the pg code, except there is nothing actually missing there like the README.md file here.)

Yes, I ran a "compare" of master to develop using GitHub... 

I looked at your comments about pg first, and reached the same conclusion. \

It seems that we should probably be more careful about how changes are merged into both develop and master. Maybe using a single source branch for pull requests to both master and develop would avoid this issue in the future.

Closing this.

Apparently we really just should get the `README.md` synced to develop, so it will not get messed up by a future merge from develop to master.

taniwallach:
I think I was bit by the same GitHub behavior as @Alex-Jordan was in https://github.com/openwebwork/pg/pull/436#issuecomment-560001360

drgrice1:
Yeah, the readme file should probably be synced.  Although a merge from develop to master wouldn't revert the changes in master.  Git is smart enough with merges.  That is also why this merge from master to develop doesn't actually revert the things in develop.

PR #1083, opened 2020-02-25
UTF-8 fix to Debug.pm

taniwallach:
`Debug.pm` open the log file using `>>:encoding(UTF-8)` to support Unicode characters in the log messages. If I recall correctly, before the fix, wide characters in log messages trigger warning messages.


taniwallach:
A slightly different change has been in production for the several months on my servers, but it makes more sense to open the log file in `UTF-8` mode than to `Encode` the `$finalMessage` as was done in the older change. The new change has been tested and works, and Unicode characters are being properly written to a UTF-8 encoded debug.log file.

PR #1085, opened 2020-02-25
Add percentage grades for each assignment to scoring csv file

taniwallach:
The current csv generation code provides assignment grades as point totals, but not as percentage grades (from 100). We found this inconvenient. 

The code adds additional columns to the generated csv files with the scores from 100. 

**Warning:**  The code continues to assume that the total number of points which can be earned on an assignment is constant, which is a convenient assumption but may not be true if a specific user's assignment was edited in a manner which changes the total.

**Note:** The "main" subroutine modified takes an argument which enables switching the new behavior on and off. However, I have not added code to add a checkbox and use that to set the switch on/off, and for now it is hard-wired as on. That is a potential future improvement.


mgage:
Looks ok to me.

PR #1086, opened 2020-02-25
Adds a new management utility script: bin/importClassList.pl 

taniwallach:
Adds a new management utility script `bin/importClassList.pl` to allow loading classlist file into courses from the command line. Useful to add standard staff members quickly to multiple courses.

### DIFF
--- a/bin/importClassList.pl
+++ b/bin/importClassList.pl
@@ -1,0 +1,155 @@
+#!/usr/bin/env perl
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2019 The WeBWorK Project, http://openwebwork.sf.net/
+# 
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+# 
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+BEGIN {
+	die "WEBWORK_ROOT not found in environment.\n"
+	  unless exists $ENV{WEBWORK_ROOT};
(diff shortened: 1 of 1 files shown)


glarose:
Looks like a reasonable addition to me.  I do a lot of this every semester, using my in house version of essentially this script.

mgage:
@glarose  is there any advantage to writing this as a wwsh script?  Would that allow productive reuse of some code or does that just make the use of this script more complicated?  Thoughts?

glarose:
@mikegage  I think it's a wash.  In this script it looks to me as if it
would save a little bit of code reuse, but not that much, and add a little
bit of overhead, but not that much.  Without spending too much time
thinking about it, it seems to me that there are three reasonable
directions one could go with this: 1. Make the decision that wwsh is the
standard, and convert this (and other scripts) to use that; this then
removes overhead by making it the way that we run all of these scripts.
2. Add this as is, and defer the discussion of wwsh vs. stand alone
scripts.  3. Figure that it is useful to have both tracks, and include
most of the admin scripts of this ilk in either form: importClassList.pl
and ww_importClassList.pl, for example.  In the long run the last is
probably not the best option, though it might provide some information
about what the best long-term policy/solution is.

On Wed, May 27, 2020 at 08:39:14AM -0700, Michael Gage wrote:
> @glarose  is there any advantage to writing this as a wwsh script?  Would that allow productive reuse of some code or does that just make the use of this script more complicated?  Thoughts? 
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/openwebwork/webwork2/pull/1086#issuecomment-634747882

-- 
-- 
P. Gavin LaRose (he/him/his) | glarose@umich.edu | 734.764.6454 | EH 3832
Karen Rhea Collegiate Lecturer & Instructional Technology Program Manager
University of Michigan Department of Mathematics

mgage:
@glarose Thanks Gavin for spelling out the choices.  I think I will go with number 2 -- differ decision until later -- mainly because I don't want to start a project of rewriting command line scripts to use wwsh.  My guess is that at some point some internals of the import scripts in the base code will change and we'll be faced with either updating two scripts (internal and command line version) or improving wwsh if necessary and making utility scripts of the import/export variety available easily available both internally and externally as command line scripts. I don't know what all the parameters of such a rewrite would be so I'll kick the can down the road.  :-)   

Thanks for the script @taniwallach .

PR #1087, opened 2020-02-25
Modifications to docker-config/docker-entrypoint.sh to improve startup time

taniwallach:
Modifications to docker-config/docker-entrypoint.sh to improve startup time, mainly intended for production servers.

Some developers may need to reenable some of the slower changes.

Also adds "a2enmod headers" which I have found useful, although in the longer term maybe that should be done in a revised Dockerfile.

### DIFF
--- a/docker-config/docker-entrypoint.sh
+++ b/docker-config/docker-entrypoint.sh
@@ -19,6 +19,9 @@ if [ $SSL -eq 1 ]; then
   a2enmod ssl && a2ensite default-ssl
 fi
 
+echo "Enabling Apache headers module"
+a2enmod headers
+
 # Build more locales
 if [ "$ADD_LOCALES" != "0" ]; then
   echo "Rebulding locales - adding: $ADD_LOCALES"
@@ -152,7 +155,7 @@ if [ "$1" = 'apache2' ]; then
         wait_for_db
         $WEBWORK_ROOT/bin/restore-OPL-tables.pl
         $WEBWORK_ROOT/bin/load-OPL-global-statistics.pl
-        $WEBWORK_ROOT/bin/update-OPL-statistics.pl
+        #$WEBWORK_ROOT/bin/update-OPL-statistics.pl
         if [ -d $APP_ROOT/libraries/webwork-open-problem-library/JSON-SAVED ]; then
           # Restore saved JSON files
           echo "Restoring JSON files from JSON-SAVED directory"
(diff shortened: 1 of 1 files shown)


mgage:
I'm getting mysql connect errors when I try to build a docker file from this:

Here are the errors:
```
db_1   | 2020-05-25 13:58:05 0 [Note] Added new Master_info '' to hash table
db_1   | 2020-05-25 13:58:05 0 [Note] mysqld: ready for connections.
db_1   | Version: '10.4.13-MariaDB-1:10.4.13+maria~bionic'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306  mariadb.org binary distribution
db_1   | 2020-05-25 13:58:05 0 [Note] InnoDB: Buffer pool(s) load completed at 200525 13:58:05
db_1   | 2020-05-25 13:58:05 8 [Warning] Aborted connection 8 to db: 'unconnected' user: 'unauthenticated' host: '172.19.0.4' (This connection closed normally without authentication)
app_1  | OPL path seems to be /opt/webwork/libraries/webwork-open-problem-library/
db_1   | 2020-05-25 13:58:06 9 [Warning] Access denied for user 'webworkWrite'@'172.19.0.4' (using password: YES)
app_1  | ERROR 1045 (28000): Access denied for user 'webworkWrite'@'172.19.0.4' (using password: YES)
app_1  | DBI connect('webwork:db:3306','webworkWrite',...) failed: Access denied for user 'webworkWrite'@'172.19.0.4' (using password: YES) at /opt/webwork/webwork2/bin/load-OPL-global-statistics.pl line 43.
db_1   | 2020-05-25 13:58:06 10 [Warning] Access denied for user 'webworkWrite'@'172.19.0.4' (using password: YES)
webwork2_app_1 exited with code 2
```
these are the last lines after running `docker-compose up`.

The procedure:
1. clone `https://github.com/taniwallach/webwork2`
2. checkout `modify_docker_entrypoint_for_faster_startup `  branch
3. modify .env to set `WEBWORK_DB_PASSWORD`
4. modify Dockerfile to set 
```
ARG WEBWORK2_GIT_URL=https://github.com/taniwallach/webwork2.git
ARG WEBWORK2_BRANCH=modify_docker_entrypoint_for_faster_startup

```
5. modify docker-compose.yml to set `MYSQL_ROOT_PASSWORD:`

One question when the mysql password is set they are for   `webworkWrite` @ `????`
what is ???? set to?

taniwallach:
You seem to want to start with a clean Docker environment. It would be easier to just mount this version of the file into an existing Docker image using a mount line in `docker-compose.yml` to override the old one.

I'm going to guess that you did not remove the old "mysql" volume, so that there is an existing database with some different password than what you set in `.env`.

Try `docker volume ls` and look for a volume called `webwork2_mysql`. It may need to be deleted with  `docker volume rm webwork2_mysql`.

mgage:
that was it.  I removed the mariadb:10.4 layer of code and all of the containers but I didn't 
remove the webwork2_mysql volume. 

Are there instructions for retrieving the root and webworkWrite passwords if you are using an old volume and you are uncertain about the password? 

Can we add optional code to reset the password for webworkWrite at least from docker-compose.yml?

I can see how to modify the entry point file in an existing docker container.  I'm less clear
about which parts of the docker-compose.yml can be modified after the docker build command is complete.  Some of the values seem to appear in both Dockerfile and in docker-compose.yml

It's been a while since I build a docker container so most of these comments relate to documentation for build docker containers from the current Dockerfile and docker-compose.yml files -- they have changed quite a bit since the last documentation on the webwork2 wiki (and that documentation was somewhat sketchy to begin with. )  Is there new 
documentation I'm overlooking?

mgage:
The admin course delivered with vanilla Dockerfile needs to be upgraded.  
`webwork2/bin/upgrade_admin_db.pl` adds achievement tables, etc. 
Is there a fresh version of the admin course that wouldn't need to be upgraded?

mgage:
The first docker box is working, after a few tweaks as described above.

A second docker box which was not set up completely from scratch is giving me errors:

```
error instantiating DB driver WeBWorK::DB::Driver::SQL for table achievement_user: DBI connect('webwork','webworkWrite',...) failed: Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2) at /opt/webwork/webwork2/lib/WeBWorK/DB/Driver/SQL.pm line 70. at /opt/webwork/webwork2/lib/WeBWorK.pm line 317.

```
I don't believe this is a password problem.  From inside the docker box I can connect to mysql
with `mysql -u webworkWrite -p -h db` without any trouble.

There is no directory /var/run/mysqld when viewed from inside the box. Where else would mysql have put this .sock file?

mgage:
more news:  changing this line in docker-compose.yml

# - ".:/opt/webwork/webwork2"
     
causes the problem.  When it is commented out the mysql connection is made.  When 
it is not commented out I get the error described above.

mgage:
Discovered the problem.  This is for future reference. 
In the bad docker box the webwork2/conf/site.conf file had been created after I had cloned the webwork2 repos.  I had even adjusted the site.conf file so that webwork2 worked.   This means however that the entry point script, seeing that a site.conf file already exists, skips the section in which entries in site.conf are modified according to information in the ENV variables set by .env and docker-compose.

mgage:
after all that I have both versions of this pull request working. 

The comments about updating the admin course and so forth should be handled in another PR.

PR #1088, opened 2020-02-25
Proposed change to main LTI grade passback code

taniwallach:
Adds a check with the LMS that the old grade and new grade differ by more that 0.001 before making a new LTI grade update call.

Reduces "no change" updates being made, which in some systems were also triggering mail updates. 

The cost is the need to query the LMS for the old grade before sending a new grade.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
@@ -32,6 +32,7 @@ use WeBWorK::Utils qw(grade_set grade_gateway grade_all_sets wwRound);
 use Net::OAuth;
 use HTTP::Request;
 use LWP::UserAgent;
+use UUID::Tiny  ':std';
 
 use Digest::SHA qw(sha1_base64);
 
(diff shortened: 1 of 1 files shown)


taniwallach:
This was in use on a server where a course of about 100 students were using "homework" (per-assignment) LTI grade pass-back for full semester with no problems.

The Apache logs did accumulate many messages 
```
Unable to read LMS grade. Error:  at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm line 304.
```
which seem to correspond to the first grade per problem (about 5000 messages, and close to 17000 answers submitted).

and 
```
Use of uninitialized value $message in string ne at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm line 303.
Use of uninitialized value $message in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm line 304.
```

Further work needed to avoid these error messages.

taniwallach:
Replaced by https://github.com/openwebwork/webwork2/pull/1177

PR #1090, opened 2020-02-25
This is the initial version of code to support html2xml using LTI

taniwallach:
This is the initial version of code to support `html2xml` using LTI authentication and grade-passback.

It has been used by several small test courses (with real students) on open-edX using LTI embedding of problems.

The LTI URL had the form: `https://webwork.technion.ac.il/webwork2/html2xml` and the following array for custom parameters was provided to edX (one sample case):
```
["sourceFilePath=Technion/prep/ComplexNumbers/AlgebraicForm/Addition1/q01/he/part01.pg", "courseid=justcomplex_daemon", "language=heb", "displayMode=MathJax", "outputformat=simple2", "resetSeedandPsvnFromUID=1", "answersSubmitted=0"]
```

It was used with inline embedding, and with open-edX set to receive a score.

---
`resetSeedandPsvnFromUID` is a new feature to set a seed and psvn based on the user ID, as LTI cannot provide different values for different user. The code to support this in in the PR.

**Note:** unlike prior use of `html2xml` to embed problems, no password is provided in the URL, just the LTI key+secret are set in the LMS so it can "sign" the initial LTI connection data which provides the authentication.

**More work is needed to improve this code.**

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -129,6 +129,19 @@ $preferred_source_of_username = "lis_person_contact_email_primary";
 
 $LTIBasicConsumerSecret = "";  #This must be set  
 
+################################################################################
+
+# List key + secret pairs for html2xml LTI support in the hash below.
+# This is the location from which lib/FormatRenderedProblem.pm
+# can get a LTI secret using a "seed_ce".
+#   A sample pair is provided.
+
+%LISConsumerKeyHash = (
+  "key_for_html2xml_LTI_course1" => "secret_for_html2xml_LTI_course1",
+);
+
+################################################################################
+
 # The purpose of the LTI nonces is to prevent man-in-the-middle attacks.
 # The NonceLifeTime (in seconds) must be short enought to prevent
 # at least casual man-in-the-middle attacks but
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -33,6 +33,8 @@ use WeBWorK::Utils qw( wwRound encode_utf8_base64 decode_utf8_base64);
 use XML::Simple qw(XMLout);
 use WeBWorK::Utils::DetermineProblemLangAndDirection;
 use Encode qw(encode_utf8 decode_utf8);
+use Digest::SHA qw(sha1_base64);
+use Date::Format;
 use JSON;
 
 our $UNIT_TESTS_ON  = 0; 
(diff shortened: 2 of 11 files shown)


taniwallach:
When using this code - you need the LTI key for your daemon course in 2 places:
1. In `course.conf` of the "daemon" course as the value of `$LTIBasicConsumerSecret` (as common for LTI for a course).
2. `webwork2/conf/authen_LTI.conf` where it should use a unique "key" name, which is used in the LMS.

The second settings is needed as there is not a full course environment available in `lib/FormatRenderedProblem.pm` but only a `seed_ce` so access to the value set in the `course.conf` is not possible from there.

Warning: The code will only work when the course name contains "daemon" as part of the course name.

taniwallach:
Warning: There are several hacks in place which were meant to be temporary, and unfinished things which I have not had time to work on yet.

In particular, I put in place a temporary hack to save just the submitted answers to the database (using the relevant user ID) when using LTI+html2xml. However, I did not complete the necessary work to create/track attempts used, or create records in the tables needed to make it possible to access the data via the WeBWorK web interface used by staff to a course. That mean that although the LMS (Moodle/edX) will save scores sent by the WeBWorK server, there is really no convenient manner even on the WeBWorK side to see the answers submitted except manual access to the database.

---

The first connection uses LTI to connect, but needs to then provide some "proof" to authenticate the XMLRPC call. At present the code authenticates the second stage (`lib/WebworkWebservice.pm`) of the initial call based on a value set for `internal_WW2_secret`. The code in the PR used the
not-so-secret value `TemporaryApproachSetSomeServerSecretHere` and until a better approach is in place - that should be replaced on a server using this code.

**The most critical issue to fix is to provide some sort of verification/digital signature that the data in the second stage is really what the LTI authentication verified. Comments about this are in the code.** 

Answer submission uses a session key, and that is linked to the data which must be maintained unchanged so that scores will be passed back for the correct user/problem/etc. The session key code is somewhat hacked so it works well enough for now, and there are many comments about what is being done, and where things may need improvement in the code.

taniwallach:
A test problem run via Moodle with this code used the URL `https://webwork2.technion.ac.il/webwork2/html2xml` and in the "custom parameters box` has the following listed:
```
courseID=LTI_daemon_course
language=heb
resetSeedandPsvnFromUID=1
displayMode=MathJax
outputformat=simple
sourceFilePath=Technion/LinAlg/Complex/he/AlgebraicCalc/z_squared.pg
answersSubmitted=0
```

mgage:
@taniwallach I'm starting to read through this code.  Is there a later version that I should be aware of?

taniwallach:
> @taniwallach I'm starting to read through this code. Is there a later version that I should be aware of?

No. It worked well enough for the urgent trial we ran of WW problems embedded via LTI+html2xml in an open-edX course, but I have not gotten back to it.

taniwallach:
Needs to be reconsidered.

PR #1056, opened 2019-12-22
Correct way to access ENV variable

mgage:
This duplicates the original PR to master #1044.  The changes make sense -- it seems unlikely that the original code would have worked well. However since I can't test shibboleth directly I'm submitting this pull request to develop and holding the hotfix to master for now. I don't want to risk breaking code that is actively being used.

PR #1091, opened 2020-02-25
Shibboleth: Reduce nesting and remove duplication

jrchamp:
I was stepping through this code and I found it a bit confusing. To help myself understand:

- I treated all `return` as leaving the code and removed the `else` nesting that just mean the subroutine continued processing. 
- I moved code that could only be reached when a statement was true (i.e. not false) to the block that checked that the statement was true.
- I noticed 3 variables were being assigned twice and removed the copies that were being overwritten 100% of the time.
- I removed a duplicate document comment that did not reflect the code that it was commenting.

Please note that I'm not primarily a Perl programmer, so please make sure these changes make sense. Thank you for taking the time to review these changes. I am primarily intending to improve readability and remove unused code.

### DIFF
--- a/lib/WeBWorK/Authen/Shibboleth.pm
+++ b/lib/WeBWorK/Authen/Shibboleth.pm
@@ -48,7 +48,6 @@ use strict;
 use warnings;
 use CGI qw/:standard/;
 use WeBWorK::Debug;
-use Data::Dumper;
 
 # this is similar to the method in the base class, except that Shibboleth
 # ensures that we don't get to the address without a login.  this means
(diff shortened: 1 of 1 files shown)


jrchamp:
On a side note: Is there a real reason that only MD5 is supported? Unless the user field has a 32 character limit, it seems like SHA-1 through SHA-512 would be viable options. [1]

[1] https://perldoc.perl.org/Digest.html

drgrice1:
This is set to merge into the master branch.  Only hotfixes should go into master.  I am rather certain this does not qualify as a hotfix.

lib/WeBWorK/Authen/Shibboleth.pm
```diff
@@ -63,57 +62,50 @@ sub get_credentials {
...
-		#    not to rely on internal logins if there's a check_user
-		#    failure.
-		$self->{external_auth} = 1;
+	debug("Shib is on!");
 
-		if ( $r->param("user") && ! $r->param("force_passwd_authen") ) {
-			return $self->SUPER::get_credentials( @_ );
-		}
-		
-		if ( defined ($ENV{$ce->{shibboleth}{session_header}}) && defined( $ENV{$ce->{shibboleth}{mapping}{user_id}} ) ) {
```
drgrice1, line 78: This re-introduces the incorrect $ENV usage fixed in PR#1056 (also on line 80 below).

jrchamp: #1056 appears to be a fix, but was not applied to master. My commit does not substantively change those lines, only the indentation.

jrchamp:
> This is set to merge into the master branch. Only hotfixes should go into master. I am rather certain this does not qualify as a hotfix.

Thank you for reviewing this, I'll see if I can rebase this on the develop branch.

jrchamp:
Okay, rebased on develop. Sorry for the branch confusion.

mgage:
Thank you @jrchamp for the improvements we appreciate your contribution.  And thank you @glennricster for the review.

PR #1092, opened 2020-02-27
Spanish translation (updating es.po)

enriqueacosta:
This es.po update is a merge, using msgmerge, of the current es.po file with the version in Transifex which is being worked on.
Thanks to Mike and Nathan for all the help on this. 
I hope I did this right!


mgage:
Cool -- let's give it a try.

PR #1094, opened 2020-03-10
Add the problem-main-form and form-inline classes to the gwquiz form.

drgrice1:
This brings the style of problems in gateway quizzes in line with the
style of problems in homework.  This fixes issue 1032.  This also fixes
other display issues that I have observed with problems in gateway
quizzes.


mgage:
changes only to css and .js files.  should be pretty safe.

PR #986, opened 2019-08-17
Feedback form expansion - REVISED

drdrew42:
# Purpose

Our institution uses the feedbackMacro_form function to direct students to ask questions on a Q&A site (https://openlab.citytech.cuny.edu/ol-webwork) where their questions are aggregated based on having the "same" problem (same pg file, different seeds). That way students have an FAQ on a per-problem basis.

We need some additional info from WeBWorK because the Q&A site does not share accounts with WeBWorK.

# Description

Okay, so - as suggested ( #957 ) - I restructured my expansion of (the necessary, for our Q&A WordPress plugin) form data sent through the `feedbackMacro_form` function. 

1. I removed my changes from `ContentGenerator.pm` and instead expanded the parameters passed to `feedbackMacro` when it is called from `output_email_instructor` in `CG\Problem.pm` and `output_footer` in `CG\ProblemUtil\ProblemUtil.pm`

2. In the process, I found multiple clusters of repeated code - so I shifted them into their own subroutines in WeBWorK::Util. The repositioned code now comprises two new subroutines: `fetchEmailRecipients` and `generateURLs` (which are made available through ContentGenerator.pm). These new functions helped to simplify the process of adding to the form data sent through the feedbackMacro

3. Please forgive the somewhat-large diffs on ContentGenerator.pm and Problem.pm - I have been using Atom as my editor, and it auto-strips trailing whitespace. The **only** relevant change to ContentGenerator.pm is line 62 (loading the new functions from Util); and in Problem.pm, the only meaningful change is to `output_email_instructor` starting on line 2116.

# Testing 

1. One should start by testing that their usual email functions are not affected by these changes.
1. As a student, use the email instructor button and confirm that it works as expected
1. Load up a JITAR set (make sure that there are a limited number of attempts!) - and, as a student, trigger the jitar-warning email by exhausting your allotted attempts. Confirm that an email is sent to the instructor (test TA account too?) warning that a student failed to complete a JITAR set.
1. add the following to the course.conf: 
`$courseURLs{feedbackFormURL} = 'http://localhost/test.php';`
`$feedback_button_name = "Test the PR"`
1. create /var/www/html/test.php with the contents:
```
<html>
<head><title>PHP Get Results</title></head>

<body>

<?php

// Show all form data submitted
foreach($_POST as $key=>$value){
    echo $key, ' => ', htmlspecialchars($value), "<br/>";
}

?>

</body>
</html>
```
[...]

PR #1097, opened 2020-04-04
feedback_form revision modification

drdrew42:
use $user->email_address instead of rfc822_mailbox, which (MIME-header) encodes the user's full name - but seems to fail in returning the email address along with the encoded full name.

I have instead switched to only send a semi-colon separated list of email addresses instead.

Refer to #986 for prior work on this functionality.


drdrew42:
Nevermind. Email addresses are being delivered. The angle brackets swallow them up when I try to dump the $_POST to be displayed in a browser.

Issue #967, opened 2019-06-26
Bug when submitting answers with non-latin characters when using XMLRPC


taniwallach:
https://github.com/openwebwork/webwork2/pull/956 added support for rendering UTF-8 problems via xmlrpc, in particular changes to `clients/sendXMLRPC.pl` and `lib/WebworkClient.pm` and the formats in the `lib/WebworkClient/' directory.

However, if an answer (or the value of any form field) includes non-latin1 characters, things break.

Some ideas of what may be causing the problem appear in the comments to https://github.com/openwebwork/webwork2/pull/956 but finding the root cause and fixing it remains to be done.

The same issue forced leaving the values of the 3 buttons (submit, preview, show correct answers) in English.

taniwallach:
I just ran into a similar problem with a different feature which used XMLRPC.

If I have a problem using `AnswerHints` as a post-filter, and the message being set uses non-Latin1 characters, things work fine when the problem is operated "normally" **but** the problem fails to render in the `instructor/sets2/setName` context when the "Render problem" icon is clicked. (That feature uses an XMLRPC call to `renderProblem`.)

After a short delay, the browser opens a popup with the text:
```
webwork2.technion.ac.il says
/webwork2/instructorXMLHandler: Forbidden
```
and the Apache error.log file get lines like:
```
Error message for command: renderProblem
<br/>faultcode: Client
<br/>faultstring: Wide character in subroutine entry at /usr/share/perl5/XMLRPC/Lite.pm line 181.

<br/>End error message<br/>
[Thu Aug 22 14:47:45.036034 2019] [perl:error] [pid 1628] [client <IP-NUMBER>:50384] [/webwork2/instructorXMLHandler] Error message for command: renderProblem \n<br/>faultcode: Client \n<br/>faultstring: Wide character in subroutine entry at /usr/share/perl5/XMLRPC/Lite.pm line 181.\n \n<br/>End error message<br/>\n * in WeBWorK::ContentGenerator::instructorXMLHandler::content called at line 233 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::go called at line 384 of /opt/webwork/webwork2/lib/WeBWorK.pm, referer: https://webwork2.technion.ac.il/webwork2/<COURSE_NAME>/instructor/sets2/xmlRPwideCharacterIssue/?key=<KEY>&effectiveUser=<eUser>&user=<User>
```

[...]

PR #1099, opened 2020-04-10
Remove hack that worked around glitches in using utf8 in xmlrpc. This…

mgage:
… appears to work alright now.

To test:
```
<!DOCTYPE html>
<html lang="es"><head>
<meta http-equiv="content-type" content="text/html; charset=UTF-8">
  <meta charset="utf-8">

  <title>Título</title>
  <meta name="author" content="GrupoLEMA">
</head>

<body>
   <h1>Hello world</h1>
   <p>
      Here is some text.
   </p>
   <iframe src="https://demo.webwork.rochester.edu/webwork2/html2xml?
   &amp;answersSubmitted=0&amp;
   &amp;sourceFilePath=local/problema.pg&amp;
   &amp;problemSeed=123567&amp;
   &amp;courseID=spanish_language_course&amp;
   &amp;userID=anonymous&amp;
   &amp;course_password=test&amp;
   &amp;showSummary=1&amp;
   &amp;displayMode=MathJax&amp;
   &amp;problemIdentifierPrefix=102&amp;
   &amp;language=es&amp;
   &amp;outputformat=sticky" width="800" height="400">
   </iframe>

</body></html>
```

The buttons will appear in English before the fix.  After the fix they will appear in Spanish except possibly for "Show correct answers" which doesn't seem to have been translated yet.

### DIFF
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -437,11 +437,6 @@ EOS
 	my $STRING_ShowCorrect = $mt->maketext("Show correct answers");
 	my $STRING_Submit      = $mt->maketext("Check Answers");
 
-# With these values - things work, but the button text is English
-# with the localized values, or any answers in UTF-8 - thing break
-$STRING_Preview = "Preview My Answers";
-$STRING_ShowCorrect = "Show correct answers";
-$STRING_Submit = "Check Answers";
 
 ######################################################
 # Return interpolated problem template


taniwallach:
My local system is already using Hebrew buttons with XMLRPC, having commented out the links this PR deletes. Thus - as far are I know this should be safe to do, on a system with all the relevant fixes in place. The problems is that mu systems have lots of small patches in place, and I am not absolutely sure what is really needed to avoid the problem the hack was working around.

I stronly suspect that the patch to `/usr/share/perl5/XMLRPC/Lite.pm` suggested in https://github.com/openwebwork/webwork2/issues/967#issuecomment-523941959 is the change which is most likely to have overcome the issue with UTF-8 text in form values and XMLRPC/html2xml which motivated the hack being removed here.

Note: There is another open PR related to fixing UTF-8 support for XMLRPC: https://github.com/openwebwork/webwork2/pull/1089 which overcomes problems with certain types of questions with UTF-8 content which could not otherwise be processed with html2xml/XMLRPC.

PR #1101, opened 2020-04-19
Tweak formats to work better with problems embedded in HTML

mgage:
This pull request contains tweaks to the simple_format  themes which are used when embedding problems into Libretext and other HTML pages. In particular the final lines of a problem listing the webwork and pg release numbers and copyright info are hidden inside a knowl so that they don't distract when several problems are embedded in a page


mgage:
Closed and replaced by PR #1112

PR #1102, opened 2020-04-22
Use current time for submission time for un-proctored gateway quizzes

drgrice1:
When checking to see if answers for a gateway quiz can be recorded or
checked, only use the version_last_attempt_time for the submission time
if the gateway quiz is proctored.  For un-proctored gateway quizzes
always use the current time.

If multiple attempts are allowed, then this was making it possible for
students that had graded the test at least once (but had not exhausted
attempts) to continue to grade the test after the quiz time had expired.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4756 for the discussion that led to this pull request.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -183,9 +183,9 @@ sub can_recordAnswers {
    # get the sag time after the due date in which we'll still grade the test
 	my $grace = $self->{ce}->{gatewayGracePeriod};
 
-	my $submitTime = ( defined($Set->version_last_attempt_time()) &&
-			   $Set->version_last_attempt_time() ) ? 
-			   $Set->version_last_attempt_time() : $timeNow;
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I tested this and it works. Note if anyone does any more testing to account for the grace period, which is 2 minutes in defaults.conf.

drgrice1:
This pull request should also be considered to be merged into master for a hot fix.  I think that this issue is a particularly relevant one considering the times we are in.  More instructors than usual will probably be trying to use online un-proctored quizzes.

taniwallach:
> This pull request should also be considered to be merged into master for a hot fix. I think that this issue is a particularly relevant one considering the times we are in. More instructors than usual will probably be trying to use online un-proctored quizzes.

I agree that a hot-fix is probably worth-while. Unfortunately, I suspect that many people who run into the issue will not yet be on WW 2.15 so a hot-fix to master will not help them much. Additional advertisement of this issue and the fix via the forums (or elsewhere) may be helpful.

PR #1105, opened 2020-05-07
Develop+po.es

mgage:
Add Enrique's new spanish translations.


mgage:
replace this with a PR that only changes the .po file

PR #1106, opened 2020-05-07
Add Enrique's updated spanish translation: es.po

mgage:
Replaces earlier PR which had additional changes.


mgage:
This updates the es.po file

PR #1108, opened 2020-05-11
MathQuill style change to support use with right-to-left languages

taniwallach:
Force 2 MathQuill CSS style items to set left-to-right text direction, so MathQuill (input boxes and the toolbar) will work properly in courses whose primary HTML text direction is right-to-left (ex. Hebrew courses). This has been in use for a while in production in Hebrew courses, and should not have any operational effect on courses whose default text direction is left-to-right.

### DIFF
--- a/htdocs/js/apps/MathQuill/mqeditor.css
+++ b/htdocs/js/apps/MathQuill/mqeditor.css
@@ -24,6 +24,7 @@ span[id^=mq-answer].incorrect /* red */
 
 span[id^=mq-answer]
 {
+    direction: ltr;
     padding: 4px 5px 2px 5px;
     border-radius: 4px !important;
     background-color: white;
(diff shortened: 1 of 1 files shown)


drgrice1:
This seems fine to me.

What is the problem without the left-to-right direction on those elements in a right-to-left language?

taniwallach:
> This seems fine to me.
> 
> What is the problem without the left-to-right direction on those elements in a right-to-left language?

I happened to have some old images inside emails from when we first worked on this.

In the input boxes it tried to order the "pieces" of the math expression from right to left which would totally mange the expression: 
  - Typing `2x - 3pi/2` would display 
MQ-input-text-reversed
  - Typing `5+sqrt(20)+sqrt(65)` would display 
MQ-input-text-reversed2

Hebrew texts keep the math as "left to right" embedded as inline or display.

In the toolbar - the buttons also were messed up for the same type of reason.
  - Before:
MathQuill-Hebrew-before-fix
  - After: 
MathQuill-Hebrew-after-fix

drgrice1:
I see.  Thanks.

PR #1113, opened 2020-06-01
Develop persistence fix

mgage:
Fixes a regression with maintaining persistence for geogebra applets.  The important change is in the subroutine create_ans_str_from_responses in ProblemUtil.pm.  

There are also name changes to make it easier to follow the code and several historical sections which had been commented out are removed since that information is now more confusing than helpful.  The historical record is available in github.

To test:

Create a problem set containing `Library/FortLewis/Calc1/06-01-Antiderivatives-graphically/AF1/AF1.pg`  Drag the pointer in the applet to some new position (say x=12) and submit it.
(Make sure the due dates are set so that the problem is active and the answer is recorded.) The new position of the pointer is maintained after the submission.  Now move away from the problem page (up to the problem set for example) and then move back.  Before the patch the pointer position will not remain at x=12 but will revert to the original default.  After the patch the pointer position will remain at x=12.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
@@ -106,18 +106,18 @@ sub process_and_log_answer{
 #    				    push @answer_order2, $response_id;
 #    				 }
 #    			}
-#    			my $answerString2 = '';
+#    			my $past_answers_string = '';
 #    			foreach my $response_id (@answer_order2) {
-#    				$answerString2.=($answersToStore2{$response_id}//'')."\t";
+#    				$past_answers_string.=($answersToStore2{$response_id}//'')."\t";
 #    			}
-#    			$answerString2=~s/\t$//; # remove last tab
-	my ($answerString2);
-	($answerString2,$encoded_answer_string, $scores2, $isEssay2) =
+#    			$past_answers_string=~s/\t$//; # remove last tab
+	my ($past_answers_string);
+	($past_answers_string,$encoded_answer_string, $scores2, $isEssay2) =
 	    WeBWorK::ContentGenerator::ProblemUtil::ProblemUtil::create_ans_str_from_responses(
 	      $self, $pg
 	    );  # ref($self) eq WeBWorK::ContentGenerator::Problem
 	        # ref($pg) eq "WeBWorK::PG::Local";
-# end new code (output is answerString2, $scores, $isEssay)
+# end new code (output is past_answers_string, $scores, $isEssay)
 ################################################################
 # 		    my $answerString = ""; my $scores = "";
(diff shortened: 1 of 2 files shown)


mgage:
This does not fix the persistence problem with file Uploads (see ).  That is caused because storing the long base64 string required to store a png file in the database breaks the encodeAnswer and decodeAnswer subroutines.

glarose:
Hi Mike,

Reading through this I don't have any reservations.

Gavin

On Mon, Jun 01, 2020 at 06:37:56AM -0700, Michael Gage wrote:
> @mgage requested your review on: openwebwork/webwork2#1113 Develop persistence fix.
> 
> -- 
> You are receiving this because your review was requested.
> Reply to this email directly or view it on GitHub:
> https://github.com/openwebwork/webwork2/pull/1113#event-3393161140

-- 
-- 
P. Gavin LaRose (he/him/his) | glarose@umich.edu | 734.764.6454 | EH 3832
Karen Rhea Collegiate Lecturer & Instructional Technology Program Manager
University of Michigan Department of Mathematics

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
```diff
@@ -381,43 +298,62 @@ sub process_and_log_answer{
...
 	my $isEssay2=0;
-	my %answersToStore2;
-	my @answer_order2;
-	my @answer_order3;
+	my %answers_to_store;
+	my @past_answers_order;
+	my @last_answer_order;
 
-	my %answerHash2 = %{ $pg->{pgcore}->{PG_ANSWERS_HASH}};
+	my %pg_anwers_hash = %{ $pg->{pgcore}->{PG_ANSWERS_HASH}};
```
drdrew42, line 327: (consistent) typo: `%pg_answers_hash`

mgage: not positive what you mean.  If you are calling attention to the capitalization in pg_answers_hash in line 327 -- that is intentional (at least for now).  The all caps convention has long been used for global variables (e.g. $PAR) or foundational subroutines (ANS_RULE()) dating to the days before perl had objects. In this case I'm trying to get away from that convention since %pg_answers_hash is used entirely within subroutines two subroutines and is not exposed to the global Safe compartment as PG_ANSWERS_HASH is in PG.pl.  However I'm not prepared yet to chase all of the uses of PG_ANSWERS_HASH down and replace it with a variable protected inside an object. Some day perhaps. :-)

drgrice1: I think drdrew42 meant it should be pg_an**s**wers_hash instead of pg_anwers_hash.

mgage:
sorry — my eyes aren’t working. :-(   ty.

PR #1115, opened 2020-06-15
Allow Achievement Items to be used more than once

Alex-Jordan:
Currently, a user either has a certain achievement item or they don't. This changes things so that they have N copies of the item. They can use an item, and N will reduce by 1. (If N is negative, the user can reuse the item with no limitation.)

This will be useful for awarding students, say, 10 copies of the item that allows for a 24-hour extension. (Which will be like MyOpenMath's "late passes" feature.) Or you can award 20 copies of an item to re-randomize the seed. Or 2 items to grant 70% on an entire assignment. Etc.

Nothing should change if you use the current achievements-with-items that come with the distribution. However, if you make new item-using achievements (or edit the defaults) you can make it so students get more than one shot with an item.

For example, you could load the default achievements with items. Then you could edit the level 1 achievement evaluator to have:
```
$globalData->{HalfCreditProb} = 3;
```
Then assign the level 1 achievement to yourself and enable it. Then submit an answer so you earn the level 1 reward. (It will say you are awarded "a" Lesser Rod of Revelation, but you really earned 3. To change that text, you separately edit the achievement's attributes.)

Then you will see in the Achievements page that you have 3 of them.  Try using them all.

Go back to the Achievements Editor and edit the level 1 evaluator to now say:
```
$globalData->{HalfCreditProb} = -1;
```
You can click to Edit Users on the level 1 achievement and uncheck yourself to make it as if you hadn't already won the level 1 achievement. Now submit anther answer, and see that you have b been awarded unlimited use of this item in the Achievements page.


lib/WeBWorK/ContentGenerator/Achievements.pm
```diff
@@ -229,9 +236,12 @@ sub body {
 			    
 		my $itemnumber = 0;
 		foreach my $item (@items) {
-		    # Print each items name and description 
+		    # Print each item's name, count, and description
 		    print CGI::start_div({class=>"achievement-item"});
-		    print CGI::h3($r->maketext($item->name()));
+		    if ($itemCounts{$item->id()} > 1)    {print CGI::h3($r->maketext($item->name() . ' (' . $itemCounts{$item->id()} . ' remaining)'))}
+		    elsif ($itemCounts{$item->id()} < 0) {print CGI::h3($r->maketext($item->name() . ' (unlimited reusability)'))}
```
drgrice1, line 242: The `count` strings will not be translated as maketext is not called on them.
Line 241 should be changed to
```perl
if ($itemCounts{$item->id()} > 1) {print CGI::h3($r->maketext($item->name() . $r->maketext('([_1] remaining)', $item->id()))}
```
and line 242 to
```perl
elsif ($itemCounts{$item->id()} < 0) {print CGI::h3($r->maketext($item->name() . $r->maketext(' (unlimited reusability)')))}
```

Alex-Jordan: Can you confirm that in each line, `$r->maketext($item->name()` should be `$r->maketext($item->name())`? I think that is what you mean, but I am second guessing if that extra close paren belongs at the end.

Also if I read this right, there will be no space between the item name and the count. Like
```
Scroll of Whatever(3 remaining)
```
instead of 
```
Scroll of Whatever (3 remaining)
```
Is it right to simply concatenate a space there?

drgrice1:
You have it.  Both of the things you said are correct.  Another parentheses is needed on those `maketext` calls, and a space should be concatenated in between the name and the count.  Good catch.

Alex-Jordan:
OK, I think I edited those lines to use `maketext()` correctly.

drgrice1:
Could you give us an idea of how to test this?

drgrice1:
Never mind, I didn't read your initial post closely.

drgrice1:
This looks good, and works as advertised.  I added a little javascript fix.  I thought I had fixed that already.

PR #1116, opened 2020-06-16
Graphtool macro for interactive graphing in problems via javascript (javascript for PG PR #484)

drgrice1:
See https://github.com/openwebwork/pg/pull/484.


drgrice1:
The corresponding pull request was merged to PG.  So this needs to be merged also.

PR #1121, opened 2020-07-15
Fix applets in gateway quizzes

drgrice1:
This updates the javascript in ww_applet_support.js with changes needed to make applet problems work in gateway quizzes.  The main changes to this file are to make Geogebra web applets work on a single page.

Some changes were also needed in the GatewayQuiz.pm file itself.  The preview hack needed to be modified to ensure that the javascript submit event occurs when the preview buttons are pressed and when a page change occurs.  This also adds a method to add javascript files to the header without repetition from a PG problem for both homework problems and gateway quizzes.  The corresponding ADD_JS_FILE method is defined in PG.pl in a corresponding pull request to the pg code.

There is one other change to the GatewayQuiz.pm file, that changes when answers are stored to the past_answer database and logged to the course answer log file.  Now the only time this is done is when answers are submitted, and not when a page change or preview occurs.  This also removes the 'submit', 'preview', or 'newPage' prefix that was prepended to the answers.  That wasn't working well in general.  Mathquill latex answers and anything in KEPT_EXTRA_ANSWERS was begin put into the database and the log.  Furthermore, the prefix was messing with the formatting when past answers are viewed.  In addition, an applet state saved in KEPT_EXTRA_ANSWERS would easily make the size of the data exceed the limits for the column in the database.

This is separated into two commits.  The first is purely code clean up.  The main content is in the second commit.

See https://github.com/openwebwork/pg/pull/489 for the corresponding pg code changes.


drgrice1:
Here is one test file.  This is a modified version of one of Drew's problems from Contrib.  This problem will work in both homework and in gateway quizzes with this pull request.  It won't work without this pull request though.  I only made the minimum changes needed to make it work in gateway quizzes, so you can see the changes that problem editors need to make to existing problems to update them (although if those changes aren't made the problems will continue to work in homework).
[two-lines-parallel.zip](https://github.com/openwebwork/webwork2/files/5462437/two-lines-parallel.zip)

You should also test the original file Contrib/CUNY/CityTech/CollegeAlgebra_Trig/setLines/two-lines-parallel.pg.  That file will work both with and without this pull request in homework, but will not work in gateway quizzes with or without this pull request.

taniwallach:
> Here is one test file. This is a modified version of one of Drew's problems from Contrib. This problem will work in both homework and in gateway quizzes with this pull request. It won't work without this pull request though. I only made the minimum changes needed to make it work in gateway quizzes, so you can see the changes that problem editors need to make to existing problems to update them (although if those changes aren't made the problems will continue to work in homework).
> [two-lines-parallel.zip](https://github.com/openwebwork/webwork2/files/5462437/two-lines-parallel.zip)
> 
> You should also test the original file Contrib/CUNY/CityTech/CollegeAlgebra_Trig/setLines/two-lines-parallel.pg. That file will work both with and without this pull request in homework, but will not work in gateway quizzes with or without this pull request.

---
Before the PR is applied:

1. The original question in a GateWay gives an error message which prevented the GateWay from loading at all:
```
Gateway Quiz applet_pr_test_gw,v1
WeBWorK Error

WeBWorK has encountered a software error while attempting to process this problem. It is likely that there is an error in the problem itself. If you are a student, report this error message to your professor to have it corrected. If you are a professor, please consult the error output below for more information.
Error messages

Can't locate object method "new" via package "myMultiAnswer" at line 33 of (eval 1123) Died within myMultiAnswer::new called at line 105 of (eval 1123)

Error details

        Problem 1
2. ERROR caught by Translator while processing this problem 
```

---

When the original question was replaces with the revised version, the GateWay did load and reported:
```
*Unable to perform an explicit initialization action (e.g. setState) on applet TwoLinesParallel because 
 TypeError: document.problemMainForm is undefined
```
6 times. At this point the GW had 2 copies of the revised question (one replacing the original one, as it would not load at all - see above).

Five "GeoGebra" spaces were visible (only 2 were expected), and only the bottom one had the dotted red line where expected.

a_mess_in_gw_working_when_old_code_active
a_mess_in_gw_working_when_old_code_active_part2

revised_question_in_gw_error_when_old_code_active


---

With the revised code from the pair of PRs installed, the revised problem is OK, but the original one is not.

revised_question_in_gw_working_PR_code_active

original_question_in_gw_error_in_parameters

taniwallach (approved):
Looked over the code, but did not attempt to follow all the change in detail. Seems OK and certainly fixes the bug with the use of applets in gateway quizzes.

The new method of requesting a special JavaScript file inclusion is a huge improvement with many potential additional uses.

@drgrice1 I believe you have been using this in production, as am approving without any serious local stress testing.

drgrice1:
The Contrib problem will not work in Gateway quizzes at all.  It won't work without this pull request, and it won't work with this pull request.  Furthermore, if that is in a gateway quiz on the same page as another Geogebra problem like the one in the attached zip, it will even mess up that problem.  There is no fix for that.  Problems will need to be updated to work in Gateway quizzes.

The problem in the attached zip does work in Gateway quizzes, just don't try to use an old Geogebra on the same page.

drgrice1:
@taniwallach:  Thanks for your review.

taniwallach:
> The Contrib problem will not work in Gateway quizzes at all. It won't work without this pull request, and it won't work with this pull request. Furthermore, if that is in a gateway quiz on the same page as another Geogebra problem like the one in the attached zip, it will even mess up that problem. There is no fix for that. Problems will need to be updated to work in Gateway quizzes.
> 
> The problem in the attached zip does work in Gateway quizzes, just don't try to use an old Geogebra problem on the same page.

Understood. We should probably attempt to notify various people who authored problems using applets to fix their code in Contrib and in the OPL. 

@drdrew42  - please see if the CUNY problems in Contrib can be fixed. If you can test these changes - an additional review would be appreciated.

drdrew42:
I did not technically review this side of the paired "applet" PRs, though I did incorporate the changes to the supporting applet scripts into the renderer and confirm that they worked successfully.

I will see what I can get done this weekend:
* batch processing my Contrib problems that use GGB (esp. those with multianswer)
* confirm these edits preserve problem functionality

I am not sure what to do, however, with regards to backwards-compatibility... Anyone who does not update to the latest PG version will be unable to render these problems? I'm thinking specifically with regards to the changes to multianswer, since I don't believe the changes that are necessary for the applet updates would be 'breaking'... is that accurate?

drgrice1:
Here is another version of the two-lines-parallel problem that will work in homework both with and without this pull request, and will work in Gateway quizzes with this pull request.  It works on a technicality, and there may be a better way to implement this.  The technicality it works on is that in this pull request the PGcore hash key SECTION_PREFIX is removed.  So prior to this pull request $PG->{SECTION_PREFIX} will be defined.  With this pull request it will not be.
[two-lines-parallel-compatible.zip](https://github.com/openwebwork/webwork2/files/6011584/two-lines-parallel-compatible.zip)

drdrew42:
Sorry, just for clarification - the SECTION_PREFIX removal is in another PR? One that has/hasn't been merged into develop (but presumably it is anticipated for 2.16)? 

_also, very clever_ ;P
The changes given here should be applicable to most of my GGB problems (though I cannot recall why I was so insistent about using multianswer on these problems where there is only ONE answer...) - if time permits, I might as well just refactor most of these problems that don't actually require multianswer.

drgrice1:
The SECTION_PREFIX removal is in this PR (well really the other part in the PG code).

taniwallach:
> Sorry, just for clarification - the SECTION_PREFIX removal is in another PR? One that has/hasn't been merged into develop (but presumably it is anticipated for 2.16)?
> 
> _also, very clever_ ;P
> The changes given here should be applicable to most of my GGB problems (though I cannot recall why I was so insistent about using multianswer on these problems where there is only ONE answer...) - if time permits, I might as well just refactor most of these problems that don't actually require multianswer.

Wouldn't it be better to allow PG code to check PG version and use that to decide "which" approach to take? Is that possible?

drgrice1:
@taniwallach: I was looking for something like that, but as of yet I have not found a way to obtain the PG version in a PG problem.  Seems like something that should be available though.  And yes, that would certainly be better.

mgage:
I suggest making it part of IO.pm which already interacts with the disk (we tried to keep all of the "dangerous" disk interaction in one place so that we could trouble shoot errors better).   Give IO.pm a subroutine that reads the PG_version file (and sanity checks that the version is written correctly). You can also have another that reads the webwork2 version file.  Then export the ability to read that value to PGcore and perhaps further to PG.pl or PGbasicmacros.pl so that the version can be easily checked.  webwork2 has access to the version files, but PG did/does not as far as  I know

taniwallach:
@drgrice1 there are now 2 merge conflicts here.
  - `htdocs/js/apps/AppletSupport/ww_applet_support.js`
  - `htdocs/js/apps/Problem/problem.js`

I assume they result from the jQuery update.

Could you make the relevant changes in your branch, and then I intend to merge this and the paired https://github.com/openwebwork/pg/pull/489, unless either @Alex-Jordan, @pstaabp,  or @drdrew42 have any reservations.

drgrice1:
Yes, almost all of my pull requests have conflicts with each other.  I will resolve them as the others are merged.  I will take care of this one shortly.

taniwallach:
> Yes, almost all of my pull requests have conflicts with each other. I will resolve them as the others are merged. I will take care of this one shortly.

Yes - the huge jQuery change and the changes to parallel files makes handling these somewhat tedious.

drdrew42 (approved):

Alex-Jordan:
I just did simple testing. I saw what I was supposed to see with the Contrib problem in a quiz, and what I was supposed to see with the new version in a quiz. After reading through the threads, I think I will merge now.

I did notice that if a student takes the quiz, and then you go to Set Detail 2 for the students quiz version, and then click "Render all", the applet problem doesn't load the applet part. And there are console errors. This seems like a separate thing to pursue. I'll post an issue about it.

drgrice1:
Good to note.  I will look into it.

drgrice1:
It seems that applet problems do not work on the Set Detail 2 page via the `Render` button at all.  This is because the javascript needed for these problems is added to the header in the problem, and the Set Detail 2 page does not get that.  So that means these also won't work in the library browser as it also does not load the header from a problem.  This is the same as before though.  That is not a change with this pull request.

drgrice1:
Still worth looking into though.

PR #1123, opened 2020-08-23
Model course setOrientation problem6 bugfix

gajennings:
Bug reported as Problem Libraries bug 4450 occurs in webwork2/courses.dist/modelCourse/templates/setOrientation/prob06.pg.  In  "image" mode this problem displays math expressions that are too small.  Rewriting in PGML fixed the display problem.


dpvc:
I would be a bit hesitant to make this change without also modifying all the other problems in the orientation set.  They are designed as a collection, and have consistent layout and approach.  Changing one problem and not the others will disrupt that.

In any case, the issue is actually in the `bootstrap.css` file, which includes `img {max-width: 100%}`, and that is causing the images to shrink.  This could affect equations in tables in *any* problem, so should probably be fixed.  (I hate it when library CSS files change the defaults for un-modified HTML tags.)  Fixing this one PG file is fixing the wrong issue.

So I would vote to reject the change.  (Sorry!)

PR #1124, opened 2020-08-25
Slight fix to errors reported by DB::validateKeyfieldValue

mikeshulman:
Some missing quotation marks, and in one case the reported regexp didn't match the one in use.

### DIFF
--- a/lib/WeBWorK/DB.pm
+++ b/lib/WeBWorK/DB.pm
@@ -2319,10 +2319,10 @@ sub validateKeyfieldValue {
     my ($keyfield,$value,$versioned) = @_;
 
     if ($keyfield eq "problem_id" || $keyfield eq 'problemID') {
-	croak "invalid characters in '".encode_entities($keyfield)." field: '".encode_entities($value)."' (valid characters are [0-9])"
+	croak "invalid characters in '".encode_entities($keyfield)."' field: '".encode_entities($value)."' (valid characters are [0-9])"
 	    unless $value =~ m/^[0-9]*$/;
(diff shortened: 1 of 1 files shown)


taniwallach (commented):
Correction to valid characters for the third case is an important correction. Adding the missing single quote after the field name is a small cleanup. 
However, I'm not certain that this is mission critical enough to bother making it a hot-fix to master. Maybe it suffices to make the fix in develop for the next version of WW?

taniwallach:
@drgrice1 - What do you think: hot-fix to master or develop?

drgrice1:
I agree with you.  I missed that this pull request was set to go into master.  This really isn't a critical fix, and should go into develop.

mikeshulman:
Sorry, my bad, I was unaware of how branches are used in this repo.  Would you like me to close it and create a new PR to develop?

drgrice1:
You can just change the branch that is is being requested to be added to.  We can take care of this for you, if you like.

drgrice1:
I guess, that is assuming that it will cleanly merge into the develop branch.  Sometimes that doesn't work so well.

drgrice1:
Yeah, it looks like you will need a new pull request.  The change to the develop branch doesn't work out.

mikeshulman:
Ok, I cherry-picked the original commit into a new branch based on develop, force-pushed it to this PR, and changed the target branch to develop.  Hope this works?

drgrice1:
Yep.  That works.  That is how I often do it, and should have thought of that.

taniwallach (approved):
@mikeshulman - Thanks for changing the target branch.

Now that it is targeted to develop - I think it is ready to merge.

PR #1125, opened 2020-08-28
Fix typo ENABLE_UTF8MB -> ENABLE_UTF8MB4

mikeshulman:
See discussion at mikeshulman/webwork2#1.

### DIFF
--- a/conf/database.conf.dist
+++ b/conf/database.conf.dist
@@ -80,7 +80,7 @@ my %sqlParams = (
 	username => $database_username,
 	password => $database_password,
 	debug    => $database_debug,
-    ($ENABLE_UTF8MB)?(mysql_enable_utf8mb4 =>1):(mysql_enable_utf8 => 1),
+    ($ENABLE_UTF8MB4)?(mysql_enable_utf8mb4 =>1):(mysql_enable_utf8 => 1),
 	# kinda hacky, but needed for table dumping
 	mysql_path     => $externalPrograms{mysql},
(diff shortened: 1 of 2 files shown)


taniwallach (approved):
I think both changes are correct. 

The fix is needed as the code is currently inconsistent, but it seems that many systems would not notice the bug. I suspect that systems where the database is not configured to default to `utf8mb4` for the connection may have issues at present.

I certainly do not encounter any issues on a Docker setup before the patch, and I could put a 4-byte UTF-8 character into a students name.

Why?
  1. My database is configured directly to use `utf8mb4` for the connection charset, etc. using `docker-config/db/mariadb.cnf` so the bug causing a failure to call `SET NAMES 'utf8mb4';` is not causing trouble.
  2. It seems that Perl's DBD when when `mysql_enable_utf8` is used (instead of `mysql_enable_utf8mb4`) does not really make any effort to prevent 4-byte UTF-8 characters from being passed between Perl and the database. Instead, to documentation instructs users of the limited 3-byte mySQL `utf8` charset that such checks are to be added by the people writing code which uses the restricted (somewhat broken) `utf8` character set.
      - https://metacpan.org/pod/DBD::mysql
     - https://github.com/perl5-dbi/DBD-mysql/blob/master/lib/DBD/mysql.pm
       - Note the lines (emphasis added)
> When the C<utf8> charset is used then **you are responsible ￼for 3-byte UTF-8 sequence checks** on input perl scalar strings.  Otherwise MySQL
￼server can reject or modify the input statement!

Issue #899, opened 2018-11-12
Gateway Quiz with non-consecutively numbered problems


heiderich:
It seems that when a homework set of type "gateway quiz" with questions numbered in a non-consecutive way is taken by a student, they may get the following error:

"The selected problem set (`name of homework set`) is not a valid set for `user`:
One or more of the problems in this set have not been assigned to you."

This was already reported in the forum by @Alex-Jordan (http://webwork.maa.org/moodle/mod/forum/discuss.php?d=3134), but I could not find a bug report for it and I am still experiencing this with WeBWorK 2.13.

mikeshulman:
I ran into this error and would like to help fix the problem.  The forum discussion suggests that the underlying bug may be unfixable without a database redesign.  Is that really true?  Looking at the code that produces that error in `ContentGenerator/GatewayQuiz.pm` my guess is that what's happening is that `@problemNumbers` is a list of the problem numbers appearing in the quiz, and the code assumes that these are sequential from 1 so that subtracting 1 from a problem number yields the 0-based index `$pIndex` of the problem.  But couldn't one just iterate through the indices directly, like in 5873fce1d79?  I don't know whether such a change would necessitate other changes elsewhere, but I just tested it very briefly on my installation and it seemed to work.

drgrice1:
mikeshulman:  This does seem to fix the problem.  I have tested it, and it works well.  Analyzing the code as it currently is, it seems that it was a mistake to decrement the problem number to achieve zero based indexing.

I would recommend that you drop the `my $problemNumber = $sortedProblemNumbers[$pIndex];` line, and just assign `$sortedProblemNumbers[$pIndex]` directly to `$pg` on line 1165.

You should put a pull request in for this.

drgrice1:
So looking at this a bit closer, it seems that the actual problem number isn't even needed.  Currently the `$pg` variable is set to the problem number, and that is pushed onto the `@pg_results array`, and then `$self->{ra_pg_results}` is set as a reference to that array.  Looking at the usage of the `ra_pg_results` key of that hash throughout the file, I see that the problem number is not actually used anywhere.  All that is important is the value of that entry in the array **not** be a reference.  So it could be set to 0, and it will work.  The only problem is decrementing the `$pIndex` variable.

mikeshulman:
Thanks for looking at this!  Would you like the pull request to do what you suggest in the second comment and not even bother to sort the array?

drgrice1:
[...]

PR #1128, opened 2020-08-30
Fix to iteration over GWQ problems in case of missing numbers

mikeshulman:
See #899.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1139,10 +1139,8 @@ sub pre_header_initialize {
 		@mergedProblems = $db->getAllMergedProblemVersions($effectiveUserName, $setName, $setVersionNumber);
 	}
 
-	foreach my $problemNumber (sort {$a<=>$b } @problemNumbers) {
+	for my $pIndex (0 .. $#problemNumbers) {
 
-		# pIndex numbers from zero
-		my $pIndex = $problemNumber - 1;
 		if ( ! defined( $mergedProblems[$pIndex] ) ) {
 			$self->{invalidSet} = "One or more of the problems " .
 				"in this set have not been assigned to you.";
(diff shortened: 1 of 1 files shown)


taniwallach:
The fix probably enables access. However, there may be additional issues, as there are indications in the discussion at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4932 that this issue may occur when a user's gateway assignment data is not consistent with the main quiz definition. 

If merging this could lead to a student being able to open a quiz which is not what they should be getting - then maybe leaving the error message is a better approach, as it would force an unassign + renumber + reassign to fix the problem and get the student the "correct" assignment data.

drgrice1:
I> The fix probably enables access. However, there may be additional issues, as there are indications in the discussion at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4932 that this issue may occur when a user's gateway assignment data is not consistent with the main quiz definition.

The issues in that forum discussion are not caused by the issue that this fixes.  User assignment data not matching the main definition is a problem that occurs for both homework assignments and gateway quizzes, and is another issue.

> If merging this could lead to a student being able to open a quiz which is not what they should be getting - then maybe leaving the error message is a better approach, as it would force an unassign + renumber + reassign to fix the problem and get the student the "correct" assignment data.

Merging this will not give students access to quizzes that they should not have access to.  That is enforced by permissions, and is unrelated to this.  It should never be the case that an assignment is assigned to a student, and yet some problems in that assignment are not.

drgrice1:
> If merging this could lead to a student being able to open a quiz which is not what they should be getting - then maybe leaving the error message is a better approach, as it would force an unassign + renumber + reassign to fix the problem and get the student the "correct" assignment data.

To add more to what I have already said, note that this does not have anything to do with the student data not matching the assignment data.  The fact is the data does match.  However, there is an indexing issue that is fixed by this pull request that causes an error for everyone that the set is assigned to.  No amount of unassigning and reassigning, or exporting and importing will fix this.  The only thing that will is making the problems be numbered consecutively (until this pull request is merged).  It doesn't make any sense to create a quiz and assign it to students that will never work, and to say that it is valid to do so to prevent students from accessing it is really crazy.  Don't assign it, or make it not visible.

taniwallach (approved):
OK - got it. I thought about the changes made here, and they certainly make sense.
I have been convinced that the issues caused by editing assignments which have collected student data which can get messed up should not hold this PR back.
I did not test myself, and am relying on the test by @drgrice1 .

drgrice1:
I think this can be merged.  I am going to do so unless there is an objection.

PR #1130, opened 2020-09-18
Develop answer log

pschan-gh:
With these modifications (assuming they work as intended):

An experimental answer log may be enabled/disabled as an optional module under Course Configuration.

Once enabled, an Answer Log button would appear on the side panel.

An extra "Answer Log Modal" button would appear after pressing "Display Past Answers" in the traditional show-past-answers form. Clicking on it sends answer log entries to a modal window for further processing in following way:

The user has to specify the number of recent attempts (counted from the most recent one) to display. I am using the answerID to track this, assuming that the manner in which mysql updates the answer records is perfectly chronological, but I don't actually know if this is a valid assumption.

The popup modal window then displays the set of selected users against the set of selected problems (selected via the traditional show-past-answers form) in a 2-dimensional table.
Given any positive integer n <= (the number of recent attempts specified), the n-th attempt of each answer of each (sub)-problem is collated into the same n-th row for each user. This is the kind of output that I would find useful, but can't seem to be able to obtain from the traditional show-past-answers output. All this reorganization is done client-side via javascript.
I figure nothing goes back to the server in the context of viewing the answer log.

The user may export the table to a csv file, now in a format which hopefully makes further analysis and manipulation easier. Correct answers are enclosed by "**".

The interface right now is a bit rough, and I wouldn't say everything is well tested. But before I proceed any further perhaps its good to hear other people's opinions.


drgrice1:
This is going to take some time to review.  I like the code clean up, but that does hinder finding relevant changes.

One thing that I noticed immediately is that you utilize ES6 features of JavaScript.  At this point we still support browsers that do not support ES6.  So either that JavaScript will need to be transpiled, or it will need to be rewritten.

pschan-gh:
> This is going to take some time to review. I like the code clean up, but that does hinder finding relevant changes.

Thank you very much.  Are you talking about ShowAnswers2.pm? Maybe I should have worked directly on ShowAnswers.pm instead, so that one can more easily see the line-by-line diffs. If this is what the community prefers I am perfectly fine with taking out the new file and work on ShowAnswers.pm instead.  Sorry about the trouble!

> 
> One thing that I noticed immediately is that you utilize ES6 features of JavaScript. At this point we still support browsers that do not support ES6. So either that JavaScript will need to be transpiled, or it will need to be rewritten.

OK, I will try to fix this.  I think I really used only the most basic features of Javascript, so hopefully it wouldn't require too much correction, but I will do it.  Thanks a lot!

drgrice1:
> Thank you very much. Are you talking about ShowAnswers2.pm? Maybe I should have worked directly on ShowAnswers.pm instead, so that one can more easily see the line-by-line diffs. If this is what the community prefers I am perfectly fine with taking out the new file and work on ShowAnswers.pm instead. Sorry about the trouble!

No, I am really talking about all of the perl files!  Again, I like the clean up.  One thing you can do to help with this is have an initial commit that is purely clean up.  Then put the significant changes into another commit.  Of course you have quite a few commits here, so that doesn't quite work.  I suspect there are some that are going to want you to clean up this pull request and put it squash all of that.

I am not quite sure if I like the addition of a ShowAnwers2.pm file.  It depends on how significant the changes are.
 
> OK, I will try to fix this. I think I really used only the most basic features of Javascript, so hopefully it wouldn't require too much correction, but I will do it. Thanks a lot!

At a quick glance, it looks like the main ES6 features you use are the "let" and "const" variable declarations.  You also used the array "includes" method.  There might be others.

pschan-gh:
> > Thank you very much. Are you talking about ShowAnswers2.pm? Maybe I should have worked directly on ShowAnswers.pm instead, so that one can more easily see the line-by-line diffs. If this is what the community prefers I am perfectly fine with taking out the new file and work on ShowAnswers.pm instead. Sorry about the trouble!
> 
> No, I am really talking about all of the perl files! Again, I like the clean up. One thing you can do to help with this is have an initial commit that is purely clean up. Then put the significant changes into another commit. Of course you have quite a few commits here, so that doesn't quite work. I suspect there are some that are going to want you to clean up this pull request and put it squash all of that.

The funny thing is, I honestly really didn't try to do any code cleanup, nor had any intention of doing so.  I think I just stumbled through the whole thing and it ended up looking like I cleaned up something.  I think the simplest thing for me to do now is to start a new branch based off of develop, and just make the changes again.  Should I close this PR then?  Once I've done the above I will start a new one.

> 
> I am not quite sure if I like the addition of a ShowAnwers2.pm file. It depends on how significant the changes are.

Yea, it might be a redundant thing to do after all.

> 
> > OK, I will try to fix this. I think I really used only the most basic features of Javascript, so hopefully it wouldn't require too much correction, but I will do it. Thanks a lot!
> 
> At a quick glance, it looks like the main ES6 features you use are the "let" and "const" variable declarations. You also used the array "includes" method. There might be others.

Got it, thanks! Is ES5 OK?

drgrice1:
You don't need to close the pull request and start a new one.  You can just force push to overwrite it.  A force push at this point should cause any problems.

Yeah, ES5 is fine.

pschan-gh:
OK, I have force-pushed onto develop-answer-log changed files which didn't have the previous "code cleanup".  (I found out that my atom editor by default removed all trailing whitespaces, so I disabled that feature. I had also subconsciously auto-indented a whole bunch of lines by sheer habit.)

In the Javascript part of the added code, all looping through of arrays and objects are now as basic as I know how to make it. Changed all 'const' and 'let' to 'var'. I hope I haven't missed anything still.

Thank you very much!

taniwallach:
@drgrice1 @pstaabp - @mgage has suggested last week seeing if this could be finished for WW 2.16. Does someone have time to look at it?

pschan-gh:
I have merged the most recent code from 2.16 into the develop-answer-log branch.

The difference between the develop-answer-log branch and WeBWorK-2.16 is now restricted to a lot fewer files than before.

The popup modal presenting the answer log in a table format is a bit ad-hoc (both in terms of the underlying code and its appearance).  Perhaps for the consideration of a release version it is more appropriate to make do with just the new Answer Log buttons on the side panel?

drgrice1:
If it helps, any concerns over es5 versus es6 are pretty much gone now.  We are no longer trying to support Internet Explorer.  Generally, the browsers that we want to still support should support es6.

pschan-gh:
OK, thanks! 

Do you think it's worthwhile to split my pull request into two: One for adding an answer log button to the side panel, and one for the popup modal table presentation? The two things are quite independent of each other.

OFF-TOPIC: it seems a lot of things regarding the presentation of various data (answer log, student progress etc) can be done on the client side, by dumping the state of the current page to the client via a JSON object. Maybe for something like a customized table-view format of the answer log, it could someday be left as some sort of elective 3rd-party "plugin" to the user.

taniwallach:
Sorry, we decided this is not going to make it into the 2.16 release now.
This should probably be retargeted to develop.

pschan-gh:
OK, thanks!

dlglin:
Closing due to inactivity.  Can be reopened if work resumes.

PR #1131, opened 2020-09-18
Correction+improvement to CSV output changes from PR 1085 (per assignment percent grades)

taniwallach:
https://github.com/openwebwork/webwork2/pull/1085 added per assignment percent grades to the CSV output, with the code in that PR always making this addition.

I discovered that the manner this was done was not consistent with the percent grade shown for assignments in the "Grades" pages generated by lib/WeBWorK/ContentGenerator/Grades.pm. That file did
```
          my $totalRightPercent = 100*wwRound(2,$total ? $totalRight/$total : 0);
```
so that the percent grade is always an integer.

https://github.com/openwebwork/webwork2/pull/1085 used 
```
          wwRound( 2, 100 * $score/ ($r_totals->[$problemValueRow]->[$j]) )
```
similar to what is done elsewhere, so the percent grade in the CVS file would include 2 decimal places.

This inconsistency led to some problems in post-WeBWorK grades processing, as student expected the displayed percent grade to be used as it was displayed, while the initial processing used the more accurate (but sometimes lower) grade with the 2 decimal places.

This PR modifies the per-assignment percent grade to match what "Grades" show.

It also adds a checkbox to control whether these additional columns are included in the CVS file. At present, the checkbox is automatically checked, so the code will default to including the additional columns, as was the case with the temporarily hard-wired behavior from https://github.com/openwebwork/webwork2/pull/1085 .

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm
@@ -59,7 +59,8 @@ sub initialize {
 	$self->{scoringFileName}=$scoringFileName;
 	
 	$self->{padFields}  = defined($r->param('padFields') ) ? 1 : 0; 
-	
+	$self->{includePercentEachSet} = defined($r->param('includePercentEachSet') ) ? 1 : 0;
+
 	if (defined $scoreSelected && @selected && $scoringFileNameOK) {
 
 		my @totals                 = ();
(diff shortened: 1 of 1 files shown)


taniwallach:
Note: I did not use `maketext` to allow translation of the text next to the new checkbox. Adding `maketext` requires making the additions to the `pot` and `po` files to avoid errors which usually break page loading, but working on changes to the translation files is quite tedious and time consuming.

Ideally it would be good to do that before merging this PR, but I don't have time to hack at the translation files now.

Since this is for instructor use only, I think it is not terrible to allow a delay in the translation support, and that we may find it convenient to merge the changes and handle the translation files work later. 

If this is done, I recommend we begin to collect a list of places where additional `maketext` calls are needed and to try to handle them all and all the changes to the translation files at once.

drgrice1:
My initial assessment is that this looks good.  I haven't had a chance to test it.  I will try to do that this weekend.

drgrice1 (approved):
This looks good.  Everything works as it should.  It is good to align the way scoring is done here as in the Grades.

There are some other issues with grades not being computed in the same way in various places, so more work needs to be done on this.

drgrice1:
I thought the policy was not to merge your own pull request?

taniwallach:
> I thought the policy was not to merge your own pull request?

Sorry - I thought it was not to merge them unapproved. This is an approved correction to code I wrote, so I figured one approval is sufficient and we should get it out of the way.

@mgage - any comment on what you think the policy should be?

drgrice1:
It isn't a big deal.  This is a minor change that was going to be merged anyway.   I think that the policy should in general be that a pull request be merged by someone else.  Perhaps an exception for approved minor pull requests is fine.

PR #1136, opened 2020-09-20
Fix debug statement in SubmitGrade.pm

sarunasb:
Fix invocation of local_escape_html() in debug statement.


Alex-Jordan:
This one was still based from `master`, and merging into `master`. I changed its target to `develop`, but that is not enough since it was based from `master`. My guess is that GitHub put you back on `master` when you thought you were still working off `develop` after creating #1135.

PR #1138, opened 2020-09-20
Fix debug statement in SubmitGrade.pm

sarunasb:
Fix invocation on local_escape_html() in debug statement.


taniwallach (approved):
This fixes the a minor bug discussed in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4770 and in my opinion is ready to be merged.

PR #1148, opened 2020-10-24
Drop user and key from URL when session_cookie session management is in use

taniwallach:
Drop the `user` and `key` values from the URLs the system generates, to make session hijacking less trivial, should students share URLS or they be passed unencrypted.

This is done only when `session_management_via` is set to `session_cookie` (which seems to be the default), as in that case - the session cookie has this data available.

Note: The `effectiveUser` is not stored in the WeBWorK cookie, so it cannot be removed from the URL at present, so currently remains in the URL. Adding it to the cookie should be possible, but there would be a need to change the cookie when starting to act as a different user, and when stopping to act as a different user.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -1867,8 +1867,16 @@ authentication.
 
 sub url_authen_args {
 	my ($self) = @_;
+	my $ce = $self->r->ce;
 
-	return $self->url_args("user", "effectiveUser", "key", "theme");
+	# When cookie based session management is in use, there should be no need
+	# to reveal the user and key in the URL. Putting it there makes session
+	# hijacking easier, in particular should a student share such a URL.
+	if ( $ce->{session_management_via} eq "session_cookie" ) {
+		return $self->url_args("effectiveUser", "theme");
+	} else {
+		return $self->url_args("user", "effectiveUser", "key", "theme");
+	}
 }
 
 =item url_state_args()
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks good.  I have thought for a while that these things should be removed from the URL.

It might be a good idea to add a the session_management_via to the appropriate config file for visibility purposes.

I would like to see the effectiveUser removed also.  I wonder if it would be better to set a separate cookie for that, rather than putting it into the same cookie with the user id and the session key.

drgrice1:
> > It might be a good idea to add a the session_management_via to the appropriate config file for visibility purposes.
> 
> It is mentioned in `defaults.config` which recommends modifying it in `localOverrides.conf` and `localOverrides.conf.dist` has a sample line to revert the setting to `key`.

I didn't see that it was mentioned in defaults.config.  I thought this was a new key that you added, but I see now that it is not.  So ignore this comment.

taniwallach:
> I would like to see the effectiveUser removed also. I wonder if it would be better to set a separate cookie for that, rather than putting it into the same cookie with the user id and the session key.

**Note:** I realized that even if we remove the `effectiveUser` from the URL's the system usually generates, we probably want to allow handling this key-value pair from an incoming URL, as the Feedback messages include it to make it easy to reach the page a student was at when sending a report.

drgrice1:
Is this a pull request we want to target for ww-2.16?

drdrew42:
> Is this a pull request we want to target for ww-2.16?

I believe so - it works as advertised. I'd still like to see effectiveUser removed from the URL, but @taniwallach is absolutely correct that we should always support it as query-string due its use in feedback emails

drdrew42 (approved):

drgrice1:
Yeah, my initial thoughts on the effectiveUser parameter were incorrect.

taniwallach:
@drgrice1, @drdrew42 - Should we mark it as for WW 2.16? Is it ready to merge?

drgrice1:
I will go ahead and merge this tomorrow unless there are objections.

Issue #1072, opened 2020-01-31
Check if we need to add the SameSite=None setting to cookies, if so probably needs to be a hotfix for versions in production use, and will require a recent CGI.pm


taniwallach:
Background: Google Chrome 80 (to be released soon) will be requiring the use of `SameSite=None` for cross-site cookies. 

Tim Payer ( http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4719 ) raised the issue of whether this scheduled change in Chrome will  cause problems with using WW over LTI.

### **Preliminary analysis:**
  1. I suspect the problem will not occur when LTI is used to open WeBWorK in a **new** window.
  2. It seems to me that when WeBWorK is used via LTI or the `html2xml` **and embedded in an iFrame**, the WeBWorK cookie will treated as a cross-site cookie, and **will probably** stop working with the upcoming  version of Chrome. 
      - I'm not absolutely certain that I am correct in my understanding that the WeBWorK cookies inside an iFrame will be treated as cross-site cookies, but that is what I understand from:
        - https://makandracards.com/makandra/71018-samesite-cookies
        - https://auth0.com/docs/sessions/concepts/cookie-attributes
        - https://github.com/apaleo/oss-traces/commit/f345c7846a6eb4e5c2a2923747388b6f6d3794dd

### **WeBWorK's cookies - a quick look + initial ideas:**

WeBWorK seems to set the cookies in
  - `lib/WeBWorK/Authen.pm`
  - `lib/WeBWorK/ContentGenerator/Logout.pm`
using a call to `$r->headers_out->set("Set-Cookie" => $cookie->as_string)` where the `$cookie` is a `WeBWorK::Cookie` object which (under `mod_perl` 2) is essentially `Apache2::Cookie`.  

`Apache2::Cookie` does not seem to be maintained, and does not seem to support the `SameSite` attribute. 

However, it seems that we can transition to using `CGI::Cookie` ( https://metacpan.org/pod/CGI::Cookie ) which does support the `SameSite` attribute.

`CGI::Cookie`
  - can take an `mod_perl` request object as an argument to the `bake()` method (and will then call `$r->headers_out->add('Set-Cookie' => $self->as_string);`
  - or we can call the `as_string()` method to create the string for the cookie, 
[...]

PR #1149, opened 2020-10-25
Cookie code refactor, Use CGI::Cookie and support samesite

taniwallach:
Refactor the Cookie code:

1. Use `CGI::Cookie` instead of `Apache2:Cookie`, as the new code needs support for the `samesite` attribute.
2. Added CGI::Cookie to bin/check_modules.pl and Dockerfile.
    - Note: The support for samesite dates to June 2019 in CGI::Cookie 4.45.
3. Remove obsolete, commented out, code using cookies from `lib/WeBWorK/ContentGenerator/Logout.pm`.
4. Drop the constant `COOKIE_LIFESPAN` and instead allow setting cookie lifespan using site / course environment configuration variables.
    - `$CookieLifeTime`  - for when cookie based session management **is** in use - default to 6 hours.
    - `$CookieLifeTime2` - for when cookie based session management is **not** in use, defaults to 30 days.
5. Allow setting value of cookies `samesite` and `secure` attribute using site / course environment configuration variable:
    - `$CookieSameSite`
    - `$CookieSecure`

---

This is motivated by the impending changes to Firefox and Chrome which ~~would~~ was expected to reject the webwork cookies in the future, as no `samesite` atttribute is set, so it ~~would be treated~~ was mistakenly expected to be treated as `none` without the `secure` attribute set. See: https://github.com/openwebwork/webwork2/issues/1072

**Update: As explained below, the change will be to treat cookies without an explicit `samesite` setting as `Lax` so they should not break for "same site" contexts. Thus, regular webwork sites are not likely to have problems in the near future.** However, the changes will allow explicit use of both cookie settings, if and when they may be needed.

---

Depending on when Firefox and Chrome begin to enforce the new policy, back-porting this as a hotfix to WW 2.15 may be needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,6 +82,8 @@ RUN git clone --single-branch --branch legacy-v2 --depth 1 https://github.com/ma
 # we need to change FROM before setting the ENV variables
 
 FROM ubuntu:18.04
+# Once ubuntu 20.04 is used, CGI.pm and CGI::Cookie will be new enough to
+# drop the cpanm install of CGI::Cookie which is needed to upgrade it.
 
 ENV WEBWORK_URL=/webwork2 \
     WEBWORK_ROOT_URL=http://localhost \
@@ -271,6 +273,8 @@ RUN echo "PATH=$PATH:$APP_ROOT/webwork2/bin" >> /root/.bashrc \
 # ==================================================================
 
 # Phase 6 - install additional Perl modules from CPAN (not packaged for Ubuntu or outdated in Ubuntu)
+
+# Ubuntu 18.04 has CGI.pm 4.38-1 which is too old to support the cookie samesite attribute added in CGI.pm 4.45 - so install CGI::Cookie here to get an upgraded version.
 
 RUN cpanm install Statistics::R::IO CGI::Cookie \
     && rm -fr ./cpanm /root/.cpanm /tmp/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -272,7 +272,7 @@ RUN echo "PATH=$PATH:$APP_ROOT/webwork2/bin" >> /root/.bashrc \
 
 # Phase 6 - install additional Perl modules from CPAN (not packaged for Ubuntu or outdated in Ubuntu)
 
-RUN cpanm install Statistics::R::IO \
+RUN cpanm install Statistics::R::IO CGI::Cookie \
     && rm -fr ./cpanm /root/.cpanm /tmp/*
 
 # Now installed from Ubuntu packages:
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -43,7 +43,7 @@ my @modulesList = qw(
 	Benchmark
 	Carp
 	CGI
-	CGI:Cookie
+	CGI::Cookie
 	Class::Accessor
 	Dancer
 	Dancer::Plugin::Database
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -32,7 +32,6 @@ my @apache1ModulesList = qw(
 
 my @apache2ModulesList = qw(
 	Apache2::Request
-	Apache2::Cookie
 	Apache2::ServerRec
 	Apache2::ServerUtil
 );
@@ -44,6 +43,7 @@ my @modulesList = qw(
 	Benchmark
 	Carp
 	CGI
+	CGI:Cookie
 	Class::Accessor
 	Dancer
 	Dancer::Plugin::Database
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -1,13 +1,12 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
-# $CVSHeader: webwork2/lib/WeBWorK/Authen.pm,v 1.63 2012/06/06 22:03:15 wheeler Exp $
-# 
+# Copyright &copy; 2000-2020 The WeBWorK Project, https://openwebworkorg.wordpress.com/
+#
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
 # Free Software Foundation; either version 2, or (at your option) any later
 # version, or (b) the "Artistic License" which comes with this package.
-# 
+#
 # This program is distributed in the hope that it will be useful, but WITHOUT
 # ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
 # FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
@@ -77,7 +76,7 @@ our $GENERIC_ERROR_MESSAGE = "";  # define in new
 ## WeBWorK-tr end modification 
 #####################
 
-use constant COOKIE_LIFESPAN => 60*60*24*30; # 30 days
+# Replaced constant COOKIE_LIFESPAN with a course environment setting which defaults to 30 days.
 #use constant GENERIC_ERROR_MESSAGE => "Invalid user ID or password.";
 
 
@@ -378,27 +377,6 @@ sub get_credentials {
 			#croak ("cookieUser = $cookieUser and paramUser = ". $r->param("user") . " are different.");
 			$self->maybe_kill_cookie; # use parameter "user" rather than cookie "user";
 		}
-# I don't understand this next segment.
-# If both key and $cookieKey exist then why not just ignore the cookieKey?
-
-# 		if (defined $cookieKey and defined $r->param("key")) {
-# 			$self -> {user_id} = $cookieUser;
-# 			$self -> {password} = $r->param("passwd");
-# 			$self -> {login_type} = "normal";
-# 			$self -> {credential_source} = "params_and_cookie";
-# 			$self -> {session_key} = $cookieKey;
-# 			$self -> {cookie_timestamp} = $cookieTimeStamp;
-# 			if ($cookieKey ne $r->param("key")) {
-# 				warn ("cookieKey = $cookieKey and param key = " . $r -> param("key") . " are different, perhaps"
-# 					 ." because you opened several windows for the same site and then backed up from a newer one to an older one."
-# 					 ."  Avoid doing so.");
-# 			$self -> {credential_source} = "conflicting_params_and_cookie";
-# 			}
-# 			debug("params and cookie user '", $self->{user_id}, "' credential_source = '", $self->{credential_source},
-# 				"' params and cookie session key = '", $self->{session_key}, "' cookie_timestamp '", $self->{cookieTimeStamp}, "'");
-# 			return 1;
-# 		} els
-
 # Use session key for verification
 # else   use cookieKey for verification
 # else    use cookie user name but use password provided by request.
@@ -866,42 +844,17 @@ sub fetchCookie {
 	
 	my $courseID = $urlpath->arg("courseID");
 	
-	# AP2 - Apache2::Cookie needs $r, Apache::Cookie doesn't
-    #my %cookies = WeBWorK::Cookie->fetch( MP2 ? $r : () );
-    #my $cookie = $cookies{"WeBWorKCourseAuthen.$courseID"};
-	
 	my $cookie = undef;
-	if (MP2) {
-		
-		my $jar = undef;
- 		eval {
-       			$jar = $r->jar; #table of cookies
-  		};
-  		if (ref $@ and $@->isa("APR::Request::Error") ) {
-			debug("Error parsing cookies, will use a partial result");
-     			$jar = $@->jar; # table of successfully parsed cookies
-  		};
-		if ($jar) {
-			$cookie = uri_unescape($jar->get("WeBWorKCourseAuthen.$courseID"));
-		};
-	} else {
-		my %cookies = WeBWorK::Cookie->fetch();
-		$cookie = $cookies{"WeBWorKCourseAuthen.$courseID"};
-		if ($cookie) {
-			debug("found a cookie for this course: '", $cookie->as_string, "'");
-			$cookie = $cookie->value;
-		}
-	}
+	my %cookies = WeBWorK::Cookie->fetch();
+	$cookie = $cookies{"WeBWorKCourseAuthen.$courseID"};
 
 	if ($cookie) {
-        #debug("found a cookie for this course: '", $cookie->as_string, "'");
-        #debug("cookie has this value: '", $cookie->value, "'");
-        #my ($userID, $key) = split "\t", $cookie->value;
-        debug("cookie has this value: '", $cookie, "'");
-        my ($userID, $key, $timestamp) = split "\t", $cookie;
+		#debug("found a cookie for this course: '", $cookie->as_string, "'");
+	        debug("cookie has this value: '", $cookie->value, "'");
+		my ($userID, $key, $timestamp) = split "\t", $cookie->value;
 		if (defined $userID and defined $key and $userID ne "" and $key ne "") {
 			debug("looks good, returning userID='$userID' key='$key'");
- 			return $userID, $key, $timestamp;
+			return $userID, $key, $timestamp;
 		} else {
 			debug("malformed cookie. returning nothing.");
 			return;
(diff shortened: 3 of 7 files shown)


taniwallach:
**Note**: The code depends on having CGI::Cookie version **4.45** (June 2019) or later. Many servers will need to update `CGI.pm`  to get a new enough version.

See: https://metacpan.org/changes/distribution/CGI

taniwallach:
Sample lines added to `course.conf`:
```
$CookieSameSite = "None";
$CookieSecure = 1;
$CookieLifeTime = "+7d";    # 7 day Cookie lifetime - for when cookie session management is in use
$CookieLifeTime2 = "+14d"; # 7 day Cookie lifetime - for when cookie session management is in use
```

The time settings can use the syntax of `CGI.pm` - search for `30 seconds from now` at https://metacpan.org/pod/CGI to see the various options.

bin/check_modules.pl
```diff
@@ -44,6 +43,7 @@
 	Benchmark
 	Carp
 	CGI
+	CGI:Cookie
```
drgrice1, line 46: This should have two colons I believe.  The check_modules.pl script gives a syntax error otherwise.

Dockerfile
```diff
@@ -272,7 +272,7 @@ RUN echo "PATH=$PATH:$APP_ROOT/webwork2/bin" >> /root/.bashrc \
 
 # Phase 6 - install additional Perl modules from CPAN (not packaged for Ubuntu or outdated in Ubuntu)
 
-RUN cpanm install Statistics::R::IO \
+RUN cpanm install Statistics::R::IO CGI::Cookie \
```
drgrice1, line 275: Shouldn't the docker setup install the libcgi-pm-perl ubuntu package instead of pulling from cpan?

lib/WeBWorK/Authen.pm
```diff
@@ -945,21 +907,25 @@ sub killCookie {
...
-	my $expires = time2str("%a, %d-%h-%Y %H:%M:%S %Z", time-60*60*24, "GMT");
-	my $cookie = WeBWorK::Cookie->new($r,
-		-name => "WeBWorKCourseAuthen.$courseID",
-		-value => "\t",
-		-expires => $expires,
-		-path => $ce->{webworkURLRoot},
-		-secure => 0,
+
+	my $sameSite = ( defined($ce->{CookieSameSite} ) ) ? $ce->{CookieSameSite} : "Strict" ;
+	my $secure   = ( defined($ce->{CookieSecure}   ) ) ? $ce->{CookieSecure}   : 0 ;
```
drgrice1, line 914: Should default values for this be defined in the appropriate configuration file (site.conf.dist I think).

drgrice1 (commented):
This is needed.  I am not sure how long it will be until Firefox and Chrome enforce this policy, but it is coming.

I think that defaults for $CookeSameSite, $CookieSecure, $CookieLifeTime, and $CookieLifeTime2 should be added to the appropriate configuration file.  Probably site.conf.dist.  If the defaults are only implemented as fall back values in the code, then they aren't very visible.  If the default values are listed (and documented) in the appropriate configuration file, then those configuring servers can find them.

Dockerfile
taniwallach: Ubuntu 18.04 has too old an version of the libcgi-pm-perl package (4.38) so does not have the samesite support needed.
When a suitable recent version of Ubuntu is used (ex. 20.04), this can be fixed.
I'm putting a note in `Dockerfile`, which will appear in an additional commit soon.

bin/check_modules.pl
taniwallach: Yes, sorry... Will be fixed in the new commit in a few minutes.

lib/WeBWorK/Authen.pm
taniwallach: Since `defaults.conf` has the setting `$session_management_via = "session_cookie";` and instructs that an override should be made in `localOverrides.conf` I think the best place is in `localOverrides.conf.dist`.

The additional commit puts it there.

It could also be added to `site.conf` or moved there.

taniwallach:
> This is needed. I am not sure how long it will be until Firefox and Chrome enforce this policy, but it is coming.

I realized that the problem is less urgent than I thought yesterday. It seems that the intention is to change to default to `Lax` (instead of `None`) which means the problems would mainly occur in embedded pages, and not for users who only use their webwork server directly.


> I think that defaults for $CookeSameSite, $CookieSecure, $CookieLifeTime, and $CookieLifeTime2 should be added to the appropriate configuration file. Probably site.conf.dist. If the defaults are only implemented as fall back values in the code, then they aren't very visible. If the default values are listed (and documented) in the appropriate configuration file, then those configuring servers can find them.

Since `defaults.conf` has the setting `$session_management_via = "session_cookie";` and instructs that an override should be made in `localOverrides.conf` I think the best place is in `localOverrides.conf.dist` For now I put sample lines there.

Dockerfile
drgrice1: I see.  I thought that might be the case.  The computer I tested this on is running Ubuntu 20.04 which has version 4.46 of the package, and as you said that is suitable.  So I didn't see the issue.

lib/WeBWorK/Authen.pm
drgrice1: localOverrides.conf.dist may be a more appropriate location for this. I was not sure.  All that matters is that these are in one of the conf files for transparency to those setting up servers.

taniwallach:
I think we want to get the secure cookie handling in for WW 2.16.

drgrice1:
I agree.

lib/WeBWorK/Authen.pm
```diff
@@ -916,27 +869,36 @@ sub sendCookie {
...
-	my $cookie = WeBWorK::Cookie->new($r,
-		-name    => "WeBWorKCourseAuthen.$courseID",
- 		-value   => "$userID\t$key\t$timestamp",
-		-path    => $ce->{webworkURLRoot},
-		-secure  => 0,
+
+	my $sameSite  = ( defined($ce->{CookieSameSite}  ) ) ? $ce->{CookieSameSite}  : "Strict" ; # Default to Strict
+	my $secure    = ( defined($ce->{CookieSecure}    ) ) ? $ce->{CookieSecure}    : 0 ;        # Default to NOT setting secure as it requires https
+	my $lifetime  = ( defined($ce->{CookieLifeTime}  ) ) ? $ce->{CookieLifeTime}  : "+6h" ;    # Default to 6 hours (for when session_management_via is session_cookie)
+	my $lifetime2 = ( defined($ce->{CookieLifeTime2} ) ) ? $ce->{CookieLifeTime2} : "+30d" ;   # Default to 30 days (for when session_management_via is NOT session_cookie)
```
drgrice1, line 880: In keeping with the practices that have been used (and I think these are good practices in this case), these settings should all be set to the defaults in defaults.config.  You added these comment out in localOverrides.conf.dist which is the correct thing to do, but the actual defaults should be in defaults.config.  Then change this to just read those values.  The point is to make these always defined, and simplify the messy logic here.

Furthermore, the defaults here do not match the defaults in localOverrides.conf.dist.

drgrice1 (changes requested):
This generally looks good, and works as it should.

There is one thing that might be nice, and that is to tie things in tighter between the $sessionKeyTimout and the $CookieLifeTime settings.  Probably there should only be one fo these for when managing the session via cookies.  It certainly should be the case that the $CookieLifeTime should be the greater of the two.

lib/WeBWorK/Authen.pm
taniwallach: In modifying `check_session` so that when
  1. `$ce->{session_management_via} eq "session_cookie"`
  2. `$ce->{CookieSecure}` is true, and
  3. `defined($self->{cookie_timestamp})`
then it will use 
```
$timestampValid = ( time <= $self->{cookie_timestamp} + $ce->{CookieLifeTime} );
```
by reenabling and modified the conditional block which had been commented out.

taniwallach:
@drgrice1  wrote:

> There is one thing that might be nice, and that is to tie things in tighter between the $sessionKeyTimout and the $CookieLifeTime settings. Probably there should only be one of these for when managing the session via cookies. It certainly should be the case that the $CookieLifeTime should be the greater of the two.

Implemented partially - see https://github.com/openwebwork/webwork2/pull/1149#discussion_r583057948 which probably should have been here instead.

lib/WeBWorK/Authen.pm
taniwallach: @drgrice1 
> ... these settings should all be set to the defaults in defaults.config.
> ... Then change this to just read those values. The point is to make these always defined, and simplify the messy logic here.

Done

lib/WeBWorK/Authen.pm
```diff
@@ -945,21 +905,25 @@ sub killCookie {
...
-	my $expires = time2str("%a, %d-%h-%Y %H:%M:%S %Z", time-60*60*24, "GMT");
-	my $cookie = WeBWorK::Cookie->new($r,
-		-name => "WeBWorKCourseAuthen.$courseID",
-		-value => "\t",
-		-expires => $expires,
-		-path => $ce->{webworkURLRoot},
-		-secure => 0,
+
+	my $sameSite = ( defined($ce->{CookieSameSite} ) ) ? $ce->{CookieSameSite} : "Strict" ;
+	my $secure   = ( defined($ce->{CookieSecure}   ) ) ? $ce->{CookieSecure}   : 0 ;
```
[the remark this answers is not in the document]
taniwallach: Done. Sorry I missed the second place.

drgrice1 (approved):
This looks good.

Could someone else please take a look.

This is an important pull request that will be needed soon.

taniwallach:
> This looks good.
> 
> Could someone else please take a look.
> 
> This is an important pull request that will be needed soon.

I added requests that @pstaabp and @drdrew42 review, as I think @mgage is too busy, and @Alex-Jordan is planning to work on testing the LTI patches, so I'ld like to free him from needing to test this one.

drgrice1:
I am bumping the priority on this (assuming priority 1 is a higher priority).  It won't be long until browsers start enforcing this beyond just a console warning.  There is a good chance this will happen before the next release of webwork, so lets get this in now.

drdrew42:
Can I get an elaboration on how I can tell whether or not this PR is working? I don't get any console logs about cookies with or without this PR...

drgrice1:
One way is to use Firefox.  Firefox shows warnings without this pull request,  but not with.

So I decided to give the information on this a closer read.  I think that Firefox's console warning is a bit misleading, and this may not be as serious as I though.

As I understand it Google Chrome is already using the new standards by default, but Firefox is not yet.  You can enable the new standards in Firefox to test things.

The new standards are that if a cookie is sent with `SameSite` not set then the default value that will be used is `Lax`.  Furthermore, if `SameSite` is set to `None`, then `Secure` must be `true` or the cookie will be rejected.

So basically, at this point the webwork cookies are being used with `SameSite` set to `Lax` on Google Chrome.  Those cookies aren't rejected, but they also aren't secure.  Firefox is still using the old standards, and sets `SameSite` to `None`.  Those cookies are also not secure, and they aren't rejected because again Firefox is still using the old standards.

Firefox's  is using the current policy to set the default of `SameSite` to `None`, but is warning about that not having `Secure` set to true as per the future policy.  This warning does not take into account that the future policy also says that the default for `SameSite` will be `Lax`, for which `Secure` does not need to be true.

So the upshot of this is that things will still continue to work for webwork with or without this pull request.  There is just the issue of whether or not we want the cookies to be secure.  So I am not sure how serious this is.

I just looked back at issue #1072, and @taniwallach said some of this in his last post there.

drgrice1:
Back to priority 2.  I see @taniwallach (or someone) classified the priorities.  I don't think there really was any doubt though.

pstaabp:
I think I need a bit more to test this.  It appears that this may only be a problem using LTI.  Is this true.  I don't have that setup currently.   

Is there a setting in FF or chrome to turn on the site-same cookie for the standard WW setup to check this?

drgrice1:
This affects all cookie based authentication.  Not just LTI.  Currently cookie based authentication is the default for webwork.

pstaabp (approved):
Not sure this is full tested because I can't see the effect this has on Chrome, but overall, this appears to future-proof us.

pstaabp:
Overall, everything seems to work,  I did verify what @drgrice1 sees on Firefox.  

I've tried setting the flags for SameSite in Chrome, but am not getting any warnings on the develop branch, which is what I expect. According to https://www.chromium.org/updates/same-site/faq there is site checking for the right flags and the current chrome (89) seems to have all the correct flags set.   Not sure what is going on.

taniwallach:
> I've tried setting the flags for SameSite in Chrome, but am not getting any warnings on the develop branch, 

From what I understand from https://www.chromium.org/updates/same-site/faq - warning will be issued by Chrome only for cross-site requests: "Chrome is displaying warnings in the Console in DevTools which highlight each **cross-site** request where cookies would be affected" (emphasis is mine). Firefox on the other hand seems to issue warnings far more broadly, even when they probably would not really have problems.

drgrice1:
We have two reviews on this.  I think it is ready to be merged in.  We will need to make sure that the installation documentation notes the a new Ubuntu package is needed (if on Ubuntu 20.04) or a new perl module needs to be installed via cpan or cpanm (if on Ubuntu 18.04).

lib/WeBWorK/Authen.pm
```diff
@@ -809,18 +787,16 @@ sub check_session {
...
-# first part of if clause is disabled for now until we figure out long term fix for using cookies
-# safely (see pull request #576)   This means that the database key time is always being used
-# even when in "session_cookie" mode
-#	if ($ce -> {session_management_via} eq "session_cookie" and defined($self->{cookie_timestamp})) {
-#		$timestampValid = (time <= $self -> {cookie_timestamp} + $ce->{sessionKeyTimeout});
-#	} else {
-		$timestampValid = (time <= $Key->timestamp()+$ce->{sessionKeyTimeout});
+	if ( $ce->{session_management_via} eq "session_cookie"
+		&& $ce->{CookieSecure} && defined($self->{cookie_timestamp}) ) {
+		$timestampValid = ( time <= $self->{cookie_timestamp} + $ce->{CookieLifeTime} );
```
drgrice1, line 792: I missed this before.  This gives an error as CookieLifeTime is not numeric.

drgrice1:
>        * Note: The support for samesite dates to June 2019 in CGI::Cookie 4.45.

Looking closely at this, it seems that CGI::Cookie had the `samesite` setting before version 4.45.  See https://metacpan.org/pod/release/LEEJO/CGI-4.38/lib/CGI/Cookie.pm.  Version 4.38 had the `samesite` setting but only allowed values of `strict` or `lax`, and not `none`.  I also tested this on my server running Ubuntu 18.04 and the Ubuntu libcgi-pm-perl package (which contains CGI::Cookie version 4.38), and this worked fine.  I was using the default `samesite` value of `strict`.

I installed the cpan version 4.45 to be safe, but there may be no need for concern if the Ubuntu package on 18.04 is used.

taniwallach:
> > ```
> >    * Note: The support for samesite dates to June 2019 in CGI::Cookie 4.45.
> > ```
> 
> Looking closely at this, it seems that CGI::Cookie had the `samesite` setting before version 4.45. See https://metacpan.org/pod/release/LEEJO/CGI-4.38/lib/CGI/Cookie.pm. Version 4.38 had the `samesite` setting but only allowed values of `strict` or `lax`, and not `none`. I also tested this on my server running Ubuntu 18.04 and the Ubuntu libcgi-pm-perl package (which contains CGI::Cookie version 4.38), and this worked fine. I was using the default `samesite` value of `strict`.
> 
> I installed the cpan version 4.45 to be safe, but there may be no need for concern if the Ubuntu package on 18.04 is used.

Good point, but it depends on whether a site's WW content will be embedded in a cross-origin setting.

Long response below.

Bottom line - I think it suffices to explain in the release notes under what circumstances a server is likely to need to be able to enable `SameSite=None` and inform administrators of the need for a more modern version of `CGI::Cookie` when that is needed.

---

I can trigger (with some work) a test where browser same-site cookie enforcement seems to require the `SameSite=None` setting - see below. (`Lax` by default triggers "bad" behavior for an end user.)

Regular" webwork servers use only first-party cookies, so should at the most trigger warnings (mainly in FireFox) at present, if they don't make an explicit setting.

That means that the version in Ubuntu 18.04 suffices to set an explicit value of `lax` (the new default or even `strict`), and that should suffice to bypass the warnings if they bother someone. For many use cases - that is more than enough.

I suspect that the sites which are likely to have a real possible need of this new code are precisely those which embed WeBWorK content in some external setting where the issue of a cross-origin request can be an issue. It seems that in such a setting - the WeBWorK cookie might sometime be needed in a cross-origin context, so would need to have the `SameSite=None` setting .


---

All the testing below was done without any special settings for SameSite on the WeBWorK servers, to check behavior with the current WW default settings. 

I did the testing using Chrome Version 89.0.4389.90 (Official Build) (64-bit) on Linux and **with a special Chrome flag enabled**:
  - `chrome://flags/#same-site-by-default-cookies`
  - SameSite by default cookies
  - Treat cookies that don't specify a SameSite attribute as if they were SameSite=Lax. Sites must specify SameSite=None in order to enable third-party usage. – Mac, Windows, Linux, Chrome OS, Android

The flag I enabled is being dropped in Chrome 91, as the new behavior will be forced in that version:
  - https://www.chromium.org/updates/same-site

That should be forcing my currently installed browser to enforce what will be enforced by default in version 91 (release scheduled for late May 2021 - https://chromiumdash.appspot.com/schedule).

**When I disable that flag, the "embedded" pages which fail to work properly do work.**

I also do not hit actual problems yet with FireFox version 78.8.0esr (64-bit).

Other useful references:
  - https://www.chromium.org/updates/same-site/faq
  - https://www.chromium.org/updates/same-site
  - https://samesite-sandbox.glitch.me/

---

The problems occurred (on Chrome with the flag mentioned above enables) when I tested a Moodle LTI link from `moodle.technion.ac.il` to a course on `webwork.iucc.ac.il` where the target page is **embedded** (in an iframe), the initial connection works, **but navigation using the menu kicks me out to the login page.**

It seems that the cookie sent at the time of the initial connection were not stored (do not appear where cookies would be shown in the developer tools) so when a navigation request is made it is sent without the cookie (none is listed in "Request Headers"). 

However, a copy of the same LTI link which was set to open in a new windows - does store a cookie, and internal course navigation does work, as a new windows creates a "first party" context for the original cookie sent by the WW server.

---

From what I see so far, in many common situations, embedding a given institutions locally hosted WW content in a locally hosted Moodle page via "LTI as embedded" **works fine**. This is apparently related to how browsers define "cross-domain" - see:
  - https://web.dev/same-site-same-origin/
  - https://publicsuffix.org/list/public_suffix_list.dat

The fact that SameSite looks at `eTLD+1` apparently means in my "local" test case that `moodle.technion.ac.il` and `webwork.technion.ac.il` are both in `technion.ac.il` domain for the purposes of SameSite.

PR #1150, opened 2020-10-27
LTI account creation patches + discussion about aborted connections being reported by the database

taniwallach:
**Update:** The discussion thread gets into matters related to aborted connections being reported in the database logs.
See the forum post:  https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4958

---

I ran into an issue where a few LTI created accounts did not have a permission level set. Those accounts were reporting when login attempts were made a `WeBWorK error` something like below (pulled out of the Apache `error.log`):
```
Unable to retrieve your permissions, perhaps due to a collision between your request and that of another user (or possibly an unfinished request of yours). Please press the BACK button on your browser and try again. at
/opt/webwork/webwork2/lib/WeBWorK/Authen.pm line 511.\n * in Carp::croak called at line 168 of
/opt/webwork/webwork2/lib/WeBWorK/Authz.pm\n * in WeBWorK::Authz::setCachedUser called at line 220 of
/opt/webwork/webwork2/lib/WeBWorK/Authz.pm\n * in WeBWorK::Authz::hasPermissions called at line 511 of
/opt/webwork/webwork2/lib/WeBWorK/Authen.pm\n * in WeBWorK::Authen::check_user called at line 329 of
/opt/webwork/webwork2/lib/WeBWorK/Authen.pm\n * in WeBWorK::Authen::do_verify called at line 216 of 
/opt/webwork/webwork2/lib/WeBWorK/Authen.pm\n * in WeBWorK::Authen::verify called at line 160 of 
/opt/webwork/webwork2/lib/WeBWorK/Authen.pm\n * in WeBWorK::Authen::call_next_authen_method called at line 213 of 
/opt/webwork/webwork2/lib/WeBWorK/Authen.pm\n * in WeBWorK::Authen::verify called at line 388 of 
/opt/webwork/webwork2/lib/WeBWorK.pm, referer: https://webwork.technion.ac.il/webwork2/104016_2021a
```

When I viewed the problematic accounts in the class list editor, the system automatically set a permission level, and reported `added missing permission level for user` (triggered in `lib/WeBWorK/ContentGenerator/Instructor/UserList2.pm` and `lib/WeBWorK/ContentGenerator/Instructor/UserList.pm`). That resolved the problem.

---

It is not totally clear how an account was created in the few such cases without any permission level set.

I suspected that these students may have had a non-standard LMS role at the time they first accessed the LTI link. The most likely cause of such a problem would be missing LMS roles which are not defined in `%LMSrolesToWeBWorKroles` a conversion table defined as a hash in `conf/authen_LTI.conf` (and possibly modified elsewhere).

However, I tried to recreate such an error by modifying the conversion table for an old course, and got a different error when attempting to have LTI create an account, and no account was created (apparently due to the `croak`). Thus, it is not clear that a missing entry in the conversion table would lead to a missing permission level.

```
WeBWorK error
An error occured while processing your request. For help, please send mail to this site's webmaster (tREMOVED), including all of the following information as well as what what you were doing when the error occured.

Tue Oct 27 13:06:55 2020

Warning messages
Use of uninitialized value in hash element at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 618.
Error messages
Cannot find a WeBWorK role that corresponds to the LMS role of Instructor. at /opt/webwork/webwork2/lib/WeBWorK.pm line 388.
Call stack

in Carp::croak called at line 618 of /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm
in WeBWorK::Authen::LTIAdvanced::create_user called at line 547 of /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm
in WeBWorK::Authen::LTIAdvanced::authenticate called at line 343 of /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm
in WeBWorK::Authen::LTIAdvanced::verify_normal_user called at line 335 of /opt/webwork/webwork2/lib/WeBWorK/Authen.pm
in WeBWorK::Authen::do_verify called at line 216 of /opt/webwork/webwork2/lib/WeBWorK/Authen.pm
in WeBWorK::Authen::verify called at line 388 of /opt/webwork/webwork2/lib/WeBWorK.pm
```

The `Use of uninitialized value in hash element at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 618.` message is addressed in the second commit of the PR, by doing the `defined` test in 2 stages.

---

In any case, I am proposing to add a fallback setting for `$LTI_webwork_permissionLevel` as `student`  level so if the test
```
     defined( $ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}} )
```
fails during the account creation process, the code will fall back to creating an account with `student` level access. To this end the `croak` is replaced by `warn`. With the patch and an intentionally damaged %LMSrolesToWeBWorKroles set in a course, account creation succeeded and a student account was created.

It is not clear if this fix will actually address the original problem of preventing an account from being created with an undefined permission level, which prevents logins from working, as I could not trigger that behavior (before applying the patch) but it will certainly reduce account creation issues in other cases.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -501,13 +501,16 @@ sub create_user {
     }
   
   my $nr = scalar(@LTIroles);
+
+  my $LTI_webwork_permissionLevel;
   if (! defined($ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}})) {
     croak("Cannot find a WeBWorK role that corresponds to the LMS role of "
-	  . $LTIroles[0] .".");
+	  . $LTIroles[0] ." so will create a student level account.");
+    $LTI_webwork_permissionLevel = $ce->{userRoles}->{student};
+  } else {
+    $LTI_webwork_permissionLevel = $ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}};
   }
-  
-  my $LTI_webwork_permissionLevel 
-    = $ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}};
+
   if ($nr > 1) {
     for (my $j =1; $j < $nr; $j++) {
       my $wwRole = $ce->{LMSrolesToWeBWorKroles}->{$LTIroles[$j]};
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -503,8 +503,9 @@ sub create_user {
   my $nr = scalar(@LTIroles);
 
   my $LTI_webwork_permissionLevel;
-  if (! defined($ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}})) {
-    croak("Cannot find a WeBWorK role that corresponds to the LMS role of "
+  if ( ! defined( $ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]} ) ||
+       ! defined( $ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}} ) ) {
+    warn("Cannot find a WeBWorK role that corresponds to the LMS role of "
 	  . $LTIroles[0] ." so will create a student level account.");
     $LTI_webwork_permissionLevel = $ce->{userRoles}->{student};
   } else {
--- a/lib/WeBWorK/Authen/LTIBasic.pm
+++ b/lib/WeBWorK/Authen/LTIBasic.pm
@@ -517,12 +517,16 @@ sub authenticate
 			}
 			
 			my $nr = scalar(@LTIroles);
-			if (! defined($ce -> {userRoles} -> {$ce -> {LMSrolesToWeBWorKroles} -> {$LTIroles[0]}})) {
+
+			my $LTI_webwork_permissionLevel;
+			if (! defined($ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}})) {
 				croak("Cannot find a WeBWorK role that corresponds to the LMS role of "
-						. $LTIroles[0] .".");
+						. $LTIroles[0] ." so will create a student level account.");
+				$LTI_webwork_permissionLevel = $ce->{userRoles}->{student};
+			} else {
+				$LTI_webwork_permissionLevel = $ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}};
 			}
-			my $LTI_webwork_permissionLevel 
-				= $ce -> {userRoles} -> {$ce -> {LMSrolesToWeBWorKroles} -> {$LTIroles[0]}};
+
 			if ($nr > 1) {
 				for (my $j =1; $j < $nr; $j++) {
 					my $wwRole = $ce -> {LMSrolesToWeBWorKroles} -> {$LTIroles[$j]};
--- a/lib/WeBWorK/Authen/LTIBasic.pm
+++ b/lib/WeBWorK/Authen/LTIBasic.pm
@@ -519,8 +519,9 @@ sub authenticate
 			my $nr = scalar(@LTIroles);
 
 			my $LTI_webwork_permissionLevel;
-			if (! defined($ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}})) {
-				croak("Cannot find a WeBWorK role that corresponds to the LMS role of "
+			if ( ! defined( $ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]} ) ||
+			     ! defined( $ce->{userRoles}->{$ce->{LMSrolesToWeBWorKroles}->{$LTIroles[0]}} ) ) {
+				warn("Cannot find a WeBWorK role that corresponds to the LMS role of "
 						. $LTIroles[0] ." so will create a student level account.");
 				$LTI_webwork_permissionLevel = $ce->{userRoles}->{student};
 			} else {


taniwallach:
Note that  Sean Fitzpatrick reported [in the WW forrms](urlhttps://webwork.maa.org/moodle/mod/forum/discuss.php?d=4958) having the same issue as originally motivated me to look into possible causes of a missing permission level for an LTI created account.

@dlglin reported also having similar problems after an attempt to change to a different database "server" and that reverting to the prior database server seems to have resolved the problem. He raised the suspicion that the connection to the database dropped or something else caused a transaction to be lost.

---

In my case, the problems were seen with mariadb 10.4 as a Docker container, using a somewhat old image.
```
docker image inspect 99c1098d5884
[
    {
        "Id": "sha256:99c1098d5884b4a2f6aa7cef1456d99fa80351f62a060778ad1b8a5a138ae301",
        "RepoTags": [],
        "RepoDigests": [
            "mariadb@sha256:6f1faac314874361a45fc946c37ba5c597ecba647666156bde783ef088d1c184"
        ],
        "Parent": "",
        "Comment": "",
        "Created": "2019-08-15T09:24:33.210821673Z",
        "Container": "c7266c83d85ccddb71bb26c0a3461aa24ea52a810116e02a896c41e4071bee35",
        "ContainerConfig": {
            "Hostname": "c7266c83d85c",
            "Domainname": "",
            "User": "",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "ExposedPorts": {
                "3306/tcp": {}
            },
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "GOSU_VERSION=1.10",
                "GPG_KEYS=177F4010FE56CA3336300305F1656F24C74CD1D8",
                "MARIADB_MAJOR=10.4",
                "MARIADB_VERSION=1:10.4.7+maria~bionic"
            ],
            "Cmd": [
                "/bin/sh",
                "-c",
                "#(nop) ",
                "CMD [\"mysqld\"]"
            ],
            "ArgsEscaped": true,
            "Image": "sha256:d02b39f5dc8b20d1d7d46095460a57f66cb8a7178ec2ebddd6f59b433b154793",
            "Volumes": {
                "/var/lib/mysql": {}
            },
            "WorkingDir": "",
            "Entrypoint": [
                "docker-entrypoint.sh"
            ],
            "OnBuild": null,
            "Labels": {}
        },
        "DockerVersion": "18.06.1-ce",
        "Author": "",
        "Config": {
            "Hostname": "",
            "Domainname": "",
            "User": "",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "ExposedPorts": {
                "3306/tcp": {}
            },
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "GOSU_VERSION=1.10",
                "GPG_KEYS=177F4010FE56CA3336300305F1656F24C74CD1D8",
                "MARIADB_MAJOR=10.4",
                "MARIADB_VERSION=1:10.4.7+maria~bionic"
            ],
            "Cmd": [
                "mysqld"
            ],
            "ArgsEscaped": true,
            "Image": "sha256:d02b39f5dc8b20d1d7d46095460a57f66cb8a7178ec2ebddd6f59b433b154793",
            "Volumes": {
                "/var/lib/mysql": {}
            },
            "WorkingDir": "",
            "Entrypoint": [
                "docker-entrypoint.sh"
            ],
            "OnBuild": null,
            "Labels": null
        },
        "Architecture": "amd64",
        "Os": "linux",
        "Size": 355289537,
        "VirtualSize": 355289537,
        "GraphDriver": {
            "Data": {
                "LowerDir": "/var/lib/docker/overlay2/143f1b19a0d2f83070e2166afc61c618779f5fb993d4d15cca41b42b1e9b259a/diff:/var/lib/docker/overlay2/e23d014d64664e959d4043b704f4c313411ad13516ff8706b42efbcc9f200496/diff:/var/lib/docker/overlay2/fb3c742b30dfe7ee82cda97536bdaf26714fc6bce159c7527114c5ba8980a264/diff:/var/lib/docker/overlay2/070e51a70240c77f152fe292ad3f3ea3ff70c170b02ec22cb848399eeeef18d1/diff:/var/lib/docker/overlay2/5847bcd06956fc343479011eefe7530b5d84dc5f1f8984a8624fc7ef7442ff0b/diff:/var/lib/docker/overlay2/d4e368d98fe75ddaed92ea2a4934bbe1667055ea78c373991f62a9c0511123fd/diff:/var/lib/docker/overlay2/3e9c6da8cff84d13ad66c692fc0e9b0b06d25804d229f3be107a05bebf7d9873/diff:/var/lib/docker/overlay2/d1d3b39ca228da87b18da2bcb73d647bdafd26039830dd1670d949e5644fb21f/diff:/var/lib/docker/overlay2/4a1b2309b4a048807b26f0c1ba8983d71126d579a8d594a8f178cf908ebdedf2/diff:/var/lib/docker/overlay2/cea28f4d67552a1cb1d5b039d47c1279ffc3fcb769bb6025a4a6eafa7fc97f02/diff:/var/lib/docker/overlay2/09ff1dbf89ebf70af1d46dbbf6945180b97215970af7827b61246713a7f81f3d/diff:/var/lib/docker/overlay2/580a89fa60591ba1f3fa5e9e3289994f977d38575d547f608ec39ba5102df50b/diff:/var/lib/docker/overlay2/fe47bb63c568fc235269f2f0d5ed6fdd654f73daf583dba9b90cbdd4299a6aae/diff",
                "MergedDir": "/var/lib/docker/overlay2/ca02b1decd13246aed2a29619fbaa0a415b5aa306365606234600dc989845653/merged",
                "UpperDir": "/var/lib/docker/overlay2/ca02b1decd13246aed2a29619fbaa0a415b5aa306365606234600dc989845653/diff",
                "WorkDir": "/var/lib/docker/overlay2/ca02b1decd13246aed2a29619fbaa0a415b5aa306365606234600dc989845653/work"
            },
            "Name": "overlay2"
        },
        "RootFS": {
            "Type": "layers",
            "Layers": [
                "sha256:6cebf3abed5fac58d2e792ce8461454e92c245d5312c42118f02e231a73b317f",
                "sha256:f7eae43028b334123c3a1d778f7bdf9783bbe651c8b15371df0120fd13ec35c5",
                "sha256:7beb13bce073c21c9ee608acb13c7e851845245dc76ce81b418fdf580c45076b",
                "sha256:122be11ab4a29e554786b4a1ec4764dd55656b59d6228a0a3de78eaf5c1f226c",
                "sha256:87e027ff77c581dc4c25282e20ea36702a72dfba80d12dc83c1bc70ca84243e2",
                "sha256:f3859e87895a1a14c9c9f2cfa8f853a20f8fae0b95809c20a510357be0bbf1c1",
                "sha256:b3be41672ebed12d8fff07042be0755509f43ea9a4e3e86e5a52c0ff42617fae",
                "sha256:9f2e7e9024dbe78b622756f27e44a4df03d7f2451bb4f87de8ee7ba7f635fff3",
                "sha256:f2236a2d3fc3ca76dd38e30a1792419fbf0edcbfabe0b8cf2cb5bfabc76b812d",
                "sha256:8d7272cae9b44ea3e8132d0e75efeb66556fe86ac9404cf928c2f1dece1dd392",
                "sha256:8195ca18cca22107296176cf9a4e6e6feb90e1a401816453fb893879f6da8a8d",
                "sha256:ed3ae67e8c8cb9f926648070305cc4960d6a6eabc8523310898d5478e0eb4134",
                "sha256:c9c4c8c0a0f16b68fee5858f197c179e74d18ba90a644dd72f5dd90fa6f7df4a",
                "sha256:97f2532be6cfb59a467f47c7b0acfb3e9fb7abcb687fe624061c4bf938495ade"
            ]
        },
        "Metadata": {
            "LastTagTime": "0001-01-01T00:00:00Z"
        }
    }
]

```

dlglin:
@taniwallach we were getting students without permission levels due to aborted DB connections.  Can you check the error log for mariadb and see if there are any errors due to aborted connections?

taniwallach:
> @taniwallach we were getting students without permission levels due to aborted DB connections. Can you check the error log for mariadb and see if there are any errors due to aborted connections?

There are certainly many aborted connection. The SQL status variables give a count. However, it could be that WW is just sloppy about closing connections. The extent of the problems encountered does not seem close to the number of aborted connections.

```
mysql> select now(); SHOW GLOBAL VARIABLES LIKE 'wait_timeout'; show global status  where ( Variable_name like 'Abort%' ) ^H or ( Variable_name = 'Uptime') or ( Variable_name = 'Connections' ); exit;
+---------------------+
| now()               |
+---------------------+
| 2020-11-07 16:47:59 |
+---------------------+
1 row in set (0.00 sec)

+---------------+-------+
| Variable_name | Value |
+---------------+-------+
| wait_timeout  | 600   |
+---------------+-------+
1 row in set (0.00 sec)

+--------------------------+--------+
| Variable_name            | Value  |
+--------------------------+--------+
| Aborted_clients          | 80220  |
| Aborted_connects         | 4      |
| Aborted_connects_preauth | 1      |
| Connections              | 81030  |
| Uptime                   | 967612 |
+--------------------------+--------+
5 rows in set (0.00 sec)
```

It turns out that the mariadb container logs to Docker's log handler. So to see the logs one would use `docker ps` to find the `CONTAINER ID` and then `docker logs CONTAINER_ID`. I found that there were very many Warning log records like
```
2020-10-27 12:19:19 19 [Warning] Aborted connection 19 to db: 'webwork' user: 'webworkWrite' host: '192.168.48.4' (Got an error reading communication packets)
```
and
```
2020-10-27 12:36:50 187 [Warning] Aborted connection 187 to db: 'webwork' user: 'webworkWrite' host: '192.168.48.4' (Got timeout reading communication packets)
```

I dumped the logs from docker into a file, so I could look more carefully and count things.
```
tani@lxtani:~/Current/2020-11-07$ grep -c "error reading communication packets" mariadb.log 
62064
tani@lxtani:~/Current/2020-11-07$ grep -c "timeout reading communication packets" mariadb.log
18166
```

There are many discussions about there being many such warning messages.
  - https://dba.stackexchange.com/questions/19135/mysql-error-reading-communication-packets
  - https://www.percona.com/blog/2016/05/16/mysql-got-an-error-reading-communication-packet-errors/
  - https://forum.iredmail.org/topic17313-mysqlmariadb-got-an-error-reading-communication-packets.html

taniwallach:
I tried using `tcpdump` to see if I see any abnormal activity in the communications between my WW server container and the MariaDB 10.4 DB container. The interface listed below was that of the Docker bridge.
```
sudo tcpdump -s 65535 -x -nn -q -tttt --interface=br-20959f22957d -c 3000 port 3306 -w /tmp/mysql.tcp
```
The data all looked fine, although WireShark showed the funky bridge packet handing as re-transmissions.

---

**It seems that the problem may be an issue with the Perl DBD driver we are using... and it seems that using the newer `DBD::MariaDB` driver instead of the old `DBD::mysql` may be a fix for this problem.**

Details below:

---

There is a mention of many aborted connections in https://github.com/gooddata/DBD-MariaDB/issues/151#issuecomment-679077264 after an upgrade from `10.3.22-MariaDB-0+deb10u1` to `10.3.23-MariaDB-0+deb10u1` It is not clear if that is related, but that new `DBD` driver seems to have forked from the older mysql `DBD` driver due to Unicode support issues (and more).

---

As a result, I decided to try out the "new" Perl DBD driver for MariaDB and mySQL: 
  - https://metacpan.org/pod/DBD::MariaDB
  - https://metacpan.org/changes/distribution/DBD-MariaDB
  - https://github.com/gooddata/DBD-MariaDB
which was forked off the older DBD::mysql
  -  https://metacpan.org/pod/DBD::mysql

The stated reason for the fork was "The goal of our effort was to fix MariaDB compatibility, Perl Unicode support and security related bugs." This new driver is packaged for Ubuntu 20.04, but will require some changes to use it in WW (ex. it always uses utf8, so does not recognize the special settings used to force utf8 in the old driver). 

---

The new Perl module is available as a package for Ubuntu 20.04, but not for 18.04. Last night, I managed to get my development server to use that DBD module and Ubuntu 20.04. It has far less load, and in the past 18 hours only showed aborted connections of the "timeout" type and none "error reading communication packets". Those are also more infrequent, and I only see a single clump of 11 warnings - all a single 5 minute period. They all look like
```
2020-11-09  5:21:50 16 [Warning] Aborted connection 16 to db: 'webwork' user: 'webworkWrite' host: '172.22.0.4' (Got timeout reading communication packets)
```

That seems to suggest that there may be some problem with how the older `DBD::mysql` driver interacts either with Maria DB 10.4 or with some sort of Unicode data.

I set up my main production server to use the new `DBD::MariaDB` driver to see what happens there - as I was seeing aborted connections due to "error reading communication packets" every few minutes. It is up 4 hours so far with no warning issues.

**Note:** I also made some changes to the `ulimits` for the number of allowed open files, but that did not seem to help before I swapped drivers.

---
Modifications needed to work with the `DBD::MariaDB` driver:

1. `/opt/webwork/webwork2/lib/WeBWorK/DB/Driver/SQL.pm`
    - Comment out the lines for `mysql_enable_utf8mb4` and `mysql_enable_utf8` as the new driver always uses UTF-8 and does not recognize those parameters.
2. Other places where the settings are used will also need to be fixed, but are less critical for student's use of the system.
3. The database DSN is somewhat different and uses a different format:
    - From old driver I had: `DBI:mysql:webwork:db:3306`
    - For the new driver I have: `DBI:MariaDB:database=webwork;host=db;port=3306`

dlglin:
I have switched my test server to DBD::MariaDB, and right now we're still seeing the "Aborted connection" errors.  I'll report back if we discover anything else about this.

taniwallach:
> I have switched my test server to DBD::MariaDB, and right now we're still seeing the "Aborted connection" errors. I'll report back if we discover anything else about this.

In several weeks of use in production, I have not seen any more aborted connections due to `(Got an error reading communication packets)`.

I still see some "rare" aborted connections, but only those for `(Got timeout reading communication packets)` which occur once the "session" timeout is reached.

drgrice1 (approved):
I think that this pull request is okay, but I am not entirely sure about it.  Looking at the discussion, it seems that the actual problem is not entirely understood, and that this pull request implements a work around.

I guess it will be okay for now, but we should look deeper into the issue with database connections being dropped.

taniwallach:
> I think that this pull request is okay, but I am not entirely sure about it. Looking at the discussion, it seems that the actual problem is not entirely understood, and that this pull request implements a work around.
>
> I guess it will be okay for now, but we should look deeper into the issue with database connections being dropped.

The PR is an attempt to make sure that LTI will at least create an account with student level privileges, when the configured conversion table for LMS role to WeBWorK privilege level fails to provide a result. It is not likely to cause problems beyond possibly creating student mode accounts for special classes of LMS authenticated users. It is possible that the missing permission level could also be caused by a database connection failure between initial account creation and the permission level being set. Since these are separate transactions with the database - this is certainly possible.

---


However, you are correct that the cause of the database connection errors is not well undestood.

The discussion about the database connection issues is somewhat orthogonal to the changes in the PR itself, and really belong in:
  - https://github.com/openwebwork/webwork2/issues/1157
  - https://github.com/openwebwork/webwork2/pull/1160

I can report that in over 3 months of using `DBD::MariaDB` in production on a smaller server - I am no longer seeing any more aborted connections due to `(Got an error reading communication packets)`. I recently reset the main server, but before I reset it I had checked for this issue and there was not any noticeable issue.

I do still see some "rare" aborted connections, but only those for `(Got timeout reading communication packets)` which occur once the `wait_timeout` "session" timeout is reached.  On the server which has not been reset recently, there are 35 such warnings over a 5 week period, where MariaDB has `wait_timeout = 86400` seconds (= 24 hours). Over these 5 weeks there is only one other warning, which is triggered at Docker startup time:
```
[Warning] Aborted connection 8 to db: 'unconnected' user: 'unauthenticated' host: '172.30.0.3' (This connection closed normally without authentication)
```

Thus, at least in my setting: MariaDB 10.4 in a Docker container on the same host as WeBWorK in Docker - the dropped connection issue is no longer occurring, while it was happening with the "older" `DBD::mysql` driver. That driver is not being very actively maintained, and there are several reports of issues with it. (That motivated the creating of the forked driver.)

> The number of seconds the server waits for activity on a connection before closing it.
from https://mariadb.com/docs/reference/mdb/system-variables/wait_timeout/

Alex-Jordan:
I tested this by merging it into develop (locally) and then a real student account in D2L (the LMS) successfully triggered account creating in WeBWorK.

I deleted my admin account in the WW course and tried to enter from the LMS to see what happened. I get the error screen indicating it will not create professor level accounts. Same as in the past. (Side note: this error screen is "frightening". Sometimes a guest instructor or staff is in the LMS course and tries to use the WeBWorK link, and sees this. Is a friendlier screen for this easy to implement?)

I commented out the roles that map to "professor" from LTI config. Then I entered from the LMS with my instructor level LMS account. It created a student level account for me in WW, which is the intent of this PR. 

I'm not sure how I feel about creating student accounts in the WW course this way. For one thing, if we are simply rejecting account creation for a too-high LMS account, why not also reject account creation for an unrecognized role?

And if we are creating accounts for one or both situations (too-high LMS roles or unrecognized LMS roles) did you consider making the WW account guest level instead of student? The instructor of the WW course could elevate permissions later to whatever is appropriate.

drgrice1:
@Alex-Jordan:  You are expressing the same concern that I had with this.  That is why I stated that I was not certain about this pull request.  I think more consideration is needed.

Alex-Jordan:
I would be comfortable approving this if the unrecognized LMS roles become "guest" accounts in WeBWorK instead of "student". And for consistency, I would like to see "too-high" LMS roles also get account creation but with "guest" WeBWorK permissions, but that could be done separately.

Yes, that still leaves things not fully understood, but it seems like a good change anyway.

drgrice1:
@Alex-Jordan: I think that sounds reasonable.

Alex-Jordan (changes requested):
@taniwallach Is it OK with you that the unclassified new accounts be given guest status instead of student? My thoughts are:

(1) we don't really think this is responsible for the instances that have occurred, right?  We think that is because of dropped connections, in which case the new people still have no permission level.

(2) At my school, it is common to have staff join the LMS course. If they follow the LTI link, this could make them be a "student" in WW. Their name will be inserted among all the users when grades are managed. And it's not obvious how to find them and weed them out. But if they have "guest" status, it will be easier to identify them.

Also if you don't object to that change, could you go ahead and make it so that users with too-high permission level are treated the same way? From time to time, an instructor has ended up with a WW course they are not enrolled in. Or a faculty mentor is in the D2L course and would like to enter the WW course. But if they try, they are recognized as having too-high permissions, and they get an error screen that gives the wrong impression that something is broken. I think these situations could allow entry, and soon after that, an admin or professor would elevate their status.

taniwallach:
> @taniwallach Is it OK with you that the unclassified new accounts be given guest status instead of student? My thoughts are:

No - I just was to try to minimize creation failures and broken accounts. Could you make a PR to my branch with the change?

We might want to make it more visible to a user that their account is a guest account, and that if they need a student account - they should write the professor, as otherwise some legitimate users may be confused by what does not work.

> (1) we don't really think this is responsible for the instances that have occurred, right? We think that is because of dropped connections, in which case the new people still have no permission level.

I suspect that most of the problems were probably dropped connections, but the data to tell does not exist. (Once some security expert recommended keeping several days of network sniffing data for all servers somewhere, and that sort of thing might have been helpful in debugging this issue.)

> (2) At my school, it is common to have staff join the LMS course. If they follow the LTI link, this could make them be a "student" in WW. Their name will be inserted among all the users when grades are managed. And it's not obvious how to find them and weed them out. But if they have "guest" status, it will be easier to identify them.

A good point.

> Also if you don't object to that change, could you go ahead and make it so that users with too-high permission level are treated the same way? 

If it can wait a bit - yes. (Currently a bit overloaded with unscheduled work demands.) If someone has a chance to PR to my branch a patch sooner, I'll merge it.

> From time to time, an instructor has ended up with a WW course they are not enrolled in. Or a faculty mentor is in the D2L course and would like to enter the WW course. But if they try, they are recognized as having too-high permissions, and they get an error screen that gives the wrong impression that something is broken. I think these situations could allow entry, and soon after that, an admin or professor would elevate their status.

Makes sense. Again - I think a "banner" indicator for "guest status" might help this approach work more smoothly.

dlglin:
It's worth noting that the current behaviour for users without a permission record is to assign them a permission level of student.  This happens the first time that the user record is accessed in the Classlist Editor:
https://github.com/openwebwork/webwork2/blob/03eca4cc1a6034d24cf28bd690f0f56ce14f284b/lib/WeBWorK/ContentGenerator/Instructor/UserList2.pm#L454-L465

drgrice1:
@dlglin:  Although it is also worth noting that if one is creating a user via the class list editor (UserList2.pm), then the user is being intentionally created.  If a user is created in the manner that this pull request is implementing, that user is not necessarily intentional.

dlglin:
@drgrice1 That code in UserList2.pm runs on all users regardless of how they were created.
I experienced the same behaviour that @taniwallach describes, where LTI created the user record, but the DB connection aborted before the permission level were set.  Simply viewing the users in the Classlist Editor triggers the assigning of student permission to any user who doesn't have a permission level.

taniwallach:
This solved a different issue and needs work.
Need to add configurable default.

Issue #1077, opened 2020-02-21
UTF8 issue with permissions saved to simple.conf


taniwallach:
There is an issue with the code of the permission editor / the code which writes to simple.conf which is triggered when changing between languages. 

(The issue was mentioned in https://github.com/openwebwork/webwork2/pull/948#issuecomment-524571050 on Aug 24, 2019 but was not a reason to hold back on the WW 2.15 release.)

The problem is not fully diagnosed, but under certain circumstances the value save for `$permissionLevels{type}` is being set using the foreign language value rather than the English value, which breaks the behavior.

A sample procedure to trigger the issue is as follows:

1. Set `simple.conf` in some course to be the contents of the file provided at the bottom of the issue.
    - Maybe use a different value of the language (which has translated names of the permission levels) instead of Hebrew so you can follow the steps below with more ease. 
    - However, Hebrew definitely triggers the problem.
2. Use the menu to open the course configuration page - two lines above the ?-mark icon in the menu. (Hebrew phrase is תצורת הקורס, in the current PO file)
3. Change the language setting from "heb" to "en_us".
4. Click the save button on the bottom of the page.
5. Example the value of `$permissionLevels{change_password}` in the saved `simple.conf`. It will show (based on my tests) the value `מרצה` which is the Hebrew term for `professor` in use.
6. Use the left menu to change to a different part of the system and them back to the configuration page. That will get the language change to take effect.
7. Now open the "Permissions" tab again. It will show the value `guest` as the value `מרצה` from the file cannot be read in.

---

Note: When the permissions are edited and saved while using a foreign language, and the language setting is **not** changes, the values saved in `simple.conf` are the proper English names of the relevant permission levels. The problem is apparently **only** triggered when a save is occurring which changes from a foreign language back to English 

---

Note: In the pull-down lists, the `<option>` items have a `value` which is the standard English name of the permission levels, but the text shown for the options is the translated phrase. When the permissions page is saved in a foreign language - the form values sent are the English values from the value field.

---

The probably cause is a call to `maketext` which is not needed or some value already saved as the "translated" value rather than the English value.

---

A prior issue with related code is at: https://github.com/openwebwork/webwork2/pull/798#issuecomment-321260258

---

The sample `simple.conf` used to trigger the issue
```
#!perl
# This file is automatically generated by WeBWorK's web-based
# configuration module.  Do not make changes directly to this
# file.  It will be overwritten the next time configuration
# changes are saved.

$language = 'heb';
[...]

PR #1152, opened 2020-11-02
Fix issue #1077

cucucucumber:
I fixed the issue #1077 by forcing UTF-8 encoding to read and write configuration files.

**Warning:** Because of the Windows text file saving mechanism, some unexpected situations may occur on Windows Server 2012/Windows 8.1 and earlier versions. But this problem does not appear after Windows 10 Build 19041 and forced UTF-8 file encoding is turned on.

PR #1153, opened 2020-11-02
Fix issue #1077

cucucucumber:
This is an attempt to fix the issue #1077 by forcing UTF-8 encoding to read and write configuration files.

Note: Because of the Windows text file saving mechanism, some unexpected situations can occur on Windows Server 2012/Windows 8.1 and earlier versions. Most commonly, in such circumstances languages other than English may experience garbled characters.  But this problem does not appear after Windows 10 Build 19041. To better support multilingual features maybe we should suggest users to upgrade to Windows Server 2018 or later.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
@@ -33,6 +33,11 @@ package configobject;
 
 use strict;
 use warnings;
+use utf8;
+
+binmode(STDIN, ':encoding(utf8)');
+binmode(STDOUT, ':encoding(utf8)');
+binmode(STDERR, ':encoding(utf8)');
 
 sub new {
 	my $class = shift;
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Following up on #1152, this pull request looks good. One or more people will test it out before merging, and it's a variable how long it takes for people to get around to that. Thanks for working on the issue, and welcome!

taniwallach (changes requested):

robert-marik:
Another attempt to solve the issue in another way: https://github.com/robert-marik/webwork2/commit/939fe2b1d86c54011c3dd28b709efb7ae4040ce4
The main idea is to put conf file in English. See my comments in the original issue https://github.com/openwebwork/webwork2/issues/1077 for an explanation.

taniwallach:
A better fix to https://github.com/openwebwork/webwork2/issues/1077 is in https://github.com/openwebwork/webwork2/pull/1352.

Issue #1144, opened 2020-10-01
Problems with accented characters in feedback emails on some systems


taniwallach:
See:  https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4916

It seems possible that the use of `encode_utf8($msg)` made in https://github.com/openwebwork/webwork2/pull/973 will need to be replaced by the use of `encode("UTF-8", $msg)`.

We need confirmation that this helps on systems experiencing such problems before making such a change.

Issue #982, opened 2019-08-09
Investigate WWSafe.pm vs. Perl's standard Safe.pm - UTF-8 motivation


taniwallach:
WeBWorK's `lib/WWSafe.pm` is based on `$Safe::VERSION = "2.16"` and does not allow the use of the Perl `Encode` module fully inside the processing of PG files.  Perl 5.24.1 has `$Safe::VERSION = "2.39";` and that version has additional code to enable the Perl `Encode` module to work better within "Safe". Updating `lib/WWSafe.pm` will require great care and expertise.

As a result some code of the limitation of `lib/WWSafe.pm` some parts of PG use `<:utf8` instead of the stricter and safer `<:encoding(UTF-8)`.

See the discussions at:
  * https://github.com/openwebwork/pg/pull/390
  * https://github.com/openwebwork/pg/pull/402
  * https://github.com/openwebwork/webwork2/pull/930

Somewhat related: 
  * https://github.com/openwebwork/webwork2/pull/908

Files (in pg):
  * `lib/WeBWorK/PG/IO.pm`
    * `read_whole_file` reads a file in `raw` mode and then uses `utf8::decode($string);`
    * we would prefer to read the file using `<:encoding(UTF-8)`
    * there is a comment about `Safe` not letting `<:encoding(UTF-8)` work
  * `lib/PGloadfiles.pm`
    * `compile_file` reads a file in `raw` mode and then uses `utf8::decode($string);`
    * we would prefer to read the file using `<:encoding(UTF-8)`
  * `lib/WeBWorK/PG/Translator.pm`
    * use `binmode(STDOUT, ":utf8")` 
    * and not something like `binmode(":encoding(UTF-8)")` 
    * or `binmode(INPUT,":encoding(UTF-8)");`
  * `lib/PGcore.pm`
    * uses `binmode(STDOUT, ":utf8");`
    * and not something like `binmode(":encoding(UTF-8)")`

taniwallach:
I think that this has been ignored for too long to want to consider it as something we are going to handle for WW 2.16.
@mgage @drgrice1 - What do you think?

drgrice1:
That is probably best unless someone has the time to work on it.

drdrew42:
[...]

PR #1154, opened 2020-11-02
UTF-8 changes

taniwallach:
- Replace
      - `encode_utf8($x)` by `Encode::encode("UTF-8",$x)`, and
      - `encode_utf8($x)` by `Encode::decode("UTF-8",$x)`
as the prior is not a proper conversion, and the documentation at https://perldoc.perl.org/Encode#encode_utf8
state that it should not be used for data exchange.
  - In several places use `Encode::encode()` explicitly, instead of just `encode()` after a `use Encode qw(encode)`.
  - Replace `binmode(STDOUT, ":utf8");` with  `binmode(STDOUT, ":encoding(UTF-8)");` as Perl 5.30 makes sysread and syswrite on ":utf8" filehandles fatal (it was deprecated in 5.24).
See: https://perldoc.perl.org/perl5300delta#Previously-deprecated-sysread()/syswrite()-on-:utf8-handles-is-now-fatal

So far I have not thoroughly tested the changes.
---

Motivation:
  - UTF-8 email issue discussed at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4916 and the issue at https://github.com/openwebwork/webwork2/issues/1144 related to that post.
  - Error messages I found in my Apache error.log files:
```
syswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 86.
Caught a SIGsyswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 87.
PIPEsyswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 88.
 at syswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 90.
-esyswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 91.
 line syswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 92.
0syswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 93.
syswrite() is deprecated on :utf8 handles. This will be a fatal error in Perl 5.30 at /usr/share/perl/5.26/sigtrap.pm line 119.
```

### DIFF
--- a/bin/importClassList.pl
+++ b/bin/importClassList.pl
@@ -31,8 +31,6 @@ use WeBWorK::File::Classlist;
 #use WeBWorK::Utils qw(readFile readDirectory cryptPassword x);
 use WeBWorK::Utils qw(cryptPassword);
 
-use Encode qw(decode_utf8 encode_utf8);
-
 use strict;
 use warnings;
 
--- a/lib/Apache/WeBWorK.pm
+++ b/lib/Apache/WeBWorK.pm
@@ -68,9 +68,9 @@ sub handler($) {
 	my $uri = $r->uri;
 
 
-	# We set the bimode for print to utf8 because some language options
+	# We set the binmode for print to utf8 because some language options
 	# use utf8 characters
-	binmode(STDOUT, "encoding(UTF-8)");
+	binmode(STDOUT, ":encoding(UTF-8)");
 	# the warning handler accumulates warnings in $r->notes("warnings") for
 	# later cumulative reporting
 	my $warning_handler;
@@ -79,11 +79,11 @@ sub handler($) {
 			my ($warning) = @_;
 			chomp $warning;
 			my $warnings = $r->notes->get("warnings");
-			$warnings = Encode::decode_utf8($warnings);
+			$warnings = Encode::decode("UTF-8",$warnings);
 			$warnings .= "$warning\n";
 			#my $backtrace = join("\n",backtrace());
 			#$warnings .= "$backtrace\n\n";
-			$warnings = Encode::encode_utf8($warnings);
+			$warnings = Encode::encode("UTF-8",$warnings);
 			$r->notes->set(warnings => $warnings);
 			
 			$log->warn("[$uri] $warning");
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -32,7 +32,6 @@ use WeBWorK::PG::ImageGenerator;
 use WeBWorK::Utils qw( wwRound encode_utf8_base64 decode_utf8_base64);
 use XML::Simple qw(XMLout);
 use WeBWorK::Utils::DetermineProblemLangAndDirection;
-use Encode qw(encode_utf8 decode_utf8);
 use JSON;
 
 our $UNIT_TESTS_ON  = 0; 
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -1261,7 +1261,7 @@ sub warnings {
 	print CGI::p("Entering ContentGenerator::warnings") if $TRACE_WARNINGS;
 	print "\n<!-- BEGIN " . __PACKAGE__ . "::warnings -->\n";
 	my $warnings = MP2 ? $r->notes->get("warnings") : $r->notes("warnings");
-	$warnings = Encode::decode_utf8($warnings);
+	$warnings = Encode::decode("UTF-8",$warnings);
 	print $self->warningOutput($warnings) if $warnings;
 	print "<!-- END " . __PACKAGE__ . "::warnings -->\n";
 
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -29,7 +29,6 @@ WeBWorK::ContentGenerator::Feedback - Send mail to professors.
 use strict;
 use warnings;
 use utf8;
-use Encode qw(encode_utf8 encode);
 use Data::Dumper;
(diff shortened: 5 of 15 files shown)


drgrice1 (approved):
This looks good.  I am not able to do any extensive testing on this as my usage rarely deals with anything that will really stress test the utf8 encoding and decoding.

I noticed that there is one instance of encode_utf8 that you did not convert to the `Encode::encode("UTF-8", ...)` method, and this is in lib/WeBWorK/EquationCache.pm.  It may be it is not needed there.  Just pointing this out.

Also, both `Encode::decode_utf8` and `Encode::encode_utf8` are used in PGcore.pm in the PG code.  `binmode(STDOUT, ":utf8")` is also used there.  Do those need to be changed also?

taniwallach:
> I noticed that there is one instance of encode_utf8 that you did not convert to the `Encode::encode("UTF-8", ...)` method, and this is in lib/WeBWorK/EquationCache.pm. It may be it is not needed there. Just pointing this out.

I'll look into that one soon.

---

> Also, both `Encode::decode_utf8` and `Encode::encode_utf8` are used in PGcore.pm in the PG code. `binmode(STDOUT, ":utf8")` is also used there. Do those need to be changed also?

Ideally yes, practically - probably not now.

`WWSafe interferes`  with `Encode::encode("UTF-8", ...)`.
  - See https://github.com/openwebwork/webwork2/issues/982 and the older discussions it references, and 
  -the comment at https://github.com/openwebwork/pg/blob/a284c3b31a26fb316c2fa6b34246df8ba2895a3e/lib/WeBWorK/PG/IO.pm#L162
  - As a result, several compromises exist in PG.
  - Apparently, more recent versions of `Safe.pm` have approaches to allow `Encode` to pre-load and work, but `WWSafe` was forked of an earlier version.
  - That was the reason I once discussed with @mgage how feasible it would be to update `WWSafe` or go back to `Safe.pm`

drgrice1:
I didn't see that issue.

taniwallach:
> > I noticed that there is one instance of encode_utf8 that you did not convert to the `Encode::encode("UTF-8", ...)` method, and this is in lib/WeBWorK/EquationCache.pm. It may be it is not needed there. Just pointing this out.
> 
> I'll look into that one soon.

That is `pg/lib/WeBWorK/EquationCache.pm` and our experience is that `WWSafe` does not let `Encode::encode("UTF-8", ...)` work properly in PG. 

The use is in `my $md5 = md5_hex(encode_utf8($tex));` so it is only using Perl's lax UTF-8 to encode a string which is them passed into `md5_hex`. I do not think that any of the potential security risks related to Perl's lax UTF-8 code can cause any significant problems in such a setting.

---

From some research 2 years ago, the issues were related to the autoloading mechanisms used by `Encode` and that more modern versions of the regular Perl `Safe.pm` had changes to allow `Encode` to behave properly inside `Safe`.

See:
  - https://github.com/openwebwork/pg/pull/390#issuecomment-470212699  
  - https://github.com/openwebwork/webwork2/issues/982
  - https://github.com/openwebwork/pg/commit/3d94218b2b1317bd4ed32c5e6943d6aa30b375dd
  - https://github.com/openwebwork/pg/blob/a284c3b31a26fb316c2fa6b34246df8ba2895a3e/lib/WeBWorK/PG/IO.pm#L162
  - https://github.com/openwebwork/pg/pull/402#issuecomment-472156239

drgrice1:
Yeah, that probably is fine. Furthermore, does anyone even use images for equations anymore? MathJax is so much better.

mgage:
There are situations where it is a good backup.  Some labs still don’t enable javaScript.  I periodically run into students who
for one reason or another can’t get the javascript to work on their computer/browser.  At the very least the image provides
a check that everything else is working.

drgrice1:
Yeah, I am not saying to remove the code for that.  I am just saying it is probably not used much, and as such, it is probably not a serious concern if utf8 encoding is proper (if that is even needed for equations in any case).

I am betting that the instances in which students have trouble with the javascript working are becoming increasingly infrequent (at least in regards to MathJax).  I have never had that happen for my students as far as I know.

taniwallach:
> I periodically run into students who for one reason or another can’t get the javascript to work on their computer/browser.

I have run into a few such cases. Often asking the student to use a different browser than their default browser will overcome the problem. A few cooperative students who disabled browser extensions/plugins and bothered to respond after trying that reported success with extensions/plugins disabled. (There are several types of plugins which block content or modify content in ways the the user's don't really understand, but have some feature they like.) I have never got a student to test enough and write back telling me which extension/plugin caused their problem. The students typically just want to do the homework and move on.

taniwallach:
> if utf8 encoding is proper (if that is even needed for equations in any case).

Submitted answers can contain UTF-8 character (even if they are not suitable), and almost everything gets rendered inside a math element in the "submitted answers" column in the results table. When using images equation display mode without the UTF-8 encoding an error page would be displayed (see the details in https://github.com/openwebwork/pg/issues/443#issue-536903798), as the MD5 hash used to create file names cannot process Unicode characters from the internal Perl encoding of strings. The patch was made to encode into UTF-8, and the the image fails to be generated as LaTeX can't actually make the image so also fails appear in the rendered image, but the tool-top text does work. Having something fail to appear on screen, is far better than there being a WeBWorK error page.

PR #1159, opened 2020-11-14
Czech translation (student's interface)

robert-marik:
Translation of most texts visible for student. Allows to work comfortably with the system when doing homeworks.
The texts for teacher and admin are kept in English in order stay in touch with community and follow better community hints and suggestions.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
@@ -54,6 +54,8 @@ use WeBWorK::Debug;
 use WeBWorK::Utils qw(timeToSec readFile listFilesRecursive cryptPassword sortAchievements);
 use DateTime;
 use Text::CSV;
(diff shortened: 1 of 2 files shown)


taniwallach:
Please split the 2 commits into different pull request. I think the UTF-8 patch for achievements should be handled separately from the addition of the Czech translation file, so that the "history" in Git is clear.
Please create the new pull requests from the **develop** branch. I do not think there is any need to make either fix a hot-fix for the current WW 2.15 (master branch). Expert users who need do not want to run the develop branch and want these fixes either before WW 2.16 can certainly handle making the necessary changes to a local server.

taniwallach (changes requested):
Request to base of develop, and split into 2 different PRs.

robert-marik:
Thank you for the suggestions. I will prepare new patches according to your instructions.

robert-marik:
Will prepare new pull requests as suggested.

Issue #1157, opened 2020-11-09
Reports of database issues - aborted connections and (hopefully) small amounts of data loss + a possible fix using the new DBD::MariaDB Perl database driver


taniwallach:
See: https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4958

Original symptom were rare cases of students getting the error `Unable to retrieve your permissions, perhaps due to a collision between your request and that of another user (or possibly unfinished request of yours)`.

That same symptom on my server led to my attempt at patching some of the LTI code at: https://github.com/openwebwork/webwork2/pull/1150 and the discussion there got into the issue of the DB logs reporting warnings like
```
2020-11-09  5:21:50 16 [Warning] Aborted connection 16 to db: 'webwork' user: 'webworkWrite' host: '172.22.0.4' (Got timeout reading communication packets)
```

@dlglin reported in the forum thread that he encountered such problems when his site tried to use a new externally hosted database.

---

Based on some preliminary tests, it seems that using the newer `DBD::MariaDB` database driver instead of the older `DBD:mysql` may help resolve this problem.

Using the new driver requires making some changes to the WW code, as explained in https://github.com/openwebwork/webwork2/pull/1150#issuecomment-724196085. Changes to allow initial testing are pretty minimal.

Packages are available for several Linux distributions: 
  - https://repology.org/project/perl:dbd-mariadb/packages
  - https://repology.org/project/perl:dbd-mariadb/versions
  - http://ftp.us2.freshrpms.net/linux/rpm2html/search.php?query=perl(DBD%3A%3AMariaDB)
but not for CentOS/RHEL.

Some comments about building the driver on CentOS 7 are in https://github.com/gooddata/DBD-MariaDB/issues/150

PR #1160, opened 2020-11-23
Modifications to support use of the DBD::MariaDB driver instead of DBD::mysql

taniwallach:
Modifications to allow use of the `DBD::MariaDB` driver instead of `DBD::mysql`.

The `DBD::MariaDB` driver handles UTF-8 by default, and does not need `mysql_enable_utf8mb4` or `mysql_enable_utf8` and would fail if they were given.

These changes are far beyond the initial one mentioned in https://github.com/openwebwork/webwork2/pull/1150#issuecomment-724196085 which was needed to allow using the new driver.

See:
  - https://github.com/openwebwork/webwork2/pull/1150
  - https://github.com/openwebwork/webwork2/issues/1157

---

Also included modifications to prevent `mysqldump` of mysql 8+ for reporting errors about `column-statistics` being missing from MariaDB databases by disabling column_statistic when the installed `mysqldump` command would support it.

See:
  - https://serverfault.com/questions/912162/mysqldump-throws-unknown-table-column-statistics-in-information-schema-1109
  - https://github.com/drush-ops/drush/issues/4410

### DIFF
--- a/bin/dump-OPL-tables.pl
+++ b/bin/dump-OPL-tables.pl
@@ -66,6 +66,16 @@ my $prepared_OPL_tables_file = "$prepared_OPL_tables_dir/OPL-tables.sql";
 
 my ($dbi,$dbtype,$db,$host,$port) = split(':',$ce->{database_dsn});
 
+# The MariaDB driver use a different DSN format
+# Ex: DBI:MariaDB:database=webwork;host=db;port=3306
+
+if ( $dbtype =~ /MariaDB/i ) {
+  ($db,$host,$port) = split(';',$db);
+  $db   =~ s/database=//;
+  $host =~ s/host=//;
+  $port =~ s/port=//;
+}
+
 $host = 'localhost' unless $host;
 
 $port = 3306 unless $port;
@@ -101,7 +111,17 @@ my $OPL_tables_to_dump = "OPL_DBsubject OPL_DBchapter OPL_DBsection OPL_author O
 
 print "Dumping OPL tables\n";
 
-`$mysqldump_command --host=$host --port=$port --user=$dbuser --default-character-set=$character_set $db $OPL_tables_to_dump  > $prepared_OPL_tables_file`;
+# Conditionally add --column-statistics=0 as MariaDB databases do not support it
+# see: https://serverfault.com/questions/912162/mysqldump-throws-unknown-table-column-statistics-in-information-schema-1109
+#      https://github.com/drush-ops/drush/issues/4410
+
+my $column_statistics_off = "";
+my $test_for_column_statistics = `$mysqldump_command --help | grep 'column-statistics'`;
+if ( $test_for_column_statistics ) {
+  $column_statistics_off = " --column-statistics=0 ";
+}
+
+`$mysqldump_command --host=$host --port=$port --user=$dbuser --default-character-set=$character_set $column_statistics_off $db $OPL_tables_to_dump  > $prepared_OPL_tables_file`;
 
 print "OPL database dump created: $prepared_OPL_tables_file\n";
 
--- a/lib/WeBWorK/Utils/CourseManagement/sql_single.pm
+++ b/lib/WeBWorK/Utils/CourseManagement/sql_single.pm
@@ -195,17 +195,38 @@ sub _get_db_info {
 	my $dsn = $ce->{database_dsn};
 	my $username = $ce->{database_username};
 	my $password = $ce->{database_password};
-	
-	die "Can't call dump_table or restore_table on a table with a non-MySQL source"
-		unless $dsn =~ s/^dbi:mysql://i;
-	
-	# this is an internal function which we probably shouldn't be using here
-	# but it's quick and gets us what we want (FIXME what about sockets, etc?)
+
 	my %dsn;
-	runtime_use "DBD::mysql";
-	DBD::mysql->_OdbcParse($dsn, \%dsn, ['database', 'host', 'port']);
-	die "no database specified in DSN!" unless defined $dsn{database};
+	if (      $dsn =~ s/^dbi:mariadb://i ) {
+		my ($dbi,$dbtype,$temp1) = split(':',$dsn);
+		( $dsn{database}, $dsn{host}, $dsn{port} ) = split(';',$db);
+		$dsn{database} =~ s/database=//;
+		$dsn{host} =~ s/host=// if ( defined $dsn{host} );
+		$dsn{port} =~ s/port=// if ( defined $dsn{port} );
+	} elsif ( $dsn =~ s/^dbi:mysql://i ) {
+		# This code works for DBD::mysql
+		# this is an internal function which we probably shouldn't be using here
+		# but it's quick and gets us what we want (FIXME what about sockets, etc?)
+		runtime_use "DBD::mysql";
+		DBD::mysql->_OdbcParse($dsn, \%dsn, ['database', 'host', 'port']);
+	} else {
+		die "Can't call dump_table or restore_table on a table with a non-MySQL/MariaDB source";
+	}
 	
+	die "no database specified in DSN!" unless defined $dsn{database};
+
+	my $mysqldump = $self->{params}{mysqldump_path};
+	# Conditionally add column-statistics=0 as MariaDB databases do not support it
+	# see: https://serverfault.com/questions/912162/mysqldump-throws-unknown-table-column-statistics-in-information-schema-1109
+	#      https://github.com/drush-ops/drush/issues/4410
+
+	my $column_statistics_off = "";
+	my $test_for_column_statistics = `$mysqldump_command --help | grep 'column-statistics'`;
+	if ( $test_for_column_statistics ) {
+		$column_statistics_off = "[mysqldump]\ncolumn-statistics=0\n";
+warn "Setting in the temporary mysql config file for table dump/restore:\n$column_statistics_off\n\n";
+	}
+
 	# doing this securely is kind of a hassle...
 	my $my_cnf = new File::Temp;
 	$my_cnf->unlink_on_destroy(1);
@@ -215,7 +236,8 @@ sub _get_db_info {
 	print $my_cnf "password=$password\n" if defined $password and length($password) > 0;
 	print $my_cnf "host=$dsn{host}\n" if defined $dsn{host} and length($dsn{host}) > 0;
 	print $my_cnf "port=$dsn{port}\n" if defined $dsn{port} and length($dsn{port}) > 0;
-	
+	print $my_cnf "$column_statistics_off" if $test_for_column_statistics;
+
 	return ($my_cnf, $dsn{database});
 }
 
--- a/bin/load-OPL-global-statistics.pl
+++ b/bin/load-OPL-global-statistics.pl
@@ -53,6 +53,18 @@ my $global_sql_file = $ce->{problemLibrary}{root}.'/OPL_global_statistics.sql';
 if (-e $global_sql_file) {
 
   my ($dbi,$dbtype,$db,$host,$port) = split(':',$ce->{database_dsn});
+
+  # The MariaDB driver use a different DSN format
+  # Ex: DBI:MariaDB:database=webwork;host=db;port=3306
+
+  if ( $dbtype =~ /MariaDB/i ) {
+    ($db,$host,$port) = split(';',$db);
+    $db   =~ s/database=//;
+    $host =~ s/host=//;
+    $port =~ s/port=//;
+  }
+
+
   
   $host = 'localhost' unless $host;
 
--- a/bin/restore-OPL-tables.pl
+++ b/bin/restore-OPL-tables.pl
@@ -66,6 +66,16 @@ my $prepared_OPL_tables_file = "$prepared_OPL_tables_dir/OPL-tables.sql";
 
 my ($dbi,$dbtype,$db,$host,$port) = split(':',$ce->{database_dsn});
 
+# The MariaDB driver use a different DSN format
+# Ex: DBI:MariaDB:database=webwork;host=db;port=3306
+
+if ( $dbtype =~ /MariaDB/i ) {
+  ($db,$host,$port) = split(';',$db);
+  $db   =~ s/database=//;
+  $host =~ s/host=//;
+  $port =~ s/port=//;
+}
+
 $host = 'localhost' unless $host;
 
 $port = 3306 unless $port;
--- a/bin/upload-OPL-statistics.pl
+++ b/bin/upload-OPL-statistics.pl
@@ -32,6 +32,16 @@ my $ce = new WeBWorK::CourseEnvironment({
 
 my ($dbi,$dbtype,$db,$host,$port) = split(':',$ce->{database_dsn});
 
+# The MariaDB driver use a different DSN format
+# Ex: DBI:MariaDB:database=webwork;host=db;port=3306
+
+if ( $dbtype =~ /MariaDB/i ) {
+  ($db,$host,$port) = split(';',$db);
+  $db   =~ s/database=//;
+  $host =~ s/host=//;
+  $port =~ s/port=//;
+}
+
 $host = 'localhost' unless $host;
 
 $port = 3306 unless $port;
@@ -53,7 +63,17 @@ $ENV{'MYSQL_PWD'}=$dbpass;
 
 my $mysqldump_command = $ce->{externalPrograms}->{mysqldump};  
 
-`$mysqldump_command --host=$host --port=$port --user=$dbuser $db OPL_local_statistics > $output_file`;
+# Conditionally add --column-statistics=0 as MariaDB databases do not support it
+# see: https://serverfault.com/questions/912162/mysqldump-throws-unknown-table-column-statistics-in-information-schema-1109
+#      https://github.com/drush-ops/drush/issues/4410
+
+my $column_statistics_off = "";
+my $test_for_column_statistics = `$mysqldump_command --help | grep 'column-statistics'`;
+if ( $test_for_column_statistics ) {
+  $column_statistics_off = " --column-statistics=0 ";
+}
+
+`$mysqldump_command --host=$host --port=$port --user=$dbuser $column_statistics_off $db OPL_local_statistics > $output_file`;
 
 print "Database File Created\n";
 
--- a/lib/WeBWorK/DB/Driver/SQL.pm
+++ b/lib/WeBWorK/DB/Driver/SQL.pm
@@ -61,6 +61,15 @@ sub new($$$) {
 	
 	my $self = $proto->SUPER::new($source, $params);
 	
+	# The DBD::MariaDB driver should not get the
+	#    mysql_enable_utf8mb4 or mysql_enable_utf8 settings,
+	# but DBD::mysql should.
+	my %utf8_parameters = ();
+	if ( $source =~ /DBI:mysql/ ) {
+	  $utf8_parameters{mysql_enable_utf8mb4} = 1;
+	  $utf8_parameters{mysql_enable_utf8} = 1;
+	}
+
 	# add handle
 	$self->{handle} = DBI->connect_cached(
 		$source,
@@ -70,8 +79,7 @@ sub new($$$) {
 			PrintError => 0,
 			RaiseError => 1,
 
-			mysql_enable_utf8mb4 => 1,
-			mysql_enable_utf8 => 1,  # for older versions of DBD-mysql Perl modules
+			%utf8_parameters,
 		},
 	);
 	die $DBI::errstr unless defined $self->{handle};
--- a/lib/WeBWorK/DB/Schema/NewSQL/Std.pm
+++ b/lib/WeBWorK/DB/Schema/NewSQL/Std.pm
@@ -269,28 +269,52 @@ sub _get_db_info {
 	my $dsn = $self->{driver}{source};
 	my $username = $self->{params}{username};
 	my $password = $self->{params}{password};
-	
-	die "Can't call dump_table or restore_table on a table with a non-MySQL source"
-		unless $dsn =~ s/^dbi:mysql://i;
-	
-	# this is an internal function which we probably shouldn't be using here
-	# but it's quick and gets us what we want (FIXME what about sockets, etc?)
+
 	my %dsn;
-	DBD::mysql->_OdbcParse($dsn, \%dsn, ['database', 'host', 'port']);
-	die "no database specified in DSN!" unless defined $dsn{database};
+	if (      $dsn =~ m/^dbi:mariadb:/i ) {
+		# Expect DBI:MariaDB:database=webwork;host=db;port=3306
+		my ($dbi,$dbtype,$temp1) = split(':',$dsn);
+		( $dsn{database}, $dsn{host}, $dsn{port} ) = split(';',$temp1);
+		$dsn{database} =~ s/database=//;
+		$dsn{host} =~ s/host=// if ( defined $dsn{host} );
+		$dsn{port} =~ s/port=// if ( defined $dsn{port} );
+	} elsif ( $dsn =~ m/^dbi:mysql:/i ) {
+		# This code works for DBD::mysql
+		# this is an internal function which we probably shouldn't be using here
+		# but it's quick and gets us what we want (FIXME what about sockets, etc?)
+		DBD::mysql->_OdbcParse($dsn, \%dsn, ['database', 'host', 'port']);
+	} else {
+		die "Can't call dump_table or restore_table on a table with a non-MySQL/MariaDB source";
+	}
 	
+	die "no database specified in DSN!" unless defined $dsn{database};
+
+	my $mysqldump = $self->{params}{mysqldump_path};
+	# Conditionally add column-statistics=0 as MariaDB databases do not support it
+	# see: https://serverfault.com/questions/912162/mysqldump-throws-unknown-table-column-statistics-in-information-schema-1109
+	#      https://github.com/drush-ops/drush/issues/4410
+
+	my $column_statistics_off = "";
+	my $test_for_column_statistics = `$mysqldump --help | grep 'column-statistics'`;
+	if ( $test_for_column_statistics ) {
+		$column_statistics_off = "[mysqldump]\ncolumn-statistics=0\n";
+		#warn "Setting in the temporary mysql config file for table dump/restore:\n$column_statistics_off\n\n";
+	}
+
 	# doing this securely is kind of a hassle...
 	my $my_cnf = new File::Temp;
 	$my_cnf->unlink_on_destroy(1);
 	chmod 0600, $my_cnf or die "failed to chmod 0600 $my_cnf: $!"; # File::Temp objects stringify with ->filename
 	print $my_cnf "[client]\n";
-	print $my_cnf "user=\"$username\"\n" if defined $username and length($username) > 0;
-	print $my_cnf "password=\"$password\"\n" if defined $password and length($password) > 0;
-	print $my_cnf "host=\"$dsn{host}\"\n" if defined $dsn{host} and length($dsn{host}) > 0;
-	print $my_cnf "port=\"$dsn{port}\"\n" if defined $dsn{port} and length($dsn{port}) > 0;
-	
+	print $my_cnf "user=$username\n" if defined $username and length($username) > 0;
+	print $my_cnf "password=$password\n" if defined $password and length($password) > 0;
+	print $my_cnf "host=$dsn{host}\n" if defined $dsn{host} and length($dsn{host}) > 0;
+	print $my_cnf "port=$dsn{port}\n" if defined $dsn{port} and length($dsn{port}) > 0;
+	print $my_cnf "$column_statistics_off" if $test_for_column_statistics;
+
 	return ($my_cnf, $dsn{database});
 }
+
 ####################################################
 # checking Fields
 ####################################################


taniwallach:
I have tested:
  - Regular course operations - the simpler changes were in use for 2 weeks on production servers.
  - Archiving a course using these changes. (Would not have worked with the change mentioned in https://github.com/openwebwork/webwork2/pull/1150#issuecomment-724196085 .)
  - Ran all of the following successfully in a Docker container using the `DBD::MariaDB` driver.
    - `bin/dump-OPL-tables.pl` 
    - `bin/restore-OPL-tables.pl`
    - `bin/load-OPL-global-statistics.pl`
    - `bin/upload-OPL-statistics.pl` (except for the stfp call, which was not enabled).

The same tests need to also be run on a server which is still using the `DBD::mysql` driver.

taniwallach:
I can report that in over 3 months of using `DBD::MariaDB` in production on a smaller server - I am no longer seeing any more aborted connections due to `(Got an error reading communication packets)`. 

I recently reset my main production server, but before I reset it I had checked for this issue and there was not any noticeable issue.

I do still see some "rare" aborted connections, but only those for `(Got timeout reading communication packets)` which occur once the `wait_timeout` "session" timeout is reached.  On the server which has not been reset recently, there are 35 such warnings over a 5 week period, where MariaDB has `wait_timeout = 86400` seconds (= 24 hours). 

Over these 5 weeks there is only one other warning, which is triggered at Docker startup time:
```
[Warning] Aborted connection 8 to db: 'unconnected' user: 'unauthenticated' host: '172.30.0.3' (This connection closed normally without authentication)
```

Thus, at least in my setting: MariaDB 10.4 in a Docker container on the same host as WeBWorK in Docker - the dropped connection issue is no longer occurring, while it was happening with the "older" `DBD::mysql` driver. That driver is not being very actively maintained, and there are several reports of issues with it. (That motivated the creating of the forked driver.)

> The number of seconds the server waits for activity on a connection before closing it.
from https://mariadb.com/docs/reference/mdb/system-variables/wait_timeout/

drgrice1 (changes requested):
This looks good and passed all my tests, but I you should add comments above the `$database_dsn` variable in site.conf explaining how to use this, and the describe format that variable should be in.

dlglin:
If we are looking at migrating to DBD::MariaDB then we also need to include this as a required perl module.  It is not currently packaged in RedHat 7, so I had to install it from CPAN on my test server.  This should be documented somewhere, and probably included in check_modules.pl.

drgrice1:
I am not sure that we should include DBD::MariaDB in check_modules.pl.  The way this pull request is set up the system administrator is given the choice to use DBD:MariaDB or DBD::mysql.  So what we need is a mechanism to have check_modules.pl check that one of DBD::MariaDB or DBD::mysql is installed.

pstaabp:
I haven't been success running these.  On anything that hits the database, I get as an example for `dump-OPL-tables.pl`:

```
mysqldump: Got error: 2002: "Can't connect to MySQL server on 'db' (36)" when trying to connect
OPL database dump created: XXX/libraries/webwork-open-problem-library//TABLE-DUMP/OPL-tables.sql
```

I'm running the following version of mariadb:
```
mysql  Ver 10.19 Distrib 10.5.9-MariaDB, for osx10.16 (x86_64)
```

The only change I made was in site.conf to set
```
$database_dsn = "DBI:MariaDB:database=webwork;host=db;port=3306";
```
and then did
```
$ENABLE_UTF8MB4 =0; 
```

I'm sure I have some other setting wrong, but can't figure that out right now.  Fortunately, changing it back worked fine.  

I think overall, if go with this, we need to do a bunch of documentation to allow people to try out.  This is probably fine to include as long as it doesn't break old things.

drgrice1:
@pstaabp:  Most likely you need to change the line `$database_dsn = "DBI:MariaDB:database=webwork;host=db;port=3306";` to `$database_dsn = "DBI:MariaDB:database=webwork";`

That is unless you have your database setup to be accessed via a different host.  In the docker setup the host is db.

pstaabp:
@drgrice1 That did it and I got everything running above, however, I also tested OPL-update and got the error:
```
DBI connect('database=webwork','webworkWrite',...) failed: Unknown attribute mysql_enable_utf8 at /Users/pstaab/code/ww-docker/webwork2/bin/OPL-update line 100.
```

Looks like the database handler might need to be altered for MariaDB in this file.

drgrice1:
@pstaabp:  You are correct.  That is an oversight.

@taniwallach:  See @pstaabp's comment.  You missed a case with the mysql_enable_utf8mb4 settings.

taniwallach:
@pstaabp - Thanks for noticing the omission. I think the second commit fixes it.

I tested only so far as making sure that an `bin/update-OPL` run would start working with the change using both drivers.

I ran under Docker once with each of the options:
  - `WEBWORK_DB_DSN: DBI:mysql:webwork:db:3306` 
  - `WEBWORK_DB_DSN: DBI:MariaDB:database=webwork;host=db;port=3306` 
and in both cases the code started to report processing PG files.

(I used `bin/restore-OPL-tables.pl` to quickly restore the tables.)

@pstaabp and @drgrice1 - If one of you would please do a quick test of `OPL-update` and remove the `Do Not Merge` label...

taniwallach:
> you should add comments above the `$database_dsn` variable in site.conf explaining how to use this, and the describe format that variable should be in.

Being added.

drgrice1:
> ... you should add comments above the `$database_dsn` variable in site.conf explaining how to use this, and the describe format that variable should be in.

Don't forget this.

drgrice1:
Cross posting!

taniwallach:
> Cross posting!

My fault, I should have recalled the second issue before pushing the changes to my branch.

Note also the `pdflatex` default setting change.

drgrice1 (approved):
I think the documentation you added is decent.  You might make a comment about case.  Your listed formats have lower case dbi, and the examples are upper case.  Perhaps point out where case matters (I believe this is only the database name `webwork`).  Or just make them all consistent to prevent confusion on this matter.

I think this pull request looks good other than that.

conf/site.conf.dist
```diff
@@ -176,8 +179,15 @@ $externalPrograms{mysqldump} ="/usr/bin/mysqldump";
...
 # server you do not need to change this first value.
-# The format is dbi:mysql:[databasename] for databases on the local machine
-# For a remote database the format is dbi:mysql:[databasename]:[hostname]:[port]
+
+# When using the DBD:mysql driver the format is dbi:mysql:[databasename]:[hostname]:[port]
+#    example:  DBI:mysql:webwork:db:3306
+#    ** Only uses colons between fields.
+#    For a database on the local machine the format is dbi:mysql:[databasename]
+# When using the DBD::MariaDB driver the format is dbi:MariaDB:database=dbname;host=hostname;port=portnum
+#    example:  DBI:MariaDB:database=webwork;host=db;port=3306
```
dlglin, line 188: This will break some of the OPL scripts.  For example https://github.com/openwebwork/webwork2/blob/c75fe032c7c174f39fadddfffe5e35728f88395b/bin/load-OPL-global-statistics.pl#L55 is expecting the dsn in the dbi:mysql:[databasename]:[hostname]:[port]
 format.  DBD:MariaDB also supports that format, so we should stick with it for consistency.

I think the better solution going forward would be to have separate config variables for hostname, port and DBname, and then let WW construct the dsn, but that is a more significant change.

dlglin: Looking at the documentation for DBD::mysql and DBD::MariaDB, they both show the same format for the database DSN, which is the one @taniwallach is using.

I don't like having the conditional in the OPL scripts.  Using either DSN format will work with either driver wherever DBI is used, so if someone uses the "wrong" format, then the only thing that will break are the OPL scripts, which is hard to diagnose.

Since the perl docs show the "DBI:mysql:database=$database;host=$hostname;port=$port" format for both drivers, I think it would be better to use that format everywhere.

My recommended fix would be something like the following in site.conf
$database_host="localhost";
$database_port="3306";
$database_name="webwork";
$database_driver="MariaDB";
$database_dsn="DBI:$database_driver:database=$database_name;host=$database_host;port=$database_port";

Since the $database_dsn variable is still defined it shouldn't affect anything that uses DBI, but it allows anything that calls mysql clients directly (like the OPL scripts) to use those other variables rather than trying to parse the DSN.

drgrice1:
I noted the pdflatex change.  It is a bit out of place here, but that is fine.

I agree with @dlglin about things being overridden that shouldn't be.  All options in site.conf.dist and localOverrides.conf.dist should be commented out unless it is an option that must be set.  This literally means that everything should be commented out in localOverrides.conf.dist.

conf/site.conf.dist
drgrice1: I agree with dlglin on this.  This would require a change in the upgrade documentation from previous releases to ww-2.16.

dlglin:
I’m working on a PR to clean up localOverrides.conf.dist so that it is back to having everything commented out.

On Mar 10, 2021, at 10:34 AM, Glenn Rice ***@***.******@***.***>> wrote:

[△EXTERNAL]




I noted the pdflatex change. It is a bit out of place here, but that is fine.

I agree with @dlglin<https://github.com/dlglin> about things being overridden that shouldn't be. All options in site.conf.dist and localOverrides.conf.dist should be commented out unless it is an option that must be set. This literally means that everything should be commented out in localOverrides.conf.dist.

conf/site.conf.dist
taniwallach: > I don't like having the conditional in the OPL scripts. Using either DSN format will work with either driver wherever DBI is used

AFAIK The new `DBD::MariaDB` does not support the old format for DBname, host and port seperated by colons.
See:
  - https://metacpan.org/pod/DBD::MariaDB

> Since the perl docs show the "DBI:mysql:database=$database;host=$hostname;port=$port" format for both drivers, I think it would be better to use that format everywhere.
> 
> My recommended fix would be something like the following in site.conf
> $database_host="localhost";
> $database_port="3306";
> $database_name="webwork";
> $database_driver="MariaDB";
> $database_dsn="DBI:$database_driver:database=$database_name;host=$database_host;port=$database_port";

Working on this + some other changes to `site.conf.dist` and `defaults.conf` based on your feedback.

> Since the $database_dsn variable is still defined it shouldn't affect anything that uses DBI, but it allows anything that calls mysql clients directly (like the OPL scripts) to use those other variables rather than trying to parse the DSN.

Will try to work on having the scripts access these values directly (and get rid of the DSN parsing).

taniwallach: I tried to replace parsing of the DSN where possible = where the new variables in the course environment are available.

However, `lib/WeBWorK/DB/Schema/NewSQL/Std.pm` and `lib/WeBWorK/Utils/CourseManagement/sql_single.pm` do not have access a course environment, so the parsing of the DSN (obtained from `$self->{driver}{source}`) is still needed in those files.

taniwallach:
The changes suggested above by @dlglin in the review comments chain were implemented in the additional commit.

As remarked there `lib/WeBWorK/DB/Schema/NewSQL/Std.pm` and `lib/WeBWorK/Utils/CourseManagement/sql_single.pm` do not have access a course environment, so the parsing of the DSN (obtained from `$self->{driver}{source}`) is still needed in those files.

---

In not totally certain that `$database_debug = 0;` should be moved to `site.conf.dist` from where it was provided (commented out) in `defaults.conf`. However, having it as something to potentially change in `defaults.conf` seems to violate how that file is supposed to be used.

Since this setting is apparently is only used in `conf/database.conf.dist` which is read by 
```
include( "./conf/database.conf.dist");  # always include database.conf.dist
```
in `defaults.conf`, but that file earlier has
```
include("conf/site.conf");
```
I think the new location change is OK.

---

@mgage - Is there a real reason that `conf/database.conf.dist` has a `.dist` in the name. It does not seem to be intended for modification?

taniwallach:
Notes for anyone testing in Docker.

I needed to add
```
          PerlPassEnv WEBWORK_DB_DRIVER
          PerlPassEnv WEBWORK_DB_NAME
          PerlPassEnv WEBWORK_DB_HOST
          PerlPassEnv WEBWORK_DB_PORT
          PerlPassEnv WEBWORK_DB_USER
```
to `webwork.conf` 
and add an environment setting
```
     WEBWORK_DB_DRIVER: MariaDB
```
to the `docker-compose.yml` (as the `WEBWORK_DB_DSN` variable no longer suffices for what it did before).

These issues will need to be addressed in additional commits to https://github.com/openwebwork/webwork2/pull/1242

taniwallach:
@drgrice1 Is there something else to change here. I think it now all uses the new approach.

conf/site.conf.dist
```diff
@@ -173,12 +176,30 @@ $externalPrograms{mysqldump} ="/usr/bin/mysqldump";
...
+
+# Select the desired DB driver:
+#$database_driver="mysql";
+$database_driver="MariaDB";
+
+$database_host="localhost";
+$database_port="3306";
+$database_name="webwork";
+
+$database_dsn="DBI:$database_driver:database=$database_name;host=$database_host;port=$database_port";
```
drgrice1, line 200: These defaults do not work.  If the host is `localhost`, then the port can not be specified.  The following error is shown:

`error instantiating DB driver WeBWorK::DB::Driver::SQL for table set_version: DBI connect('database=webwork;host=localhost;port=3306','webworkWrite',...) failed: Connection error: port cannot be specified when host is localhost or embedded at /opt/webwork/webwork2/lib/WeBWorK.pm line 317.`

So the default `$database_dsn` will need to be

`$database_dsn="DBI:$database_driver:database=$database_name"`

This may include the host, but it is not necessary.

pstaabp:
This is probably a documentation question, but I updated site.conf by copying the lines directly from site.conf.dist as
```
$database_driver="MariaDB";

$database_host="localhost";
$database_port="3306";
$database_name="webwork";

$database_dsn="DBI:$database_driver:database=$database_name;host=$database_host;port=$database_port";
```

and am currently getting the error: 
```
error instantiating DB driver WeBWorK::DB::Driver::SQL for table locations: DBI connect('database=webwork;host=localhost;port=3306','webworkWrite',...) failed: Connection error: port cannot be specified when host is localhost or embedded at /XXX/webwork2/lib/WeBWorK/DB/Driver/SQL.pm line 79. at /XXX/webwork2/lib/WeBWorK.pm line 317.
```

If instead, I replace the last line above with:
```
$database_dsn="DBI:$database_driver:database=$database_name";
```

it seems to work.   Either we need to have the default be the latter value of `$database_dsn` or maybe have some check or additional documentation.

pstaabp:
Additionally, I pulled in develop with the `updateOPLextras.pl` code and that doesn't work.  I'll fix that and send a PR to this branch.

drgrice1 (commented):
I think @pstaabp commented on all of the problems that I am seeing.

drgrice1:
@pstaabp:  You should add the verbosity flag to the help in `updateOPLextras.pl`.

taniwallach:
>```
> $database_dsn="DBI:$database_driver:database=$database_name;host=$database_host;port=$database_port";
> ```
> 
> and am currently getting the error:
> 
> ```
> error instantiating DB driver WeBWorK::DB::Driver::SQL for table locations: DBI connect('database=webwork;host=localhost;port=3306','webworkWrite',...) failed: Connection error: port cannot be specified when host is localhost or embedded at /XXX/webwork2/lib/WeBWorK/DB/Driver/SQL.pm line 79. at /XXX/webwork2/lib/WeBWorK.pm line 317.
> ```
> 
> If instead, I replace the last line above with:
> 
> ```
> $database_dsn="DBI:$database_driver:database=$database_name";
> ```
> 
> it seems to work. Either we need to have the default be the latter value of `$database_dsn` or maybe have some check or additional documentation.

  - It turns out that `localhost` needs a different format.  See:  https://metacpan.org/pod/DBD::MariaDB#port 
  - On Docker the database is in a different Docker container, so the DSN format was OK for that setting.
  - For `localhost` there is the option of using a Unix socket to `TCP` to 127.0.0.1, so I added a new config switch to select between those 2 modes when the `$database_host` is `localhost`.

On my machine, using Docker things worked before, and work after the new change.

@pstaabp and @drgrice1 I hope it will now work on your setups with the DB on `localhost`.

---

In order to work around the recent changes to develop and overcome a merge conflict, I merged develop into this branch again, and the fixed the conflict.

conf/site.conf.dist
taniwallach: I hope this is now resolved by the conditional code managing the definition of `$database_dsn` when `database_host` is `localhost`. That needs to be tested by someone with such a setup.

drgrice1:
Read the documentation that you referenced carefully.  You only need to specify localhost via the loopback IP address if you want to access the database via TCP.  That is not something we want as it is a slow route to access the database on the local machine.  To access the database on the local machine directly you should not specify localhost or the port at all.  This introduces an inefficient way of accessing the database for those not using docker.

I am sorry, but this still needs work.

taniwallach:
> Read the documentation that you referenced carefully. You only need to specify localhost via the loopback IP address if you want to access the database via TCP. That is not something we want as it is a slow route to access the database on the local machine. To access the database on the local machine directly you should not specify localhost or the port at all. This introduces an inefficient way of accessing the database for those not using docker.
> 
> I am sorry, but this still needs work.

I tried to, and attempted to cover all the relevant cases.

Other parts of the WW code expected a `hostname` and `port` which used to be pulled out of the DSN and get fallback defaults.

I think the current approach, suggested by @dlglin and implemented here, of requiring all those settings to be provided as single settings in `site.conf` file makes lots of sense, but we then need to build the correct DSN for the various cases.

By default, with a DB on `localhost` it uses the Unix socket, as the recommended code sets `$database_use_socket_if_localhost=1;`. 

However, I coded an option to change to using the TCP connection  (possibly needed in some strange settings, such as if the local port is really a tunnel to somewhere else), which requires **modifying** that setting, and which will then force the use of `127.0.0.1` and  the provided port number, which seems to be what the driver documentation requires should such be desired.

```
# For a DB on localhost - default to using Unix socket.
# Change to 0 to use a TCP connection to 127.0.0.1.
$database_use_socket_if_localhost=1;

if ( $database_host eq "localhost" ) {
        if ( $database_use_socket_if_localhost ) {
                $database_dsn="DBI:$database_driver:database=$database_name";
        } else {
                $database_dsn="DBI:$database_driver:database=$database_name;host=127.0.0.1;port=$database_port";
        }
} else {
        $database_dsn="DBI:$database_driver:database=$database_name;host=$database_host;port=$database_port";
}
```

Feel free to propose an alternate mechanism to allow all these cases to work depending on the piecewise settings made.

One option would be to drop the internal conditional, and just provide a commented out alternate line to use to get the TCP connection `$database_dsn` for `localhost`. Which approach is easier for an end-user to notice and use is debatable.

drgrice1:
I apologize.  I didn't see what that option was doing, or ready the comment carefully enough there.  That is good then.  I will review this more carefully soon.

Alex-Jordan:
I am trying something unrelated to testing for this release on the server `https://webwork-dev.aimath.org`. And I am also getting the error message Peter reported. 

> error instantiating DB driver WeBWorK::DB::Driver::SQL for table locations: DBI connect('database=webwork;host=localhost;port=3306','webworkWrite',...) failed: Connection error: port cannot be specified when host is localhost or embedded at /XXX/webwork2/lib/WeBWorK/DB/Driver/SQL.pm line 79. at /XXX/webwork2/lib/WeBWorK.pm line 317.

(If you visit `https://webwork-dev.aimath.org/webwork2/admin` you might see it.)

But I am getting this on a pure `develop` (bf11e63), not this unmerged branch. The only thing I did change is `VERSION` to 2.16 for a certain PreTeXt tool to work as expected after sniffing it as 2.16.

Does this make sense to anyone?

Alex-Jordan:
Nevermind. My error message at https://webwork-dev.aimath.org/webwork2/admin is not the same.

taniwallach:
> I apologize. I didn't see what that option was doing, or ready the comment carefully enough there. That is good then. I will review this more carefully soon.

We're all trying to do quite a bit at present... at the 11-th hour for this RC.
Sorry if I'm getting snappy. It's quite late here.

pstaabp:
I'll try to review early tomorrow.

drgrice1:
I am re-targeting this to the WeBWorK-2.16 branch.

bin/dump-OPL-tables.pl
```diff
@@ -100,7 +97,17 @@
...
-`$mysqldump_command --host=$host --port=$port --user=$dbuser --default-character-set=$character_set $db $OPL_tables_to_dump  > $prepared_OPL_tables_file`;
+# Conditionally add --column-statistics=0 as MariaDB databases do not support it
+# see: https://serverfault.com/questions/912162/mysqldump-throws-unknown-table-column-statistics-in-information-schema-1109
+#      https://github.com/drush-ops/drush/issues/4410
+
+my $column_statistics_off = "";
+my $test_for_column_statistics = `$mysqldump_command --help | grep 'column-statistics'`;
+if ( $test_for_column_statistics ) {
+  $column_statistics_off = " --column-statistics=0 ";
+}
```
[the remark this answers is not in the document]
taniwallach: Yes - this occurs.

In Docker running this branch I did:
```
cd /opt/webwork/webwork2/bin
./dump-OPL-tables.pl
cp dump-OPL-tables.pl dump-OPL-tables.pl2
vim dump-OPL-tables.pl2
# force $test_for_column_statistics = 0;
diff dump-OPL-tables.pl dump-OPL-tables.pl2
./dump-OPL-tables.pl2
rm dump-OPL-tables.pl2
/usr/bin/mysqldump --version
```
and get
```
root@localhost:~# cd /opt/webwork/webwork2/bin
root@localhost:/opt/webwork/webwork2/bin# ./dump-OPL-tables.pl
OPL path seems to be /opt/webwork/libraries/webwork-open-problem-library/
Dumping OPL tables
OPL database dump created: /opt/webwork/libraries/webwork-open-problem-library//TABLE-DUMP/OPL-tables.sql
root@localhost:/opt/webwork/webwork2/bin# cp dump-OPL-tables.pl dump-OPL-tables.pl2
root@localhost:/opt/webwork/webwork2/bin# vim dump-OPL-tables.pl2
root@localhost:/opt/webwork/webwork2/bin# diff dump-OPL-tables.pl dump-OPL-tables.pl2
105c105
< my $test_for_column_statistics = `$mysqldump_command --help | grep 'column-statistics'`;
---
> my $test_for_column_statistics = 0;
root@localhost:/opt/webwork/webwork2/bin# ./dump-OPL-tables.pl2
OPL path seems to be /opt/webwork/libraries/webwork-open-problem-library/
Dumping OPL tables
mysqldump: Couldn't execute 'SELECT COLUMN_NAME,                       JSON_EXTRACT(HISTOGRAM, '$."number-of-buckets-specified"')                FROM information_schema.COLUMN_STATISTICS                WHERE SCHEMA_NAME = 'webwork' AND TABLE_NAME = 'OPL_DBsubject';': Unknown table 'COLUMN_STATISTICS' in information_schema (1109)
OPL database dump created: /opt/webwork/libraries/webwork-open-problem-library//TABLE-DUMP/OPL-tables.sql
root@localhost:/opt/webwork/webwork2/bin# rm dump-OPL-tables.pl2
root@localhost:/opt/webwork/webwork2/bin# /usr/bin/mysqldump --version
mysqldump  Ver 8.0.23-0ubuntu0.20.04.1 for Linux on x86_64 ((Ubuntu))
```
That error is what led to the changes your are pointing out.

lib/WeBWorK/Utils/CourseManagement/sql_single.pm
```diff
@@ -195,17 +195,38 @@ sub _get_db_info {
...
+	my $mysqldump = $self->{params}{mysqldump_path};
+	# Conditionally add column-statistics=0 as MariaDB databases do not support it
+	# see: https://serverfault.com/questions/912162/mysqldump-throws-unknown-table-column-statistics-in-information-schema-1109
+	#      https://github.com/drush-ops/drush/issues/4410
+
+	my $column_statistics_off = "";
+	my $test_for_column_statistics = `$mysqldump_command --help | grep 'column-statistics'`;
+	if ( $test_for_column_statistics ) {
+		$column_statistics_off = "[mysqldump]\ncolumn-statistics=0\n";
+warn "Setting in the temporary mysql config file for table dump/restore:\n$column_statistics_off\n\n";
```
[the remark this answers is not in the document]
taniwallach: Done. I also commented it out. It was added during testing and accidentally was left active.

pstaabp (approved):
Ran all the scripts listed here and generally navigation, especially the library browser looks great.

And a big bonus, I see speedup of a couple of factors using this branch.

taniwallach:
> One other thing that is unrelated to this pull request, and you don't need to fix it here. There is one thing that has bugged me in the `load-OPL-global-statistics.pl` script. That script uses the `DBI` package, connects to the database, and drops the `OPL_global_statistics` file. None of that is needed. The mysqldump command will drop that table in any case if it exists as the file `OPL_global_statistics.pl` has `` DROP TABLE IF EXISTS `OPL_global_statistics`; `` in it (as is standard for dump files created via `mysqldump`).

I agree that the default behavior of `mysqldump` is to include the `DROP TABLE` line before the `CREATE TABLE` command, so typically having the  `load-OPL-global-statistics.pl` script make the `DROP TABLE` call is not needed. However, since the `DROP TABLE` line from the dump file can be left out (when `mysqldump` is called with the `--skip-add-drop-table` option), I guess I  thought it was better to play it safe than depend on the `DROP TABLE` line being in future versions of the `OPL_global_statistics` file.

It can certainly be changed.

drgrice1:
> > One other thing that is unrelated to this pull request, and you don't need to fix it here. There is one thing that has bugged me in the `load-OPL-global-statistics.pl` script. That script uses the `DBI` package, connects to the database, and drops the `OPL_global_statistics` file. None of that is needed. The mysqldump command will drop that table in any case if it exists as the file `OPL_global_statistics.pl` has `` DROP TABLE IF EXISTS `OPL_global_statistics`; `` in it (as is standard for dump files created via `mysqldump`).
> 
> I agree that the default behavior of `mysqldump` is to include the `DROP TABLE` line before the `CREATE TABLE` command, so typically having the `load-OPL-global-statistics.pl` script make the `DROP TABLE` call is not needed. However, since the `DROP TABLE` line from the dump file can be left out (when `mysqldump` is called with the `--skip-add-drop-table` option), I guess I thought it was better to play it safe than depend on the `DROP TABLE` line being in future versions of the `OPL_global_statistics` file.
> 
> It can certainly be changed.

It is a safety measure I guess.   There is no need to change it.

drgrice1:
We have two approvals on this.  I tested it one more time to be sure.  I think it is time for this to go in.

drgrice1:
@taniwallach:  This pull request is causing issues with LTI authentication when the MariaDB driver is in use.  I get the error
`
error instantiating DB driver WeBWorK::DB::Driver::SQL for table user: DBI connect('database=webwork','webworkWrite',...) failed: Unknown attribute mysql_enable_utf8 at /opt/webwork/webwork2/lib/WeBWorK/DB/Driver/SQL.pm line 70. at /opt/webwork/webwork2/lib/WeBWorK.pm line 317.
`

drgrice1:
Here is the error I am getting on my production server running Ubuntu 18.04 if I enable the MariaDB driver:
`
error instantiating DB driver WeBWorK::DB::Driver::SQL for table problem_user: DBI connect('database=webwork','webworkWrite',...) failed: Can't connect to local MySQL server through socket '/tmp/mysql.sock' (2) at /opt/webwork/webwork2/lib/WeBWorK.pm line 317.
`

taniwallach:
> Here is the error I am getting on my production server running Ubuntu 18.04 if I enable the MariaDB driver:
> `error instantiating DB driver WeBWorK::DB::Driver::SQL for table problem_user: DBI connect('database=webwork','webworkWrite',...) failed: Can't connect to local MySQL server through socket '/tmp/mysql.sock' (2) at /opt/webwork/webwork2/lib/WeBWorK.pm line 317.`

Can you try using the settings to force a network socket connection (`$database_use_socket_if_localhost=1`) and see if that helps. Also check carefully what DSN is being used.

---

Google found a page in Chinese, which when translated
https://translate.google.com/translate?hl=en&sl=zh-TW&u=https://www.itread01.com/p/1151446.html&prev=search&pto=aue
suggests the following changes to `/etc/my.conf`:

> After the server starts normally, check /var/lib/mysql to automatically generate the mysql.sock file.
>
> But my problem is still not solved.
> 
> The problem is finally solved:
> Method: Modify `/etc/my.conf`:
```
[mysqld]
datadir=/usr/local/mysql/data
socket=/var/lib/mysql/mysql.sock

[mysql.server]
user=mysql
basedir=/usr/local/ mysql
```
> If there is not currently a section called `[client]`, add one at the bottom of the file and copy the `socket=` line under the `[mysqld]` section such as:
```
[client]
socket=/var/lib/mysql/mysql
```
> The sock found that it was still the case, and the execution of `/etc/init.d/mysql` start reported an error: 
> `Starting MySQLCouldn't find MySQL manager or server`
> is the mysqld service did not start, execute `/usr/local/mysql/bin/mysqld_safe &;` 
> the problem is solved.

taniwallach:
See also: http://manpages.ubuntu.com/manpages/groovy/man3/DBD::MariaDB.3pm.html
> `mariadb_socket`
>               It is possible to choose the Unix socket that is used for connecting to the
>               server. This is done, for example, with
>
> `               my $dsn = 'DBI:MariaDB:database=test;'`
> `                        . 'mariadb_socket=/var/run/mysqld/mysqld.sock';`
>
>               Usually there's no need for this option, unless you are using another location for
>               the socket than that built into the client.

It could be that we need to add another config setting to allow setting the socket in use, but I would have assumed that the Ubuntu package of `libdbd-mariadb-per` would have the default socket location set to match that of their packaging of the databases.

drgrice1:
`$database_use_socket_if_localhost=1` is what I was using (and want to use).  I did try `$database_use_socket_if_localhost=0` and that does work.

In any case the problem is that when cpan installs `DBD::MariaDB` it calls `mariadb_config` to "guess" the configuration of the mariadb server and client.  It seems that on Ubuntu 18.04 this does not return the correct thing.  It returns that the socket is `/tmp/mysqld.sock` even though the actual socket in use is `/var/run/mysqld/mysqld.sock`.  On Ubuntu 20.04 `mariadb_config`  returns the correct socket that is in use. I have not figured out how to make `mariadb_config` return the correct thing yet.  The configuration files are set up correct in `/etc/mysql`.  `mariadb_config` doesn't seem to be looking at those.

I saw the `mariadb_socket` option in the DBI string also.  It should be possible to avoid that.

Note, this is not a problem with this pull request.  It is just Ubuntu 18.04's setup of the mariadb.  It would be nice to figure out how to use MariaDB in this case for release documentation purposes.

taniwallach:
> `$database_use_socket_if_localhost=1` is what I was using (and want to use). I did try `$database_use_socket_if_localhost=0` and that does work.
> 
> In any case the problem is that when cpan installs `DBD::MariaDB` it calls `mariadb_config` to "guess" the configuration of the mariadb server and client. It seems that on Ubuntu 18.04 this does not return the correct thing. It returns that the socket is `/tmp/mysqld.sock` even though the actual socket in use is `/var/run/mysqld/mysqld.sock`.

Maybe see if a symbolic link from `/tmp/mysqld.sock` to `/var/run/mysqld/mysqld.sock` will work.

> I saw the `mariadb_socket` option in the DBI string also. It should be possible to avoid that.

That might require overriding the system wide default set up incorrectly, which could break on OS package upgrade.

Setting the DSN on the machine in a custom manner might be the simplest solution for now.

> It would be nice to figure out how to use MariaDB in this case for release documentation purposes.

Yes. We just need to make sure to remember to put it in.

drgrice1:
> Maybe see if a symbolic link from `/tmp/mysqld.sock` to `/var/run/mysqld/mysqld.sock` will work.

This doesn't work.  I don't think file system links work for sockets.
 
> > I saw the `mariadb_socket` option in the DBI string also. It should be possible to avoid that.
> 
> That might require overriding the system wide default set up incorrectly, which could break on OS package upgrade.

I am not sure what you mean by this?  The `mariadb_socket` option is just another argument you add to the DBI string used by webwork.  It doesn't require overriding the system default, and won't break on OS package upgrade (other than perhaps needing to change the setting).  This is just telling webwork how to connect.

> Setting the DSN on the machine in a custom manner might be the simplest solution for now.

I also don't know what you mean by this.  This seems unrelated to the issue.

taniwallach:
> > Maybe see if a symbolic link from `/tmp/mysqld.sock` to `/var/run/mysqld/mysqld.sock` will work.
> 
> This doesn't work. I don't think file system links work for sockets.

Sorry, I was not sure, but it was worth trying.

> > Setting the DSN on the machine in a custom manner might be the simplest solution for now.
> 
> I also don't know what you mean by this. This seems unrelated to the issue.

Try something like:
```
$database_dsn="DBI:$database_driver:database=$database_name"
     . 'mariadb_socket=/var/run/mysqld/mysqld.sock';
```
as a manual setting after the regular block in `site.conf`.

I know you wanted to avoid this, and for Ubuntu 20.04 it apparently is not needed, but it should hopefully work.

drgrice1:
Oh, you meant the database DSN.  I was thinking the general network DSN.  Sorry.

Yeah, that works for general access, but fails when creating a course archive (and probably restoring one).  It messes up the parsing of the database_dsn string that is done in the webwork code.  The apache2 error.log shows messages about an unknown host `mariadb_socket=/var/run/mysqld/mysqld.sock`.

drgrice1:
Wait, this is a problem with this pull request.  Course archiving is broken now in general.  Even with the mysql driver.  I am getting errors like:
`Warning: Failed to dump table course_id_setting' with command '2>&1 /usr/bin/mysqldump --defaults-file=/tmp/NE7XaluFC1 webwork course_id_setting > /opt/webwork/courses/course_id/DATA/mysqldump/setting.sql' (exit=2 signal=0 core=0): mysqldump: Got error: 2005: "Unknown MySQL server host 'mysql' (-3)" when trying to connect`

drgrice1:
This happens on Ubuntu 20.04 also.

drgrice1:
Anytime you are using the mysql DBD driver course archiving is broken.  The MariaDB driver works with this on Ubuntu 20.04.

taniwallach:
> Oh, you meant the database DSN. I was thinking the general network DSN. Sorry.
> 
> Yeah, that works for general access, but fails when creating a course archive (and probably restoring one). It messes up the parsing of the database_dsn string that is done in the webwork code. The apache2 error.log shows messages about an unknown host `mariadb_socket=/var/run/mysqld/mysqld.sock`.
>
> Wait, this is a problem with this pull request. Course archiving is broken now in general. Even with the mysql driver. I am getting errors like:
> `Warning: Failed to dump table course_id_setting' with command '2>&1 /usr/bin/mysqldump --defaults-file=/tmp/NE7XaluFC1 webwork course_id_setting > /opt/webwork/courses/course_id/DATA/mysqldump/setting.sql' (exit=2 signal=0 core=0): mysqldump: Got error: 2005: "Unknown MySQL server host 'mysql' (-3)" when trying to connect`

 I have archived courses with older versions of the code from this PR, but using a "remote" database where `host` and `port` are both set.

The changes in the PR were supposed to replace most of the cases where the database DSN is being parsed, but there are 2 places where that remained necessary.
  - `lib/WeBWorK/DB/Schema/NewSQL/Std.pm`
  - `lib/WeBWorK/Utils/CourseManagement/sql_single.pm`
 where new code manually does it for the new driver. Now that we have additional DSN formats (for socket use on localhost) - we have a problem which was not anticipated. The older code used `DBD::mysql->_OdbcParse()` which also was a hack and did not address local socket access to the database.

I'm checking now with the most recent code.

drgrice1:
Note that this just generally fails.  I was testing without the mariadb_socket option added in, but with the mysql driver enabled.

taniwallach:
> Note that this just generally fails. I was testing without the mariadb_socket option added in, but with the mysql driver enabled.

I can archive courses (in Docker WW) with either setting of `WEBWORK_DB_DRIVER`:
  - `WEBWORK_DB_DRIVER=mysql`
  - `WEBWORK_DB_DRIVER=MariaDB`
when using
```
WEBWORK_DB_HOST=db
WEBWORK_DB_USER=webworkWrite
WEBWORK_DB_PORT=3306
WEBWORK_DB_PASSWORD=the real secret password
WEBWORK_DB_NAME=webwork
```

I think the problem is `socket` specific, and we will need to fix it by adjusting the special code in
  - `lib/WeBWorK/DB/Schema/NewSQL/Std.pm`
  - `lib/WeBWorK/Utils/CourseManagement/sql_single.pm`
to handle alternate formats of the DSN (without `host=` and `port=`).

Did archiving work via `socket` with the older code. It also was designed to handle only `host` and `port`.

taniwallach:
Note the old code in those files had:
```
	# this is an internal function which we probably shouldn't be using here
	# but it's quick and gets us what we want (FIXME what about sockets, etc?)
```
so sockets were not covered before either.

drgrice1:
I am not using docker.  Sockets aren't the problem.  I say again, I am not using the socket option.

drgrice1:
This all worked with the old code without the socket option, and still should.

drgrice1:
This is definitely this pull request that caused this.  If I revert back to before this pull request, course archiving works as it should.

taniwallach:
> This is definitely this pull request that caused this. If I revert back to before this pull request, course archiving works as it should.

If you revert - getting this code back in will be a mess.

Please test on your setup with a network based database DSN setting and look at what changed in the code in the 2 files I just mentioned.

I looked at `/usr/lib/x86_64-linux-gnu/perl5/5.28/DBD/mysql.pm` on my machine to see what `_OdbcParse()` does...

drgrice1:
That is not what I am saying.  I am saying that if I locally revert to before this pull request, then archiving works.  I am also saying that this is a big problem that needs to be fixed, or it is possible that this will need to be reverted.

taniwallach:
I agree it needs to be fixed, and dropped other work to work on this now.

We need to understand how https://github.com/openwebwork/webwork2/blob/WeBWorK-2.16/lib/WeBWorK/DB/Schema/NewSQL/Std.pm is broken now... 

At present, the only hint I find of what is wrong is in
```
   mysqldump: Got error: 2005: "Unknown MySQL server host 'mysql' (-3)" when trying to connect
```
from the error message you posted.

I think it would help for you to see what is being placed in the temporary config file created by the archiving process in **both** settings, as well as the command line being executed.

Commenting out the
```
    $my_cnf->unlink_on_destroy(1);
```
in `sub _get_db_info` might help grab the file, and adding something to output `$dump_cmd` to a file will help save that.

drgrice1:
I am not sure what has changed, but before this pull request `_OdbcParse` did not set `$dsn{host}` or `$dsn{port}` with the dsn `dbi:mysql:webwork`, and now (hacking in that same dsn) it sets `$dsn{host}=mysql`, and `$dsn{port}=webwork`.

drgrice1:
If you change line 285 of `Std.pm` to
```perl
DBD::mysql->_OdbcParse($dsn, \%dsn, ['dbi', 'driver', 'database', 'host', 'port']);
```
archiving works again with the default settings in site.conf as you have them with the exception of `$database_driver="mysql"`.

drgrice1:
Another things that works is to just use the code you have above for `dbi:mariadb` also for `dbi:mysql`.

drgrice1:
By the way, this also works (either way) if the mariadb_socket is added into the dsn.

taniwallach:
> Another things that works is to just use the code you have above for `dbi:mariadb` also for `dbi:mysql`.

Then maybe that is the best approach. We need to make a call on what seem the best solution 

I did not want to change what was being done for the old driver, as it has case base handling for different ways of writing the database DSN.

Until the middle of this PR's review process DSN was set directly and then parsed in many places. Some (archive/unarchive courses) used the "old" driver's `_OdbcParse()` routine which allows `host=` or `hostname=` and `dbname=` or `db=`. (The code is below.) In order to avoid changing existing behavior unnecessarily and with little interest in maintaining more parsing code - I figures leaving the old approach for the old driver was safest. It turns out that this is a mistake for `localhost` databases for some cases.

I'll happily experiment later or tomorrow with a revised file via Docker and using  both driver options. However, I'm not set up to test a `localhost` DB.

---

> If you change line 285 of Std.pm to
> ...

That line should not be active for the "new" driver, but only when $database_driver="mysql".

I can't really figure out how to see what is happening on your machine. I think a detailed "trace" of the relevant variables and the values they get along the process may help.

Could you post the code of `DBD::mysql`'s internal `_OdbcParse` function on your box.

I have `/usr/lib/x86_64-linux-gnu/perl5/5.30/DBD/mysql.pm` in the Docker container, and it has:
```
sub _OdbcParse($$$) {
    my($class, $dsn, $hash, $args) = @_;
    my($var, $val);
    if (!defined($dsn)) {
	return;
    }
    while (length($dsn)) {
	if ($dsn =~ /([^:;]*\[.*]|[^:;]*)[:;](.*)/) {
	    $val = $1;
	    $dsn = $2;
	    $val =~ s/\[|]//g; # Remove [] if present, the rest of the code prefers plain IPv6 addresses
	} else {
	    $val = $dsn;
	    $dsn = '';
	}
	if ($val =~ /([^=]*)=(.*)/) {
	    $var = $1;
	    $val = $2;
	    if ($var eq 'hostname'  ||  $var eq 'host') {
		$hash->{'host'} = $val;
	    } elsif ($var eq 'db'  ||  $var eq 'dbname') {
		$hash->{'database'} = $val;
	    } else {
		$hash->{$var} = $val;
	    }
	} else {
	    foreach $var (@$args) {
		if (!defined($hash->{$var})) {
		    $hash->{$var} = $val;
		    last;
		}
	    }
	}
    }
}
```

taniwallach:
I think I see the issue - the new code uses `database=` in the DSN, but the `_OdbcParse` parser allows `db=` or `dbname=` but not `database=`.

Since we are not building the database DSN by "our standard" in `conf/site.conf.dist` and expect sites to adopt that standard in their `site.conf` I think the best solution is to drop the cases, and just use the local "parser" of the DSN in the 2 relevant files.

Install instructions and release notes should make it very clear that modifying the DNS too much could break archiving.

Maybe we should rewrite the "parser" to allow multiple orders and skipped parameters, now that we are using the format with explicit 
  - `database=`
  - `host=`
  - `port=`
so no longer have a need to implicitly assume a fixed order.

taniwallach:
A better fix, but possible very complicated to do would be to get the "raw" settings from `site.conf` somewhere the database code can access directly, and drop the special DSN parsing in these 2 files.

drgrice1:
I found it.  The difference is what is on line 274 in Std.pm before this pull request, and what you changed that to on line 281.  It was a substitution that stripped of the "dbi:mysql" before the pull request.  Now it is a match that does not strip off that part.  In sql_single.pm you left it as a substitution.  So if you change that back, you get the old behavior back and it works.

drgrice1:
I think that it would be best to do as you suggest, and just use the new code for both drivers and both files.

taniwallach:
> I found it. The difference is what is on line 274 in Std.pm before this pull request, and what you changed that to on line 281. It was a substitution that stripped of the "dbi:mysql" before the pull request. Now it is a match that does not strip off that part. In sql_single.pm you left it as a substitution. So if you change that back, you get the old behavior back and it works.

That seems to be the root bug... But as I posted in https://github.com/openwebwork/webwork2/pull/1284#issuecomment-809632124 I did not manage to trigger the issue...

PR #1163, opened 2020-12-13
Fix the WebworkWebservice permission handling

drgrice1:
It is probably not a good idea to allow those with proctor_quiz_login permissions the full capabilities of what the WebworkWebservice can do.  So now each method exposed by the WebworkWebservice can specify its permission level.

This also disables the webwork2_rpc and webwork2_wsdl endpoints and the WebworkSOAP module by default in the webwork.apache2.4-config.dist file.

It would probably be a good idea for someone else to look at the permissions that I chose for each method to ensure that they are correct.


lib/WebworkWebservice.pm
```diff
@@ -922,7 +923,7 @@ sub getCourseSettings {
 sub updateSetting {
 	my $class = shift;
 	my $in = shift;
-	my $self = $class->initiate_session($in);
+	my $self = $class->initiate_session($in, "modify_problem_sets");
```
drdrew42, line 926: Because this is modifying course settings, this doesn't seem right. There is only `create_and_delete_courses` and `modify_course_files` as far as course-level permissions are concerned. Create-and-delete seems inappropriate, as that's usually admin-only, but updating course settings should default to profs... so, `modify_course_files`?

lib/WebworkWebservice.pm
```diff
@@ -911,7 +912,7 @@ sub listSetUsers {
 sub getCourseSettings {
 	my $class = shift;
 	my $in = shift;
-	my $self = $class->initiate_session($in);
+	my $self = $class->initiate_session($in, "modify_problem_sets");
```
drdrew42, line 915: This is just a 'get' - so this seems like it's more on par with `access_instructor_tools`?

drdrew42 (approved):

lib/WebworkWebservice.pm
drgrice1: Yeah, I wasn't sure what to do with this one (and the other below).  I will change these to your suggested permissions.

conf/webwork.apache2.4-config.dist
```diff
@@ -189,27 +189,27 @@ if ($webwork_url eq "") {
 ####################################################################
```
pstaabp, line 189: Add the $permission argument to the documentation on line 184.

lib/WebworkWebservice.pm
```diff
@@ -225,6 +225,7 @@ sub initiate_session {    # close to being a "new" subroutine
     ###############################
```
pstaabp, line 225: Add information in the documentation on line 184 about the permission argument.

pstaabp:
I think the list of permissions looks good.  However, I haven't run any code.  Is there anything in particular I should test with this?

lib/WebworkWebservice.pm
drgrice1: The apache2 config file is not the correct place for documentation on this.  Also, the access point at line 184 is not the access point that is used by the WebworkWebservice.  The access point for the WebworkWebservice is at line 250 and currently has no documentation.  Furthermore, the permissions argument is not set by the web request.  It is an argument that is passed to the initiate_session method by the methods defined in WebworkWebservice.pm.  I will add documentation to the WebworkWebservice.pm file where the method is documented.

drgrice1:
The main things to test is that the library browser still works as it should,  the homework sets editor pages can still render problems as they should, the new problem grader still submits grades for someone with the correct permissions to do so, and that the webwork2/html2xml endpoint still works for rendering problems.  I think those are all of the places where the WebworkWebservice is currently used at this point.

pstaabp (approved):
Looks like it works as advertised.

Alex-Jordan (approved):
I did some testing of the things Glenn mentioned. I don't feel that it was exhaustive, but I did not have any issues.

Issue #1015, opened 2019-09-19
Upgrade javascript libraries


taniwallach:
We should consider upgrading some of the other javascript libraries that are utilized also.  For example, the versions of jQuery (1.7.2) and jQuery UI (1.9.0) that WeBWorK uses are quite old (released in 2012).

_Originally posted by @drgrice1 in https://github.com/openwebwork/webwork2/issues/1013#issuecomment-532908420_

taniwallach:
Right now WW uses the deprecated `bower` to help handle JavaScript dependencies and quite a few outdated JavsScript libraries.

Below are some mails about discussions to replace it...

We should consider whether the JavaScript dependencies can be kept out of the main Git repo and pulled in later as "sub-modules" or in some other manner during installation. 

---
On Sun, Aug 26, 2018 at 4:12 PM Nathan Wallach <taniwallach@gmail.com> wrote:

Dear Mike, Peter, and Jason,

The webwork2/htdocs/js/README_js_organization file refers to the use of bower to manage JavaScript libraries, so there should be a bower.json file somewhere... but the webwork2 git repository does not seem to have one.

I'm trying something which will require updating jQuery to a newer version (at least 1.9.1) and installing the jstree plugin which is available for bower (https://github.com/vakata/jstree.git 

Can one of you send me a relatively recent `bower.json` file?

Thanks,
Nathan
http://webwork.maa.org/pipermail/webwork-devel/2013-June/001069.html
http://webwork.maa.org/pipermail/webwork-devel/2013-June/001070.html

---

On Aug 26, 2018, at 11:00 AM, Nathan Wallach <tani@mathnet.technion.ac.il> wrote:

This is starting to look even more daunting to me... 
 1. jsquery had major upgrades to 1.9 and 3.0 which apparently required work to handle the changes to the framework, 
 2. there seem to be several versions of jQuery scattered in the htdocs/js tree and custom versions of jquery-ui 
 3. bower has been deprecated in favor of Yarn:   https://yarnpkg.com/en/docs/getting-started 
Anyone have a suggestion on how to work on migrating WW to a recent jQuery, and who is likely to be a helpful partner in trying to
work on that?

jQuery update notes:
 *  https://jquery.com/upgrade-guide/ 
 *  https://stackoverflow.com/questions/16514477/upgrade-jquery-1-4-2-to-1-9-1 
 *  https://stackoverflow.com/questions/45789166/migrating-jquery-from-1-4-to-3-0 
[...]

PR #516, opened 2014-11-06
This adds the description field.

aorinevo:
Fixes #316.

PR #1167, opened 2020-12-18
Make webwork use newer versions of jquery and jquery-ui installed via npm

drgrice1:
This removes all jquery and jquery-ui files from the git repository.  The files are now installed via npm.  Install npm from the Ubuntu repositories, and then run `npm install` from the `webwork2/htdocs` directory.

All of the third-party javascript and css files (those in js/vendor and css/vendor) should be worked into this scheme eventually.  At least as much as possible.

This will not work with the docker installation yet.  Edit:  This is not true anymore.

I am adding this as a pull request to make it more visible.  See https://github.com/openwebwork/webwork2/issues/1015.


pstaabp:
Trying to test this.  I switched to this branch and ran `npm install` and notice that the new packages are in the `node_modules` directory, but it's unclear that they are getting pulled in. 

Searching through the code, it only appears that the jquery-ui.min.css file is being pulled in--doing a search through all .pm files. 

Anything else that needs to be done or what am I missing?

drgrice1:
After you run `npm install` you should see jquery, jquery-ui-dist, and jquery-ui-themes in the node_modules directory.  These files are loaded in webwork everywhere that they were before.  That doesn't happen much in the perl library, but there are a few places.  Most of these are in the template files.

pstaabp:
I finally got a development machine running. First of all, here's a list of things I checked from the list of changed files that you have.  

- [ ] no pages on basic load seems to have any js errors
- [ ] I checked moving problem order in the homework set editor. no errors
- [ ] I checked achievements and they seem to pop open ok.
- [ ] Checked the library browser and no errors.

Here's some things I see: 

1. In the Classlist Editor, clicking on the pencil icon shows the error:
```
[Error] TypeError: null is not an object (evaluating 'document.getElementById("filter_select").onchange = classlist_add_filter_elements')
	(anonymous function) (jquery.min.js:2:31678)
```
This isn't a jQuery call, but I think it's getting handled by jQuery because being wrapped in a `$()` block to run on load.   This could be an error in the develop branch--I'll try to check this later. 

2. I luckily stumbled on a program that uses DraggableProof and it appears that we need to update that in the openwebwork-problem-library.  This may just need to point to the nestableSortable.js that you updated.    Are there any other problem macros to check?

pstaabp:
Just noticed that issue #516 in the pg repository mentions this.

mgage:
That’s my pull request to attempt to solve the problem.

drgrice1:
The problem with the pencil icon in the classlist editor was not created by this pull request.  You get the same error without this pull request.  Well, almost the same.  The difference is before this pull request the error comes through addOnLoadEvent.js.  Now the error comes through jQuery because I replaced the addOnLoadEvent with jQuery's ready function.

Note that in some cases I replaced the addOnLoadEvent with window.addEventListener("load", ...).  I checked all of those cases and determined which was the correct one to use.  Maybe I should not have used jQuery's ready function at all, and should have used DOMContentLoaded in those cases.  jQuery's ready function is a bit more convenient though as it is guaranteed to run even if the DOMContentLoaded event has already fired.  In any case the addOnLoadEvent.js code is obsolete, and has been since at least when IE 9 was released 9 years ago.

Furthermore, none of these changes have anything to do with the error coming from that pencil icon.  If I get a chance, I will try to fix that.

drgrice1:
Ahh, so the problem is that the elements that have the ids `filter_select` and `export_select_target` do not exist on the page that is opened when you click on the pencil icon (the user editor page).  I added the fix to this pull request.

taniwallach:
We should probably also fix `macros/CanvasObject.pl` in `pg` which has a line:
```
<script language="javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.4.2/jquery.min.js"></script>
```

drgrice1:
@taniwallach:  I don't think that macro was ever fully developed.  It is not used by anything in the OPL or Contrib, and I doubt it is used by anyone.  If someone knows otherwise, let me know, and give me a functional example pg file to test with.  For now, I am going to leave that file alone.

pstaabp:
I'm getting a couple of errors.  Not sure what changed: 
* In any problem view:
```
jQuery.Deferred exception: e.indexOf is not a function. (In 'e.indexOf(" ")', 'e.indexOf' is undefined) (2)
(anonymous function) — jquery.min.js:2:84839
(anonymous function) — problem.js:3
e — jquery.min.js:2:30011
(anonymous function) — jquery.min.js:2:30309
undefined
```
This seems to be from line 3 of problem.js (js/apps/Problem/problem.js) related to the achievements.  I had turned on achievements to test something, but seems to throw the error either way.  

I can't figure out why this error is thrown, though. 

* On the Homework Sets Editor, I'm now getting three `Can't find variable: addOnLoadEvent` errors for
   * `form_checker_hwsets.js`
   * `hwsets_handler.js`
   * `show_hide.js`

They are all of the form:
```
addOnLoadEvent(function (){
	document.getElementById("take_action").onclick = check_form_hmwk_sets;
});
```
perhaps we should change these to 
```
$( window ).load(function() {

});
```
instead.

drgrice1:
This sounds like a case of browser caching.  I recommend that you enable the feature (available in most browsers) of the developer tools to disable the network cache.  Clear your browsers cache and then try again.  The problems you are describing are common things I have seen when the browser is using a mix of not updated javascript from the cache, and new javascript from the pull request.

pstaabp:
That did it and explains why I hadn't seen it before. Thanks.

drgrice1:
That last commit to this pull request shouldn't need any testing.

pstaabp:
I am getting some errors related to flash applets.   Since flash is now dead, we just probably just pull those from the OPL.

pstaabp (approved):
This is looking pretty solid now.

drgrice1:
Are you trying to load the flash problems?  That will cause errors.  There are also java applet problems in the OPL.  They will also cause errors.  All browsers have now dropped support for flash and java.  We need to remove all of the flash and java support in webwork2 and pg.  Although that is general clean up that we can defer until after the release.  As long as you don't try to load those problems there is no issue.

drgrice1:
Another thing we need to discuss is how far back we want to go with browser support.  At this point, I think the only sane thing to do is to go no further back than IE 11.  Although, if no one is opposed, I would be up for dropping all IE support entirely.  That really opens up doors.

drgrice1:
By the way, if I am not mistaken, MathJax 3 requires at least IE 11.

drgrice1:
Please consider merging this so that we can move forward.

taniwallach:
I had tested this using the version from a while back. It is in use on a semi-public development server already for several days with no complaints from the people using that machine.

I am putting it up on my production servers tonight via a personal branch https://github.com/taniwallach/webwork2/tree/jquery-update_and_PR_1190 .

After it is up and running on the production server, I'll run some quick tests again and do the merge, unless someone asks to wait longer. 

My 2 cents: I think getting this merged is more important than being 100% certain that we did miss some loose ends, as until many people put it into use - we are likely to miss some places (ex. in some OPL problems) where there may be dependencies which remain to be fixed. We will need to handle any such hard-to-find issues as they get reported/discovered.

taniwallach (approved):
I have tested this on several servers, and have it in use on production servers using a private branch merging this PR and one other on top of develop.
I think we have done the best possible testing for such a large change, and am going to merge this now.

Alex-Jordan:
On the development server I am using, I have noticed that the page loads at first, and there is a noticeable flicker that I attribute to some CSS being applied. For example on the Homework Sets page, the table that lists homework sets is initially a more crude, less styled version, and then some fraction of a second later it is in its final form. Occasionally it takes as long as (I'd estimate) half a second.

Using git bisect, I attribute it to this pull request, which is not surprising given what this one is all about. 

Does anyone else see the flicker? Is there anything we can do about it?

drgrice1:
For me the pages load pretty much the same as they always have.  However, I have always seen what you are seeing on all pages in webwork.

Alex-Jordan:
To see the magnitude of the flicker compared to what I have if I rewind, I made a video: http://spot.pcc.edu/~ajordan/Flicker.mp4

I didn't narrate the video well and I cut out the transition. During the second half, I've rewound to before the jquery changes.

The behavior is consistent if I switch back and forth. That is, it's not faster on the earlier head because something cached stuff from the page load on the later head. (But the older stuff may be cached in the browser.)

I'm not complaining but this was a behavior change I thought worth reporting.

drdrew42:
tagging @apizer for putting this into the release notes -- this is a significant change to the install/upgrade process.

PR #1170, opened 2021-01-04
Add code to automatically remove spurious "published" column

mgage:
This automatically deletes the "published" column in several of the tables.  This column has not been used in a decade or more and is not created for recent courses. It does however hang around in tables that were created more than 10 years ago and produces annoying "Harmless:  published column is extra" messages. This deletes the columns and gets rid of the messages (after the deletion is reported).

The column "published" is not present no operation is performed.

To test unarchive an old course (more than 5 or 6 years old, perhaps more) from the admin course. 
Then upgrade that course.

### DIFF
--- a/lib/WeBWorK/Utils/CourseIntegrityCheck.pm
+++ b/lib/WeBWorK/Utils/CourseIntegrityCheck.pm
@@ -239,10 +239,16 @@ sub checkTableFields {
  	my %database_field_names =  map {${$_}[0]=>[$_]} @$result;             # drill down in the result to the field name level
                                                            #  result is array:  Field      | Type     | Null | Key | Default | Extra 
  	foreach my $field_name (sort keys %database_field_names) {
- 	    if ($field_name eq "published") {  #hack to stop warning about residual "published" fields.
- 	    	warn "Harmless: $table_name has extra column 'published' for older course. \n";
- 	        next;
- 	    }
+			if ($field_name eq "published") {  #hack to stop warning about residual "published" fields.
+				warn "Harmless: $table_name has extra column 'published' for older course. We'll try to delete it\n";
+				if ( $db->{$table}->can("drop_column_field") ) {
+					warn "dropping field $field_name in  table $table\n";
+					$db->{$table}->drop_column_field($field_name);
+				}
+				next;
+			} 
+ 	
+
  		my $exists = exists($schema_override_field_names{$field_name} );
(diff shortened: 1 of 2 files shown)


pstaabp:
@mgage I can't seem to find any old archived courses.   I did add a published column to the problem set table in the database, ran the upgrade courses, which showed the new message and then deleted the column.  

Anything else to test.  Should I try to get an old archived course and rerun?

taniwallach:
@mgage - Maybe package up a relatively empty course archive (at least one with no confidential personal information) with the old format for testing.

lib/WeBWorK/Utils/CourseIntegrityCheck.pm
```diff
@@ -274,11 +280,13 @@ sub updateTableFields {
...
-     	if ( $schema_obj->can("add_column_field") ) {
-     		$msg.= "Added column '$field_name' to table '$table'".CGI::br() if $schema_obj->add_column_field($field_name);
-     	}
+     	if ( $fieldStatus->{$field_name}->[0] == ONLY_IN_A) { 
+     		my $schema_obj = $db->{$table};
+			if ( $schema_obj->can("add_column_field") ) {
+				$msg.= "Added column '$field_name' to table '$table'".CGI::br() if $schema_obj->add_column_field($field_name);
+			}
+		}
+
```
drgrice1, line 289: Why is this code changed?  Is there a preference for this approach over the next unless approach?  Just curious.

Also, the first two lines have inconsistent indentation with the rest of the changed code.

lib/WeBWorK/Utils/CourseIntegrityCheck.pm
```diff
@@ -239,10 +239,16 @@ sub checkTableFields {
...
- 	    }
+			if ($field_name eq "published") {  #hack to stop warning about residual "published" fields.
+				warn "Harmless: $table_name has extra column 'published' for older course. We'll try to delete it\n";
+				if ( $db->{$table}->can("drop_column_field") ) {
+					warn "dropping field $field_name in  table $table\n";
+					$db->{$table}->drop_column_field($field_name);
+				}
+				next;
+			} 
+
```
drgrice1, line 250: There seems to be an extra tab level of indentation here relative to the other code.

drgrice1 (approved):
This looks good.  I don't have an old course to test this on, but the code looks like it should work for its purpose.

PR #1175, opened 2021-01-13
Fix variable used to configure old answer macros

dpvc:
The variable associated with the use-old-answer-checkers configuration was incorrect (it is in `specialPGEnvironmentVariables`, not `ansEvalDefaults`).  Not that anyone is using old answer checkers at this point, I expect.  Perhaps the option should be removed?  The new ones have been in place for over a decade.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1731,7 +1731,7 @@ $ConfigValues = [
 		  type => 'boolean'
 		},
 
-		{ var => 'pg{ansEvalDefaults}{useOldAnswerMacros}',
+		{ var => 'pg{specialPGEnvironmentVars}{useOldAnswerMacros}',
 		  doc => 'Use older answer checkers',
 		  doc2 => 'During summer 2005, a newer version of the answer checkers was implemented for answers which are functions and numbers.  The newer checkers allow more functions in student answers, and behave better in certain cases.  Some problems are specifically coded to use new (or old) answer checkers.  However, for the bulk of the problems, you can choose what the default will be here.  <p>Choosing <i>false</i> here means that the newer answer checkers will be used by default, and choosing <i>true</i> means that the old answer checkers will be used by default.',
 		  type => 'boolean'
--- a/lib/WeBWorK/Localize.pm
+++ b/lib/WeBWorK/Localize.pm
@@ -384,7 +384,7 @@ my $ConfigStrings = [
 		  type => 'boolean'
 		},
 
-		{ var => 'pg{ansEvalDefaults}{useOldAnswerMacros}',
+		{ var => 'pg{specialPGEnvironmentVars}{useOldAnswerMacros}',
 		  doc => x('Use older answer checkers'),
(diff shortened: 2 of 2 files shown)


drdrew42 (approved):

pstaabp:
This looks like a simple enough change.  Is there anything in particular that needs testing with this?

dpvc:
> Is there anything in particular that needs testing with this?

Well, I can think of two things:  first, that the configuration panel shows the correct default value (when the variable earlier in the `defaults.conf` file is changed)., and second, that the value of `$useOldAnswerMacros` within a problem correctly reflects the value set in the configuration panel.  (The latter may have been true without this change, as the `ansEvalDefaults` hash may also set variables, but the former would not.)

Alex-Jordan:
TLDR: it works as expected, testing what Davide suggested.

> Perhaps the option should be removed?

I would approve of that, for what it's worth. Maybe part of a larger cleanup of Course Config later though.

Does this mean that even if/when people wanted to use old answer macros on Math Objects, that in recent years they were unable to do that via the Course Config?


Prior to this change:

I'm on `develop`. In a course's Course Config, I see the default for "Use older answer checkers" is False and it's not changed from the default. In a course, `TEXT($useOldAnswerMacros);` returns `0`.

I change the setting in course config to True. This writes `$pg{ansEvalDefaults}{useOldAnswerMacros} = 1;` to `simple.conf`. Now `TEXT($useOldAnswerMacros);` still returns `0`. (Which is basically the problem that will be fixed.)

After this change:

I'm on `fix-config`. I cleared `simple.conf`. In a course's Course Config, I see the default for "Use older answer checkers" is False and it's not changed from the default. In a course, `TEXT($useOldAnswerMacros);` returns `0`.

I change the setting in course config to True. This writes `$pg{specialPGEnvironmentVars}{useOldAnswerMacros} = 1;` to `simple.conf`. Now `TEXT($useOldAnswerMacros);` returns `1`.

Also, changing `defaults.conf` so that `$pg{specialPGEnvironmentVars}{useOldAnswerMacros} = 1;`, I do see that reflected in the default value shown in Course Config.

Alex-Jordan (approved):

dpvc:
> Does this mean that even if/when people wanted to use old answer macros on Math Objects, that in recent years they were unable to do that via the Course Config?

I wasn't sure if that was the case or not, as I didn't check to see if the original setting actually got passed through, but according to your tests, it did not, so the answer to your question is "yes, people did not get the old answer macros".

But it was not for MathObjects, which always use the MathObject answer checkers.  The setting was for the traditional checkers, like `num_cmp()`.  Shortly after MathObejct were incorporated into WeBWorK, I rewrote the traditional checkers to call MathObjects internally.  That means all the answer checkers were using MathObjects (other than the string checker, and I think the complex checkers.  I did write one for complex numbers, (and it is in `pg/lib/Parser/Legacy/PGcomplexmacros.pm`, but I don't think it was ever activated).

The `$useOldAnswerMacros` turned the traditional checkers back into their original (non-MathObject) versions.  It didn't affect the checkers used by MathObjects themselves.

It may be  time to retire the traditional checker code, especially since the setting to enable it hasn't been working in a long time.  Nothing in the OPL seems to set this variable themselves, so I think it is safe to say the old ones are no longer being used.

dpvc:
> It's sort of like "use the original version of the old checkers".

Exactly!  That is a much better (and more concise) way of saying it.

drgrice1:
This looks fine to me.

I am in agreement with removing this option in the near future.

drgrice1 (approved):

Alex-Jordan:
3 approvals, no outstanding concerns. Merging.

Issue #1066, opened 2020-01-02
beginproblem()  fix delicate behavior of \pgmlSetup


mgage:
among other things this would fix a bug introduced in PR #1058

Here is the conversation after that PR was merged:
conf/snippets/ASimpleCombinedHeaderFile.pg 

@drgrice1
drgrice1 2 hours ago Member

Switching to PGML here seems to cause problems with hardcopy generation for problem sets that use this header file. The problem is that this causes the \pgmlSetup command to be used early in the tex file before it is defined.
@mgage
mgage 35 minutes ago Author Member

Good catch. I suspect that adding beginproblem() to header file will fix the hardcopy generation problem. It isn't completely natural since beginproblem() also outputs the
number of points a problem is worth. It's time to refactor beginproblem(). It's a hold over from WW1.9 when it was the only place we could initialize problems. It's performing too many functions.
 
it appears to be complicated. it would be worth asking @dpvc for advice.

This works fine:
```
loadMacros(
	"PG.pl",
	"PGbasicmacros.pl",
    "PGML.pl",
    "unionLists.pl",
	"PGcourse.pl",
);
#TEXT($BEGIN_ONE_COLUMN);
TEXT(beginproblem());
####################################################
#
# The item below printed out only when a hardcopy is made.
#
####################################################
```
If I comment out `TEXT(beginproblem())`  it breaks.
If I uncomment `TEXT($BEGIN_ONE_COLUMN)` and the matching $END_ONE_COLUMN
at the end of the file it also breaks.  I think this is because braces are introduced
into the TeX code which undefines \pgmlSetup where it is needed.  I seem to recall that this was a problem with scaffolding sections as well. It was important to enter certain header items in exactly the right order.

mgage:
additional comment.  I think that beginproblem()  all alone without surrounding TEXT() will 
initialize pgmlSetup without printing points for the problem which is what is desired.

mgage:
This version of the top of the header file works. The only caveat is that because beginproblem() is performing several functions it automatically assigns 0 pts to the header file and there does not seem to be an easy way to remove it.
```
loadMacros(
	"PG.pl",
	"PGbasicmacros.pl",
	#"source.pl", #uncommenting this will display "show source" button if that is permitted
	"PGcourse.pl",
);

TEXT($BEGIN_ONE_COLUMN);
loadMacros("PGML.pl");
TEXT(beginproblem() );

[...]

PR #1178, opened 2021-01-19
Remove PGML from ASimpleCombinedHeaderFile.pg.

drgrice1:
This was added in PR #1058 and breaks hardcopy generation.

### DIFF
--- a/conf/snippets/ASimpleCombinedHeaderFile.pg
+++ b/conf/snippets/ASimpleCombinedHeaderFile.pg
@@ -7,7 +7,6 @@ DOCUMENT();
 loadMacros(
 	"PG.pl",
 	"PGbasicmacros.pl",
-	"PGML.pl",
 	#"source.pl", #uncommenting this will display "show source" button if that is permitted
 	"PGcourse.pl",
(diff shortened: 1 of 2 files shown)


taniwallach (approved):
There was really no real need to PGML in the header file, and we had flagged the issue as a problem before. This changes things back to avoid the problem.

I think this minor correction should be merged ASAP.

taniwallach:
Mike opened an issue requesting a fix in https://github.com/openwebwork/webwork2/issues/1066. If someone eventually fixes the issue, we can then try PGML again in the header file.

@drgrice1 - @mgage  is apparently very busy with other things. I think we can merge without his review of such a minor change.

drgrice1:
I agree.  You can go ahead and merge it.  It has become cumbersome to continue to need to change the set hardcopy header file every time you are testing things with new sets on the develop branch.

mgage:
Go ahead and merge it.  I was hoping we could fix the hardcopy issue quickly.  I often use the header to list topics describing the homework set and the problems and the PGML markdown lists are very convenient.

PR #1185, opened 2021-02-01
Fix the problem urls in the emails sent using the "E-Mail Instructor" feedback page

drgrice1:
This was introduced in pull request #986.

That pull request switched from obtaining the set and problem ids cached
in the WeBWorK::ContentGenerator::Problem object to attempting to find
the set and problem ids from the url parameters.  The problem is that
the feedback page does not have those url parameters.


Alex-Jordan:
Thanks Glenn, that was fast. I'll review and test Monday.

drdrew42:
wow, yeah - I started working on a fix without realizing you'd already put this out. This looks good, thanks for cleaning up my mess!

drdrew42 (approved):

Alex-Jordan (approved):
Works as advertised.

Alex-Jordan:
I'm going to merge now. It's working on my server and @drdrew42 approves. That seems like enough.

PR #1186, opened 2021-02-01
resolve errors with emailed url

drdrew42:
As @Alex-Jordan pointed out in #986, some changes were made to the feedback form which broke the url that was sent when student's generated an email from a problem or a problem set. This PR resolves that error.


drdrew42:
Closed in favor of #1185

PR #1190, opened 2021-02-10
Revised "WeBWorK errors" page code + error.log records

taniwallach:
Various WW sites (including mine) have been subjected to security review using commercial scanning tools, which have reported various "security flaws" due to the information provided by "WeBWorK error" pages.

This PR is intended to allow a site to optionally:
  1. Minimize what is provided by the HTML"WeBWorK error" pages.
    - Controlled by the new `MIN_HTML_ERRORS` environment variable, set to true (`1`)to enable the new behavior.
  2. Store the data in the Apache `error.log` file in a JSON structure.
    - Controlled by the new `JSON_ERROR_LOG` environment variable, set to true (`1`)to enable the new behavior.

In all cases, the revised code assigns a UUID to each error record to allow matching up between the HTML error output (which students may get) and the data recorded in the Apache `error.log` file. This is particularly important when the `MIN_HTML_ERRORS` setting is turned on.

Setting the environment variables (and having the passed into the Apache mod_perl process) is necessary to enable the new options. I do not believe that this can be controlled via the `course.conf` file or similar configuration files, as errors may potentially occur when no course environment will be available, and none seems to be available to the code in `lib/Apache/WeBWorK.pm` where the main changes are.

Using the `MIN_HTML_ERRORS` should hopefully suffice for WeBWorK to no longer trigger significant warnings from the "WeBWorK error" pages.

Pay attention to the need to add `PerlPassEnv` lines for the new environment variables to the `/etc/apache2/conf-enabled/webwork.conf` file.

---

  * The first commit message has some details.
  * The second commit is just some white-space cleanup.

### DIFF
--- a/lib/Apache/WeBWorK.pm
+++ b/lib/Apache/WeBWorK.pm
@@ -37,9 +37,16 @@ use Date::Format;
 use WeBWorK;
 use Encode;
 use utf8;
-
+use JSON::MaybeXS;
+use UUID::Tiny  ':std';
 
 use constant MP2 => ( exists $ENV{MOD_PERL_API_VERSION} and $ENV{MOD_PERL_API_VERSION} >= 2 );
+
+# Should the minimal (more secure) HTML error output be used?
+use constant MIN_HTML_ERRORS => ( exists $ENV{"MIN_HTML_ERRORS"} and $ENV{"MIN_HTML_ERRORS"} );
+
+# Should Apache logs get JSON formatted record?
+use constant JSON_ERROR_LOG => ( exists $ENV{"JSON_ERROR_LOG"} and $ENV{"JSON_ERROR_LOG"} );
 
 # load correct modules
 BEGIN {
@@ -128,16 +135,30 @@ sub handler($) {
 		my $exception = $@;
 		
 		my $warnings = MP2 ? $r->notes->get("warnings") : $r->notes("warnings");
-		my $htmlMessage = htmlMessage($r, $warnings, $exception, @backtrace);
+		my $htmlMessage;
+		my $uuid = create_uuid_as_string(UUID_SHA1, UUID_NS_URL, $r->uri )
+		  . "::" . create_uuid_as_string(UUID_TIME);
+		my $time = time2str("%a %b %d %H:%M:%S %Y", time);
+
+		if ( MIN_HTML_ERRORS ) {
+			$htmlMessage = htmlMinMessage($r, $exception, $uuid, $time);
+		} else {
+			$htmlMessage = htmlMessage($r, $warnings, $exception, $uuid, $time, @backtrace);
+		}
 		unless ($r->bytes_sent) {
 			$r->content_type("text/html");
 			$r->send_http_header unless MP2; # not needed for Apache2
-			$htmlMessage = "<html><body>$htmlMessage</body></html>";
+			$htmlMessage = "<html lang=\"en-US\"><head><title>WeBWorK error</title></head><body>$htmlMessage</body></html>";
 		}
 		
 		# log the error to the apache error log
-		my $textMessage = textMessage($r, $warnings, $exception, @backtrace);
-		$log->error($textMessage);
+		my $logMessage;
+		if ( JSON_ERROR_LOG ) {
+			$logMessage = jsonMessage($r, $warnings, $exception, $uuid, $time, @backtrace);
+		} else {
+			$logMessage = textMessage($r, $warnings, $exception, $uuid, $time, @backtrace);
+		}
+		$log->error($logMessage);
 
 		$r->custom_response(FORBIDDEN,$htmlMessage);
 
@@ -186,7 +207,7 @@ sub backtrace {
 
 =over
 
-=item htmlMessage($r, $warnings, $exception, @backtrace)
+=item htmlMessage($r, $warnings, $exception, $uuid, $time, @backtrace)
 
 Format a message for HTML output reporting an exception, backtrace, and any
 associated warnings.
@@ -194,10 +215,10 @@ associated warnings.
 =cut
 
 sub htmlMessage($$$@) {
-	my ($r, $warnings, $exception, @backtrace) = @_;
-	
+	my ($r, $warnings, $exception, $uuid, $time, @backtrace) = @_;
+
 	# Warnings have html and look better scrubbed. 
-	
+
 	my $scrubber = HTML::Scrubber->new(
 	    default => 1,
 	    script => 0,
@@ -212,7 +233,7 @@ sub htmlMessage($$$@) {
 
 	$warnings = $scrubber->scrub($warnings);
 	$exception = $scrubber->scrub($exception);
-	
+
 	my @warnings = defined $warnings ? split m|<br />|, $warnings : ();  #fragile
 	$warnings = htmlWarningsList(@warnings);
 	my $backtrace = htmlBacktrace(@backtrace);
@@ -220,14 +241,13 @@ sub htmlMessage($$$@) {
 	# $ENV{WEBWORK_SERVER_ADMIN} is set from $webwork_server_admin_email in site.conf
 	# and $ENV{SERVER_ADMIN} which is set by ServerAdmin in httpd.conf is used as a backup
 	# if an explicit email address has not been set.
-	
+
 	$ENV{WEBWORK_SERVER_ADMIN} = ($ENV{WEBWORK_SERVER_ADMIN}) ?$ENV{WEBWORK_SERVER_ADMIN}:$ENV{SERVER_ADMIN};
 	$ENV{WEBWORK_SERVER_ADMIN}= $ENV{WEBWORK_SERVER_ADMIN}//''; #guarantee this variable is defined. 
 
 	my $admin = ($ENV{WEBWORK_SERVER_ADMIN}
 		? " (<a href=\"mailto:$ENV{WEBWORK_SERVER_ADMIN}\">$ENV{WEBWORK_SERVER_ADMIN}</a>)"
 		: "");
-	my $time = time2str("%a %b %d %H:%M:%S %Y", time);
 	my $method = htmlEscape( $r->method  );
 	my $uri = htmlEscape(  $r->uri );
 	my $headers = do {
@@ -235,22 +255,24 @@ sub htmlMessage($$$@) {
 		join("", map { "<tr><td><small>" . htmlEscape($_). "</small></td><td><small>" .
 		                htmlEscape($headers{$_}) . " </small></td></tr>" } keys %headers);
 	};
-	
+
 	return <<EOF;
 <div style="text-align:left">
- <h2>WeBWorK error</h2>
+ <h1>WeBWorK error</h2>
  <p>An error occured while processing your request. For help, please send mail
  to this site's webmaster $admin, including all of the following information as
  well as what what you were doing when the error occured.</p>
  <p>$time</p>
- <h3>Warning messages</h3>
+ <h2>Error record identifier</h3>
+ <p style="color: #dc2a2a"><code>$uuid</code></blockquote>
+ <h2>Warning messages</h3>
  <ul>$warnings</ul>
- <h3>Error messages</h3>
- <blockquote style="color:red"><code>$exception</code></blockquote>
- <h3>Call stack</h3>
+ <h2>Error messages</h3>
+ <blockquote style="color: #dc2a2a"><code>$exception</code></blockquote>
+ <h2>Call stack</h3>
    <p>The information below can help locate the source of the problem.</p>
    <ul>$backtrace</ul>
- <h3>Request information</h3>
+ <h2>Request information</h3>
  <table border="1">
   <tr><td>Method</td><td>$method</td></tr>
   <tr><td>URI</td><td>$uri</td></tr>
@@ -264,7 +286,67 @@ sub htmlMessage($$$@) {
 EOF
 }
 
-=item textMessage($r, $warnings, $exception, @backtrace)
+###############################################################################
+
+=item htmlMinMessage($r, $exception, $uuid, $time)
+
+Format a minimal message for HTML output reporting an error ID number, and NOT providing much
+additional data, which will instead be in the log files.
+
+=cut
+
+sub htmlMinMessage($$$@) {
+	my ($r, $exception, $uuid, $time) = @_;
+
+	# Warnings have html and look better scrubbed.
+
+	my $scrubber = HTML::Scrubber->new(
+	    default => 1,
+	    script => 0,
+	    comment => 0
+	    );
+	$scrubber->default(
+	    undef,
+	    {
+		'*' => 1,
+	    }
+	    );
+
+	$exception = $scrubber->scrub($exception);
+
+	# Drop any code reference from the error message
+	$exception =~ s/ at \/.*//;
+
+	# $ENV{WEBWORK_SERVER_ADMIN} is set from $webwork_server_admin_email in site.conf
+	# and $ENV{SERVER_ADMIN} which is set by ServerAdmin in httpd.conf is used as a backup
+	# if an explicit email address has not been set.
+
+	$ENV{WEBWORK_SERVER_ADMIN} = ($ENV{WEBWORK_SERVER_ADMIN}) ?$ENV{WEBWORK_SERVER_ADMIN}:$ENV{SERVER_ADMIN};
+	$ENV{WEBWORK_SERVER_ADMIN}= $ENV{WEBWORK_SERVER_ADMIN}//''; #guarantee this variable is defined.
+
+	my $admin = ($ENV{WEBWORK_SERVER_ADMIN}
+		? " (<a href=\"mailto:$ENV{WEBWORK_SERVER_ADMIN}\">$ENV{WEBWORK_SERVER_ADMIN}</a>)"
+		: "");
+
+	return <<EOF;
+<div style="text-align:left">
+ <h1>WeBWorK error</h2>
+ <p>An error occured while processing your request. For help, please send mail
+ to this site's webmaster $admin, including all of the following information as
+ well as what what you were doing when the error occured.</p>
+ <p>$time</p>
+ <h2>Error record identifier</h3>
+ <p style="color: #dc2a2a"><code>$uuid</code></blockquote>
+ <h2>Error messages</h3>
+ <blockquote style="color: #dc2a2a"><code>$exception</code></blockquote>
+
+</div>
+EOF
+}
+
+################################################################################
+
+=item textMessage($r, $warnings, $exception, $uuid, $time, @backtrace)
 
 Format a message for HTML output reporting an exception, backtrace, and any
 associated warnings.
@@ -272,7 +354,7 @@ associated warnings.
 =cut
 
 sub textMessage($$$@) {
-	my ($r, $warnings, $exception, @backtrace) = @_;
+	my ($r, $warnings, $exception, $uuid, $time, @backtrace) = @_;
 	
 	#my @warnings = defined $warnings ? split m/\n+/, $warnings : ();
 	#$warnings = textWarningsList(@warnings);
@@ -280,8 +362,48 @@ sub textMessage($$$@) {
 	my $backtrace = textBacktrace(@backtrace);
 	my $uri = $r->uri;
 	
-	return "[$uri] $exception\n$backtrace";
+	return "[$uuid] [$uri] $exception\n$backtrace";
 }
+
+=item jsonMessage($r, $warnings, $exception, $uuid, $time, @backtrace)
+
+Format a JSON message for log output reporting an exception, backtrace, and any
+associated warnings.
+
+=cut
+
+sub jsonMessage($$$@) {
+	my ($r, $warnings, $exception, $uuid, $time, @backtrace) = @_;
+
+	my @warnings = defined $warnings ? split m/\n+/, $warnings : ();
+
+	my %headers = MP2 ? %{$r->headers_in} : $r->headers_in;
+	# Was getting JSON errors for the value of "sec-ch-ua" in my testing, so remove it
+	my %headers2 = ( %headers,
+			 "sec-ch-ua" => "REMOVED",
+			);
+
+	return join("",
+		    "{ ", # Start object
+		    encode_json("Error record identifier"), ": ",
+		    encode_json($uuid), ", ",
+		    encode_json("Time"), ": ",
+		    encode_json($time), ", ",
+		    encode_json("Method"), ": ",
+		    encode_json($r->method), ", ",
+		    encode_json("URI"), ": ",
+		    encode_json($r->uri), ", ",
+		    encode_json("HTTP Headers"), ": ",
+		    encode_json( {%headers2} ), ", ",
+		    encode_json("Warnings"), ": ", jsonWarningsList(@warnings), ", ",
+		    encode_json("Exception"), ": ",
+		    encode_json(chomp $exception), ", ",
+		    encode_json("Backtrace"), ": ", jsonBacktrace(@backtrace),
+		    " }" # End object
+		   );
+}
+
+################################################################################
 
 =item htmlBacktrace(@frames)
 
@@ -312,6 +434,20 @@ sub textBacktrace(@) {
 	return join "\n", @frames;
 }
 
+
+=item jsonBacktrace(@frames)
+
+Converts a list of stack frames in a backtrace into a JSON array.
+
+=cut
+
+sub jsonBacktrace(@) {
+	my (@frames) = @_;
+	return encode_json( [ @frames ] );
+}
+
+################################################################################
+
 =item htmlWarningsList(@warnings)
 
 Formats a list of warning strings as list items for HTML output.
@@ -340,6 +476,19 @@ sub textWarningsList(@) {
 	}
 	return join "\n", @warnings;
 }
+
+=item jsonWarningsList(@warnings)
+
+Converts a list of warnings into a JSON array.
+
+=cut
+
+sub jsonWarningsList(@) {
+	my (@warnings) = @_;
+	return encode_json( [ @warnings ] );
+}
+
+################################################################################
 
 =item htmlEscape($string)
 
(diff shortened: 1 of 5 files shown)


taniwallach:
Here is a sample of the JSON record from an Apache `error.log` after passing it through `json_pp`.
  - Only the JSON portion of the `error.log` line should be passed into `json_pp`.
  - 2 parts of the `Cookie` record were replaced by the string `REMOVED` relative to what was really in the file.
  - The `sec-ch-ua` field is replaced by `removed` in the code, as the JSON output was broken when it was left there during my testing. Something in the data in that header was not being handled well by `encode_json` and I did not try to overcome the issue.
```
{
   "Backtrace" : [
      "in Carp::croak called at line 2325 of /opt/webwork/webwork2/lib/WeBWorK/DB.pm",
      "in WeBWorK::DB::validateKeyfieldValue called at line 1243 of /opt/webwork/webwork2/lib/WeBWorK/URLPath.pm",
      "in WeBWorK::URLPath::visitPathTypeNode called at line 1281 of /opt/webwork/webwork2/lib/WeBWorK/URLPath.pm",
      "in WeBWorK::URLPath::visitPathTypeNode called at line 1281 of /opt/webwork/webwork2/lib/WeBWorK/URLPath.pm",
      "in WeBWorK::URLPath::visitPathTypeNode called at line 1128 of /opt/webwork/webwork2/lib/WeBWorK/URLPath.pm",
      "in WeBWorK::URLPath::getPathType called at line 866 of /opt/webwork/webwork2/lib/WeBWorK/URLPath.pm",
      "in WeBWorK::URLPath::newFromPath called at line 145 of /opt/webwork/webwork2/lib/WeBWorK.pm"
   ],
   "Method" : "GET",
   "Error record identifier" : "edb672c7-9764-5817-9927-de629d69d08d::f820971e-6bc9-11eb-ac5b-985c78788a99",
   "Exception" : 1,
   "URI" : "/webwork2/Madar_devel/phase-plane-testin++g/",
   "Time" : "Wed Feb 10 20:01:16 2021",
   "Warnings" : [],
   "HTTP Headers" : {
      "Host" : "localhost:8080",
      "sec-gpc" : "1",
      "sec-ch-ua" : "removed",
      "Accept" : "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9",
      "Connection" : "keep-alive",
      "Sec-Fetch-Dest" : "document",
      "Cache-Control" : "max-age=0",
      "dnt" : "1",
      "Upgrade-Insecure-Requests" : "1",
      "Cookie" : "WeBWorKCourseAuthen.Madar_devel=tani%REMOVED; csrftoken=REMOVED",
      "Accept-Encoding" : "gzip, deflate, br",
      "Sec-Fetch-Mode" : "navigate",
      "sec-ch-ua-mobile" : "?0",
      "Sec-Fetch-User" : "?1",
      "Accept-Language" : "en-US,en;q=0.9,he;q=0.8",
      "User-Agent" : "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36",
      "Sec-Fetch-Site" : "none"
   }
}
```

lib/Apache/WeBWorK.pm
```diff
@@ -264,25 +286,125 @@ sub htmlMessage($$$@) {
...
+
+	# Drop any code reference from the error message
+	$exception =~ s/ at \/.*//;
+
+	# $ENV{WEBWORK_SERVER_ADMIN} is set from $webwork_server_admin_email in site.conf
+	# and $ENV{SERVER_ADMIN} which is set by ServerAdmin in httpd.conf is used as a backup
+	# if an explicit email address has not been set.
+
+	$ENV{WEBWORK_SERVER_ADMIN} = ($ENV{WEBWORK_SERVER_ADMIN}) ?$ENV{WEBWORK_SERVER_ADMIN}:$ENV{SERVER_ADMIN};
+	$ENV{WEBWORK_SERVER_ADMIN}= $ENV{WEBWORK_SERVER_ADMIN}//''; #guarantee this variable is defined.
```
drgrice1, line 325: This can be done in a single line with
```perl
$ENV{WEBWORK_SERVER_ADMIN} = $ENV{WEBWORK_SERVER_ADMIN} // $ENV{SERVER_ADMIN} // '';
```

taniwallach: Dates back to 2015: https://github.com/openwebwork/webwork2/commit/25e5901c0bbbe744a2cc17d6dd2dc5fcbec59881
Being fixed.

lib/Apache/WeBWorK.pm
```diff
@@ -212,45 +233,45 @@ sub htmlMessage($$$@) {
...
 		join("", map { "<tr><td><small>" . htmlEscape($_). "</small></td><td><small>" .
 		                htmlEscape($headers{$_}) . " </small></td></tr>" } keys %headers);
 	};
-	
+
 	return <<EOF;
 <div style="text-align:left">
- <h2>WeBWorK error</h2>
+ <h1>WeBWorK error</h2>
  <p>An error occured while processing your request. For help, please send mail
```
drgrice1, line 261: There seems to be a tag mismatch issue here.

taniwallach: Fixed. Sorry - sloppy search replace work (missed the "global" per line).

lib/Apache/WeBWorK.pm
drgrice1: I didn't realize this was just code clean up.  I was misinterpreting the diff.

lib/Apache/WeBWorK.pm
```diff
@@ -212,45 +233,45 @@ sub htmlMessage($$$@) {
...
 <div style="text-align:left">
- <h2>WeBWorK error</h2>
+ <h1>WeBWorK error</h1>
  <p>An error occured while processing your request. For help, please send mail
  to this site's webmaster $admin, including all of the following information as
  well as what what you were doing when the error occured.</p>
  <p>$time</p>
- <h3>Warning messages</h3>
+ <h2>Error record identifier</h2>
+ <p style="color: #dc2a2a"><code>$uuid</code></blockquote>
```
drgrice1, line 266: This ends a blockquote that seems to not have been started.  Should the <p> be a blockquote tag instead?

lib/Apache/WeBWorK.pm
```diff
@@ -264,25 +285,124 @@ sub htmlMessage($$$@) {
...
+
+	return <<EOF;
+<div style="text-align:left">
+ <h1>WeBWorK error</h1>
+ <p>An error occured while processing your request. For help, please send mail
+ to this site's webmaster $admin, including all of the following information as
+ well as what what you were doing when the error occured.</p>
+ <p>$time</p>
+ <h2>Error record identifier</h2>
+ <p style="color: #dc2a2a"><code>$uuid</code></blockquote>
```
drgrice1, line 337: This seems to be another case of the <p> </blockquote> mismatch.

taniwallach: mea culpa.
Being fixed in the next commit, and getting rid of the use of `blockquote` to do indentation in this file. Abused of the HTML tag and bad for accessibility.
Also did some more HTML cleanup for better accessibility compliance.

taniwallach:
Background to some of the recent changes.

IBM's accessibility checker (Chrome extension) from: https://www.ibm.com/able/toolkit/ reported:

```
Verify that <blockquote> is used only for quotations, not indentation
Needs review
Use <blockquote> only for quotations, not indentation
```

```
The page does not provide a way to quickly navigate to the main content (WAI-ARIA "main" landmark or a skip link)
```

The tool still reports a recommendation for requirement 1.3.1 Info and Relationships that A complex data table should have a 'summary' or an 'aria-describedby' that references an overview of the table even though the outer table has aria-labelledby="req_info_summary1" set. Filed a bug report at https://github.com/IBMa/equal-access/issues/320
It complains

drgrice1 (commented):
This pull request generally looks good.

Is there a simple way to make one of these errors occur so that I can test it better?  I know that I have seen these things at times, but I can't seem to make it happen when I need it to!

Edit:  Nevermind.  I found ways to force errors to occur.  Obviously you can just throw some "die" calls in somewhere.

drgrice1 (approved):
This looks good.  I tested this with various combinations of the MIN_HTML_ERRORS and JSON_ERROR_LOG environment variables.  It seems to me that using the MIN_HTML_ERRORS option without JSON_ERROR_LOG isn't wise.  You would lose the HTTP information that was previously displayed.

One thing to note is that if the MIN_HTML_ERRORS option is enabled, then instructors without admin access will not be able to obtain this information directly anymore.

taniwallach:
@drgrice1 Thanks.

> It seems to me that using the MIN_HTML_ERRORS option without JSON_ERROR_LOG isn't wise. You would lose the HTTP information that was previously displayed.

I think the error.log would still have the data, just in many lines, which makes it hard to handle.

> One thing to note is that if the MIN_HTML_ERRORS option is enabled, then instructors without admin access will not be able to obtain this information directly anymore.

Yes. At the level where the errors are being processed, there is no course environment, so logging per course would be somewhat difficult. It could be that a cron job or something else could parse the records being places into the error.log file and place those with an obvious relation to a specific course in a file under the course directory tree. It certainly bears consideration. In the old approach, teaching staff is still mostly dependent on the end-user forwarding the error messages. Many of my reports have screen-shots rather than a text form, so even now - pulling the text form out of the error.log is often helpful.

drgrice1:
> > It seems to me that using the MIN_HTML_ERRORS option without JSON_ERROR_LOG isn't wise. You would lose the HTTP information that was previously displayed.
> 
> I think the error.log would still have the data, just in many lines, which makes it hard to handle.

I tested this again with MIN_HTML_ERRORS set but JSON_ERROR_LOG not set, and I do not see the http request information (method, headers, cookies and such) in the error.log.  Without MIN_HTML_ERRORS that is displayed on the page.  With MIN_HTML_ERRORS and JSON_ERROR_LOG that information is in the error.log.  With MIN_HTML_ERRORS set but JSON_ERROR_LOG not set all I see is the backtrace and the referer url.

taniwallach:
> I tested this again with MIN_HTML_ERRORS set but JSON_ERROR_LOG not set, and I do not see the http request information (method, headers, cookies and such) in the error.log. Without MIN_HTML_ERRORS that is displayed on the page. With MIN_HTML_ERRORS and JSON_ERROR_LOG that information is in the error.log. With MIN_HTML_ERRORS set but JSON_ERROR_LOG not set all I see is the backtrace and the referer url.

Yes, as that was all that was logged in the past - the error message/warnings and the backtrace.

I will look soon into adding the additional information for the case where `MIN_HTML_ERRORS` is set but `JSON_ERROR_LOG` is not set.

taniwallach:
Code revised to add (in JSON format) data which was not in the `error.log` but had been on the more detailed HTML output.

While testing by requesting a problem of index beyond the end of an assignment I saw the following in the `error.log`.
```
Use of uninitialized value $path in pattern match (m//) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 399.
Use of uninitialized value $path in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 401.
```
which appear before the main error message `Can't use an undefined value as a HASH reference`.

I'm not trying to fix those excess error messages now, not to make the error reported more meaningful, but maybe in the future we should:
  1. trigger a more informative message for a problem number which does not exist, and 
  2. avoid the use of an uninitialized `$path` (ex. by adding `return 0 if ( !defined( $path ) );` near the top of `sub path_is_subdir`.
I'm not making the change to `sub path_is_subdir` in this PR as such a change it could have consequences elsewhere.

---
**Log message:**:
```
[Thu Feb 18 17:08:48.652865 2021] [perl:warn] [pid 95] [client 172.29.0.1:39314] [/webwork2/Madar_devel/junk1/7] Use of uninitialized value $path in pattern match (m//) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 399.
[Thu Feb 18 17:08:48.652915 2021] [perl:warn] [pid 95] [client 172.29.0.1:39314] [/webwork2/Madar_devel/junk1/7] Use of uninitialized value $path in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 401.
[Thu Feb 18 17:08:48.662767 2021] [perl:error] [pid 95] [client 172.29.0.1:39314] [f3c8adea-ad5d-5bf0-a3bb-268693a5ea66::33e9cc3e-71fb-11eb-bc86-980c1f751a5b] [/webwork2/Madar_devel/junk1/7] { "Error record identifier": "f3c8adea-ad5d-5bf0-a3bb-268693a5ea66::33e9cc3e-71fb-11eb-bc86-980c1f751a5b", "Time": "Thu Feb 18 17:08:48 2021", "Method": "GET", "URI": "/webwork2/Madar_devel/junk1/7", "HTTP Headers": {"Upgrade-Insecure-Requests":"1","Accept-Language":"en,en-US;q=0.5","Host":"localhost:8080","User-Agent":"Mozilla/5.0 (X11; Linux x86_64; rv:78.0) Gecko/20100101 Firefox/78.0","Accept":"text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8","Connection":"keep-alive","Cookie":"WeBWorKCourseAuthen.Madar_devel=tani%094TLzxx6UdzYXn7ghBD2BUsImldLQZ0qe%091613660615; default-theme=ngax","Accept-Encoding":"gzip, deflate","Cache-Control":"max-age=0"}, "Warnings": ["Use of uninitialized value $path in pattern match (m//) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 399.","Use of uninitialized value $path in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 401."],  } Can't use an undefined value as a HASH reference at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 1038.\n * in WeBWorK::ContentGenerator::Problem::nav called at line 155 of /opt/webwork/webwork2/lib/WeBWorK/Template.pm\n * in WeBWorK::Template::template called at line 610 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::content called at line 372 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm\n * in WeBWorK::ContentGenerator::Problem::content called at line 233 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm\n * in WeBWorK::ContentGenerator::go called at line 386 of /opt/webwork/webwork2/lib/WeBWorK.pm
```

drgrice1 (commented):
I again tested all the various combinations of MIN_HTML_ERRORS and JSON_ERROR_LOG.

I am a bit dubious of the need for JSON_ERROR_LOG option though.  If MIN_HTML_ERRORS=1 then the only difference between JSON_ERROR_LOG=1 and JSON_ERROR_LOG=0 is that the exception and backtrace are part of the json output in the first case, and not in the latter.  Maybe just kill that option and always give json output in the error.log.

I noticed that you are using the perl module JSON::MaybeXS for this.  So JSON::MaybeXS should be added to the check_modules.pl script.  On Ubuntu 20.04 this is provided by the libjson-maybexs-perl package.  However, I noticed that if that package is installed, but the libjson-xs-perl package is not (which contains the JSON::XS perl package), then this fails with an internal server error regardless of the settings for MIN_HTML_ERRORS.  So JSON::XS will probably also need to be added to the check_modules.pl script.

On a related note, the Ubuntu package libjson-xs-perl depends on the package libtypes-serialiser-perl.  For some reason on Ubuntu 20.04 this package seems to be troublesome with WeBWorK.  For example, if AskSage is used (from IO.pm of PG) in a problem, then you get the warnings `Can't locate package Types::Serialiser::BooleanBase for @JSON::PP::Boolean::ISA`.  This package on Ubuntu 18.04 did not cause this issue.  So I am not sure what is going on there.  I haven't tried installing these things from cpan yet to see if the same problem occurs.

drgrice1:
So I did some testing, and see that this works fine directly using the JSON package as is done everywhere else in the webwork2 code base.  So why are you using JSON::MaybeXS here?

taniwallach:
Thanks @drgrice1 

> So I did some testing, and see that this works fine directly using the JSON package as is done everywhere else in the webwork2 code base. So why are you using JSON::MaybeXS here?

Probably as I cribbed the `use` line from somewhere else and the XS versions are supposed to be more efficient. 

For now, I will change it to the plain JSON package.The change will be in the next commit.

We can decide whether to move to a better implementation of JSON support later.

> JSON::PP is a Pure-Perl implementation that can be found in the recent releases of Perl core. According to comments on this article Mojo::JSON, which is also a pure-Perl implementation, is 50% faster than JSON::PP.
>
>According to my measurements Cpanel::JSON::XS is the fastest implementation and despite its name, the only connection to Cpanel is that they sponsor the maintenance of it.
>
>JSON::MaybeXS is a wrapper around a number of these implementations. It will use the XS (and therefore much faster) version if it is installed, and will use the Pure-Perl version if the XS version is not installed.
 from: https://perlmaven.com/json

lib/Apache/WeBWorK.pm
```diff
@@ -212,77 +233,219 @@ sub htmlMessage($$$@) {
...
+		    encode_json($time), ", ",
+		    encode_json("Method"), ": ",
+		    encode_json($r->method), ", ",
+		    encode_json("URI"), ": ",
+		    encode_json($r->uri), ", ",
+		    encode_json("HTTP Headers"), ": ",
+		    encode_json( {%headers} ), ", ",
+		    encode_json("Warnings"), ": ", jsonWarningsList(@warnings), ", ",
+		    encode_json("Exception"), ": ",
+		    encode_json(chomp $exception), ", ",
```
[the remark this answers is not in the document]
taniwallach: Being fixed in the next commit. Thanks!

drgrice1:
As I understand it the JSON package tries to use the JSON::XS package if it is available, and falls back to using JSON::PP if not.  See https://metacpan.org/pod/release/MAKAMAKA/JSON-2.59/lib/JSON.pm#BACKEND-MODULE-DECISION.  Executing `perl -MJSON -e "print JSON->backend"` on my system when libjson-xs-perl is not installed returns `JSON::Backend::PP`, and running that with libjson-xs-perl installed returns `JSON::Backend::XS`.  So that seems to be working as documented.

I have read the Cpanel::JSON::XS is supposed to be the fastest, and JSON::MaybeXS is supposed to choose that backend first if available.  The odd thing on my Ubuntu 20.04 system is that Cpanel::JSON::XS is installed, but yet for some reason I am getting the server not configured correctly error with this pull request when JSON::XS is not installed.

taniwallach:
> I again tested all the various combinations of MIN_HTML_ERRORS and JSON_ERROR_LOG.

Thanks!


> I am a bit dubious of the need for JSON_ERROR_LOG option though. If MIN_HTML_ERRORS=1 then the only difference between JSON_ERROR_LOG=1 and JSON_ERROR_LOG=0 is that the exception and backtrace are part of the json output in the first case, and not in the latter. Maybe just kill that option and always give json output in the error.log.

If anyone is doing any scripted processing of their error log files, such a change might break their scripts. I would prefer not to do that. In the longer term - I think the JSON encoded data will be easier to handle scripts processing the error logs.

> I noticed that you are using the perl module JSON::MaybeXS for this. So JSON::MaybeXS should be added to the check_modules.pl script. On Ubuntu 20.04 this is provided by the libjson-maybexs-perl package. However, I noticed that if that package is installed, but the libjson-xs-perl package is not (which contains the JSON::XS perl package), then this fails with an internal server error regardless of the settings for MIN_HTML_ERRORS. So JSON::XS will probably also need to be added to the check_modules.pl script.

My locally modified `Dockerfile` was installing `libjson-xs-perl` and `libjson-perl` and not `libjson-maybexs-perl` and the code was working. For now, as discussed in the prior reply, I'm changing to a plain `use JSON;` and we can decide what JSON implementation we want in a separate change made consistently to the entire codebase.

> On a related note, the Ubuntu package libjson-xs-perl depends on the package libtypes-serialiser-perl. For some reason on Ubuntu 20.04 this package seems to be troublesome with WeBWorK. For example, if AskSage is used (from IO.pm of PG) in a problem, then you get the warnings `Can't locate package Types::Serialiser::BooleanBase for @JSON::PP::Boolean::ISA`. This package on Ubuntu 18.04 did not cause this issue. So I am not sure what is going on there. I haven't tried installing these things from cpan yet to see if the same problem occurs.

We should definitely look into this separately, and was an extra factor in my decision to just change to plain `use JSON;` for now.

> Edit: The cpan packages cause the same issue.

It could be related to changes in the core of Perl between the version in Ubuntu 18.04 and Ubuntu 20.04. I think we should try to move the Docker setup for WW 2.16 to Ubuntu 20.04 and if the issue is there - we will need to determine how to fix it before the WW 2.16 release.

drgrice1:
I don't think it is the JSON::MaybeXS package that is causing the problem.  Even without this pull request, if the JSON::XS package is installed, the AskSage problems give that warning.  The problem occurs when Sage returns a value of true for "success" in the returned JSON data.  The JSON::XS package uses Types::Serialiser::true and Types::Serialiser::false for this.  The thing is that this works fine from a perl script run from the command line, and even works fine in the webwork2 portion of the code.  However, in the AskSage method in PG/IO.pm it does not.  Is is possible that the safe compartment is interfering with the package dependencies in some way?

drgrice1 (approved):
Everything looks good here for this pull request.

We just need to figure out this JSON::XS issue.

It would be good to switch to JSON::MaybeXS to allow the system administrator to choose which backend to use.  There does seem to be some another issue with that package though in that it does not seem to be properly choosing the backend.  If Cpanel::JSON::XS is installed, that should be used regardless of whether JSON::XS is installed according to the JSON::MaybeXS documentation, but it is not for some reason.

taniwallach:
@drgrice1 I copied your posts about the AskSage and JSON package matters to: https://github.com/openwebwork/webwork2/issues/1205 where we can focus on those matters.

I'll post a few initial thoughts there.

drgrice1:
Okay, so now I understand what is going on with the JSON::MaybeXS package, and why there is an error if it is used when JSON::XS is not installed.

So when JSON::MaybeXS is used and JSON::XS is not installed, then Cpanel::JSON::XS is the backend that is chosen (if available as it is on my system).  If JSON::MaybeXS is used and JSON::XS is installed, then the JSON::XS backend is chosen because it has already been loaded by the other JSON usage instances in webwork (as the documentation at https://metacpan.org/pod/JSON::MaybeXS says).  Of course if neither of Cpanel::JSON::XS or JSON::XS are installed, then it falls back to using JSON::PP.

The problem is a difference in the parameter that the encode_json method of JSON::PP and JSON::XS accept, versus the parameter that the encode_json method of Cpanel::JSON::XS accepts.  That is that the encode_json method of JSON::PP and JSON::XS accept a scalar (strings and numbers), while that of Cpanel::JSON::XS does not (by default).  For Cpanel::JSON::XS the encode_json method will only accept a reference to an array or hash.  This can be seen by executing the following from the command line:
```
perl -MJSON -e "encode_json("foo")"
perl -MJSON::XS -e "encode_json("foo")"
perl -MCpanel::JSON::XS -e "encode_json("foo")"
```
The first two execute with no error, while the last has the error `hash- or arrayref expected (not a simple scalar, use allow_nonref to allow this) at -e line 1.`
Note that the default for `allow_nonref` for JSON::PP and JSON::XS is true, while the default for Cpanel::JSON::XS is false.  That is why this occurs.

So the reason that this is a problem with this pull request is that you are calling encode_json on string scalars.

So this leads me to request that you change this pull request a little further.  Although you don't have to if you don't want to.  I approved it as is, but I think these changes are an improvement and work regardless of the backend chosen.  In any case, the requested changes are in the attached patch.  This switches back to JSON::MaybeXS (you can determine if that is best or not) and it switches to calling encode_json on hash references.  It is much cleaner this way also.
[changes.zip](https://github.com/openwebwork/webwork2/files/6016878/changes.zip)

drgrice1:
There is another change that I recommend also.  Add
```
PerlPassEnv MIN_HTML_ERRORS
PerlPassEnv JSON_ERROR_LOG
```
to conf/webwork.apache2.4-config.dist before the `<Perl>` section (probably about line 43), and add
```
#$ENV{MIN_HTML_ERRORS} = 1;
#$ENV{JSON_ERROR_LOG} = 1;
```
to that file inside the `<Perl>` section with comments about what un-commenting those lines will do.

This is to make these options transparent to those that install webwork2 without using docker.

taniwallach:
Thanks @drgrice1 - patch merged in, other suggestion implemented + some related changes to `Dockerfile`.

taniwallach:
> As I understand it the JSON package tries to use the JSON::XS package if it is available, and falls back to using JSON::PP if not. See https://metacpan.org/pod/release/MAKAMAKA/JSON-2.59/lib/JSON.pm#BACKEND-MODULE-DECISION. Executing `perl -MJSON -e "print JSON->backend"` on my system when libjson-xs-perl is not installed returns `JSON::Backend::PP`, and running that with libjson-xs-perl installed returns `JSON::Backend::XS`. So that seems to be working as documented.
> 
> I have read the Cpanel::JSON::XS is supposed to be the fastest, and JSON::MaybeXS is supposed to choose that backend first if available. The odd thing on my Ubuntu 20.04 system is that Cpanel::JSON::XS is installed, but yet for some reason I am getting the server not configured correctly error with this pull request when JSON::XS is not installed.

With the last commits, including installing all of
  - `libjson-maybexs-perl`
  - `libjson-xs-perl`
  - `libcpanel-json-xs-perl`

Running `perl -MJSON -e "print JSON->backend"` shows
```
JSON::Backend::XS
```
as you reported. However, I don't think approach was intended to be able to determine if it is `JSON::XS` or `Cpanel::JSON::XS` running in the backend.

Running `perl -MCpanel::JSON::XS -e "encode_json("[\"ff\",\"aa\"]")"` works fine.

There is a discussion about renaming  the "Cpanel" fork: https://github.com/rurban/Cpanel-JSON-XS/issues/142

drgrice1:
Thanks for adding those changes.  This pull request looks rather nice now.

taniwallach:
@drgrice1 Thanks for the many reviews and suggestions.

drgrice1:
I think I am going to merge this soon, unless there are objections.

PR #1196, opened 2021-02-12
Achievements and UTF-8

robert-marik:
Without this patch

1. UTF-8 characters in the name or description of achievements produce error "Wide character in print at /usr/lib/x86_64-linux-gnu/perl/5.30/IO/Handle.pm line 418."  The file is created fine, anyway.
2. The UTF-8 characters in the names are broken when exporting score table for achievements
3. The import of achievements with UTF-8 in name or description fails. Only the first non-UTF-8 lines are imported, the rest is silently ignored.

The patch solves these problems.


robert-marik:
Will prepare new patch. This one still produced error.

PR #1197, opened 2021-02-14
Fix to iteration over GWQ problems in case of missing numbers part 2

drgrice1:
PR #1128 fixed gateway quizzes so that problem numbering need not be
consecutive in order for the quiz to function.  However it fails to
match the answers to the correct problem id.  This results in the
answers being lost when the page changes away from any problem after the first
point that problem numbering is not consecutive.  This fixes that issue.

Another issue was found and fixed on the grades page.


drgrice1:
To test this create a gateway quiz with, say, 4 problems, and set the quiz to have 1 problem per page.  Then delete problem 2.  So you end up with a quiz that has 3 problems numbered 1, 3, 4.

Now take the quiz.  This will work as of PR #1128, at least in terms of being able to start the quiz and not get the warning about some problems not being assigned to you.

Try the following without this pull request:  Enter an answer into each of the problems in order.  Then switch back to page/question 2.  The answer you entered is gone.  Switch back to page/question 3.  That answer is gone now to.  The answer in question 1 is still there.  If you grade the quiz some answers will not be submitted (which ones depends on which page you are on when you submit -- at least one of 2 and 3 won't be there).

If you try the above with this pull request the answers do not disappear, and all answers are submitted.

drdrew42 (approved):
Following the given steps, I was able to take a GWquiz with a gap in the problem numbers. Responses were recorded during page navigation, and all responses were submitted accurately.

Alex-Jordan:
Confirmed the issue using Glenn's walkthrough, and confirmed the issue when merging this branch.

Interestingly, before I saw Glenn's walkthrough, I tried my own version. I had a quiz with 7 questions already. I deleted 2, so I had 1,3,4,5,6,7. The quiz was set to two per page, so I saw 1,3;4,5;6,7 but labeled as 1,2;3,4;5,6. They were not in random order. I was able to type answers into answer blanks on page 2, navigate to page 3, and then back to page 2, and what I had typed persisted. That's when I went back to this thread and saw Glenn's walkthrough.

So either I messed up testing that first time, or something about showing 2 problems per page is different than 1 per page.

PR #1200, opened 2021-02-15
Update the mathquill javascript.

drgrice1:
The main change is just pulling in upstream improvements.

One notable fix is that you can now cut and paste into a text block.  Previously this resulted in errors in the console and unexpected behavior in the mathquill input.

I have also included a minor tweak to the text output that is sent to webwork.  That is when an operator name like arcsec is input, the text will no longer include 'operatorname{}'.  This was spurred by a recent forum conversation (see https://webwork.maa.org/moodle/mod/forum/discuss.php?d=5023).

I added another minor tweak that fixes braces entered in text mode.  See the discussion at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=5027.


drdrew42 (approved):

drdrew42:
I still never quite figured out the issues around set-builder, but it seems that mathquill answers (set-builder in textmode and mathmode, arcsec() and the like) are working as expected with this PR.

drgrice1:
With the current MathQuill javascript in develop or before, if you type `arcsec(x)` in an answer the text that is sent to webwork is `operatorname{arcsec}(x)`.  With this pull request it is `arcsec(x)`.

For the set builder issue if you type `"{x|x<3}` the text answer that is sent to webwork is correct, but the latex answer that is input back into the MathQuill after the answer is submitted is `\text{\{x|x<3\}}` which is correct, but MathQuill's text block was unable to parse that, and so the answer in the MathQuill ends up disappearing.  This pull request makes it work.  However, I still discourage using text mode for math answers.  This is the reason we added the vert button to the toolbar.

drdrew42:
I did see the difference with `operatorname` -- that is definitely resolved.
I've double checked with the `\text` version of set-builder (disappears from answer blank before this PR and persists after this PR), and it works.

What I'm seeing now:
- submitted answers' mathjax are not rendering - appears in AttemptsTable as `[math]`
- mathjax in hint/solution knowls are not rendering (appears as raw string with tex delimiters: `\(...\)` so PGML is parsed, but unprepared for mathjax)
- knowl console error on click: [Uncaught TypeError: Cannot read property 'Queue' of undefined] (knowl.js:100)

drgrice1:
@drdrew42:  So there are issues with testing pull requests now.  The actual branch of this pull request does not contain all of the mathjax and jquery changes.  So there are conflicts.  If you merge the develop branch into this pull request branch (which in reverse is what will happen if this pull request is merged), then you won't see those issues.

drgrice1:
To simplify things.  I merged the develop branch into this pull request.  So now those issues won't occur.

drdrew42:
Excellent, I'll add the merging of develop into my default testing workflow. You're right, that's how these should be tested.

Alex-Jordan:
I want to check that what I'm doing for testing makes sense, since it is a little backward from what Glenn outlined.

If I want to test drgrice1/featurebranch, I have been doing:
* check out develop
* pull in case there is an update to develop
* fetch drgrice1/featurebranch
* checkout a new branch that I name "featurebranch" but it is copied from develop
* merge drgrice1/featurebranch into my local featurebranch
* restart apache
* clear browser cache

Is there anything off with that?

drgrice1:
@Alex-Jordan:  That will work too.  That is in a sense more like what actually happens when the pull request is merged in.

Alex-Jordan:
> One notable fix is that you can now cut and paste into a text block. Previously this resulted in errors in the console and unexpected behavior in the mathquill input.

I'm testing this. I have made sure to clear the cache.

I enter an answer blank, press `"` to enter text mode and type some text. Now if I use my mouse to try to select some of that text (say, starting from the middle of the text) the console floods with error messages and nothing highlights. This sounds like the description of the original problem. I can, however, click just outside the text and then select it all. It pastes like `\text{asdfkjasdlkjf}`, which I think is expected.

The arcsec thing checks out.

The braces in text mode persist, so that checks out too.

drgrice1:
> I enter an answer blank, press `"` to enter text mode and type some text. Now if I use my mouse to try to select some of that text (say, starting from the middle of the text) the console floods with error messages and nothing highlights. This sounds like the description of the original problem. I can, however, click just outside the text and then select it all. It pastes like `\text{asdfkjasdlkjf}`, which I think is expected.

Ahh, so you found that bug.  I know of that one as well.  That is not the same bug as the one that I am talking about.  The bug that is fixed allows you to paste into text blocks.  What you found is another bug that I reported (see https://github.com/mathquill/mathquill/issues/923).

Alex-Jordan:
OK I see. I successfully pasted into the text block, so that part works. With two approvals and no outstanding things in the thread, I will merge now.

PR #1203, opened 2021-02-19
Fix the capaLibrary link in the modelCourse

drgrice1:
The capaLibrary link created in #1058 should point directly to the
location in Contrib, and not point to the Contrib link to the actual
Contrib location.  The isSymLink method of FileManager.pm does not fully
resolve link paths, and resolves this link to a path inside the course
directory.  As such it allows users to browse into the CAPA library path
in the File Manager, and edit files there.

This also corrects the $pg{specialPGEnvironmentVars}{CAPA_GraphicsDirectory}
variable value in defaults.config.  That should be the same as in
localOverrides.dist, and both of those should be the actual location of
the CAPA_Graphics directory in Contrib/CAPA.


taniwallach (approved):

Alex-Jordan:
Even if #1213 is merged, this is still a fine thing to do. Especially the bit about the graphics folder. Merging it now.

Issue #1158, opened 2020-11-13
UTF-8 not enabled in Achievements score csv, patch attached.


robert-marik:
UTF-8 is not enabled in Achievements score csv. A short patch is attached.
[webwork_utf8_achievements.patch.zip](https://github.com/openwebwork/webwork2/files/5536912/webwork_utf8_achievements.patch.zip)

taniwallach:
This looks fine. I have never tried out achievements, and apparently no one else actively using the UTF-8 support and a non-English laguage from WW 2.15 has either, so had not run into the issue.

1. Please make a pull request just with the change.
    - I do no think you need `use utf8;` as that is only needed when there are UTF-8 characters in the Perl source file, not for processing UTF-8 data in other places.
2. If possible, post the error message you were getting before the patch in the discussion thread here or that of the PR, so that people who encounter the problem will have a chance to find this patch via a search using Google or GitHub.

robert-marik:
Another issue is that the export of achievements issues error messages if UTF-8 is used in the name or description of the achievement. The attached patch solves both issues.  The pull request is on my todo list, but I am still confused with the github system :(

[AchievementEvaluator.pm_merged_2.patch.txt](https://github.com/openwebwork/webwork2/files/5956450/AchievementEvaluator.pm_merged_2.patch.txt)

robert-marik:
UTF-8 characters in the name or description of the achievement also make troubles. The export works but with error message (the file if fine, though) and the import fails silently. The attached patch solves the problems. It replaces the previous one. Will do a pull request, but I was not able to make the route to submit the pull request to work.

[Achievements-utf8.zip](https://github.com/openwebwork/webwork2/files/5968887/Achievements-utf8.zip)

robert-marik:
The pull request is at https://github.com/openwebwork/webwork2/pull/1206

robert-marik:
Fixed in https://github.com/openwebwork/webwork2/pull/1206

PR #1206, opened 2021-02-20
Allow UTF-8 in achievement export/import/score

robert-marik:
Without this patch

- UTF-8 characters in the name or description of achievements produce error "Wide character in print at /usr/lib/x86_64-linux-gnu/perl/5.30/IO/Handle.pm line 418." The file is created fine, anyway.
- The UTF-8 characters in the names of students are broken when exporting score table for achievements
- The import of achievements with UTF-8 in name or description fails. Only the first non-UTF-8 lines are imported, the rest is silently ignored.

The patch solves these problems. Replaces previously submitted (and closed) pull request https://github.com/openwebwork/webwork2/pull/1196

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
@@ -54,6 +54,8 @@ use WeBWorK::Debug;
 use WeBWorK::Utils qw(timeToSec readFile listFilesRecursive sortAchievements);
 use DateTime;
 use Text::CSV;
+use Encode;
+use open IO => ':encoding(UTF-8)';
 
 #constants for forms and the various handlers
 use constant BLANK_ACHIEVEMENT => "blankachievement.at";


robert-marik:
This is related to the issue reported at https://github.com/openwebwork/webwork2/issues/1158

taniwallach (changes requested):
It looks like all you meant to change was `lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm` but additional commits were added to try to bring your branch in sync with WeBWorK-2.16.

There are now merge conflicts, so it seems that it would be easier to test if you rebased from the current head of WeBWorK-2.16 or just redid the change from there, and force pushed to the new branch.

robert-marik:
Rebased. Thank you.

lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
```diff
@@ -591,7 +593,7 @@ sub score_handler {
 	    my $user_id = $userRecord->user_id;
 	    next unless $db->existsGlobalUserAchievement($user_id);
 	    next if ($userRecord->{status} eq 'D' || $userRecord->{status} eq 'A');
-	    print SCORE "$user_id, $userRecord->{last_name}, $userRecord->{first_name}, $userRecord->{section}, ";
+	    print SCORE Encode::encode("UTF-8","$user_id, $userRecord->{last_name}, $userRecord->{first_name}, $userRecord->{section}, ");
```
taniwallach, line 596: The `Encode` call in this line needs to be dropped. It caused a double encoding which leads to mojibake.

taniwallach (changes requested):
At present there is a bug, and the generated `coursename_achievement_scores.csv` has Mojibake instead of the student names in UTF-8. This is a result of a double encoding as the `print SCORE` line calling `Encode::encode("UTF-8", ... )` when the file is already being opened using `':encoding(UTF-8)'` based on the `use` line at the top of file.

Dropping the extra encoding in the `print SCORE` line fixes the problem for me.

Fixed in https://github.com/robert-marik/webwork2/pull/3 to @robert-marik's branch.


After that change - the student names appear properly in a score file, the achievement name and description appear properly in an "export file" and an import of a file with UTF-8 data works properly.

lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
taniwallach: Fixed in https://github.com/robert-marik/webwork2/pull/3 to @robert-marik's branch.

You can merge, and then rebase so the PR will be nice and clean.

robert-marik:
Fixed, thank you for catching this bug.

taniwallach (approved):
Works as expected with this version of the PR.
Merging.

PR #1207, opened 2021-02-21
Switch from opening a new window to opening a dialog when an image is clicked

drgrice1:
The current method of having a new browser window or tab open when an image in a PG problem is clicked is intrusive.  Furthermore the behavior that occurs is different depending on which browser is in use.

This changes to opening a bootstrap modal dialog  when an image is clicked instead.  That dialog can be moved around and the image in the dialog can be resized using the zoom in and zoom out buttons in the dialog.  The image can also be resized using the +/- keys or the mouse wheel.

This is the webwork2 part that contains the javascript and css, and loads this in the appropriate placed (Problem.pm, GatewayQuiz.pm, etc).

For this to work you also need to checkout https://github.com/openwebwork/pg/pull/529.


Alex-Jordan:
I tried this and it's not working. I am viewing a problem form the OPL. I clicked the "Try it" eyeball icon to open it in an undefined set. When I click on an image in this problem, it opens in a new (small) window, not a dialog. The same happens if I click the image while still in the Library Browser.

drgrice1:
Did you also checkout the pg side of this?

Alex-Jordan:
Nope :) Will do that later today, sorry.

Alex-Jordan (approved):
This works as advertised.

taniwallach:
@drgrice1 There are now some minor merge conflicts with develop. Could you clean the up?

taniwallach (changes requested):
There are the small merge conflicts to be fixed.
Testing with this branch and the paired https://github.com/openwebwork/pg/pull/529 - it works as advertised.
The accessibility side seems well handled:
  - The zoom feature has keyboard control.
  - Escape can be used to leave the modal dialog.
  - Tab moves between "items" and enter will "press" the buttons when selected.
  - The modal has `role="dialog"` set.
  - 
I'm not a real expert, but from my recollections of WCAG 2.0 requirements and looking at
  - https://developer.yoast.com/blog/the-a11y-monthly-making-modals-accessible/
  - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

I think this should be OK. If someone has access to a screen-reader, a test with that might be a good thing.

taniwallach (approved):
Tested and ready to merge IMHO.

taniwallach:
@pstaabp would you give a quick look at the javascript code before we merge this. (Pay attention that you need the paired PR for pg.) Functionality seems fine already.

drdrew42:
https://user-images.githubusercontent.com/3912882/110045160-48dc0d00-7d18-11eb-9e6c-e9dcb20e3721.mov

I'm not seeing the default image dimensions being respected when initially clicked, and it seems that zoom only affects the width?

drgrice1:
@drdrew42:  Is there anything in the console?  I will try that problem, and see if I can see the same thing.  It is in Contrib, right?

drdrew42:
Yes, just prepend 'Contrib/' to the path shown in the video.

In the console:
`[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952`

drgrice1:
@drdrew42:  Okay, so this is an odd one.  I tested this in Firefox with that problem and did not see the problem you see.  I didn't see anything in the developer console either.  Then I tested this in Google Chrome.  I also did not see the problem.  The image opened at its natural dimensions, and zoom worked as expected.  Then I opened the developer console in Google Chrome.  Then it started to happen.  Then I closed the developer console, and it stopped.  I will see if I can fix that, but it only happens with Google Chrome with the developer console open.  I also see the warning you are talking about.

drdrew42:
confirmed, I have similar behavior -- it didn't occur to me that having the developer console open would make a difference here...

drgrice1:
Yeah, I have started to notice other issues with things when the developer console is open recently.  In both Google Chrome and Firefox, if the developer tools are open and and you have MathJax displayed on the page, and something calls the typeset promise method to typeset additional MathJax loaded later (like in a solution or hint knowl), then the already typeset stuff flickers.  This is with MathJax v3.  This doesn't happen if the developer tools are closed.

drdrew42:
Screen Shot 2021-03-04 at 8 06 13 PM

In both Safari and Chrome, I'm getting `maximum call stack size exceeded` with this update. (loading the same pg problem as earlier)

Alex-Jordan:
I think this and the pg counterpart now have 3 approvals. Merging.

PR #1213, opened 2021-02-24
Prevent following any symbolic link in File Manager

Alex-Jordan:
This is to address the concern discussed at https://github.com/openwebwork/webwork2/commit/dcaa14400e71e36991eb90ae18b55c7063049f4e#commitcomment-47292260


drgrice1 (approved):
This looks good.

drgrice1:
This is a rather straightforward change.  It would be nice to have another review.

Basically, after this a symbolic link will not be allowed to be followed from the file browser unless it is explicitly listed in `$webworkDirs{valid_symlinks} = [];` which can be set in `localOverrides.conf`.

If there are any objections to this new behavior, then please say so now.  Otherwise I will plan to merge this in a day or two.

PR #1214, opened 2021-02-24
Update the translation pot and po files to pull in strings that have been added in the code.

drgrice1:
This is simply the result of running the script bin/update_localization_files.

Note that the excessively long string in Localize.pm was marked as not translatable so that xgettext.pl will not add it to the po and pot files, but a comment in the file notes this.

This should be merged soon so that translators have time to update po files before the release.

This will probably need to be done again once the major pull requests are in, as they will most likely have additional translatable strings.


Alex-Jordan:
> as they will most likely have additional translatable strings

At a minimum, #1213 adds a new string.

taniwallach:
> This should be merged soon so that translators have time to update po files before the release.

I agree that we should clean this up as well as we can and merge soon.

> This will probably need to be done again once the major pull requests are in, as they will most likely have additional translatable strings.

Hopefully some of the current translators can be asked to review the updated translation files in their language once the release candidate is ready. Changes to the translation files by a trusted translator to a specific file should be merged if the visual changes do not noticeably break the structure of the translation file.

drgrice1:
I had fixed the `fuzzy` strings in the en.po and en_us.po files in the original pull request.  I forgot to do this when I ran the script again to pull in new strings from later pull requests.

There is no need to mark the `fuzzy` strings in the other languages.  That is what `fuzzy` is for.  It is the mark.  It is rather silly to mark a mark.  The way that gettext works is that when new strings are added it tries to fill in a string from an already translated close match, and then it marks them as `fuzzy`.  Of course it is not perfect.  That is what the `fuzzy` mark means.

The en.po and en_us.po files do not need a translation unless the msgstr is different from the msgid.  This only is needed for the incorrectly dealt with translations that you mentioned (_ACHIEVEMENTS_EDITOR_DESCRIPTION and the like).  Those strings should be fixed to be handled correctly in the code.  A method should be written that as called to retrieve and translate the strings from the lexicon instead of this hackish way of dealing with it.  For now what should be done is for all of those translations in the en.po and en_us.po file be removed except for these `fake` translations.

taniwallach:
> I had fixed the `fuzzy` strings in the en.po and en_us.po files in the original pull request. I forgot to do this when I ran the script again to pull in new strings from later pull requests.

It is too easy to forget it... 

> There is no need to mark the `fuzzy` strings in the other languages. That is what `fuzzy` is for. It is the mark. It is rather silly to mark a mark. 

The mark in the comment helps "us" = the developers.

However, end users will see the "guessed translations" which in quite a few cases seem to be incorrect. A visible mark can help the end users report the issue to their site admins, and maybe will lead to fixed translations being submitted.

> The en.po and en_us.po files do not need a translation unless the msgstr is different from the msgid. This only is needed for the incorrectly dealt with translations that you mentioned (_ACHIEVEMENTS_EDITOR_DESCRIPTION and the like). Those strings should be fixed to be handled correctly in the code. A method should be written that as called to retrieve and translate the strings from the lexicon instead of this hackish way of dealing with it. 

I agree that a better mechanism is needed for those special strings. @mgage essentially said the same in some prior discussions of that hack.

> For now what should be done is for all of those translations in the en.po and en_us.po file be removed except for these `fake` translations.

Not urgent - so long as it works as is. 

Eventually, we would like to be able to get Transifex updated, and things set up so we can sync changes to/from Transifex in a dependable manner. Hopefully that would allow updating `webwork2.pot` from updated code, uploading that to Transifex, and having Transifex prepare the revised `po` files. For now - Transifex is almost certainly out-dates in many ways.

drgrice1:
I stripped all translations from the en.po and en_us.po files.  Then there was an error when viewing the past answers page.  The reason for that is that the string was missing from the lexicon.  Once it was added it worked.  Perhaps that is why @mgage  thought those strings needed to be translated?

drgrice1:
It should be that we have a translator for each `po` file that is in the repository.  In the proper release cycle we should have a string freeze.  Then the translators go to work.  The `fuzzy` strings are dealt with by them.  It is incorrect to mark them in a way that is visible to the end user as you say.  It is exceedingly easy to find the word `fuzzy` in the po file.

taniwallach:
> @mgage and whoever wanted "short" placeholders, and since those placeholders need to be replaced in English, they need values in `en.po` and `en_us.po`. It was a hack that would be better solved otherwise, as it makes more work maintaining the translation files.
> 
> I suggest dropping the English translations, but where these are used - if the `msgstr` is the same as the original `msgid` forcing an override from a fixed dictionary.

From @drgrice1 's post https://github.com/openwebwork/webwork2/pull/1214#issuecomment-794579535  it seems that the proper settings in the "Lexicon" fix this already.

I still need to test this...

drgrice1:
> > Okay, so why is it that the strings like _ACHIEVEMENTS_EDITOR_DESCRIPTION and the like need anything in the en.po and en_us.po files?
> 
> @mgage and whoever wanted "short" placeholders, and since those placeholders need to be replaced in English, they need values in `en.po` and `en_us.po`. It was a hack that would be better solved otherwise, as it makes more work maintaining the translation files.
> 
> I suggest dropping the English translations, but where these are used - if the `msgstr` is the same as the original `msgid` forcing an override from a fixed dictionary.

Sometimes I don't think you read what I am saying correctly, or maybe I was not detailed enough.  There is already a `fixed dictionary` that does exactly this.  It is the `%Lexicon` in Localize.pm.  None of the English translations are needed already.  I tested this and said so.  This is what the Locale:Maketext:Lexicon perl package does.

> > The `fuzzy` strings are dealt with by them. It is incorrect to mark them in a way that is visible to the end user as you say. It is exceedingly easy to find the word `fuzzy` in the po file.
> 
> Unless we recruit people to fill the role of "language file maintainer for X" we will end up with students seeing incorrect translations. Let's see what we can do about asking Marianna to look for "language file maintainers" once the release candidate branch is ready.

I agree that we have a shortage of translators, and need to recruit.  Doing what you say is still not the correct way to deal with this.  The students will still see the incorrect translations just with question marks.  If we are going to make a change, the only viable option is to completely remove the translation and the `fuzzy` marks.  Then the students will see the correct string, but of course still in English.

taniwallach:
> If we are going to make a change, the only viable option is to completely remove the translation and the `fuzzy` marks. Then the students will see the correct string, but of course still in English.

Sold - let's make this a policy - all non-English `fuzzy` translations get an empy `msgstr` (unless we are certain that the `msgid` was changed in a trivial manner, ex. fixing typos or punctuation). We can leave the "fuzzy" guess as a comment, which may be of some help to the people fixing the dictionaries.

Maybe we do this in a scripted manner as an automatic post-processing step at the end of `bin/update_localization_files`. (This is not a request to try to code that now.)

drgrice1:
It would not be hard to write a script to remove the `fuzzy` marks and change corresponding `msgstr`'s to the empty string.

It would be hard to make it so only those that are "nontrivial" are dealt with though.  Really, the only way to do that is to go through them one at a time.  There you are right back to the job of the translator.

Also, how are you editing po files?  The way that a translator should be doing this is via a po file editor.  There are many of these available online for free, as well as applications that can be installed (like gtranslator and lokalize).  By modifying the po files and removing the `fuzzy` mark, you are messing up things for editing in a po file editor.

drgrice1:
I was looking in the wrong place for the option to disable fuzzy matching.  I was trying to find the option for `xgettext.pl`, but of course that isn't what creates the po files.  `msgmerge` does that.  I added that option in the last commit.

drgrice1:
This should be merged now so that translators can start working on translating.

drgrice1:
It definitely should be part of the release candidate.

Alex-Jordan:
One or two PRs I just opened touch on `maketext` instances and would impact this. If they go into the RC, then this should be updated again. I'm ready to merge this, but wanted to point that out. If you say the word (one more time, sorry), I will merge it.

drgrice1:
Yeah, we can wait a bit for those.

taniwallach:
I don't have time to do a proper review and test of this in the next few days, but the approach is sound, and undefined translations stings (`msgid` not defined in the PO file) can totally break server functionality and is far worse than places where there is an empty translation and a fallback to the source language.

I hope to be able to install the 2.16 branches on my development server relatively soon, and then to try to update the Hebrew translation file. While doing that, I may discover minor issues which can be addressed later.

I would recommend moving ahead and merging this PR.

drgrice1:
This has been updated for all strings in the code, and there are no pending pull requests that will change any strings.  So it is time for this to go in now.

Alex-Jordan:
OK, I'll push the button.

drgrice1:
@Alex-Jordan:  Note that I had to change your `\xA0` into a `&nbsp;`.  The gettext.pl script complained about that being an invalid multibyte character.

Alex-Jordan:
Thanks, good to file away for maketext-able strings. Probably better for the translator to actually see the `&nbsp;` to consider if they would use that in their language.

I had a thought that maybe the translation should only be on "version" and `[_1]` should be the number _and_ the nbsp. But that would assume all languages go like "version 4" and not like "4 version"...

drgrice1:
Yeah, and that is the right way to go.  This is one of those cases where translators may ask what that is and how to deal with it.

PR #1215, opened 2021-02-24
Update mathjax to version 3

drgrice1:
Note that to test this you will need to checkout the master branch of MathJax, and change the appropriate line to
`$webworkURLs{MathJax} = "$webworkURLs{htdocs}/mathjax/es5/tex-chtml.js";` in localOverrides.conf.

This is paired with https://github.com/openwebwork/pg/pull/532


conf/defaults.config
```diff
@@ -329,7 +329,7 @@ $webworkURLs{AuthorHelpURL}           ='http://webwork.maa.org/wiki/Category:Aut
 # <link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="htdocs"-->/css/math.css"/>
 
 # Location of MathJax script, used for the MathJax display mode.
-$webworkURLs{MathJax}       = "$webworkURLs{htdocs}/mathjax/MathJax.js?config=TeX-MML-AM_HTMLorMML-full";
+$webworkURLs{MathJax}       = "$webworkURLs{htdocs}/mathjax/es5/tex-mml-chtml.js";
```
dpvc, line 332: I would just load `tex-chtml.js` rather than `tex-mml-chtml.js` unless you really want MathML support.  I'm assuming the reason that `TeX-MML-AM` was uses was to get the AsciiMath, and there were not combined files that were just TeX and AsciiMath.  It would be possible to build a custom combined component (the tools are part of the mathjax-full npm package) that includes just TeX and AsciiMath, together with CommonHTML output.  It's not hard to do that, but would be additional steps in the Dockerfile.

htdocs/js/apps/ProblemGrader/problemgrader.js
```diff
@@ -1,19 +1,18 @@
...
-    });
-
-}) 
+	$(".preview").click(function(evt) {
+		var previewBtn = $(evt.target);
+		previewBtn.attr("data-content",
+			"<span>" + previewBtn.siblings("textarea").val().replace(/</g, '< ').replace(/>/g, ' >')) + "</span>";
+		previewBtn.popover('toggle');
+		if (window.MathJax) {
+			MathJax.typeset([ '.popover-content' ]);
```
dpvc, line 15: If you don't load `tex-chtml-full.js`, then you can run into issues here if the math needs to load any extensions (e.g., uses `\color` or `\verb` or a bunch of other things).

Otherwise, you might want to use `MathJax.typesetPromise(['.popover-content'])` instead, or better yet

``` javascript
MathJax.startup.promise = MathJax.startup.promise.then(() => MathJax.typesetPromise(['.popover-content']));
```
so that you don't accidentally end up calling `MathJax.typesetPromise()` twice concurrently.

htdocs/js/apps/ProblemSetDetail2/problemsetdetail2.js
```diff
@@ -1,356 +1,356 @@
...
+				response = $('<div/>',{style:'font-weight:bold','class':'ResultsWithError'}).text('No Such File or Directory!');
+			}
+			if (/"server_response":"","result_data":""/i.test(response)) {
+				response = $('<div/>',{style:'font-weight:bold','class':'ResultsWithError'}).text('There was an error rendering this problem!');
+			}
+
+			$('#psr_render_area_'+id).html(response);
+			// run typesetter depending on the displaymode
+			if(displayMode=='MathJax')
+				MathJax.typeset();
```
dpvc, line 337: Again, this may want to be

``` javascript
MathJax.startup.promise = MathJax.startup.promise.then(() => MathJax.typesetPromise());
```

htdocs/js/apps/SetMaker/setmaker.js
```diff
@@ -418,7 +418,7 @@ function randomize(filepath, el) {
 	      $('#'+el).html(data);
 	      // run typesetter depending on the displaymode
 	      if(displayMode=='MathJax')
-		  MathJax.Hub.Queue(["Typeset",MathJax.Hub,el]);
+              MathJax.typeset([ el ]);
```
dpvc, line 421: Similarly

``` javascript
MathJax.startup.promise = MathJax.startup.promise.then(() => MathJax.typesetPromise([el]));
```

htdocs/js/vendor/other/knowl.js
```diff
@@ -97,8 +97,7 @@ function knowl_click_handler($el) {
       if(window.MathJax == undefined) {
             $knowl.slideDown("slow");
       }  else {
-            MathJax.Hub.Queue(['Typeset', MathJax.Hub, $output.get(0)]);
-            MathJax.Hub.Queue([ function() { $knowl.slideDown("slow"); }]);
+            MathJax.typesetPromise([ $output[0] ]).then(function() { $knowl.slideDown("slow"); });
```
dpvc, line 100: In keeping with using `MathJax.startup.promise` as a quing mechanism, this might want to be

``` javascript
MathJax.startup.promise = MathJax.startup.promise.then(() => MathJax.typesetPromise([$output[0]]));
MathJax.startup.promise.then(function () {$knowl.slideDown("slow")});
```

(Here, the next MathJax doesn't have to wait for the knowl, so you don't need to save the promise for that.)

htdocs/js/vendor/other/knowl.js
```diff
@@ -121,8 +120,7 @@ function knowl_click_handler($el) {
...
```
dpvc, line 123: Same here as previous comment.

lib/WeBWorK/Utils/AttemptsTable.pm
```diff
@@ -360,7 +360,7 @@ sub previewAnswer {
 	} elsif ($displayMode eq "images") {
 		$imgGen->add($tex);
 	} elsif ($displayMode eq "MathJax") {
-		return '<span class="MathJax_Preview">[math]</span><script type="math/tex; mode=display">'.$tex.'</script>';
+		return '<span>\['.$tex.'\]</span>';
```
dpvc, line 363: Yo don't really need the `<span>` element, here and below, but I guess it doesn't hurt.  (A `<div>` might be better, since the content is display style, so block level).

htdocs/themes/math4/gateway.template
```diff
@@ -42,6 +42,7 @@
 <link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="theme"-->/math4-overrides.css"/>
 <!--#endif-->
 <script type="text/javascript" src="<!--#url type="webwork" name="htdocs"-->/js/vendor/jquery/jquery.js"></script>
+<script type="text/javascript" src="<!--#url type="webwork" name="MathJax"-->" id="MathJax-script" async defer></script>
```
dpvc, line 45: If you want to support IE11 (the oldest IE that v3 supports), you will need to include 

``` html
<script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
```

as well.  It could be inside an IE if-then comment:

``` html
<!--[if IE 11]>
<script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
<![endif]-->
```

This will need to be included in all the templates and the WeBWorK clients that load MathJax.

I don't think you want both `async` and `defer`.  Both cause the script to be loaded asynchronously, but `async` will have it run as soon as it is loaded, while `defer` puts it off until the page is fully parsed.  I would say `defer` is probably a reasonable choice.  If you put the polyfill script before the MathJax one, you would also be able to use `defer`, since `defer` preserves the order of execution (while `async` doesn't).

dpvc (changes requested):

drgrice1:
Thanks for the comments.  I will look into the needed changes.

lib/WeBWorK/Utils/AttemptsTable.pm
drgrice1: I'll just remove the span then.  Or would it be better to have it wrapped in a div?

dpvc: I don't think it needs either one.

htdocs/themes/math4/gateway.template
drgrice1: I didn't know about this `<!--[if IE 11]>...<![endif]>` construct.  Is that an html thing, or is that a webwork template implementation?

drgrice1: Never mind.  Google searching is a bit tedious for this sort of thing, but I found it.

drgrice1:
I have noticed with the content that is typeset after the page loads (knowls, grader comments, and such) that the rest of the page content flickers.  It seems as if mathjax is re-typesetting the entire page.  This is with (or without) the changes you suggested to the typesetPromise code.  Is it set up right to only typeset the new elements?

htdocs/themes/math4/gateway.template
dpvc: It's an HTML thing (actually, an IE thing).  To all other browsers, it looks like a comment (between `<!-- and -->`), but IE interprets it and only does the HTML in between if the condition is met.

When I was looking up the reference for it, it turns out that it is no longer supported in IE 11 (who knew?), so don't bother with it.  Here's the [Wikipedia page](https://en.wikipedia.org/wiki/Conditional_comment) for it.  The Microsoft pages have all been taken down.

drgrice1: Okay, I will just add the polyfill unconditionally.  Presumably polyfill.io is smart enough to load only what is needed.

pstaabp:
I just pull this down (even the new couple that you just posted).  I'm getting the following javascript errors:
```
tex-chtml.js:1 Uncaught RangeError: Maximum call stack size exceeded
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
    at o (VM594 tex-chtml.js:1)
```

this is on a problem with about 10 [math] boxes on the page.  I just updated MathJax, so I'm running 3.1.2 of MathJax.

Wondering if I have some configuration problem.

drgrice1:
I am not sure.  I am not seeing that.

I don't think this would give the issues that you are seeing, but did you change the line in localOverrides.conf to `$webworkURLs{MathJax} = "$webworkURLs{htdocs}/mathjax/es5/tex-chtml.js";`

drgrice1:
@pstaabp:  Also, just to make sure... You have checked out both this pull request and the corresponding pg pull request, right?

drgrice1:
@dpvc: Does MathJax 3 no longer allow selections?  I can't seem to copy and paste MathJax stuff anymore.  This isn't a problem for me as I didn't do this before, but I am just curious.

pstaabp:
I had missed the pg branch.  Those errors went away.

pstaabp (approved):

dpvc:
> I have noticed with the content that is typeset after the page loads (knowls, grader comments, and such) that the rest of the page content flickers. It seems as if mathjax is re-typesetting the entire page. This is with (or without) the changes you suggested to the typesetPromise code. Is it set up right to only typeset the new elements?

Yes, but typesetting new content can cause MathJax to change the stylesheet that it uses.  It may be that that is causing the flicker you are seeing.  What browser are you using.  I haven't tried this branch yet, but will see if I get the same result.  If that is the cause, then switching to the sag output might be a way to avoid it (as the SVG output doesn't need to update the CSS as often).

> Does MathJax 3 no longer allow selections? I can't seem to copy and paste MathJax stuff anymore.

You are correct, you can't select the math in v3.  This is due to the way the CommonHTML output uses CSS `content` styles to insert the characters, and it turns out that `content` can't be selected.  This will probably change in the next major release this summer, but for now, that is true.  I did work out a [couple of alternatives](https://groups.google.com/g/mathjax-users/c/1BzTQFINbqY/m/xGYAlhDNAwAJ) that would allow copying and pasting.  One of these could he added to the `mathjax-config.js` file if you wanted to add that.  (I haven't looked carefully to see if they would need any updating since those were written, but you could give it a try if you wanted.)

I will try out the branch, but your changes look good.

dpvc (approved):
I've tested out the branch, and it works for me.  I don't see the flickering you indicate.  There is a bug in Safari that causes it to sometimes clip the expressions incorrectly, however, and perhaps that is part of what you are seeing, if you are using Safari.  For that reason, you may want to use the SVG output rather than CHTML.  We are hoping to address that issue during the font update this summer.

drgrice1:
Not being able to select the MathJax is fine with me.  If others want that, then we can see about implementing that.

I am using the Firefox browser on linux.  However, I just noticed something with the flickering.  It only occurs if the developer tools are open when the page loads.  If the developer tools are not open when the page loads (even if they are opened after the page loads) this flickering does not occur.  I tested this with Google Chrome as well, and the same behavior is exhibited there.

I almost always have the developer tools open, but students (and most other people) of course will not.  So this is not an issue.

drgrice1:
I guess this is an issue if there is a problem on Safari.  I don't have Safari to test, but we can make the switch if that is needed.

htdocs/js/apps/ProblemGrader/problemgrader.js
```diff
@@ -1,19 +1,18 @@
...
-    });
-
-}) 
+	$(".preview").click(function(evt) {
+		var previewBtn = $(evt.target);
+		previewBtn.attr("data-content",
+			"<span>" + previewBtn.siblings("textarea").val().replace(/</g, '< ').replace(/>/g, ' >')) + "</span>";
+		previewBtn.popover('toggle');
+		if (window.MathJax) {
+			MathJax.startup.promise = MathJax.startup.promise.then(() => MathJax.typesetPromise(['.popover-content']));
```
dpvc, line 15: I just remembered:  the `=>` construct will not work in IE11 (I'm too used to using Typescript and down-compiling to es5).  To support IE11, replace

``` javascript
() => MathJax.typesetPromise([...])
```

with

``` javascript
function () {return MathJax.typsetPromise([...])}
```

here and the other places it is used.

drgrice1: Yeah, stupid on my part. I should have seen that. I will fix it.

Alex-Jordan:
I tested this and it works for me. I also tried Safari and I couldn't see any clipping. But it sounds like it only happens in certain circumstances. With 2 approvals, plus Glenn's I assume, I will merge this now. (And the pg counterpart.)

htdocs/js/apps/MathJaxConfig/mathjax-config.js
```diff
@@ -0,0 +1,20 @@
+if (!window.MathJax) {
+	window.MathJax = {
+		tex: {
+			autoload: { color: [], colorV2: ['color'] },
```
Alex-Jordan, line 4: This configuration essentially says that if `\color` is used in some math, load the `colorV2` extension, not the `color` extension. And so `\color` will be the version from `colorV2`.

If you don't know, the difference is that the `colorV2` version is targeted. So you use it like `\color{blue}{this is blue}`. If you use it like `\color{blue}this is blue` then only the "t" will be blue. With the `color` extension, it works like in LaTeX affecting the color of all that follows within some scope. So with the `color` version, `\color{blue}this is blue` will make all of "this is blue" be blue.

When I put `This is \(\color{blue}blue\)` or `This is \(\color{blue}{blue}\)` into a problem, it doesn't work. MathJax does not recognize `\color` as a macro. 

When I return the configuration to use `color` tor the `\color` macro instead of `colorV2`, then it works.

I have two questions. Am I right that this appears to be a MathJax bug?

And are we sure that we want WeBWorK's default to use `colorV2`? That keeps the behavior from before this version of WW, when MathJax 2 is supposed to be used. But it also makes the `\color` macro behave differently in HTML and hardcopy.

drgrice1: When I chose these options, I chose them to conform to the set up of webwork before the transition to MathJax v3.  The colorV2 extension does this.  As to if this is what we want or not, I don't know.

Previously there was no choice, but now that we have the choice, I agree that it would be better to choose consistency between HTML and hardcopy.

drgrice1: I am seeing the issue you see as well.  If I add `\(\color{blue} ...\)` I don't get color.  I noticed you can do `\(\require{color}\color ...\)` or `\(\require{colorv2}\color ...\)` and it takes effect from there on in the page, but that isn't really what we want.

@dpvc:  Is the configuration not correct to make `\color` work without having to call `\require` first, or is this a MathJax bug as @Alex-Jordan suggests?

dpvc: It should be `colorv2` not `colorV2`.  We normalized all the names to lower case in the last release.  I did make compatibility translations in most places, but apparently missed the autoload options.

dpvc: PS, the LaTeX-color extension was available in v2 as well, but was not the default.  Now it is in v3.

drgrice1: I see. I misread or misunderstood the documentation.

dpvc: You may have been looking at an older version of the documentation from before we made the change.

Alex-Jordan: I don't object to keeping it default to `colorv2` instead of `color`.

Is it important for this configuration file to live where it lives instead of in `webwork2/conf/`?

drgrice1: > Is it important for this configuration file to live where it lives instead of in `webwork2/conf/`?

I think we should just use the MathJax v3 default.  I think that consistent behavior between HTML and hardcopy output is better.

We could make a link from its current location to the conf dir, but it is javascript, and can't directly be used in the conf dir.

drgrice1: Another option is to move this code into the math4.js file, and then you can use a custom theme and override that file.  Or even just use the math4-overrides.js file to do this.  You lose the `defer` attribute, but that would still work as by the time the MathJax javascript loads it would be defined.

drgrice1: Actually, since math4.js is `defer`ed we may need to `defer` the math4-overrides.js loading as well for that to work right.

drgrice1: Although, then the above suggestion for the MathJax config won't work.

drgrice1:
There is yet another option.  If we stick with `colorv2` then you can do `{\color{blue}{5x+3}}`, and this will work the same both in HTML and in hardcopy.

PR #1216, opened 2021-02-26
Update jsxgraph to the latest version.

drgrice1:
There was a bug in version 1.1.0 that could cause curve's not to appear due to being unable to generate points.

At this point this only affects the recently added parserGraphTool.pl macro.


drgrice1:
To test this use the attached problem, and in that problem graph the curve through the points (0, -3), (2, -11), and (6, -12).  Without this javascript update, you will see the points graphed, but the curve will not be visible.  With the javascript in this pull request you will see the curve.
[CustomThreePointParabolaTool.pg.txt](https://github.com/openwebwork/webwork2/files/6062317/CustomThreePointParabolaTool.pg.txt)

Alex-Jordan:
Works for me. (I almost thought it was not working, but I realized the browser had cached jsxgraph.) Since this is just updating an external dependency, and the test works, I will merge.

PR #1218, opened 2021-03-01
Remove the onmouseover event from the attempts table answer previews

drgrice1:
Remove the onmouseover event from the attempts table answer previews,
and instead of hacking that with javascript to turn it into a bootstrap
popover, just directly use a bootstrap popover.  The onmouseover event
frequently does trigger when the page loads and shows the wztooltip if the
mouse happens to be in the right position when the page loads.


Alex-Jordan:
Before merging this, I was getting the popover and I saw the onmouseover event in the inspector.

After this, the event is not in the `td` anymore. And I do see the answer-preview span. But clicking it does not cause anything to happen.

In the examples I tried, the answer preview is rendered by MathJax. Maybe MathJax 3 is interacting differently with this span than 2 was? In any case, we should think about how the MathJax settings might be such that clicking triggers the zoom, and if that would be fighting with this.

drgrice1:
@Alex-Jordan: You need to clear your browser's cache.

drgrice1:
You may also have needed to merge origin/develop into the test branch.  I pushed that merge here to ease testing for you.

One thing I recommend that you do is disable network caching in the tab or window you are viewing the page in any time you are testing a pull request that changes javascript or css.  This can be done in the network tab of the browser developer tools.

drdrew42:
https://user-images.githubusercontent.com/3912882/110043614-b63a6e80-7d15-11eb-9385-c340379d9dba.mov

I pulled this, disabled cache, hard-refresh, etc. This was the result every time, tooltips appear onClick, rather than onMouseOver, and only disappear onClick as well. 

Chrome Version 88.0.4324.192 (Official Build) (x86_64)

drgrice1:
@drdrew42:  Isn't that what they are supposed to do?  They have never appeared on mouseover before.

drdrew42:
I must have misunderstood - I think I latched onto the 'onmouseover' part of the description. It clearly says 'remove', d'oh!

The video is proof then that this works as advertised.

drdrew42 (approved):

drgrice1:
If we want to switch to an actual mouseover tooltip we could do that.

Previously the data for a popover was stored in the inline `onmouseover` attribute, and then javascript hacked into that (in math4.js) and turned it into a popover.

This pull request just removes that hackery, and makes it directly a popover (still activated in math4.js, but in the usual bootstrap way).

drgrice1:
The point is that occasionally, if the mouse is in the correct position when the page loads, the old `onmouseover` handler was actually called.  That resulted in an ugly wztooltip floating mysteriously in the page.

Alex-Jordan:
Sorry for all my missteps recently, many of which relate to caching. Testing now, it is working for me.

Nits: 

(1) If I change MathJax settings, I might be someone who sets the zoom trigger to be "click". Testing indicates this overrules the click for the popover. (The person clicks and they get the zoom, and do not get the popover.) Should we try to make sure click-to-zoom is always an option? 

(2) You put the popover on a span that surrounds the answer. This span surrounds MathJax content with a 1em margin. An effect is that the table cell is taller than it would otherwise be. The clickable area for the popover is still just the span, which occupies roughly the middle third of the cell. Could we make it so the clickable area for the popover is any place within the cell? Instead of using this new span, could the `td` get the popover data and the entire td is the clickable area? This would sort of mitigate (1), because then clicking the central region would trigger zoom, and clicking the margin would trigger the popover.

A mouseover trigger for the popup might address some of what I mentioned, but I think mouseover should be avoided. Is it even meaningful for a user on a tablet or phone?

drgrice1:
@Alex-Jordan:  Note that the MathJax click to zoom issue also occurs without this pull request.  This doesn't change that.  Are you asking me to fix that?

Also the 1em space above and below is there without this pull request, and is added by MathJax, not the new span element.

drgrice1:
By the way, I initially tried to put the popover directly on the `td`, but that causes problems with the table when the popover is added.

drgrice1:
The clickable area is also the same both with and without this pull request.

Alex-Jordan:
OK I thought that since the onmouseover event was in the `td`, that previously the popover click trigger area was the entire table cell area. And therefore (I thought) this was reducing the clickable area. But I see that's not how it currently works.

Alex-Jordan:
If I put `display:inline-block;width:100%;` on that span, it seems to make the entire table cell have the clickable for the popover. I am unsure if that has unwanted consequences though.

drgrice1:
You beat me too it.  I was going to try that.

drgrice1:
Setting the style in that way also makes it so that if the MathJax click to zoom option is set, then you can still click outside of the MathJax area and get the popover.

However, the popover does open a bit lower relative to the answer.  This is probably not a big deal though, and maybe is even desirable.

Alex-Jordan:
If you think it is an improvement, you could add it to this PR. Or I will merge this one and add that separately for review. (I would put that styling in `math4.css` (and `math3.css`?) for class "answer-preview", which I think is a new class you made for this span [and therefore there would be no unintended consequences of messing with its styling].)

drgrice1:
I will add it shortly.  I will probably do a bit better with the css selector.

drgrice1:
Hmm, so the math3 theme needs some work.  Can we drop support for that?  Does anyone really use that anymore?

drgrice1:
Okay.  That math3 theme is fixed.  Both for this pull request, and for the MathJax merge (but out of place in this pull request).

Alex-Jordan (approved):
Approved, but I did not test with math3. I'm in favor of discontinuing support for math3. I think math4 has been around since somewhere around 2012–2013. We should ask @mgage if he thinks there is reason to continue supporting math3.

pstaabp:
A little silly asking this, but the popover is showing the entered text, which in the left column.  The functionality is there with the popover, I'm just curious about the usefulness of this.  I understand this fixes the onmouseover problem.  Do people use this feature?

pstaabp (approved):
Seems to be working as advertised.

drgrice1:
> A little silly asking this, but the popover is showing the entered text, which in the left column. The functionality is there with the popover, I'm just curious about the usefulness of this. I understand this fixes the onmouseover problem. Do people use this feature?

Note that the entered column is not always there. There is a course configuration option to have that not shown.

pstaabp:
> > A little silly asking this, but the popover is showing the entered text, which in the left column. The functionality is there with the popover, I'm just curious about the usefulness of this. I understand this fixes the onmouseover problem. Do people use this feature?
> 
> Note that the entered column is not always there. There is a course configuration option to have that not shown.

Didn't realize this.  Learning something else today.  :)

drgrice1:
> A mouseover trigger for the popup might address some of what I mentioned, but I think mouseover should be avoided. Is it even meaningful for a user on a tablet or phone?

@Alex-Jordan:  I am not trying to bring this back up, but just want to point out bootstraps behavior on this.  If the popup wer set to 'hover' (bootstrap's equivalent of mouseover), then the popover will appear when the mouse hovers over the answer, and on a touch screen (I assume this is what you meant by a tablet or phone), the popover will appear on touch.  I just wanted to point out that behavior for future reference.  I also don't want this to appear when the mouse hovers over the answer.  I would find that annoying.  A click is better.

PR #1223, opened 2021-03-04
Page Load speed tweak

drgrice1:
Defer loading of the the math4.js script and remove the jQuery ready
function.  Just execute it as soon as it is loaded.

In newer version of jQuery, jQuery's ready function is supposed the execute when the DOMContentLoaded event occurs, but it seems to actually occur a bit later.  Even the DOMContentLoaded event seems to run to late to prevent flickering.

This seems to be the best option without really fixing things the right way.  That will probably need to be deferred until after the ww-2.16 release though.

### DIFF
--- a/htdocs/themes/math4/gateway.template
+++ b/htdocs/themes/math4/gateway.template
@@ -48,7 +48,7 @@
 <script type="text/javascript" src="<!--#url type="webwork" name="htdocs"-->/js/vendor/bootstrap/js/bootstrap.js"></script>
 <link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="htdocs"-->/node_modules/jquery-ui-dist/jquery-ui.min.css"/>
 <script type="text/javascript" src="<!--#url type="webwork" name="htdocs"-->/node_modules/jquery-ui-dist/jquery-ui.min.js"></script>
-<script type="text/javascript" src="<!--#url type="webwork" name="theme"-->/math4.js"></script>	
+<script type="text/javascript" src="<!--#url type="webwork" name="theme"-->/math4.js" defer></script>	
 <!--#if exists="math4-overrides.js"-->
 <script type="text/javascript" src="<!--#url type="webwork" name="theme"-->/math4-overrides.js"></script>	
 <!--#endif-->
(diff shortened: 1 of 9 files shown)


drgrice1:
Many of the other instances where the jQuery ready function and window load event are used should also be reviewed.  The load event occurs later than both the jQuery ready function the window DOMContentLoaded event.  As stated on https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event "It is a common mistake to use the load event where the DOMContentLoaded event should be used."

drgrice1:
@Alex-Jordan:  I don't think this fixes the issue.  I suspect that this may improve once bootstrap is also updated.

drgrice1:
The real problem here is that almost none of what is done by javascript in the math4 jQuery ready function (or with this pull request in the DOMContentLoaded event handler) should be done by javascript at all.  Most of what is done there is adding css classes.  Those classes should be added by the page directly (in the ContentGenerator).  Then the styles would be applied immediately at page load.

This is why I have always seen this flickering even before the jQuery update.  The jQuery update just emphasized this.

drgrice1:
Alex, try this.  This should fix the flickering that you observed.

Alex-Jordan:
I tried this. I'm afraid it's not noticeably better. If it's clear that these are good changes anyway, I can merge it.

drgrice1:
That is odd. It is noticably faster for me.

drgrice1:
When you tested this, are you sure that your browser was loading the updated javascript?  Either clear your browser cache, or disable network caching and try again.  I see a significant improvement on the speed at which the styling on the buttons is applied by the javacript (which as I have said is still not the correct way for this to be done).

Alex-Jordan:
Yes, I think that I did clear the cache. I was very cache-aware after something else bit me yesterday. 

But then I think what happened is I loaded the page and saw the ~0.5 second flicker and called it based on that without more reloads. I didn't consider that this was the first retrieval of the .js. Now I'm still seeing the flicker but it is indeed noticeably better. Except, as you might expect, the first time after I clear the cache.

Alex-Jordan (approved):

PR #1224, opened 2021-03-04
Navigation restrictions for LMS

limefrogyank:
With excellent direction from Alex Jordan, I have added some modifications that give the ability to restrict navigation within WeBWorK.  

### Problem
WeBWorK can be used within an LMS where the problem sets are graded separately within the LMS.  However, this only works properly if the specific problem set that the LMS expects a grade for is navigated to directly from the LMS.  If the student then decides to do another problem set by navigating within WeBWorK, the grade for the second problem set is NOT sent back to the LMS.  It will only transfer if the link from the LMS is clicked.  This creates annoying sync issues.

If we restrict navigation by turning off all links within WeBWorK other than 
1. The User Settings link
2. Individual problems
3. Email TA
4. Hardcopy
then we can force the student to exit WeBWorK and navigate to the next problem set using the LMS.

### Solution
I created a new permission in `defaults.conf` called "navigation_allowed".  This is set to "guest" by default meaning that the default for everyone is allowed.

However, if you set `$permissionLevels{navigation_allowed} = "login_proctor";` or higher, then the students will no longer be able to see the sidebar links (except for "User Settings") nor the breadcrumb links.  Where the sibling window is supposed to show other problem sets, nothing is shown instead.  (The sibling window for the problem itself, where it shows other problems within a set, still shows.)  

Finally, when the student navigates to "User Settings" or "Hardcopy", they need to be able to navigate backwards.  If used within an LMS, they are most likely seeing an IFRAME and the back button on the browser won't work.  So I have rendered a "Go Back" button that simply runs a simple javascript command: "window.history.back()".

image

image

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -770,6 +770,22 @@ sub links {
 	my %systemlink_args;
 	$systemlink_args{params} = \%params if %params;
 
+    # A permission that is normally set to guest, but can be set to 'login_proctor' so that students don't see navigation links.
+	# This would be very useful when having students navigate to problem sets from an LMS and preventing the student from navigating
+	# to a different problem set from within WeBWorK.  Doing so prevents the LMS from receiving scores for the navigated problem sets. 
+	# Those problem sets must navigated to directly from the LMS for the grade to be sent back to the LMS gradebook. 
+	# This modification will only include the User Settings link so that students can switch between text and MathJax.
+	unless ($authz->hasPermissions($userID, "navigation_allowed")){
+	if (defined $courseID) {
+		if ($authen->was_verified) {
+			print CGI::start_ul();
+			print CGI::li(&$makelink("${pfx}Options", urlpath_args=>{%args}, systemlink_args=>\%systemlink_args));
+			print CGI::end_li();
+			}
+		}
+		return "";
+	}
+
 	print CGI::start_ul();
 	print CGI::start_li({class => "nav-header"});
 	print CGI::h2($r->maketext("Main Menu"));
@@ -1095,10 +1111,14 @@ associated with the current request.
 sub path {
 	my ($self, $args) = @_;
 	my $r = $self->r;
+    my $authz = $r->authz;
 
 	my @path;
 
 	my $urlpath = $r->urlpath;
+    my $courseID = $urlpath->arg("courseID");
+	my $userID = $r->param('user');
+
 	do {
 	    my $name = $urlpath->name;
 	    # If its a problemid for a jitar set (something which requires
@@ -1115,6 +1135,13 @@ sub path {
 	} while ($urlpath = $urlpath->parent);
 
 	$path[$#path] = ""; # we don't want the last path element to be a link
+
+    # restrict navigation capabilities if not allowed
+	if (defined $courseID) {
+		unless ($authz->hasPermissions($userID, "navigation_allowed")){
+			return "";
+		}
+	}
 
 	#print "\n<!-- BEGIN " . __PACKAGE__ . "::path -->\n";
 	print $self->pathMacro($args, @path);
--- a/lib/WeBWorK/ContentGenerator/ProblemSet.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSet.pm
@@ -109,10 +109,17 @@ sub nav {
 	my ($self, $args) = @_;
 	my $r = $self->r;
 	my $urlpath = $r->urlpath;
+    my $authz = $r->authz;
+	my $userID = $r->param('user');
 
 	my $courseID = $urlpath->arg("courseID");
 	#my $problemSetsPage = $urlpath->newFromModule("WeBWorK::ContentGenerator::ProblemSets",  $r, courseID => $courseID);
 	my $problemSetsPage = $urlpath->parent;
+
+    # If navigation is restricted, we don't want to allow a student to navigate to a different problem set.
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+	    return "";
+    }
 	
 	my @links = ($r->maketext("Homework Sets") , $r->location . $problemSetsPage->path, $r->maketext("Homework Sets"));
 	return $self->navMacro($args, '', @links);
@@ -190,6 +197,12 @@ sub siblings {
 				     && !$LTIRestricted;
 				           }   @setIDs;
 	}
+
+    # restrict navigation to other problem sets if not allowed
+	unless ($authz->hasPermissions($user, "navigation_allowed")){
+		return "";
+	}
+
 
 	print CGI::start_div({class=>"info-box", id=>"fisheye"});
 	print CGI::h2($r->maketext("Sets"));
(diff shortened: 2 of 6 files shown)


limefrogyank:
I should mention that we can further restrict navigation by removing or altering the Masthead using a theme template.  That's not part of the scope of this PR though.

Alex-Jordan:
Thanks Lee. I did some testing with this and here are some comments:

* In the code, could you change all leading whitespace to use tabs instead of spaces? More broadly, there are [these coding standards](https://webwork.maa.org/wiki/Coding_Standards). You may find that the old code adjacent to your new code is not conforming to those standards, and you could edit adjacent code too. But don't overhaul an entire file just for coding standards, or it will make the diff hard to read.
* I think that in practice this will be used on a course by course basis. Most typically when the instructor is using "homework" grade passback mode. So it should probably be an option in Course Config, in the permissions tab. Could you add that? I think it can all be done in [defaults.config](https://github.com/openwebwork/webwork2/blob/03eca4cc1a6034d24cf28bd690f0f56ce14f284b/conf/defaults.config#L1615)
* It works! Just as described.
* If I go to User Settings, I can change a setting, and click Save. This reloads the page. Now if I hit the Go Back button, I am back in User Settings from before the page reload. So I need to use Go Back again. I don't know what to do about that, while keeping the (clever!) back-navigation technique. 
* Probably, we should also let students visit the Achievements page (if Achievements are in use). This would be done the same way as with User Settings.
* I see no reason to let students visit Grades. I mention it in case others reading this want to bring that up, because it came up in an earlier conversation about this kind of feature. Grades would be a hassle, because that page will provide links to all homework sets, defeating the purpose. For the one set you are visiting, you can see problem scores right there. And also most likely grade passback is in use and students can see grades in the LMS.
* For testing, I just put the permission status as 'ta' and then logged in as a student using a password rather than honestly enter through the LMS. So I landed on the Homework Sets page, but without navigation. This was weird [but not problematic] because I can choose a problem set to go into, and then I am stuck there. OK fine. For the main use case, this will never happen.
* But the above made me realize. I made my way to a problem set: `https://webwork-dev.pcc.edu/webwork2/coursename/problemsetname/`. It won't take long for some students to figure out they can cut the "problemsetname/" from the URL, getting to `https://webwork-dev.pcc.edu/webwork2/coursename/`. I know it's impossible to truly cut off inappropriate navigation by savvy students who edit the address bar, but this route stands out as too easy. What ideas do we have about this? If a student has restricted navigation, and they go to `https://webwork-dev.pcc.edu/webwork2/coursename/`, what might we show them? Leave it as it is? Or print some message about the instructor needing to provide homework set specific links?

Nice work here, and I'm going to ask a few of the other active people to review this. I did not examine the actual code (yet). Others here are more likely to spot possible issues with those details.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -109,10 +109,17 @@ sub nav {
...
+	my $userID = $r->param('user');
 
 	my $courseID = $urlpath->arg("courseID");
 	#my $problemSetsPage = $urlpath->newFromModule("WeBWorK::ContentGenerator::ProblemSets",  $r, courseID => $courseID);
 	my $problemSetsPage = $urlpath->parent;
+
+    # If navigation is restricted, we don't want to allow a student to navigate to a different problem set.
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+	    return "";
+    }
```
drgrice1, line 122: It would be nicer to use the Perl idiom of `return "" unless ...` here.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -191,6 +198,12 @@ sub siblings {
 				           }   @setIDs;
 	}
 
+    # restrict navigation to other problem sets if not allowed
+	unless ($authz->hasPermissions($user, "navigation_allowed")){
+		return "";
```
drgrice1, line 203: Also use the `return "" unless ...` here.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -770,6 +770,22 @@ sub links {
...
+	unless ($authz->hasPermissions($userID, "navigation_allowed")){
+	if (defined $courseID) {
+		if ($authen->was_verified) {
+			print CGI::start_ul();
+			print CGI::li(&$makelink("${pfx}Options", urlpath_args=>{%args}, systemlink_args=>\%systemlink_args));
+			print CGI::end_li();
+			}
+		}
+		return "";
+	}
```
drgrice1, line 787: I am not sure about handling this here.  It might be better to move this down into the handling below (inside the `if (defined $courseID)` block beginning on what is now line 797.  Wrap the "Homework Sets" and "Grades" in checks on the "navigation_allowed" permission, but show the rest.  This would show the "Main Menu" header and "Courses" link (doesn't hurt).  The instructor tools don't need anything further as they won't be shown anyway.  The bug reporter could be debated on as to if that should be shown.

drgrice1 (commented):
On all of the pages that have a "Go Back" button now, you need to deal with the case that there is nothing in the window's history.  This happens on the "Show Me Another" page because that opens in a new tab or window, but could happen on any page.  For example, the instructor may provide a direct link to a specific problem in the LMS.

For the "Go Back" button it would be best not to use inline javascript (the `onClick` attribute), but rather add javascript in math4.js that sets this up.  Then you can deal with the lack of a window history and such better.  Though this is part of a bigger project for the webwork2 code in general.  Inline javascript is bad practice.

This is a good start on this concept.

limefrogyank:
Thanks for the feedback.  I'm in full grading mode right now, so I'll get to these quickly when I'm done grading all of my labs...

As far as truly restricting navigation, there are three scenarios with an LMS:  
1.  User logs into webwork completely separately.
    - I could disable the problemsets page completely and just display a message that the user must access this course from their LMS
    - Navigating by editing the url would still be an issue.  However, we could generate a hashcode from the problemset id and user id and require that as a parameter to view the assignment.  The hashcode generating code would be public, but this is not security... this is just a roadblock that is much harder to pass.
2.  User gets to webwork via the LMS using an IFRAME window.
   - This is the easiest scenario.  We just need to fix what is mentioned in the review.
3.  User gets to webwork via the LMS using a new browser window.
       - Navigating by editing the url would still be an issue.  Same solution as above.

Alex-Jordan:
There's no rush, of course. We could get this going on the PCC server at any time. And even if it doesn't make the 2.16 release candidate branch, field testing at PCC may be enough to work it in to 2.16 as well.

I think it would be wise to make a high-level decision before more coding: is this a feature for LMS users using "homework" grade passback? Or is this an orthogonal feature (restricted navigation) that happens to pair well with "homework" grade passback?

If it is the former, then I think it should be controlled by something in the LTI config instead of a general permission. Perhaps a third option for `$LTIGradeMode`. It could still reveal the Instructor Tool menus if the user has access to those.

If it's the latter, then it's better to plan abstractly, and not make decisions based on assumptions the user is using LTI. 

Personally I think there is value in doing it the way you have done it, as a permission, and orthogonal to LMS use. But at the same time, if you could assume LTI is in use, that could make things easier and seems to cover 99.99% of use cases.

PR #1226, opened 2021-03-04
Resolving memory leak issues

drdrew42:
Preamble: this PR will require that you have `Class::Accessor::Lite` installed (either libclass-accessor-lite-perl via apt, or cpan Class::Accessor::Lite).

The WWSafe compartment orphans all objects created within, causing significant memory leakage with every pg file that is processed (every preview, submitted attempt, LaTeX2PDF).

This PR attempts to resolve this issue by iterating through the WWSafe symbol-table in order to drop these references before the safe is closed.

With this WWSafe cleanup process in place, and the changes in openwebwork/pg#543, the core WeBWorK pipeline (PG::Local -> Translator -> WWSafe -> PGCore -> PGanswer/responsegroups and other PG children) unravels itself when the problem has finished rendering.  There are still some lingering issues with Value::Context and Parser::Context objects - so this is not a complete fix, but it should represent a significant improvement in the rate at which apache processes consume RAM.

**Changes:**
- CourseEnvironment: be a little pickier with what we 'pull' out of the safe and into the CE hash
- PG.pm: free method no longer needs to iterate through any of its references, just drop the safe reference from the translator (see comments in the code)
- AttemptsTable: the current use of `Class::Accessor` is preventing AttemptsTable objects from deallocating. We can get by with `Class::Accessor::Lite` in its place, which does not prevent deallocation. (Note to future self: this affects check_modules and Dockerfile)
- There is still a lingering issue with WeBWorK::Localize:I18N which I believe is related to AttemptsTable (via maketext). The changes to maketext are not significant, but they do allow Class::Accessor to 'protect' the coderef without throwing a warning that a r/o method is being replaced when `maketext` is declared.
- Finally, WWSafe: before emptying the symbol-table for the safe, iterate through the symbol-table and empty (if relevant) and delete references to objects from within the safe. 

Add the following at WWSafe, line 198, to watch the objects that are caught.
(This works in the standalone renderer, but IDK where STDOUT/STDERR are visible in ww2)
```perl
printf("[%-22s] %75s", $val, "$pkg$key should be removed ($isref)\n");
```


drgrice1:
I am closing this.  The corresponding PG pull request was closed, so this one should be as well.

Issue #1225, opened 2021-03-04
Problem Editor 3 bugs


Alex-Jordan:
From `develop`, I am experiencing a bug in Problem Editor 3. Video demonstrating the bug at:
http://spot.pcc.edu/~ajordan/ViewTakeActionBug.mp4

The "Take Action" button is not doing anything helpful. I think it reloads the page, but it does not do anything that responds to the tab selection (like View, Revert). In the video, the problem was a temporary verions so the Revert tab was present. I just tested with a different problem that had the Append tab present. This time, View still did not work. Append did append a copy of the problem to the problem set as problem 4. Then when I go to problem 4 and click Edit3, I see the Revert tab right away. But how could this be since I haven't yet had a chance to edit the problem and View it?

taniwallach:
I don't have develop running at present, but forked off develop and with some early merges of other changes, and I don't seem to see this issue/
  - https://github.com/taniwallach/webwork2/tree/jquery-update_and_PR_1190
  - https://github.com/taniwallach/webwork2/compare/jquery-update_and_PR_1190...openwebwork:develop

Maybe see if refreshing the page or bypassing the browser cache helps - maybe the major JS changes are not yet being properly loaded.

drgrice1:
I am also not seeing this either.

drgrice1:
I have been able to reproduce this now.  There is a minor issue with the javacript caused in a change in the way that jQuery handles things.  jQuery has stopped improperly allowing the checked property of a checkbox to be determined using the `attr` method.  I will create a pull request to fix this soon.

Although there are larger problems with the pg editor that are not new.  There are issues with how temporary file creation is dealt with.

drgrice1:
I want to add that most of this issue that you have stated is not javascript related, and is not new to the develop branch.  It may have been highlighted by the javascript issue though.  In any case that is what I was referring to in the previous comment about the issues with how temporary file creation is handled.

drgrice1:
One problem that I see using ww-21.5 (still present in develop) is the following:

Open a problem in a homework set that is from the OPL.  Then click `edit3`.

`View` is the default selected tab.  Now click `Take Action`.  This shows the problem (in a bootstrap modal by default).  It also creates a temporary file in tmpEdit even though no editing has been done, and there is no need for this.

Now close the editor window, and reload the problem in the homework.  The source file for the current user is now set to the temporary file.

Now click `edit3` again.

[...]

PR #1231, opened 2021-03-05
Fix the problem editor 3 take action button

drgrice1:
The checked property of a checkbox or radio button can not be determined using the .attr method for newer versions of jQuery.  jQuery was lenient in allowing this to begin with.

This fixes the parts of issue #1225 introduced by the change to the newer jQuery version.  This does not fix the parts of that issue that already existed.

### DIFF
--- a/htdocs/js/apps/PGProblemEditor3/pgproblemeditor3.js
+++ b/htdocs/js/apps/PGProblemEditor3/pgproblemeditor3.js
@@ -46,43 +46,43 @@ $(function(){
     
 });
 
-window.addEventListener("load", function () {
-    $('#submit_button_id').on('click',function() {
-	/* NOTE:  This makes a lot of the updateTarget()/setTarget() JS
-	   in the main pgeditor3 code superfluous.  Since we are doing a 
-	   just in time check to see what the target should be, it doesn't
-	   matter if we keep the target up to date before submit is pressed.
-	*/
+window.addEventListener("DOMContentLoaded", function () {
+	$('#submit_button_id').on('click', function() {
+		// NOTE:  This makes a lot of the updateTarget()/setTarget() JS
+		// in the main pgeditor3 code superfluous.  Since we are doing a
+		// just in time check to see what the target should be, it doesn't
+		// matter if we keep the target up to date before submit is pressed.
 
-	// action0 = view
-	// action1 = update
-	// action2 = new version
-	// action3 = append
-	// action 4 = revert
-	
-	var inWindow = $("#newWindow").attr('checked');
-	var target = "pg_editor_frame";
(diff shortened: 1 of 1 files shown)


drdrew42:
How should this PR be tested?

drgrice1:
It would be good to check that it works.  I of course tested it, and it works as before the jQuery merge for me.  @Alex-Jordan should at least check that the basic functionality of the `Take Action` button that he reported in issue #1225 is correct.

Alex-Jordan (approved):
With this PR, I'm unable to reproduce the issue as I recorded it in #1225.

drdrew42 (approved):
I have tested this PR and found that it successfully recognizes the checked property of 'open in a new tab' as well as other radio-button options in the pgProblemEditor3 interface.

pstaabp:
Walking through what @drgrice1 including in #1225, I was finally able to see what's going on.   

I switched back to develop and couldn't reproduce the bug that @Alex-Jordan posted on #1225.  Did the jquery update fix this automatically.  Since I couldn't reproduce, if @Alex-Jordan is satisfied with this fix, then I'm okay with it as well.

pstaabp (approved):
I just left a note on this that I couldn't reproduce, but am fine with this.

drgrice1:
The jquery merge caused this problem.  To reproduce the issue open a problem and click on `edit3` below.
Leave the `Open in New Window` checkbox unselected below and leave it set to `View`.
Then click on `Take Action`.
On the current develop branch the problem will open in a new window even though `Open in New Window` is not checked.
With this pull request the problem will open in a bootstrap modal in the current window.

To see the difference you could add the following code to the beginning of the `click` handler at the end of prproblemeditor3.js:
```javascript
	console.log("jquery attr check newWindow", $("#newWindow").attr('checked'));
	console.log("jquery attr check action2", $('#action2').attr('checked'));
	console.log("jquery attr check action3", $('#action3').attr('checked'));
	console.log("jquery attr check action4", $('#action4').attr('checked'));

	var win = document.getElementById('newWindow');
	var action2 = document.getElementById('action2');
	var action3 = document.getElementById('action3');
	var action4 = document.getElementById('action4');

	console.log("elt property checked newWindow", win.checked);
	console.log("elt property checked action2", action2.checked);
	console.log("elt property checked action3", action3.checked);
	console.log("elt property checked action4", action4.checked);

```
The jquery attr method returns `checked` for newWindow always, and undefined for the rest always.
The attribute properties of the elements return the correct values.

drgrice1:
Note that I could have also used `$('#eltid').is(":checked");` to stick with a jQuery solution that works.  I took the opportunity to start moving away from jQuery.

PR #1233, opened 2021-03-05
Problem Set Detail: view problem in versioned quiz, not undefined_set

Alex-Jordan:
Currently suppose a student has taken a quiz and has a quiz version. Make your way to the Set Detail page for that student, for that version: Classlist Editor, click the `#/#` for sets assigned to that student, and click the versioned quiz. Before this commit, if you click the eyeball icon, you will see the problem in an Undefined Set.

Since there is a version for this student, it seems better to use that. So now with this commit, you will be viewing the problem in the student's versioned quiz set.

This only affects the behavior of viewing the problem. If you click to _edit_ that problem, then view it from the editor, it is still usse an Undefined_Set. That sounds appropriate to me, but I haven't fully thought it through. Please comment if you feel that behavior should also change.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -2188,7 +2188,7 @@ sub body {
 				$viewProblemPage =
 					$urlpath->new(type =>'gateway_quiz',
 						      args => { courseID => $courseID,
-								setID => "Undefined_Set",
+								setID => ($editingSetVersion > 0) ? $fullSetID : "Undefined_Set",
 								problemID => "1" } );
 
 				my $seed = $problemToShow ? $problemToShow->problem_seed : "";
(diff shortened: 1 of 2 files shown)


drgrice1:
This doesn't seem to quite work as expected.  If you have a quiz with problems on multiple pages, then no matter which problem you click on your get the first page of the quiz.  Not the actual problem that you want to view.

Alex-Jordan:
I see. I was not testing with enough quiz variation.

So maybe this is not possible without a lot of new engineering, and maybe that's why it is the way it is. Because with a Gateway, there is not a URL to each problem or even to each page with a subset of the questions. It's all one HTML page, and as I move "page to page" or question to question, javascript is just hiding/unhiding things.

The individual questions do have id~s, starting with "prob0". That could be leveraged to jump to a particular problem if they are all visible, but even that leaves the issue of a target problem that is hidden on the opening page.

I'll let this sit a bit in case there is an idea to make it work. If not, I can close it.

drgrice1:
It is a good idea.  It will just take a bit more work to achieve.

PR #1237, opened 2021-03-07
Add a couple of missing periods on the show me another page.

drgrice1:
These have bugged me for a while.

Once merged I will update the translation file strings for this in PR #1214.


taniwallach (approved):
Other than the need to fix the translation files, this is just a minor change to some text messages.
I think this can be merged, and we can afford the delay before the translation files are updated via additions to https://github.com/openwebwork/webwork2/pull/1214

PR #1241, opened 2021-03-09
avoid a comparison with an undefined variable

Alex-Jordan:
If you have a set definition file that does not declare a reduce scoring date, then when you try to import it, there will be a comparison of that undefined date against 0, which gives an error/warning message. This edit simply pulls that little block of code up inside another conditional that was already present, that first checks if the reduced scoring date is defined.

To test, take a set definition file and manually edit it so that there is no reduced scoring date. Then import the problem set on`develop` and you should see the error. Then delete the problem set.

After merging this commit, import the problem set again and there will be no error.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList2.pm
@@ -2004,12 +2004,14 @@ sub readSetDef {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
I think I have seen this occur.  Clearly it would based on the logic of the code.  Good catch.  I tested this and it works as it should.

I am going to merge this.

PR #1245, opened 2021-03-12
Remove dragmath - Just a little clean up

drgrice1:
We no longer support this, and it is no longer used.


Alex-Jordan:
Should this also remove the folder `htdocs/dragmathedit`?

drgrice1:
Yes, it probably should.  I missed that.

Alex-Jordan:
As long as I was looking here....

Do these three things do anything, and should they be cleared too?
```
$pg{specialPGEnvironmentVars}{MathView} = 0;
$pg{specialPGEnvironmentVars}{WirisEditor} = 0;
$pg{specialPGEnvironmentVars}{MathQuill} = 0;
```

Instead of these booleans, there is `$pg{specialPGEnvironmentVars}{entryAssist}`.

I will merge this either way. But I'll wait to discuss this last thing I just brought up.

drgrice1:
I don't think they are used.  I didn't remove those as they are not related to dragmath.  This may be a point to mention on @dlglin's pull reqest #1247.

PR #1246, opened 2021-03-12
Correct URLs for mathjax-config.js in templates

drgrice1:
Fix the urls in the templates for the mathjax-config.js file, and add the mathjax javascript to the gateway.template for the math3 theme.

Some things I did incorrectly or missed in the mathjax update.


Alex-Jordan (approved):
The three math4 changes look right on a visual inspection of the code. No testing.

I think math3 could be unofficially retired with this release and officially cleared with the next one. I'm not sure what `polyfill.io` is and how to review that. But I am personally not concerned if we stop maintaining math3.

drgrice1:
I am in agreement on dropping math3.

`pollyfill.io` is javascript needed for MathJax on older browsers (IE11).  If we officially drop support for IE11, then we don't need that.

PR #1247, opened 2021-03-12
Clean up localOverrides.conf and defaults.config

dlglin:
Fix localOverrides.conf.dist so that no values are set in that file.
Values that were defined in both files are now commented out in localOverrides.conf.dist.  If the values differed, then the value from localOverrides.conf.dist was moved to defaults.config and then commented out.
This means that a fresh install of 2.16 will have the same defaults as a fresh install of 2.15, though if someone has an older version of localOverrides.conf along with a recent version of defaults.config, this may change a couple of settings for them.


conf/localOverrides.conf.dist
```diff
@@ -281,52 +281,44 @@ $webworkURLs{MathJax} = "$webworkURLs{htdocs}/mathjax/es5/tex-chtml.js";
...
- $pg{specialPGEnvironmentVars}{DragMath} = 0;
- $pg{specialPGEnvironmentVars}{CAPA_Tools}             = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_Tools/",
- $pg{specialPGEnvironmentVars}{CAPA_MCTools}           = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_MCTools/",
- $pg{specialPGEnvironmentVars}{CAPA_GraphicsDirectory} = "$courseDirs{templates}/Contrib/CAPA/CAPA_Graphics/",
- push @{$pg{directories}{macrosPath}},  
-   "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_Tools",
-   "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_MCTools";
-
- # The link Contrib in the course templates directory should point to ../webwork-open-problem-library/Contrib 
+# $pg{specialPGEnvironmentVars}{DragMath} = 0;
```
drgrice1, line 321: Just remove the DragMath option entirely.  It doesn't exist anymore.  Although, I have another pull request that removes this option anyway.

conf/localOverrides.conf.dist
```diff
@@ -349,9 +341,9 @@ $pg{options}{periodicRandomizationPeriod} = 5;
...
 
-$showeditors{pgproblemeditor3} = 1;
-$showeditors{pgproblemeditor2} = 0;
-$showeditors{simplepgeditor}   = 0;
+#$showeditors{pgproblemeditor3} = 1;
+#$showeditors{pgproblemeditor2} = 0;
+#$showeditors{simplepgeditor}   = 0;
 
 # This will turn off ww3 
 #$showeditors{ww3} = 0;
```
drgrice1, line 349: Delete the `ww3` option.  This is webwork2.  Furthermore, this option does not even work if enabled.  It is not in the code.  Also remove `ww3` from the %showeditors hash in defaults.config.

drgrice1 (approved):
This generally looks good, but I think a few more improvements are needed.  See my comments.

Alex-Jordan:
I resolved a conflict that had arisen, addressed Glenn's comments, removed the obsolete answer entry variables, and on `localOverrides` I did a once-through to (mostly) conform to coding standards.

One thing I changed that is not quite trivial. In addition to Glenn's suggested changes to the hardcopy theme options, I changed it so the theme name hash comes first. Then the theme options are defined as the keys of the theme name hash. I hope that was not a mistake.

drgrice1:
I added a bit more insignificant clean up (a quick `s/\s*$//` to remove white space at the end of lines, and a few minor indentation issues).  @Alex-Jordan covered my original comments.  I will go ahead and merge it.

Alex-Jordan:
It occurred to me that by using `keys` for the hardcopy themes, that will lead to the themes being presented in random order each time the hardcopy generator page is loaded. The default will still be selected, but the random order might be suboptimal.

One option is to return it to the way it was. I didn't like the required redundancy in that which was why I changed it. But I wouldn't object to changing it back.

Another option is that in `Hardcopy.pm`, it could alphabetize the options each time.

drgrice1:
Yeah, that is true.  I thought it was a good idea initially as well, but the order should be preserved that is set in the configuration.  So it will probably need to go back to the way it was.

PR #1248, opened 2021-03-13
Mathjax typeset on rerandomize in library browser

drgrice1:
In the library browser (setmaker) when the `Randomize` button is clicked on a problem, the TeX in the problem is not being typeset by MathJax.  The reason is that the html id of the element is passed and used and MathJax needs the element itself.

Without this pull request find some problems and click the randomize button.  The problem is reloaded and all TeX in the problem appears as `\(...\)`.

This was not introduced by the recent MathJax upgrade, but is also present in WeBWorK 2.15.
Edit:  Scratch that.  This was not present in WeBWorK 2.15.  My initial test was still on the develop branch.  I forgot my production server isn't on ww 2.15 anymore.

### DIFF
--- a/htdocs/js/apps/SetMaker/setmaker.js
+++ b/htdocs/js/apps/SetMaker/setmaker.js
@@ -412,7 +412,7 @@ function randomize(filepath, el) {
 	      // run typesetter depending on the displaymode
 	      if(displayMode=='MathJax')
               MathJax.startup.promise = MathJax.startup.promise.then(function() {
-				  return MathJax.typesetPromise([document.getElementById(el)]);
+				  return MathJax.typesetPromise(['#' + el]);
 			  });
 	      if(displayMode=='jsMath')
 		  jsMath.ProcessBeforeShowing(el);
--- a/htdocs/js/apps/SetMaker/setmaker.js
+++ b/htdocs/js/apps/SetMaker/setmaker.js
@@ -411,7 +411,9 @@ function randomize(filepath, el) {
 	      $('#'+el).html(data);
 	      // run typesetter depending on the displaymode
 	      if(displayMode=='MathJax')
-              MathJax.startup.promise = MathJax.startup.promise.then(function() { return MathJax.typesetPromise([el]); });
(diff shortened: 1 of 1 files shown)


htdocs/js/apps/SetMaker/setmaker.js
```diff
@@ -411,7 +411,9 @@ function randomize(filepath, el) {
 	      $('#'+el).html(data);
 	      // run typesetter depending on the displaymode
 	      if(displayMode=='MathJax')
-              MathJax.startup.promise = MathJax.startup.promise.then(function() { return MathJax.typesetPromise([el]); });
+              MathJax.startup.promise = MathJax.startup.promise.then(function() {
+				  return MathJax.typesetPromise([document.getElementById(el)]);
```
dpvc, line 415: The alternative would be to use `["#$el"]`, as MathJax v3 uses CSS selectors for the strings, so that the selection is more flexible.

drgrice1: I see.  I assume you mean `"#" + el`.

dpvc: Sorry, mixing up javascript and perl.  D'oh.  I just think of WeBWorK as perl, and didn't process that it is javascript.

drgrice1: Yeah, that is an easy trap to fall into.  I thought you were think (somewhat incorrect) ES6 (`` `#${el}` ``).

Alex-Jordan (approved):
Tested and it happens as described, both before and after the PR. Note for anyone testing, to clear the js cache between runs.

The randomizing is cleaner than it has been in the past. All I see is the numbers change, instead of a temporary compression of the bits in the problem.

drgrice1:
I suspect that is due to the new version of MathJax being faster.  @dpvc can confirm this.

dpvc:
Version 3 *is* faster, but the real reason is that version 3 doesn't do screen updates while it is running, like version 2 did.  So here, the update is followed immediately by MathJax processing the math, with no screen update in between.

PR #1250, opened 2021-03-14
don't hide list labels on lists in problem set detail problems

Alex-Jordan:
This is just a quick fix for an issue that bugged me as I am setting up courses for the next term.

Find a problem that has a list in it. For example, Library/PCC/BasicAlgebra/LinearInequalities/CompoundInequalityOr10.pg has an unordered list with some instructions for entering answers. Put this problem in a set, and then go to the Problem Set Detail page. Render all problems. The bullets for the ul will be missing.

This is because there is CSS that removes list labels, but the intent is just to remove the list labels on the problems themselves in the context of the problem set detail page, not on deeper lists within the problems. So the change to the CSS does that.

Perhaps it would be better to put a class on the problem set detail list and style that. This is a quick fix.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -1125,7 +1125,7 @@ div.problem_sets_options {
     padding-left: 0px;
 }
 
-#psd_list > li {
+#psd_list li.psd_list_row {
     list-style-type: none;
 }
 
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -1125,7 +1125,7 @@ div.problem_sets_options {
     padding-left: 0px;
 }
 
-#psd_list li {
+#psd_list > li {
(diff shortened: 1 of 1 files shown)


drgrice1 (commented):
This needs a bit more thought.  This messes up just-in-time assignments.

drgrice1:
Try this selector instead: `#psd_list li.psd_list_row`

Alex-Jordan:
Thanks, I didn't understand the jitar complication. I made that change.

drgrice1 (approved):
This looks good, and makes the css selector what it should have been.

This is a trivial change, so I will go ahead and merge it.

drgrice1:
> Thanks, I didn't understand the jitar complication. I made that change.

Yeah, I don't use jitar, but I keep an assignment on my test server with that for just this sort of thing.

PR #1257, opened 2021-03-17
Add output_CSS to gateway quizzes and add css files requested by problems to gatway quizzes

drgrice1:
Add an output_CSS method to GatewayQuiz.pm and the corresponding item in the template.  Add the method for adding css files passed from PG problems or via the extra_css_files key of the specialPGEnvironmentVars hash to that method.

Also make it so that js and css files can be added that are not in the css or js sub-directories, but any subdirectory of htdocs on the server.

Generally clean up the coding of those methods.


drgrice1:
Oops.  Something isn't right here.

PR #1259, opened 2021-03-18
Install codemirror and jquery.nestedSortable via npm

drgrice1:
Add codemirror and jquery.nestedSortable to the list of javascript libraries installed by npm.


Alex-Jordan:
Tested by intentionally not running npm install after switching branches. That broke what was supposed to break. Then npm install brought it back.

PR #1261, opened 2021-03-18
fix CAPA_Graphics symlink location

Alex-Jordan:
Just noticed this symlink was incorrect. Not sure if the OPL changed where it keeps CAPA_graphics, but right now, it is here: https://github.com/openwebwork/webwork-open-problem-library/tree/master/Contrib/CAPA/CAPA_Graphics


drgrice1:
That link has been broken since it was created.  Good fix.

PR #1263, opened 2021-03-18
Fix issues when viewing problems on another page from the library browser.

drgrice1:
Points are not defined for problems in the library browser, so don't try to show them.

The library browser does not record attempts, and so don't try to detect a browser refresh there.  This fixes issue openwebwork/pg#548.

These are both my fault.  I though I had checked the first, but didn't try the second.  I apparently don't use the library browser enough to catch these things.


Alex-Jordan:
If I am on develop and go through the steps I listed in openwebwork/pg#548, thigns are different. Now I get a pink PG error/warning message. So it just looks like more things evolved with develop since I first experienced this.

But with this PR, either version of the issue is gone.

PR #1264, opened 2021-03-18
Tweak gateway header

drgrice1:
Move the problem status message (about whether or not the answers were checked and recorded) up above the problem header in gateway quizzes.  Now that the problem points (and possibly source file path) are shown on the same line, this looks better.


Alex-Jordan:
OK, I'm back online for testing, and this checks out. Looks much better with this than without.

PR #1265, opened 2021-03-18
Move the definition of the TEXINPUTS environment variable

drgrice1:
Move the definition of the TEXINPUTS environment variable out of webwork.apache2.4-config.dist and just define it directly in
Hardcopy.pm.

The previous method works, but this makes it so that things don't break if a sysadmin does not update the webwork.apache2.4-config file.

On the other hand, if one had a desire to override the default location then one would now have to edit Hardcopy.pm.

So the decision as to whether or not this is the way to go comes down to whether or not this should be configurable.

### DIFF
--- a/conf/webwork.apache2.4-config.dist
+++ b/conf/webwork.apache2.4-config.dist
@@ -94,9 +94,6 @@ eval "use lib '$pg_dir/lib'"; die $@ if $@;
 $WeBWorK::Constants::PG_DIRECTORY = $pg_dir;
 $ENV{PG_ROOT} = $pg_dir;
 
-# Location of common tex input files for hardcopy generation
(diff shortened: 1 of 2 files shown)


drgrice1:
So I guess the configurability decision was made!

Alex-Jordan:
Yeah, I'm kind of in a mindset that there is only one or two more days before we need the pre-release branch, and, well, executive decisions. Also, getting pulled every which way by kids today. This is the 4th time I tried to finish writing this post.

But thinking again, could you make it so that it's:
(1) in the config file
(2) but if `$ENV{TEXINPUTS}` is undefined, then Hardcopy.pm uses the default that you just put there?

drgrice1:
Yeah, if we want it configurable then I can do that.  I thought someone might ask for that.

PR #1266, opened 2021-03-18
Hardcopy theme order

Alex-Jordan:
My development server is not fully functioning and PCC IT support is busy right now. So I'm not able to do real testing. I built this entirely within GitHub and could not test it, even for syntax typos.

But does this look like it will work?


Alex-Jordan:
Now that I'm back online for real testing, I see that `sort` is not allowed.

drgrice1:
Yeah, sort is not allowed in the safe zone.  PGsort is available though.  I am not keen on the sort idea though.  That doesn't preserve the given order, as the given order may not be alphabetical.  The current set up has twoColumn before oneColumn, so that would get changed.

Alex-Jordan:
This version used sort, but also commented something like "if you want some explicit order, do that here instead of this one-liner".

But new version coming up.

drgrice1:
I see.  I didn't read the file changes carefully.  I see that now.

PR #1267, opened 2021-03-18
Hardcopy theme order

Alex-Jordan:
This is to replace #1266. 

Also, I noticed `--shell-escape` was still present. We decided this needed to go, is that right?

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -164,17 +164,25 @@ $defaultThemeTemplate = "system";
 ################################################################################
 
 # Available Hardcopy themes (located in snippets)
-# Hardcopy theme names (pretty names for themes)
+# (their "internal" name and their pretty name for end users)
 $hardcopyThemeNames = {
 	oneColumn => 'One Column',
 	twoColumn => 'Two Columns',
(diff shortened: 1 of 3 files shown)


drgrice1:
We did decide to disable that, but we were going to go with `--no-shell-escape` as that completely disables TeX's `\write18` command.  Just removing the `--shell-escape` leaves that command active, but only allows specific commands listed in the texmf.cnf file.  Also @taniwallach also has this in pull request #1160.

Alex-Jordan:
I put `--no-shell-escape` in the three places, and put in some commentary.

drgrice1:
Looks good.

You gave up on using PGsort?

drgrice1 (approved):

Alex-Jordan:
> You gave up on using PGsort?

I did this before learning about PGsort from your comment.

drgrice1:
Yeah, could use it like:
```perl
$hardcopyThemes = PGsort(sub { return $_[0] cmp $_[1]; }, keys(%{$hardcopyThemeNames}));
```

drgrice1:
You don't need `return` in that.

PR #1268, opened 2021-03-19
Rename PGProblemEditor3 to PGProblemEditor, and remove all of the others

drgrice1:
In addition to removing all but one problem editor the inline javascript is mostly removed (all but one method), and opening in a new window is disabled for all actions except for those that view the problem.

Note that when you look at the diff of the PGProblemEditor.pm file in this pull request, you are actually comparing PGProblemEditor3.pm to the old PGProblemEditor.pm file.


pstaabp:
Are you getting rid of the modal popup completely?  It seems that in the ProblemEditor on View, there isn't a new window open, but replaces this.  I thought we had talked about only using that modal. 

My only thought is that I wonder if this will throw some people's workflow.  I wonder if we should allow the option of the popup modal versus the replacement of the page. We could do this with a simple flag. 

Lastly, I added the _Put comment in release notes_ label.  I think we need to document.

drgrice1:
@pstaabp:  I am not sure, but I think you might be having a caching issue again.  The modal is the only in-window option now.  It no longer reloads the page in the current window.  I don't think anyone needs the behavior of opening in the current window (reloading the page like PGProblemEditor2.pm did).

I don't think there really is anything that needs to be documented here.  This will just work.

pstaabp:
@drgrice1  yep.  Once again, I forgotten to clear the cache :(. This is what I expect now.

pstaabp (approved):
Looks solid and simplifies things.

Alex-Jordan:
An observation: when the modal popup happens, I momentarily see the WW side panel navigation there too. Then I think CSS kicks in and it clears away. Do we care?

Alex-Jordan:
If I am very fast, I am able to click that navigation and go places within the popup.

Alex-Jordan:
Although the modal popup has breadcrumb navigation too, so if we don't care about that, we wouldn't care about the side panel navigation either.

Alex-Jordan:
OK, aside from those comments, I had no issues with the problem editor. I used it from a problem ("Edit" link), from the Library Browser (pencil icon) and from Set Detail editor (pencil icon). Does anyone object to a merge, given the flash of the sidebar navigation? If that is a concern, it could be made into an issue to address later.

drgrice1:
It is javascript that kicks in. If it were css you wouldn't see it. css styles apply immediately. In any case that is how it was before, and is not new. I am working on that though. That won't be in this pull request.

Alex-Jordan:
> In any case that is how it was before, and is not new. 

I thought maybe that too. But in my current 2.15 production server, that is not the case (either I don't get that flash, or it is too fast to perceive). Do you mean it is "not new" as of the larger javascript changes in develop between 2.15 and 2.16?

drgrice1:
I mean it is the same as it was for WeBWorK 2.15.  The flicker is perhaps a bit more perceptive for the same reason as the other things you noticed that with.  It needs to be changed to use the DOMContentLoaded event instead of the jQuery ready function also.

pstaabp:
I would agree that we should hide the breadcrumbs too.  Maybe we can target this for rc2

taniwallach:
> I would agree that we should hide the breadcrumbs too. Maybe we can target this for rc2

Make an issue or a card on the project board. Hopefully then it will get done.

PR #1253, opened 2021-03-17
Cookie code refactor, Use CGI::Cookie and support samesite - PR # 1149 - for revisions

taniwallach:
This is a resubmit of https://github.com/openwebwork/webwork2/pull/1149 to add secure cookie support and support for "same-site" as it was decided that additional fixes are needed.

Please see the discussion at  https://github.com/openwebwork/webwork2/pull/1149 .

---

Main points = refactor the Cookie code:

1. Use `CGI::Cookie` instead of `Apache2:Cookie`, as the new code needs support for the `samesite` attribute.
2. Added CGI::Cookie to bin/check_modules.pl and Dockerfile.
    - Note: The support for samesite dates to June 2019 in CGI::Cookie 4.45.
3. Remove obsolete, commented out, code using cookies from `lib/WeBWorK/ContentGenerator/Logout.pm`.
4. Drop the constant `COOKIE_LIFESPAN` and instead allow setting cookie lifespan using site / course environment configuration variables.
    - `$CookieLifeTime`  - for when cookie based session management **is** in use - default to 6 hours.
    - `$CookieLifeTime2` - for when cookie based session management is **not** in use, defaults to 30 days.
5. Allow setting value of cookies `samesite` and `secure` attribute using site / course environment configuration variable:
    - `$CookieSameSite`
    - `$CookieSecure`

PR #1269, opened 2021-03-19
Cookie refactor same site v2 - replacing PR #1149

taniwallach:
Moved code to an updated branch of develop.

This is a resubmit of https://github.com/openwebwork/webwork2/pull/1149 to add secure cookie support and support for "same-site" as it was decided that additional fixes are needed.

Please see the discussion at  https://github.com/openwebwork/webwork2/pull/1149 .

---

Main points = refactor the Cookie code:

1. Use `CGI::Cookie` instead of `Apache2:Cookie`, as the new code needs support for the `samesite` attribute.
2. Added CGI::Cookie to bin/check_modules.pl and Dockerfile.
    - Note: The support for samesite dates to June 2019 in CGI::Cookie 4.45.
3. Remove obsolete, commented out, code using cookies from `lib/WeBWorK/ContentGenerator/Logout.pm`.
4. Drop the constant `COOKIE_LIFESPAN` and instead allow setting cookie lifespan using site / course environment configuration variables.
    - `$CookieLifeTime`  - for when cookie based session management **is** in use - default to 6 hours.
    - `$CookieLifeTime2` - for when cookie based session management is **not** in use, defaults to 30 days.
5. Allow setting value of cookies `samesite` and `secure` attribute using site / course environment configuration variable:
    - `$CookieSameSite`
    - `$CookieSecure`

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -940,18 +940,15 @@ $CookieSameSite = "Strict";
 # Set the value of the secure cookie attribute:
 $CookieSecure = 0; # Default is 0 here, as 1 will not work without https
 
-# At present the CookieLifeTime / CookieLifeTime2 only effect how long the
+# At present the CookieLifeTime setting only effect how long the
 # browser is to supposed to retain the cookie.
 #     https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
-
-# Set cookie lifetime for when $session_management_via eq "session_cookie"
 $CookieLifeTime = "+7d";
 
-# Set cookie lifetime for when $session_management_via ne "session_cookie"
-$CookieLifeTime2 = "+30d";
-
-# NOTE: the cookie lifespan settings use the CGI.pm relative time settings.
+# NOTE: In general the cookie lifespan settings use the CGI.pm relative time settings.
 # Search for "30 seconds from now" at https://metacpan.org/pod/CGI to see the various options.
+# A WW option is to set it to "session", in which case the Cookie will expire when the
+# browser session ends (a "session cookie").
 
 ################################################################################
 # PG subsystem options
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -940,6 +940,10 @@ $CookieSameSite = "Strict";
 # Set the value of the secure cookie attribute:
 $CookieSecure = 0; # Default is 0 here, as 1 will not work without https
 
+# At present the CookieLifeTime / CookieLifeTime2 only effect how long the
+# browser is to supposed to retain the cookie.
+#     https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
+
 # Set cookie lifetime for when $session_management_via eq "session_cookie"
 $CookieLifeTime = "+7d";
 
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -925,6 +925,31 @@ $session_management_via = "session_cookie";
 $caliper{enabled} = 0;
 
 ################################################################################
+# Cookie control settings
+################################################################################
+
+# The following variables can be set to control cookie behavior.
+
+# Set the value of the samesite attribute of the WeBWorK cookie:
+#    See: https://blog.chromium.org/2019/10/developers-get-ready-for-new.html
+#         https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+#         https://tools.ietf.org/html/draft-west-cookie-incrementalism-00
+
+$CookieSameSite = "Strict";
+
+# Set the value of the secure cookie attribute:
+$CookieSecure = 0; # Default is 0 here, as 1 will not work without https
+
+# Set cookie lifetime for when $session_management_via eq "session_cookie"
+$CookieLifeTime = "+7d";
+
+# Set cookie lifetime for when $session_management_via ne "session_cookie"
+$CookieLifeTime2 = "+30d";
+
+# NOTE: the cookie lifespan settings use the CGI.pm relative time settings.
+# Search for "30 seconds from now" at https://metacpan.org/pod/CGI to see the various options.
+
+################################################################################
 # PG subsystem options
 ################################################################################
 
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -787,16 +787,18 @@ sub check_session {
 	my $keyMatches = (defined $possibleKey and $possibleKey eq $Key->key);
 	
 	my $timestampValid=0;
-	if ( $ce->{session_management_via} eq "session_cookie"
-		&& $ce->{CookieSecure} && defined($self->{cookie_timestamp}) ) {
-		$timestampValid = ( time <= $self->{cookie_timestamp} + $ce->{CookieLifeTime} );
-	} else {
+# first part of if clause is disabled for now until we figure out long term fix for using cookies
+# safely (see pull request #576)   This means that the database key time is always being used
+# even when in "session_cookie" mode
+#	if ($ce -> {session_management_via} eq "session_cookie" and defined($self->{cookie_timestamp})) {
+#		$timestampValid = (time <= $self -> {cookie_timestamp} + $ce->{sessionKeyTimeout});
+#	} else {
 		$timestampValid = ( time <= $Key->timestamp() + $ce->{sessionKeyTimeout} );
 		if ($keyMatches and $timestampValid and $updateTimestamp) {
 			$Key->timestamp(time);
 			$db->putKey($Key);
 		}
-	}
+#	}
 	return (1, $keyMatches, $timestampValid);
 }
 
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -381,7 +381,7 @@ sub get_credentials {
 # else   use cookieKey for verification
 # else    use cookie user name but use password provided by request.
 
-		if (defined $r -> param("key")) {
+		if (defined $r->param("key")) {
 			$self->{user_id} = $r->param("user");
 			$self->{session_key} = $r->param("key");
 			$self->{password} = $r->param("passwd");
@@ -398,19 +398,24 @@ sub get_credentials {
 			$self->{login_type} = "normal";
 			$self->{credential_source} = "cookie";
 			$self->{user_id}     = trim($self->{user_id});
-			debug("cookie user '", $self->{user_id}, "' key '", $self->{session_key}, "' cookie_timestamp '", $self->{cookieTimeStamp}, "'");
+			debug(  "cookie user '", $self->{user_id},
+				"' key '", $self->{session_key},
+				"' cookie_timestamp '", $self->{cookieTimeStamp}, "' "
+			);
 			return 1;
 		} else {
-			$self -> {user_id} = $cookieUser;
-			$self -> {session_key} = $cookieKey; # will be undefined
-			$self -> {password} = $r->param("passwd");
-			$self -> {cookie_timestamp} = $cookieTimeStamp;
-			$self -> {login_type} = "normal";
-			$self -> {credential_source} = "params_and_cookie";
+			$self->{user_id} = $cookieUser;
+			$self->{session_key} = $cookieKey; # will be undefined
+			$self->{password} = $r->param("passwd");
+			$self->{cookie_timestamp} = $cookieTimeStamp;
+			$self->{login_type} = "normal";
+			$self->{credential_source} = "params_and_cookie";
 			$self->{user_id}     = trim($self->{user_id});
 			$self->{password}    = trim($self->{password});
-			debug("params and cookie user '", $self->{user_id}, "' params and cookie session key = '",
-				 $self->{session_key}, "' cookie_timestamp '", $self->{cookieTimeStamp}, "'");
+			debug(  "params and cookie user '", $self->{user_id},
+				"' params and cookie session key = '", $self->{session_key},
+				"' cookie_timestamp '", $self->{cookieTimeStamp}, "' "
+			);
 			return 1;
 		}	
 	}
@@ -435,7 +440,10 @@ sub get_credentials {
 		$self->{login_type} = "normal";
 		$self->{credential_source} = "cookie";
 		$self->{user_id}     = trim($self->{user_id});
-		debug("cookie user '", $self->{user_id}, "' key '", $self->{session_key}, "' cookie_timestamp '", $self->{cookieTimeStamp}, "'");
+		debug(  "cookie user '", $self->{user_id},
+			"' key '", $self->{session_key},
+			"' cookie_timestamp '", $self->{cookieTimeStamp}, "' "
+		);
 		return 1;
 	}
 }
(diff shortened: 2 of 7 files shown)


taniwallach:
This still has the issue which was discoved by @drgrice1 after the code was modified to try to check the cookie did not expire... see https://github.com/openwebwork/webwork2/pull/1149#discussion_r593836887

I ran into problems on my first attempt to fix that, and will resume on Sunday.

conf/defaults.config
```diff
@@ -924,6 +924,35 @@ $session_management_via = "session_cookie";
...
+
+# At present the CookieLifeTime / CookieLifeTime2 only effect how long the
+# browser is to supposed to retain the cookie.
+#     https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
+
+# Set cookie lifetime for when $session_management_via eq "session_cookie"
+$CookieLifeTime = "+7d";
+
+# Set cookie lifetime for when $session_management_via ne "session_cookie"
+$CookieLifeTime2 = "+30d";
```
drgrice1, line 951: I was thinking that it would be better to just have one `$CookieLifeTime` variable here.  It is silly to have two variables that are used in exactly the same way for the same reason, but conditionally on the `session_management_via` setting.  Just document here that when using session management via keys you probably want a longer setting than when using session management via cookies.

taniwallach:
In https://github.com/openwebwork/webwork2/pull/1269/commits/dfba7fc7f3e4e743596f678032e6668badab5feb I reverted back to the old approach of determining whether they session key is still valid, which is used whether or not session management depends on the cookie or not.

The following commit has some additions to documentation inside the code/config files, and has some formatting changes.

lib/WeBWorK/Authen.pm
```diff
@@ -916,27 +883,37 @@ sub sendCookie {
...
 	);
 
 	if ($ce->{session_management_via} ne "session_cookie") {
-		my $expires = time2str("%a, %d-%h-%Y %H:%M:%S %Z", time+COOKIE_LIFESPAN, "GMT");
-		$cookie -> expires($expires);
+		$cookie->expires( $lifetime2 );
+		$cookie->max_age( $lifetime2 );
+	} else {
+		$cookie->expires( $lifetime );
+		$cookie->max_age( $lifetime );
```
drgrice1, line 910: As I said in my other comment, I recommend getting rid of `CookieLifeTime2` (and `$lifetime2` here).  Furthermore, I recommend changing this to
```perl
if ($lifetime ne 'session') {
	$cookie->expires( $lifetime );
	$cookie->max_age( $lifetime );
}
```
Then add to the documentation to the config files explaining that the setting of `session` for `$CookieLifeTime` means that cookies will expire with the session.  Not calling `expires` and `max_age` is how you make that happen.

taniwallach: Done

conf/defaults.config
taniwallach: Done

lib/WeBWorK/Authen.pm
taniwallach: Done

taniwallach:
@pstaabp - This is pretty similar to what was in https://github.com/openwebwork/webwork2/pull/1149 when you reviewed it (but with only one `$CookieLifeTime` setting).

Due to the bug explained there and below tPR https://github.com/openwebwork/webwork2/pull/1149 was reverted after it was merged.

The bug I introduced late on was in https://github.com/openwebwork/webwork2/pull/1149/commits/0280ff6780a884a4fdaa0269cb8340324a0a2ce when I modified  `lib/WeBWorK/Authen.pm` with
```
	if ( $ce->{session_management_via} eq "session_cookie"
		&& $ce->{CookieSecure} && defined($self->{cookie_timestamp}) ) {
		$timestampValid = ( time <= $self->{cookie_timestamp} + $ce->{CookieLifeTime} ); 

```
as explained in https://github.com/openwebwork/webwork2/pull/1149#discussion_r583057948 .

The problem is that the changes treated `$ce->{CookieLifeTime}` as if it was a numeric value in seconds, while it is a specially parsed string, and that triggers an error.

In any case, upon further thought - trusting the cookie's timestamp value which may be tampered with on the client side is not really a good idea. The potential of such abuse was mentioned in https://github.com/openwebwork/webwork2/pull/576#issuecomment-98900894 and that PR had added to code *not** to use the cookie's timestamp.

I'm pretty confident that this version is OK, but it needs someone to test it.

pstaabp (approved):
this is looking good.  I tested in Safari, which doesn't give an error, but does detect that SameSite is strict.  Also tested in FF and without this patch, lots of warnings, but with the patch FF seems happy.

lib/WeBWorK/Authen.pm
```diff
@@ -77,7 +76,7 @@ our $GENERIC_ERROR_MESSAGE = "";  # define in new
 ## WeBWorK-tr end modification 
 #####################
 
-use constant COOKIE_LIFESPAN => 60*60*24*30; # 30 days
+# Replaced constant COOKIE_LIFESPAN with a course environment setting which defaults to 30 days.
```
drgrice1, line 79: We don't need comments referring to now non-existent constants.  So delete the comment.  It is also not correct for the actual defaults in defaults.config.  There the default is set to 7 days.

lib/WeBWorK/Cookie.pm
```diff
@@ -35,13 +34,13 @@ use warnings;
...
 
-# This class inherits from Apache::Cookie under mod_perl and Apache2::Cookie under mod_perl2
+# This class inherits from Apache::Cookie under mod_perl and CGI::Cookie under mod_perl2
 BEGIN {
 	if (MP2) {
-		require Apache2::Cookie;
-        require APR::Request::Error;
-		Apache2::Cookie->import;
-		push @WeBWorK::Cookie::ISA, "Apache2::Cookie";
+		#require APR::Request::Error;
```
drgrice1, line 40: This line is not needed anymore.  Delete!

lib/WeBWorK/Authen.pm
```diff
@@ -916,27 +883,40 @@ sub sendCookie {
...
  	my $timestamp = time();
-	
-	my $cookie = WeBWorK::Cookie->new($r,
-		-name    => "WeBWorKCourseAuthen.$courseID",
- 		-value   => "$userID\t$key\t$timestamp",
-		-path    => $ce->{webworkURLRoot},
-		-secure  => 0,
+
+	my $sameSite  = $ce->{CookieSameSite};
+	my $secure    = $ce->{CookieSecure};    # Warning: use 1 only if using https
```
drgrice1, line 892: There is no need to defined new variables here.  Just use `$ce->{CookieSameSite}` below in the one place that the variable `$sameSite` is used.  Same for `$secure`.

In general it is bad coding practice to define a variable if its value is only used once.  It is inefficient and clutters the variable namespace.  Furthermore, there is generally no need to define a variable for accessing a hash key.

You seem to do this sort of thing frequently, and the code could be made much more concise and efficient by not doing this.  It also improves readability of the code as it is easier to backtrack to where the value in use comes from.

lib/WeBWorK/Authen.pm
```diff
@@ -945,21 +925,25 @@ sub killCookie {
...
-	my $expires = time2str("%a, %d-%h-%Y %H:%M:%S %Z", time-60*60*24, "GMT");
-	my $cookie = WeBWorK::Cookie->new($r,
-		-name => "WeBWorKCourseAuthen.$courseID",
-		-value => "\t",
-		-expires => $expires,
-		-path => $ce->{webworkURLRoot},
-		-secure => 0,
+
+	my $sameSite  = $ce->{CookieSameSite};
+	my $secure    = $ce->{CookieSecure};    # Warning: use 1 only if using https
```
drgrice1, line 932: Again, just use `$ce->{...}` below.

drgrice1 (approved):
This looks good now.  The previous authentication methods are preserved.  This basically just replaces the outdated Apache::Cookie module with CGI::Cookie, and adds the functionality that module provides.  Everything works as it should.

I added a few code clean up suggestions, but other than that this is looking good.

lib/WeBWorK/Authen.pm
taniwallach: Done.

lib/WeBWorK/Authen.pm
taniwallach: Done. I'll try to avoid doing this in the future.

lib/WeBWorK/Authen.pm
taniwallach: Done

drgrice1:
This looks good.  I am re-targeting for WeBWorK-2.16.

drgrice1:
Nevermind.  Looks like that has been done.

lib/WeBWorK/Authen.pm
```diff
@@ -916,27 +882,37 @@ sub sendCookie {
...
-		-name    => "WeBWorKCourseAuthen.$courseID",
- 		-value   => "$userID\t$key\t$timestamp",
-		-path    => $ce->{webworkURLRoot},
-		-secure  => 0,
+
+	my $cookie = WeBWorK::Cookie->new(
+		-name     => "WeBWorKCourseAuthen.$courseID",
+		-value    => "$userID\t$key\t$timestamp",
+		-path     => $ce->{webworkURLRoot},
+		-samesite => $ce->{CookieSameSite};
```
Alex-Jordan, line 894: This has a semicolon where a comma should be. I am getting errors restarting apache now.

Alex-Jordan: (and same on next line)

pstaabp: Trying to think of the best way to handle this.  Should I rebase this?  If I can remember how to?

lib/WeBWorK/Authen.pm
```diff
@@ -945,21 +921,22 @@ sub killCookie {
...
-		-secure => 0,
+
+	my $cookie = WeBWorK::Cookie->new(
+		-name      => "WeBWorKCourseAuthen.$courseID",
+		-value     => "\t",
+		'-max-age' => "-1d", # 1 day ago
+		-expires   => "-1d", # 1 day ago
+		-path      => $ce->{webworkURLRoot},
+		-samesite  => $ce->{CookieSameSite};
+		-secure    => $ce->{CookieSecure};    # Warning: use 1 only if using https
```
Alex-Jordan, line 934: Same here, with semicolons and commas.

pstaabp: Alex, I just made a direct commit from the website.  Do a git pull  on the webwork2.16 branch and should be set.

drgrice1:
Something is wrong here.  This pull request did not merge correctly.

pstaabp:
There were errors, but try a `git pull` on this branch and restart.

pstaabp:
I mean on `WeBWorK-2.16` branch.

drgrice1:
@pstaab:  This is still not working.  No one do anything more.  I am working to fix it.  Don't merge anything.

pstaabp:
Sorry about that.  Not sure what I did.   I'm not having any troubles with the 2.16 branch.

drgrice1:
I don't think it is what you did.  I think it is how @taniwallach redid the branch.

drgrice1:
I think the only way to fix this is to reset the branch back to before this was merged.  The only thing after this was my TEXINPUTS change to the apache2 config file.   I will have to resubmit that, but @taniwallach is going to have to rework this pull request.  Do not merge it until I check it over.  I should have tested the merge before.

pstaabp:
What issue are you seeing?

drgrice1:
It is done.  You will most likely need to run `git reset --hard origin WeBWorK-2.16` (and possible the same for develop) after running `git pull` (or better `git fetch`).

@pstaab: Almost none of this pull request actually went in.

@taniwallach:  I can help you fix this pull request.

Alex-Jordan:
What I did just now.

1. checkout develop
2. fetch origin (github.com/openwebwork/webwork2)
3. pull develop
4. delete local WeBWorK-2.16 branch
5. checkout new WeBWorK-2.16 from origin/WeBWorK-2.16

Should that leave me in good shape?

drgrice1:
That will work.

pstaabp:
That's roughly what I did as well.

drgrice1:
Okay.  This pull request is in now.  I had to rebuild the pull request into a new commit.

pstaabp:
@drgrice1 Are you saying this is now merged correctly in WeBWorK-2.16?

lib/WeBWorK/Authen.pm
taniwallach: Ouch - that was a late change made late at night, but I did not see if trigger an error. Sorry.

taniwallach:
I apologize for whatever made the mess, I had cherry-picked the commits from the old branch onto an update copy of develop (which was at bf11e63ee6bbe8ea9ee64e6c760f173e9c5ec8bb). That certainly worked on my branch, but apparently due to the merge+revert of the original branch - merging the new branch did not work.

Alex-Jordan:
No problem. But is someone able to assess if this will be resolved (as in, the intended commits get merged) within the hour or so? I could wait to being upgrading my production server, but not much longer than that. Is it helpful that this make it in, for field testing?

taniwallach:
> No problem. But is someone able to assess if this will be resolved (as in, the intended commits get merged) within the hour or so? I could wait to being upgrading my production server, but not much longer than that. Is it helpful that this make it in, for field testing?

@Alex-Jordan - I just checked out a clean copy of the WeBWorK-2.16 branch (after a git fetch) and compared using `diff -r` to the code in the branch this PR was supposed to be sourced from.  The code which this PR was supposed to modify is all done using @drgrice1's new 0db5cd6297b13c49684563c6dfe98755edb4cc73, and the other changes all seem related to several recently merged PRs, so it looks to me like everything in the WeBWorK-2.16 branch should be good now.

@drgrice1 Thanks for cleaning up my mess. Sorry. Do you know what caused the problem, so I can avoid it happening again.

pstaabp:
@taniwallach I think the cherry-picking can be dangerous.  In hindsight, you probably should have stayed with the original PR.   If a PR is going into a branch like `develop`, you can merge `develop` into that branch (especially if it has been a while since since branching from the original).

drgrice1:
@taniwallach:  I am not entirely certain.  I just know that none of the config file changes actually appeared in the code after the merge was done.  The cherry-picking should have worked.  I have seen things like this before.  That is where GitHub shows things in the GUI in the browser, but what actually happens is different.

It is all fixed now, so no harm done.

taniwallach:
I don't think the original PR branch could work once the original PR was reverted. The commits were the same ones reverted - so apparently cannot be "reapplied" as is from the same "feature branch". When I quickly tried to make https://github.com/openwebwork/webwork2/pull/1253 and target master by accident and tried to retarget to develop it just made a huge mess. Apparently this is well-known to the experts, but confused plenty of people:

  - https://stackoverflow.com/questions/27852143/how-to-pr-and-merge-again-after-reverting-pr-using-github-revert-button
  - https://medium.com/@shanikae/create-pull-request-from-a-reverted-git-branch-27219cadf9b5
  - https://itnext.io/git-revert-the-revert-88b1e66d71d4

I suspect that the cherry picking from reverted commits is somehow a problem, but I am not certain.

If I am pulling out the correct lines from my `.bash_history` file, what I did was:
```
git checkout develop
git checkout -b cookie-refactor-same-site-v2
git checkout cookie-refactor-same-site
git log
git checkout cookie-refactor-same-site-v2
git cherry-pick edf75ca0d44c29de100f527ce3185a0b5bd03089..78b819dfc918ed7cb3ec3772e5b121927771c737
git status
vim conf/localOverrides.conf.dist
git add conf/localOverrides.conf.dist
git cherry-pick --continue
vim conf/defaults.config
git add conf/defaults.config
git cherry-pick --continue
vim lib/WeBWorK/Cookie.pm lib/WeBWorK/Authen.pm
git push --set-upstream tani cookie-refactor-same-site-v2
```
and then additional changes and commits were made.

---

We certainly have observed that comparing master and develop in GitHub shows something quite different from what `diff -r` would show.

---

I think we may find it safer to make small reverting changes directly using a new PR instead of reverting a merged PR for a small issue. The "bug" in the original PR (absolutely my fault) could have been corrected without reverting the entire PR.

Sorry again for all the mess.

PR #1270, opened 2021-03-20
Make it possible to override the default location of the common tex input files.

drgrice1:
Make it possible to override the default location of the common tex input files so that it is possible for the hardcore tex enthusiasts to modify those files.

This was suggested by @Alex-Jordan in #1265.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -14,11 +14,6 @@
 # Artistic License for more details.
 ################################################################################
 
-BEGIN {
-	$ENV{TEXINPUTS} = ".:$ENV{WEBWORK_ROOT}/conf/snippets/hardcopyThemes/common:"
-	unless defined($ENV{TEXINPUTS});
-}
-
 package WeBWorK::ContentGenerator::Hardcopy;
 use base qw(WeBWorK::ContentGenerator);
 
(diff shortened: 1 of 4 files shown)


pstaabp (approved):
Looks great.

taniwallach:
I see that the setting does work during PDF generation, as I could have it use a modified version of `PGML.tex` from a special location set using this setting.
However, I don't think that the (new) ZIP file creation will collect the files from the modified location when this is in use.

drgrice1:
@taniwallach:  Good catch.  I will fix that.

drgrice1:
@taniwallach:  The zip file that is generated now uses the configured location.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -866,7 +869,7 @@ sub generate_hardcopy_tex {
 
 	# Copy the common tex files into the bundle directory
 	my $ce = $self->r->ce;
-	my $commonTeXDir = "$ce->{webworkDirs}{conf}/snippets/hardcopyThemes/common";
+	my $commonTeXDir = $ENV{TEXINPUTS} =~ s/^\.:([^:]*):$/$1/r;
```
Alex-Jordan, line 872: It would (should) never happen, but the internet tells me it is legal for a linux filname to have a colon. If `$ENV{TEXINPUTS}` had a colon in the middle of it, that would make `$commonTeXDir` come out as null, I think.

Is it safe to just use  `s/^\.:(.*):$/$1/r`?

taniwallach: @drgrice1 Would it be possible to instead allow a `course.conf` / CourseEnvironment setting to provide a (single) path to the "local" tex common files, and then to build the value of `$commonTeXDir` and `$ENV{TEXINPUTS}` at run-time from that setting?

drgrice1: @Alex-Jordan: Although a linux file or directory name may have a colon in it, a directory with a colon in it will not work in a PATH environment variable.  See https://stackoverflow.com/questions/14661373/how-to-escape-colon-in-path-on-unix for example.  The TEXINPUTS variable uses the PATH environment variable structure.

@taniwallach:  It should be possible to do that.  I was going for a somewhat quick solution.  I don't expect that this will get used much.  It is very easy to break hardcopy generation with this.

Alex-Jordan: OK. Unless either of you would like to pursue Tani's idea, I would go ahead and merge this. Tani's idea could be filed as an issue. If Tani reads this and agrees, Tani, please merge it and make that issue.

drgrice1:
I am working on implementing Tani's idea, so give me a few minutes.

drgrice1:
@taniwallach, @Alex-Jordan:  Okay.  I think this is good. Test it again of course.

Alex-Jordan:
Tested, and I was able to change configuration in the localOverrides, and also in the apache config file, with the desired result.

drgrice1:
Uh, I meant to delete the TEXINPUTS stuff from webwork.apache2.4-config.dist. That won't work anymore.  I must have just deleted from my local version.

Alex-Jordan:
Oh, oops. I was scratching me head a little about that. I thought maybe both mattered. Well, it didn't "work" but it of course caused no issues. It defaulted to the default.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
taniwallach: Saw late, after this was merged. Will try to test the revised version tomorrow in the WeBWorK-2.16 branch.

PR #1273, opened 2021-03-22
Remove Dancer and Dancer::Plugin::Database from bin/check_modules.pl and Dockerfile.

drgrice1:
Those perl modules are not used by webwork2 (and will not be used by ww3 either).

The fact that these are here has bugged me for a while now.


Alex-Jordan:
Merging without testing.

PR #1277, opened 2021-03-23
Update OPL macro folders

Alex-Jordan:
Some OPL macro folders were omitted from defaults.config. Which was hard to tell because they were not in alphabetical order.


drgrice1 (approved):
This looks good.  The list matches the directories I currently see in the OPL.

drgrice1:
This can be merged.

PR #1279, opened 2021-03-23
provision for undefined  in AttemptsTable.pm

Alex-Jordan:
I don't recall why, but my production server has an edit with `$rh_answers//()` replacing `$rh_answers`. It must have come up at some point. It seems like a good change, or at least harmless.

While I was there, I cleaned indentation for this block.


lib/WeBWorK/Utils/AttemptsTable.pm
```diff
@@ -166,22 +166,22 @@ sub new {
 	ref($rh_answers) =~/HASH/ or die "The first entry to AttemptsTable must be a hash of answers";
 	my %options = @_; # optional:  displayMode=>, submitted=>, imgGen=>, ce=> 
 	my $self = {
-		answers      		=> $rh_answers,
+		answers             => $rh_answers//(),
 		answerOrder         => $options{answerOrder}//(),
```
drgrice1, line 170: `answers` needs to be a hash reference.  So this should be `answers => $rh_answer // {},`.

Note that `answerOrder` needs to be an array reference, and so that should be `answerOrder => $options{answerOrder} // [],`.

I am not sure that it matters too much though.  Looking at all of the places that AttemptsTable is used in the code, these are always guaranteed to be defined.

Alex-Jordan:
I made the `{}` and `[]` and uniformly put spaces around the `//` operator in this block.

drgrice1:
Looks good.  Nice clean up.  Merging.

PR #1281, opened 2021-03-23
turn GW 'test' into 'version'

Alex-Jordan:
Instructors use Gateway for more than things that are "test"s. For example, quizzes.

Also, instructors have actual tests where they put "Test" in the set name. This has led to a student seeing confusing things like "Test 1 (test 2)" where "Test" is in the set name, and "test" is really the version. Or "Take Test 1 test".

So this commit tries to change the language so that "version" is used in all the appropriate places.


drgrice1 (approved):
This has always bugged me to.

drgrice1:
There are a few of these strings that are untranslated.  The reason for this is the "prettification" of the set names that is done using regular expression replacement.  That should be done differently so that translations work.  You don't need to fix that now though.

I think this is good.  I will merge it.

PR #1282, opened 2021-03-23
display problem sets by urgency, with GW versions under their GW temp…

Alex-Jordan:
…late

Before this commit, the problem sets on the main course page are displayed:
* all non-GW sets first
* then GW templates
* then GW versioned sets

This can mean an urgent quiz is not visible until scrolling down off screen, if there are many non-GW sets.

The commit changes it so that non-GW and GW templates are mixed, sorted by urgency. Then, immediately following each GW template set, its versions come in their natural order.

(It would be nice if the versioned GW sets could be indented under their GW template set. But this whole thing is built as a table, and I'm not sure the right way to do that.)

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -413,11 +413,6 @@ table.classlist-table .tbody {
 	padding-left:40px;
 }
 
-/* 80px chosen so "Start a version of" never has to wrap at default zoom */
-.problem_set_table tr.gw-version td.gw-version {
-	min-width:80px;
-}
-
 /* these are for indenting problems on jitar sets */
 .problem_set_table .nested-problem-1 {
     margin-left: 10px;
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -413,6 +413,10 @@ table.classlist-table .tbody {
 	padding-left:40px;
 }
 
+.problem_set_table a.gw-parenthetical {
+	font-weight:normal;
+}
+
 /* these are for indenting problems on jitar sets */
 .problem_set_table .nested-problem-1 {
     margin-left: 10px;
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -413,6 +413,11 @@ table.classlist-table .tbody {
 	padding-left:40px;
 }
 
+/* 80px chosen so "Start a version of" never has to wrap at default zoom */
+.problem_set_table tr.gw-version td.gw-version {
+	min-width:80px;
+}
+
 /* these are for indenting problems on jitar sets */
 .problem_set_table .nested-problem-1 {
     margin-left: 10px;
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -409,6 +409,9 @@ table.classlist-table .tbody {
     white-space: nowrap;
 }
 
+.problem_set_table tr.gw-version td.gw-version {
+	padding-left:40px;
+}
 
 /* these are for indenting problems on jitar sets */
 .problem_set_table .nested-problem-1 {
--- a/lib/WeBWorK/ContentGenerator/ProblemSets.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSets.pm
@@ -478,9 +478,9 @@ sub setListRow {
 	      }  
 	      if ( $preOpenSets ) {
 		# reset the link
-		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of [_1]", $display_name));
+		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of[_1][_2]", CGI::br(), $display_name));
 	      } else {
-		$interactive = $r->maketext("Start a version of [_1]", $display_name);
+		$interactive = $r->maketext("Start a version of[_1][_2]", CGI::br(), $display_name);
 	      }
 	      $control = "";
 	      
@@ -506,20 +506,20 @@ sub setListRow {
 	      if ($setIsOpen ||  $preOpenSets ) {
 		# reset the link
 		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL},
-				      $r->maketext("Start a version of [_1].", $display_name));
+				      $r->maketext("Start a version of[_1][_2].", CGI::br(), $display_name));
 		$control = "";
 	      } else {
 		$control = "";
-		$interactive = $r->maketext("Start a version of [_1].", $display_name);
+		$interactive = $r->maketext("Start a version of[_1][_2].", CGI::br(), $display_name);
 	      }
 	    } else {
 	      $status = $r->maketext("Closed.");
 	    
 	      if ( $authz->hasPermissions( $user, "record_answers_after_due_date" ) ) {
-		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of [_1]", $display_name));
+		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of[_1][_2]", CGI::br(), $display_name));
 	      
 	      } else {
-		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of [_1]", $display_name));
+		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of[_1][_2]", CGI::br(), $display_name));
 	      }
 	    }
 	  } 
@@ -639,7 +639,7 @@ sub setListRow {
 	  
 	  return CGI::Tr(($gwtype == 1) ? {class => 'gw-version'} : {},
 		  CGI::td($control).
-		  CGI::td(($gwtype == 1) ? {class => 'gw-version'} : {},$interactive).
+		  CGI::td(($gwtype == 1) ? {class => 'gw-version'} : ($gwtype == 2) ? {class => 'gw-template'} : {},$interactive).
 		  CGI::td([$status,$score,$startTime]));
 	}
       }
--- a/lib/WeBWorK/ContentGenerator/ProblemSets.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSets.pm
@@ -266,9 +266,9 @@ sub body {
 	    print CGI::Tr(
 		CGI::th({-scope=>"col"},CGI::div({class=>"sr-only"},$r->maketext("Download Hardcopy"))),
 		CGI::th({-scope=>"col"},$nameHeader),
-		CGI::th({-scope=>"col"},$r->maketext("Test Score")),
-		CGI::th({-scope=>"col"},$r->maketext("Test Date")),
 		CGI::th({-scope=>"col"},$statusHeader),
+		CGI::th({-scope=>"col"},$r->maketext("Score")),
+		CGI::th({-scope=>"col"},$r->maketext("Start Date")),
 	        );
 	}
 
@@ -280,23 +280,31 @@ sub body {
 	@gwSets = grep {$_->assignment_type !~ /proctored/ || $viewPr} @gwSets;
 	
 	if ( $sort eq 'name' ) {
-	    @nonGWsets = sortByName("set_id", @nonGWsets);
-	    @gwSets = sortByName("set_id", @gwSets);
+		@sets = sortByName("set_id", @nonGWsets, @gwSets);
 	} elsif ( $sort eq 'status' ) {
-	    @nonGWsets = sort byUrgency  @nonGWsets;
-	    @gwSets = sort byUrgency @gwSets;
+		@sets = sort byUrgency (@nonGWsets, @gwSets );
 	}
-# we sort set versions by name
+# build a tree for set versions
+# first sort set versions by parent set name, then by version
 	@vSets = sortByName(["set_id", "version_id"], @vSets);
+	my %vSetTree = ();
+	# initialize keys as the parent GW set ids
+	for my $vset (@vSets) {
+		$vSetTree{$vset->set_id} = [];
+	}
+	#push in versioned sets as values
+	for my $vset (@vSets) {
+		push(@{$vSetTree{$vset->set_id}},$vset);
+	}
 
-# put together a complete list of sorted sets to consider
-	@sets = sort byUrgency (@nonGWsets, @gwSets );
-	
 	debug("End preparing merged sets");
 
-# we do regular sets and the gateway set templates separately
-# from the actual set-versions, to avoid managing a tricky test
-# for a version number that may not exist
+# Regular sets and gateway template sets are merged, but sorted either by name or urgency.
+# Immediately following a gateway template set comes its set versions.
+# Note: this assumes that the set_id of a versioned GW set cannot differ from all of the corresponding GW versions.
+# In a future version of WeBWorK where instructors can change names of sets, that may not be a safe assumption.
+# Similarly, properties like visibility of a set version are directly interpreted from that property of the template set.
+# If that changes in the future, then this needs to be revisited.
 	foreach my $set (@sets) {
 		die "set $set not defined" unless $set;
 		
@@ -304,7 +312,7 @@ sub body {
 			print $self->setListRow($set, $authz->hasPermissions($user, "view_multiple_sets"), $authz->hasPermissions($user, "view_unopened_sets"),$existVersions,$db);
 		}
 		if (defined($gwSetNames{$set->set_id})) {
-			foreach my $vset (@vSets) {
+			foreach my $vset (@{$vSetTree{$set->set_id}}) {
 				die "set $vset not defined" unless $vset;
 				if (($set->set_id eq $vset->set_id) && ($vset->visible || $authz->hasPermissions($user, "view_hidden_sets"))) {
 					print $self->setListRow($vset, $authz->hasPermissions($user, "view_multiple_sets"), $authz->hasPermissions($user, "view_unopened_sets"),$existVersions,$db,1, $gwSetsBySetID{$vset->{set_id}});  # 1 = gateway, versioned set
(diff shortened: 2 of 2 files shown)


drgrice1 (approved):
Another nice little improvement that I have also thought would be better.  I also agree it would be nice to indent the versioned gateway quizzes.  Ditto on the difficulty.

Clearly we are of the same mind on a lot of these things, and I suspect many others are as well.

drgrice1:
Does anyone disagree with this new ordering on the homework sets page?  If not I will merge this.

drdrew42:
I think this is an excellent modification

Alex-Jordan:
Stand by. I would like to try something that will make the version quizzes be indented.

Alex-Jordan:
Also, I just noticed I left commentary in the file that describes the prior ordering. And I think I made a mistake for when someone wants to sort by name.

drgrice1:
I messed with the indentation briefly.  I tried injecting a `&nbsp;`.  It worked but is not enough.  I didn't put any more thought into it though.

Alex-Jordan:
OK, this is ready for another look. I rebased because of the "test" to "version" change already merged. And so then I had to force push this.

Alex-Jordan:
Changes:
* The new row ordering already described.
* I put the two quiz-specific columns at the far right. Once quizzes and regular problem sets are mixed, it was too busy to have those columns in their original location (3rd and 4th out of 5).
* There is a hard break after "Start a version of" so that the quiz name starts on a new line.
* Removed the quiz name from the versions, since that is redundant now that they follow the quiz template.
* The versions are indented.

Alex-Jordan:
When there are no GW quizzes, now things look off. The status column fills out the table width, which I think is not a good look. This appears to be some side effect of the math4 edits I made. Any suggestions for doing it better?

Alex-Jordan:
Nevermind about my last comment. It is a change from 2.15, but it is not a change having to do with this PR.

Alex-Jordan:
Screenshot of the new order and layout:

Screen Shot 2021-03-23 at 6 01 34 PM


And here is the layout when there are no GW quizzes. But this is how it looks without this PR too.

Screen Shot 2021-03-23 at 6 02 02 PM

htdocs/themes/math4/math4.css
```diff
@@ -409,6 +409,14 @@ table.classlist-table .tbody {
...
 }
 
+.problem_set_table tr.gw-version td.gw-version {
+	padding-left:40px;
+}
+
+/* 80px chosen so "Start a version of" never has to wrap at default zoom */
+.problem_set_table tr.gw-version td.gw-version {
+	min-width:80px;
+}
```
drgrice1, line 419: These css selectors are identical, so just put the styles together like
```css
.problem_set_table tr.gw-version td.gw-version {
	padding-left:40px;
	/* 80px chosen so "Start a version of" never has to wrap at default zoom */
	min-width:80px;
}
```

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -280,35 +280,45 @@ sub body {
...
-	    @gwSets = sortByName("set_id", @gwSets);
+		@sets = sortByName("set_id", @nonGWsets, @gwSets);
 	} elsif ( $sort eq 'status' ) {
-	    @nonGWsets = sort byUrgency  @nonGWsets;
-	    @gwSets = sort byUrgency @gwSets;
+		@sets = sort byUrgency (@nonGWsets, @gwSets );
 	}
-# we sort set versions by name
+# build a tree for set versions
+# first sort set versions by parent set name, then by version
```
drgrice1, line 288: Please indent these comments to the same level as the code.  I know that is the way that it was, but that was bad.  It makes it harder to find where code blocks start and end.

Do the same with the other comments you modified/added.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -280,35 +280,45 @@ sub body {
...
 
-# we do regular sets and the gateway set templates separately
-# from the actual set-versions, to avoid managing a tricky test
-# for a version number that may not exist
+# Regular sets and gateway template sets are merged, but sorted either by name or urgency.
+# Immediately following a gateway template set comes its set versions.
+# Note: this assumes that the set_id of a versioned GW set cannot differ from all of the corresponding GW versions.
+# In a future version of WeBWorK where instructors can change names of sets, that may not be a safe assumption.
+# Similarly, properties like visibility of a set version are directly interpreted from that property of the template set.
+# If that changes in the future, then this needs to be revisited.
```
drgrice1, line 307: No need to comment about future versions of WeBWorK and what may happen there.  If the changes you mention happen, everything will need to be revisited regarding set names.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -456,7 +466,7 @@ sub setListRow {
 	    # reset the link to give the test number
 	    my $vnum = $set->version_id;
 	    $interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(), href=>$interactiveURL},
-				  $r->maketext("[_1] (version [_2])", $display_name, $vnum));
+				  $r->maketext("(version\xA0[_2])", $display_name, $vnum));
```
drgrice1, line 469: Why is the `\xA0` here? A space will work the same here.

Also, remove the `$display_name` from the list of substitutions.  Change it to
`$r->maketext("(version [_1])", $vnum));`

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -468,9 +478,9 @@ sub setListRow {
 	      }  
 	      if ( $preOpenSets ) {
 		# reset the link
-		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of [_1]", $display_name));
+		$interactive = CGI::a({class=>"set-id-tooltip", "data-toggle"=>"tooltip", "data-placement"=>"right", title=>"", "data-original-title"=>$globalSet->description(),href=>$interactiveURL}, $r->maketext("Start a version of[_1][_2]", CGI::br(), $display_name));
 	      } else {
-		$interactive = $r->maketext("Start a version of [_1]", $display_name);
+		$interactive = $r->maketext("Start a version of[_1][_2]", CGI::br(), $display_name);
```
drgrice1, line 483: This is a bit of an odd construction for a translatable string.  It would probably be best to just change it to:
`$interactive = $r->maketext("Start a version of [_1]", CGI::br() . $display_name);`
The two consecutive substitutions can be confusing to the translator.  The space in between "of" and `[_1]` won't matter in the html output.

Do the same with the other places that you do this.

drgrice1 (commented):
This looks good, but there are a few changes I suggested.

The issue when there are no versioned sets could be resolved with some css style.  You say that is different than for WW 2.15?  The only difference in this file from WW 2.15 is your "test" to "version" change.  I don't think there has been anything else that would affect the display of the sets table either.  The jQuery/MathJax changes wouldn't do that.

Alex-Jordan:
So, the fact that these examples have long reduced scoring messages is what gave me that impression. Typically (including where I looked on 2.15 servers) that long message is not present, and I didn't think about that.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
Alex-Jordan: The `\xA0` (non breaking space) is important. On a small screen, the number will break onto the next line without this. At least it did in my testing, which probably included long reduced scoring messages that influenced that column getting too narrow.

Not sure what I was thinking with `[_2]`.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
Alex-Jordan: I had it that way at first. But I think there are other languages where it is awkward to put the object of this phrase at the end of the phrase. In Hungarian for example you may need to put a grammatical suffix after the name of the set in this phrase. Your suggestion would force the set name to be the last thing. Combine that with the need (or desire, anyway) to break to a new line before printing the set name, and this is what I came up with. A translation could ignore the `[_1]` if that translation doesn't need the newline.

Alex-Jordan:
Thanks for the review, and I'll circle back to address those items soon before calling it a day.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
drgrice1: I am not sure that you should be trying to prevent the line from breaking.  If the screen becomes so narrow that it does not fit onto the line, the text should be allowed to go onto the next line.  Although, even better is if the table is made so that the content fits.  It seems the reduced scoring messages should be on a new line in the status column.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
drgrice1: That is true about sentence structure in other languages.  That may be an argument not to have the set name on another line though.  Another thing to consider is to change the text from `Start a version of [_1]` to just `Start [_1]` without the new line.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
Alex-Jordan: I think things like "version 2" are a unit that should never be broken. Similar to "Dr. Frankenstein". Or more similar to "Theorem 2.5.1" from a textbook. We use unbreakable spaces for things like that in PreTeXt, and that guides my thinking here. It's better to make the long string in the status column break a little more, but the web browser algorithm isn't going to do that on its own.

Now, I did this before reaching the issue that led me to put a `min-width` in the CSS for the "Start a new version of" string. Once that is there, this non-breaking space is moot. But I'm more inclined to think the `min-width` is the mistake. 

I'm going to look into some hard line breaks in the reduced scoring messages. My thinking is that will be a good thing because it will make them easier to read. A bonus side effect might be that the `min-width` is unnecessary, and I can cut that part out.

htdocs/themes/math4/math4.css
Alex-Jordan: This is a mistake. The `min-width` was supposed to go on `td.gw-template`, not `td.gw-version`. That would only matter where there is a quiz you haven't taken yet.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
Alex-Jordan: One thing that set me off on putting the set name on its own line, is if you click to sort alphabetically by name, it is by the set name, not by the phrase "Start a new...". If the set name is not on a new line, it's hard to follow the alphabetical sequencing of it all.

I wish English made `Start a version of [_1]` more concise. I think `Start [_1]` may be confusing when there is already a version started. It could be read as "start working where I left off". But it's not going to do that; it's going to start a new version.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
Alex-Jordan: I've done what I described in the last paragraph. On a computer/laptop/tablet it breaks in good places. On a small screen (specifically, as small as a phone), we see `Start a version\nof\nQuiz` but I think that is OK.

Alex-Jordan:
Ready for another look. Here is a screenshot full of quizzes with lots of different settings.

Screen Shot 2021-03-23 at 11 07 00 PM

Alex-Jordan:
There are problems with any of the messages, including `Start [_1]` and `Take [_1]`. Between how many versions are permitted in the current time interval, whether or not the current time interval is infinite, and how many versions the student has already started in the current time interval, there's always a scenario where the wording is misleading.

When you click a template set, one of two things happens. If a new version is available to you in the current time interval, you start a new version. Even if you already started a version previousely. Otherwise, you are sent to the most recent version you initiated. (This is why "Start" is misleading.) The logic for which thing happens is in Gateway.pm and ProctoredGateway.pm, not in ProblemSets.pm. So it is too much right now to pre-determine which will happen from ProblemSets.pm and have casewise options like "Start" and "Continue". Even if you could do that, if the screen sits long enough for a time interval to elapse, you would have bad information on the page.

What if it all goes away? Consider:
* Gateway template sets just print their name. No "Start" or "Start a new version of".
* Versions appear immediately below their template, and indented.

Something should still indicate that the set is a quiz. Or the student could set off a timer before they are ready. (I know we want a confirmation page, but that is a future project.) Now in many cases it will be there in the set name, like "Week_1_Quiz". But what is a reliable, accessible way to indicate that this set is a quiz?

One option is an icon of a clock. All quizzes are timed, although for some of theme the close date is the end of the time interval. (And there must be a close date.) A clock icon could precede or follow the template set name.

Here is a mockup screenshot:, including a "solid" clock icon and a "regular" clock icon:

Screen Shot 2021-03-24 at 2 15 03 PM

drgrice1:
Sounds good.  You could use the font awesome icon `fas fa-clock` or `far fa-clock`.

Alex-Jordan:
It would also be easy to make the clock icon only appear when there is a defined time limit (as opposed to the close date being the stand-in for the time limit).

drgrice1:
I like your mock up.  Make it so!

drgrice1:
I was also thinking that it wouldn't be hard to make it so that when you click on a link for a timed quiz, instead of opening the quiz immediately, it opens a confirmation dialog.  I could try to work that in.

Alex-Jordan:
To confirm: I'm only going to put the clock icon there if it is a timed set (not counting the close date) and the set has not yet closed. Any reason to do it differently?

Also I cannot decide if the "regular" clock or "solid" clock is better. So just going with regular. Holler if solid is better.

Alex-Jordan:
Coding standards question: do we care about `if ( conditions )` versus `if (conditions)`?

drgrice1:
> Coding standards question: do we care about `if ( conditions )` versus `if (conditions)`?

If you ask me I will say `if (conditions)`.  If you ask @taniwallach, I am betting he will say `if ( conditions )`.

drdrew42:
> Coding standards question: do we care about `if ( conditions )` versus `if (conditions)`?

perltidy would say `if ( conditions )`

drgrice1:
perltidy has options that also make it say `if (conditions)`.

drdrew42:
sorry, "default" perltidy would say it ;P

drgrice1:
Parenthesis in proper grammar should never have the spaces.  I have always been for following proper grammar in coding syntax, unless there is a good reason for the divergence.  I have not heard a good reason to diverge in this case.

drgrice1:
@Alex-Jordan:  I put in a pull request to update that icon.  You can use that if you want.

Alex-Jordan:
OK, it's in there. Clock icon on sets that are still open and have a timer. And I merged that icon replacement. Turns out that was not in a place where I did lots of code cleanup.

drgrice1 (approved):
This looks nice.

@drdrew42:  Could you take a look?

Alex-Jordan:
GitHub thinks there is a conflict. I think it is a false alert, some consequence of the branch that made that icon change. I'm going to resolve it by going 100% with `GWorder` version, which does not overlap where you made the icon change.

dlglin:
> 
> 
> Parenthesis in proper grammar should never have the spaces. I have always been for following proper grammar in coding syntax, unless there is a good reason for the divergence. I have not heard a good reason to diverge in this case.

Social distancing?

drgrice1:
@Alex-Jordan:  I think the conflict was caused by me merging your LMS name pull request.

@dlglin:  Nice!

Alex-Jordan:
@drgrice1 Ah, thanks for catching that. I just put `LMS_name` back in.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -437,85 +444,124 @@ sub setListRow {
...
+					title=>"",
+					"data-original-title"=>$globalSet->description(),
+					href=>$interactiveURL
+				},
+				$r->maketext("(version\xA0[_1])", $vnum)
+			);
+		} else {
+			my $t = time();
+			if ($set->{version_time_limit} > 0 && $t < $set->due_date()) {
+				$display_name = CGI::i({class=>"far fa-clock", title=>$r->maketext("This is a quiz with a time limit")}, '') . ' ' . $display_name;
```
drgrice1, line 474: Should this have the screen reader alternate text for the clock icon?  To add that change this to
```perl
$display_name = CGI::i({class=>"icon far fa-clock", title=>$r->maketext("This is a quiz with a time limit"), data_alt => $r->maketext("Timed quiz")}, '') . ' ' . $display_name;
```
So add the `icon` class, and the `data_alt` attribute.  I am not sure what the text should be there.  With what I put in there the screen reader would get `Timed quiz $display_name`.

Alex-Jordan: Absolutely. Thanks for catching that. I went with something that ends with a period so there is a pause before the set's name is read.

drgrice1:
I think this is ready now.  I am going to merge it.

Alex-Jordan:
I visited a test course as a student, where there are quizzes present. For a student, the download column is not check boxes. It has download icons. It looks like this:

Screen Shot 2021-04-20 at 10 03 19 PM

I momentarily thought something was broken from this PR. The download icons mess with my visual perception of the nesting. So in that shot for example, it sort of looks like "Quiz" (a GW set) is a child of "Testing" (a homework set). And "Quiz"'s version 1 is harder to perceive as a child of "Quiz" because of its download icon.

Does anyone else think this should be made better? And what ideas do you have to do that? One idea is to swap what is currently the first and second columsn, so the download icons are in the second column. Another idea is to merge the columns, maybe putting the download icons after the set name. I'm not in love with either idea.

drgrice1:
Another option is to remove the download icon from the page entirely, and make it so that to download an assignment the students need to enter the homework assignment the same as an instructor does.  To be quite frank, I am betting that students use that less than instructors do if even at all.

dlglin:
I agree with @Alex-Jordan that this makes the nesting less obvious.

I had two thoughts about how to fix this:

1. Instead of giving the download link/checkbox its own cell, just prepend it to the text of the Name cell.
2. Split the Name column into two columns.  In the case of an assignment (and the header), use a colspan to cover both columns.  In the case of a GW quiz, use the first of these columns for the download link/checkbox, and the second for the text.

If memory serves, someone tried option 1 a long time ago, and there was some problem, but I don't remember what it was.  Option 2 might be unnecessarily complicated to code.

dlglin:
When you mouseover the clock icon next to a GW quiz, you now get "Quiz with time limit".  How hard would it be to include the actual time limit in that message.

I should probably put this next suggestion in its own issue, but how hard would it be to add a js confirm box when clicking on a GW quiz that shows the time limit and number of submissions?

Alex-Jordan:
I can't speak for pandemic times, but in the before times, our students used the hard copy. It was common to go to the library or tutoring centers and see students with printouts from WeBWorK. Of course that could be done from the problem set page, but my sense is that these download icons get their attention, and the blue button down at the bottom of the page is less noticed.

> the same as an instructor does

I usually get the PDFs from the main homework sets page where these icons are. Except for an instructor, they are checkboxes so I could select more than one if I wanted to.

@dlglin What would you suggest for your item 2 for the instructor version that currently uses check boxes?

Add to the drawing board: the downloading thing (icon/check box) could be moved to the far right column.

drgrice1:
What do you not like about putting the download icon in the second column?  That seems like it would look good.

Alex-Jordan:
It might look good. But because parent quizzes don't get that icon, it will not run uniformly down the column. That kind of irregularity was why I moved the quiz version scores and start times to the far right. The "Status" column has this clean left edge, and maybe (maybe) those icons would disrupt that.

I'll mock up these ideas (and any more that come in) and see if there is a clear winner. Or post pictures to get opinions.

drgrice1:
After doing some quick hacking and experimentation, to me putting the checkbox/icon at the end looks good.  In fact, I like that better than how it was before.

Alex-Jordan:
That's good enough for me. If you would like to submit that, please do. Otherwise I will do this over the weekend.

PR #1283, opened 2021-03-23
add formattedReducedScoringDate environment variable

Alex-Jordan:
This is a nice variable to have in the environment for set header files, if you want to say something there about the reduced scoring date.


drgrice1 (approved):
Looks good.  This makes that set date work like the rest.  I will go ahead and merge this one.

PR #1285, opened 2021-03-24
LMS url in LTI config

Alex-Jordan:
This formalizes a clever hack that @drgrice1 posted in #1280. Separate from `$LMS_name`, you have `$LMS_url`. If the latter exists and is not empty, things that previously used plain `$LMS_name` will make a link with `$LMS_name` as the text, and `$LMS_url` as the href.

I made this change at all places where `$LMS_name` was used except in ProblemSets.pl. In that location, the LMS is referenced in a problem set's status field to say "you will need to go in through the LMS to get to this set". In practice in my own courses, the LMS may have release conditions that do not yet allow the student to get to the ELT link for that assignment. So it doesn't feel right to encourage the student to actually follow a link at that time.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Login.pm
+++ b/lib/WeBWorK/ContentGenerator/Login.pm
@@ -206,15 +206,11 @@ sub body {
 
 	if ($externalAuth ) {
 		my $LMS = ($ce->{LMS_url}) ? CGI::a({href => $ce->{LMS_url}},$ce->{LMS_name}) : $ce->{LMS_name};
-		if ($authen_error) {
-			if ($r -> authen() eq "WeBWorK::Authen::LTIBasic") {
-				print CGI::p($r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $LMS));
-			} else {
-				print CGI::p($r->maketext("_EXTERNAL_AUTH_MESSAGE", CGI::strong($course), $LMS));
-			}
+		if (!$authen_error || $r->authen() eq "WeBWorK::Authen::LTIBasic") {
+			print CGI::p($r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $LMS));
 		} else {
-			print CGI::p({}, $r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $LMS));
-		} 
+			print CGI::p($r->maketext("_EXTERNAL_AUTH_MESSAGE", CGI::strong($course), $LMS));
+		}
 	} else {
 		print CGI::p($r->maketext("Please enter your username and password for [_1] below:", CGI::b($course)));
 		if ($ce -> {session_management_via} ne "session_cookie") {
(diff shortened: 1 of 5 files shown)


lib/WeBWorK/ContentGenerator/Login.pm
```diff
@@ -205,14 +205,15 @@ sub body {
...
 			if ($r -> authen() eq "WeBWorK::Authen::LTIBasic") {
-				print CGI::p($r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $ce->{LMS_name}));
+				print CGI::p($r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $LMS));
 			} else {
-				print CGI::p($r->maketext("_EXTERNAL_AUTH_MESSAGE", CGI::strong($course), $ce->{LMS_name}));
+				print CGI::p($r->maketext("_EXTERNAL_AUTH_MESSAGE", CGI::strong($course), $LMS));
 			}
 		} else {
-			print CGI::p({}, $r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $ce->{LMS_name}));
+			print CGI::p({}, $r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $LMS));
```
drgrice1, line 216: This could be done a bit simpler.  You can leave it, but it could be done as:
```perl
if (!$authen_error || $r->authen() eq "WeBWorK::Authen::LTIBasic") {
	print CGI::p($r->maketext('The course [_1] uses an external authentication system ([_2]). Please return to that system to access this course.', CGI::strong($course), $LMS));
} else {
	print CGI::p($r->maketext("_EXTERNAL_AUTH_MESSAGE", CGI::strong($course), $LMS));
}
```
I held back on commenting on this in your last pull request, but I can't hold back the second time I see this.

drgrice1 (approved):
This looks okay.  If one wants the link in all of the locations, one can just use the `LMS_name` as I suggested before.  Although this is a bit directed toward your usage.

Alex-Jordan:
>  If one wants the link in all of the locations, one can just use the LMS_name as I suggested before. 

They could. It seems wrong to not separate the text from the href though. Can we imagine using `$LMS_name` with output that is not HTML output? In principle, I could, even though I probably never will in practice. Like I could distribute problem sets where the set header file mentions `$LMS_name`. As in "log in to `$LMS_name` to upload all your scratch work for this exam." And then as PDF output, the HTML content wouldn't come out well.

drgrice1:
I mention it because I most likely will.  I want the link in the locations you omitted as well.  I don't use LMS access in the same way that you do.  I just won't use $LMS_name in the set header, or if I do I will use a regular expression to extract the name portion.

drgrice1:
Actually, I realized the two places in ProblemSets.pm don't matter to me.  My students never see that as I use an approach similar to #1224 to prevent students from going there.  Of course they can still get there by modifying the url in the address bar, but I doubt they ever do.

lib/WeBWorK/ContentGenerator/Login.pm
Alex-Jordan: This is the conditional structure that was there before, it looks like for about the past 6 years. I can tidy it up though.

drgrice1: I know.  I was just suggesting a clean up.

drgrice1 (approved):
This looks good.

drgrice1:
I think I will go ahead an put it in.

PR #1287, opened 2021-03-25
Install Mathjax via npm

drgrice1:
This is a simple change that simplifies the webwork install process a bit.

### DIFF
--- a/htdocs/mathjax
+++ b/htdocs/mathjax
@@ -1,1 +1,1 @@
-../../MathJax/
+node_modules/mathjax
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -28,6 +28,11 @@
             "resolved": "https://registry.npmjs.org/jquery-ui-themes/-/jquery-ui-themes-1.12.0.tgz",
             "integrity": "sha1-pXugrZaADYRSL+dok//L3mcIHVE="
         },
+        "mathjax": {
+            "version": "3.1.2",
+            "resolved": "https://registry.npmjs.org/mathjax/-/mathjax-3.1.2.tgz",
+            "integrity": "sha512-BojKspBv4nNWzO1wC6VEI+g9gHDOhkaGHGgLxXkasdU4pwjdO5AXD5M/wcLPkXYPjZ/N+6sU8rjQTlyvN2cWiQ=="
+        },
         "nestedSortable": {
             "version": "1.3.4",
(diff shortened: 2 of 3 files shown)


drdrew42 (approved):

drdrew42:
makes total sense

drgrice1:
I was looking at the release notes, and realized we could remove that step quite easily.

drdrew42:
everyone has to `npm install` for this update anyhow

Alex-Jordan (approved):
Looks good and works.

Testing this made me look closely at math rendering on the page. Maybe we already talked about it, but this is another thing that happens slowly now, compared to 2.15. Not because of this PR. But on 2.15 servers, math for a problem is rendered pretty fast, while now it's taking around a full second.

drgrice1:
> Testing this made me look closely at math rendering on the page. Maybe we already talked about it, but this is another thing that happens slowly now, compared to 2.15. Not because of this PR. But on 2.15 servers, math for a problem is rendered pretty fast, while now it's taking around a full second.

That is in the difference between MathJax version 2 and 3.  The initial rendering on the page is a bit slower to kick in, but the entirety of the page being completed is much faster.  Particularly if there is quite a bit of TeX to be dealt with.  I have also noticed that rendering of things added to the page later is much faster though.  @dpvc can probably give better details.

dpvc:
I think there are two issues at play.  First, the v3 MathJax download is larger (not is all in one file, rather than several files, so only one download, but larger).  That means MathJax can't start as quickly (but as Glenn says, it runs faster once it does).

The second affect is subtler.  Because MathJax v2 was both slower and had to wait for other files to load, it would stop periodically to allow the screen to refresh.  Version 3 doesn't do that.  So in v2, the page would refresh before Mathjax would run, and it would refresh after every few equations were processed.  That gaves the psychological impression that MathJax was operating faster than it really was.  With v3, the math doesn't show up until it is all processed, and although that occurs faster than in v2, v2 could show you some results earlier.  In fact, in some cases, with v3 you may not see the page at all until the math is processed (no flash as math first appears untypeset, then typeset), so the effect may seem longer, even though the full page is ready faster than it was in v2.

pstaabp (approved):
In the documentation, we could add that the `$WEBWORK_ROOT/MathJax` directory can be deleted.

drgrice1:
Yeah, if upgrading that will not be needed anymore.

taniwallach (approved):
Great idea.

PR #1288, opened 2021-03-25
Add an option to disable MathQuill answer boxes per response 

drgrice1:
Add an option to disable MathQuill answer boxes per response (not quite per answer blank, but good enough for most cases).
    
To do so use `$ans->cmp(mathQuillOpts => "disabled")`.
    
This also changes the way the mathQuillOpts answer hash key works.  Previously to disable the spaceBehavesLikeTabs option (i.e. to enable typing of spaces) and the disable rootsAreExponents you would need to do
      `$ans->cmp(mathQuillOpts => "spaceBehavesLikeTab: false, rootsAreExponents: false")`
Now you will do
      `$ans->cmp(mathQuillOpts => { spaceBehavesLikeTab => 0, rootsAreExponents => 0 })`

This is paired with the pg pull request https://github.com/openwebwork/pg/pull/555.

### DIFF
--- a/htdocs/js/apps/MathQuill/mqeditor.css
+++ b/htdocs/js/apps/MathQuill/mqeditor.css
@@ -32,7 +32,7 @@ span[id^=mq-answer]
     margin-left: 0
 }
 
-input[type=text].codeshard
+input[type=text].codeshard.mq-edit
 {
     display: none !important;
 }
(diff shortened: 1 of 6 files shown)


Alex-Jordan:
This was not working with `$ans->cmp(mathQuillOpts => "disable")`. I looked at the code, and saw that to make it work needs `$ans->cmp(mathQuillOpts => "disabled")`. It does work once I make that change.

Do you want to leave it with "disabled" or change it to work for "disable"?

I also needed to use `spaceBehavesLikeTab` instead of `spaceBehavesLikeTabs`. That worked, as well as `rootsAreExponents`.

Alex-Jordan (approved):
Approved both here and in the PG counterpart.

PR #1289, opened 2021-03-25
Stop making symlinks to flash applets in the OPL when OPL-update is run.

drgrice1:
Just a little clean up.

### DIFF
--- a/bin/OPL-update
+++ b/bin/OPL-update
@@ -610,11 +610,6 @@ sub pgfiles {
 	my @textproblems = (-1);
 #print "\n$name";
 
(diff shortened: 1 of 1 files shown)


taniwallach (approved):
With flash no longer working in modern browsers, this makes sense.

Having this in `OPL_update` and the approach whereby the OPL tables get loaded from a downloaded database dump don't really work together.

If there are future needs for OPL content to make something available under `$ce->{webworkDirs}->{htdocs}` we will need a better approach to handling that.

Alex-Jordan (approved):

PR #1290, opened 2021-03-26
Trim leading and trailing space when passwords are set.

drgrice1:
There is some code clean up that I did while initially thinking I would trim passwords where they were set.  Then I found the quick and simple solution.  Passwords are trimmed when cryptPassword is called.

Do we also want to trim passwords when they are entered to sign in?  That is also not hard to accomplish.

### DIFF
--- a/bin/change_user_id
+++ b/bin/change_user_id
@@ -27,7 +27,7 @@ BEGIN {
 use lib "$ENV{WEBWORK_ROOT}/lib";
 use WeBWorK::CourseEnvironment;
 use WeBWorK::DB;
-use WeBWorK::Utils qw(runtime_use readFile cryptPassword);
+use WeBWorK::Utils qw(runtime_use readFile);
 use Data::Dumper;
 
 if((scalar(@ARGV) != 3)) {
--- a/bin/delcourse
+++ b/bin/delcourse
@@ -96,7 +96,7 @@ use lib "$ENV{PG_ROOT}/lib";
 
 use WeBWorK::CourseEnvironment;
 use WeBWorK::DB;
-use WeBWorK::Utils qw(runtime_use readFile cryptPassword);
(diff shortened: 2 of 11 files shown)


Alex-Jordan:
> Do we also want to trim passwords when they are entered to sign in?

I think yes. There is the scenario of an instructor sending a new password to a student over email, and then the student copy-pastes and accidentally grabs a space.

drgrice1:
Okay, I will add that.

pstaabp:
This looks to me.   I did the following
1. made a new student
2. changed the password to " password"

Logging with either "password" or " password" worked fine. 

Question: we talked about how non-WW authentication handles this.  Just looking at code changes, if you're scrubbing only `cryptPassword`, then it appears we leave the other authentication method alone.  Correct?  

I looked at this for LDAP, which is what we use on our campus.   I don't see that this new change will affect that, which I think it good.

drgrice1:
It looks like you are correct, except in the fail over case when checking the password.  Although I am not certain.  How does the password get set locally for that fail over case?

pstaabp:
Definitely, in the failsafe method.  That is, if say, you can't connect to the authentication server, it falls back on the webwork authentication and I think that's what we want. 

My fear was that if the password on the external authentication has a leading space, and we trim before sending then you can never authenticate, since WW has no control on that.  As I can tell, there is no trimming before going to an external authentication.  Is that correct?

drgrice1:
I think you are right, unless that was already the case due to what @mgage had in Authen.pm.

taniwallach (approved):
The code looks good.

I was able to change a user's password in both the "Classlist editor" and in "User Settings" to passwords with leading and trailing spaces, and then to log in when the number of leading/trailing spaces was different. (Spaces in the middle of the password remained unchanged). Thus, this seems to work exactly as intended.

pstaabp (approved):

PR #1291, opened 2021-03-27
Let TA's and above use SMA early

Alex-Jordan:
A small thing that has bothered me in the past.

Before these commits: if SMA is enabled (globally and for a problem) but the problem set has not yet opened, and if an instructor visits a problem set...

The SMA button is grayed out and there is a message something like  "you must attempt so many times before you may use this feature". This happens even when you have SMA set to a 0 attempt threshold.

This was because of bad logic. This PR makes a new permission `can_use_show_me_another_early` and sets it to `ta` by default. Then in the above scenario, the instructor get access to the button.

If you change the permission level to undef, and try the same experiment, the button is grayed out and its tooltip text says a message about the problem set not yet being open.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -763,7 +763,7 @@ $authen{xmlrpc_module}  = "WeBWorK::Authen::XMLRPC";
 	create_and_delete_courses      => "professor",
 	fix_course_databases           => "professor",
 	modify_tags                    => "admin",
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
This looks good to me.

drdrew42 (approved):

PR #1292, opened 2021-03-27
Improve the user interface of the PG problem editor.

drgrice1:
The main changes are improvements to the modal dialog when viewing a problem in the current window.  However, this also removes the usage of the legacy javascript tabber.js (just us bootstrap), and the ugly tooltips from wz_tooltips.js (again just use bootstrap).

There is also a bug fixed in Problem.pm.  When editing a problem and problem randomization is enabled the temporary file was saved as the source file for the user's actual problem.

This also cleans out some of the legacy javascript usage (wz_tooltips.js and tabber.js are gone).


drgrice1:
There are a few problems I need to resolve yet, but this works well for PG problems and set headers.  I just noticed something is wrong for the course header though.

drgrice1:
Okay.  Course info file editing is fixed.

Alex-Jordan:
`openwebwork` has a `pgproblem-editor-ui-improvements` branch now, along with a few others. I think maybe you pushed to the wrong remote at some point recently. More concerned that `drgrice1` (on GitHub) has the latest updates than anything else.

drgrice1:
I cleaned up the branches.  The pull request on GitHub is based on the branch in my fork, and has the latest updates.

What is happening is that I create a branch from the origin remote (the openwebwork repository), make my changes, and then push it to the remote for my fork.  However, git doesn't automatically set the upstream branch to the one in my fork.  Then I make more changes, and push again (forgetting to change the upstream branch), and then it ends up on origin.  However, I always check GitHub after pushing to see my changes there, and then remember to change the upstream branch and push again.  The last step in this situation is to remember to delete the accidental push to origin.

I have it figured out now so this won't happen again.  I need to add the `-u` flag when pushing, to set the upstream branch to the one being created.  Although, it seems to me that this should be the default behavior when creating a new remote branch.

pstaabp:
Looks much better.  One other possibility might be to hide the Show Past Answers and Email button at the bottom in this view.  They don't make much sense in this context.

pstaabp:
Also, looks like nothing else in the code uses wz_tooltips.js.  How about delete that from /js/legacy/vendor ?

I was going to say the same with tabber.js, but there's a few other places that uses it.  That could be another cleanup task.

drgrice1:
> One other possibility might be to hide the Show Past Answers and Email button at the bottom in this view. They don't make much sense in this context.

Yeah, I agree.  Although, that is a bit trickier to do with the current structure.  There are also several other things that shouldn't be there, like `Previous Problem`, `Problem List`, and `Next Problem` buttons and the `edit` link, etc.  I might add something I am also working on locally that changes to utilizing the webwork webservice for rendering a problem instead when using the modal.  Then you  don't get any of that.

> Also, looks like nothing else in the code uses wz_tooltips.js. How about delete that from /js/legacy/vendor ?

I will check into that.   If this is the last case of the wz_tooltips.js usage, I will delete it.  I think there are still other cases where the tabber.js file is still used, as you say though.  I will see how hard it would be to completely remove that also.

pstaabp:
> I will check into that. If this is the last case of the wz_tooltips.js usage, I will delete it. I think there are still other cases where the tabber.js file is still used, as you say though. I will see how hard it would be to completely remove that also.

It looks like the `AchievementEditor`, `AchievementList`, `ProblemSetDetail`, `ProblemSetList2` and `UserSetList2` uses tabber.  We'd have to switch over all of those to bootstrap.  I would say wait until upgrading bootstrap to version 4 or 5 do to these.

drgrice1:
Yeah, maybe.  Although, it might be easier to upgrade bootstrap if those were already switched over.  It would certainly make that step more uniform.

htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
```diff
@@ -17,46 +13,46 @@ window.addEventListener("DOMContentLoaded", function () {
...
-		$("#pg_editor_frame_id").contents().find('#toggle-sidebar')
-			.addClass('hidden');
+		if (busyIndicator) {
+			busyIndicator.remove();
+			busyIndicator = null;
+		}
+		var contents = $('#pg_editor_frame_id').contents();
+		if (contents[0].URL == "about:blank") return;
+		contents.find("head").append("<style>#site-navigation,#toggle-sidebar,#masthead,#breadcrumb-row,#footer{display:none;}</style>");
+		contents.find('#content').removeClass('span10');
```
pstaabp, line 35: to hide the other elements in the page, wouldn't:
```
contents.find("#problem-nav").addClass("hidden");
contents.find("#probem-footer").addClass("hidden");
```

easily hide those?

pstaabp:
I notice that there are also changes to ProblemSetList2.pm and AchievementEditor and AchievementList.  I can't figure out particularly what has changed.  Is it tooltips? or more substantial?

pstaabp:
Also, I don't know about graphTool.  I found it used by `parserGraphTool.pl`, but couldn't find anything the library that uses it.  Are my grepping skills lacking?

htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
drgrice1: @pstaabp:  Yes, but it is slower.

drgrice1:
> I notice that there are also changes to ProblemSetList2.pm and AchievementEditor and AchievementList. I can't figure out particularly what has changed. Is it tooltips? or more substantial?

Do you mean that behavior has changed, or that the code has changed?  I removed all tabber.js usage, so the code has changed.

> Also, I don't know about graphTool. I found it used by `parserGraphTool.pl`, but couldn't find anything the library that uses it. Are my grepping skills lacking?

The `parserGraphTool.pl` macro is the only thing that uses the `graphTool` javascript.  I explicitly wrote the javascript for that macro.

pstaabp:
> Do you mean that behavior has changed, or that the code has changed? I removed all tabber.js usage, so the code has changed.

I just trying to test everything thoroughly.  I realized this changed.  Are these code changes in response to removing tabber.js from he codebase?  If that's the case, we're just looking to test the tab functionality is working, correct?



> The `parserGraphTool.pl` macro is the only thing that uses the `graphTool` javascript. I explicitly wrote the javascript for that macro.

Is there something that I should test for this?  If so, can you point toward a problem that uses it?

drgrice1:
Yes, all of the code changes in the other files are for removing `tabber.js`.  Just test to make sure the tabs work as they should on those pages.  It was pretty straightforward to make the change.

The `parserGraphTool.pl` macro was added in https://github.com/openwebwork/webwork2/pull/1116 and https://github.com/openwebwork/pg/pull/484.  There are examples posted in https://github.com/openwebwork/pg/pull/484#issuecomment-699567751.

drgrice1:
There are more basic examples of the parserGraphTool.pl macro in https://github.com/openwebwork/pg/pull/484#issue-435143658.

pstaabp:
> There are more basic examples of the parserGraphTool.pl macro in [openwebwork/pg#484 (comment)](https://github.com/openwebwork/pg/pull/484#issue-435143658).

I guess I wasn't paying enough attention. super cool.

htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
pstaabp: You can mentioned trying to call the renderer in a different way.  Is that your thought to further clean this up?

pstaabp (approved):
This both simplifies the UI of the Problem Editor, but also cleans up the code in terms of removing the old `tabber.js` code from webwork.

Alex-Jordan:
@drgrice1 I just reviewed and it was all good. With @pstaab's review in as well, I would go ahead and merge this. Would that complicate anything for you with #1301?

This is a poll for anyone reading: change "edit" to an icon?

Alex-Jordan (approved):

drgrice1:
Nope.  That wouldn't be a problem.  I can resolve any merge conflicts.

PR #1294, opened 2021-03-28
Change the MathJax config to use its default color setup instead of colorv2

drgrice1:
The current setup was compatible with how MathJax version 2 behaved by default, and that is how webwork has used it.  However, this changes to the MathJax 3 default.  The difference is in how the TeX `\color` command behaves.  With the MathJax 3 default it is the same as how LaTeX works (that is `\color{blue}This is text` makes all of "This is text" blue.  With the MathJax 2 default (and colorv2 in MathJax 3) only the first "T" would be blue.  The point of this change is that with this setting both HTML output and hardcopy output will be the same.  See the discussion in #1215.

This is up for discussion as to if this is the right way to go.  If this is incorrect, and we want to maintain the previous behavior then that needs to be fixed.  I had put in `colorV2` instead of `colorv2`.  I believe at the time I wrote that code (almost a year ago), that was correct.

This also adds the `defer` attribute to the loading of the `math4-overrides.js` file (if it exists) so that it loads after `math4.js` and can actually do overrides.

### DIFF
--- a/htdocs/js/apps/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/apps/MathJaxConfig/mathjax-config.js
@@ -1,7 +1,6 @@
 if (!window.MathJax) {
 	window.MathJax = {
 		tex: {
-			autoload: { color: [], colorV2: ['color'] },
 			packages: {'[+]': ['noerrors']}
 		},
 		loader: { load: ['input/asciimath', '[tex]/noerrors'] },
--- a/htdocs/themes/math4/gateway.template
+++ b/htdocs/themes/math4/gateway.template
@@ -50,7 +50,7 @@
 <script type="text/javascript" src="<!--#url type="webwork" name="htdocs"-->/js/vendor/bootstrap/js/bootstrap.js"></script>
 <script type="text/javascript" src="<!--#url type="webwork" name="theme"-->/math4.js" defer></script>	
 <!--#if exists="math4-overrides.js"-->
-<script type="text/javascript" src="<!--#url type="webwork" name="theme"-->/math4-overrides.js"></script>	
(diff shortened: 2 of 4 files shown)


Alex-Jordan:
I searched the OPL and there are six instances of `\color`.

Four of those instances are well-written with a container scope to work for both MathJax v2 and LaTeX, like:
```
This only has one {\color{blue}{word}} that is colored.
```

The other two instances are written in a way that assumes the LaTeX (MJ3) behavior.

So only considering OPL problems, it seems that moving to the MJv3 default would be an improvement.

Note that any use of `\color` that assumes one version or the other can be edited to be like the above so that it works in both settings.

drdrew42:
I have had issues with `\color`, as one of our profs is a big fan, but doesn't always properly scope their usage. Check in `Contrib/CUNY` for plenty of examples.

Because usage is inconsistent among these problems, I will plan to batch-process them once the dust settles on our decision for implementation (this PR or #1299)

Alex-Jordan:
I think I'm in favor of this one instead of #1299.

pstaabp:
I agree with @Alex-Jordan on this.  Perhaps we can also provide a script for anyone using the old version to adapt to the more-standard scoping version of `\color`.

Alex-Jordan:
I see three in favor of this one, no expression in favor of #1299. Worst case is some problems become a lot more colorful, and could quickly be fixed individually. As far as the OPL instances go, this one (#1294) only makes things better.

@pstaabp One thing about a script is how would you tell what the intent was? In the abstract, `\color{blue}x-axis` might have been intended to go either way. In which case you can't automate deciding to make it `{\color{blue}x}-axis` or something else. I suppose if you were sure all of the problems you are applying it to only used the MJv2 convention.

You could always change the MJ config file too. Let's be sure to document this in release notes.

In the interest of progress I am going to merge this one and close the other one.

PR #1295, opened 2021-03-28
tweak message for when a roster file is uploaded as csv instead of lst

Alex-Jordan:
This is just a tiny improvement in messaging. At least where I am, it is common to upload a roster file as .csv. If you use Excel to create it, that's what you get, not .lst. You need to change the extension, either before or after upload. If you upload it as csv, the warning message is not helpful. And more importantly, some newbie faculty don't realize there is a need to make it lst.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -1174,7 +1174,10 @@ sub checkFileLocation {
 	return if $dir =~ m/^$location$/;
 	$location =~ s!/\.\*!!;
 	return if $dir =~ m/^$location$/;
-	$self->addbadmessage($r->maketext("Files with extension '.[_1]' usually belong in '[_2]'",$extension,$location));
+	$self->addbadmessage(
+		$r->maketext("Files with extension '.[_1]' usually belong in '[_2]'",$extension,$location)
+		. (($extension eq 'csv') ? $r->maketext(". If this is a class roster, rename it to have extension '.lst'") : '')
+	);
 }
 
 ##################################################


drgrice1:
This is fine.  However, I think it would be better to just make it so that webwork can work with either the `.lst` extension or the `.csv` extension.

Alex-Jordan:
I don't disagree, but I had a thought about that. If you have a .csv in your templates folder that is *not* a properly formatted roster, and it is an option to import, and someone imports it, it can create a mess. It would be nice if the only files presented to you for import could be vetted beforehand as properly formatted roster files. The current system forces people to apply this nonstandard 'lst' extension, which does an imperfect job of preventing people from importing things that cause messes.

taniwallach:
I think we could try to improve the import class list functionality.

I looked at https://webwork.maa.org/wiki/Classlist_Files and it seems the standard for "data" lines is more or less as follows (ignoring fields which have no specific conditions applied to them by default).
  * 11 fields, of which the last 2 may be totally omitted.
  * Field 1 is `student_id` and **must be unique or blank**.
    * What happens if that `student_id` is already in the database?
    * I don't see any enforcement about it being unique in the file or in the database.
  * Field 4 is `status` and has a fixed set of allowed values.
    * The standard settings are in the Wiki page, but they can be modified in `global.conf`.
    * We could try to enforce valid values, and use a configurable fallback value for invalid values.
  * Field 8 is `user_id` and **must be unique** and can contain only letters, digits, hyphen (-), period (.), and underscore (_). 
    * **Cannot be left blank.**
    * I do not see any enforcement of it being unique. I think a repeated records would trigger an error, as I believe that a second call to `$db->addUser($User);` in `importUsersFromCSV()` for the same "user" would fail.
    * It would seem that "extra" lines for the same `user_id` could be rejected in `parse_classlist()`.
    * Behavior when `student_id` is already in the database is based on the setting of the pull-down menu.
  * Field 11 is `permission` and is either blank or an integer (?). When blank the permission will be set to 0 (i.e. student).
    * The standard settings are in the Wiki page, but they can be modified in `global.conf`.
  * Field 8 is `email_address` and can be left blank.
    * There seems to be no enforcement that it "looks like" a valid email address, but if I recall correctly, the DB code has some restrictions on allowed characters.

I would think that we can parse a `.csv` file or `.lst` file into a temporary structure, validate it seems OK, and either run all the DB changes at once or reject the file if not.

  * The import code is in `importUsersFromCSV()` in `lib/WeBWorK/ContentGenerator/Instructor/UserList2.pm`.
    * It is marked as:
    * ` FIXME REFACTOR this belongs in a utility class so that addcourse can use it!`
    * `(we need a whole suite of higher-level import/export functions somewhere)`
  * That depends on `parse_classlist()` from `lib/WeBWorK/File/Classlist.pm` which does check that there are at least `$MIN_FIELDS` (currently 9) fields (allows permission to be missing) and at most `$MAX_FIELDS` (currently 11) which apparently allows 
  * It would seem that better validation could be done in `parse_classlist()` and "bad" rows can be reported.

PR #1297, opened 2021-03-29
log LTI authentication to course's login.log

Alex-Jordan:
When a user comes in through an LMS, if it is a new WW user, there is a record in the course's login.log. But if they authenticate and enter WW, there is currently no record in the login.log. Instructors cannot verify a student was in the course when they report that they were. This commit changes that.

All I really wanted was for LTIAdvanced, to log a "LOGIN OK" message. To mirror what happens with basic password logins, I put a message "AUTH LTI: user authenticated" message as a counterpart to "AUTH WWDB: password accepted". I really don't know what the "AUTH WWDB" is supposed to convey, so I really don't know if "AUTH LTI" is a good counterpart.

And then there are all these points where authentication could fail, so I felt I should put log messages there too. I'm not confident I made these messages say the right thing. And each one is preceded by `$self->{log_error} .= ...`, so I suspect it is redundant to add what I added. However I am not sure where the messages in `$self->{log_error}` end up...

And then even though I only cared about LTIAdvanced, it seemed appropriate to do the same things in LTIBasic.

I've only tested this for a student user that enters through the LMS, using LTIAdvanced, and successfully enters. Testing the fail points seems unnecessary, at least compared with the effort it would take to engineer each fail.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -340,10 +340,12 @@ sub verify_normal_user {
   if ($auth_result eq "1") {
       $self->{session_key} = $self->create_session($user_id);
       debug("session_key=|" . $self->{session_key} . "|.");
+      $self->write_log_entry("LOGIN OK");
       return 1;
     } else {
       $self->{error} = $auth_result;
       $self-> {log_error} .= "$user_id - authentication failed: ". $self->{error};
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I'm not sure what conditions led to me putting this commit on my GH's WeBWorK-2.16 branch. I'll move it so there is no confusion and in case I accidentally update that branch to something else.

PR #1298, opened 2021-03-29
LTIAdvanced: assign all Achievements to a new user

Alex-Jordan:
This is supposed to make it so that when a new WW account is created for a user coming in through LTI, they will get all the Achievements assigned to them automatically (similarly to what happens with "visible" problem sets).

Updated:

~~There are 2 or 3 things wrong with it:~~

1. ~~This sort of works. A new user will show up in the instructor's Achievement Editor as having all Achievements assigned to them. And when the new user answers their first question, from that point onward all seems fine. But before they answer their first question, if they go to Achievements, there is a message that they have no global Achievements data. Somehow that needs to be initialized. (A user who gets Achievements assigned to them in the usual manual way does _not_ see a message like this. They see grayed out badges.)~~
2. I only did this in LTIAdvanced. If it works, I suppose it should be done in LTIBasic too?
3. ~~It assigns all Achievements, but perhaps it should only assign all _enabled_ Achievements. That would sort of be parallel to how the only problem sets that get assigned are all "visible" problem sets.~~

I'm content with this for my own server as it is. But if anyone has input on 1, 2, and/or 3 I would complete this and it could become official.


taniwallach:
My 2 cents, but as I have not used achievements in production - take them as an opinion.

  * Point 3: I think you should only assign achievements other students can earn, so apparently only enabled ones.
  * Point 1: It would be good to initialize to avoid the strange temporary situation. Maybe where the error message is triggered the initialization could be done? 
  * Point 2: When it works fully, it makes sense to add to LTIBasic.

I would say that if point 3 is successfully addressed this can be added to 2.16.
@drgrice1 - What do you think?

taniwallach:
@Alex-Jordan : We were trying to close up loose ends last night so we can close up WWW 2.16. The code on this looks fine, and I would like to test this. Could you try to rebase the PR (or if easier/necessary replace it with a new one)? When doing that, maybe merge the two commits into one, and do a force push. Thanks!

Alex-Jordan:
The conflict was just a place adjacent to the new code here, where a trailing space character was removed in some other branch that has been merged. I fixed the conflict here in GitHub because it was small enough. Thanks for testing!

taniwallach (approved):
Tested. Works as intended. I just added a commit to fix 3 lines of the new code which were indented with spaces to use tabs (as the rest of the new code was).

I would recommend "squash merge" as there is no need to keep the 4 small commits.

drgrice1 (approved):
I think this is good.  I haven't really tested it, but I assume that you and @taniwallach did.

PR #1299, opened 2021-03-29
Mathjax configuration color fix (alternate to #1294)

drgrice1:
This fixes the current setup for compatability with MathJax version 2.

Either this pull request of #1294 needs to be chosen and merged.  Currently the `\color` TeX command does not work.

This also adds the defer attribute to the loading of the math4-overrides.js file (if it exists) so that it loads after math4.js and can actually do overrides (same as #1294).


Alex-Jordan:
Closing in favor of #1294.

PR #1301, opened 2021-03-30
Improve renderViaXMLRPC (webwork2/html2xml) and use it in ProblemSetDetail2.pm and SetMaker.pm

drgrice1:
The formatRenderedProblem method used by renderViaXMLRPC has been improved to make it do many of the things that Problem.pm and GatewayQuiz.pm do when rendering problems.  The renderProblem method (in RenderProblem.pm) that is also in the chain of that rendering approach has been modified a bit to allow the use of MathQuill answer boxes if enabled for the course.

In ProblemSetDetail2 and in SetMaker, this now loads problems asynchronously via javascript through renderViaXMLRPC.pm instead of loading all problems on the initial page load.  Each problem is loaded successively instead of all at one time, to avoid bogging down the browser.  The point of this change is that many problems currently do not work on the problem set detail page or in the library browser.  This puts each problem into its own iframe and solves those issues.  Most likely all problems will now work.

Previously when problems were loaded via javascript after the page loads the instructorXMLHandler was used.  Now renderViaXMLRPC is used for this as well.

The simple_format.pl needed some minor changes to make this work well.

The json_format.pl has been cleaned up and improved also.  In the case that this format is used the header is set with the appropriate content type.  Related to this is a special case added for the library browser (SetMaker.pm).  If the option `send_pg_flags` is set then the return will also be in json format, but will only contain two keys.  Those are
`html` and `pg_flags`.  The `html` key contains the full html of the format, and the `pg_flags` the `flags` of the pg problem.  This allows the library browser to obtain the insructor comments.  This could be useful in other scenarios as well.

The file lib/WeBWorK/Utils/DetermineProblemLangAndDirection.pm has been renamed to lib/WeBWorK/Utils/LanguageAndDirection.pm and the methods in that file improved to simplify their usage throughout the code.

This also removes the local iframe-resizer javascript and installs a newer version (4.3.1) via npm.

The legacy javascript file htdocs/js/legacy/java_init.js has been deleted and its usage removed from the code.

I also added some improvements to the appearance of the problem set detail 2 page, and eliminated the addition of css style classes via javascript.

This is targeted for the develop branch for now, but it would be nice to get it into the release.


Alex-Jordan:
Testing this out. So far it works, and I just have cosmetic/layout/design observations in the Library Browser. At least so far, I'm not looking at the code, so I may be off base and citing things that aren't part of this, but I'm seeing them stand out now.

* In the Library Browser, if I reveal the file path of a problem and the file path is even medium length, it butts into the usage statistics and expands the width of the entire block. It's jarring. (a) Can we think of a reason to not simply always show file paths? (b) Could the file path be positioned on the line under the usage statistics, randomize button, etc?
* Also in the Library Browser, I find it jarring when I click to add a problem to a set, and the "(in target set)" text gets inserted between the usage statistics and the icons. Would it be good right-aligned under the icons? (In combination with the previous item, now two things would be on the second line. But less likely to butt into each other.)
* I'm finding it hard to intuit whether a given "Add" button applies to the problem above or below that Add button. It's easy when you are at the top of the list or the problem has been grouped with similar problems so it gets the black border around it and the Add button. But otherwise, viewing the middle of the list, it's unclear. It gets even more muddled if I click something that brings in the results table. I propose styling things in some way that makes it clear which exercise the Add button is tied to. That may mean thinking about the next item:
* I propose changing the styling for the grouped problems. Instead of a border (which might be useful to address the Add button issue) maybe something that indicates an accordion or carousel. Instead of the M button, there could be something that lets you cycle through a carousel.

drgrice1:
Yeah, so those are all things that need improvement, but those are also all things that were that way before.

drgrice1:
You must not have looked at the set detail page yet. That has many of these sort of issues as well. I wanted to do a lot more to improve both of these pages, but stopped short before I went to far with this pull request. Perhaps I will try to do that in another pull request yet.

Alex-Jordan:
What is new is the Preview/Check/ShowCorrectAnswers buttons. Their presence and visual similarity to the Add button makes it less clear that the Add button which is immediately below them is actually paired with the following exercise, not the same exercise those three buttons go with. And then if you use them, the results table appears, also adding a new element of sprawl.

drgrice1:
I don't think there is quite as much difficulty in determining which problem the add button goes with as you say.  The problem is contained in the gray box, and when the page first loads the add button is at the top of the first problem.  Really, none of those things you are mentioning are changed by this pull request.

In any case, one way to fix this is just to put the border lines around all problems.  However, doing what you suggest for the problems with more is to progressive at this point, and would take a lot more work and is beyond what needs to be done.

Another option is to remove the problem submit buttons (and just make them inert viewable problems instead of active workable).  I had debated on creating a new "even simpler" format that would be used here that does that (doing so with the simple format could be achieved, but is not a very efficient way to handle it).

drgrice1:
Another simple solution that would work for now (better could be done later) would be to just put a horizontal rule between the problems.

Alex-Jordan:
The other new thing is that before, the width of a problem is 100%, minus the side navigation panel, but with this branch, something about the iframe and its surrounding div is coming out at a restricted width. And this is what makes the first two items in my list "new". With a wider problem, the butting up doesn't happen except very rarely.

> I don't think there is quite as much difficulty in determining which problem the add button goes with as you say. 

So I really was mixed up which was which, after scrolling down halfway through a page of with 40 OPL problems. I'm not just hypothesizing; it's empirical UX data. I'm sure it's all about the "Add" button for problem 23 falling right under the Preview etc buttons for problem 22. The Preview etc buttons clearly go with 22 because of the background. It just feels more natural (to me at minimum) that the "Add" goes with those buttons. And only my experience that the "Add" goes above a problem gave my brain a counterargument, which a new instructor user doesn't have. Yes, I could go back to the top to reorient myself, and commit to memory what the scheme is, and then try to scroll back down and find the problem I was at before. But why would I do that if I'm in a state of mind where the "Add" button is for 22?

I think all I am saying is that the user experience elements are in a place where we should do something (and I'd be happy to work on this branch too) before merging. I'm not sure how I feel about merging something, even into develop, when we know it needs follow-up work. Even more important if it goes into 2.16, which you mentioned in the opening. I'd like this in 2.16 as well, since I'd like to use more javascript-reliant problems with simple_format.pl.

If you don't beat me to it, the next chunk of time I do things here, I'll put into trying things on this branch. That would be Tuesday or Thursday at the earliest.

drgrice1:
The width restriction is not new either.  Again, everything you are talking about is the way it was before.

> I'm not just hypothesizing; it's empirical UX data.

Say what now?

drgrice1:
Everything in webwork needs follow up work.  We are never going to have any progress if we try to do everything in one pull request.  It would be better to merge, and then improve further.

drgrice1:
The point is that none of the issues you are discussing are caused by this pull request.  All of them existed before this pull request.

It is also quite debatable that any of those issues are made worse by this pull request as you claim.  Your empirical UX data set is limited to one data value (you).   That outcome is quite different than my empirical data set with my one data value (me).  Putting that data set together we have two data values with completely different outcomes.  Statistically speaking we can make no conclusion.

Furthermore, the changes that you are requesting are difficult to obtain.  This is already an extensive pull request, and adding these things would make it more so.

The changes in this pull request are more important than the minor visual issues that you are observing, and those can be dealt with later.

drgrice1:
@Alex-Jordan:  I think that my recent commits have addressed your concerns.  I think that your suggested changes to the "More/Less" implementation would be nice, but are more extensive and can be done later.

Alex-Jordan:
> The point is that none of the issues you are discussing are caused by this pull request. All of them existed before this pull request.

This is not the case. Probably words can't explain it, and I will make a video to show you the differences. And they are things that make using WeBWorK worse, so they should not be merged yet.

drgrice1:
This is the case.  You can make a video, and it still won't change that fact.  I see what you are trying to say are differences.  They aren't.  The primary difference is your perception that when buttons have the same style and are anywhere near each other, then they are grouped together.  Even if there are clear indicators that they are not grouped together, like a gray container around the ones above, and the "Add" button not in that below.  My latest commits to this make it painstakingly clear that those buttons are not grouped.

Alex-Jordan:
OK, thanks for the edits. I'll take a look again when I return to this.

>  The primary difference is your perception that when buttons have the same style and are anywhere near each other, then they are grouped together.

In general, I take my perception of things like this as an indicator of how people in general will perceive it. Don't we all? That's what I meant by a user experience data point.

Also it seems we are using words differently. I would say it _is_ a change when the layout changes. But I think you are saying that it's not a change, because the edits here didn't touch on layout or styling code in that way. They made other things better, and old layout and styling decisions are the problem.

drgrice1:
> In general, I take my perception of things like this as an indicator of how people in general will perceive it. Don't we all? That's what I meant by a user experience data point.

You are assuming that your perceptions are the same as (or similar to) other people's perceptions.  That is usually not a valid assumption.  To obtain valid user experience data, you need many more data values (poll the public).

> Also it seems we are using words differently. I would say it _is_ a change when the layout changes. But I think you are saying that it's not a change, because the edits here didn't touch on layout or styling code in that way. They made other things better, and old layout and styling decisions are the problem.

I knew what you meant.  I think I responded poorly, but I think that your assumption that I didn't know what you meant irritated me.

lib/FormatRenderedProblem.pm
```diff
@@ -24,54 +23,12 @@ FormatRenderedProblem.pm
...
-};
-	if ($@ or not ref($seed_ce)){
-		warn "Unable to find environment for WebworkClient: 
-			 webwork_dir => $WeBWorK::Constants::WEBWORK_DIRECTORY 
-			 pg_dir      => $WeBWorK::Constants::PG_DIRECTORY";
-	}
-
-
-
-our %imagesModeOptions = %{$seed_ce->{pg}->{displayModeOptions}->{images}};
```
taniwallach, line 60: Did you check that removing the `our $imgGen` block and `use WeBWorK::PG::ImageGenerator;` above does not break use of images display mode for XMLRPC rendering?

lib/FormatRenderedProblem.pm
```diff
@@ -80,469 +37,408 @@ sub new {
...
+	# since sha1_base64 doesn't pad we have to do so manually
+	while (length($bodyhash) % 4) {
+		$bodyhash .= '=';
 	}
 
-	my $localStorageMessages = CGI::start_div({id=>'local-storage-messages'});
-	$localStorageMessages.= CGI::p('Your overall score for this problem is'.'&nbsp;'.CGI::span({id=>'problem-overall-score'},''));
-	$localStorageMessages .= CGI::end_div();
-		
-	# my $pretty_print_self  = pretty_print($self);
```
taniwallach, line 432: I think this was here to help with debugging, and is the reason the file has a `pretty_print` subroutine.

lib/WeBWorK/ContentGenerator/renderViaXMLRPC.pm
```diff
@@ -187,6 +187,8 @@ sub pre_header_initialize {
 	$xmlrpc_client->{sourceFilePath}  = $inputs_ref{sourceFilePath};
 	$xmlrpc_client->{inputs_ref} = \%inputs_ref;  # contains form data
 	# print STDERR WebworkClient::pretty_print($r->{paramcache});
```
taniwallach, line 189: Another debugging tool.

lib/WebworkClient/simple_format.pl
```diff
@@ -73,15 +73,9 @@
 	</form>
 </div>
 </div></div>
-
-<div id="footer">
-WeBWorK &copy; 1996-2019 | host: $SITE_URL | course: $courseID | format: simple | theme: math4
```
taniwallach, line 78: Copyright range can be fixed here and in other format files.

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -352,144 +294,116 @@ sub renderProblem {
...
+		print STDERR "RenderProblem.pm: problem source is included in the request \n" if defined($rh->{source});
 	}
-    # warn "problem Record is $problemRecord";
 	# now we're sure we have valid UserSet and UserProblem objects
-	# yay!
 
-##################################################
-# Other initializations
-##################################################
-	#debug( "envir->displayMode", WebworkWebservice::pretty_print_rh($rh->{envir}));
```
taniwallach, line 405: debug code.

taniwallach (commented):
This is just some comments from a look through the Perl code.
I'm not comfortable enough with JavaScript to quickly understand and review the changes to the JS code.
I did not look at the SetMaker changes that much yet.
Overall the code I looked at seems to cleans things up and/or improves various code, and  seems to be consistent. Testing the functionality is certainly necessary.
Dropping the Java applet support, and moving the iFrame resizer to be a npm installed package is certainly a good change.
I would tend to leave the commented out debug code alone, as it clearly can help during development work to be able to record and view the relevant data structures. I made several remarks about this where I noticed such commented out code being removed.

I will try to get to do some functional testing in the next few days.

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -500,29 +414,28 @@ sub renderProblem {
...
-	###################
-	
-	$out2->{flags}->{PROBLEM_GRADER_TO_USE} = undef;
+	# END XML DEBUGGING CODE
+
+	$out2->{flags}{PROBLEM_GRADER_TO_USE} = undef;
+
 	my $endTime = new Benchmark;
 	$out2->{compute_time} = logTimingInfo($beginTime, $endTime);
-	# warn "flags are" , WebworkWebservice::pretty_print_rh($pg->{flags});
```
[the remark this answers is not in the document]
drgrice1: This is the wrong way to insert debugging code.  Code like this can always be inserted as needed if you are wanting to debug something at a particular instance.  It should not be left in the code base.  Doing so and expecting this commented out code to be un-commented when debugging is desired can result in this un-commented code being left in for a future pull request, and is generally bad coding practice.

If you want debugging that is left in the code base, the correct way is to use a debugging flag that can be changed to make the debugging code active.

This applies to all of the comments that you left in this regard.

lib/FormatRenderedProblem.pm
```diff
@@ -2,14 +2,13 @@
 
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
-# $CVSHeader: webwork2/lib/WebworkClient.pm,v 1.1 2010/06/08 11:46:38 gage Exp $
-# 
+# Copyright &copy; 2000-2021 The WeBWorK Project, http://openwebwork.sf.net/
```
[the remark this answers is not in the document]
drgrice1: Good point.  I will add that.

lib/FormatRenderedProblem.pm
drgrice1: The image generator is created directly by WeBWorK::Utils::AttemptsTable now, instead of passing one in here.  Now a more valid course environment option is passed in instead.  Only one of these is needed by the WeBWorK::Utils::AttemptsTable.  Also, I have tested it, and it works.

lib/FormatRenderedProblem.pm
drgrice1: This did not actually output anything.  It just generated a string containing the pretty_print result that was never used.  That is why this and the pretty_print subroutine was deleted from this file.  We have enough pretty_print subroutines defined throughout the code.  We don't need one for each file.  And again, this is bad programming practice.  Inject the debugging code you need as you need it, and remove it before committing.

drgrice1: Further note that if one wants to obtain a `pretty_print` of `$self` as is done here one could use `my $pretty_print_self = WebworkClient::pretty_print($self)` in this file and do so.  That works and gives the same output.  The `pretty_print` subroutine defined in `WebworkClient` is identical to the one defined in this file (other that an improvement to prevent certain secure information from being displayed -- lines 756 - 762 in that file).

Alex-Jordan:
I haven't moved on past the Library Browser. Thanks for the things you did. In the future when it's small like that, I think it will be good practice for me to just play around, and if it I hit something good, I could just open a PR to the branch. It's reflexive to just report what I see as an issue, but not worth other people's time if I can address it.

For this next thing, it's some javascript thing that I don't know how to address. In my production server (2.16 from a few weeks back) if I go to Library Browser, OPL, Calculus - Single Variable, Differentiation, Quotient Rule without trig, with Shown: 20, then clicking View Problems, it takes about 3 seconds and all problems are loaded, with all math rendered. Good, or at least acceptable.

If I try the same thing on this branch, it first takes about 20 seconds for all of the iframes to at least have something in them. Then (since I have MathQuill on) about 4 more seconds for MathQuill to do its thing to all of the answer blanks. Then about 4 more seconds for all of the math to render. Throughout all that, the iframeresizer is reacting slowly as well. And I have 5 console warnings about iframeresizer not responding within 20 seconds. Once it's all settled, there is a small amount of vertical scroll in each problem within its iframe.

Do you have similar behavior? It may be that my development server is resource-poor and these things are amplified.

drgrice1:
I don't mind the request for changes.  It was just something in the wording of your second comment that ruffled my feathers.  I need to do what my old chair always said she does and wait several minutes and look over my responses before I hit send if I feel that I am rattled.

In any case, the behavior that you are describing unfortunately comes with this change.  It is not your server that is slow, it is your computer and the browser that is slow.  This doesn't increase the load on the server (not much at least).  However, if you are loading 20 problems, this is the equivalent of opening 20 tabs/windows at once in your browser.  That is quite the load for your browser and computer to handle.  Each page has its own instance of MathJax and all of the other javascript that must load, and the page render, and then finally the iframe resizer kicks in.  I have it set up so that each iframe loads subsequently (async/await javascript), and that means that the first several iframes should load rather quickly.  On my computer (which is running the test server as well as the browser) the first several problems in the page are completely loaded in a fraction of a second.  The rest continue to load below, but you can start looking at those problems in the mean time.  My computer is more than ten years old so that isn't to bad.  With your choice of problems and 20 problems shown it took about 23 seconds for the entire page to render.

I don't see the small amount of scroll usually.  Occasionally I see the warnings from the iframe resizer about the child page not responding, but not usually.  One thing to note is that the page loads faster if the browser dev tools pane is not open.

This gives an idea of the things that need to be worked out if we switch to using the standalone renderer concept.

I will see if there is anything more that I can do to speed things up and alleviate stress on the client end.

drgrice1:
By the way, do you have the network cache disable in your browser dev tools?  If you do and you have the dev tools open, it seriously slows down the page loading.

drgrice1:
Also, do you have the latest changes in this pull request?  I changed some things that did improve the loading speed some (commits 34018ac43a and 4a04d9c9bf).

drgrice1:
Another thing to try is to load several problems that use geogebra or the graphtool macro with this pull request.  Compare to what happens before this pull request.

Alex-Jordan:
> By the way, do you have the network cache disable in your browser dev tools?

I doubt it. Not if that's something I would have actively set.

> Also, do you have the latest changes in this pull request?

Yes, I pulled it fresh before testing this afternoon.

> Another thing to try is to load several problems that use geogebra or the graphtool macro with this pull request. Compare to what happens before this pull request.

No doubt that is better. It's just broken as things are now.

drgrice1:
> > By the way, do you have the network cache disable in your browser dev tools?
> 
> I doubt it. Not if that's something I would have actively set.

It is something you have to set.  I was just checking.  The list of problems in your example takes about 50 seconds to render on my computer if the dev tools are open and network caching is disabled.

> > Another thing to try is to load several problems that use geogebra or the graphtool macro with this pull request. Compare to what happens before this pull request.
> 
> No doubt that is better. It's just broken as things are now.

Yeah, I just wanted to remind you that there is a point to this pull request.

I am trying something that should cut the initial rendering time in half.  That is not rendering all the hidden problems when the page loads, and instead rendering them when the `M` button is clicked (or maybe better in the background after the visible problems load).  That is a large part of what is slowing the loading down.  The problems from the OPL you list in your example is actually a list of 47 problems.  All of those are currently loading when the page opens.

Alex-Jordan:
This might be a dumb question, but is it possible to condition the use of an iframe on whether or not the problem loads its own javascript?

drgrice1:
> This might be a dumb question, but is it possible to condition the use of an iframe on whether or not the problem loads its own javascript?

That isn't really feasible.  There is no reliable way to determine if a problem uses its own javascript.  Any problem could load an external javascript via the TEXT method, or can even inject javascript directly into the problem without loading an external script.  Furthermore, there are other issues besides javascript.  If you did that the appearance of the problems would be inconsistent.  You would also be adding to the work the server has to do because it would have to preprocess all of the problems again the way it did before.

Have you tried this pull request with the latest improvements?  The loading time is considerably less now.  The first few problems are ready to work almost immediately, and the rest load quite quickly while you are looking at those.  Furthermore, with the last changes, this actually reduces the load on the server from the current setup in many cases.  For example with the problems you mentioned (Calculus single variable - differentiation - quotient rule without trig - show 20), there are 47 problems loaded.  With the current setup, all 47 of those problems are processed by the server.  With this pull request only 20 of them are (although more will be processed on demand).

I did some timing.  I loaded your example (Calculus single variable - differentiation - quotient rule without trig - show 20) on the current WeBWorK-2.16 branch and compared that to this pull request.  The timing was from when the "View Problems" button was pressed (via a submit handler) to the time the the window load event was called (for the WeBWorK-2.16 branch) or to the time the last iframe's load event was called (for this pull request).  I used localStorage to save the start time so the javascript could compute the duration at the end.  On my computer it took on average about 6.5 seconds on the WeBWorK-2.16 branch, and about 16 seconds with this pull request.  With the pull request, I also timed each problem from the submit button press until it was viewable.  The first problem was viewable in less than 2 seconds every time.  The first 8 were viewable in the same 6 - 7 second time frame as the WeBWorK-2.16 branch.

So there is a big difference here that you are failing to observe.  On the WeBWorK-2.16 branch you see nothing until the end of those 6 to 7 seconds.  While with this pull request you have some problems ready to go in less than that, and 8 problems in the same time as the current setup.  Those first problems are fully set up and even workable in that 6 to 7 seconds.

This test was performed with my more than 10 year old computer running both the browser and the server.  This should be considerably less with a more modern computer that is not also running the server.

So if you are still thinking that this is that slow, then I have to ask ... how slow is your computer?

Alex-Jordan:
It is much better with the new handling of the M problems. I think it's not just the timing; there was another effect from the perceived stalling while "nothing" happened (while the hidden problems processed).  And no I hadn't had a chance to test. I just posed a question I thought worth asking. Thank you for the informed answer.

I find that if I set `scrolling: false` on line 411 of `setmaker.js`, I no longer have the small vertical scrolling issue. I can't think of bad side effects from making that change, but maybe you can.

Alex-Jordan:
In PreTeXt, we have pre-rendered all questions in use as static versions. The page loads quickly with all of the questions. Each question has a button to "Make interactive". Clicking the "Make interactive" calls on the `raw` format, and the div with the question gets replaced with a new div that has the question content, check/show answer buttons, randomize button, and reset button. Right now this does not work for a problem that loads some javascript.  But I think that I can look into the `raw` content and see script tags, and use that conditionally to replace the div with an iframe instead, maybe one that calls the simple format or some new format tailored for PreTeXt. That's where my head was. But not needing to load all problems at once, plus having access to pre-rendered static versions of questions makes the PreTeXt situation easier.

drgrice1:
I think that the `scrolling` option could be removed from `setmaker.js`.  In fact I will do so.

I added that when I was working with this in `problemdetail2.js`.  The issue on the problem detail 2 page is that the problems must fit into a narrower region.  As a result some problems end up having a portion clipped to the right and there is no way to see what is there (the only problems that I saw this with were some of @drdrew42's Geogebra problems).  Enabling the scrolling option makes the horizontal scrollbar come up so you can get to the right.

In set maker the problems are allotted more horizontal width, so this shouldn't be a problem.

What I really wanted was to only enable horizontal scrolling and not vertical to cover these situations, because I saw the same issue with the vertical scrollbar that you see.  This seems to be a limitation of iframes in that there is no option to separately enable/disable the horizontal and vertical scroll.  Another thing to note is that the `scrolling` attribute of iframes used by the iframeresizer javascript is deprecated.

Okay, this comment has been sitting here waiting for me to submit it for a while as I work on this.  I will leave my discussion above, but I found a better way to handle scrolling.  I will commit it soon with some other tweaks.

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -327,12 +269,12 @@ sub renderProblem {
 		$setRecord->answer_date(time()+60*60*24*7*3); # in three weeks
 		$setRecord->psvn($rh->{envir}->{psvn}||0);
 	}
-	#warn "set Record is $setRecord";
```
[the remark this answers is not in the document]
drgrice1: Really?  Do you think that the warn output of `set Record is WeBWorK::DB::Record::UserSet=HASH(0x55d0142a8178)` is that useful ever?

lib/WeBWorK/ContentGenerator/renderViaXMLRPC.pm
drgrice1: This is not a useful debugging tool.  At least not in the form it is in here.  This prints html formatted output to STDERR which is not very readable.  If the `WebworkWebservice::print_print_rh` method were used instead it would at least be readable.  Although in general this is one of those warn statements that is out of place.

Furthermore, why are you even commenting on this here?  I didn't remove this line in this pull request.

lib/FormatRenderedProblem.pm
```diff
@@ -80,469 +37,408 @@ sub new {
...
-	if ( defined ($rh_result->{WARNINGS}) and $rh_result->{WARNINGS} ){
-		$warnings = "<div style=\"background-color:pink\">
-		             <p >WARNINGS</p><p>".decode_utf8_base64($rh_result->{WARNINGS})."</p></div>";
+	if ($ce->{pg}{specialPGEnvironmentVars}{entryAssist} eq 'MathQuill') {
+		$problemHeadText .= qq{<link href="$ce->{webworkURLs}{htdocs}/js/apps/MathQuill/mathquill.css" rel="stylesheet" />} .
+			qq{<link href="$ce->{webworkURLs}{htdocs}/js/apps/MathQuill/mqeditor.css" rel="stylesheet" />} .
+			qq{<script src="$ce->{webworkURLs}{htdocs}/js/apps/MathQuill/mathquill.min.js" defer></script>} .
+			qq{<script src="$ce->{webworkURLs}{htdocs}/js/apps/MathQuill/mqeditor.js" defer></script>};
 	}
-	#warn "keys: ", join(" | ", sort keys %{$rh_result });
```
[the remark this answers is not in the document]
drgrice1: Why is this one useful on a regular basis?  And most notably for this hash, why here?  The contents of the `$rh_result` hash reference might be desired at various points in the html2xml pipeline, but it is best if those are injected as needed when working on things, and at the location that is pertinent for what is being worked on.  This one is not a good one to keep in my opinion.

lib/WebworkWebservice/RenderProblem.pm
drgrice1: The entire `$rh` hash is already output via `debug` at the beginning of this method.  There is no need to output the contents of this particular key again at this point.

pstaabp:
IMO, this improves things quite a bit.  Often in the library browser would timeout.  I don't see any of those.  I'm running my development machine (reasonably new laptop) and things are super speedy.  

One thing that has been bugging me about ProblemSetDetail2 is that the Display mode at the top of the problem list: 

image

always has "Images" as the default even though all of the other settings that I have have MathJax as the default.  @drgrice1 is this hardcoded or picking up some weird setting that I have?

drgrice1:
@pstaabp:  You can change this by changing the order of the values in `$pg{displayModes}` in the `localOverrides.conf` file (or just comment that variable out as the defaults now put "MathJax" first).

pstaabp:
I have 
```
$pg{options}{displayMode}        = "MathJax";
```
in `localOverrides.conf` and  everywhere else the default is MathJax.

Alex-Jordan:
You can do like this:
https://github.com/openwebwork/webwork2/blob/WeBWorK-2.16/conf/defaults.config#L950

which puts MathJax first. That's the default now.

Yes, it would be nice if `$pg{options}{displayMode}        = "MathJax";` alone was enough to make that be the option that is used here. I'm not sure if it is hard to do, but the reordering of the array ref entries is what Glenn was referring to.

drgrice1:
I agree that it would be nice for the currently set display mode to be active by default on the set detail page and on the library browser page.  It wouldn't be hard to implement.

Alex-Jordan (approved):
I did not scrutinize the code for this, but I just tested again and it is working well. I have no basis to weigh in on the commented out debug code discussion. Aside from resolving that, this seems ready for merging.

pstaabp:
@drgrice1 I think as we talked yesterday that we should get this into ww2.16.  Can you switch the branch to `WeBWorK-2.16`.  We should test against that instead of `develop`.

drgrice1:
@pstaabp:  I re-targeted it.  Although, note that at this point the two branches are identical.

taniwallach (changes requested):
At present the new set maker code does not work on a local Docker setup, but the fix seems to be relatively simple.

When I tried to test this in a local Docker setup on my machine:
  - The problems loaded (quite quickly), but the MathJax math did not display.
  - I saw things like `\(\displaystyle f(x) = (x^2-1)(x^2-4).\)` in the problem text.
  - The answer boxes did not use MathQuill.
  - The iFrame resizer did not work.
  - The browser developer tools show many requests to `localhost` while Docker has the server on `localhost:8080`, which explains much of the trouble.
  - I also could not get any of the buttons (`Preview my answers`, `Check answers`, `Show correct answers`) to work. All reported:
```
Unable to connect
Firefox can’t establish a connection to the server at localhost.
```

until I overcome the problem by manually adding
```
    ro.forcePortNumber = 8080;
```
to `htdocs/js/apps/SetMaker/setmaker.js` near line 365 (in the settings for `ro`).

With that small change, the MathJax renderer, MathQuill was working, the buttons worked, all `localhost` requests were to `localhost:8080`.

I had a similar issue with the ProblemSetDetail2 page, and adding the same hacked setting to `htdocs/js/apps/ProblemSetDetail2/problemsetdetail2.js` fixed it also.

@drgrice1  he challenge is that the server will need to provide a hint for when such a setting is needed and what the correct value for the `ro.forcePortNumber` value is, and the JavaScript code will need to make the addition when it is needed.

Explanation: The `html2xml` subsystem supports an optional `forcePortNumber` setting which was intended to add the port number where needed, exactly for settings where a port number is needed like the local Docker setup. See:
  - https://github.com/openwebwork/webwork2/commit/e4c108cc113b0a7e119efdd4fea15acce583f2b2

lib/WebworkWebservice/RenderProblem.pm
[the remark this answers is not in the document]
taniwallach: I think that you put back some sort of debugging code.
I'm not saying which of the bits were "good" and which were bad (it's not code I wrote, but I have found such bits help start debugging, as many of them do provide access to data which is hard to check otherwise)

I just felt that wholesale deletion was to far, and would have made it harder to debug.
Hopefully what you put as a revised option strikes a reasonable balance. For now, since you looked into it and tried to provide a replacement - I'm dropping the objection.

lib/FormatRenderedProblem.pm
taniwallach: So maybe put that option as a comment... The idea is to leave some reasonable "hints" for future use of how to access the data structures when debugging. I'm not sure I used this one, but I certainly did turn on a disabled `pretty_print` during some work on the `html2xml` routine in the past, and having the sample made it easy to do.

lib/WeBWorK/ContentGenerator/renderViaXMLRPC.pm
taniwallach: The comment was out of place, but accessing the `paramcache` to understand what is getting into the `$xmlrpc_client->{inputs_ref}` has been helpful several times now, such as in understanding what went wrong with @Alex-Jordan's iframe in the PreTeXt book. Doing any such debugging gets frustrating at times.

lib/WebworkWebservice/RenderProblem.pm
taniwallach: OK

taniwallach:
@drgrice1 - Overall this is a really nice change and improvement in my opinion. I would appreciate an attempt to make it work in the local Docker setup (or other settings where a non-standard port number is needed), but if that is not possible to do quickly, maybe put in the 2 JavaScript files a comment about what setting can be used to overcome the problem, and it can be noted in the release notes and left to fix in the future.

> I did not scrutinize the code for this, but I just tested again and it is working well. I have no basis to weigh in on the commented out debug code discussion. Aside from resolving that, this seems ready for merging.

@drgrice1 put in some replacement debugging code, and I'm trusting his calls on what to put back and how. I'm not sure I'm happy with the removal of all the "prettyy_print" from `lib/FormatRenderedProblem.pm` but the suggestion of how to get a similar result via `my $pretty_print_self = WebworkClient::pretty_print($self);` in https://github.com/openwebwork/webwork2/pull/1301#discussion_r608175176 seems to mean that a small comment with the revised version would make it easy to figure out, and would also use the more secure `pretty_print` from the `WebworkClient`.

pstaabp:
@taniwallach Not knowing a ton about docker, did you set the port in `webwork.apache2-config`?  This seems to be an apache setup issue.

pstaabp:
Inside the ProblemSetDetail2, if you have rendered all of the problems, it appears that any of the three buttons changes the seed of the problem.  I was shocked to get a chain rule problem wrong, only to realize the problem changed.  Seems like after any button has been clicked once, the seed stays the same. 

Doing the same in the Library Browser does not seem to change the seed.

drgrice1:
@taniwallach: I will look into the docker setup.

drgrice1:
@pstaabp:  I am seeing that too.  I will look into it.

taniwallach:
> Seems like after any button has been clicked once, the seed stays the same.

`html2xml` keeps the `problemSeed` and `psvn` from call to call be keeping them in hidden form fields, and carries them over from call to call. (This is know to be not perfect, as a user could tamper with those values, but that is how `html2xml` has worked thus far. A signed JWT token might provide a more secure approach for this in the future.)

> Inside the ProblemSetDetail2, if you have rendered all of the problems, it appears that any of the three buttons changes the seed of the problem. 

I confirm that I see this also... Did not notice it before, as I only tested a problem without a random parameter.

When the iFrame loads it seems to have
```
	       <input type="hidden" name="problemSeed" value="1">
```
and that remains on the following calls, but that apparently was not the seed actually used when the problem was generated the first time. It seems that the code to set `ro.problemSeed` is not working properly. The code should really be getting the `problemSeed`of the `effectiveUser` so it will remain the same, but JavaScript probably will not have access to the correct data, and in such cases `html2xml` provides some defaults which have no dependence on the values in the database. Most likely, a similar issue occurs with `psvn` for problems which use it, as that is also persisted by `html2xml` using a hidden form field.

The SetMaker case does not have this problem, as it just sets a random value for `ro.problemSeed` on the first call, and then that value should be retained via the hidden form field. However, the SetMaker code should also fix a value for `psvn`.

taniwallach:
Here is a sample PG file which fails to render for me in the Set Details 2 page. From a vague recollection, the Matrix is somehow the cause of the problem.
```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
);
TEXT(beginproblem());

Context('Matrix');

$A = Matrix([[0,-1],[1,0]]);

Context()->texStrings;
BEGIN_TEXT
Enter \( $A \) below.
$BR
\{ $A->ans_array \}
END_TEXT
Context()->normalStrings;

ANS( $A->cmp );

ENDDOCUMENT();
```

pstaabp:
I'm seeing this too both in ProblemSetDetail2 and the LibraryBrowser.  Here's the error:
```
faultstring: Not an ARRAY reference at /Users/pstaab/code/ww-docker/webwork2/lib/WebworkWebservice/RenderProblem.pm line 507.
```

However, in the Student view of the problem (I forget the name of the perl module), there is not error.   Trying to look at line 507 of that file, looks like it is handling of the xml.

taniwallach:
I suspect that only when the set editor is open for a specific single
student and showing that student's seeds should any attempt be made to set
the specific seed+psvn for that user's problems be made (so that student's
questions will render). I did not think carefully about what the JS code
was doing in the first case,but I suspect that was the idea.

In other cases, maybe text that a random seed was used could precede the
problem and then the approach from the library broswer setting a random
initial value for ro.problemSeed could be used.

drgrice1:
@pstaabp, @taniwallach:  Note that the issue of the problem re-rerandomizing itself in the set detail page does not occur if you are viewing the details for a particular student (or your own user).  It only happens when you are viewing the set detail for ALL students.  Furthermore, I have a fix for that and will commit it soon.

@taniwallach:  There really is no need for the library browser to set the psvn.  There is no problem set, so it can just use the default psvn.

@taniwallach:  Yes that broken problem will not work via the xmlrpc method as we have already seen.  That isn't new.  We know that problem needs to be fixed to add the line (commented out in what you posted) to set the context back to normal strings.  There shouldn't be that many of them in the OPL.

taniwallach:
My recollection is that with the string setting still on tex strings,
something gets into the data structures which the WW XML serialization code
cannot handle. That code is a recursive "parser" and serializer and makes
assumptions which usually are OK. Not when this happens.

If I recall such problems made trouble in the set editor internal rendering
before this PR... so this is not a new problem, just something I recalled
so tested and pointed out. It could be that better tests in that code might
solve the problem.

The issue is almost certainly a "bad" test which determines that the tex
string form of the correct_answer (whatever it is really called in the data
structure) is a Perl array reference when it is not.

On my phone and signing off for tonight, so not checking more details now.

drgrice1:
> My recollection is that with the string setting still on tex strings, something gets into the data structures which the WW XML serialization code cannot handle. That code is a recursive "parser" and serializer and makes assumptions which usually are OK. Not when this happens. If I recall such problems made trouble in the set editor internal rendering before this PR... so this is not a new problem, just something I recalled so tested and pointed out. It could be that better tests in that code might solve the problem. The issue is almost certainly a "bad" test which determines that the tex string form of the correct_answer (whatever it is really called in the data structure) is a Perl array reference when it is not. On my phone and signing off for tonight, so not checking more details now.

Yeah, I tested this particular problem as I was originally working on this pull request, and knew it wasn't going to work.  I also expected you would probably do the same, and bring it up.  The xml serialization code needs to be investigated further.

drgrice1:
@taniwallach:  Is there any real reason to set the `base href` in the html2xml formats?  This predates anything that we have done with webwork, and as far back as I can see in the git repository this was set by @mgage.  In any case, I don't see the point of that.  Is that a security thing?  In any case, the simple fix for the docker build is to remove that.

drgrice1:
I went with a simpler javascript solution for fixing the docker build.

lib/FormatRenderedProblem.pm
drgrice1: I added this one back for you.

drgrice1:
It is worth noting that the troublesome problem mentioned here does work if used directly with the html2xml endppoint.  For example with
`http://localhost/webwork2/html2xml?courseID=myTestCourse&userID=course_login&password=secret&course_password=secret&displayMode=MathJax&outputformat=simple&problemSeed=17&sourceFilePath=Library/Rochester/setLinearAlgebra11Eigenvalues/ur_la_11_30.pg`

drgrice1:
The Library/Rochester/setLinearAlgebra11Eigenvalues/ur_la_11_30.pg problem is now fixed with this pull request as of the last commit.

drgrice1:
Okay, now that problem is fixed.  I added a safety check for hash and array references before trying to use them as such.  There are potentially a multitude of things that could end up passing a case insensitive test for containing the string "HASH" or "ARRAY".  Blessed objects that are not actually valid hash or array references could pass this test.  This prevents that issue.

taniwallach:
> @pstaabp, @taniwallach: Note that the issue of the problem re-rerandomizing itself in the set detail page does not occur if you are viewing the details for a particular student (or your own user). It only happens when you are viewing the set detail for ALL students. Furthermore, I have a fix for that and will commit it soon.

Thanks - I realized later that the single user case was what the code was probably handling.

> @taniwallach: There really is no need for the library browser to set the psvn. There is no problem set, so it can just use the default psvn.

Good point - sorry.

> @taniwallach: Yes that broken problem will not work via the xmlrpc method as we have already seen. That isn't new. We know that problem needs to be fixed to add the line (commented out in what you posted) to set the context back to normal strings. There shouldn't be that many of them in the OPL.

Yes, but while were fixing things in this part of the code-base, we have the chance to see if we can fix it... Looking at the following comments and new changes.

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -265,32 +205,27 @@ sub renderProblem {
...
+	# Define the set and problem information from
+	# data in the environment if necessary
+	my $setName = (defined($rh->{set_id})) ? $rh->{set_id} :
+		(defined($rh->{envir}->{setNumber}) ? $rh->{envir}->{setNumber}  : '');
 
 	my $setVersionId = $rh->{version_id} || 0;
-	
+
 	my $problemNumber =  (defined($rh->{envir}->{probNum})   )    ? $rh->{envir}->{probNum}      : 1 ;
-	my $problemSeed   =  (defined($rh->{envir}->{problemSeed}))   ? $rh->{envir}->{problemSeed}  : 1 ;
```
taniwallach, line 285: Here you removed the local definition of `$problemSeed` by dropping line 227 (old numbering, between 220 and 221 in the new numbering), so maybe line 240 (old numbering, now 233) which issues a `debug()` call with the value should be removed or relocated.

lib/FormatRenderedProblem.pm
```diff
@@ -268,6 +268,9 @@ sub formatRenderedProblem {
 	my $footer = $showFooter && $showFooter eq "no" ? ''
 		: "<div id='footer'>WeBWorK &copy; 2000-2021 | host: $SITE_URL | course: $courseID | format: $self->{inputs_ref}{outputformat} | theme: math4</div>";
 
+	# For debugging purposes add $pretty_print_self to the output format in use and uncomment below.
+	#my $pretty_print_self = WebworkClient::pretty_print($self);
```
taniwallach, line 272: @drgrice1 Thank you!

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -460,58 +460,70 @@ sub xml_filter {
...
-				if ( ! Scalar::Util::looks_like_number( $filtered_value ) &&
-				     $filtered_value =~ /[^\x00-\x7f]/ # Some non 7-bit character included
-				   ) {
-					# UTF-8 encoding needed
-					$input->{"xmlrpc_UTF8_encoded_$item"} = encode("UTF-8", $filtered_value );
-					delete( $input->{$item} ); # remove the original value
+		eval { my %test = %{$input}; };
+		if ($@) {
+			print DEBUGCODE "($input) looks like a HASH reference but is not (type is $type)\n" if $debugXmlCode;
+			$input = "HASH reference";
```
taniwallach, line 466: Maybe the value of `$input` should be set to something like `Misunderstood as a HASH reference but is not (type is $type)`
Could the two `eval` tests be done before the start of this `if-elsif-else` block, so we could replace the tests like `elsif( $type =~/HASH/i or "$input"=~/HASH/i)` with something where the second test is not only "string match" based but depends on the `eval` test if the "string match" test would be true and trigger entering the following code.

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -460,58 +460,70 @@ sub xml_filter {
...
-		my $tmp = [];
-		foreach my $item (@{$input}) {
-		    # print DEBUGCODE "-----checking $item of type\n",ref($item) if $debugXmlCode;
-			$item = xml_filter($item,$level);
-			push @$tmp, $item;
-			# print DEBUGCODE "-----end checking $item\n" if $debugXmlCode;
+		eval { my @test = @{$input}; };
+		if ($@) {
+			print DEBUGCODE "($input) looks like an ARRAY reference but is not (type is $type)\n" if $debugXmlCode;
+			$input = "ARRAY reference";
```
taniwallach, line 512: Here to the `$input` can explain that it was misuderstood to be an array reference, but if the `eval` test was done before the `if-elsif` block we could probably avoid this false positive getting here at all.

lib/WebworkWebservice/RenderProblem.pm
```diff
@@ -460,58 +460,70 @@ sub xml_filter {
...
 		}
-		$level--;
-		print DEBUGCODE "  "x$level."HASH reference completed \n" if $debugXmlCode;
 	} elsif( $type=~/ARRAY/i or "$input"=~/ARRAY/i) {
-		print DEBUGCODE "  "x$level."ARRAY reference with ".@{$input}." elements will be investigated\n" if $debugXmlCode;
-		$level++;
-		my $tmp = [];
-		foreach my $item (@{$input}) {
-		    # print DEBUGCODE "-----checking $item of type\n",ref($item) if $debugXmlCode;
-			$item = xml_filter($item,$level);
```
taniwallach, line 591: So far we are still assuming that arrays will not have wide Unicode characters. If this is getting to be a dangerous assumption, something like the test above on the value might be needed, but would require more work, as we cannot use a modified hash key to inform the decoder that the value was encoded into UTF-8 so need to be decoded on the other side.

I was to busy and lazy to to look for a good solution when I have not yet seen any case where a solution to handle this would be needed.

I hope webwork3 will not use any of this "xml filter" code, and the replacement will natively support UTF-8, so we need not fix what is not needed now, and will hopefully be replace by a better solution in the near future.

taniwallach (commented):
@drgrice1 - The changes are certainly good, but once you are fixing the `xml_filter` I think we can probably do it better.
Some inline review comments based on reading the changes, but I have not yet tested the code.

taniwallach (commented):
> @taniwallach: Is there any real reason to set the `base href` in the html2xml formats? This predates anything that we have done with webwork, and as far back as I can see in the git repository this was set by @mgage. In any case, I don't see the point of that. Is that a security thing? In any case, the simple fix for the docker build is to remove that.

I don't think this is the best thing to do, but it also predates my work. I don't think it was a security thing. I suspect that the idea was to get things to work (or work "better") within `iframe` embedding, but still to leave the `href` values as relative values, as is typical elsewhere in the code-base.

I think a better solution might be to optionally force the main `href` to add the entire "base" at the start, as if a page were embedded in a different page but not inside an iframe, modifying the `base_href` is the wrong thing to do.

On my server, I have a locally modified version of `lib/WebworkClient/json_format.pl` where the server portion of the URL was hardwired, which was a temporary approach which worked well enough for the proof-of-concept (single problems embedded using a modified `LTI` subsystem into an open-edX MOOC course). Hardwired values are not meant to be a long-term solution, but allowed things to work.

Part of the problem is that without a server side (maybe course setting) hard-coded value for the fixed initial URL portion, that setting needs to be persisted to the next call to `html2xml` which current approaches would do using a hidden form field and that is very open to abuse.

The `JWT` approach of persisting data which should not be tamperered with on the client side is being used already in some of the standalone render code at least in https://github.com/drdrew42/renderer/pull/49 . Such an approach is a better long-term solution to persisting data which should not be tampered with, and other parts of PG could put it to use. I don't think we have the time to get such changes into the current RC, and probably not even into 2.16 at all. @drdrew42 could probably explain how they are starting to use `JWT` in the standalone renderer.

> The Library/Rochester/setLinearAlgebra11Eigenvalues/ur_la_11_30.pg problem is now fixed with this pull request as of the last commit.

Will test soon. Put in some review comments on the code changes and some related thoughts now while I have a small bit of time to work on WW.

> There is something seriously wrong with the code in the `xml_filter` subroutine. 

Yes, it is ugly, complicated, and far from perfect. It made this subsystem work well enough over 10 years ago, and the UTF-8 support brought many problems to my attention. When we replace `XMLRPC::Lite` or when we move to webwork3, hopefully it will be totally replaced.

> Clearly you can't use a variable as a reference to a hash or array unless it is a reference to a hash or array. I am surprised this code doesn't break more frequently.

That bad test is 16 years old: https://github.com/openwebwork/webwork2/blame/68f0f146f8ac021fd50e11fc52aba098c96978db/lib/WebworkWebservice/RenderProblem.pm#L537

But it worked well enough, often enough, for long enough that apparently people just avoid using problems where it failed.

> Okay, now that problem is fixed. I added a safety check for hash and array references before trying to use them as such. There are potentially a multitude of things that could end up passing a case insensitive test for containing the string "HASH" or "ARRAY". Blessed objects that are not actually valid hash or array references could pass this test. This prevents that issue.

Yeah! Will try to test soon, but that may only be early next week.

lib/WebworkWebservice/RenderProblem.pm
drgrice1: That was a shadow definition.  Look up.  `$problemSeed` is defined and set equal to `$rh->{envir}->{problemSeed}` on line 113.  Then it is checked that `$problemSeed` is defined on line 116.   Then this shadow definition what put in that does all of that again.  If `use warnings` were in this file (not sure why it is not) we would see warnings.  The same was true of the `$courseName` and `$user` variables.

taniwallach: Sorry. I did not look back far enough in the file.

lib/WebworkWebservice/RenderProblem.pm
drgrice1: I don't see a valid way to do the `eval` check before the `if-elsif-else` block.  The "misunderstanding" hasn't occurred yet.  So you would have to double up the messy `$type =~/HASH/i or "$input"=~/HASH/i` check.  Basically, you would have to do the `if-elseif-else` block twice because you have to know what type to test the `$input` for.  Note that the `$type =~ /.../i` check can also give false positives.  The only proper check that validates being able to use `$input` as a hash or array reference is the check I just introduced.  How do you envision your suggestion working, and why do you think it is better?

I will adjust the wording a bit on the debugging output. You are dropping the "($input)" on the output which I won't do.  We already know that `$input` can be stringified at this point, so it is useful to display it in the debugging output.

lib/WebworkWebservice/RenderProblem.pm
drgrice1: I will let you check this and figure out how to do it, if you have reason to believe it is needed.

lib/WebworkWebservice/RenderProblem.pm
taniwallach: Sounds good.

lib/WebworkWebservice/RenderProblem.pm
taniwallach: OK - we'll deal with it if (when?) we discover a need. I was just pointing out the potential issue as more and more potential UTF-8 usage occurs.

pstaabp (approved):
All of the issues that I see are fixed. 

@taniwallach Clearly there's more work to do with the xmlrpc, but I see that being replaced this summer, so I don't know if we want to spend a lot of effort to fix the little bugs that have been there for a long time.

taniwallach:
> All of the issues that I see are fixed.
> 
> @taniwallach Clearly there's more work to do with the xmlrpc, but I see that being replaced this summer, so I don't know if we want to spend a lot of effort to fix the little bugs that have been there for a long time.

Overall I agree. Since more of the system now depends on it - fixing the failed load was probably worth the effort, especially since WW2 will remain in production for a good while after the development team and early adopters move to WW 3.

@drgrice1 - Thanks for working on all these fixes and coding the main changes.

taniwallach (changes requested):
Most of the prior issues are fixed.

There is still an issue with non-stable randomization behavior in "Set Detail 2" (`htdocs/js/apps/ProblemSetDetail2/problemsetdetail2.js`) both for a single student and the general case when the PG file uses `SRAND($psvn);`. The problem is apparently that the fallback value for `psvn` in `html2xml` is `54321` (which is what gets into the hidden form field) but that is not what is used in the initial render. The problem below displays the value of `$psvn` and it changes to `54321` when on of the buttons is clicked.

Sample PG file which exhibits the issue:
```
DOCUMENT();
loadMacros(
  "PGstandard.pl",
  "MathObjects.pl"
);

TEXT(beginproblem());

Context("Numeric");


BEGIN_TEXT
psvn \( = $psvn \)
$PAR
END_TEXT

# Seed the RNG so that a sequence of problems will use the same seed.
SRAND($psvn);

$a = random(1,5,1);
$b = 10 * random(2,5,1);

$ans1 = Real( $a + $b );

Context()->texStrings;
BEGIN_TEXT
\( a = $a, b = $b \)
$BR
\( a + b = {} \)
\{ $ans1->ans_rule(5) \}
END_TEXT
Context()->normalStrings;

ANS( $ans1->cmp );

ENDDOCUMENT();
```

drgrice1:
@taniwallach:  Okay.  I have a fix for that.  I was going to add that when I added the fix for the problem seed, and should have.

taniwallach (approved):
These last commits fix the issue with `psvn` in `Set Detail 2`.

@drgrice1 - Great work and patience with all the tedious issues and corrections.

drgrice1:
I am glad it is getting thorough testing. There were a lot of issues to resolve.  This is a big change, and needs a lot of testing.

Thanks @Alex-Jordan, @pstaabp, and @taniwallach for all of your reviews and testing thus far.

pstaabp:
I see 3 positive reviews.  Let's merge this in so it can be tested on the RC!!

taniwallach:
@drgrice1 In commit https://github.com/openwebwork/webwork2/pull/1301/commits/c08d4faa83f4944c02eae386d2a65e026d0de1d3 you persist the `$extra_header_text` only for simple format. 

Is there some reason what it is not relevant in the other "html" style formats like 
  - `standard_format.pl`
  - `sticky_format.pl`
  - `simple2_format.pl`

It looks like some may need other updates to get them back in sync with other recent changes.

In any case, any such fixes will wait for later... @pstaabp merged this, which is good, as now this code is in the RC.

drgrice1:
The other formats could use some work.  Notice there were several things added to the simple_format.pl in this pull request that were not done to the others (the footer, header, etc).  Those formats will still work, but yes they should get some attention.  I will work on adding that.

PR #1306, opened 2021-04-01
Fix POD generation errors.

drgrice1:
In addition to repairing some issues with POD in the code, this pull request adds a scripts for generating POD from the webwork2 and pg source code.  These can be used to create the POD published on webwork.maa.org.  The scripts are modified from the scripts in https://github.com/mgage/admintools.git.
    
I placed them in a dev_scripts subdirectory of the bin directory as they are not scripts that those installing webwork will use.  This prevents them from appearing in the PATH.

### DIFF
--- a/bin/dev_scripts/PODtoHTML.pm
+++ b/bin/dev_scripts/PODtoHTML.pm
@@ -1,0 +1,245 @@
+#!/usr/bin/perl
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2021 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+use strict;
+use warnings;
+
+package PODtoHTML;
+
+use File::Find qw(find);
+use File::Path qw(make_path);
+use IO::File;
+use Pod::Find qw(contains_pod);
+use POSIX qw(strftime);
+use PODParser;
+
+our @sections = (
+	'/'     => "(root)",
+	bin     => "Scripts",
+	conf    => "Config Files",
+	doc     => "Documentation",
+	lib     => "Libraries",
+	macros  => "Macros",
+	clients => "Clients"
+);
+
+sub new {
+	my ($invocant, %o) = @_;
+	my $class = ref $invocant || $invocant;
+
+	my @section_list = ref($o{sections}) eq 'ARRAY' ? @{$o{sections}} : @sections;
+	my $section_hash = {@section_list};
+	my $section_order = [ map { $section_list[2 * $_] } 0 .. $#section_list / 2 ];
+	delete $o{sections};
+
+	my $self = {
+		%o,
+		idx => {},
+		section_hash => $section_hash,
+		section_order => $section_order,
+	};
+	return bless $self, $class;
+}
+
+sub convert_pods {
+	my $self = shift;
+	my $source_root = $self->{source_root};
+	my $dest_root = $self->{dest_root};
+
+	find({ wanted => $self->gen_pod_wanted, no_chdir => 1 }, $source_root);
+	$self->write_index("$dest_root/index.html");
+}
+
+sub gen_pod_wanted {
+	my $self = shift;
+	return sub {
+		my $path = $File::Find::name;
+		my $dir = $File::Find::dir;
+		my ($name) = $path =~ m|^$dir(?:/(.*))?$|;
+		$name = '' unless defined $name;
+
+		if ($name =~ /^\./) {
+			$File::Find::prune = 1;
+			return;
+		}
+		unless (-f $path or -d $path) {
+			$File::Find::prune = 1;
+			return;
+		}
+		if (-d _ and $name =~ /^(\.git|\.github|t|htdocs)$/) {
+			$File::Find::prune = 1;
+			return;
+		}
+
+		return if -d _;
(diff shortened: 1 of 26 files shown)


taniwallach (approved):
All the syntax seems correct, and the links work.
I did not check output in POD format, just the code here.

drgrice1:
I should have added a comment to the pull request perhaps.  These are issues seen when running the script @mgage gave me to generate the POD documentation.  There are no code changes, just POD changes.

If you have `perldoc` or `pod2html`, you can test with those.  `perldoc` won't show the links though.

The scripts use the Pod::Html perl package.

drgrice1:
I removed the addition of the external links.  I am going to switch from Pod::Html (which seems to be out of date and unmaintained), to Pod::Simple::XHTML which is able to resolve these links as they are.

pstaabp:
What I could tell, it seems like testing the `generate-ww-pg-pod.pl` script is what is needed here. A few things I noticed:
1.   Should this work for generating some local files?  I'm trying to test on my laptop where I develop and was hoping this would work as a folder of html files that I could navigate with a browser under `file:///` None of the links however are working.   (I even played with the -b flag).   If this isn't possible with this script, then that's okay.
2.  I wonder if the environmental variable `PG_ROOT` is not set, but `WEBWORK_ROOT` is, that it could try `WEBWORK_ROOT/../pg` 
3. It seems like if the environmental variable `WEBWORK_ROOT` is set that this script should run.  You get the help/usage menu, however the last part of it:
    ```
    Note that at least one of the options --webwork-root or --pg-root must
        be provided (or there is nothing to do!).
    ```
    The reason it won't run is that `output-dir` is not set.  I would suggest either giving this error or changing the help menu to emphasize that `--output-dir` must be set.

drgrice1:
1.  This will work (somewhat) for local files If you use something like
`./bin/dev_scripts/generate-ww-pg-pod.pl -b file:///home/user/pod -o /home/user/pod`
When clicking on the initial PG or WeBWorK links on the primary index, you will then have to click on `index.html`.  A local file system link does not automatically load index.html.  However, the links inside the POD html files will work.

2.  That is inconsistent with all of the other scripts and usage in WeBWorK, and not reliable.  You already have the option to set both the webwork root and pg root, so I don't see this as needed.

3.  The output when the option `--output-dir` is not set is
```
Usage:
    generate-ww-pg-pod.pl [options]

     Options:
       -w|--webwork-root     Directory containing a git clone of webwork2.
                             If this option is not set, then the environment
                             variable $WEBWORK_ROOT will be used if it is set.
       -p|--pg-root          Directory containing  a git clone of pg.
                             If this option is not set, then the environment
                             variable $PG_ROOT will be used if it is set.
       -o|--output-dir       Directory to save the output files to. (required)
       -b|--base-url         Base url location used on server. (default: /)
                             This is needed for internal POD links to work correctly.
       -v|--verbose          Increase the verbosity of the output.
                             (Use multiple times for more verbosity.)

    Note that at least one of the options --webwork-root or --pg-root must
    be provided (or there is nothing to do!).
```
Note the word `required` by the `--output-dir` option.

pstaabp:
I ran the script similar to: 

> 1. This will work (somewhat) for local files If you use something like
>    `./bin/dev_scripts/generate-ww-pg-pod.pl -b file:///home/user/pod -o /home/user/pod`
>    When clicking on the initial PG or WeBWorK links on the primary index, you will then have to click on `index.html`.  A local file system link does not automatically load index.html.  However, the links inside the POD html files will work.

and I am seeing this working.  As local files, however, any link that is not an HTML however, opens up the Finder (MacOS) for me and shows the file.  It appears to be an HTML file, but not with the proper extension.  For example, this happens in Scripts and Config Files.  

If this will link properly on the server, then I'm fine with this, however, a server may not properly serve the file without the HTML extension.

bin/dev_scripts/generate-ww-pg-pod.pl
```diff
@@ -0,0 +1,129 @@
...
+<title>WeBWorK/PG POD</title>
+</head>
+<body>
+<h1>WeBWorK/PG POD</h1>
+<h2>(Plain Old Documentation)</h2>
+<div>
+<ul>
+EOF
+
+	print $fh q{<li><a href="pg">PG</a></li>} if $pg_root;
```
pstaabp, line 123: You could put `href="pg/index.html"` here to ensure the top links work.

drgrice1:
> If this will link properly on the server, then I'm fine with this, however, a server may not properly serve the file without the HTML extension.

These links work fine on the maa server.  They also work fine on my local computer (linux must handle this better).

bin/dev_scripts/PODtoHTML.pm
```diff
@@ -0,0 +1,245 @@
...
+
+	$self->update_index($subdir, $html_rel_path, $pod_name);
+	make_path($html_dir);
+	my $html = $self->do_pod2html(
+		pod_path => $pod_path,
+		pod_name => $pod_name
+	);
+	my $fh = new IO::File($html_path, '>')
+		or die "Failed to open file '$html_path' for writing: $!\n";
+	print $fh $html;
```
pstaabp, line 139: It looks like we could check that the `$html_path` has extension `.html` here and if not add it.  This could ensure that links open.

bin/dev_scripts/generate-ww-pg-pod.pl
drgrice1: This is designed for webwork.maa.org, and works well there.  On an actual server you usually don't want to include "index.html" in the href.  That is the point of an "index.html" file.

pstaabp:
> > If this will link properly on the server, then I'm fine with this, however, a server may not properly serve the file without the HTML extension.
> 
> These links work fine on the maa server. They also work fine on my local computer (linux must handle this better).

If they work on the server, then never mind about my comments.  That's the important place to work.

drgrice1:
> If they work on the server, then never mind about my comments. That's the important place to work.

Just to make sure, check them on the server.  Go to https://webwork.maa.org/pod/ and see if those links work for you.

pstaabp (approved):
Looks great.  I'll merge this.

bin/dev_scripts/PODtoHTML.pm
drgrice1: I think this is not a bad idea.  Since you already merged this I will put in another pull request that adds this.  The modification needs to be done on line 128 and also to the relative path on line 129 to work right.

PR #1307, opened 2021-04-01
SameSite cookie default to Lax

taniwallach:
Change SameSite cookie setting to Lax as default, and add some explanations about the options and why Lax was selected as the default value.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -918,7 +918,13 @@ $caliper{enabled} = 0;
 #         https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
 #         https://tools.ietf.org/html/draft-west-cookie-incrementalism-00
 
-$CookieSameSite = "Strict";
+
+# Notes about the $CookieSameSite options:
+#   The "None" setting should only be used with HTTPS and when $CookieSecure = 1; is set below. The "None" setting is also less secure and can allow certain types of cross-site attacks.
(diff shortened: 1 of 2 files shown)


conf/defaults.config
```diff
@@ -918,7 +918,13 @@ $caliper{enabled} = 0;
 #         https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
 #         https://tools.ietf.org/html/draft-west-cookie-incrementalism-00
 
-$CookieSameSite = "Strict";
+# Notes about the $CookieSameSite options:
+#   The "None" setting should only be used with HTTPS and when $CookieSecure = 1; is set below. The "None" setting is also las secure and can allow certain types of cross-site attacks.
+#   The "Strict" setting creak the links in the system generated fe
+edback emails when read in a web mail client.
+#   Due to those factors, the "Lax" setting is probably the optimal choice for typical WeBWorK servers.
```
drgrice1, line 925: It looks like there are some typos that need to be fixed in both files.  `las secure`, `creak the links`?

There also seems to be a linebreak that shouldn't be (also in both files).

taniwallach: Typing while exhausted and almost cross-eyed. Thanks for noticing. Force pushed the correction.

drgrice1:
Looks good.

drgrice1 (approved):
This is ready to merge.

PR #1309, opened 2021-04-05
Add a script that will upgrade the types of columns in the database.

drgrice1:
This can be used by those upgrading from a version of WeBWorK prior to WeBWorK 2.15.  This script will also be need to be run anytime a course archive is unarchived that was created with a version of WeBWorK prior to 2.15.

See the documentation in the script for its detailed usage.

Eventually the things this script does can be worked into the course administration database check/upgrade methods, but this is the quick solution for now.


bin/upgrade-database-to-utf8mb4.pl
```diff
@@ -0,0 +1,306 @@
...
+##############################################################################
+
+=head1 NAME
+
+upgrade-database-to-utf8mb4.pl -- Upgrade webwork course database tables from
+latin1 to utf8mb4.
+
+=head1 SYNOPSIS
+ 
+  upgrade-database.pl [options]
```
pstaabp, line 24: Should this be `upgrade-database-to-utf8mb4.pl` ?

drgrice1: Yep.

drgrice1: Fixed.

pstaabp:
I ran the script with the backup option, verbose on one archived course and got the following error:
```
mysqldump: Got error: 1045: "Access denied for user 'webworkWrite'@'localhost' (using password: NO)" when trying to connect
```

It seems like on line 191 of the script, it's not passing the password to `mysqldump`.  Isn't this needed?

drgrice1:
Yes it should. I have a .my.cnf file in my home directory that has my my credentials so it works for me without that, but I meant to add that.  I will fix it when I have time.

pstaabp (approved):
This is working now and it did convert over the course from latin1 to `utf8mb4_general_ci`. 

It seems my other courses are `utf8_general_ci` which I created natively from develop in the past few months. I guess I should upgrade these as well.

drgrice1:
You can test the script on those courses, but if they were created on the develop branch the script won't actually do anything. So you are testing to see that that is the case.

pstaabp:
> You can test the script on those courses, but if they were created on the develop branch the script won't actually do anything. So you are testing to see that that is the case.

The script did update the type from `utf8_general_ci` to `utf8mb4_general_ci`. 

> You can test the script on those courses, but if they were created on the develop branch the script won't actually do anything. So you are testing to see that that is the case.

It did convert all of the columns in the course from `utf8_general_ci` to `utf8mb4_general_ci`.  This was not what was intended?

drgrice1:
Yeah, it should have. I didn't the collation closely.  Of those were created with develop and your database was configured to use utf8mb4 then the script would have done nothing. You must have had your database configured to default to utf8 when you created the courses.

pstaabp:
Got it.  I wasn't thinking about the datatypes when recreating a new development machine a few months ago.  It was probably whatever was currently in the develop pipeline.

taniwallach:
> Got it. I wasn't thinking about the datatypes when recreating a new development machine a few months ago. It was probably whatever was currently in the develop pipeline.

The setting of whether to use the 3-byye MySQL `uf8` or the 4 byte `utf8mb4` is `$ENABLE_UTF8MB4 =1;` which is in `conf/site.conf.dist` and needs to be set in `site.conf` to get the full `utf8mb4` use. It could be you don't have that set.


See: 
  - https://github.com/openwebwork/webwork2/blob/03eca4cc1a6034d24cf28bd690f0f56ce14f284b/conf/site.conf.dist#L187
  - https://github.com/openwebwork/webwork2/blob/05c982497e68d105829e190475ad73979df89c38/conf/site.conf.dist#L230

dlglin:
This doesn't currently work with CentOS 7.  `selectall_array` was added in DBI version 1.635.  CentOS currently provides version 1.627.

drgrice1:
@dlglin:  Please test with the latest commit.  Now the script shouldn't work on your system unless you use the --no-backup option.  It should work if you do.  Of course, you should make sure you do have a backup before testing.

dlglin:
> @dlglin: Please test with the latest commit. Now the script shouldn't work on your system unless you use the --no-backup option. It should work if you do. Of course, you should make sure you do have a backup before testing.

The script now aborts if the mysqldump doesn't complete successfully, with a reasonable error message.

dlglin:
This runs on CentOS now.

When running on a copy of my prod DB I got the following error:

> Failed to modify 'user_id' in 'Diagnostic_Test_key' from 'varchar(16)' to 'tinyblob.
> It is recommended that you restore a database backup.  Make note of the
> error output below as it may help in diagnosing the problem.  Note that
> the most common reason for this error is the existence of a data value
> in a column that does not fit into the smaller size data type that was
> needed for the utf8mb4 change.
> DBD::MariaDB::db do failed: BLOB/TEXT column 'user_id' used in key specification without a key length at /opt/webwork/webwork2/bin/upgrade-database-to-utf8mb4.pl line 242, <> line 1.
> 

The structure of the table is as follows:
MariaDB [wwprod]> show create table Diagnostic_Test_key;

> +---------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> | Table               | Create Table                                                                                                                                                                                                        |
> +---------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> | Diagnostic_Test_key | CREATE TABLE `Diagnostic_Test_key` (
>   `user_id` varchar(16) NOT NULL,
>   `key_not_a_keyword` text,
>   `timestamp` bigint(20) DEFAULT NULL,
>   UNIQUE KEY `user_id` (`user_id`)
> ) ENGINE=MyISAM DEFAULT CHARSET=latin1 |
> +---------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> 

Since the user_id field is set to 16 characters there was no need to explicitly set a key length, but when converting this to tinyblob it requires a specified key length, which is why it is failing for me.

This particular DB has been through many versions of WW and several different DB architectures.  At one point we were using a DB cluster that couldn't handle blobs in keys, so it's quite possible that I manually altered the table at that time, so this may only apply to my DB.  Is it worth having the script explicitly set the key length before changing the type of the columns?

PR #1313, opened 2021-04-09
Fix an unitialized usage in join or string issue in WeBWorK.pm.

drgrice1:
This shows up when errors happen and the sec-ch-ua parameter is not defined.  This happens anytime that errors occur through the WebworkWebservice as that parameter is not set there.


Alex-Jordan:
Merging untested (aside from testing that I could restart apache) since the code change is so clear and just makes sense.

PR #1314, opened 2021-04-09
Fix errors caused by attempting to return utf8 base64 encoded errors in the WebworkWebservice.

drgrice1:
Apparently there are characters in these errors that cause `faultstring: Wide character in subroutine entry at /usr/share/perl5/XMLRPC/Lite.pm`.


drgrice1:
I don't think this is actually causing the problem I thought it was.  Closing for now.

PR #1316, opened 2021-04-13
Exclude files from evaluating achivements and incrementing counters.

robert-marik:
There is a file `preamble.at` which excludes certain set or problem from assigning achievements. However, points counting to the next level are still assigned since the code which increments `$globalData->{'completeProblems'}` and  `$achievementPoints` preceeds the lines where `preamble.at` takes effect. The patch solves this problem.

### DIFF
--- a/courses.dist/modelCourse/course.conf
+++ b/courses.dist/modelCourse/course.conf
@@ -89,4 +89,14 @@ $dbLayoutName = 'sql_single';
 # $LTIBasicConsumerSecret = "xxxxxxxxx";
 # $LTIGradeMode = "homework";
 
+# For achievements you may set the preamble.at file as described at
+# https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4209 to exclude
+# certain sets or questions from achievement system.  However, these
+# questions still count to the score used to determine the level.  The
+# variable $achievementExcludeSet allows to exclude certain set completely
+# from the achievement system and these questions do not count to the level.
+# Remember that the names of the sets do not contain spaces.
+#
+# $achievementExcludeSet = ["00_Introduction","99_Summary","Sample_exam"];
+
 1;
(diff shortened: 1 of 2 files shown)


robert-marik:
btw: Another thing which should be fixed is the test that the points for a certain problem have been assigned. Currently we have `if ($thisProblem->status == 1 && $thisProblem->num_correct == 1)` in `AchievementEvaluator.pm` which is not sufficient.

drdrew42:
You are correct that preamble.at is only applied to each individual achievement.at file, and global achievement data is indeed updated regardless of 'ignored' sets in the preamble. This patch would make it possible to truly bypass achievements for a specified problem set. However, I think it would be more useful if course.conf supported an array of setIDs to be excluded from achievements, rather than being limited to a single set.

> btw: Another thing which should be fixed is the test that the points for a certain problem have been assigned. Currently we have `if ($thisProblem->status == 1 && $thisProblem->num_correct == 1)` in `AchievementEvaluator.pm` which is not sufficient.

Aren't we assured that `problem->status` and `problem->num_correct` don't change unless the problem has been assigned to the user (and the attempt occurs between open and close dates)?

robert-marik:
Thank you for comments. I am not comfortable with Perl so I am sorry that the patch was poor.

Concerning the `problem->num_correct` check your email, please. I sent you a private message. Roughly: the current setting allows to earn really high level by answering just a single problem.

robert-marik:
I did the second commit according to the suggestion from drdrew42. More sets can be excluded now, for example the introductory set and the exam sample. If someone wants to exclude just selected problems, then she/he should still use the `preamble.at` method. However, in this case the answer has influence on the global variables counting answered points and level points.

robert-marik:
The intended changes are in fact just two small changes in files  `lib/WeBWorK/AchievementEvaluator.pm` and `conf/localOverrides.conf.dist`. I tried to rebase in order to resolve conflict with further patches but evidently with little success.

drgrice1:
You seem to have rebased onto the WeBWorK-2.16 branch.  That is fine, but you will need to re-target this pull request to that branch.  I can do that for you if that is what you intended.

robert-marik:
Thank you. It seems that I did it.

drdrew42 (approved):

PR #1317, opened 2021-04-13
Increment number of items in achievements rather than setting to 1.

robert-marik:
The items earned by achievements can be used more than once (see https://github.com/openwebwork/webwork2/commit/e78c63b004ac3d3a5e07b2fe02c41395fa94b60f ). For this reason we should increment number of items earned on each level rather than set to 1.

### DIFF
--- a/courses.dist/modelCourse/templates/achievements/level_eight.at
+++ b/courses.dist/modelCourse/templates/achievements/level_eight.at
@@ -16,7 +16,7 @@ my $newLevelThreshold = 1200;
 
 if ($achievementPoints >= $nextLevelPoints) {
     #set new threshold and return 1 
-    $globalData->{DoubleSet} = 1;
+    $globalData->{DoubleSet}++;
     $nextLevelPoints = $newLevelThreshold;
     return 1;
 } else {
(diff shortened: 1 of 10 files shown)


Alex-Jordan:
Did you check if these values ever initialized to 0 in the first place? Or do they start out as undefined, and then `++` would (I think?) not work. I'm sorry for not checking myself, just short on time and thinking you may know.

robert-marik:
I checked it and it worked. Not sure why some of the variables need to be initialized and other not. I am not skilled in Perl and in details of WeBWorK. However I tested the code. There was no warning and the system behaved as expected. (To earn points easily I  used really large value for `Achievement Points Per Problem` but this could not introduce an issue.)

drgrice1:
An undefined hash key will be automatically initialized and set to 1 by the `++` operator if not defined.

robert-marik:
Thank you for clear explanation.

Alex-Jordan:
I tested this and it works as described. I also modified the level 2 achievement to give the same item as level 1, and it worked (I then had two of the items).

Alex-Jordan:
I'm going to merge it into 2.16.

PR #1318, opened 2021-04-13
Fix for empty reduced_scoring_date when using certaing items.

robert-marik:
If the course does not use reduced scoring the following message is produced when the user wants to use items which extend the due date.

```Warning -- There may be something wrong with this question. Please inform your instructor including the warning messages below.
```
and
```
    Argument "" isn't numeric in addition (+) at /opt/webwork/webwork2/lib/WeBWorK/AchievementItems.pm line 253.
    Argument "" isn't numeric in numeric gt (>) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 775.
    Argument "" isn't numeric in numeric gt (>) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 775.
    Argument "" isn't numeric in numeric gt (>) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 775.
    Argument "" isn't numeric in numeric gt (>) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 775.
    Argument "" isn't numeric in numeric gt (>) at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 775.
```

I believe that this patch solves the problem and does not introduce another issue.

### DIFF
--- a/lib/WeBWorK/AchievementItems.pm
+++ b/lib/WeBWorK/AchievementItems.pm
@@ -250,7 +250,7 @@ sub use_item {
     my $userSet = $db->getUserSet($userName,$setID);
     
     #add time to the reduced scoring date, due date, and answer date; remove item from inventory
-    $userSet->reduced_scoring_date($set->reduced_scoring_date()+86400) if defined($set->reduced_scoring_date());
+    $userSet->reduced_scoring_date($set->reduced_scoring_date()+86400) if defined($set->reduced_scoring_date()) && $set->reduced_scoring_date();
     $userSet->due_date($set->due_date()+86400);
     $userSet->answer_date($set->answer_date()+86400);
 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Probably something that fixes something which is broken should target `WeBWorK-2.16` right now instead of `develop`. @drgrice1 is managing how the two branches interact and will know for sure.

drgrice1:
If it is desired to get this into the next release, then it should target the WeBWorK-2.16 branch.  Anything that targets the develop branch will be assumed to be meant for a later release.  Of course, we (the release team) can always review these sort of things and determine if they should go into the release.

drdrew42:
The warnings from the first post that come from Utils.pm (L775: `sub after`) seem to come from `is_jitar_problem_closed` (L1532-1606) which calls `after($mergedSet->reduced_scoring_date))` and this is the only call to `after` with `reduced_scoring_date` as an argument.

My assumption is that the `if` statement logic for `is_jitar_problem_closed` should also check `$mergedSet->reduced_scoring_date` before calling `after` (inserted around line 1590 in lib/WeBWorK/Utils.pm).

It also seems that checking `defined($set->reduced_scoring_date)` is effectively moot -- so likely we don't need to keep the defined checks in the `if` logic for these modifications to AchievementEvaluator.pm? Just strip the `defined()` wrapper on each of these instead of adding a second check.

Alex-Jordan (approved):

drdrew42 (approved):

PR #1319, opened 2021-04-13
Use current submission result in achievement calculations

drdrew42:
Thanks to @robert-marik for pointing out that there is an exploitable loophole in the achievement evaluator. I believe this should be considered a priority fix for v2.16.

The oversight here is based on the differences between the `$pg->{state}` and `$pg->{result}` hashes. Essentially, we only want to consider the 'result' of the current submission, rather than the overall 'state' of the problem. By considering the 'state' of a problem, we can **repeatedly** end up with `state->{recorded_score} == 1 && num_correct == 1`. However, we can only achieve `result->{score} == 1 AND num_correct == 1` **once**.

### DIFF
--- a/lib/WeBWorK/AchievementEvaluator.pm
+++ b/lib/WeBWorK/AchievementEvaluator.pm
@@ -88,9 +88,10 @@ sub checkForAchievements {
     # updated $problem with the new results from $pg.  So we cheat and update the 
     # important bits here.  The only thing that gets left behind is last_answer, which is
     # still the previous last answer. 
-    # 
     
-    $problem->status($pg->{state}->{recorded_score});
+    # $pg->{result} reflects the current submission, $pg->{state} holds the best result 
+    # close the unlimited achievement points loophole by only using the current result!
+    $problem->status($pg->{result}->{score});
     $problem->sub_status($pg->{state}->{sub_recorded_score});
     $problem->attempted(1);
     $problem->num_correct($pg->{state}->{num_of_correct_ans});


drgrice1:
You should probably target this for the WeBWorK-2.16 branch.

drgrice1 (approved):
This looks rather straightforward.  The explanation makes sense.  I haven't tested it yet, but it probably doesn't really need that much.

robert-marik:
Thank you for looking at this issue. However I am not sure about this solution. Sometimes we require the score for the current answer as in your patch. This is useful if we check whether it is the first correct answer and we should increment global score. Also useful to decide whether it is a first correct answer after two incorrect answers and we should give third_time.at, for example. But sometimes we really need the saved status. For example if we check that all questions in the set are finished we check the saved state. I think that we should keep both saved status and score of the current submission and in each case decide which is appropriate.

robert-marik:
My suggestion is something like the solution from the attachment. However, needs to check also the other achievements and needs more testing. Among others, I do not use gateway tests so I do not have experiences in this field. I am also not sure if it makes sense to turn achievements on for gateway tests, but it is included in WeBWorK and should not get broken. So I consider this as a temporary patch.
[temporary.txt](https://github.com/openwebwork/webwork2/files/6311610/temporary.txt)

drdrew42:
> But sometimes we really need the saved status. For example if we check that all questions in the set are finished we check the saved state.

Checking whether or not to increment `completeSets` **does** depend on the 'state' of all problems *except* the current one. If we attempt to rely on the 'state' of the current problem, once again, we will end up in a position where a student could repeatedly trigger "completion" of the same set. We must only increment `completeSets` when the set is completed for the _first_ time. That means all problems have been completed except for the current one -- their 'state' has `recorded_score == 1`; and the current problem is 100% for the *first* time -- 'result' has `score == 1` and `num_correct == 1`.

I just don't see any use-case for the 'state' of the current problem.

robert-marik:
Yes, for the current answer we need the score for this submission. However, I was not sure what will be status of the other questions in the set. Sometimes student try to experiment with other ways to enter an answer and the problem may be solved but with incorrect last answer entered later. I supposed that in this case the other problems will be treated as incorrect. I will do more tests.

drdrew42:
> Sometimes student try to experiment with other ways to enter an answer and the problem may be solved but with incorrect last answer entered later.

In such a case, that problem would still have 'state' with `recorded_score == 1` (even though its most recent submission had result: `score < 1`). This is precisely what was causing the original loophole: 'state' holds the *best* score for the problem.

robert-marik:
OK. Thank you for explanation.

Alex-Jordan (approved):

Alex-Jordan:
Oops, I thought this was already merged. I will merge it now since I merged one of @robert-marik 's PRs that rely on this.

PR #1320, opened 2021-04-14
Add an option for setting the application (dvisvgm or pdf2svg) used by the PGtikz.pl macro to generate svg images

drgrice1:
Unfortunately the versions of latex and dvisvgm on Ubuntu 18.04 do not work well for generating svg images with the PGtikz.pl macro.  However, the versions on 20.04 do.  The pdf2svg application works okay in any case.  So this adds a configuration option used by the macro to set the application used by PGtikz.pl to be either dvisvgm or pdf2svg.  The default (set in defaults.config) is pdf2svg.  If your system is running Ubuntu 20.04 you can change this in localOverrides.conf to dvisvgm.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,6 +108,7 @@ RUN apt-get update \
 	apache2 \
 	curl \
 	dvipng \
+	dvisvgm \
 	gcc \
 	libapache2-request-perl \
 	libarchive-zip-perl \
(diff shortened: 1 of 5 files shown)


pstaabp (approved):
This seems perfectly reasonable.

Alex-Jordan (approved):
This is working for me with https://github.com/openwebwork/pg/pull/557.

taniwallach (approved):
Reviewed only the "code" changes. Looks good.

pstaabp (approved):
I wasn't able to test the `pdf2svg` however overall I think this good.

PR #1321, opened 2021-04-14
Consider individual deadlines when evaluating achievements.

robert-marik:
I believe that the achievements "the early bird", "the really early bird" and "last minute math" should take into account the open_time and due_time for the individual student. The current state takes into account the global setting. 

Consequently, the member of the group with shifted due_time may get the message "finished in last 30 minutes" even if there could be still days to his/her due_time.

### DIFF
--- a/lib/WeBWorK/AchievementEvaluator.pm
+++ b/lib/WeBWorK/AchievementEvaluator.pm
@@ -44,7 +44,7 @@ sub checkForAchievements {
     my $cheevoMessage = '';
     my $user_id = $problem->user_id;
     my $set_id = $problem->set_id;
-    our $set = $db->getGlobalSet($problem->set_id);
+    our $set = $db->getMergedSet($user_id,$problem->set_id);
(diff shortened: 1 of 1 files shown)


drdrew42 (approved):

drdrew42:
This is a good catch -- individual settings should be used instead of global ones. This PR will make sure that all sets use the individual problem-set info across all achievement tests.

Alex-Jordan:
The code change looks good. I merged and tested to a point. I did not actually arrange for myself to earn the badges that rely on due date after giving myself a close date that differs from the main set close date. But bringing this in didn't break anything.

drgrice1:
Is this a pull request that is meant to go into the 2.16 release?  It was targeted to the develop branch.  If it is meant to go into the 2.16 release I will merge it in.  Please let me know.

robert-marik:
I used the develop branch following the instructions for developers. If it is possible to merge with 2.16 it would be great.

PR #1322, opened 2021-04-14
Wrap answer previews in `<script type="math/tex"></script>` tags in the attempts table

### LINKED ISSUE #1315: less than sign math issue, possibly relating to MathJax3
In a 2.15 server, this minimal example comes out as expected:

```
BEGIN_TEXT
\(1<x\)
END_TEXT
```

But on my production 2.16 server, the above does not process as desired  in HTML. Basically it comes out in such a way that the web browser is interpreting `<x` as the start of an HTML tag. You see the `\(1` in output.

@dpvc Is this a MathJax3 issue? Or should we edit the math_ev3 routine (or whatever its name is, I forget right now) in WeBWorK to do insert spaces after less than signs?

drgrice1:
This also adds the appropriate MathJax configuration to handle this to the mathjax-config.js file.

This is paired with openwebwork/pg#564.

This fixes #1315.

### DIFF
--- a/htdocs/js/apps/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/apps/MathJaxConfig/mathjax-config.js
@@ -14,6 +14,22 @@ if (!window.MathJax) {
 				}
 				return MathJax.startup.defaultReady()
 			}
+		},
+		options: {
+			renderActions: {
+				findScript: [10, function (doc) {
+					document.querySelectorAll('script[type^="math/tex"]').forEach(function(node) {
+						var display = !!node.type.match(/; *mode=display/);
+						var math = new doc.options.MathItem(node.textContent, doc.inputJax[0], display);
+						var text = document.createTextNode('');
+						node.parentNode.replaceChild(text, node);
+						math.start = {node: text, delim: '', n: 0};
+						math.end = {node: text, delim: '', n: 0};
+						doc.math.push(math);
+					});
+				}, '']
+			}
 		}
+
 	};
 }
--- a/htdocs/js/apps/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/apps/MathJaxConfig/mathjax-config.js
@@ -28,7 +28,8 @@ if (!window.MathJax) {
 						doc.math.push(math);
 					});
 				}, '']
-			}
+			},
+			ignoreHtmlClass: 'tex2jax_ignore'
 		}
 
 	};
--- a/lib/WeBWorK/Utils/AttemptsTable.pm
+++ b/lib/WeBWorK/Utils/AttemptsTable.pm
@@ -362,7 +362,7 @@ sub previewAnswer {
 	} elsif ($displayMode eq "images") {
 		$imgGen->add($tex);
 	} elsif ($displayMode eq "MathJax") {
-		return '\['.$tex.'\]';
+		return '<script type="math/tex; mode=display">' . $tex . '</script>';
 	}
 }
 
@@ -384,7 +384,7 @@ sub previewCorrectAnswer {
 		$imgGen->add($tex);
 		# warn "adding $tex";
 	} elsif ($displayMode eq "MathJax") {
-		return '\['.$tex.'\]';
+		return '<script type="math/tex; mode=display">' . $tex . '</script>';
 	}
 }
 


Alex-Jordan (approved):
These work for me.

There is a bonus improvement. Before when the page first loads, you see the unprocessed math content for a second before MathJax processes. With this, the script tag content is not visible until the math is processed. I think that is an improvement, although it's worth noting that the rest of the text will tend to move more now that there is not unprocessed math serving as a placeholder.

pstaabp (approved):
Looks good and should take care of this.

dpvc (approved):
Looks good (though the issue remains with the plainText and essay output needing to be quoted in `AttemptsTable`).

drgrice1:
> Looks good (though the issue remains with the plainText and essay output needing to be quoted in `AttemptsTable`).

You mentioned that before.  However, those were not dealt with in any special way prior to the MathJax 3 upgrade, and so doing so would be a change from how things were before.  Are you sure those should be dealt with?  What sort of quoting do you mean here?  The script tags certainly won't work in those cases.

drgrice1:
> There is a bonus improvement. Before when the page first loads, you see the unprocessed math content for a second before MathJax processes. With this, the script tag content is not visible until the math is processed. I think that is an improvement, although it's worth noting that the rest of the text will tend to move more now that there is not unprocessed math serving as a placeholder.

I think the flash of unprocessed math was bad.  Even with that there was the possibility of text moving as the unprocessed math would frequently not have the same width as the processed math.

drgrice1:
So something clearly needs to be done about plainText mode.  That is certainly not working.  We need to tell MathJax to skip processing of the math mode delimited text in that case.

@dpvc: Any advice from the resident MathJax expert?  Or are you going to make me sift through the MathJax documentation on this!

drgrice1:
Although, it is not in the AttemptsTable that plainText is not working.  The raw latex equations are not processed by MathJax in that case as they should not be.  It is the latex equations in the problem that aren't working.

drgrice1:
So it seems that we need the MathJax 3 equivalent of the `tex2jax_ignore` class that MathJax 2 heeded.

drgrice1:
Okay, nevermind.  MathJax 3 still has the `tex2jax_ignore` class option.  I will add that.

dpvc:
> those were not dealt with in any special way prior to the MathJax 3 upgrade, and so doing so would be a change from how things were before.

I suspect plainText is rarely used, and essay text is unlikely to include `<` that would look like a tag or `&` that would look like an entity, so it may never have been reported as an issue.

> Are you sure those should be dealt with?

Pretty sure.  You are inserting raw LaTeX into the page, and that can include HTML special characters.  I haven't looked to see if these modes do any special processing to proceed their `preview_latex_string`, but since that is supposed to be LaTeX code, inserting that into the page without quoting HTML special characters could be problematic.  For example, if an essay included `<script>...</script>` then that would be processed as a script.

> What sort of quoting do you mean here?

HTML special character quoting.  Any time that material from the student is being inserted into the page like this, it should be HTML escaped.

> We need to tell MathJax to skip processing of the math mode delimited text in that case.

It seems that yo have found the `tex2jax_ignore` feature that I was going to suggest.

drgrice1:
There is processing done in PGessaymacros.pl in the `essay_cmp` subroutine.  That seems to deal with the issue for essay type answers.  I tested this and tried to inject `<script>document.body.innerHTML = ""</script>` in an essay answer, and it just displays the script tag as is.  The script is not processed (or presumable the page content would have been emptied).

drgrice1:
I also tried inserting `` `1<x` `` and `\(1<x\)` in an essay answer, and those worked as they should.

dpvc:
> There is processing done in `PGessaymacros.pl in the `essay_cmp` subroutine.

As I said, I didn't look.  It turns out that, as you say, `essay_cmp` uses `text2PG()`, which adjusts the result dependent on the `$displayMode`.  So, despite its name, `preview_latex_string` is **not** actually LaTeX when the display mode is not `TeX`.

This is one of the long-time issues with the AnswerHash structure:  there is a text version and a LaTeX version, but no HTML version, and so answer checkers have had to corrupt one or the other (or both) in order to get things to work properly, and so you can't count on these fields to actually contain the correct kind of string.

Sorry for the false alarm, there.

Alex-Jordan:
Perhaps worth noting: when I applied this to my production server, for most users math just stop displaying at all. That was because they had cached copy of `mathjax-config.js` and with that the script tags were not displayed. On that scale, during the term, it was too much to ask everyone including students to clear their cache. I edited the template files to look for a versioned js file, like `mathjax-config.js?1234` and that forced browsers to redownload the file.

drgrice1:
@Alex-Jordan:  I had the same problem.  Adding the url parameter is a good idea for this sort of thing.  At least temporarily when javascript changes.

PR #1324, opened 2021-04-15
Fixing some achievements (third_time, persistence, super_persistence)

robert-marik:
Fix for three achievements. They do not work as advertised. Without this fix it would be possible to earn achievement "Solve a problem on the third submission" by entering correct answer followed by two incorrect answers to the same question. Similarly for other two achievements. 

This should be installed after https://github.com/openwebwork/webwork2/pull/1319

### DIFF
--- a/courses.dist/modelCourse/templates/achievements/third_time.at
+++ b/courses.dist/modelCourse/templates/achievements/third_time.at
@@ -1,6 +1,6 @@
 #This checks to see if the student has a certain number of attempts
 
-    if ($problem->num_correct == 1 and $problem->num_incorrect == 2) {
+    if ($problem->num_correct == 1 and $problem->num_incorrect == 2 and $problem->status == 1) {
 	return 1;
     }    else {
 	return 0;
--- a/courses.dist/modelCourse/templates/achievements/persistance.at
+++ b/courses.dist/modelCourse/templates/achievements/persistance.at
@@ -1,7 +1,7 @@
 #This checks to see if the student has solved the problem after a certain
 # number of incorrect attempts
 
-     if ($problem->status == 1 && $problem->num_incorrect >= 20) {
+     if ($problem->status == 1 && $problem->num_correct == 1 && $problem->num_incorrect >= 20) {
 	return 1
      }    else {
 	return 0;
--- a/courses.dist/modelCourse/templates/achievements/super_persistance.at
+++ b/courses.dist/modelCourse/templates/achievements/super_persistance.at
@@ -1,7 +1,7 @@
 #This checks to see if the student has solved the problem after a certain
 # number of incorrect attempts
 
-     if ($problem->status == 1 && $problem->num_incorrect >= 100) {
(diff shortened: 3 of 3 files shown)


courses.dist/modelCourse/templates/achievements/third_time.at
```diff
@@ -1,6 +1,6 @@
 #This checks to see if the student has a certain number of attempts
 
-    if ($problem->num_correct == 1 and $problem->num_incorrect == 2) {
+    if ($problem->num_correct == 1 and $problem->num_incorrect == 2 and $problem->status == 1) {
```
Alex-Jordan, line 3: An achievement can only be earned once, is that right?

Once #1319 is resolved, if I understand correctly, `$problem->status == 1` implies `$problem->num_correct >= 1`.  And `$problem->num_correct > 1` would imply there was an earlier attempt where `$problem->num_correct == 1` and `$problem->status == 1` and the achievement was already earned.

So I'm wondering if `$problem->num_correct == 1` can be removed from all of these just to simplify the code.

robert-marik: Consider the following scenario
1. Incorrect answer entered.
2. Correct answer entered, problem solved.
3. Incorrect answer entered, just trying to write the solution in another way. Say some algebraic simplifications with typo.
4. A correct answer, fixed the typo in the previous attempt.
This would give the message "Solved on third attempt" which is confusing. Just tested, to be sure.

I observed problems of this type since I advised the students to play with some other forms how to enter answers even if the problem is solved. They try different ways (if they wish) to ensure, that they can enter the formulas in WeBWorK reliably.
In the usual scenario, when the student solves the problem and continues to another problem these things do not matter. But some students are not strong in the ability to code math answers in computer notation and play with the answer fields even if the problem is solved. Just to ensure that on the exam they will not loose too much time by typing formulas.

Alex-Jordan: I see, thank you for that example.

drdrew42 (approved):

Alex-Jordan:
Ugg. So I wasn't paying attention and I see that I merged this into `develop`. Also #1321. What is the right thing to do right now to get these into 2.16 without causing a mess with `develope?

drgrice1:
I will take care of it.

drgrice1:
I will presume that all are intended for the release.

drgrice1:
At least of the ones just merged that is.

Alex-Jordan:
imho they should all go into the release. They fix things that were not working as they should have been. And they are so small scale that there's no need for more testing.

drgrice1:
Yeah, the two branches are synchronized now.

Alex-Jordan:
Thanks. I remembered to change to `WeBWorK-2.16` yesterday with #1317, but forgot about it today.

robert-marik:
Thank you for including to 2.16.

PR #1325, opened 2021-04-15
Fixing Night Owl, Last Minute and Crack o dawn achievements

robert-marik:
Without this fix the student gets achievement if he/she solves the set at any time, returns later after the midnight, submits another answer again, then returns in the morning, submits another answer again and finally returns 30 minutes before the due time and submits any answer again.
This adds the test that the submission is correct and the problem has one correct answer. I hope that this prevents the mentioned problem. I hope that it ensures, that this problem has been solved just now. The rest is usual - checking the time and checking the status of other questions.

### DIFF
--- a/courses.dist/modelCourse/templates/achievements/crack_o_dawn.at
+++ b/courses.dist/modelCourse/templates/achievements/crack_o_dawn.at
@@ -1,4 +1,14 @@
-# checks to see if a studen finished a set between 5AM and 7AM
+# checks to see if a student finished a set between 5AM and 7AM
+
+# Check that this is not a repeated submission of problem which has
+# been solved before.
+# Do not consider this achievement if this is a repeated submission
+# of previously solved problem.
+
+return 0 if $problem->status != 1 or $problem->num_correct != 1;
+
+# now check the time
+
     my @timeData = localtime(time);
 
     if ($timeData[2] < 5 || $timeData[2] > 6) {
--- a/courses.dist/modelCourse/templates/achievements/last_minute.at
+++ b/courses.dist/modelCourse/templates/achievements/last_minute.at
@@ -1,5 +1,12 @@
 # checks to see if a studen finished a set close to the due date
 
+# Check that this is not a repeated submission of problem which has
+# been solved before.
+# Do not consider this achievement if this is a repeated submission
+# of previously solved problem.
+
+return 0 if $problem->status != 1 or $problem->num_correct != 1;
+
     #test to see if it is within 30 min of the due date
     if (($set->due_date - time()) > 1800) {
 	return 0;
--- a/courses.dist/modelCourse/templates/achievements/night_owl.at
+++ b/courses.dist/modelCourse/templates/achievements/night_owl.at
@@ -1,4 +1,12 @@
 # checks to see if a studen finished a set between 12AM and 2AM
+
+# Check that this is not a repeated submission of problem which has
+# been solved before.
+# Do not consider this achievement if this is a repeated submission
+# of previously solved problem.
+
+return 0 if $problem->status != 1 or $problem->num_correct != 1;
+
     my @timeData = localtime(time);
 
     #test to see if it is between midnight and 2am


drdrew42 (approved):

drdrew42:
just a minor comment -- and it is certainly subject to personal taste...

With if statements that impose negated conditions, I prefer to use the perl `unless` rather than `if` `not`
```perl
return 0 unless ( $problem->status == 1 && $problem->num_correct == 1 );
```
instead of
```perl
return 0 if $problem->status != 1 or $problem->num_correct != 1;
```
Personally, I find that it improves readability. Maybe others disagree?

Regardless, this comment is irrelevant to the fact that this PR does what it claims to do -- 'completing' a set means that the current problem has been solved correctly for the _first_ time, and all other problems in the set have `status == 1`.

drgrice1:
> With if statements that impose negated conditions, I prefer to use the perl `unless` rather than `if` `not`

I agree.

Alex-Jordan:
I opened https://github.com/robert-marik/webwork2/pull/2 to reverse the negated condition syntax.

Alex-Jordan:
I am testing this. I got the new `.at` files into a testing course. I assigned a set to myself with one question. I have all of the achievements assigned to myself, but I do not yet have these. It is roughly 9:30 PM Pacific time, so I changed my time zone to America/New_York so it would be 12:30 AM, in the window for Night Owl. I answered the one question correctly, completing the set. If I look at the past answer time I see `04/17/2021 at 12:40am EDT`, so it was in the right time window.

And yet I did not earn the achievement.

Is it possible it is using the server's time instead of the time that is adjusted for time zone?

Alex-Jordan:
I think I confirmed that is the case. It tests the time using the server's time, not the adjusted time considering the time zone in the the configuration. For example it was 10:05 PM here. I edited the code to award the Night Owl between 10 PM and 12 AM. I still have the time zone set to New York where it is 1:05 AM, so not the Night Owl should not be awarded to me. And yet it was. I know this is nothing to do with this PR, which I'm sure is doing what it is supposed to do.


I'm unsure how to access the time zone to shift things appropriately. I would want something like `localtime(time + shiftForLocale)` but after looking a bit I could not find what I need.

Alex-Jordan:
Note that for most cases, the locale matches the server's time zone. But for example with the MAA server, each course should set its own time zone, with courses all over the place using that one server.

robert-marik:
Thank you for good remark. You are right, this should be fixed. I think that the safe container which evaluates the achievement should receive an information about the local time zone of the course (thus the achievement will be capable to convert the timestamp to proper local time). Another possibility is to evaluate the local time inside `AchievementEvaluator.pm` and pass the correct local time in a special variable to the achievements. I will look if I succeed to find how to pass this information to the achievements. Nowadays with distant learning there are also students of the same course in different timezones, but I think that the course should keep the same timezone for the whole course.

If I succeed, should I do pull request to develop or to WeBWorK2.16? When will be 2.16 frozen?

robert-marik:
I think that something like the attached txt patch could be the way. Extract locale info  in the same way as in Util.pm, find the information and pass to the achievement safe container. Then the `*.at` files should be updated in order to use this information. However, did not finish all tests yet. I am attaching this partial work and will continue as time permits (weekend etc.). But it seems that it detects the changes in the course timezeone setting.

I am not comfortable with WeBWorK internals. Maybe this solution does something nasty, maybe I try to dig an information which can be obtained in an easy way. I do not like repeating code from Util.pm related to timezone, but I think that it is necessary to ensure that the timezone has the same value as in the rest of the course. However anybody can suggest a better solution and meanwhile I will work on the final patch.

I think that `shiftForLocale` is not optimal solution because of daylight saving time, necessity to track the shift when trying to find the day of week etc.

BTW: there are discussions about daylight saving time in Europe last two years. How is the daylight saving time
[dateTime.txt](https://github.com/openwebwork/webwork2/files/6328924/dateTime.txt)
 handled in WeBWorK? If the goverment states (for example) that from certain day there will be a daylight saving time for the whole year, will WeBWorK utilize this information?

Alex-Jordan:
Thanks Robert, I will test to help with this. 

>How is the daylight saving time handled in WeBWorK? 

My understanding is that this is what the locale information handles. So for example, I indicate I am in "America/Los_Angeles"
not that I am in "Pacific Daylight Time". I am assuming that somewhere, something knows when this region moves from "Pacific Standard Time" to "Pacific Daylight Time". I believe it is a different date in Canada, and "America/Vancouver" is a different locale for that reason.

If my assumption is correct, it's not something WeBWorK handles. There must be some utility that looks that up on the internet somewhere. Because as you say, the details of the dates do change occasionally.

> I think that shiftForLocale is not optimal solution because of daylight saving time, necessity to track the shift when trying to find the day of week etc.

My first guess would be that as long as you take the result mod 24, it would behave as intended. But yeah, this could be tricky.

Alex-Jordan:
Let me do more testing of this PR as it is, just to confirm it fixes the thing where people earn the award when they don't deserve it. Then I will merge this and we can tackle the other issue separately.

robert-marik:
Thank you. I will continue with the issue related to different course/server timezones and will do another pull request. The attempt in my txt file basically works, but there should be declaration `our @courseDateTime  ...` I I have one other issue. Will commit the full new PR which addresses this today or tomorrow.

mgage:
The DateTime.pm  perl module handles the time zones and the transitions to daylight savings time.

robert-marik:
The PR in WW2.16 branch which tries to handle timezones is https://github.com/openwebwork/webwork2/pull/1331

Alex-Jordan:
Tested and it works. If I do not have Night Owl and I have completed an assignment, and I visit the assignment during the Night Owl time window and submit; then before this PR I am granted Night Owl and after I am not. Also if I complete the assignment for the first time during the Night Owl time window, I am granted the badge.

PR #1328, opened 2021-04-16
Make sure that generated POD documentation files have the html extension

drgrice1:
This is a minor tweak that was suggested by @pstaabp in #1306.


pstaabp:
I just noticed that the link in the footer to the POD doesn't work for my local files.  Looks like the header info for the `-b` switch may not be passed to it.   Again, no big deal since it works on the webwork server.

PR #1329, opened 2021-04-16
Update the other formats with the features added to the simple_format in #1301

### LINKED ISSUE #1335: webwork2/html2xml doesn't use math4-overrides.css
Embedded problems created using `html2xml` don't use `math4-overrides.css`, and any custom css is ignored. I noticed this in the problem library since it appears to now be using this to render the problems when viewed and also confirmed it happens in the few examples I've used embedded problems on other pages.

drgrice1:
@pstaabp observed that there were improvements made to the simple_format in #1301, and suggested that those be added to the other formats as well.  So this does that.

There are also a few tweaks to the simple format as well.  Mainly just adding the answer blank coloration (but only if it is not an answer preview and is a submit or check answers).

This now also adds theme selection to the formats which fixes #1335.

This also now makes it possible to hide any of the submit buttons in the formats if desired.  To hide the buttons add the GET or POST parameters `showPreviewButton=no`, `showCheckAnswersButton=no`, or `showCorrectAnswersButton=no`.  The simple2 format has been removed, as this is all that it does that the simple format does not anyway.

Final addition:   The standard format is deleted, and the debug format is beefed up to replace it.


Alex-Jordan:
This looks good.

taniwallach:
Just to reference @Alex-Jordan's other JSON output - it is the "raw" format from https://github.com/openwebwork/webwork2/pull/1120/

The `json_format.pl` could probably be revised to replace the CSS and JS sections from being HTML code to an array of what files are needed.

lib/WebworkClient/json_format.pl
```diff
@@ -16,18 +16,20 @@
...
+$json_output->{head_part100_VI} = <<'ENDPROBLEMTEMPLATE';
 <!-- CSS Loads -->
 <link rel="stylesheet" type="text/css" href="/webwork2_files/js/vendor/bootstrap/css/bootstrap.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/js/vendor/bootstrap/css/bootstrap-responsive.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/node_modules/jquery-ui-dist/jquery-ui.min.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/node_modules/@fortawesome/fontawesome-free/css/all.min.css"/>
-<link rel="stylesheet" type="text/css" href="/webwork2_files/themes/math4/math4.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/css/knowlstyle.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/js/apps/ImageView/imageview.css"/>
+<link rel="stylesheet" href="$themeDir/math4.css"/>
```
taniwallach, line 27: The lines with `$themeDir` which need to be processed later by variable interpolation need to be in single quotes.
@drgrice1 This is fixed in the PR I a making to your branch.

taniwallach (changes requested):
@drgrice1 - Great stuff. Tested and mostly working.
  1. The formats: `simple`, `debug` and `sticky` all seem to work as expected.
  2. `sticky`'s use of local storage to save the last answers seems fine.However, I do not see colors in the results table for the `stick` format. I think this is worth fixing, unless it requires an inordinate amount of effort. I did not look into why it does not work now.
  3. `json` needs some small changes, and this is also a chance to replace the JS and CSS loading data with arrays of the addresses, rather than a block of full HTML code as recently discussed. I made a PR https://github.com/drgrice1/webwork2/pull/3 to the branch of thus PR with what I think needs to be changed for the `json_format`.

Two more minor comment - which need not hold back merging the improvements:
  * Why use `no` for the new settings like: `showPreviewButton=no`? Maybe at least allow `0` also for no?
  * For `showCorrectAnswersButton` maybe the default should be `no` although that changes the existing default behavior.

lib/WebworkClient/json_format.pl
drgrice1: They don't need to be in single quotes with the current setup.  They need to be in single quotes with the other changes in your pull request though.

drgrice1:
>     2. `sticky`'s use of local storage to save the last answers seems fine.However, I do not see colors in the results table for the `stick` format. I think this is worth fixing, unless it requires an inordinate amount of effort. I did not look into why it does not work now.

I see what you see.  I will look into it.

>     3. `json` needs some small changes, and this is also a chance to replace the JS and CSS loading data with arrays of the addresses, rather than a block of full HTML code as recently discussed. I made a PR [drgrice1#3](https://github.com/drgrice1/webwork2/pull/3) to the branch of thus PR with what I think needs to be changed for the `json_format`.

Your pull request looks good.  I will merge it.  More could be done here.  I think that instead of using prefixed keys for the hidden inputs like `hidden_input_filed_sourceFilePath`, it would be nicer to use a sub object.  So one key `hidden_input_fields` that is itself an object, and each of the inputs a sub key.  It takes a little more work to set up, but is better structurally.

>     * Why use `no` for the new settings like: `showPreviewButton=no`? Maybe at least allow `0` also for no?

I am fine with using '0' instead (I think).  I would rather not allow multiple settings as I don't want to overly complicate the code in FormatRenderedProblem.pm.

>     * For `showCorrectAnswersButton` maybe the default should be `no` although that changes the existing default behavior.

I would rather not change the existing default behavior unless there is a consensus to do so.

drgrice1:
@taniwallach:  I have addressed all of the things you mentioned except the default behavior for the show correct button.  I am fine with making that change, but I will let others weigh in on that idea first.

@Alex-Jordan:  You probably use the html2xml formats more than most others.  What do you think of defaulting to not showing the "Show Correct Answers" button.

taniwallach (approved):

taniwallach:
@pstaabp - I think this is ready to go, except the possible change to the default. Could you take a look?

lib/WebworkClient/simple_format.pl
```diff
@@ -11,9 +11,11 @@
 <link rel="stylesheet" type="text/css" href="/webwork2_files/js/vendor/bootstrap/css/bootstrap-responsive.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/node_modules/jquery-ui-dist/jquery-ui.min.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/node_modules/@fortawesome/fontawesome-free/css/all.min.css"/>
-<link rel="stylesheet" type="text/css" href="/webwork2_files/themes/math4/math4.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/css/knowlstyle.css"/>
 <link rel="stylesheet" type="text/css" href="/webwork2_files/js/apps/ImageView/imageview.css"/>
+<link rel="stylesheet" href="$themeDir/math4.css"/>
```
Alex-Jordan, line 16: Something about this is not working. Here is the AIM server, now using 2.16. For normal use, things appear to be OK. But for using the simple format, `$themeDir` is resolving as the empty string, as you can see in the source of the page below. So the CSS and js is not loading. Here is an example page:

https://webwork-ptx.aimath.org/webwork2/html2xml?courseID=anonymous&userID=anonymous&password=anonymous&course_password=anonymous&answersSubmitted=0&displayMode=MathJax&outputformat=simple&problemSeed=6&problemSource=RE9DVU1FTlQoKTsKbG9hZE1hY3JvcygiUEdzdGFuZGFyZC5wbCIsIk1hdGhPYmplY3RzLnBsIiwiUEdNTC5wbCIsIkFuc3dlckZvcm1hdEhlbHAucGwiLCJQR2NvdXJzZS5wbCIsKTtURVhUKGJlZ2lucHJvYmxlbSgpKTsKQ29udGV4dCgnTnVtZXJpYycpOyRhID0gcmFuZG9tKDEsIDksIDEpOwokYiA9IHJhbmRvbSgxLCA5LCAxKTsKJGMgPSBDb21wdXRlKCRhICsgJGIpOwoKQkVHSU5fUEdNTApDb21wdXRlIHRoZSBzdW0gb2YgW2BbJGFdYF0gYW5kIFtgWyRiXVx0ZXh0ezp9YF0KCltgWyRhXSArIFskYl0gPWBdIFtfXXskY317NX0gW0BBbnN3ZXJGb3JtYXRIZWxwKCdudW1iZXJzJylAXSoKCgpFTkRfUEdNTAoKQkVHSU5fUEdNTF9TT0xVVElPTgpbYFskYV0gKyBbJGJdID0gWyRjXVx0ZXh0ey59YF0KCgpFTkRfUEdNTF9TT0xVVElPTgoKRU5ERE9DVU1FTlQoKTs%3D

I've searched through the config files and I can't find something that should fix this. `$themeDir` is defined with ` my $themeDir = "$ce->{webworkURLs}{htdocs}/themes/$theme";` in `FormatRenderedProblem.pm`. And in `defaults.conf`, I find `$webworkURLs{htdocs}        = "$webwork_htdocs_url";`. And in `site.conf` I find `$webwork_htdocs_url  = "/webwork2_files";`. But I have trouble seeing how `$themeDir` resolves as the empty string.

Alex-Jordan: I copied the line
```        my $themeDir = "$ce->{webworkURLs}{htdocs}/themes/$theme";```
in `FormatRenderedProblem.pl` down to immediately before the interpolation at line 357, and now things work. So something is not working with its original location at line 130.

**Comment by drgrice1** [comment]
I am not able to reproduce this.  Even going to the link that you posted seems to work.  This may be because you have made the change you mention on your server.  Although, I also can't see how making that change would fix anything.  That variable depends on the `$ce` variable which is defined above line 130.

**Comment by Alex-Jordan** [comment]
Yes, it fixed once I copied line 130 to line 357, but that is all I did.

I will use a different server to try to cleanly reproduce it.

**Comment by Alex-Jordan** [comment]
I'm not able to reproduce it on the next server I tried. And I deleted the line I pasted in the AIM server, restarted apache, cleared cache, and I can't reproduce it there now either. 

Note that the issue as I understood it is not about `$ce`. Even if `$ce` were undefined, the line
```
my $themeDir = "$ce->{webworkURLs}{htdocs}/themes/$theme";
```
should leave `$themeDir` as a nonempty string with "themes" inside it. But what was happening was in the HTML source for the page like that URL I posted, instead of lines like
```
<link rel="stylesheet" href="/webwork2_files/themes/math4/math4.css"/>
```
I was getting lines like
```
<link rel="stylesheet" href="/math4.css"/>
```

OK, well nevermind then. I'm never sure if caching somewhere is messing with me. If it comes up again, I will carefully try to make it happen in a reproducible way on some other development server.

drgrice1:
There is certainly something odd that occurred.

Issue #1323, opened 2021-04-15
URLs are not properly generated for a set named "0" in function generateURLs from Utils.pm


somiaj:
Hello,

I have installed a new server using the develop branch and have noticed the following when I was testing it out.

Emails sent using the 'Email Instructor' button from a problem in a homework set do not fully keep track of which problem the email was sent from. The link `$emailalbleURL` sent with the email only links to the user who sent the problem, and no longer links to the problem from which the email was sent from. In addition clicking `cancel email` or `return to your work` returns to the homework list and not the problem from which the email was sent from. The links I see are just `https://server/webwork2/Course/?effectiveUser=practice2` and no longer contain info about the problem the user was working on.

The email details contain the info about which set and problem the email was sent from, but this information is not being recorded in the links like I am use to on my other server.

somiaj:
Okay, I think I've tracked down the actual issue, and for the most part it is working. When I was testing this, I decided to use the default `set0.def` to send an email from. It appears that if the set name is `0` that the links don't get created properly. But if I used a set with an actual name, here I used `setMAAtutorial`, the links all generate as expected. So there is some issue in generating the links of the set name is `0`.

I also tested this using a set named `1`, and that worked fine as well. So it seems this issue only affects any set that has the name `0`.

I have updated the issue title.

drgrice1:
This is now fixed.

PR #1330, opened 2021-04-16
Fix the feedback urls for set0.  This only affects set0. 

drgrice1:
Very specific fix!  This fixes issue #1323.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -1764,7 +1764,7 @@ sub generateURLs {
 	if ($userName) {
 		my $modulePath;
 		my @args;
-		if ($params{set_id}) {
+		if (defined $params{set_id} && $params{set_id} ne "") {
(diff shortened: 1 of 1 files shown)


pstaabp:
I guess this is about as specific as it gets.   Funny that I'm sure this has existed for 20+ years and this is the first time.  I did a quick check for 
```
if ($params{set_id})
```
in the codebase, just to make sure there aren't other places like this.  This might be a good opportunity to double check this.

drgrice1:
Yeah, it certainly isn't a new issue.  It is difficult to do a thorough check as set_id is also used as setID, or setId, or setName, etc.

drdrew42 (approved):
😂

Alex-Jordan:
I tested this by manually applying the same edit to my production server. I don't have email running on other ones. It worked (of course). I tested sending emails from set 0, set 0 problem 1, some other set, some other set problem 1. They all came back with the right links.

PR #1331, opened 2021-04-17
Pass information on the course timezone to the achievements

robert-marik:
This tries to solve the problem mentioned by @Alex-Jordan in https://github.com/openwebwork/webwork2/pull/1325
The patch allows to pass information on the current time to the achievement. This allows to properly handle achievements with courses in different time zones on the same server.

Tested as follows:
1. create new course, new users, new short homeworks with few simple questions, assign achievements to the users.
2. solve the hw
3. change the timezone in confiuguration of the course to the timezone with time just after the midnight.
4. solve with another user or solve another set. You should get "Night owl"
5. change timezone to the land where the current time is between 5 and 7 in the morning
6. solve another hw or solve another set. You should get the "Crack of dawn"

When you merge, the code here contains changes did in https://github.com/openwebwork/webwork2/pull/1325 again. Not sure if it will merge automatically.

Hope it will work. When I played and tested, I observed some problems. But these problems were probably caused by the fact that I repeatedly deleted and added homeworks and users with names 1,2,3, ... and this does not remove the rows in the past_answers table. When I manually deleted old answers and did the tests again, everything  behaved as expected.

### DIFF
--- a/courses.dist/modelCourse/templates/achievements/crack_o_dawn.at
+++ b/courses.dist/modelCourse/templates/achievements/crack_o_dawn.at
@@ -1,7 +1,15 @@
-# checks to see if a studen finished a set between 5AM and 7AM
-    my @timeData = localtime(time);
+# checks to see if a student finished a set between 5AM and 7AM
 
-    if ($timeData[2] < 5 || $timeData[2] > 6) {
+# Check that this is not a repeated submission of problem which has
+# been solved before.
+# Do not consider this achievement if this is a repeated submission
+# of previously solved problem.
+
+return 0 unless ($problem->status == 1 && $problem->num_correct == 1);
+
+# now check the time
+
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
I merged #1325 and then there were a few conflicts to resolve here. But I did that and now this is ready to test.

Alex-Jordan (approved):
Changed my time zone to America/New_York so I could earn Night Owl, and then to Atlantic/Reykjavik so I could earn Crack of Dawn. Worked as expected.

I will let others review as well before merging this one since it edits `AchievementEvaluator.pm` and not just `.at`~s.

drdrew42 (approved):

taniwallach (approved):
Tested and could get the 2 different achievements by adjusting the course timezone.
Note: If this the course was created from a model course with old versions of the achievement files, they will need to be updated to test this.

Issue #1164, opened 2020-12-15
Localization of time broken in the set editor


robert-marik:
WeBWorK allows time localization, but the set editor assumes time in English format when parsing start time, due time etc.

Steps to reproduce the problem:

1. set  `$siteDefaults{locale}="cs_CZ";`  in site.conf file, restart apache.
2. create hw set which starts in afternoon, say at 3:00pm
3. if you open the set again (to edit problem, for example), the start time is `...(date).. at 03:00odp.`, where "odp." is localization for "pm"
4. after saving the changes, WeBWorK assumes English date format, see the file Utils.pm where you can find
```
sub unformatDateAndTime {
        my ($string) = @_;
        my $orgString = $string;

        $string =~ s|^\s+||;
        $string =~ s|\s+$||;
        $string =~ s|at| at |i; ## OK if forget to enter spaces or use wrong case
        $string =~ s|AM| AM|i;  ## OK if forget to enter spaces or use wrong case
        $string =~ s|PM| PM|i;  ## OK if forget to enter spaces or use wrong case
        $string =~ s|,| at |;   ## start translating old form of date/time to new form

        # case where the at is missing: MM/DD/YYYY at HH:MM AMPM ZONE
        unformatDateAndTime_error($orgString, "The 'at' appears to be missing.")
                if $string =~ m|^\s*[\/\d]+\s+[:\d]+|;

        my ($date, $at, $time, $AMPM, $TZ) = split /\s+/, $string;
```
and then (few lines below)
```
	# default value for $AMPM
	$AMPM = "AM" unless defined $AMPM;
```
whis is followed by 
```
	$pm = 0;
	$pm = 12 if ($AMPM =~/PM/ and $hour < 12);
	$hour += $pm;
	$hour = 0 if ($AMPM =~/AM/ and $hour == 12);
```
5. as a consequence of bad parsing, the time is silently changed from 3:00pm to 3:00am

In other words, the AM and PM strings are hardwired to WeBWorK. Could be resolved in https://github.com/openwebwork/webwork2/pull/1333

robert-marik:
Fixed in #1615

robert-marik:
closing

PR #1332, opened 2021-04-17
Use en_US format whenever date and time are intended to be in a form fields

robert-marik:
This solves https://github.com/openwebwork/webwork2/issues/1164

Explanation: WeBWorK assumes on some places the date and time in the format "%m/%d/%Y at %I:%M%P" and the parser expects the date and the time in this format. However, if the course uses non-english locales, the %P string gets its local value (for example "odp." stands for "pm" in cs_CZ) and the parser fails. In this case WeBWorK is not able to parse his own string back.

As a consequence: If I open the editor for HW with due_time at 5:00pm, the value in the form field changes to 5:00am. If I change something (say give different point values to problems) and save without fixing the time, the due_time is shifted from 5pm to 5am. This is annoying if the lecturer intends to open HW for 90 minutes in the afternoon as an exam. I think that two possible solutions are either to improve the parser or to force date and time to be in en_US whenewer required. I followed the second possibility, since this seems to be easier and there is lower risk that something gets broken. This is also compatible with date picker which is used to enter dates and times.


robert-marik:
Oops, should go to 2.16. Closing.

drgrice1:
@robert-marik:  You don't need to close it.  You can just change the target.

robert-marik:
Thank you. Still learning Git. I already issued another PR. I am very sorry for causing inconvenience for developers. Will remember for future. Thank you again.

PR #1333, opened 2021-04-17
Use en_US format whenever date and time are intended to be in form fields

robert-marik:
This solves #1164

Explanation: WeBWorK assumes on some places the date and time in the format "%m/%d/%Y at %I:%M%P" and the parser expects the date and the time in this format. However, if the course uses non-english locales, the %P string gets its local value (for example "odp." stands for "pm" in cs_CZ) and the parser fails. **In this case WeBWorK is not able to parse his own string back.**

As a consequence: If I open the editor for HW with due_time at 5:00pm, the value in the form field is in cs_CZ 5:00odp. which is parsed back as 5:00am. (see #1164 for the corresponding code pieces). If I change something (say give different point values to problems) and save without fixing the time, the due_time is shifted from 5pm to 5am. This is annoying if the lecturer intends to open HW for 90 minutes in the afternoon as an exam. I think that two possible solutions are either to improve the parser or to force date and time to be in en_US whenewer required. I followed the second possibility, since this seems to be easier and there is lower risk that something gets broken. This is also compatible with date picker which is used to enter dates and times. Finally, it is compatible with the fact that WeBWorK supposes to separate the day and time by English "at" clause.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -608,9 +608,9 @@ sub FieldHTML {
 	$userValue = (defined($userValue)) ? ($labels{$userValue // ""} || $userValue) : $blankfield; # this allows for a label if value is 0
 
 	if ($field =~ /_date/) {
-		$globalValue = $self->formatDateTime($globalValue,'','%m/%d/%Y at %I:%M%P') if defined $globalValue && $globalValue ne $labels{""};
+		$globalValue = $self->formatDateTime($globalValue,'','%m/%d/%Y at %I:%M%P', 'en_US') if defined $globalValue && $globalValue ne $labels{""};
 		# this is still fragile, but the check for blank (as opposed to 0) $userValue seems to prevent errors when no user has been assigned.
-		$userValue = $self->formatDateTime($userValue,'','%m/%d/%Y at %I:%M%P') if defined $userValue && $userValue =~/\S/ && $userValue ne $labels{""};
+		$userValue = $self->formatDateTime($userValue,'','%m/%d/%Y at %I:%M%P', 'en_US') if defined $userValue && $userValue =~/\S/ && $userValue ne $labels{""};
 	}
 
 	if ( defined($properties{convertby}) && $properties{convertby} ) {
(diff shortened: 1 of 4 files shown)


Alex-Jordan:
I looked at the code here. I wondered if:

- the format string could pass through `maketext`. Like `maketext("[_1] at [_2]", '%m/%d/%Y', '%I:%M%P')`
- `formatDateTime` could default to use the server's locale when no locale is passed to it as an argument.

Alex-Jordan:
That is probably not a solution to the issue since it does not touch on the `unformatDateAndTime` routine. Better parsing there would be good, as you allude to.

Alex-Jordan:
It does not feel right to center English and/or US date format (MM/DD/YYYY versus DD/MM/YYYY). Apparently the only countries that use the MM/DD/YYYY system are the US, the Philippines, Palau, Canada, and Micronesia.

But this seems like a really hard nut to crack. Your solution fixes something, at the cost of digging deeper with the English/US date format.

What if we merge this approach for 2.16, but work on some better solution for 2.17? In the absence of some universal date format parser, we could slowly add parser code that is case-wise base on locale and timezone. For example, with locale "cs_CZ" it could recognize "odp". Time zone (for example "America/El_Salvador") might matter as well in some cases. It would just take effort from international contributors to not revert to defaults, which might be US defaults or some international-minded defaults.

robert-marik:
Thank you for looking at this. 
I do not know, how many users are affected. Sure that not all non-US speaking. A colleague in my country keeps locale in en_US and she does not have problems like this. In fact, setting locale to cs_CZ is required only if you want to properly show names of days in the hw list. Without this the user is happy to use en_US locale and the local language of the course.

When we will look for a solution we should be aware also of the following.

-  Locale and language setting for the course need not to be the same, see above.
- It would be necessary to patch date picker as well.
- Finally, as you wrote, there are also problems with MM/DD/YYYY versus DD/MM/YYYY format.

I think that if we do not want to dig the en_US date format deeper (which sounds as a good point), we can just wait for the feedback from other users. I am sure that there are many non-US users and they may help to find the best approach. **So we need not to merge now, we can just keep track here and wait. This will prevent that something may get broken.**

I suggested to keep the English date format since I am not a big fan of localized software. With English software we can just copy error message to google and we immediately find solution to most issues. With localized software the first step is guessing the original English phrase :) . Also using tutorials is impossible if the names of buttons and links are localized. So this is just few words to explain my approach to this issue. I did not translate the admin part of WeBWorK to Czech since for me it is easier to work with English version and this is compatible with the suggestion to keep English date format.

So we can wait for the feedback from other users from other countries.

robert-marik:
Should I rebase this PR to devel? Or close?

Alex-Jordan:
I think leave it as it is for at least a little while. This edit would do no harm (I think?) and does make something behave better. That issue where 5:00pm turns to 5:00am is worth fixing. Maybe @taniwallach and others with internationalization concerns will weigh in.

taniwallach:
I think we should consider this in the future, so the internal value is as universal as possible, but the displayed value can be nicely localized. I think we should hold off on this, at least until after WW 2.16.

PR #1337, opened 2021-04-20
Fix typo in CourseIntegrityCheck.pm

dlglin:
The warning message was using the wrong variable name.


drgrice1 (approved):
This looks good.  I will go ahead and merge it, since it is rather trivial.

drgrice1:
I assume that you meant this to go into 2.16.  So I will also merge it to there.

PR #1351, opened 2021-04-29
Keep simple.conf file in English

robert-marik:
This PR solves issue https://github.com/openwebwork/webwork2/issues/1077
The problem is that localized phrases for professor, student, true, false can be written to simple.conf file when saving configuration. However, WeBWorK is not capable to read localized strings and the file is messed.

The problems is in pull down menus. If they are on the current tab of configuration editor, the user sees translated phrase in pull down menu. The value field is in English and the corresponding part of the simple.conf file is correct. 
However, if the pull down menu is NOT on the active tab, the translated phrase is written int simple.conf file and remains there even if the language is changed. 

I believe that the correct approach is not to write translated strings to the conf file. I also believe, that the value in the pull down menu should be the same as the value written in the config file.  Therefore it is sufficient to remove `maketext` commands.

I was not able to rebase commit https://github.com/robert-marik/webwork2/commit/939fe2b1d86c54011c3dd28b709efb7ae4040ce4 and therefore made another one.

Only few languages (Hebrew, French, Hungarian) translated these strings and only these languages are affected.
This issue is somewhat related to https://github.com/openwebwork/webwork2/pull/1333 where similar situation appears for date/time string: WeBWorK is not capable to parse nonEnglish text which may appear if locales is not en_US.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
@@ -193,8 +193,8 @@ package configboolean;
 sub display_value {
   my ($self, $val) = @_;
   my $r = $self->{Module}->r;
-  return $r->maketext('True') if $val;
-  return $r->maketext('False');
+  return 'True' if $val;
+  return 'False';
 }
 
 sub save_string {
@@ -204,14 +204,14 @@ sub save_string {
 	my $newval = $self->convert_newval_source($newvalsource);
 	my $displayoldval = $self->comparison_value($oldval);
 	return '' if($displayoldval eq $newval);
-	return('$'. $varname . " = " . ($newval eq $r->maketext("True") ? 1 : 0) .";\n");
+	return('$'. $varname . " = " . ($newval eq "True" ? 1 : 0) .";\n");
 }
 
 sub entry_widget {
 	my ($self, $name, $default) = @_;
 	my $r = $self->{Module}->r;
-	my $true = $r->maketext('True');
-	my $false = $r->maketext('False');
+	my $true = 'True';
+	my $false = 'False';
 	return CGI::popup_menu(
 		-name => $name,
 		-default => ($default ? $true: $false),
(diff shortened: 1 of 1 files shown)


drgrice1 (commented):
There is often little effort put into making things in webwork work nicely for instructors.  This philosophy is one of the things that is seriously holding webwork back from being adopted at many institutions.

Unfortunately this pull request is directly in line with that philosophy.  It is a quick hack to make things work, rather than to fix the actual problem.  It would be better to fix the handling of translations than to just remove them.  This can be done by properly using labels and values of html inputs.

As such, I can not approve this pull request.

drgrice1:
There is often little effort put into making things in webwork work nicely for instructors.  This philosophy is one of the things that is seriously holding webwork back from being adopted at many institutions.

Unfortunately this pull request is directly in line with that philosophy.  It is a quick hack to make things work, rather than to fix the actual problem.  It would be better to fix the handling of translations than to just remove them.  This can be done by properly using labels and values of html inputs.

As such, I can not approve this pull request.

drgrice1:
Oops.  Didn't mean to close it though.

drgrice1:
@robert-marik: Test #1352 and see if that fixes this issue without dropping translations.

robert-marik:
Thank you. I was not able to fix things in other way than removing translations. And since I was asked to try to resolve this old issue, I choosed the approach presented here. 

Confirm that #1352 works great, then translated roles are not written in the config file. Feel free to close this one. Now or when #1352 will be merged.

robert-marik:
> There is often little effort put into making things in webwork work nicely for instructors. This philosophy is one of the things that is seriously holding webwork back from being adopted at many institutions.
> 

Agree. I can help to resolve some of them. 

> Unfortunately this pull request is directly in line with that philosophy. It is a quick hack to make things work, rather than to fix the actual problem. It would be better to fix the handling of translations than to just remove them. This can be done by properly using labels and values of html inputs.
> 
> As such, I can not approve this pull request.

I like your approach that you not only reject my ideas, but provide a better solution. Thank you.

Anyway, as a user of localized software I think that the options should be kept in English. If someone follows tutorial to do some steps, this tutorial will be probably in English. It is in most cases better to stay at English setting. However it is not important here. The admin may prepare his course in English and switch to the local language later. There are more important issues such as missing support for multiple plurals or missing support for localization of messages from pg  library.

taniwallach:
@robert-marik, @drgrice1  - Thanks!
Idea for the future - a setting to allow the admin interfaces to be used in `en_us` when the main course language is not English.
There are several other foreign language users who commented that due to the dearth of local documentation, managing the course is better handled in English, at least in the short term.
I put in an issue to consider that in the future: https://github.com/openwebwork/webwork2/issues/1354

PR #1352, opened 2021-04-30
Make sure that translations are not saved to simple.conf when editing the course configuration

drgrice1:
When an instructor makes changes to the course configuration the translated display values should not be used in any way in that process.  This completely separates the underlying values of the configuration objects from what is displayed by using labels and values for the html inputs.

This replaces #1351 and fixes https://github.com/openwebwork/webwork2/issues/1077.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -190,6 +190,7 @@ $hardcopyTheme = "twoColumn";
 # Achievements
 ################################################################################
 $achievementsEnabled = 0;
+$achievementItemsEnabled = 0;
 $achievementPointsPerProblem = 5;
 $achievementPreambleFile = "preamble.at";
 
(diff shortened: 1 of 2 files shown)


robert-marik:
Does the job for me.  

If I change roles in configuration and then switch language to heb and save configuration two times, the translated roles are written in config file. Seems to better approach than #1351 .  #1351 Should be closed after including this one.

taniwallach (approved):
Fixes issue https://github.com/openwebwork/webwork2/issues/1077. I tested several types of setting changes, and noting triggered the type of bug reported in that issue.
The variable name changes, certainly make the code more readable.
@drgrice1 and @robert-marik Thanks for working on the issue and coding fixes.

PR #1355, opened 2021-04-30
Add a confirmation dialog when a timed gateway quiz is clicked on on the homework sets page

drgrice1:
This is rather quickly hacked in and is only a partial solution.  This will not work when entering a quiz directly via LTI authentication (or some other external link).


Alex-Jordan:
I have a quiz that is not yet open. While acting as a student, now I can click on that not-yet-open quiz and I get the confirmation dialog. Of course it leads to an error if I click to proceed to the quiz.

drgrice1:
Okay.  I will look into it.  I didn't try that case.

drgrice1:
@Alex-Jordan:  The closed set issue is resolved now.  I also made it so that the dialog is not shown when acting as a student.

Alex-Jordan (approved):

Alex-Jordan:
Just confirming that it works as expected.

drdrew42 (approved):

pstaabp (approved):
Nice feature.

dlglin (approved):
Looks good.  Thanks!

Alex-Jordan:
One thing I just noticed about this while testing something else.

Say I start a timed quiz version. Then, while this version has not yet timed out, I find myself back on the main sets page. I click the parent quiz template. This is not actually starting a new version, it is (appropriately) directing me to my version that is already running. But the confirmation dialog still pops up with some misleading statements:

'You will have 1 minute to complete the quiz. Click "Begin" below to start.'

drgrice1:
This will be a challenging one to fix.  It will be hard to determine if a quiz version is in progress in ProblemSets.pm at the time that the gateway quiz template is displayed.  I will see what I can do.

drgrice1:
I am not sure about your misleading comment though.  Is your gateway quiz set to be completed in 1 minute?  When I try this the dialog shows the total amount of time that the quiz must be completed in every time, even when a quiz is in progress.

Alex-Jordan:
Yes, it's set to 1 minute. It's a test quiz that I have that I want to run out of time quickly.

The "misleading" part is that it is suggesting I am about to start a new version, when actually I'm about to be sent to an older version that maybe has much of its time already used up.

I doubt this will cause lots of students any trouble because they should be staying in their quiz after starting it the first time, or clicking the "version 1" link if they are trying to get back to it. But it was a quirky thing I ran into when I was trying to trigger multiple versions for testing purposes.

drgrice1:
My plan is to disable the dialog if a quiz is in progress so that the quiz is just re-entered.  So the misleading comment won't be an issue.  I just have to figure out how to get the needed information to determine that there is a quiz in progress.  I will figure it out.

PR #1358, opened 2021-05-05
Use student id for password

### LINKED ISSUE #1356: Default password using user_id instead of student_id
On a server using the 2.16 branch I have noticed that if I add a student using the Classlist Editor, the default password is being set to the user_id and not the student_id. (If I add a group of users just a class .lst file, student_id is used).

drgrice1:
The student id should be used for the password by default when adding a user from the classlist editor. This was unintentionally changed in #1290.

This fixes #1356.


drdrew42 (approved):

taniwallach (approved):
This reverts the setting to what it was in the past, and is an urgent fix to the WeBWorK-2.16 and and develop branches.
https://github.com/openwebwork/webwork2/blob/03eca4cc1a6034d24cf28bd690f0f56ce14f284b/lib/WeBWorK/ContentGenerator/Instructor/AddUsers.pm

PR #1359, opened 2021-05-07
fix ptx format answers

Alex-Jordan:
I think this was a mistake introduced in #1301. After making this change, what I do with ptx output on a 2.16 server is working again.


drgrice1 (commented):
Yep.  That is a mistake.  Thanks for cleaning up for me.

PR #1362, opened 2021-05-14
Don't try to show the student nav in gateway quizzes when the set is invalid

drgrice1:
Currently an error occurs if you attempt to open a new gateway quiz that is closed when acting as a student.  This fixes that and lets you see the page that says, "The selected problem set (Gateway_Past_Due) is not a valid set for tguy:  This set is closed. No new set versions may be taken."

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1166,7 +1166,7 @@ sub nav {
 	if ($r->authz->hasPermissions($user, "become_student") && $effectiveUser ne $user) {
 		my $setName = $self->{set}->set_id;
 
-		return "" if $setName eq "Undefined_Set";
+		return "" if $setName eq "Undefined_Set" || $self->{invalidSet};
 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
What I found was that for a quiz that had past its close date, it was already the case that I would be taken to the "...is not a valid set..." page. But in the cases I tested, the student did already have versions. You say "new" here, so maybe it was only an error screen when there are no existing versions already?

For a quiz that had not yet opened at all, I would get the error screen before this patch, and after the patch I get the "...is not a valid set..." page.

Alex-Jordan (approved):

drgrice1:
Yes, this occurs for me on a quiz that is past due, but the student had not taken any versions of.  I see now that it also happens for a quiz that has not opened yet.

drgrice1:
This pull request fixes that case also!

Alex-Jordan:
> This pull request fixes that case also!

Yes, that is what I meant to be reporting. I think it's all good and I approve. It's clearly a good edit to the code that is so small, I am tempted to just merge it.

PR #1364, opened 2021-05-14
Use svg WeBWorK logo and tweak masthead

Alex-Jordan:
I wanted to change the logo to be a cleaner looking SVG image.

While I was in there, I tweaked some styling details with the masthead. To see the differences in behavior, try narrowing the size of your browser window before this PR. You should see the MAA logo shrink as you approach the critical width where it transitions to small-screen styling. Also you will see the login information start to wrap for no real good reason.

After the PR, in addition to the logo being an SVG image, neither of those things should happen if you are at default viewing zoom level on a browser.

### DIFF
--- a/htdocs/images/webwork_logo.svg
+++ b/htdocs/images/webwork_logo.svg
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created by manually editing webwork_square.svg -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   width="80"
+   height="65"
+   id="svg2"
+   version="1.1">
+  <metadata
+     id="metadata7">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <g
+     id="layer1"
+     transform="translate(-137.21391,-210.45965)">
+    <g
+       id="g5899">
+      <g
+         id="g5972"
+         transform="translate(-61.042075,-18.507438)">
+        <g
+           transform="translate(-0.11138,1.2987676)"
+           id="g5816">
+          <g
+             id="g5922">
+            <g
+               id="g5932" />
+            <g
+               id="g5961">
+              <g
+                 id="g5951">
+                <path
+                   id="path4218"
+                   d="m 224.2781,233.62123 c 0.42645,5.38525 0.53172,10.78991 -2.46322,14.62117 -4.41017,5.55354 -16.72809,10.29041 -16.66061,10.12704 0,0 11.43376,4.24683 15.51724,9.31035 4.08349,5.06352 8.16696,17.80399 8.16696,17.80399 0,0 3.17736,-10.35947 12.02941,-13.60571 8.11384,-2.97553 27.77228,0.82982 27.77228,0.82982 0,0 -8.93054,-6.17148 -11.87065,-12.86839 -2.94011,-6.69692 -1.30671,-22.3775 -1.30671,-22.3775 0,0 -8.65699,3.92015 -15.19057,4.41016 -6.53357,0.49002 -15.99413,-8.25093 -15.99413,-8.25093 z"
+                   style="fill:#ffffcc;fill-opacity:1;stroke:#000000;stroke-width:1;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:5;stroke-opacity:1;stroke-dasharray:none;marker-start:none;marker-mid:none;marker-end:none" />
+                <path
+                   style="fill:none;stroke:#000000;stroke-width:1px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1"
+                   d="m 206.21826,258.27235 29.97049,-1.5363"
+                   id="path4220"/>
+                <path
+                   style="fill:none;stroke:#000000;stroke-width:1px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1"
+                   d="m 224.26497,233.37852 11.27042,23.19419"
+                   id="path4222"/>
+                <path
+                   style="fill:none;stroke:#000000;stroke-width:1px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1"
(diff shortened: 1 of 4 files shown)


htdocs/themes/math4/system.template
```diff
@@ -80,12 +80,12 @@
...
 	<div class="span2 webwork_logo">
-		<a href="<!--#url type="webwork" name="root"-->"><img src="<!--#url type="webwork" name="htdocs"-->/images/webwork_transparent_logo_small.png" alt="to courses page" />WeBWorK</a>
+		<a href="<!--#url type="webwork" name="root"-->"><img src="<!--#url type="webwork" name="htdocs"-->/images/webwork_logo.svg" alt="to courses page" />WeBWorK</a>
 	</div>
-	<div class="span4 maa_logo">
+	<div class="span6 maa_logo">
 		<a href="http://www.maa.org"><img src="<!--#url type="webwork" name="htdocs"-->/images/maa_logo_small.png" alt="to MAA (Mathematical Association of America) main web site" /></a>
 	</div>
-	<div id="loginstatus" class="offset3 span3">
+	<div id="loginstatus" class="offset3 span4">
```
drgrice1, line 88: Bootstraps row-fluid system is designed to have the span and offset parts add up to 12.  With your changes it now adds to 15.  That breaks bootstraps design.  You hack over this with the float in the math4.css file, but that isn't the right way to do it.  Remove the `offset3` class and drop the `margin-left` and `float` styles on the `#loginstatus` and you will achieve the same effect, but done the right way for bootstrap's system.

drgrice1:
Another related thing that I have been thinking about implementing is to show the actual user's name instead of the user id in the masthead.  In many cases the user id is not meaningful.  What do you thing of that?

htdocs/themes/math4/system.template
Alex-Jordan: OK, thanks. I was definitely inferring incorrectly how bootstrap worked without understanding it. Maybe with this understanding I can make it even better.

Alex-Jordan:
I had some thoughts to change that content too. I held back since it's a separate issue from what this PR tries to accomplish. 

User name sounds good. I might also remove the "Logged in as" and the period. And I'm on the fence about having both the "Log Out" text and the icon at the same time. Maybe just one is sufficient.

A related thought I've had is to move the User Settings access to over there. When we were looking at removing navigation like in #1224, there was a need to keep access to User Settings, and therefore take up real estate with the navigation panel. But if it were accessible through that area of the masthead, that need goes away.

drgrice1:
I was thinking it might be nice to include the WeBWorK text in the svg image with the logo.  Currently the masthead is too thick (in my opinion), and the reason for that is that the text goes onto the next line when that span becomes to narrow.  This happens for browser window widths anywhere between 768 and 1216 pixels.  If the text were in the logo, then the logo would scale to the width, but the text would stay on the same line.  Thus resulting in a sleek thin masthead.

Although, when the window width gets close to 768 pixels the image would probably scale down perhaps too small.

Alex-Jordan:
I can put the text inside the SVG to the right of the spider web. There are issues as you predicted. If that div width narrows because of resizing your browser, the entire image will shrink, shrinking the text and the spider web too. Maybe there is some way to style the text within the SVG so that it will rescale while the spider web stays the same size. But I'm unsure how to do that.

Here are screen shots. I'm happy with it and do not mind the resizing.

Screen Shot 2021-05-14 at 3 56 51 PM
Screen Shot 2021-05-14 at 3 57 10 PM
Screen Shot 2021-05-14 at 3 57 21 PM

Based on the screenshots, do you have any design requests? I could put them in before I push.

I have plans after this to attack the MAA logo. I want to generalize it to be a generic place for the host institution (which might be MAA) to put its logo. I know you can already do this with localoverrides.js, but the internals still call it the "maa_logo".

drgrice1:
Yeah, it was just a thought.  I will let you decide if that idea works.  It am not sure if the text can be made to shrink while the rest of the image stays the same.

Alex-Jordan:
I'm fine with this. If you don't have any comments on styling based on the screenshots, I'll clean up, commit, and push that for review.

drgrice1:
Looks good in the screenshots.

Alex-Jordan:
OK, this is ready to review.

Alex-Jordan:
OK, I'm done again. I increased the height of the svg's container div to match the height that is needed to fit two logout buttons. Then I tweaked the font size and position of "WeBWorK" in the svg.

drgrice1 (approved):
This is looking pretty good now.  Once we upgrade to Bootstrap 5 this can be done better.  Bootstrap 2.3.2 uses such old css techniques.

pstaabp (approved):
Looks good.  Nice improvement.  I'm guessing PR #1365 will replace the logo in this one.

Alex-Jordan:
This an #1365 are separate. This one gives the svg WeBWorK logo in the top right of the masthead, and #1365 is only about the favicon in your browser tabs. But if you are like me and were recently switching branches between these and didn't fully flush the cache, the effect of one of these PRs may have appeared in your browser while you were actually on the other branch.

drdrew42:
It's really throwing me off to see WeBWorK in all-caps in the screenshots above... yet the code doesn't seem to suggest this behavior - maybe it was changed after the screenshots?

htdocs/images/webwork_logo.svg
```diff
@@ -0,0 +1,87 @@
...
+                   style="fill:none;stroke:#000000;stroke-width:1px;stroke-linecap:butt;stroke-linejoin:bevel;stroke-opacity:1"
+                   d="m 228.02178,240.89213 c 0.21306,1.67416 2.07245,6.77696 -0.72048,10.98053 -3.01023,4.53061 -6.95647,5.68009 -6.95647,5.68009 0,0 2.77677,0.98003 6.53357,4.73684 3.75681,3.7568 4.61999,8.90361 4.61999,8.90361 0,0 3.96475,-4.03192 7.73116,-5.31984 5.51947,-1.88738 14.56431,-0.44999 14.56431,-0.44999 0,0 -1.05352,-0.75601 -5.73429,-7.29732 -3.62703,-5.06874 -1.09688,-10.27745 -1.09688,-10.27745 0,0 -4.53348,0.64703 -10.15092,0.15041 -5.21371,-0.46093 -8.78999,-7.10688 -8.78999,-7.10688 z"
+                   id="path4230"/>
+              </g>
+            </g>
+          </g>
+        </g>
+        <text
+           xml:space="preserve"
+           style="font-size:30px;font-style:normal;font-variant:small-caps;font-weight:500;font-stretch:normal;line-height:125%;letter-spacing:0px;word-spacing:0px;fill:#FFFFFF;fill-opacity:1;stroke:none;font-family:'Helvetica Neue', Helvetica, Arial, sans-serif"
```
Alex-Jordan, line 80: @drdrew42 The small-caps font variant is in the code of the SVG image here.

Alex-Jordan:
The small caps is set inside the svg. But you are correct that some things changed since those screen shots, aminly from Glenn's additional commit.

pstaabp:
> This an #1365 are separate. This one gives the svg WeBWorK logo in the top right of the masthead, and #1365 is only about the favicon in your browser tabs. But if you are like me and were recently switching branches between these and didn't fully flush the cache, the effect of one of these PRs may have appeared in your browser while you were actually on the other branch.

Got it.  I wasn't paying enough attention.

PR #1365, opened 2021-05-14
sharpen favicon

Alex-Jordan:
I took the SVG from #1364, thickened its lines, and then used ImageMagick to convert that to a new favicon.ico. 

Here is a screenshot with the current favicon, then one where I thickened the lines in the SVG  to stroke-width 2, then one where I went to stroke-width 3.

Screen Shot 2021-05-14 at 1 49 47 PM

This PR has the version with stroke width 3. Should I go with 2? Or something else like 2.5?


drgrice1:
I think I vote for the one with stroke-width 2.  It seems to look rather dark with the sroke-width of 3.

Alex-Jordan:
I changed to the one based on width 2 and pushed. Here is a screenshot from Firefox where you can see it with the tab both in and out of focus.

Screen Shot 2021-05-14 at 3 11 08 PM

drgrice1:
Looks good to me.

drgrice1 (approved):
Much better favicon than before.  It looks crisp.

pstaabp:
I think this looks good.  @Alex-Jordan, can you fancy up the MAA logo now.  :) It looks so 1990s.

Alex-Jordan:
> can you fancy up the MAA logo now. 

Heh, no. But I do have an intention to replace that with a generic image you set in a configuration, and it should represent the host institution. So could be MAA (until they cease hosting) but for most of us it would be your school's logo image.

Alex-Jordan:
Oh I see this was about favicon. Sorry, my comment is about the MAA logo in the masthead, which is also GeoCities-esque.

drdrew42 (approved):
looks great!

PR #1366, opened 2021-05-15
style the tabs in course configuration page

Alex-Jordan:
This styles the tabs in the Course Config page.

Really, these should be buttons with the form for all the areas fully present, and there are no links to other pages to get to a different subsection of the config options. That is a heavier project.

Here is what I styled it to be like, going for minimalism. If you hover over one of the other tabs, that tab turns a darker gray.

Screen Shot 2021-05-14 at 5 48 57 PM

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1745,7 +1745,7 @@ $ConfigValues = [
 		  type => 'permission'
 		},
 	],
-	['PG - Problem Display/Answer Checking',
+	['Problem Display/Answer Checking',
 		{ var => 'pg{displayModes}',
 		  doc => 'List of display modes made available to students',
 		  doc2 => '<p>When viewing a problem, users may choose different methods of rendering formulas via an options box in the left panel. Here, you can adjust what display modes are listed.</p><p>Some display modes require other software to be installed on the server. Be sure to check that all display modes selected here work from your server.</p><p>The display modes are <ul><li> plainText: shows the raw LaTeX srings for formulas.<li> images: produces images using the external programs LaTeX and dvipng.<li> MathJax: a successor to jsMath, uses javascript to place render mathematics.</ul></p></p>You must use at least one display mode. If you select only one, then the options box will not give a choice of modes (since there will only be one active).</p>',
--- a/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
@@ -499,13 +499,15 @@ sub print_navigation_tabs {
 	my $str = '';
 	for my $tab (0..(scalar(@tab_names)-1)) {
 		if($current_tab eq "tab$tab") {
-			$tab_names[$tab] = $r->maketext($tab_names[$tab]);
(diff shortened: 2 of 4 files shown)


lib/WeBWorK/ContentGenerator/Instructor/Config.pm
```diff
@@ -675,7 +677,6 @@ sub body {
 	$tabnumber =~ s/tab//;
 	my @configSectionArray = @{$ConfigValues->[$tabnumber]};
 	my $configTitle = shift @configSectionArray;
-	print CGI::p(CGI::div({-align=>'center'}, CGI::b($r->maketext($configTitle))));
```
drgrice1, line 678: I haven't looked closely at this, but you deleted this line, and don't seem to have replaced it with anything.  Was this intentional?  What did it do?

drgrice1 (commented):
Looks nicer than before.

lib/WeBWorK/ContentGenerator/Instructor/Config.pm
Alex-Jordan: > Was this intentional? What did it do?

Yes, it was intentional. That line printed text saying which section you were currently in, like "Optional Modules". It is redundant once the styling makes that obvious. But I was just thinking if something should be done for accessibility. I'm not sure how a screen reader user would directly know which section they are in. I think that moving to buttons with the form all on one page would help with that. But short of doing that larger project, does anything occur to you? (Aside from bringing back the redundant text?)

drgrice1: I see.  I am not sure how to make this more accessible other than putting the line back at this point.  It would be nice to move to a tabbed setup like in the homework sets editor and class list editor and such though.  I assume that is what you mean.  That is a bigger project.

Alex-Jordan:
I think I may change the `h1` from "Course Configuration" to be like "Course Configuration: Permissions". It may be more appropriate anyway to make the `h1` change when the page reloads. Does that raise any flags?

Alex-Jordan:
Editing the `h1` would be more difficult than I expected. I put back the text, but made it an `h2`. Now it looks like this, where my mouse was hovering over the Email tab:

Screen Shot 2021-05-14 at 8 59 35 PM

drgrice1 (approved):
This looks good.

lib/WeBWorK/ContentGenerator/Instructor/Config.pm
pstaabp: I think you can add aria attributes to the tabs to help indicate what is going on, but it would be better to set up as tabs.  Push that to the bootstrap update in 2.17.

pstaabp (approved):
Overall, looks better.

conf/defaults.config
```diff
@@ -1745,7 +1745,7 @@ $ConfigValues = [
 		  type => 'permission'
 		},
 	],
-	['PG - Problem Display/Answer Checking',
+	['Problem Display/Answer Checking',
```
drdrew42, line 1748: 🙏  🙏  🙏  🙏  🙏  🙏

drdrew42 (approved):
so grateful to drop the "PG - " from that configuration!

Issue #1363, opened 2021-05-14
Warning when submitting answers when $permissionLevels{dont_log_past_answers} is set.


somiaj:
I was testing out some problems as my admin account and when I try to submit an answer to a problem I get an error.

```
WeBWorK Warnings

WeBWorK has encountered warnings while processing your request. If this occured when viewing a problem,
it was likely caused by an error or ambiguity in that problem. Otherwise, it may indicate a problem with the
WeBWorK system itself. If you are a student, report these warnings to your professor to have them corrected.
If you are a professor, please consult the warning output below for more information.

Warning messages

    Use of uninitialized value in concatenation (.) or string at
    /home/webwork/webwork2/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm line 186.
```

In testing I only get this warning when submitting a answers as a professor or admin. Submitting answers as a ta or lower doesn't give me this warning. Here I'm unsure if the goal is to not allow profs/admins to store answers in the database or not, but in either case I don't think getting a warning should happen.

Also it appears there is a spelling error with `occured` in that output.

drgrice1:
Professors and admins are allowed to submit answers.  How are you getting this warning?  You will need to be more specific as to how to reproduce this.

somiaj:
I get this warning when I submit answers to any question as my admin user. I do get feedback if the answer was correct or not, but it is not saved to the database.

drgrice1:
Is this on WeBWorK 2.16?  I am not seeing this.  If I submit answers as an admin user, there is no error, and the answers are stored in the database.  This is true for all versions of WeBWorK in fact.  No error and answers are stored.

somiaj:
Okay, I think I have tracked down the issue. I went and compared my `localOverrides.conf` to `localOverrides.conf.dist` and noticed that I uncommitted the following line: `$permissionLevels{dont_log_past_answers} = "professor";`

If I go and comment out that line, I no longer get the warning. Maybe I misunderstood what that did, I was thinking it would log past answers in the past answer database (since I do lots of testing figured I didn't need to log my past answers), but thought it would at least keep the most current `last_answer`.

taniwallach:
Fixed by https://github.com/openwebwork/webwork2/pull/1367

PR #1367, opened 2021-05-15
Fix the dont_log_past_answers permission.

drgrice1:
When a user has this permission the create_ans_str_from_responses subroutine of ProblemUtil.pm still needs to be called for other things that are done in the process_and_log_answer subroutine.

This fixes issue #1363.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
@@ -83,36 +83,26 @@ sub process_and_log_answer{
     my ($encoded_last_answer_string, $scores2, $isEssay2);
 	my $scoreRecordedMessage = "";
 
-	if ( defined($answer_log ) and defined($pureProblem)) {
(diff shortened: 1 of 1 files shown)


somiaj:
I have tested this patch and am no longer getting warnings when submitting answers with `permissionLevels{dont_log_past_answers} = "professor";` set. It also appears that the intended design of `permissionLevels{dont_log_past_answers} = "professor";` is to still put the previous answer in `last_answer`, but not to log previous last answers. So provided that is the design it is working as intended.

Though a bit beyond this, I was wanting an option that would not populate `last_answer` or just disable the submit button altogether. The reason I initially set this option was when writing problems I didn't want to have my previous submitted answers appearing when I was checking other random versions of a problem. With `last_answer` being recorded the previous answer still populates problems when viewing them from the problem editor interface.

drgrice1:
@somiaj:  I am rather certain that the way that this pull request sets things up is as intended.  The `dont_log_past_answers` permission name seems to indicate that.  Others can weigh in as to if that permission should also not save the `last_answer`.

drdrew42 (approved):

PR #1374, opened 2021-05-21
Adjust ADD_JS_FILE handling code to deal with new third parameter

drgrice1:
Adjust the code for dealing with the files added by PG.pl's ADD_JS_FILE method to add attributes provided in the new optional third argument.

The code is also adjusted to ensure that scripts are loaded in the order that they are requested, instead of in the random order that perl gives with hash keys.  Should this be done with the ADD_CSS_FILE handling code also?  Order matters for css loading as well.

This is paired with https://github.com/openwebwork/pg/pull/576.


taniwallach (commented):
Code looks fine, but it would be nice if someone tested in all 3 contexts: homework, gateway, and html2xml, and verified that order is maintained, and that the 3-rd option is working for all 3 contexts.

PR #1375, opened 2021-05-21
Fix datepicker icon in set import form on the hmwk sets editor page

drgrice1:
Fix the icon on the "shift dates" input of the import sets form. Currently the button is detached from the input instead of appended to
the end.  I think I caused this in an earlier pull request.

Also remove the inline javascript that adds the datepicker and move it into the datepicker.js file, and remove the DatePickerScripts.pm module, and move the datepicker css into a separate file.


pstaabp:
Is this separated?  This image is the current branch. 

image

drgrice1:
Yes.  It should look the same here as it does on the set detail page.

Here is what it looks like in Firefox on my system.  Of course this will vary with the browser and system.
Screenshot from 2021-05-26 05-59-32

drgrice1:
Here is what it looks like in Firefox on my system without this pull request for comparison:
Screenshot from 2021-05-26 06-03-17

pstaabp (approved):
looks good.

drdrew42 (approved):

PR #1376, opened 2021-05-21
switch SQL::Abstract to SQL::Abstract::Classic

### LINKED ISSUE #1236: Database breaks with update to version 2.000001 of SQL::Abstract
There's a big discussion thread on slack about this, but in short, version  2.000001 of [SQL::Abstract](https://metacpan.org/pod/SQL::Abstract) breaks the database.  I haven't figured out exactly how, but after upgrading to this version, there was  a `[table missing]` error.  Could it have been vaguer? 

I'll try to circle back to this with a small script illustrating the error, but want to file this before I forget.  Hopefully if we can figure out how it affect us, we could file a bug report with SQL::Abstract. 

This could become a problem if many standard distribution pull in this version when updating perl modules. I got this version using cpanm to update.  I think we do need to put some documentation out there (perhaps with the ww2.16 release) about this problem.  

Also, I was able to roll back using by downloading version 1.87 from [metacpan site of SQL::Abstract](https://metacpan.org/pod/SQL::Abstract) and then using `cpanm cpanm SQL-Abstract-1.87.tar.gz`

pstaabp:
This fixes #1236 by switching the package.   

The package `SQL::Abstract::Classic` may need to be installed to test and then make sure things that hit the database still work.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -105,7 +105,7 @@ my @modulesList = qw(
 	Scalar::Util
 	SOAP::Lite
 	Socket
-	SQL::Abstract
+	SQL::Abstract::Classic
 	Statistics::R::IO
 	String::ShellQuote
 	Template
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -1,8 +1,44 @@
 #!/usr/bin/env perl
 #
+
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2021 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+=head1 NAME
+
+check_modules.pl - check to ensure that all applications and perl modules are installed. 
+
+=head1 SYNOPSIS
+ 
+check_modules.pl [options]
+ 
+ Options:
+   -a|--apache-version	 Which apache version to use.  Defaults to 2. 
+   -m|--modules          Lists the perl modules needed to be installed. 
+   -p|--programs       	 Lists the programs/applications that are needed. 
+   -A|--all         		 checks both programs and modules (Default if -m or -p is not selected)
+
+=head1 DESCRIPTION
+ 
+Lists all needed applications for webwork as well as a perl modules.
+ 
+=cut
 use strict;
 use warnings;
 use version;
+use Getopt::Long qw(:config bundling);
 
 my @applicationsList = qw(
 	convert
@@ -105,7 +141,6 @@ my @modulesList = qw(
 	Scalar::Util
 	SOAP::Lite
 	Socket
-	SQL::Abstract::Classic
 	Statistics::R::IO
 	String::ShellQuote
 	Template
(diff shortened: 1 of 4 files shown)


drgrice1:
The only thing that I don't like about this is that now we are forcing Ubuntu users to need to install SQL::Abstract::Classic from CPAN when previously a distribution package sufficed.  That package worked perfectly well as things were before.  Is there a way that this could be conditional?  If the SQL::Abstract::Classic package is available then use that, but if not fall back to trying to use SQL::Abstract.

pstaabp:
This sounds like a decent idea.  Looks like Ubuntu 20.04 is on version 1.86 of `SQL::Abstract` so that's fine.  Can we do something like

```
if ($SQL::Abstract::VERSION>1.87) {
  use SQL::Abstract::Classic;
} else {
  use SQL::Abstract;
}
```

does it need to be in a `BEGIN` block?

pstaabp:
It doesn't appear a switch like this works.  Looks like on compilation it looks within all branches.   I've looked at `Module::Load`, but that doesn't appear to work with `use base`.  Also, looked at `eval` with:
```
if ($SQL::Abstract::VERSION>1.87) {
	eval 'use base SQL::Abstract::Classic' or die($@);
} else {
	eval 'use base SQL::Abstract' or die($@);
}
```
but getting errors with this too.  

This seems to be more trouble than it's worth unless someone else has a good idea.

drgrice1:
@pstaabp:  I put in a pull request to your branch that fixes the detection.

pstaabp:
Thanks for fixing.  I've got this working on my develop machine (MacOS) and tested with both SQL::Abstract version 1.87 and 2.00001 as well as ubuntu 20.04 using the standard packages (no Classic version).

taniwallach:
Can `check_modules.pl` do a conditional test based on the version of `SQL::Abstract` it finds. Probably special code is needed in the main routing, but it should make the report more useful.

The changes to the 2 comments should probably refer to A or B, as we seem to now allow either option.

pstaabp:
> Can `check_modules.pl` do a conditional test based on the version of `SQL::Abstract` it finds. Probably special code is needed in the main routing, but it should make the report more useful.
> 
> The changes to the 2 comments should probably refer to A or B, as we seem to now allow either option.

good idea.  I'll add that.

pstaabp:
Updated `check_modules.pl` and had a version with some more modern `Getopt::Long` and POD documentation that I included.

bin/check_modules.pl
```diff
@@ -207,4 +244,27 @@ sub check_modules {
...
+	print "\n checking for SQL::Abstract\n\n";
+	eval "use SQL::Abstract";
+	my $sql_abstract = not($@); 
+	my $sql_abstract_version = $SQL::Abstract::VERSION if $sql_abstract; 
+
+	eval "use SQL::Abstract::Classic";
+	my $sql_abstract_classic = not($@);
+
+	if($sql_abstract_classic) {
+		print "You have SQL::Abstract::Classic installed.  This package will be used.\n";
```
taniwallach, line 262: If I understand the code in `lib/WeBWorK/DB/Utils/SQLAbstractIdentTrans.pm` correctly, `SQL::Abstract::Classic` will only be used when either `SQL::Abstract` is not installed, or is and have a version `> 1.87`. Thus the message should be

> You have SQL::Abstract::Classic installed.  This package will be used if either the installed version of `SQL::Abstract` is version `> 1.87` or if that package is not installed.

Maybe making the test more parallel to the login in `lib/WeBWorK/DB/Utils/SQLAbstractIdentTrans.pm` would make sense.

bin/check_modules.pl
```diff
@@ -207,4 +244,27 @@ sub check_modules {
...
+	} elsif ($sql_abstract && $sql_abstract_version <= 1.87) {
+		print "You have version $sql_abstract_version of SQL::Abstract installed.  This will be used";
+	} else {
+		print "You need either SQL::Abstract version 1.87 or SQL::Abstract::Classic installed.\n";
```
taniwallach, line 266: I think the message needs to have `<=` added, so it will be
```
You need either SQL::Abstract version <= 1.87 or SQL::Abstract::Classic installed.
```

taniwallach (commented):
Looks fine in terms of the code, but I did not have a chance to test it. Some small comments.

taniwallach (approved):

bin/check_modules.pl
```diff
@@ -207,4 +247,29 @@ sub check_modules {
...
+	my $sql_abstract_version = $SQL::Abstract::VERSION if $sql_abstract; 
+
+	eval "use SQL::Abstract::Classic";
+	my $sql_abstract_classic = not($@);
+
+	if($sql_abstract_classic) {
+		print qq/ You have SQL::Abstract::Classic installed. This package will be used if either 
+ the installed version of SQL::Abstract is version > 1.87 or if that package is not installed.\n/;
+	} elsif ($sql_abstract && $sql_abstract_version <= 1.87) {
+		print "You have version $sql_abstract_version of SQL::Abstract installed.  This will be used";
```
drgrice1, line 268: A new line might be nice here.

drgrice1: That is in the output.  Not in the code.

pstaabp: I'm getting a new line in the output at 266.  Is this what you mean?

drgrice1: Line 268 does not have a new line in the output.

drgrice1 (approved):
This looks good.

PR #1377, opened 2021-05-21
Make output_JS in ProblemSets.pm return the empty string.  Oops.

drgrice1:
When I added the output_JS method to ProblemSets.pm I forgot to have it return the empty string.  So now you get 1's where they shouldn't be.


pstaabp:
Looks fine.  Okay to merge?

drgrice1:
Currently if you open the problem sets page (the first page you see when you log into a course), there is a 1 at the top of the page above the masthead.  This removes that 1.

This shouldn't need a lot of consideration I should think!  It is a simple one line change.

PR #1382, opened 2021-05-27
Fix tab order hack for scaffolds

drgrice1:
This fixes a bug with the tab order hack in which a tooltip on an element within the scaffold causes the scaffold to also close.  This bug was seen in a problem with a graphTool element inside a scaffold.

Another way to fix this is to upgrade to bootstrap 4 or newer.  Those versions of bootstrap properly deal with the tab order, and remove the need for this hack.

Make the scaffold javascript more precise on the selector used to find the accordion body elements that belong to the scaffold.  Also make the event handlers only act if the target is actually the accordion body, and not an element within.  In other words, ignore events that bubble up from within.

### DIFF
--- a/htdocs/js/apps/Scaffold/scaffold.js
+++ b/htdocs/js/apps/Scaffold/scaffold.js
@@ -1,6 +1,7 @@
 window.addEventListener('DOMContentLoaded', function() {
-	var sections = $('.section-div .collapse');
-	sections.on('show', function() {
+	var sections = $('.section-div > .accordion-body.collapse');
+	sections.on('show', function(e) {
+		if (e.target != this) return;
 		this.style.display = 'block';
 
 		// Reflow MathQuill answer boxes contained in the section so that their contents are rendered correctly.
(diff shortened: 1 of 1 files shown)


taniwallach:
Could you provide a small sample problem where the bug occurs which the patch fixes?

drgrice1:
I have attached a problem that demonstrates the bug.
[gt-in-scaffold.zip](https://github.com/openwebwork/webwork2/files/6556609/gt-in-scaffold.zip)

drgrice1:
To see the bug in action, open the third scaffold section in the previously attached problem and move the mouse over one of the tool buttons at the bottom of the graph, and then move the mouse away.  As soon as the mouse leaves the button the tooltip will close and so will the scaffold.  You won't be able to reopen the scaffold either (unless you utilize the browser dev tools to remove the `display: none` style from the accordion body div).

Alex-Jordan:
This fixes another issue as well. (Or maybe it is the same issue, just a different face of it.)

I was looking at nested scaffolding (to work on numbering them). I found that if section 1 starts open, and subsection 1 also starts open, and I click to close subsection 1, then everything closes and cannot be opened back up with clicking.

Alex-Jordan (approved):

drgrice1:
Yes, this would also fix that issue.

pstaabp (approved):
Looks like this fixes the bug. !

taniwallach (approved):
Tested and fixes the bug in the sample problem.

PR #1384, opened 2021-05-28
use CSS to control the grader comment field width

Alex-Jordan:
The comment field in the grader is great and I've been using it a lot. I find that its default 30x3 size is almost always too small for what I want to write though.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -1,5 +1,5 @@
 /* WeBWorK Online Homework Delivery System
- * Copyright � 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
+ * Copyright &copy; 2000-2021 The WeBWorK Project, http://openwebwork.sf.net/
  * $CVSHeader: webwork2/conf/templates/math2/math2.css,v 1.1.2.1 2008/06/24 17:17:36 gage Exp $
  * 
  * This program is free software; you can redistribute it and/or modify it under
@@ -738,7 +738,7 @@ input.correct[type="text"]{ /* green */
     outline: thin dotted \9;
     /* IE6-9 */
     
-    -webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(81,153,81, ,.6);
+    -webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(81,153,81,.6);
     -moz-box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(81,153,81,.6);
     box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(81,153,81,.6);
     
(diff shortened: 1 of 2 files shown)


drdrew42 (approved):
Count me in.

drgrice1 (changes requested):
I agree that the comment box is to narrow.  I copied the number of cols from the original problem grader, and have also noticed that it could use a wider setting.  However, 70 seems to be to much in many cases.  Here is a screenshot:
Screenshot from 2021-05-28 20-53-33

Notice that the comment box extends to far to the right ... past the horizontal rules above and below the grader section.  It even extends beyond the containing div element.  The browser window width in this screen shot is not that narrow.  It is at 1200px.

Furthermore, due to the extensive width of the comment box the other elements of the problem grader are forced to wrap to the left.

I suggest using css styling to achieve a width that is more responsive to the browser window size, rather than using the textarea `cols` attribute.

drgrice1:
@Alex-Jordan:  I submitted a pull request to your branch with suggested css tweaks.

Alex-Jordan:
Thanks @drgrice1 , I merged your PR.

With CSS in control, is there any reason to do anything more with the `@cols` value? Any reason to lower it some from 70?

drgrice1:
The `cols` attribute doesn't matter much now.  The css styling overrides that.  You can leave it at 70.

I am still not entirely content with the problem grader appearance, but this is good in a pinch.  The real problem is a design flaw in the html markup to begin with.  I followed suite on the sort of thing that is done in current WeBWorK markup and used and html table.  I really should have used `div`'s and Bootstrap's form layout system to obtain a better responsive setup.  That is something that we will definitely need to change to generally for ww 3.  Tables are for data, not for forms or layout in general.  Part of the problem is that Bootstrap 2.3.2 doesn't do this well either.

drgrice1:
What do you think of the height of `100px` that I chose?  I also feel that the default height was not enough, but also don't want the height to be too much.  I chose the height of the textarea that Github uses initially for its comment boxes (the one that I am currently typing in).  Feel free to adjust that, or remove the added css height if you think it is too much.

Alex-Jordan:
This is an [interesting idea](https://codepen.io/shshaw/pen/c72208b279d750f1ab46b6a45bbd04d7). It uses javascript so I would be reluctant to try it. For starters, I am unsure where to put javascript for something like this. But also I just don't feel competent using javascript.

I think that 100px is fine for the height. But at default zoom etc on my browser that fits five lines of text with some left over. I can take it down to 87px, and at 86px is when five lines starts to need a vertical scroll bar. So you could get away with something like 90px and not lose any lines of text.

Alex-Jordan:
> I can take it down to 87px, and at 86px is when five lines starts to need a vertical scroll bar.

I take this back. Before I was changing things in the DOM with the inspector, but now that I load this branch and try things as they are, 100px is just right for 5 lines.

Alex-Jordan:
This is ready for anyone more to review and/or merge.

drgrice1:
> This is an [interesting idea](https://codepen.io/shshaw/pen/c72208b279d750f1ab46b6a45bbd04d7). It uses javascript so I would be reluctant to try it. For starters, I am unsure where to put javascript for something like this. But also I just don't feel competent using javascript.

We could try something like that if we want to be fancy.  If done right, things like that can be nice.  I think for now the simple css will suffice though.

drgrice1 (approved):
This looks good.

pstaabp (approved):
Looks good to me.  I'll merge this in.

PR #1385, opened 2021-05-30
Sticky student nav

Alex-Jordan:
This makes the problem and user navigation buttons sticky to the top of the screen. So for example if you have scrolled down to the bottom of a problem or quiz, you can move to the next student or problem without scrolling back up.

### DIFF
--- a/htdocs/themes/math4/system.template
+++ b/htdocs/themes/math4/system.template
@@ -90,6 +90,27 @@
 	</div>
 </div>
 
+<!-- Breadcrumb -->
+<div id="breadcrumb-row" class="row-fluid">
+	<div class="span12">
+		<a href="#" class="btn btn-primary btn" id="toggle-sidebar">
+			<span id="toggle-sidebar-icon">
+				<i class="icon fas fa-chevron-left" aria-hidden="true" data-alt="close sidebar"></i>
+			</span>
+		</a>
+		<div id="breadcrumb-navigation" role="navigation" aria-label="breadcrumb navigation">
+			<ul class="breadcrumb">
+				<!--#path style="bootstrap" text="<li><span aria-hidden='true' class='divider'>/</span></li>"-->
+			</ul>
+		</div>
+	</div>
+</div>
+
+<!-- Run breadcrumb js since it gets loaded early -->
+<script type="text/javascript">
+	$('#toggle-sidebar').removeClass('btn-primary')
+</script>
+
 <div id="body-row" class="row-fluid">
 
 <!-- Navigation -->
@@ -131,25 +152,13 @@
 
 <!-- Main Content Area -->
 <div id="content" class="span10" >
-	<!-- Breadcrumb -->
-	<div id="breadcrumb-row" class="row-fluid">
-	  <div class="span12">
-	    <a href="#" class="btn btn-primary btn" id="toggle-sidebar">
-			<span id="toggle-sidebar-icon">
-				<i class="icon fas fa-chevron-left" aria-hidden="true" data-alt="close sidebar"></i>
-			</span>
-	    </a>
-	    <div id="breadcrumb-navigation" role="navigation" aria-label="breadcrumb navigation">
-	      <ul class="breadcrumb">
-			<!--#path style="bootstrap" text="<li><span aria-hidden='true' class='divider'>/</span></li>"-->
-	      </ul>
-	    </div>
-	</div></div>
-	<!-- Run breadcrumb js since it gets loaded early -->
-	<script type="text/javascript"> 
-	  $('#toggle-sidebar').removeClass('btn-primary')
-	</script>
-	
+	<!--#if can="nav"-->
+	<div class="problem-nav row-fluid sticky-nav" role="navigation" aria-label="problem navigation">
+		<!-- Question Navigation, e.g.: Prev, Up, Next for homeworks -->
+		<!--#nav style="buttons" imageprefix="/webwork2_files/images/nav" imagesuffix=".gif" separator="  "-->
+	</div>
+	<!--#endif-->
+
 	<!-- Page Title -->
 	<!--#if can="title"-->
 	<div class="row-fluid"><div class="span12">
@@ -161,13 +170,6 @@
 	<div id="Message" class="Message row-fluid">
 		<!-- Message for the user -->
 		<!--#message-->
-	</div>
-	<!--#endif-->
-
-	<!--#if can="nav"--> 
-	<div class="problem-nav row-fluid sticky-nav" role="navigation" aria-label="problem navigation">
-		<!-- Question Navigation, e.g.: Prev, Up, Next for homeworks -->
-		<!--#nav style="buttons" imageprefix="/webwork2_files/images/nav" imagesuffix=".gif" separator="  "-->
 	</div>
 	<!--#endif-->
 
--- a/htdocs/themes/math4/system.template
+++ b/htdocs/themes/math4/system.template
@@ -93,7 +93,7 @@
 <!-- Breadcrumb -->
 <div id="breadcrumb-row" class="row-fluid">
 	<div class="span12">
-		<a href="#" class="btn btn-primary btn" id="toggle-sidebar">
+		<a href="#" class="btn" id="toggle-sidebar">
 			<span id="toggle-sidebar-icon">
 				<i class="icon fas fa-chevron-left" aria-hidden="true" data-alt="close sidebar"></i>
 			</span>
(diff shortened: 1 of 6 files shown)


pstaabp:
Is there ever a case where there will be more than one of these buttons on page?  If so, will this result in problems? 

This seems like a relatively new  (past few years) css feature adopted.  I'm guessing if a browser doesn't support it, it will scroll off the top of the page like the current situation.

Alex-Jordan:
Here is a [chart](https://caniuse.com/css-sticky). It looks like IE never supported it, and Chrome was as late as 2016.

It can be tricky to get it to work. If some ancestor of the sticky object has its overflow value set in certain ways, then the sticky won't work. It's tricky because you might not know that the ancestor overflow values matter, and also you may have a ton of ancestors to examine.

But yeah, if it doesn't work, it should just scroll off the screen.

For now I can't think of a scenario where two of these would be on one screen. I believe the second one would scroll up and cover the first one if that scenario ever comes about. They would both be stuck to the top of the screen, tied with z-index, and I assume the latter one wins the z-index tie to appear "on top".

Probably it is not great to name a class "sticky".

drgrice1 (commented):
I guess this is okay.

I am not entirely fond of the sticky style in this instance though.  I don't like that this makes the problem nav sticky on all pages that have it and for students.  So on the problem sets page the `Homework Sets` button is sticky, and on a homework problem page for students or instructors not acting as a student the `Previous/Next Problem` buttons are sticky even if the student nav is not present.

drgrice1 (approved):

Alex-Jordan:
Maybe I can add classes so it only happens to the buttons that instructors use to navigate. I guess that would mean doing things at a deeper level than the template.

This started when an instructor requested that these buttons reappear at the bottom of pages, and using sticky seemed like an easier way to achieve the real goal of not having to scroll back up. And perhaps, it's better since it doesn't add a new thing to the page.

Alex-Jordan:
I'm trying things here. Status quo is like:

```
div
    div#problem-nav
        div.user-nav
        problem navigation buttons not in their own div
    div#message (message like "this set is visible to students")
```

In the above, the outer div, div#problem-nav, and div#message are created by the system.template. The div.user-nav and the buttons are created by Problem.pm.


I'm looking for reaction/suggestions to the following.

1. In Problem.pm I can put the problem navigation buttons in their own div, parallel to the div.user-nav.
2. Move the div#message out of here, positioning it to the right of the div that contains the h1. This simplifies the navigation area but also seems like a good move for other reasons. If that message is important, it should come earlier in the DOM for a screen reader, and immediately following the h1 is a good place.
3. Now the original div#problem-nav is unnecessary. So there would be the outer div from the system template, and then div.user-nav and a new div.problem-nav (not the original one) that are built by Problem.pm.
4. I still can't isolate sticky to the div.user-nav. The fact that it is within that outer div means that this can only work if that outer div is sticky. The outer div is built by the sytem.template, where I cannot test if the user navigation will be present, so I don't see how to conditionally apply sticky to it. So if someone else sees a way to do that, good. Otherwise I think that sticky is only possible if it always applies to the entire navigation div. And maybe that is not bad? Making the Previous/Next buttons sticky seems not bad, even for students. (Making the "Homework Sets" button sticky does not feel helpful, but also seems harmless.)
5. If the entire navigation div is sticky and the div#message is out of there, I could explore putting the user navigation and problem navigation to the left and right instead of stacked as two rows the way it currently is. Unsure if that is better, but I could look at it.

drgrice1:
Be careful with 5 if you go that route. Three buttons are already to wide for narrow screens (like phones).  Part of the problem there is the excessive padding on the body. We should use a media query to reduce that for narrow screens. That wouldn't fix the problem if all six buttons were on the same row unless care is taken to do the layout right.  This can be accomplished with bootstrap's responsive layout.

Sticky contains the element in its parent. Your current setup works because the parent of the user-nav div is the main content div which is almost as tall as the page (all but the masthead and footer I believe).

Another option is to used fixed positioning instead. This takes JavaScript to make this work though, so it is not a pure css solution. However it can be nicer if done well.

Alex-Jordan:
I have the responsive layout going with things like span5 and offset1, thanks to something you showed me a few weeks ago. Not sure I like it better than stacking.

>Sticky contains the element in its parent.

This took me a while to parse. I get it now: using sticky constrains the element to be within its parent. And if the parent gets pulled off screen, too bad. I get that. I'm wondering if there is a way to put a (sticky) class on the parent div here only when the content deeper inside has student navigation. There's no CSS way, since `:has()` is not implemented in browsers. I'm not up for a javascript method. So that leaves some kind of logic conditional when the page is built. But to me I don't see how that could be done. The conditional needs to execute inside the system.template, and I couldn't see how to set that up.

I'm going to leave this PR as it is for now, and open a separate one tonight or tomorrow to see a new layout alternative.

drgrice1:
Attached is a patch demonstrating a javascript solution that only makes the user nav "sticky".  I am not sold on the concept of anything sliding with the page though.

Would you prefer this "sticky" approach, or would you rather just have the user nav at the bottom of the page as well as the top. Would you want that right before or after the submit buttons?  If you would rather have the buttons at the top and bottom, then I will implement that.  I have a way that it can be done without having to re-obtain the student data from the database.  I think I would prefer that to the "sticky" approach.

Alex-Jordan:
I'd be happy to take an opinion poll, but I think the sticky solution is better than a second button at the bottom. I'm coming from the experience of reviewing "incorrect" answers in quizzes. Sometimes I just want to move on to the next student, and I'm neither at the top nor the bottom of the page at that moment.

As I say that, maybe it suffices to only make the user-nav sticky for quizzes, which is easier since the other stuff is not present.

Alex-Jordan:
Following https://github.com/Alex-Jordan/webwork2/pull/8, what do you think if I change system.template to put this navigation div immediately under the breadcrumb navigation? So the title and message go under it.

Screen Shot 2021-06-09 at 5 43 06 PM

And then in sticky:

Screen Shot 2021-06-09 at 5 47 58 PM



This way it doesn't really travel far before becoming sticky. Also this div goes the full width to the right, and to me that clashes with how the problem itself is span10. Making this div span10 caused other problems. But moving this div up above the title makes that clash that I perceive less noticeable.

drgrice1:
I guess that is okay.  As to the clash of the width of the nav with the width of the problem, I noticed that as well.  Adding bootstrap 2.3.2's span10 class will mess up the flex styling.  This is another deficiency of bootstrap 2.3.2.  Bootstrap versions 4 and 5 use flex styling for the grid layout and this could have been done directly with their `col-??` class (which replaces the bootstrap 2 `span??` classes).  Although, I actually don't like that we use `span10` for the problem content.  I think it looks better if we use a `span12` there, and give the problem the width of the page instead of putting an empty `span2` div to the right.

Alex-Jordan:
I made further CSS changes and one structural change.

In the end there are two functional changes in this PR.

1. The student-nav and problem-nav are sticky now.
2. The breadcrumb now comes before the main nav. In my view it was not right to have it where it was previously. More specifically, the open/collapse nav button was in the wrong place. To see what I mean, go to small screen (prior to this PR). At first, the main navigation should be collapsed. But then use the button to open it, and the button gets sent to the bottom of the menu. It should stay at the top so you can close the menu right away. This PR moves it so that happens.

And then there is styling to go along with these things. Here are some screen shots of what things look like with all this. These screenshots don't show the "sticky" but you can see the thing that becomes sticky.

# Small screen 
Screen Shot 2021-06-10 at 5 02 51 PM

# Gateway
Screen Shot 2021-06-10 at 5 02 29 PM

# Homework set, Problem level
Screen Shot 2021-06-10 at 5 02 10 PM

# Homework set
Screen Shot 2021-06-10 at 5 01 57 PM

# Problem sets page
Screen Shot 2021-06-10 at 5 01 31 PM

htdocs/themes/math4/system.template
```diff
@@ -90,6 +90,27 @@
...
+				<!--#path style="bootstrap" text="<li><span aria-hidden='true' class='divider'>/</span></li>"-->
+			</ul>
+		</div>
+	</div>
+</div>
+
+<!-- Run breadcrumb js since it gets loaded early -->
+<script type="text/javascript">
+	$('#toggle-sidebar').removeClass('btn-primary')
+</script>
```
drgrice1, line 112: While you are at it delete this script and delete the `btn-primary` class (and the preceding comment and the extra `btn` class) from the `id="toggle-sidebar"` element.  It is rather pointless to add the class in html, and then use inline javascript to remove it.

Alex-Jordan: Got it.

drdrew42 (approved):

PR #1386, opened 2021-06-01
Change default server user and group

pstaabp:
This changes the server user and group to standard values for Ubuntu, RHEL and CentOS.


taniwallach (changes requested):
We should modify 
I made https://github.com/pstaabp/webwork2/pull/39 to have `docker-entrypoint.sh` set the correct values for Ubuntu in `site.conf`.

taniwallach:
Thanks for making the PR and merging my patch.

Just for the "records" - these settings mainly influence data printed in a few places.

However, it does have one important operational impact. The script `bin/setfilepermissions` does several `chgrp` operations using a value which comes from the value of `$server_groupID` set in `site.conf`. That script would not behave well (bad `chgrp`) if no value were set, so I made another change - to `die` when needed, and also to try fallback values for the user for the `chown` commands and to die if the new owner to be set is not known.

Sorry I did not think to check where these variables matter sooner.

pstaabp:
@taniwallach good catch.  I don't think I've used that script to fix file permissions, but I would bet someone does.

taniwallach:
> @taniwallach good catch. I don't think I've used that script to fix file permissions, but I would bet someone does.

I do not either - but not I might. I only found it using `grep` as I realized that the current `site.conf.dist` if not used properly could leave the 2 variables unset, and wondered if something might break as a result.

taniwallach (approved):

bin/setfilepermissions
```diff
@@ -50,6 +50,22 @@ chomp($me);
...
+	} elsif ( `/usr/bin/grep -c apache /etc/passwd`) {
+		warn "No login name available, using apache as the user name for the chown calls.";
+		$me = "apache";
+	} else {
+		die "Aborting: No login name available, and cannot fall back to www-data or apache for the chown calls.";
+	}
+}
+if ( ! $servergroup ) {
+	die "Aborting: server_groupID not set - so cannot do the chgrp calls. Check your conf/site.conf file.";
+}
```
drgrice1, line 67: Another possibility is to obtain the actual server user and group using `apach2ctl -S`.  This outputs
```
...
User: name="www-data" id=33 not_used
Group: name="www-data" id=33 not_used
```

taniwallach: My main intention in modifying `bin/setfilepermissions` was to avoid it running when the values to use for user/group are available. Giving a reasonable error message when necessary, and using some reasonable defaults seemed reasonable.


@drgrice1 - Your approach will more accurately determine a better default value for the user - but I'm not sure it is worth the effort of parsing the output from `apache2ctl -S` just to get a better default. If you think the effort is justified - we can check the improved code you provide.

I think the original intention was that the user would be the non-privileged user of the "main sysadmin" to allow that person to modify course files with relative ease. Adding a default allows the script can be used in some cases where the current code would fail (ex. inside a bash shell inside a Docker container started with `docker container exec -it webwork2_app_1 bash`.) Having the default be the user under which Apache is expected to run seemed a reasonable default. Your approach will do a better job of determining the correct username for Apache.

About the group - the intention is certainly to allow the web-server (and sysadmins) to all access and modify the relevant directories/files/ At least for WW 2.15, when a `wwdata` group was used (as in the approach from https://webwork.maa.org/wiki/Installation_Manual_for_2.15_on_Ubuntu_20.04_Server) the web-server user was added to that group (as were sysadmins), but the server itself was not running as that group (but just as the OS standard for the relevant distribution). Thus, polling for the group under which Apache is running is not really what is intended. I think that may justify `die`-ing and forcing the user to make the expected setting in `site.conf` to be able to use the script.

It could be that a better and more generic solution would be to have a `bin/setfilepermissionsTo` script which expects 2 arguments (user and group) and does all the "real work" assuming that the values are valid, modifying the current `bin/setfilepermissions` script to try to collect reasonable choices and use them to call the `bin/setfilepermissionsTo` script. Again - this may be an unnecessary effort.

drgrice1: I really don't think it is worth the effort to do this.  Particularly for a script that I am not sure is used by anyone.  I was just throwing an idea out there.

taniwallach: Then I think the current quick solution is good enough for now.

drgrice1 (approved):
This looks fine.

PR #1387, opened 2021-06-02
Fix a typo from the latest ADD_JS_FILES update that causes warnings

drgrice1:
Without this pull request run `apache2ctl -S` to see
`%jsFiles{...} in scalar context better written as $jsFiles{...} at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 2365.`

With this pull request run the same thing an enjoy not seeing that warning.

This typo did not happen in the other two places the ADD_JS_FILES handling occurs.


taniwallach (commented):
Thanks for fixing this. I noticed it in my docker startups and decided it was not urgent enough to fix yet.

PR #1348, opened 2021-04-27
log LTI authentication to course's login.log

Alex-Jordan:
This replaces #1297.

When a user comes in through an LMS, if it is a new WW user, there is a record in the course's login.log. But if they authenticate and enter WW, there is currently no record in the login.log. Instructors cannot verify a student was in the course when they report that they were. This commit changes that.

All I really wanted was for LTIAdvanced, to log a "LOGIN OK" message. To mirror what happens with basic password logins, I put a message "AUTH LTI: user authenticated" message as a counterpart to "AUTH WWDB: password accepted". I really don't know what the "AUTH WWDB" is supposed to convey, so I really don't know if "AUTH LTI" is a good counterpart.

And then there are all these points where authentication could fail, so I felt I should put log messages there too. I'm not confident I made these messages say the right thing. And each one is preceded by `$self->{log_error} .= ...`, so I suspect it is redundant to add what I added. However I am not sure where the messages in `$self->{log_error}` end up...

And then even though I only cared about LTIAdvanced, it seemed appropriate to do the same things in LTIBasic.

I've only tested this for a student user that enters through the LMS, using LTIAdvanced, and successfully enters. Testing the fail points seems unnecessary, at least compared with the effort it would take to engineer each fail.

PR #1388, opened 2021-06-02
Change LTIAdvanced.pm so that all new logins get logged + some other fixes

taniwallach:
This is intended to solve the same problem as https://github.com/openwebwork/webwork2/pull/1348, but by forcing the regular logging code to trigger by setting `$self->{initial_login} = 1;` on all logins.

The change was only made to `LTIAdvanced.pm`.

Now all new logins get logged,even when
  1. `$ce->{LMSManageUserData}` is false, and when
  2. `maybe_update_user()` would not modify the user due to the user's permission level being > `$ce->{userRoles}->{$ce->{LTIAccountCreationCutoff}}`.
Both of those cases failed to be logged in the past.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -248,7 +248,7 @@ sub get_credentials {
       # User id and address is at this point
       if ( $ce->{debug_lti_parameters} ) {
 	warn "=========== summary ============";
-	my $tmpIDmsg = ( $used_LMS_user_id == 1 ) ? " was the remote LMS LTI provided user_id\n" : "\n" ;
+	my $tmpIDmsg = ( $used_LMS_user_id == 1 ) ? " (the remote LMS LTI provided user_id)\n" : "\n" ;
 	warn "User id is |$self->{user_id}| $tmpIDmsg";
 	warn "User mail address is |$self->{email}|\n";
 	warn "Student id is |", $self->{student_id}//'undefined',"|\n";
@@ -287,7 +287,7 @@ sub check_user {
     return $self->SUPER::check_user(@_);
   }
 
-  if (!defined($user_id) || (defined $user_id and $user_id eq "")) {
+  if (!defined($user_id) || (defined $user_id && $user_id eq "")) {
     $self->{log_error} .= "no user id specified";
     $self->{error} = $r->maketext("There was an error during the login process.  Please speak to your instructor or system administrator.");
     return 0;
@@ -453,13 +453,9 @@ sub authenticate {
       $self->{error} .= $r->maketext("There was an error during the login process.  Please speak to your instructor or system administrator.");
       $self->{log_error} .= "Construction of OAuth request record failed";
       return 0;
-  } # was an elsif here, but we would have returned in the case that the if condition held
+  }
 
-  $verifyOK = $request->verify || $altrequest->verify;
-
-  # Space for future LTI verification options
-
-  if ( ! $verifyOK ) {
+  if ( ! $request->verify && ! $altrequest->verify) {
       debug("LTIAdvanced::authenticate request-> verify failed");
       debug("OAuth verification Failed ");
 
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -464,15 +464,12 @@ sub authenticate {
 			}
 		}
       }  elsif ($ce->{LMSManageUserData}) {
-		# Existing user.  Possibly modify demographic information and permission level.
-		if ($self->maybe_update_user()) {
-			$self->{initial_login} = 1; # Set here so login gets logged, even for accounts which maybe_update_user() would not modify
-		} else {
-			$r->maketext("There was an error during the login process.  Please speak to your instructor or system administrator.");
-			  $self->{log_error} .= "Failed to update user $userID.";
-			if ( $ce->{debug_lti_parameters} ) {
-				warn("Failed to updateuser $userID.");
-			}
+		$self->{initial_login} = 1; # Set here so login gets logged, even for accounts which maybe_update_user() would not modify or when it fails to update
+		# Existing user. Possibly modify demographic information and permission level.
+		unless ( $self->maybe_update_user() ) {
+			# Do not fail the login if data update failed
+			# FIXME - In the future we would like the message below (and other warn messages in this file) to be sent via maketext.
+			warn("The system failed to update some of your account information. Please speak to your instructor or system administrator.");
 		}
       } else {
 	$self->{initial_login} = 1; # Set here so login gets logged when $ce->{LMSManageUserData} is false
@@ -625,8 +622,10 @@ sub maybe_update_user {
   my $user = $db->getUser($userID);
   my $permissionLevel = $db->getPermissionLevel($userID);
   # We don't alter records of users with too high a permission
-  unless (defined($permissionLevel->permission) &&
-	  $permissionLevel->permission > $ce->{userRoles}->{$ce->{LTIAccountCreationCutoff}}) {  
+  if (defined($permissionLevel->permission) &&
+	$permissionLevel->permission > $ce->{userRoles}->{$ce->{LTIAccountCreationCutoff}}) {
+    return 1; # Treat as OK as not allowed to modify
+  } else {
     # Create a temp user and run it through the create process
     my $tempUser = $db->newUser();
     $tempUser->user_id($userID);
@@ -638,7 +637,7 @@ sub maybe_update_user {
     $tempUser->first_name($first_name);
     $tempUser->email_address($self->{email});
     $tempUser->status("C");
-    $tempUser-> section($self->{section} // "");
+    $tempUser->section($self->{section} // "");
     $tempUser->recitation($self->{recitation} // "");
     $tempUser->student_id($self->{student_id} // "");
     
@@ -647,8 +646,8 @@ sub maybe_update_user {
       $ce->{LTI_modify_user}($self,$tempUser);
     }
 
-    my @elements = qw(last_name first_name 
-		      email_address status 
+    my @elements = qw(last_name first_name
+		      email_address status
 		      section recitation student_id);
 
     my $change_made = 0;
@@ -656,21 +655,29 @@ sub maybe_update_user {
     for my $element (@elements) {
       if ($user->$element ne $tempUser->$element) {
 	$change_made = 1;
-	warn "WeBWorK User has $element: ".$user->$element." but LMS user has $element ".$tempUser->$element."\n" 
+	warn "WeBWorK User has $element: ".$user->$element." but LMS user has $element ".$tempUser->$element."\n"
 	  if ( $ce->{debug_lti_parameters} );
       }
     }
 
     if ($change_made) {
       $tempUser->comment(formatDateTime(time, "local"));
-      $db->putUser($tempUser);
-      $self->write_log_entry("Demographic data for user $userID modified via LTIAdvanced login");
-      warn "Existing user: $userID updated.\n"
-	if ( $ce->{debug_lti_parameters} );
+      eval { $db->putUser($tempUser) };
+      if ($@) {
+        $self->write_log_entry("Failed to update user $userID in LTIAdvanced login: $@");
+        warn "Failed to update user $userID in LTIAdvanced login.\n"
+	  if ( $ce->{debug_lti_parameters} );
+        return 0;
+      } else {
+        $self->write_log_entry("Demographic data for user $userID modified via LTIAdvanced login");
+        warn "Existing user: $userID updated.\n"
+	  if ( $ce->{debug_lti_parameters} );
+	return 1;
+      }
+    } else {
+      return 1; # No changes needed
     }
   }
-
-  return 1;
 }
 
 ################################################################################
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -105,9 +105,9 @@ sub  request_has_data_for_this_verification_module {
 
   # We need at least these things to verify an oauth request
   if (!(defined $r->param("oauth_consumer_key"))
-      or !(defined $r->param("oauth_signature"))
-      or !(defined $r->param("oauth_nonce"))
-      or !(defined $r->param("oauth_timestamp")) ) {
+      || !(defined $r->param("oauth_signature"))
+      || !(defined $r->param("oauth_nonce"))
+      || !(defined $r->param("oauth_timestamp")) ) {
     debug("LTIAdvanced returning that it has insufficent data");
     return(0);
   } else {
@@ -213,7 +213,7 @@ sub get_credentials {
       # after @
 
       if (!defined($self->{user_id})
-	  || ( $self->{email} ne "" # modified to ne "" as fallback value defined above
+	  || ( $self->{email} ne ""
 	       && defined($ce->{preferred_source_of_username})
 	       && $ce->{preferred_source_of_username} eq "lis_person_contact_email_primary")) {
 	$self->{user_id} = $self->{email};
@@ -287,7 +287,7 @@ sub check_user {
     return $self->SUPER::check_user(@_);
   }
 
-  if (!defined($user_id) or (defined $user_id and $user_id eq "")) {
+  if (!defined($user_id) || (defined $user_id and $user_id eq "")) {
     $self->{log_error} .= "no user id specified";
     $self->{error} = $r->maketext("There was an error during the login process.  Please speak to your instructor or system administrator.");
     return 0;
@@ -297,11 +297,12 @@ sub check_user {
 
   if (!$User) {
     if (    defined($r->param("lis_person_sourcedid"))
-	 or defined($r->param("lis_person_sourced_id"))
-	 or defined($r->param("lis_person_source_id"))
-	 or defined($r->param("lis_person_sourceid"))
-	 or defined($r->param("lis_person_contact_email_primary"))
-         or $ce->{allow_lti_to_use_provided_user_id}
+	 || defined($r->param("lis_person_sourced_id"))
+	 || defined($r->param("lis_person_source_id"))
+	 || defined($r->param("lis_person_sourceid"))
+	 || defined($r->param("lis_person_contact_email_primary"))
+	 || $ce->{allow_lti_to_use_provided_user_id}
+	 || $ce->{force_lti_to_use_provided_user_id}
        ) {
       debug("User |$user_id| is unknown but may be an new user from an LSM via LTI. About to return a 1");
       return 1;  #This may be a new user coming in from a LMS via LTI.
@@ -498,7 +499,6 @@ sub authenticate {
 		}
       }  elsif ($ce->{LMSManageUserData}) {
 		$self->{initial_login} = 1; # Set here so login gets logged, even for accounts which maybe_update_user() would not modify or when it fails to update
-		$self->{was_LTI} = 1;
 		# Existing user. Possibly modify demographic information and permission level.
 		unless ( $self->maybe_update_user() ) {
 			# Do not fail the login if data update failed
@@ -507,7 +507,6 @@ sub authenticate {
 		}
       } else {
 	$self->{initial_login} = 1; # Set here so login gets logged when $ce->{LMSManageUserData} is false
-	$self->{was_LTI} = 1;
       }
 
       # If we are using grade passback then make sure the data
@@ -641,7 +640,6 @@ sub create_user {
   }
 
   $self->{initial_login} = 1;
-  $self->{was_LTI} = 1;
 
   return 1;
 }
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -465,13 +465,17 @@ sub authenticate {
 		}
       }  elsif ($ce->{LMSManageUserData}) {
 		# Existing user.  Possibly modify demographic information and permission level.
-		unless ($self->maybe_update_user()) {
+		if ($self->maybe_update_user()) {
+			$self->{initial_login} = 1; # Set here so login gets logged, even for accounts which maybe_update_user() would not modify
+		} else {
 			$r->maketext("There was an error during the login process.  Please speak to your instructor or system administrator.");
 			  $self->{log_error} .= "Failed to update user $userID.";
 			if ( $ce->{debug_lti_parameters} ) {
 				warn("Failed to updateuser $userID.");
 			}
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -465,13 +465,17 @@ sub authenticate {
...
 		# Existing user.  Possibly modify demographic information and permission level.
-		unless ($self->maybe_update_user()) {
+		if ($self->maybe_update_user()) {
+			$self->{initial_login} = 1; # Set here so login gets logged, even for accounts which maybe_update_user() would not modify
+		} else {
 			$r->maketext("There was an error during the login process.  Please speak to your instructor or system administrator.");
 			  $self->{log_error} .= "Failed to update user $userID.";
 			if ( $ce->{debug_lti_parameters} ) {
 				warn("Failed to updateuser $userID.");
 			}
```
drgrice1, line 475: The log_error that is set in this portion will never make it into the log.  The reason is that the `authenticate` subroutine later returns 1, and as a result the `verify` method of Authen.pm considers this a valid sign in (but not an initial sign on).  This method should return 0 here.  This is also true in the section preceding this in the case that user creation fails.

Now for this portion of the code the above argument is moot.  The reason for this is that this portion of the code only executes if `$self->maybe_update_user()` returns a false value.  However, that will never happen because that subroutine absolutely always returns 1.

drgrice1 (approved):
This tests out, and I think this is the right way to deal with this.

On a related note, however, there are some other broken things going on in this file.  See my comments.  These perhaps do not need to be fixed now, but they need to be looked at.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: I modified `maybe_update_user()` to return a meaningful result, but am only triggering a `warn` message if the update fails. I see no reason to block the login if the LMS provided different data (ex. a change to the user's name). I moved the debug messages into the method, triggered if the `$db-putUser()` call fails.

taniwallach:
> This tests out, and I think this is the right way to deal with this.
> 
> On a related note, however, there are some other broken things going on in this file. See my comments. These perhaps do not need to be fixed now, but they need to be looked at.

Fixed.

lib/WeBWorK/Authen/LTIAdvanced.pm
drgrice1: Yeah, that is probably correct.  If the user is valid, but something fails in the data update, it should still be a valid sign in.  I realized that later.  It is just that the log_error is not handled for valid sign ins.  It looks like you took care of that.

taniwallach:
I added a bit of new code in the third commit.
The main purpose was to add 2 new settings:
  - `force_lti_to_use_provided_user_id` - when true the LMS provided `user_id` will always be used as the WW `user_id`.
    - Use case: A setting where the LMS side wants to minimize what personal information it send to WW due to privacy concerns.
  - `allow_lti_to_use_provided_user_id` - when true can fall back to using the LMS provided `user_id` as the WW `user_id` when all else fails.
The new `$self->{was_LTI} = 1` is related to local code I have to get `html2xml` to work with LTI authentication, and is a small change - so I included it here already. The larger change goes where I put a placeholder, but depends on changes to additional files.

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -190,55 +190,78 @@ sub get_credentials {
...
+      # or if the user_id is still undefined at this point
+      # then replace the user_id with the full email address.
       # if strip_address_from_email ==1  strip off the part of the address
       # after @
 
       if (!defined($self->{user_id})
-	  or (defined($self->{email})  
-	      and defined($ce->{preferred_source_of_username})
-	      and $ce->{preferred_source_of_username} eq "lis_person_contact_email_primary")) {
+	  || ( $self->{email} ne "" # modified to ne "" as fallback value defined above
```
Alex-Jordan, line 216: Do we want comments like this in the code? It is more of a comment on this change being made by this commit than a comment on the code itself.

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -253,31 +276,33 @@ sub check_user {
...
     return 0;
   }
-  
+
   my $User = $db->getUser($user_id);
-  
+
   if (!$User) {
-    if ( defined($r->param("lis_person_sourcedid"))
+    if (    defined($r->param("lis_person_sourcedid"))
```
Alex-Jordan, line 299: Sorry for the nitpick: spacing issue here. And I think 5 lines down at line 304. Should all of these `or` become `||`?

Alex-Jordan (approved):
Tested that it works to solve the original issue. The rest of it looks OK, but I did not test things introduced in the 2nd and 3rd commit. Two comments about spacing and code commentary that you can take or leave.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: I thought it would help explain to anyone who looked in the future why specifically here `defined` is not the correct test.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: About the spacing - I added the extra 3 spaces, as there was a series of defined tests for request parameters due to the multiple "options" for the "same" thing. I think in this case lining them up makes it easier for people reading the code to follow. The main issue (when that code was written) was the lack of proper compliance with the LTI standard where different LMSs used (at least historically) somewhat different strings to name this "field".

About "or": I changed to `or` to `||` there and in another location. This code was using `or` for no real reason here. https://perldoc.perl.org/perlop#Operator-Precedence-and-Associativity and  https://perldoc.perl.org/perlop#List-Operators-(Rightward) explain when `or` is needed.

taniwallach:
For the record, the current use case of `$force_lti_to_use_provided_user_id=1` was LTI form an open-edX MOOC, as standard edX installations prevent LTI from providing personal identification information (student's name or email) over LTI. Thus, essentially the only reasonable thing to use is the edX `user_id` which is some ugly UUID style string, but is better than nothing.

taniwallach:
I edited https://webwork.maa.org/wiki/LTI-Advanced_Authentication to make the manner in which WW `user_id` is set precise, as the old version of the documentation was far from precise. I also documented there the old `$preferred_source_of_student_id` option (until now only explained in the code and in `authen_LTI.conf.dist`, and the 2 new settings.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: @Alex-Jordan Please reapprove, as the 4th commit made changes (basically `or` to `||`) - so the approval seems to have disappeared. About the one location of human friendly spacing and that comment - I prefer to leave them as they are.

@drgrice1 - I have tested the `$force_lti_to_use_provided_user_id = 1` new feature, but an additional test would be nice. I'm hoping we can finish this PR off and merge it.

lib/WeBWorK/Authen/LTIAdvanced.pm
drgrice1: I agree with @Alex-Jordan on this.  This comment is not needed.  Anyone reading the code can see why defined is not the correct test as `$self->{email}` is defined to be the empty string just a few lines above.  So this comment is just giving information that should be commented on in this pull request, but not in the code.

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -464,14 +498,17 @@ sub authenticate {
...
       }  elsif ($ce->{LMSManageUserData}) {
-		# Existing user.  Possibly modify demographic information and permission level.
-		unless ($self->maybe_update_user()) {
-			$r->maketext("There was an error during the login process.  Please speak to your instructor or system administrator.");
-			  $self->{log_error} .= "Failed to update user $userID.";
-			if ( $ce->{debug_lti_parameters} ) {
-				warn("Failed to updateuser $userID.");
-			}
+		$self->{initial_login} = 1; # Set here so login gets logged, even for accounts which maybe_update_user() would not modify or when it fails to update
+		$self->{was_LTI} = 1;
```
drgrice1, line 502: Delete lines 502, 511, and 645.  You set `$self->{was_LTI}` on those lines, but never use that anywhere.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: Comment removed using a forced push, so that change is in the 4th commit which mainly fixes the `or`.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: Deleted in a second forced push of the last commit. I had explained the rationale above - but it can wait.

> The new `$self->{was_LTI} = 1` is related to local code I have to get `html2xml` to work with LTI authentication, and is a small change - so I included it here already. The larger change goes where I put a placeholder, but depends on changes to additional files.

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -253,31 +276,34 @@ sub check_user {
...
 
   # See comment in get_credentials()
   if ($r->{xmlrpc}) {
     #debug("falling back to superclass check_user (xmlrpc call)");
     return $self->SUPER::check_user(@_);
   }
-  
-  if (!defined($user_id) or (defined $user_id and $user_id eq "")) {
+
+  if (!defined($user_id) || (defined $user_id and $user_id eq "")) {
```
drgrice1, line 290: If you are changing the `or` to `||` on this line, you might as well also change the `and` to `&&`.

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -190,55 +190,78 @@ sub get_credentials {
...
-     }
-      
       # For setting up its helpful to print out what the system think the
-      # User id and address is at this point 
+      # User id and address is at this point
       if ( $ce->{debug_lti_parameters} ) {
 	warn "=========== summary ============";
-	warn "User id is |$self->{user_id}|\n";
+	my $tmpIDmsg = ( $used_LMS_user_id == 1 ) ? " was the remote LMS LTI provided user_id\n" : "\n" ;
+	warn "User id is |$self->{user_id}| $tmpIDmsg";
```
drgrice1, line 252: In the case that the LMS provided user_id is used this reads, "User id is |lti_user_id| was the remote LMS LTI provided user_id" which isn't a properly structured sentance.  I suggest adding parentheses around the `$tmpIDmsg` part to obtain, "User id is |lti_user_id| (was the remote LMS LTI provided user_id)".

lib/WeBWorK/Authen/LTIAdvanced.pm
drgrice1: I tested the `$force_lti_to_use_provided_user_id` option.  It seems to work fine.

Although, I am not certain that this is the right way to go.  One thing that has always bugged me is the `$preferred_source_of_username` option.  The way that currently works is that if it is set to `lis_person_contact_email_primary` then the provided email address is used (and the address stripped if the `$strip_address_from_email` is also set).  If that option is undefined or set to anything else, the option is ignored.  So it seems that the option with its current usage should be named something like `$use_email_for_username` and simply have it be a boolean value.  Another option is to leave the name of that option, and actually make it work for other values.  For example it could be set to `user_id`, and in that case use the LTI provided `user_id`.  In other words, make this work much like the `$preferred_source_of_student_id` works.

drgrice1 (approved):
This generally looks good, but there are a few thinks that could use some changes.

lib/WeBWorK/Authen/LTIAdvanced.pm
drgrice1: I didn't see the explanation above about the new `$self->{was_LTI}`.  Although, you certainly should not be adding code to achieve things that you have in your local code.  If this is part of a possible later pull request then add it to that pull request.  Don't preemptively add code that is currently not used.  That is why we have things in the code now that still are not used.  Add the entire finished product when it is ready, or put it in a draft pull request.

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -425,19 +453,25 @@ sub authenticate {
...
       $self->{log_error} .= "Construction of OAuth request record failed";
       return 0;
-    } elsif (! $request->verify && ! $altrequest->verify) {
+  } # was an elsif here, but we would have returned in the case that the if condition held
+
+  $verifyOK = $request->verify || $altrequest->verify;
+
+  # Space for future LTI verification options
+
+  if ( ! $verifyOK ) {
```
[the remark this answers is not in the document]
drgrice1: Was your point here that the previous `if` condition if true would have returned, and as such the `elseif` is unneccesary and an `if` serves the purpose?  In other words, to improve readability of the code?  If that was what you meant, then that is fine.  The comment in the code isn't needed though.  The new `$verifyOK` variable isn't either.  Just literally change the `elsif` to an `if` on the next line and that end is achieved.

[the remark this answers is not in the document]
taniwallach: There is a closed PR  https://github.com/openwebwork/webwork2/pull/1090  which had sample code to enable LTI authentication to `html2xml`. The `was_LTI` was needed for that, that code needed to add a new verification stage here. I was trying to keep local and main code easy to compare the placeholder helped. Since I still have that provisional code active on one server, I had intended to reduce the amount of extra work needed to keep things in sync.

Being changed in a new commit to just a new
`  if ( ! $request->verify && ! $altrequest->verify) {`
rather than an `elsif`.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: Being done in the next commit.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: Being done in the next commit, but I am dropping the work "was" inside the parenthesis.

taniwallach:
> I tested the `$force_lti_to_use_provided_user_id` option.  It seems to work fine.
> 
> Although, I am not certain that this is the right way to go.  One thing that has always bugged me is the `$preferred_source_of_username` option. ... So it seems that the option with its current usage should be named something like `$use_email_for_username` and simply have it be a boolean value.

This code is quite old and the suboptimal behaviour is "not our fault", but is in frequent production use while using the current names for the settings.

I agree in principle, but such a change would break behavior for existing courses, and backwards compatibility seems to be more valuable than the variable having the most accurate name. I edited https://webwork.maa.org/wiki/LTI-Advanced_Authentication 2 days ago to make this whole messy situation as clear as possible, and emphasized the poor name of the variables.

> Another option is to leave the name of that option, and actually make it work for other values.  For example it could be set to `user_id`, and in that case use the LTI provided `user_id`.  In other words, make this work much like the `$preferred_source_of_student_id` works.

That would also potentially break behavior of existing courses. Right now any value other than `lis_person_contact_email_primary` is disregarded, and a functionality change would change the login time `user_id` also for existing courses. Avoiding that would require an additional switch to flip from the old and bad behavior to the new and improved behavior.

I think we are best leaving "working now and good enough" alone for WW2, and fixing this properly for WW3.

lib/WeBWorK/Authen/LTIAdvanced.pm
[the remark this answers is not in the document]
taniwallach: > Was your point here that the previous if condition if true would have returned, and as such the elseif is unneccesary and an if serves the purpose?

Yes.

> In other words, to improve readability of the code? If that was what you meant, then that is fine.

> The comment in the code isn't needed though.

Agree and being removed.

> The new $verifyOK variable isn't either. Just literally change the elsif to an if on the next line and that end is achieved.

That is what is in the new commit, soon to be on GitHub. Testing again before I push the commit.

lib/WeBWorK/Authen/LTIAdvanced.pm
taniwallach: > If this is part of a possible later pull request then add it to that pull request. Don't preemptively add code that is currently not used. 
Agree, but our time is also a valuable resource, and at times it is easier to save a bit of time and preemptively add code that is needed later on.

In any case, it has already been removed for now.

lib/WeBWorK/Authen/LTIAdvanced.pm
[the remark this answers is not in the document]
taniwallach: There is a closed PR  https://github.com/openwebwork/webwork2/pull/1090  which had sample code to enable LTI authentication to `html2xml`. The `was_LTI` was needed for that, that code needed to add a new verification stage here. I was trying to keep local and main code easy to compare the placeholder helped. Since I still have that provisional code active on one server, I would like to reduce the amount of extra work needed to keep things in sync, and that ...

dlglin:
> > Another option is to leave the name of that option, and actually make it work for other values.  For example it could be set to `user_id`, and in that case use the LTI provided `user_id`.  In other words, make this work much like the `$preferred_source_of_student_id` works.
> 
> That would also potentially break behavior of existing courses. Right now any value other than `lis_person_contact_email_primary` is disregarded, and a functionality change would change the login time `user_id` also for existing courses. Avoiding that would require an additional switch to flip from the old and bad behavior to the new and improved behavior.
> 
> I think we are best leaving "working now and good enough" alone for WW2, and fix this properly for WW3.

On my prod server I modified the code to do exactly what @drgrice1 suggests since we are using a different value (student ID) as the login.  I consider the naming of the variable to be a bug, so I'm okay with modifying the behaviour even if it isn't backwards compatible.  It will only affect systems where it is misconfigured, and it can be explicitly mentioned in the release notes.

taniwallach:
> On my prod server I modified the code to do exactly what @drgrice1 suggests since we are using a different value (student ID) as the login. I consider the naming of the variable to be a bug, so I'm okay with modifying the behaviour even if it isn't backwards compatible. It will only affect systems where it is misconfigured, and it can be explicitly mentioned in the release notes.

OK - I will work on making the non-backwards-compatible change, merging into this PR. If you have code to start from - please send it to me, or make a PR to my branch.

As I see it:
  - The first choice for `user_id` for WW should be the field requested by `$preferred_source_of_username`, so long as it was provided by the LMS.
  - Fallback options according to an order set in an array, which the system/course config can override. How about:
    1. `lis_person_sourcedid` (in the various spellings)
    2. `lis_person_contact_email_primary` (with/without stripping domain, depending on `strip_address_from_email`)
    3. LMS `user_id`

Note: The LTI 1.1 spec states that `user_id` is formally only recommended in the "basic launch data" (https://www.imsglobal.org/specs/ltiv1p1/implementation-guide#toc-3) but I strongly l suspect that is the best possible final fallback value, and better than just using the email as the fallback value.

Once we finalize the decisions - we can edit https://webwork.maa.org/wiki/LTI-Advanced_Authentication to refer to the WW 2.16 flow, and leave the explanation about earlier versions in a section at the bottom.

dlglin:
> > On my prod server I modified the code to do exactly what @drgrice1 suggests since we are using a different value (student ID) as the login. I consider the naming of the variable to be a bug, so I'm okay with modifying the behaviour even if it isn't backwards compatible. It will only affect systems where it is misconfigured, and it can be explicitly mentioned in the release notes.
> 
> OK - I will work on making the non-backwards-compatible change, merging into this PR. If you have code to start from - please send it to me, or make a PR to my branch.
> 

My local changes really just add the line
`$self->{user_id} = $r->param($ce->{preferred_source_of_username});`

Since I know that the attribute is defined for my instance, I didn't bother to write any fallback logic.

> As I see it:
> 
> * The first choice for `user_id` for WW should be the field requested by `$preferred_source_of_username`, so long as it was provided by the LMS.
> * Fallback options according to an order set in an array, which the system/course config can override. How about:
>   
>   1. `lis_person_sourcedid` (in the various spellings)
>   2. `lis_person_contact_email_primary` (with/without stripping domain, depending on `strip_address_from_email`)
>   3. LMS `user_id`
> 
> Note: The LTI 1.1 spec states that `user_id` is formally only recommended in the "basic launch data" (https://www.imsglobal.org/specs/ltiv1p1/implementation-guide#toc-3) but I strongly l suspect that is the best possible final fallback value, and better than just using the email as the fallback value.
> 
> Once we finalize the decisions - we can edit https://webwork.maa.org/wiki/LTI-Advanced_Authentication to refer to the WW 2.16 flow, and leave the explanation about earlier versions in a section at the bottom.

If a `$preferred_source_of_username` is defined, then I'm not sure that having fallback options is a good idea.  This could lead to unpredicted behaviour for the administrator.  Consider the following scenario:

1. The administrator sets a custom attribute as the preferred source of username.
2. A user clicks on the LTI link, but that attribute is not defined.
3. WeBWorK goes through the fallback list, and assigns a username from one of those attributes.

In such a scenario, the administrator may intend for the username to be something like the student's ID number, but could end up with some users having their email address as their username.

lib/WeBWorK/Authen/LTIAdvanced.pm
Alex-Jordan: > I think in this case lining them up

But from my perspective, the three spaces are not aligning things. I think maybe your tab display is set to something that made it look aligned to you, but the first line having all spaces at the front messes that up.

Here in this GitHub diff, I see the following, not aligned:

```
    if (    defined($r->param("lis_person_sourcedid"))
       || defined($r->param("lis_person_sourced_id"))
       || defined($r->param("lis_person_source_id"))
```

where the 2nd and 3rd lines lead with a tab that GH uses 6 spaces for, followed by a space; and the first line leads with four spaces. (Elsewhere in GH, I see GH using tabs that are 8 wide, and then it lines up.)

I don't really care about the spaces. But I did want to let you know that I understand and respect the desire to align things. :) It's just that you are only succeeding with that according to your own tab settings.

taniwallach: @Alex-Jordan  - Thanks. You are right, and @drgrice1 has reminded me of this in the past. I should remember the pitfall and not try.  Will try to remember to drop it in the revised PR.

For anyone who runs into this discussion: The WeBWorK [Coding Standards](https://webwork.maa.org/wiki/Coding_Standards) page, which is not well followed in the existing code, was to use tabs to indent lines (but many files don't do this, and often it is easiest to just stick to whatever style was in use in the file). Lining up using spaces to a tab on a later line is definitely not reliable.

taniwallach:
Closing and replacing by https://github.com/openwebwork/webwork2/pull/1402 without modifications to the handling of `user_id` but with a minor modification for `student_id`.

Revised code to handle `user_id` will be in a different (soon to be created) PR, building on top of the https://github.com/openwebwork/webwork2/pull/1402.

PR #1389, opened 2021-06-02
fixed a bug with db passwords with special characters.

pstaabp:


drdrew42 (approved):

drgrice1 (approved):

taniwallach (approved):
Reverts a change (my fault) from https://github.com/openwebwork/webwork2/commit/a086e50a68dbcbc598d4708b85c57d966776a604#diff-abc09f3a937b2832dd3e516b5f6cb47eeab9fe1c55502063865eec6ec52030ec which dropped the quoting of these settings in the generated config file and broke passwords with special characters (specifically `#`).

Based on https://dev.mysql.com/doc/refman/5.7/en/option-files.html it seems that the quoting is probably only needed if there is a '#' character inside the user name or the password.

Issue #1336, opened 2021-04-20
Removing "published" column not compatible with older versions of mysql/MariaDB


dlglin:
https://github.com/openwebwork/webwork2/blob/9985302e6031983f5c5384957dd7a66eaddbe66b/lib/WeBWorK/DB/Schema/NewSQL/Std.pm#L383 uses the "ALTER TABLE ... DROP COLUMN IF EXISTS ..." syntax.

The IF EXISTS option for dropping a column was only added in MariaDB 10.0.2, so this will fail on any server running an older version of MariaDB or mysql.  CentOS 7 is currently still using MariaDB 5.5, so this fails on a standard CentOS installation.

I expect that the "IF EXISTS" was included to prevent the command from returning an error if there is no such column, but I believe that this code is only called after verifying that the column is present, so there shouldn't be an issue with removing it.

taniwallach:
Being fixed by https://github.com/openwebwork/webwork2/pull/1391

somiaj:
Fixed in #1391 (From what I can tell). Closing.

PR #1391, opened 2021-06-02
drop the 'if exists' from drop column stmt

drdrew42:
MySQL does not support `ALTER TABLE DROP COLUMN ... IF EXISTS` statement construction.


drgrice1 (approved):
Looks good.

dlglin:
This change fixes #1336, but there is an issue with the underlying code.

The code to delete the published column is inside the CheckTableFields routine, which is only intended to identify issues with DB columns, which are then fixed by the UpdateTableFields routine.  This means that in the admin course, as soon as you click on "Upgrade Courses" link it will delete the extra column, rather than initially reporting what changes it will make, and then not making the changes until you click the "Upgrade Courses" confirmation button at the bottom.

PR #1394, opened 2021-06-03
Create xmlrpc server variable

pstaabp:
This creates a `$xmlrpc_root_url` variable to allow xmlprc to run on a separate server or in my case under http://  since the https:// is taken care of on a proxy level.  

If no variable is set, it should use `$server_root_url` as the url.

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -48,6 +48,11 @@ $webwork_url       = '/webwork2';
 $server_root_url   = '';   # e.g.  'https://webwork.yourschool.edu' or 'http://localhost'
                            # Note, if running a secure (ssl) server, you probably need 'https://...'
 
+# if the XMLRPC runs on a different server or uses http internally, uncomment the following
+# line and set the proper URL.  For example, 'http://192.168.1.1'
+
+# $xmlrpc_root_url = ''; 
+
 # The following two variables must match the user ID and group ID respectively
 # under which apache is running.
 # In the apache configuration file (often called httpd.conf) you will find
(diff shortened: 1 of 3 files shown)


pstaabp:
Easiest way to test this is to go to the library browser and make sure nothing fails when selecting problems from the library.

taniwallach:
Are you sure that the `$FORM_ACTION_URL` should also be changed? I think that it is used by generated problems as the address to submit to - so should probably remain the regular externally available URL.

You should be able to tell if it should remain unchanged by checking a problem embedded in an iFrame using the method from 
http://michaelgage.blogspot.com/2015/06/whether-writing-full-text-book-or-just.html and 

@mgage - Any comments?

pstaabp:
Finally dug into this a bit.  My server has `X-Frame-Options` set to `SAMEORIGIN` to prevent the loading into a iframe, so I wasn't able to test this.   Here's some other info on that: [https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options). 

@taniwallach I guess unless I relax the `X-Frame-Options`, this wouldn't work, but perhaps I can try another way.

pstaabp:
Just tested this on a development server that I was able to upgrade to 2.16 without the `X-Frame-Options` set to `SAMEORIGIN`.  

I changed 
```
$xmlrpc_root_url = 'http://127.0.0.1';
```

and have success with both internal xmlrpc calls like the library browser as well as the external iframe version like @taniwallach suggested above.  

Hopefully my logic is correct, and would be great for others to test.

PR #1397, opened 2021-06-05
Give output formats a consistent display order on the hardcopy generation page

drgrice1:
This ensures that the "Adoby PDF" and "TeX Source" hardcopy format radio buttons appear in the same order everytime when they are both present on the hardcopy generator page.

I find it rather annoying when trying to debug a problem that these radio buttons keep switching order, and I end up checking the wrong one.  I am sure I am not the only one that has found this annoying.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -64,7 +64,7 @@ our %HC_FORMATS = (
 	tex => { name => x("TeX Source"), subr => "generate_hardcopy_tex" },
 	pdf => { name => x("Adobe PDF"),  subr => "generate_hardcopy_pdf" },
 );
-our @HC_FORMAT_TYPES = ('tex', 'pdf');
+our @HC_FORMAT_DISPLAY_ORDER = ('tex', 'pdf');
 
 # custom fields used in $self hash
 # FOR HEAVEN'S SAKE, PLEASE KEEP THIS UP-TO-DATE!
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
Looks right, and simple enough.

pstaabp (approved):
didn't test it, but seems straightforward enough.

Alex-Jordan:
Alternatively, you could add order like:

```
our %HC_FORMATS = (
    	tex => { name => x("TeX Source"), subr => "generate_hardcopy_tex", order => 1},      
    	pdf => { name => x("Adobe PDF"),  subr => "generate_hardcopy_pdf", order => 2},
);
```

and then use 
```
sort {$HC_FORMATS{$a}->{order} <=> $HC_FORMATS{$b}->{order}} keys %HC_FORMATS
```

to sort them.

Or something like that, I didn't test it out.

I'm not expressing a preference, just pointing out there may be a way to avoid the redundancy.

taniwallach (approved):
I also only read the code and did not test it. It seems fine. Maybe calling it `@HC_FORMAT_DISPLAY_ORDER` would make the purpose clearer. Since I doubt there are likely to be additions here, such a rename is not critical.

drgrice1:
@Alex-Jordan:  I am not sure that sorting is better.  Generally using an array is the way that this is done in Perl.  Either that or using a module that has ordered hashes (like Tie::IxHash), but that is probably excessive for this case.

@taniwallach:  I changed the variable name.

taniwallach (approved):

PR #1398, opened 2021-06-08
Limit html2xml grade pass-back to when it is explicitly allowed.

taniwallach:
Prevent the `html2xml` subsystem from being able to make LTI grade pass-back submissions unless explicitly allowed via a course (or system) setting due to the potential security issues.

Fix some other issues along the way:
  1. Prevent LTI grade pass-back for occurring when not a "submit" or when render fails.
  2. Add a `hideWasNotRecordedMessage` setting to hide the message which refers to grades not being saved (in WeBWorK), so that message will not be displayed if not desired. Default is still to show it.

The first of the 4 commits is just white-space changes, mainly adding indentation - so the other commits are easier to follow.

### DIFF
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -203,24 +203,27 @@ sub formatRenderedProblem {
 	my $showAnswerNumbers = $self->{inputs_ref}{showAnswerNumbers} // 1;
 
 	# Attempts table
-	my $tbl = WeBWorK::Utils::AttemptsTable->new(
-		$rh_result->{answers} // {},
-		answersSubmitted    => $self->{inputs_ref}{answersSubmitted} // 0,
-		answerOrder         => $rh_result->{flags}{ANSWER_ENTRY_ORDER} // [],
-		displayMode         => $displayMode,
-		showAnswerNumbers   => $showAnswerNumbers,
-		ce                  => $ce,
-		showAttemptPreviews => $previewMode || $submitMode || $showCorrectMode,
-		showAttemptResults  => $submitMode || $showCorrectMode,
-		showCorrectAnswers  => $showCorrectMode,
-		showMessages        => $previewMode || $submitMode || $showCorrectMode,
-		showSummary         => (($showSummary and ($submitMode or $showCorrectMode)) // 0) ? 1 : 0,
-		maketext            => WeBWorK::Localize::getLoc($formLanguage),
-		summary             => $problemResult->{summary} // '', # can be set by problem grader
-	);
-	my $answerTemplate = $tbl->answerTemplate;
-	my $color_input_blanks_script = (!$previewMode && ($checkMode || $submitMode)) ? $tbl->color_answer_blanks : "";
-	$tbl->imgGen->render(refresh => 1) if $tbl->displayMode eq 'images';
+
+	# Increase indent below - to make coming changes easier to track
+
+		my $tbl = WeBWorK::Utils::AttemptsTable->new(
+			$rh_result->{answers} // {},
+			answersSubmitted    => $self->{inputs_ref}{answersSubmitted} // 0,
+			answerOrder         => $rh_result->{flags}{ANSWER_ENTRY_ORDER} // [],
+			displayMode         => $displayMode,
+			showAnswerNumbers   => $showAnswerNumbers,
+			ce                  => $ce,
+			showAttemptPreviews => $previewMode || $submitMode || $showCorrectMode,
+			showAttemptResults  => $submitMode || $showCorrectMode,
+			showCorrectAnswers  => $showCorrectMode,
+			showMessages        => $previewMode || $submitMode || $showCorrectMode,
+			showSummary         => (($showSummary and ($submitMode or $showCorrectMode)) // 0) ? 1 : 0,
+			maketext            => WeBWorK::Localize::getLoc($formLanguage),
+			summary             => $problemResult->{summary} // '', # can be set by problem grader
+		);
+		my $answerTemplate = $tbl->answerTemplate;
+		my $color_input_blanks_script = (!$previewMode && ($checkMode || $submitMode)) ? $tbl->color_answer_blanks : "";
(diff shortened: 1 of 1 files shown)


lib/FormatRenderedProblem.pm
```diff
@@ -380,8 +408,12 @@ sub saveGradeToLTI {
 	my $consumer_secret = $ce->{'LISConsumerKeyHash'}{$consumer_key};
 	my $score = $rh_result->{problem_result} ? $rh_result->{problem_result}{score} : 0;
 
-	# This is boilerplate XML used to submit the $score for $sourcedid
-	my $replaceResultXML = <<EOS;
+	my $LTIGradeMessage = '';
+
+        if ( ! $forbidGradePassback ) {
```
drgrice1, line 413: This is indented with 8 spaces instead of a tab.  So it only works correctly with a tabstop of 8.

drgrice1 (commented):
This generally looks good.  There are a couple of changes needed though.

lib/FormatRenderedProblem.pm
```diff
@@ -93,6 +98,13 @@ sub formatRenderedProblem {
 
 	my $mt = WeBWorK::Localize::getLangHandle($self->{inputs_ref}{language} // 'en');
 
+	if ( $forbidGradePassback == 1 &&
+	     defined( $ce->{html2xmlAllowGradePassback} ) &&
+	     $ce->{html2xmlAllowGradePassback} eq "This course intentionally enables the insecure LTI grade pass-back feature of html2xml." ) {
```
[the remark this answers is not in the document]
taniwallach: Yes, but this is a severely flawed feature at present.
If I was not worried that someone "really needs it" I would say we could just hard-wire it off for now.

lib/FormatRenderedProblem.pm
taniwallach: Sorry. Fixed being fixed in the next commit.

lib/FormatRenderedProblem.pm
```diff
@@ -75,11 +76,15 @@ sub formatRenderedProblem {
 	my $rh_result   = $self->return_object() || {};  # wrap problem in formats
 	$problemText    = "No output from rendered Problem" unless $rh_result;
 
-	if (ref($rh_result) and $rh_result->{text}) {
+	my $forbidGradePassback = 1; # 1 = feature blocked, 2 = due to rendering error, 3 = not in submit mode
+
+	if (ref($rh_result) && $rh_result->{text}) {
 		$problemText = $rh_result->{text};
 	} else {
 		$problemText .= "Unable to decode problem text:<br>$self->{error_string}<br>" .
```
Alex-Jordan, line 84: This is not related to the PR, but since I am looking at the code...

I can't work out if it is appropriate that raw HTML `<br>` is used here. Is it?

taniwallach: It may not be perfect, but under the assumption that the `$problemText` is expected to be handled by a browser, it should work. I'm not sure we can do much better for passing this back with some attempt to get it to render nicely in a more general manner.

Alex-Jordan (approved):
I'm sure it works, but I can't really test with this. I did read the code changes line by line and it seems fine. 

Nitpicking: I'm not a fan of encoding the 1, 2, and 3 to have meaning. And then sometimes that meaning is helpfully repeated in a nearby comment like line 87, but then sometimes not, like line 259. Consider using strings like `$forbidGradePassback = 'blocked'`, `$forbidGradePassback = 'error'`, `$forbidGradePassback = 'notsubmit'`. They will still evaluate to true in the two places where `$forbidGradePassback` is used as a boolean.

 But I approve either way.

taniwallach:
> I'm sure it works, but I can't really test with this. I did read the code changes line by line and it seems fine.
> 
> Nitpicking: I'm not a fan of encoding the 1, 2, and 3 to have meaning. And then sometimes that meaning is helpfully repeated in a nearby comment like line 87, but then sometimes not, like line 259. Consider using strings like `$forbidGradePassback = 'blocked'`, `$forbidGradePassback = 'error'`, `$forbidGradePassback = 'notsubmit'`. They will still evaluate to true in the two places where `$forbidGradePassback` is used as a boolean.

Will work on that soon. Marking as not ready to merge.

taniwallach:
@drgrice1 @pstaabp @mgage: I thought some about @Alex-Jordan's recommendation to replace the numeric values being used for the variable `$forbidGradePassback` in this PR, and am not sure that it makes sense to change them to string values as he suggested. In several places the code tests for a specific value to decide how to act, and doing that test using a string `eq` will certainly be less efficient than a numeric `==`. I would like some advice on whether to make such a change or not.

drgrice1:
It is less efficient to compare string values than to check numeric values.  It may be possible to restructure the code a bit so that the value of the variable could be boolean in nature (either 0 or 1).  Then the variable name has intrinsic meaning, and the desire for adding meaning via string values is moot.  One way to do this is to use two variables instead of one.  One that is boolean as described above, and another that is also boolean (something like `$renderErrorOccured`) that tracks if an error has occurred.

taniwallach:
> It is less efficient to compare string values than to check numeric values. It may be possible to restructure the code a bit so that the value of the variable could be boolean in nature (either 0 or 1). Then the variable name has intrinsic meaning, and the desire for adding meaning via string values is moot. One way to do this is to use two variables instead of one. One that is boolean as described above, and another that is also boolean (something like `$renderErrorOccured`) that tracks if an error has occurred.

@drgrice1 - Thanks. That seems to make sense. Will try to implement it soon.

taniwallach:
Sorry - I had fixed some typos (`$renderErrorOccured` -> `$renderErrorOccurred`) but forgot to commit all of them... so fixed it and git a forced push.

drgrice1 (approved):
This looks good.

@Alex-Jordan: Is this approach with the variables better?  Does it make more sense this way?

Alex-Jordan:
Thanks @taniwallach and sorry for the extra work. I hereby re-approve :)

PR #1399, opened 2021-06-09
Update site.conf.dist

mgage:


mgage:
using double quotes caught Frank Wolfs since his database password had an @ sign in it.  I've run into this before, including with people specifying email addresses directly in configuration files using double quotes.

drgrice1:
This looks good.  Since this only adds comments and a minor quote change, I will go ahead and merge it.

PR #1402, opened 2021-06-10
Change LTIAdvanced.pm so that all new logins get logged + some other fixes - version 2

taniwallach:
This replaces https://github.com/openwebwork/webwork2/pull/1388.
  - The first 2 commits are identical to those in https://github.com/openwebwork/webwork2/pull/1388.
  - The 4th commit here is just a cleanup of trailing whitespace.
  - The 3rd commit:
    1. Handles the "agreed" changes from the discussion there which in the last 3 commits of https://github.com/openwebwork/webwork2/pull/1388 
      - Avoid warning when email not provided by LMS.
      -  Replace `or` with `||` and `and` with `&&` where used just for logic in control flow.
      - Has a few minor code fixes.
    2. **New:** Make sure the WW `student_id` is only defined using `preferred_source_of_student_id` and fall back to an empty string to avoid warnings.
  - This PR does **not** modify the code to set `user_id` - which will be addressed in a different PR.

Given the reviews of  https://github.com/openwebwork/webwork2/pull/1388 the only **new** change to be carefully reviewed is the handling of `student_id` which is in accordance with a recent decision by the development team to make it `preferred_source_of_student_id` behave as the name means and avoid using any less-than-stable fall-back value (the form parameter `custom_student_id` was acting as a fall-back source for the value).

Once reviewed - I think a "squash" merge is recommended. The PR will provide the internal details.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -118,10 +118,11 @@ $preferred_source_of_username = "lis_person_contact_email_primary";
 # LTI Preferred source of Student Id
 ################################################################################
 
-# If preferred_source_of_student_id is not set, WeBWorK will look in the LTI
-# parameter 'custom_student_id.' If preferred_source_of_student_id is set, then
-# WeBWorK will look in the LTI parameter specified below. This parameter may be
-# dependendent on the LMS. E.G., In D2L, student_id is stored in OrgDefinedId and
+# If preferred_source_of_student_id is not set, WeBWorK will set student_id
+# to be an empty string. You should use debug_lti_parameters in order to determine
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
This looks good.  Everything checks out in my tests.

It would be good for someone else to look at this.

Alex-Jordan (approved):
Looks good to me too.

Alex-Jordan:
I would squash and merge as suggested, except I don't know what our guidelines are with the commit messages. Keep them all as is?

drgrice1:
I think just take them all as they are in this case.  I think @taniwallach has it set up so that it will work well.

PR #1403, opened 2021-06-10
LTI (Advanced) modify user_id selection

taniwallach:
This is a totally reworked approach to setting the WW `user_id` for `LTIAdvanced.pm`.

It is based off the code in https://github.com/openwebwork/webwork2/pull/1402, and was rebased off `WeBWorK-2.16`after that PR was squash-merged. 

Two settings are used:
  - `preferred_source_of_username` (**required**) names the LTI parameter to use as the first choice for WW's `user_id`.
  - `fallback_source_of_username` (**optional**) provides a single fallback option.

As in the past when a setting has the value `lis_person_sourcedid`, several options are checked due to LMS "spelling" issues.

As in the past when `lis_person_contact_email_primary` is selected as the option to use, the setting `strip_address_from_email` can be used to remove the part after the `@`.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -77,32 +77,45 @@ $authen{user_module} = [
 # LTI preferred and fallback source of WW user_id
 ################################################################################
 
-# If both lis_person_sourcedid (or one of its variants)
-# and lis_person_contact_email_primary are present,
-# then one needs to state which one is preferred
-# as the WeBWorK username.
-# The considerations are (1) ensuring that usernames
-# are unique and (2) being as compatible as possible with the practices
-# of the institutions that are being served in a site.
-# If a site is only being used be students from a single
-# institution and the value  lis_person_sourcedid  is
-# unique across the population and matches the  logon username
-# at that institution, then lis_person_sourcedid is
-# probably the better choice.
-# On the other hand, if a site is serving a population
-# from several institutions and  lis_person_sourcedid  is not
-# necessarily unique across the population, then
-# lis_person_contact_email_primary is the better choice.
-# Because the WeBWorK username must be unique across
-# the population served by a site, the default if
-# neither line is uncommented will be the value of
-# lis_person_contact_email_primary.
+# You MUST set what LTI field is used to set the WeBWorK user_id.
+#
+# The lis_person_sourcedid (or one of its variants) is formally an OPTIONAL field
+#    https://www.imsglobal.org/specs/ltiv1p1/implementation-guide
+# but is relatively consistently available.
+# However, some LMS systems, Blackboard in particular, do not send lis_person_sourcedid.
+#
+# The email address lis_person_contact_email_primary is often a more understandable
+# value but may not exist.
+#
+# The LTI standard recommends providing a "user_id" value, which needs to be a unique
+# indentifier for each student in the LMS. The value is not expected to be people
+# friendly by may be available when neither of the others are.
+#
+# You need to make sure to use a setting such that
+#    (1) usernames are unique and
+#    (2) the setting is as compatible as possible with the practices
+#        of the institutions that are being served in a site.
+#
+# If each course is only being used be students from a single institution
+# and the value  lis_person_sourcedid is unique across that population
+# and matches the logon username at that institution, then lis_person_sourcedid
+# is probably the better choice.
+#
+# On the other hand, if a site is serving a population from several institutions
+# or if the lis_person_sourcedid is not necessarily unique across the population,
+# then lis_person_contact_email_primary is the better choice.
+#
+# NOTE: As of WeBWorK 2.16 some setting MUST be made.
+#
+#       If no setting is made, all LTI logins will fail and an error will be reported.
+#
+#       See the comment further down on how to get WeBWorK 2.16 to behave similarly to
+#       the prior behavior.
 
-# NOTE: Even if a course management system sends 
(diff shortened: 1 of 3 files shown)


taniwallach:
This needs to be tested by someone else who can do tests with LTI and various values of 
  - `preferred_source_of_username` (required) and
  - `fallback_source_of_username`
Below are some lines from a `course.conf` from a course I tested with.
```
#undef $preferred_source_of_username; # Intended to trigger an error
#$preferred_source_of_username = "lis_person_contact_email_primary";
#$preferred_source_of_username="";
#$preferred_source_of_username="evil"; # Intended to fail
#$preferred_source_of_username="user_id";

#$fallback_source_of_username="lis_person_sourcedid";
#$fallback_source_of_username="user_id";
#$fallback_source_of_username="lis_person_contact_email_primary";
#$fallback_source_of_username="evil"; # Intended to fail
```

taniwallach:
@dlglin raised the issue that using `$fallback_source_of_username` can lead to a user account being created initially with the value that setting requests, and should the LMS later provide the `$preferred_source_of_username` requested value for that user - a new account will the be created for the same LMS user.

This problem also existed in the old code which had built-in fall-over and no option to avoid it. 
  - When `lis_person_sourcedid` (in the various forms) did not exist, there was an automatic fall-over to `lis_person_contact_email_primary`, regardless of the setting of `$preferred_source_of_username`
  -  In the old code, `$preferred_source_of_username` was only used to override the default choice to use `lis_person_sourcedid` if the value of `$preferred_source_of_username` was `lis_person_contact_email_primary` and an email was provided.
  - Thus, even when `$preferred_source_of_username` was `lis_person_contact_email_primary` if the LMS failed to provide the email, there was an automatic fall-over to using `lis_person_sourcedid`.

---

The code in the PR has the following advantages over the current code:
  1. The value of `$preferred_source_of_username` has a meaning which fits the variable name and a **different** LTI field other than `lis_person_sourcedid` or `lis_person_contact_email_primary` **can be used** to set the WeBWorK `user_id`.
  2. There is **no automatic fail-over** to either of the 2 "standard" options as in the past, only an explicitly permitted option can be used. Thus, permitting a fallback option is something which must be **explicitly** allowed.
  3. By a suitable pair of settings, the new code can behave as the old code did, with automatic fail-over in the desired order.

---

**Change in default behavior**:  Use of **only** `$preferred_source_of_username` will **not allow** a fail-over to occur, as happened in the old code.

---

The documentation in `conf/authen_LTI.conf.dist` was revised to make all this as clear as possible.

---

The default value in `conf/authen_LTI.conf.dist` remains
```
$preferred_source_of_username = "lis_person_contact_email_primary";
```
as in the past.

dlglin:
I tested with Moodle and D2L and things work as expected for me.  Here are the tests I ran and the results.  For reference both my D2L and Moodle instances do not send lis_person_sourcedid, and do send the email address as lis_person_contact_email_primary.  

- With $preferred_source_of_username = "lis_person_contact_email_primary" and $fallback_source_of_username undefined, logins work as expected with the username set to the email address.
- With $preferred_source_of_username = "lis_person_sourcedid" and $fallback_source_of_username undefined, login fails with the warning message "LTI is not properly configured. Please contact your instructor or system administrator. at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 292.", which is expected since it can not find a source of username.
- With $preferred_source_of_username = "lis_person_sourcedid" and $fallback_source_of_username = "lis_person_contact_email_primary" login succeeds setting the email as the username as expected.
- With $preferred_source_of_username = "user_id" login succeeds and sets the username to whatever the LMS sends as user_id (in both of my LMSs this is an internal identifier).

The only other observation I have is that the strip_address_from_email is only available if the source of username is "lis_person_contact_email_primary".  That option should probably apply to any source of username.  There could be a scenario where the email is passed from the LMS as a different attribute, and currently the code would be unable to strip the address from the email.  In fact, looking at the code, with the new approach I think all of the conditional code around "lis_person_contact_email_primary" could probably be removed.

dlglin:
One more thought: if the code can't find a valid source of user id, then the user gets the warning message
"LTI is not properly configured. Please contact your instructor or system administrator. at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 292."

The same message is written to the apache error log.  Nowhere is it recorded that the misconfiguration has to do with the fact that $preferred_source_of_username and $fallback_source_of_username are not being sent by the LMS.  It would be very helpful to administrators for debugging if this were at least recorded in the error log.

taniwallach:
@dlglin - Thanks for testing.

> The only other observation I have is that the strip_address_from_email is only available if the source of username is "lis_person_contact_email_primary". That option should probably apply to any source of username.

I think it is somewhat dangerous to automatically drop everything from an `@` and on as a general option. Right now the feature maintain backwards compatibility with what the "strip domain from email" of the old code, and I think that is good enough for now.

> One more thought: if the code can't find a valid source of user id, then ... Nowhere is it recorded that the misconfiguration has to do with the fact that $preferred_source_of_username and $fallback_source_of_username are not being sent by the LMS. It would be very helpful to administrators for debugging if this were at least recorded in the error log.

I just modified the `warn` calls which issue this message to explain what triggered the message. There are 2 cases, configuration is totally missing `preferred_source_of_username` and when no `user_id` can be determined using that and the `fallback_source_of_username` if it is also defined.

PR #1404, opened 2021-06-10
Add support for a 2 stage Docker build process + support graceful Apache restart

taniwallach:
Docker improvements:

1. Add files for a 2 stage Docker build process.
2. Set things up so Apache gets stopped "gracefully" (current requests get completed).
  - https://httpd.apache.org/docs/2.4/programs/apachectl.html
  - Also, it is now possible to do `docker container exec -it webwork2_app_1 /usr/sbin/apachectl graceful`:
    - This is a faster restart if code is mounted from outside the container.
    - It should also help with log rotation.
3.   Also some minor fixes.

---

### 2 stage build process

```
# cd into webwork2 directory (as usual)
docker build --tag webwork-base:forWW216 -f DockerfileStage1 .
vim docker-compose.yml
# uncomment the line
#    dockerfile: DockerfileStage2
docker-compose build
```

The first stage does all the OS level work, and involves installing several hundred ubuntu packages (and some CPAN Perl packages), so takes quite some time.
The second stage installs WeBWorK, and is relatively fast.

---

### Force rebuild of OS later
```
# cd into webwork2 directory (as usual)
docker build --no-cache --tag webwork-base:forWW216 -f DockerfileStage1 .
```

### DIFF
--- a/DockerfileStage1
+++ b/DockerfileStage1
@@ -1,0 +1,146 @@
+# This is the Stage 1 Dockerfile, which builds a base OS image (webwork-base)
+# on top of which the WeBWorK parts will be installed by the Stage 2 Dockerfile.
+
+FROM ubuntu:20.04
+
+# ==================================================================
+
+# Phase 1 - set base OS image install stage ENV variables
+#
+# We only need install time ENV variables, not those needed by the WeBWorK system
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    DEBCONF_NONINTERACTIVE_SEEN=true
+
+# ==================================================================
+
+# Phase 2 - Ubuntu 20.04 base image + required packages
+
+# Packages changes/added for ubuntu 20.04:
+#       libcgi-pm-perl (for CGI::Cookie), libdbd-mariadb-perl
+
+# Do NOT include "apt-get -y upgrade"
+# see: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
+
+RUN apt-get update \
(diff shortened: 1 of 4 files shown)


drgrice1 (approved):
This looks good.  I have tested the build steps and it works.  There is an issue with the permissions on htdocs/tmp with the default build that needs to be resolved.

mgage:
the htdocs/tmp permission issue  is addressed in my upcoming pull request to patch the 2 stage docker process.

drgrice1:
> the htdocs/tmp permission issue is addressed in my upcoming pull request to patch the 2 stage docker process.

I see that. I see that you did it without the recursive flag as suggested.

taniwallach:
Note: this PR does not fix the `sed` code for creating `site.conf` which no longer properly sets the database password, as the `site.conf.dist` was fixed in https://github.com/openwebwork/webwork2/pull/1399 to use single quotes on the password, but `docker-config/docker-entrypoint.sh` was not yet fixed to handle that change. That is among the several things fixed in https://github.com/openwebwork/webwork2/pull/1433.

drgrice1:
I will go ahead and merge this unless someone objects.  It only affects the docker build, and seems to work well for that aside from the mentioned issues that are fixed in @mgage's pull request.

mgage:
I'm in favor of merging this PR.  It will make it easier for me to organize the additional minor changes I'd like to add.

mgage (approved):

PR #1405, opened 2021-06-10
WW 2.16 - mods to admin page for collecting site info / security group signup

taniwallach:
Add a section to the admin course main page about signing up for the WW-security-announce Google group / mailing list.

Provide a Google form based site registration form (replacing the MAA site) + modified text.

Right now this is a form I own, it would be good to move it to a Google account belonging to TWP.


taniwallach:
To do:
  - New form URL: https://forms.gle/vrqpKiRHritnQNf37 
  - Put a link to the form, and don't embed in an iFrame.
  - Set the link via a config setting.

dlglin:
Some comments/questions about the form:

1. Who is our audience for the form?  The "Server type" question has options for third-party hosting.  Do we want to include those in the same form?  Is the rest of the form relevant for those using Edfinity or MAA hosting?
2. We are asking for multiple email addresses: the person filling out the form and the sysadmin.  Having both of those be mandatory will probably lead to some abandoned registrations.  Also since we are collecting contact info the form should make it clear what that info will be used for.

taniwallach:
Data from the old form is under the map at https://webwork.maa.org/wiki/WeBWorK_Sites and several of the fields were based on what appears there. 

I'm pretty sure I sent an message (email or Slack) asking for feedback on the draft form.
> Some comments/questions about the form:
> 
> 1. Who is our audience for the form? 

Server admins / or the academic staff "responsible" for a server if the tech staff do not handle the functions handled via the admin course. One primary use would be to send announcements of new releases or of major bug fixes.

The intention is to reach people who do not read the forums or other main WeBWorK sites often.

> The "Server type" question has options for third-party hosting.  Do we want to include those in the same form?  Is the rest of the form relevant for those using Edfinity or MAA hosting?

Probably not. Maybe we should create 2 versions of the form - one for "locally operated" servers (where we ask for an email address for a local system administrator or alternate contact), and one for users of a "hosted" server, where only an academic content is likely available.

> 2. We are asking for multiple email addresses: the person filling out the form and the sysadmin.  Having both of those be mandatory will probably lead to some abandoned registrations.  Also since we are collecting contact info the form should make it clear what that info will be used for.

Good point. Some can be optional.

I think we can state that the sysadmin contact info will only be used to send important announcements (ex. new releases) and information letters no more frequently than once in X months unless there is a truly urgent need to inform site administrators of something.

We could add a checkbox to permit sending other less critical information (ex. information on pending releases, WeBWorK events, etc.) for each email address collected.

For academic contacts - is it reasonable to commit to sending no more than 3-4 emails per year?

taniwallach:
Recreated on a current copy of `WeBWorK-2.16` and revised to get URLs from `conf/default.conf` and to just give a link to the Google form without using an iFrame. There is a small hack to set the `subDisplay` hidden field which is needed to get the `Hide the banner` button to work, as the banner itself would not have set it as needed.

@dlglin @drgrice1 @pstaabp - I think this is now ready for review, after the changes done in response to what was discussed at the meeting.

Along the way a few `http` URLS in `conf/defaults.conf` were changes to `https`.

---

@mgage - What do you think about the idea of using a different form to collect data on users of "hosted" courses. Such submissions would not be filling in several of the fields. If we want to stick with one form - then at least 2 more fields should be optional:
  - System Admin contact name
  - Primary admin email address

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -3479,111 +3478,63 @@ sub display_registration_form {
...
+	CGI::p(join("",
+		"Eventually your site will be listed along with all of the others on the ",
+		CGI::a({href=>$ce->{webworkURLs}{SiteMap}, target=>"_blank"}, "site map"),
+		" on the main ",
+		CGI::a({href=>$ce->{webworkURLs}{WikiMain}, target=>"_blank"}, "WeBWorK Wiki"),".<hr>")),
+	CGI::p(join("",
+		'You can hide this "registration" banner for the future by clicking the button below. ',
+		'It writes an empty file (',
+		CGI::code('registered_versionNumber'),
+		') to the admin course directory.')),
```
drgrice1, line 3514: I don't think it is necessary to explain to the admin user the details of how the banner is hidden.  I recommend just deleting the additional commentary in lines 3512 - 3514.

drgrice1 (approved):
This looks good.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
taniwallach: I had kept that from the old `sub registration_form` code line 3557. It has been deleted (so there is no need for a `join` either any more).

drgrice1: I think that since you added the button to click to hide the registration form it makes the comment unnecessary.

pstaabp (approved):
Looks good.

taniwallach:
I accidentally made changes meant for a different branch and pushed, so used `git reset --soft HEAD~1;  git push --force` to get this back to where it belongs.

PR #1406, opened 2021-06-11
Allow the content div in the system template take up as much width as it can

drgrice1:
Allow the content div in the system template take up as much width as it can instead of leaving empty space on the right.  This is done by removing the empty span2's, and using span12 instead of span11 when the side nav is hidden.

This makes homework problems more consistent with gateway quizzes, and generally looks better to me. This also uses space better on other content pages as well.

@Alex-Jordan and I disagree on this.  I like this consistency, and @Alex-Jordan doesn't like the long lines.  See our discussion at https://github.com/Alex-Jordan/webwork2/pull/9 and https://github.com/Alex-Jordan/webwork2/pull/10.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -866,7 +866,8 @@ table.attemptResults span.answer-preview {
 /*styles for the instructor comment box */
 
 .answerComments {
-    margin-left: 2em;
+    margin-left: auto;
+    margin-right: auto;
     max-width: 80%;
     border-style: outset;
     border-width: 1px;
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -14,13 +14,13 @@ if ($.fn.button.noConflict) $.fn.bootstrapBtn = $.fn.button.noConflict();
 	var hideSidebar = function () {
 		$('#site-navigation').remove();
 		$('#toggle-sidebar-icon i').removeClass('fa-chevron-left').addClass('fa-chevron-right');
-		$('#content').removeClass('span10').addClass('span11');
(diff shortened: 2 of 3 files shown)


Alex-Jordan:
I don't disagree with having the div.content being so wide. With it's background coloring and border, it's a good change (as long as there is nothing to actually put off to the right, like say problem notes in ww3).

But I am not sold about having lines of text that students read possibly become that wide. With some time, I'd like to explore things like (just for example) using a max-width on `p` elements. Or possible PGML-generated `p` elements. Or...  

(Yes, somewhere else I wrote max-height where I meant max-width.)

drgrice1:
I am not  in favor of putting css style on generic html elements (like a `p` element) in general.  That is generally a bad idea.  css classes should be added to elements that the style is desired to be placed on.

Alex-Jordan:
Totally, and rest assured I wasn't suggesting that. In the GitHub post, for brevity, I may say "put CSS on the `p`s" but it's just because I don't need to get into the weeds here with too much detail.

Alex-Jordan:
Here are some screenshots of a statistics exercise with wide text. All at regular browser width on a MacBookPro. I'm not making a preference statement, just giving us something to look at.

## Current
Screen Shot 2021-06-11 at 4 24 14 PM

## With this change
Screen Shot 2021-06-11 at 3 58 51 PM

## With this change, and `p` children of `#output_problem_body` having `max-width: 600px;` (for example)
Screen Shot 2021-06-11 at 4 03 10 PM


An approach based on the `p` will not do much good. The first paragraph of a non-PGML problem is not actually in a `p` unless the author opened the statement using `$PAR`. And PGML is not using `p` to mark paragraphs, but rather it inserts a spacer div in between paragraphs. So something based on `p` is not worth thinking about much more.

PR #1407, opened 2021-06-13
Gateway navigation table

Alex-Jordan:
There are two commits here. The first is more or less trivial, inserting some space characters in between Gateway status messages so they don't come out like:

Your recorded score on this test (version 1) is 0/5.You exceeded the allowed time.

The second commit rebuilds the page/problem navigation table. There was a lot of excess things in the navigation table like nbsp characters for spacing and entire cells using `[` and `]` for delimiting pages. With this, all the excess stuff is cleared away and a table cell with a colspan is used for a page with multiple problems in it.

Then I also added a third row to the navigation table with problem scores. As an instructor with a 20-question quiz with 4 questions on each page, it can help to know that the student's only incorrect answer is on page 3, or something like that. Also if you are permitting students to see scores on previously graded submissions, this is good. I've had students complain that they graded a test and it told them a few were wrong. They went to preview some answer and that cleared the display of what was right/wrong and they didn't know how to get that info back. I was only allowing one additional graded submission, so they had a fair complaint.

Once scores are printed, it makes sense to list all of the problem numbers, even for problems on other pages. But I did not make them into links.

I put classes, colgroup elements, and col elements into the table to do better styling. Also I put an "active" class on the cell for the current page. All I did for now with styling was to put left/right borders on the colgroups, give everything some padding, highlight the active page cell, and style the row headers with CSS instead of how they had been styled with a `b` tag. I think more attention to styling would be great, but I don't know what direction to take it in.

More broadly, this could be improved upon. But it is an improvement from what is there now.

### DIFF
--- a/htdocs/themes/math4/gateway.css
+++ b/htdocs/themes/math4/gateway.css
@@ -98,12 +98,28 @@ div.gwProblem
 }
 
 .gwNavigation {
-    margin-top: 1em;	
-    margin-bottom: 1em;
+	margin-top: 1em;
+	margin-bottom: 1em;
+	text-align: center;
+}
+
+.gwNavigation th, td {
+	padding: 0 0.5rem;
 }
 
 .gwNavigation th {
     text-align:right;
+    font-weight: bold;
+    padding-right: 1rem;
+}
+
+.gwNavigation colgroup.page {
+	border-left: solid 1pt black;
+	border-right: solid 1pt black;
+}
+
+.gwNavigation .page.active {
+	background-color: #ffeeaa;
 }
 
 div.gwDivider {
--- a/htdocs/themes/math4/gateway.css
+++ b/htdocs/themes/math4/gateway.css
@@ -103,7 +103,7 @@ div.gwProblem
 	text-align: center;
 }
 
-.gwNavigation th, td {
+.gwNavigation th, .gwNavigation td {
 	padding: 0 0.5rem;
 }
 
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1978,7 +1978,7 @@ sub body {
 		# subsequent pages of a multipage test
 		print CGI::hidden({-name=>'pageChangeHack', -value=>''}),
 			CGI::br();
-        print CGI::hidden({-name=>'startTime', -value=>$startTime});
+		print CGI::hidden({-name=>'startTime', -value=>$startTime});
 		if ($numProbPerPage && $numPages > 1) { 
 			print CGI::hidden({-name=>'newPage', -value=>''});
 			print CGI::hidden({-name=>'currentPage', -value=>$pageNumber});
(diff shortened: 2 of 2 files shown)


Alex-Jordan:
Here are some screen shots to get a visual. Styling is of course up for changes.

## Multiple questions per page

Screen Shot 2021-06-12 at 6 00 00 PM

## One question per page

Screen Shot 2021-06-12 at 6 00 20 PM

## All questions on one page

Screen Shot 2021-06-12 at 6 00 32 PM

drgrice1 (approved):
This looks much better than before.

There are a couple of things that I notice though.

First, would it be possible to only show the score row after the quiz has been graded?

The second thing is really an existing problem that this doesn't fix, and would probably take a complete restructuring to fix.  So this is probably something that is not to be done now.  The issue is with narrow screens, quizzes with a large number of problems, quizzes with a large number of pages, or those factors combined.  In these cases the table goes off the screen to the right, and you have to scroll to reach those far right problems or pages.  This is always an issue when using tables for layout.

drgrice1 (changes requested):
Actually, there is more to the first issue I listed that makes this something that needs to be changes.  That ignores the `can showScore` settings.

drgrice1:
Furthermore, my initial issue was enough to require changes.  The scores should not be shown until the quiz is graded.  That is consistent with how it was before.

drgrice1 (approved):
Wait, I missed line 2030 where it does check the canShowProblemScores setting.  My bad.

drgrice1 (changes requested):
Wait once more.  Something is still not right though.  The score row is shown for a student before the quiz is graded.  The scores are all zeros, but that could be confusing for a student.  The score row shouldn't be shown at all until after the quiz is graded with at least one submission.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -1995,50 +1993,42 @@ sub body {
...
 		}
+		my @tableRows;
+		my @cols = (CGI::colgroup(CGI::col({class => 'header'})));
 		if ($numProbPerPage && $numPages > 1) {
-			my $pageRow = [ CGI::th( {scope=>"row"}, CGI::b($r->maketext('Jump to Page:'))),
-					CGI::td(CGI::b(' [ ' )) ];
+			push (@cols, (CGI::colgroup({class => 'page'},CGI::col({class => 'problem'}) x $numProbPerPage) x $numPages));
+			my @pages;
 			for my $i (1 .. $numPages) {
 				my $pn = ($i == $pageNumber) ? $i :
```
drgrice1, line 2014: This looks better to me if the `&nbsp;` is also added for the current page number.  Then the colums all have uniform spacing.  Although, perhaps even better would be to drop the `&nbsp;` for both cases, and modify the `.gwNavigation colgroup.page` rule in gateway.css (of the math4 theme) by adding `padding: 0 0.25rem;`.  You can twiddle the actual padding setting if you like.  `0.25rem` is a little less than the `&nbsp;` gives.

drgrice1: Looking closer I see that you already have a padding of `0.5rem` that applies (and my suggested rule is preempted by that).  So ignore the styling changes I suggested.  Just drop the `&nbsp` and this looks better already.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -1995,50 +1993,42 @@ sub body {
...
-					CGI::a({-href=>"#",
-							-onclick=>"jumpTo($pn);return false;"},
-						"$pn") . "&nbsp; ");
-				push(@$probRow, CGI::b(" ] ")) if ($i == $endProb);
-			} elsif (!($i % $numProbPerPage)) {
-				push(@$probRow, " &nbsp;&nbsp; ",
-					" &nbsp;&nbsp; ", " &nbsp;&nbsp; ");
+				push(@$probRow, CGI::a({-href=>"#", -onclick => "jumpTo($pn);return false;"}, "$pn"));
+			} else {
+				push(@$probRow, "$pn");
```
drgrice1, line 2002: There shouldn't be quotes on `$pn` on lines 2000 and 2002.  I realize that they were there before, but that was incorrect.  Single variables should not be interpolated.  That is an inefficiency to avoid in Perl.

Alex-Jordan: Got it in an upcoming commit.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
Alex-Jordan: Thanks, I had missed those nbsps in the links surrounding the number. Removed in an upcoming commit.

Alex-Jordan:
> The score row shouldn't be shown at all until after the quiz is graded with at least one submission.

Agreed. I'll change this later today.

Alex-Jordan:
>     The score row shouldn't be shown at all until after the quiz is graded with at least one submission.

OK, it is behaving this way now.

drgrice1 (approved):
This looks good now.  Much nicer than before.

It would still be nice to do something about the case that the page/problem nav is so wide that it make scrolling necessary to see the right part of it.

Alex-Jordan:
Without rebuilding the table structure, we could:

* Remove the `Move to ` and `Jump to ` from the row headers
* Shrink the padding

I think it could support up to about a 15-question quiz on a phone screen that way without needing to scroll. And on a regular screen, more than 3x as many. Is this worth doing? Or not, if you'd like to try something more responsive like with flex boxes or something?

drgrice1:
I think for now this is good.  It would take switching to a flex box layout to make it responsive.  It would probably better to put this off until we have a newer version of bootstrap.  Bootstrap versions 4 and 5 use flex box for their responsive layouts and make this sort of thing much easier to accomplish.

drgrice1:
I just noticed that on narrower screens the `Move to` and `Jump to` text wrap anyway.  So removing those won't help.

The changes in this pull request make the page/problem nav narrower than before, so this does help at least.

htdocs/themes/math4/gateway.css
```diff
@@ -98,12 +98,28 @@ div.gwProblem
...
 
 .gwNavigation {
-    margin-top: 1em;	
-    margin-bottom: 1em;
+	margin-top: 1em;
+	margin-bottom: 1em;
+	text-align: center;
+}
+
+.gwNavigation th, td {
```
Alex-Jordan, line 106:
```suggestion
.gwNavigation th, .gwNavigation td {
```

drdrew42 (approved):

PR #1410, opened 2021-06-13
Disable the quiz begin confirmation dialog for a gateway quiz if a quiz is in progress

drgrice1:
This fixes an issue with the new gateway quiz confirmation dialog that was pointed out by @Alex-Jordan.  See https://github.com/openwebwork/webwork2/pull/1355#issuecomment-860138358.

The issue is that if a quiz is in progress and the student returns to the problem sets list page, then the confirmation dialog is displayed again.  This pull request disables the dialog in that case.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSets.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSets.pm
@@ -304,19 +304,27 @@ sub body {
 	# Note: this assumes that the set_id of a versioned GW set cannot differ from all of the corresponding GW versions.
 	foreach my $set (@sets) {
 		die "set $set not defined" unless $set;
-		
-		if ($set->visible || $authz->hasPermissions($user, "view_hidden_sets")) {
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
Tested, and it works to stop the confirmation dialog. And if I load the problem sets page after a quiz version has expired, I can start a new version and the confirmation dialog is back.

There is a side effect. If I start a quiz version, then load the problem sets page, then walk away for some period of time and the quiz version times out (including the grace period) then I can start a new quiz version without getting the confirmation dialog. Of course that is because that problem set page was generated when the first version was still active.

This side effect is much more likely to come up when someone like me is testing this compared with a real student. Addressing this at the moment the student clicks a quiz compared to the moment the problem sets page is written seems like a significant project and not worth the small thing it would do. So I think this is great as it is.

drgrice1:
Yeah, that is much harder to address with the current set up.

PR #1411, opened 2021-06-13
show all sets that are successfully hidden

Alex-Jordan:
This is a very small thing that I ran into today which bothered me.

I went into the admin course to hide a handful of courses and the message I received was:

```
The following courses were successfully hidden: a-course-was-listed
```

Only one course was listed, even though I had hidden three. This change makes it list them all. Not that I need to see them all, but it's better than just listing one when you actually succeeded with all of them.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -3138,11 +3138,7 @@ sub do_hide_inactive_course {
 	}	
 	
 	if ($succeeded_count) {
-		if ($succeeded_count < 6) {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Wow.  The code you replaced was pretty messed up.  If less than six courses were hidden, then it would show "The following courses were successfully hidden: first_hidden_course", and if six or more courses were hidden then it would show "num_hidden courses were successfully hidden."

There is one change that I think you should make though.  The exact same issue happens when unhiding courses.  Make essentially the same changes on lines 3207-3210.

Alex-Jordan:
Thanks, I should learn to scan for similar code blocks. I changed the unhidden block, and force pushed.

drdrew42 (approved):

drdrew42 (approved):

PR #1412, opened 2021-06-13
Allow instructors to use spaces in set names (that get converted to u…

Alex-Jordan:
…nderscore)

I went through every place I could think of where an instructor has the option to type in the name of a set, and made it so they can type spaces. In all cases, the spaces are converted to underscore before anything happens with the set name. It is tedious to explain to new faculty users that they must use underscores when they want spaces. (Even more tedious to explain this repeatedly to instructors who forget).

Places where this has effect:

- Hmwk Sets Editor filter tab
- Hmwk Sets Editor import tab 
- Hmwk Sets Editor create tab 
- Library Browser create a new set button (actually there this was already happening, but not in a way that could be managed consistently with other places)
- Instructor Tools create a new set button
- Problem Set Detail page conditional release field

In each place, you should be able to use spaces in a set name, like for example "Quiz 2". Behind the scenes, it will become "Quiz_2".

There is one other change here with the filtering tool in Hmwk Sets Editor. Before, if you filter set names with say "Quiz", it only counts as a match if there is a set whose name is actually "Quiz". Or you could give a list like "Quiz_1, Quiz_2" and just those sets would appear. This did not seem especially useful. Rather than typing those out, why not just click to select them and filter that way?

Now with this PR, if you use "Quiz" then any set with that string in its name will match. So you might end up seeing Quiz 1, Quiz 2, Readiness Quiz. This seems more useful, like if you wanted to filter to only see your quizzes or some other subset where you have a naming convention that can be leveraged.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor.pm
@@ -657,6 +657,12 @@ sub read_dir {  # read a directory
 	return sort @files;
 }
 
+sub format_set_name {
+	my $string = shift;
+	$string =~ s/ /_/g;
+	return $string;
+}
+
 =back
 
 =cut
--- a/lib/WeBWorK/ContentGenerator/Instructor/Index.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Index.pm
@@ -238,19 +238,20 @@ sub pre_header_initialize {
 	};
 	
 	defined param $r "create_set" and do {
-	  my $setname = $r->param("new_set_name");
-	  if ($setname && $setname ne 'Name for new set here') {
-		if ($setname =~ /^[\w.-]*$/) {
-		$module = "${ipfx}::SetMaker";
-		$params{new_local_set} = "Create a New Set in this Course";
-		$params{new_set_name} = $setname;
-		$params{selfassign} = 1;
-		  } else {
-		push @error, E_BAD_NAME;
-		  }
-	  } else {
-		push @error, E_SET_NAME;
-	  }
+		my $setname = $r->param("new_set_name");
+		if ($setname && $setname ne 'Name for new set here') {
+			if ($setname =~ /^[\w .-]*$/) {
+				$setname = WeBWorK::ContentGenerator::Instructor::format_set_name($setname);
+				$module = "${ipfx}::SetMaker";
+				$params{new_local_set} = "Create a New Set in this Course";
+				$params{new_set_name} = $setname;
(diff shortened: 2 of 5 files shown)


pstaabp:
I think if we're doing this, then printing out the name consistently would be helpful.  For example on the SetDetail page, I would convert the _ to [space] everywhere on the top of the page.  (It appears to be in 4 locations). Also on Student progress.  The _ is converted to [space] on the list of homework sets, but once selected the _ comes back.

Alex-Jordan:
> I think if we're doing this, then printing out the name consistently would be helpful. For example on the SetDetail page, I would convert the _ to [space] everywhere on the top of the page. (It appears to be in 4 locations). Also on Student progress. The _ is converted to [space] on the list of homework sets, but once selected the _ comes back.

If it's worth it, I will attempt this. Do others think it is worth doing this?

There are many places where the underscores are already displayed as spaces. From what I can see, it's done in an ad hoc manner, like a line that uses `s/_/ /g` right before display. So I think doing this right would not only be about hunting down places where we currently see underscores, but also it would be good (if we are doing that) to pull in all these ad hoc instances into a uniform utility alongside the `sub format_set_name` that this PR is introducing.

mgage:
I agree that the conversions should be located in just one place.  I have mixed feelings about allowing spaces and converting them to _ vs. training people not to use spaces in identifiers but to use underscore (or -) instead. (I realize that the latter is a bias because I'm used to coding. If spaces are never allowed things are clear, if they are sometimes allowed, or only allowed if there are quotes then there is always some uncertainty. ).  If we are going to allow spaces then we need to create a consistent way of converting the spaces to _ and converting them back when they are displayed so that users are not surprised by cases where spaces cause silent errors.

drgrice1:
@Alex-Jordan:  I would like to get this into the upcoming release.  Many of the places where sets were displayed with underscores have already been removed.  This can be done immediately to any parameter that involves a set being entered by an instructor into a text input.  There is no chance of silent failure (as @mgage put it) if it is done to the parameter value when it is assigned (checking definedness as needed).  The `format_set_name` method that is added in this pull request needs to have a more concise name.  That name will only serve to cause confusion.  `format_set_name` for what?  It is formatting it for storage in the database or internal use, but that name does not convey that.  Seeing the name I would assume the method is formatting for display purposes.

In any case, if you would like to update this pull request and put it in then I will let you do so.  If not, then I will implement this.

drgrice1:
This can be closed as it is now in the code.
Thanks for starting the work in this @Alex-Jordan.

PR #1413, opened 2021-06-14
Export confirm

Alex-Jordan:
This started out because it has always bothered me that when I export problem sets, I must pass through a secondary confirmation screen. It seems unnecessary.

But it was too much to remove that confirmation step, and besides, the community may want to keep it. Still, I found the wording of messages misleading. Faculty here occasionally don't understand why their export "failed" and it turns out that it's just because they did not click Take Action on the confirmation screen. So this changes some of the wording when you are using the Export tab to try to make it more clear that a second step is needed.

Then I notices all these extra space characters and colons, and crowding of the input fields butted up against their labels. So there is some styling changes to that.

And after all that, I went through to do similar things with the classlist editor.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -1453,7 +1453,7 @@ table#grades_table pre{
     margin-bottom:10px;
 }
 
-.tab-pane select, input {
+.tab-pane select, .tab-pane input {
 	margin-left: 0.5rem;
 }
 
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -1453,6 +1453,10 @@ table#grades_table pre{
     margin-bottom:10px;
 }
 
+.tab-pane select, input {
+	margin-left: 0.5rem;
+}
+
 .small-table-text,
 .small-table-text td {
     font-size:small;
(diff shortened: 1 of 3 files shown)


htdocs/themes/math4/math4.css
```diff
@@ -1453,6 +1453,10 @@ table#grades_table pre{
     margin-bottom:10px;
 }
 
+.tab-pane select, input {
+	margin-left: 0.5rem;
+}
```
drgrice1, line 1458: I think that you meant `.tab-pane select, .tab-pane input` for the selector.  Otherwise this applies to all inputs.

drgrice1: Here are comparative screenshots to show that this does affect something negatively.  The first without this pull request and the second with this pull request.

cc-wo

cc-w

drgrice1 (approved):
This looks good other than the minor css issue.

htdocs/themes/math4/math4.css
```diff
@@ -1453,6 +1453,10 @@ table#grades_table pre{
     margin-bottom:10px;
 }
 
+.tab-pane select, input {
```
Alex-Jordan, line 1456:
```suggestion
.tab-pane select, .tab-pane input {
```

htdocs/themes/math4/math4.css
Alex-Jordan: If I understand, that is just because of my CSS oversight, yes? I just edited it here in GitHub.

drgrice1:
Looks good.

pstaabp (approved):
looks good.

taniwallach:
Merging.
@drgrice1 This one also has some `maketext` changes. We'll need to redo the translation files again.

drgrice1:
I will do this.  Thanks @taniwallach.

PR #1414, opened 2021-06-14
use a download icon atop the column of download controls

Alex-Jordan:
From the instructor view in the problem sets page, the column of checkboxes for download has no indication of its purpose (unless you read the dowload button that is down and to the left and figure it out). This puts an icon with tooltip text atop that column, which helps and does not take up as much space as a word would take up. The same icon gets placed there for student view, which I think is harmless.

Also I shortened "Download PDF or TeX Hardcopy for Selected Sets" to just "Download Hardcopy for Selected Sets". That button is very wide.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -468,8 +468,13 @@ table.classlist-table .tbody {
     padding-left: 35px;
 }
 
-.problem_set_table .hardcopy-link {
-    font-size:16px;
+.problem_set_table .hardcopy, .problem_set_table .hardcopy-link {
+	font-size: 16px;
+	text-align: center;
+}
+
+.problem_set_table .hardcopy input[type="checkbox"] {
+	margin-right: 0;
 }
 
 /* Info Box */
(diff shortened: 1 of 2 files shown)


pstaabp:
I might suggest that the button (and tooltip) is "Generate Hardcopy for Select Sets", since clicking the button actually takes to the generate page, not download.

Alex-Jordan:
I changed the button to say "Generate Hardcopy for Select Sets". I changed the tooltip to just say "Generate Hardcopy" because it is the same icon and tooltip that students see, and for them there are no "selected sets". They can only do one at a time.

pstaabp (approved):
Looks good now.

mgage:
@drgrice1 we'll need to rebuild the .po files after this pull request.  The labels on some buttons have been changed.

I'd like to make a detailed description of how to scrape the code and find out which strings need to be changed in the .pot file.  Can you help with this -- my experience is out of date. If we can automate this even better.  --- Mike

drgrice1:
It is all completely automated already.  Just run the script update_localization_files.

PR #1416, opened 2021-06-14
log when a user actively logs out

Alex-Jordan:
This makes it so that when a user clicks the button to log out, there is entry in the login.log for that.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Logout.pm
+++ b/lib/WeBWorK/ContentGenerator/Logout.pm
@@ -45,7 +45,7 @@ sub pre_header_initialize {
 #	}
 
 	$authen -> killSession;
-	$self->WeBWorK::Authen::write_log_entry("LOGGED OUT");
+	$authen->WeBWorK::Authen::write_log_entry("LOGGED OUT");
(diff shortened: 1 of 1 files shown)


pstaabp:
The log file does say "LOGGED OUT", but there's no information about who was logged out.  Do you want to log it before ending the session, so perhaps the `user_id` is present on the log.

Alex-Jordan:
Good catch, I assumed that `write_log_entry` was handling that, but I guess `$self` here is not the right thing. I'll work on it.

Alex-Jordan:
OK, one small change to get the right thing calling to write to the log. Now I get entries like:

```
[Tue Jun 15 22:19:28 2021] LOGGED OUT user_id=profa login_type=normal credential_source=cookie host=10.14.0.254 port=62738 UA=Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:89.0) Gecko/20100101 Firefox/89.0
```

pstaabp (approved):
Looks good now.

drdrew42 (approved):

PR #1417, opened 2021-06-14
change the name of the answer log file from answer_log to answer.log

Alex-Jordan:
I think it has just been a typo that the name of the answer log file is `answer_log`. This changes it to `answer.log`. If a course already has an `answer_log` file, that file will stay there, but new entries will be added to `answer.log`.


taniwallach (approved):
This idiosyncrasy has always been somewhat annoying to me.
I'm all for changing the name.

pstaabp (approved):

drgrice1 (approved):

PR #1418, opened 2021-06-14
copy simple.conf when making a new course

Alex-Jordan:
This makes it so that when you create a new course from the admin course, there is an option to copy the simple.conf file along with the templates folder.  It is checked by default.

Note I would like to do the same thing with the course.conf file, but it's not as simple as just copying that one. That one has this in it:
```
$pg{specialPGEnvironmentVars}{PRINT_FILE_NAMES_FOR}
```

which has the course instructor listed. So you don't want to copy course.conf and overwrite the original course.conf, in case it's a new instructor. If it were OK to take that out of course.conf (and manage PRINT_FILE_NAMES_FOR as a permission) then I could copy both config files this way. Would there be any objections to that?

Copying config files is nice for an instructor who has some settings they like and use term after term. It would be nice to fully clone their old courses when making a new course for them.

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -323,6 +323,10 @@ h2.page-title {
     text-align:center;
 }
 
+.new-course-options label span {
+	vertical-align: middle;
+}
+
 /* Home Page */
 ul.courses-list {
 	list-style-type: none;
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -510,9 +510,21 @@ sub add_course_form {
 		),
 	);
 	
-	print CGI::p($r->maketext("To add the WeBWorK administrators to the new course (as administrators) check the box below."));
 	my @checked = ($add_admin_users) ?(checked=>1): ();  # workaround because CGI::checkbox seems to have a bug -- it won't default to checked.
-	print CGI::p({},CGI::input({-type=>'checkbox', -name=>"add_admin_users", @checked }, $r->maketext("Add WeBWorK administrators to new course")));
+	print CGI::div(
+		{class => 'new-course-options'},
+		CGI::p($r->maketext("To add the WeBWorK administrators to the new course (as administrators) check the box below.")),
+		CGI::label(
+			{for => "add_admin_users"},
+			CGI::input({-type=>'checkbox', -name=>"add_admin_users", @checked }, ''),
+			CGI::span($r->maketext("Add WeBWorK administrators to new course"))
+		),
+		CGI::label(
+			{for => "add_config_file"},
+			CGI::input({-type=>'checkbox', -name=>"add_config_file", @checked }, ''),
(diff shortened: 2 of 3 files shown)


taniwallach:
I think it would be better to keep support for the old `$pg{specialPGEnvironmentVars}{PRINT_FILE_NAMES_FOR}` for now, as removing it would change the behavior of existing courses in a manner which might surprise many "teaching staff" end-users.

However, if we added support for a permission to provide this setting, then a typical `course.conf` file could omit this.

Even with this setting - I suspect that it is not that terrible to copy `course.conf` and "warn" the admin course user that it may be advisable to remove the setting in the new course.

Alex-Jordan:
I just noticed that it's possible for a course to not have a `simple.conf` file, and then if this option is checked, there is an error message. I'll circle back to fix that.

Alex-Jordan:
I addressed the thing from my last comment.

Regarding copying `course.conf`, I think that could be done later. It will take more effort than I can prioritize for a while, and just copying `simple.conf` seems useful enough. And I think it should be harmless to expand to `course.conf` if/when the time comes.

htdocs/themes/math4/math4.css
```diff
@@ -323,6 +323,10 @@ h2.page-title {
     text-align:center;
 }
 
+.new-course-options label span {
+	vertical-align: middle;
+}
+
```
drgrice1, line 329: This doesn't work as expected.  I assume you are trying to make the label and checkbox align better.   For Google Chrome it improves this a bit, but for Firefox it doesn't change a thing.  The best that I have found for this in this case that is good for both of these browsers is:
```suggestion
.new-course-options label.checkbox input {
	margin-top: 0;
	float: unset;
	vertical-align: unset;
}
```

I can't guarantee this be an improvement on other browsers though.

The real problem here is bootstrap 2.3.2.  It is adding style that doesn't work right, and causes the misalignment.  That style may have worked 5 years ago, but browsers have changed.  The style changes I put above are undoing what bootstrap does in this instance.  The proper set up for browsers now is to not have the input inside its own label.  Newer versions of bootstrap are updated for that.  Although, to get that to work correctly you also have to add their form-control styles.

I am not sure that it is worth it to fix the appearance of these checkboxes and bootstrap 2.3.2's problems.  The other labels and inputs on the page are badly misaligned as well, so this is a very minor improvement.

drgrice1 (approved):
This looks good.  The simple.conf copy option works as expected.

Alex-Jordan:
> but for Firefox it doesn't change a thing

That's not what I am seeing. If I toggle this in Firefox using the inspector I get a change:

## with

Screen Shot 2021-06-18 at 3 53 44 PM

## without

Screen Shot 2021-06-18 at 3 53 53 PM

htdocs/themes/math4/math4.css
Alex-Jordan: Agreed that this is not worth too much effort/investigation. It's just for a user of the admin page to see.

If you confirm that your suggestion is indeed better than the `vertical-align`, I'll accept that change here. But as I noted, for me on a Mac with Firefox, the `vertical-align` makes it work.

PR #1421, opened 2021-06-16
Update docker-entrypoint.sh

mgage:
changed "passwordRW" to 'passwordRW' in  -e 's/database_password ='passwordRW'/database_password =$ENV{"WEBWORK_DB_PASSWORD"}/' \
literally specifying passwords in the configuration files allowing interpolation wrecks havoc if the  password has special characrers such as @ and %
so the quoting of passwordRW must match the quotes used in site.conf.dist (which are single quotes)


Alex-Jordan (approved):

pstaabp (approved):

drgrice1 (changes requested):
You will need to escape the single quotes for this to work.  This takes a little care to work in this case.  Look at the other escaped single quotes on the next line to see how this should be done.

PR #1423, opened 2021-06-16
Ww216 patches admincourse

mgage:
if admin course is added then its tables are also updated and  user: admin with password: admin are added to the admin course.

### DIFF
--- a/docker-config/docker-entrypoint.sh
+++ b/docker-config/docker-entrypoint.sh
@@ -76,8 +76,8 @@ if [ "$1" = 'apache2' ]; then
                     -e 's/^\$database_host="localhost"/$database_host = $ENV{"WEBWORK_DB_HOST"}/' \
                     -e 's/^\$database_port="3306"/$database_port = $ENV{"WEBWORK_DB_PORT"}/' \
                     -e 's/^\$database_name="webwork"/$database_name = $ENV{"WEBWORK_DB_NAME"}/' \
-                    -e 's/database_username ="webworkWrite"/$database_username =$ENV{"WEBWORK_DB_USER"}/' \
-                    -e 's/database_password ='\''passwordRW'\''/$database_password =$ENV{"WEBWORK_DB_PASSWORD"}/' \
+                    -e 's/database_username ="webworkWrite"/database_username =$ENV{"WEBWORK_DB_USER"}/' \
+                    -e 's/database_password ='\''passwordRW'\''/database_password =$ENV{"WEBWORK_DB_PASSWORD"}/' \
                     -e 's/mail{smtpServer} = '\'''\''/mail{smtpServer} = $ENV{"WEBWORK_SMTP_SERVER"}/' \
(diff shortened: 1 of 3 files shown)


taniwallach:
I have never needed to do anything to have the `admin` user created. It seems to be handled by the load from the class list file in
```
$WEBWORK_ROOT/bin/addcourse admin --db-layout=sql_single --users=$WEBWORK_ROOT/courses.dist/adminClasslist.lst --professors=admin
```
My guess is that you may have manually needed to fix the DB if the DB connection was not working when the course was created.

mgage:
I'll try it again to double check and report back before this is checked in.  I don't think it was because of a bad connection to the database.

mgage:
@taniwallach I think you are right, if the addcourse admin script runs completely the admin user and its credentials are also installed.

docker-config/docker-entrypoint.sh
```diff
@@ -76,8 +76,8 @@ if [ "$1" = 'apache2' ]; then
                     -e 's/^\$database_host="localhost"/$database_host = $ENV{"WEBWORK_DB_HOST"}/' \
                     -e 's/^\$database_port="3306"/$database_port = $ENV{"WEBWORK_DB_PORT"}/' \
                     -e 's/^\$database_name="webwork"/$database_name = $ENV{"WEBWORK_DB_NAME"}/' \
-                    -e 's/database_username ="webworkWrite"/database_username =$ENV{"WEBWORK_DB_USER"}/' \
-                    -e 's/database_password ="passwordRW"/database_password =$ENV{"WEBWORK_DB_PASSWORD"}/' \
+                    -e 's/^\$database_username ="webworkWrite"/$database_username =$ENV{"WEBWORK_DB_USER"}/' \
+                    -e 's/^\$database_password ='\''passwordRW'\''/$database_password =$ENV{"WEBWORK_DB_PASSWORD"}/' \
```
taniwallach, line 80: This line is a mission critical correction, as the line in `site.conf.dist` was changed to use single quotes.
The change on these 2 lines to match from the start of the line so it should only modify the active definition lines and not other lines is also good.

docker-config/docker-entrypoint.sh
```diff
@@ -96,6 +96,9 @@ if [ "$1" = 'apache2' ]; then
         $WEBWORK_ROOT/bin/addcourse admin --db-layout=sql_single --users=$WEBWORK_ROOT/courses.dist/adminClasslist.lst --professors=admin
         chown www-data:www-data -R $APP_ROOT/courses
         echo "Admin course is created."
+	$WEBWORK_ROOT/bin/upgrade_admin_db.pl
```
taniwallach, line 99: I have yet to be able to trigger a case where these extra lines are needed.
If I delete the course directory, but left the database tables alone - the course seems to work after being recreated by the prior call to `bin/addcourse`.
My impression is that these lines can be deleted (or commented out, if @mgage want's to retain them in case some special circumstances do make them necessary).

taniwallach (changes requested):
As noted, I think the extra lines to make a second attempt to fix the admin course tables and to add the admin user are not needed.

mgage:
this has been addressed in patch5

PR #1429, opened 2021-07-01
Localization update (one last time)!

drgrice1:
This should signify a string freeze before the release I think.


taniwallach:
@drgrice1 - sorry, but there are still string changes pending:
  - https://github.com/openwebwork/webwork2/pull/1431 has one about the 100 character limit. We could drop the `maketext` there for now if that helps.
  - If we try to get https://github.com/openwebwork/webwork2/pull/1130 done - it seems to also have some.

taniwallach (approved):
These are updated translation files @drgrice1 generated from the recent code changes.
In order to prevent breaking foreign languages - we want these files updated before the release.

PR #1430, opened 2021-07-01
Stop making ping calls to webwork.maa.org during Apache startup

taniwallach:
Drop the code which triggered a ping to webwork.maa.org on Apache startup.

That was added in https://github.com/openwebwork/webwork2/commit/60c520392859594181d56186d48e5f832918713f
as part of https://github.com/openwebwork/webwork2/pull/976

It does not seem necessary to force sites to send those `ping` packets automatically, and expecting people to realize it is happening and to opt-out is unlikely to occur.

### DIFF
--- a/conf/webwork.apache2-config.dist
+++ b/conf/webwork.apache2-config.dist
@@ -122,19 +122,6 @@ eval "use lib '$pg_dir/lib'"; die $@ if $@;
 
 require Apache::WeBWorK; # force compilation of pretty much everything
 
-
-print "Check in with MAA site\n";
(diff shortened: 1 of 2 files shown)


drgrice1:
I agree with this change.

@mgage: Are you okay with this?  I don't think that the webwork.maa.org is set up to handle these pings anymore in any case.

mgage:
I am ok with this change.  We'll figure out a better way to keep track of WW sites.

mgage (approved):
I'm ok with these changes.  we need a better method of keeping track of WW sites than this ping method provided

drgrice1 (approved):
Looks good.

PR #1431, opened 2021-07-01
Past answer table: resize fields + limit allowed length of a set name accordingly

taniwallach:
@dlglin pointed out in the meeting this week that the change to the field lengths in the `past_answer` table causes problems for instructors who are used to being able to use the previously allowed 100 characters in their set names.

This PR is intended to change the limit back to what it used to be, to prevent such issues.

The field lengths were changed in 
https://github.com/openwebwork/webwork2/commit/b840e28f802dd69f6251aa4c4096bcca43c2ad90 in order for the total key-length to remain under the limit when `utf8mb4` is in use (as mySQL/MariaDB will reserve 4 bytes per character when `utf8mb4` is in use).

Both `set_id` and `user_id` were returned to the prior `VARCHAR(100)`. In other tables these are `TINYBLOB`which can actually hold 255 bytes.

In order to remain under the key length limit, `course_id` was reduced down to ` VARCHAR(40)` (from the current `VARCHAR(100)` which was also`VARCHAR(100)` before WeBWorK-2.15).

Note: This fields does not appear in any other course table, and ideally would not be here either. However, there is existing code which expects this column to exist.

For practical purposes it seems unlikely that this change can break anything in an operational course. mySQL has a limit of 64 characters for a table name (https://dev.mysql.com/doc/refman/8.0/en/identifier-length.html) and the longest table name suffix in use in WeBWorK is `_global_user_achievement` (24 characters) which effectively limits the `course_id` of a fully functional course to 40 characters.

Under standard settings, this is certainly sufficiently long, as the admin course limits the length of the `course_id` which can be created using the admin interface using `-maxlength=>$ce->{maxCourseIdLength}` in the attributes on the input box. By default this is set by `conf/defaults.config` to 40 characters. Creating a course called `MaxLongNameViaAdmin123456789012345678901` (exactly 40 characters) seems to have worked properly: all 15 database tables were created, and that course works properly.

Although in principle that limit may have been overridden on some installations, and use of the `bin/addcourse` script does not set a limit on the length of the `course_id`, the mySQL limit of 64 characters in a table name should mean that any attempt to create a course with a `course_id` of over 40 characters will fail to work.

For example, an attempt to create a course with 41 characters in the name using `bin/addcourse` triggers an error, and 
```
root@localhost:/opt/webwork/webwork2# bin/addcourse --users=../courses/adminClasslist.lst MaxLongNameViaAdmin1234567890123456789012
addcourse:  WeBWorK root directory set to /opt/webwork/webwork2
addcourse:  WeBWorK server is starting
addcourse:  WeBWorK root directory set to /opt/webwork/webwork2 in webwork2/conf/webwork.apache2-config
addcourse:  The following locations and urls are set in webwork2/conf/site.conf
addcourse:  PG root directory set to /opt/webwork/pg
addcourse:  WeBWorK server userID is www-data
addcourse:  WeBWorK server groupID is www-data
addcourse:  The webwork url on this site is http://localhost/webwork2
DBD::MariaDB::db do failed: Incorrect table name 'MaxLongNameViaAdmin1234567890123456789012_global_user_achievement' at /opt/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm line 891.
```
and so did an attempt to do so using the admin course:
```
An error occured while creating the course AMaxLongNameViaAdmin123456789012345678901:

DBD::MariaDB::db do failed: Incorrect table name 'AMaxLongNameViaAdmin123456789012345678901_global_user_achievement' at /opt/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm line 891.
```

Also note that since this field is really of no value as this is a per-course table, any data-loss due to possible truncation is really not important.

The only potential problem I can think of is that some code uses a version if `listProblemPastAnswers` of the form `listProblemPastAnswers($courseID,$userID,$setID,$problemID)` where the `course_id` is used as a `where` condition on the database all.  If someone was able to create a functional course with a `course_id` of over 40 characters (and probably missing the `_global_user_achievement` table) - the code may misbehave when processing where conditions with the full `course_id`.

---

On the other side of things, the system did not enforce that a `set_id` should not exceed 100 characters. Creating such a set was possible before the code in this PR, but when an answer is submitted an error occurs:
```
Error messages

DBD::MariaDB::st execute failed: Data too long for column 'set_id' at row 1 
```

This PR both uses a `maxlength` parameter on the input box, and checks `length($newSetID)` inside the `create_handler` to prevent a set being created with a `set_id` of over 100 characters.

**Future work**: Handling the import functionality to enforce the limit.

### DIFF
--- a/bin/addcourse
+++ b/bin/addcourse
@@ -189,6 +189,11 @@ $ce = WeBWorK::CourseEnvironment->new({
 	courseName => $courseID
 });
 
+# Do not attempt to create a course whose name is too long.
+if ( length($courseID) > $ce->{maxCourseIdLength} ) {
+	die "Aborting addcourse: Course ID cannot exceed " . $ce->{maxCourseIdLength} . " characters.";
+}
+
 if ($dbLayout) {
 	die "Database layout $dbLayout does not exist in the course environment.",
 			" (It must be defined in defaults.config.)\n"
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -576,11 +576,13 @@ $dbLayoutName = "sql_single";
 *dbLayout     = $dbLayouts{$dbLayoutName};
 
 # This sets the max course id length.  It might need to be changed depending
-# on what database tables are present.  Mysql allows a max table length of 68
+# on what database tables are present.  Mysql allows a max table length of 64
 # characters.  With the ${course_id}_global_user_achievement table that means
-# the max ${course_id} is about 40 characters
+# the max ${course_id} is exactly 40 characters.
 
 $maxCourseIdLength = 40;
+
+# Reference:  https://dev.mysql.com/doc/refman/8.0/en/identifier-length.html
 
 ################################################################################
 # Problem library options
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -380,7 +380,7 @@ sub body {
 	);
 	
 	print( CGI::p({style=>"text-align: center"}, $self->display_registration_form() ) ) if $self->display_registration_form();
-	
+
 	my @errors = @{$self->{errors}};
 	
 	
@@ -493,7 +493,7 @@ sub add_course_form {
 	print $self->hidden_authen_fields;
 	print $self->hidden_fields("subDisplay");
 	
-	print CGI::p($r->maketext("Specify an ID, title, and institution for the new course. The course ID may contain only letters, numbers, hyphens, and underscores."));
+	print CGI::p($r->maketext("Specify an ID, title, and institution for the new course. The course ID may contain only letters, numbers, hyphens, and underscores, and may have at most [_1] characters.", $ce->{maxCourseIdLength}));
 	
 	print CGI::table({class=>"FormLayout"},
 		CGI::Tr({},
@@ -653,7 +653,10 @@ sub add_course_validate {
 	if (grep { $add_courseID eq $_ } listCourses($ce)) {
 		push @errors, $r->maketext("A course with ID [_1] already exists.", $add_courseID);
 	}
-	
+	if ( length($add_courseID) > $ce->{maxCourseIdLength} ) {
+                @errors, $r->maketext("Course ID cannot exceed [_1] characters.", $ce->{maxCourseIdLength});
+	}
+
 	if ($add_initial_userID ne "") {
 		if ($add_initial_password eq "") {
 			push @errors, $r->maketext("You must specify a password for the initial instructor.");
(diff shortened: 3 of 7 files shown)


drgrice1 (approved):
This looks good.

taniwallach:
@pstaabp @mgage @dlglin - Anyone else planning to test this?

pstaabp:
When I run the `addCourse` script as above, I do get the error, however, two tables are still made in the database (I'm running MariaDB 10.5.9):

```
MaxLongNameViaAdmin1234567890123456789012_password
MaxLongNameViaAdmin1234567890123456789012_setting
```

There is also a directory made in the `courses` directory.   I didn't see an error when using the CourseAdmin page, however the database names are truncated--I think as you indicated this is truncated by the INPUT html box.  The course seems to work fine.  

If we're worried about the course_id too long via the addCourse script, we should probably put in a limit on that. 

The same is true with set_id in that using the UI, I wasn't able to exceed the 100 characters.

taniwallach:
@pstaabp - Thanks for the feedback. I did some more work to try to prevent bypassing/ignoring the limit in an additional commit. Please take a look after the changes.

---

> When I run the `addCourse` script as above, I do get the error, however, two tables are still made in the database (I'm running MariaDB 10.5.9):

> If we're worried about the course_id too long via the addCourse script, we should probably put in a limit on that.

I had not tried to fix the `addCourse` script, just indicated that it would not work properly for names which are too long. 

I just added a length check which will cause the script to `die` before it does anything. That will prevent partial work by the script.

---

> ... I didn't see an error when using the CourseAdmin page, however the database names are truncated--I think as you indicated this is truncated by the INPUT html box. The course seems to work fine.

The input box has a length limit, so would prevent passing the limit (essentially truncating what you might have tried to paste in) and that should prevent problems under typical use. So were the input boxes for a new name for the rename and unarchive operations. 

I just added code to test the length and to try to warn and fail if the name is too long in any of these.

@drgrice1 - Sorry that means there are `maketext` changes again.

In order to test the new limits - I modified the `maxlength` limit on the input box for adding a course and then entered too many characters. It now reports
```
An error occured while creating the course MaxLongNameViaAdmin1234567890123456789012:

Course ID cannot exceed 40 characters. at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/CourseAdmin.pm line 801.
```
but the also issues
```
WeBWorK Warnings

WeBWorK has encountered warnings while processing your request. If this occured when viewing a problem, it was likely caused by an error or ambiguity in that problem. Otherwise, it may indicate a problem with the WeBWorK system itself. If you are a student, report these warnings to your professor to have them corrected. If you are a professor, please consult the warning output below for more information.

Warning messages
Warning: the course root directory '/opt/webwork/courses/MaxLongNameViaAdmin1234567890123456789012' does not exist. Attempting to delete the course database and other course directories...
```

The secondary error is apparently caused when a failed add leads to an attempt to run a delete. Since someone needs to override or otherwise bypass the `maxlength` limit on the input box to even try to use a too long name, I think it is good enough this way.

For unarchive and rename, the new error messages behave more nicely (there is no secondary error).

---

> The same is true with set_id in that using the UI, I wasn't able to exceed the 100 characters.

And for `set_id` there was already an error message being generated if someone tried to "hack" past the `maxlength` limit, as I had just added the enforcement.

---

**Note:** There is special code in the `_unarchiveCourse_move_away` subroutine in `lib/WeBWorK/Utils/CourseManagement.pm` which I did not try to fix. It is used to try to move aside an existing course directory when unarchiving a course which will be renamed, when the name from the `tar.gz` file name is in use by an existing course. That whole mode of operations is dangerous (moving aside an existing course briefly), and the manner in which a temporary name is selected can exceed the length limit. The whole concept of how this is done needs to be reconsidered. It might be reasonable to unarchive in a temporary directory, rename the directory, and then move it into the course directory, before continuing on. For now, this was just set aside.

taniwallach:
Note: I tested the changes to `bin/addcourse` and to the admin course (add, rename, unarchive to new name).
However, I did not test the new code in `lib/WebworkWebservice/CourseActions.pm` as I'm not really sure how to easily get that to run. Since it is a new `if` block similar to the prior ones, it should work.

pstaabp (approved):
Looks good now.

PR #1434, opened 2021-07-05
Modify listCourses to also test for the existence of the course_user table

taniwallach:
In `listCourses` test also for the existence of the `course_user` table and ignore directories for which there is no such table.

I chose the `course_user` table, as that is one whose absence will quickly cause problems.

This is intended to address the issue that the current code would assume that any subdirectory in the courses directory corresponded to a "real" course. 

That issue also was mentioned in https://github.com/openwebwork/webwork2/pull/1424 as `bin/update-OPL-statistics`  misbehaves where there are subdirectories which do not correspond to courses. Several other parts of the code-base will have similar various problems handling incorrect names from the old code.

For example, until this PR, the code which generates the list of courses for the "home page" at "/webwork2" on the server could include links for names of subdirectories which are not valid courses.

### DIFF
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -83,7 +83,46 @@ Lists the courses defined.
 sub listCourses {
 	my ($ce) = @_;
 	my $coursesDir = $ce->{webworkDirs}->{courses};
-	return grep { not (m/^\./ or m/^CVS$/) and -d "$coursesDir/$_" } readDirectory($coursesDir);
+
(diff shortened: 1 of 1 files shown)


mgage (commented):
code looks clear and good.  I'll pull and test it later today.

mgage (approved):
worked well for me.  I introduced a few empty directories and they did not appear in the list.  There was no noticeable speed hit. The empty directories also did not appear in the admin view of courses

pstaabp (approved):

drgrice1 (approved):
I think this is good.  I tested it quickly, and it works as promised.

PR #1435, opened 2021-07-05
Patch5 for docker two stage

mgage:
I believe this is in good enough shape to be looked at. I have not yet done all the tests
that I want to do to make sure it's solid.  I'm open to suggestions for improvements.

htdocs/tmp is now owned by www-data
if either the admin directory or the admin course tables are missing then they are replaced.

for the time being I've put off including install instructions -- we can discuss that later.


docker-config/docker-entrypoint.sh
```diff
@@ -82,21 +82,58 @@ if [ "$1" = 'apache2' ]; then
...
         cd $APP_ROOT/courses
         wait_for_db
         $WEBWORK_ROOT/bin/addcourse admin --db-layout=sql_single --users=$WEBWORK_ROOT/courses.dist/adminClasslist.lst --professors=admin
         chown www-data:www-data -R $APP_ROOT/courses
         echo "Admin course is created."
+        echo "user: admin password: admin added to course admin and tables upgraded"
+    elif [ ! $ADMIN_TABLE_EXISTS ]; then
+        echo "admin course db tables need updating"
+        $WEBWORK_ROOT/bin/upgrade_admin_db.pl
+        $WEBWORK_ROOT/bin/wwsh admin ./addadmin
```
taniwallach, line 116: This does not work as is. If the `admin_user` table is dropped, then on the `docker-compose up` the table will be recreated, but the admin user is not added.

I think it needs to be `$WEBWORK_ROOT/bin/wwsh admin $WEBWORK_ROOT/bin/addadmin` or
```
cd $WEBWORK_ROOT/bin/
./wwsh admin ./addadmin
```

docker-config/docker-entrypoint.sh
```diff
@@ -158,7 +195,7 @@ if [ "$1" = 'apache2' ]; then
         echo "Restoring OPL tables from the TABLE-DUMP/OPL-tables.sql file"
         wait_for_db
         $WEBWORK_ROOT/bin/restore-OPL-tables.pl
-        $WEBWORK_ROOT/bin/load-OPL-global-statistics.pl
+        #$WEBWORK_ROOT/bin/load-OPL-global-statistics.pl
```
taniwallach, line 198: I do not think that the call to `$WEBWORK_ROOT/bin/load-OPL-global-statistics.pl` should be commented out.

A test for whether the table exists can be done and then the invocation can be conditional on the table being missing.

If that table fails to exist (and this is the only place that `docker-entrypoint.sh` would create it, then attempting to view some problems via the library browser fails and reports:
```Warning messages

    Couldn't find the OPL global statistics table. Did you download the latest OPL and run load-OPL-global-statistics.pl? at /opt/webwork/webwork2/lib/WeBWorK/Utils/LibraryStats.pm line 95. 

Error messages

Table 'webwork.OPL_global_statistics' doesn't exist at /opt/webwork/webwork2/lib/WeBWorK/Utils/LibraryStats.pm line 97.
Call stack

The following information can help locate the source of the problem.

    in WeBWorK::Utils::LibraryStats::getGlobalStats called at line 1043 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
    in WeBWorK::ContentGenerator::Instructor::SetMaker::make_data_row called at line 1671 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
    in WeBWorK::ContentGenerator::Instructor::SetMaker::body called at line 155 of /opt/webwork/webwork2/lib/WeBWorK/Template.pm
    in WeBWorK::Template::template called at line 577 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
    in WeBWorK::ContentGenerator::content called at line 232 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
    in WeBWorK::ContentGenerator::go called at line 386 of /opt/webwork/webwork2/lib/WeBWorK.pm
```

docker-compose.yml
```diff
@@ -225,7 +225,7 @@ services:
       #ADD_LOCALES: "he_IL ISO-8859-8\nhe_IL.UTF-8 UTF-8\n"
 
       # Extra Ubuntu packages to install during startup
-      #ADD_APT_PACKAGES: vim telnet
+      ADD_APT_PACKAGES: vim less
```
taniwallach, line 228: I would recommend leaving the `ADD_APT_PACKAGES` line commented out, and explaining how to use it in the documentation.

Adding packages here causes a significant start-up delay for them to be installed. Users who need to access the shell inside a running docker container can either install things when needed, or can make the decision to pay the penalty in start-up time.

Installing extra `support` related packages to the base image is not considered to be appropriate due to DevOps concerns about keeping the image as small/clean as possible. (There are old posts from @xcompass about that, and the consensus seemed to be to do so.)

taniwallach (changes requested):
In my testing there are several bugs.
  - The `admin_user` table is not being recreated due to an error in the `bash` script condition.
  - The `wwsh` command is not executing what is intended when it would be called.
  - Dropping the global statistics table leads to it not be recreated, as the relevant line was commented out.
Corrections are proposed in the "line comments".

taniwallach:
Even with the patches discussed in my review, there is still an issue with the "local statistics" being missing, which triggers an error:
```
Warning messages

    Couldn't find the OPL local statistics table. Did you download the latest OPL and run update-OPL-statistics.pl? at /opt/webwork/webwork2/lib/WeBWorK/Utils/LibraryStats.pm line 69. 

Error messages

Table 'webwork.OPL_local_statistics' doesn't exist at /opt/webwork/webwork2/lib/WeBWorK/Utils/LibraryStats.pm line 71.
Call stack

The following information can help locate the source of the problem.

    in WeBWorK::Utils::LibraryStats::getLocalStats called at line 1057 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
    in WeBWorK::ContentGenerator::Instructor::SetMaker::make_data_row called at line 1671 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
    in WeBWorK::ContentGenerator::Instructor::SetMaker::body called at line 155 of /opt/webwork/webwork2/lib/WeBWorK/Template.pm
    in WeBWorK::Template::template called at line 577 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
    in WeBWorK::ContentGenerator::content called at line 232 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
    in WeBWorK::ContentGenerator::go called at line 386 of /opt/webwork/webwork2/lib/WeBWorK.pm
```

I think that for "developers" and "simple" Docker use, the best  solution is probably to disable local statistics by adding
```
$problemLibrary{showLibraryLocalStats} = 0;
```
into `conf/localOverrides.conf`.

taniwallach:
> Doing a find/replace on the config files is fragile, as evidenced by the changes to the password quoting. For changes to localOverrides.conf, would it make more sense to either have the docker build generate that file from scratch (which should be safe since the .dist version is entirely commented out), or simply add new config options at the end of the file (before the closing line)? This way if the text changes in the .dist version it wouldn't break the find/replace.

About how to efficiently and safely modify `conf/localOverrides.conf` it might be simplest to try using a `patch` file to modify the end of the file, or just to put a minimal file (without all the commented out lines) under `docker-config` and copy it into the image with a `COPY` line.

docker-config/docker-entrypoint.sh
```diff
@@ -82,21 +82,58 @@ if [ "$1" = 'apache2' ]; then
...
+                sed -i -e 's/#$pg{specialPGEnvironmentVars}{Rserve} = {host => "r"};/$pg{specialPGEnvironmentVars}{Rserve} = {host => "r"};/' $WEBWORK_ROOT/conf/localOverrides.conf
+                echo "$WEBWORK_ROOT/conf/$i has been modified."
             fi
         fi
+        
     done
     # create admin course if not existing
-    if [ ! -d "$APP_ROOT/courses/admin"  ]; then
+    # check first if the admin courses directory exists then check that at 
+    # least one of the ables associated with the course (the admin_user table) exists
```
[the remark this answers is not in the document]
mgage: ty

docker-config/docker-entrypoint.sh
```diff
@@ -82,21 +82,58 @@ if [ "$1" = 'apache2' ]; then
...
+#    echo mysql -u $ENV{WEBWORK_DB_USER}  -pENV{$WEBWORK_DB_PASSWORD} -B -N -h db -e "select count(*) from information_schema.tables where table_schema='webwork' and table_name = 'admin_user';"  2>/dev/null
+#    
+   
+    
+    # use case for the extra check for the admin:
+    # In rebuilding a docker box one might clear out the docker containers, 
+    # images and volumes including mariaDB, BUT leave the 
+    # contents of ww-docker-data directory in place.  It now holds the shell of the courses 
+    # including the admin course directory. This means that once you rebuild the box 
+    # you can't access the admin course (be cause the user_admin table is missing)
```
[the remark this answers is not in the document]
mgage: ty

docker-config/docker-entrypoint.sh
```diff
@@ -82,21 +82,58 @@ if [ "$1" = 'apache2' ]; then
...
+    elif [ ! $ADMIN_TABLE_EXISTS ]; then
+        echo "admin course db tables need updating"
+        $WEBWORK_ROOT/bin/upgrade_admin_db.pl
+        $WEBWORK_ROOT/bin/wwsh admin ./addadmin
+        echo "admin course tables created with one user: admin   whose password is admin"
+    else 
+        echo "using pre-existing admin course and admin tables"
     fi
+    
+#    echo $WEBWORK_ROOT/bin/courseUserTableExists.sh admin  $WEBWORK_DB_USER $WEBWORK_DB_PASSWORD
```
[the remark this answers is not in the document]
mgage: I've removed them.
They were used to debug the script courseUserTableExists.sh in webwork2/bin -- which we were never able to get to run properly inside docker. We gave up.

docker-config/docker-entrypoint.sh
```diff
@@ -82,21 +82,58 @@ if [ "$1" = 'apache2' ]; then
...
+    if [ ! -d "$APP_ROOT/courses/admin" ]; then
         newgrp www-data
         umask 2
         cd $APP_ROOT/courses
         wait_for_db
         $WEBWORK_ROOT/bin/addcourse admin --db-layout=sql_single --users=$WEBWORK_ROOT/courses.dist/adminClasslist.lst --professors=admin
         chown www-data:www-data -R $APP_ROOT/courses
         echo "Admin course is created."
+        echo "user: admin password: admin added to course admin and tables upgraded"
+    elif [ ! $ADMIN_TABLE_EXISTS ]; then
```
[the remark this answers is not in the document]
mgage: OK.  I'll look further at this.  I don't quite understand why there would be a difference.  There is one other twist. If the passwords don't match then a null (or null string) is returned instead of either a 1 or a 0.  Now that I have gotten ELIF constructions to work I might be able to use that to give extra information about why things are going wrong

docker-config/docker-entrypoint.sh
mgage: This was commented out during the debugging of the docker-entrypoint file -- probably while trying to correct something else that was going wrong.  do you think a conditional test is necessary?

mgage: what about the update-OPL-statistics.pl file?  that was also commented out because it seemed to be causing problems

docker-config/docker-entrypoint.sh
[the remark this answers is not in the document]
taniwallach: I think it is just a matter of Perl vs. `bash`. Try putting the following in a file, `chmod u+x` it and run it.
```
#!/bin/bash
  
ADMIN_TABLE_EXISTS=0

if [ ! $ADMIN_TABLE_EXISTS ]; then
  echo 'negated test passed'
fi

if [ $ADMIN_TABLE_EXISTS == 0 ]; then
  echo 'Was == 0'
fi

ADMIN_TABLE_EXISTS=1

if [ $ADMIN_TABLE_EXISTS ]; then
  echo 'direct test passed after value set to 1'
fi
```
I get the output:
```
Was == 0
direct test passed after value set to 1
```

docker-config/docker-entrypoint.sh
taniwallach: I don't think that a conditional test is needed for the global statistics file. The file is included with the OPL, and loading it is quite fast.

About the local statistics file:
  1. I don't think it should be recreated on each startup by default. A production Docker server could have many courses, so this could create a large delay in startup time.
  2. Hopefully, the main problem you reported with the `update-OPL-statistics.pl` misbehaving if there were non-course directories in the courses directory will be fixed by the recent PR.
  3. I really think the best option is to have users run `update-OPL-statistics.pl` only when they want to, which means that disabling the local statistics by default for Docker is probably the option. I already suggested that in a separate post in the thread: https://github.com/openwebwork/webwork2/pull/1435#issuecomment-875026890

docker-compose.yml
mgage: I see their point, but not having less and an editor available inside the container is a real disadvantage for a brand new user who might want to fix some minor glitch in the local container file or just wants to have a look around.  @xcompass suggested using environment variables to differentiate between production and development (and in this case newbie) builds.  I notice that the installation of vim is not very clean (lots of warnings), and in any case perhaps nano would be a better editor to include. more is included automatically so perhaps we could make do without "less". (pun intended)  I'll think some more about this.

mgage:
_I think that for "developers" and "simple" Docker use, the best solution is probably to disable local statistics by adding
$problemLibrary{showLibraryLocalStats} = 0;
into conf/localOverrides.conf._

I think you are right, the developers and simple Docker use is unlikely to be used heavily for real students

docker-config/docker-entrypoint.sh
```diff
@@ -82,21 +82,58 @@ if [ "$1" = 'apache2' ]; then
                     -e 's/mail{smtpSender} = '\'''\''/mail{smtpSender} = $ENV{"WEBWORK_SMTP_SENDER"}/' \
```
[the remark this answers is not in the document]
mgage: got it

docker-config/docker-entrypoint.sh
mgage: I think you are right. There are unlikely to be "real" students in a simple build for first timers or in a develop build.

pstaabp:
@mgage I've decided to get off my anti-docker horse and dig in.  Can you give me some basics to try this out?  I ran `docker build .` with all the appropriate arguments however I think that built the whole thing.

taniwallach:
With the changes in this PR, you should not need to first edit `docker-compose.yml` as was necessary when https://github.com/openwebwork/webwork2/pull/1404 was first being used.

So roughly you want to do:
```
# cd into webwork2 directory (as usual)
docker build --tag webwork-base:forWW216 -f DockerfileStage1 .
docker-compose build
```

Another `docker-compose build` would only rebuild the "webwork" layers.

---

If you want to force a rebuild of the base OS image (the one which installs lots of Ubuntu packages, etc.) you would run:
```
# cd into webwork2 directory (as usual)
docker build --no-cache --tag webwork-base:forWW216 -f DockerfileStage1 .
```
using the `--no-cache` setting to force Docker not to use cached layers from the prior builds.

taniwallach:
@mgage - I still think the condition to trigger the recreation of the admin course databases is broken. The change I suggested above fixes it for me.

Here is my test protocol:

Update the branch and rebuild the (stage2) image.

```
docker-compose build
docker-compose up -d
```

Enter the container and delete the `admin_user` table to trigger the condition to rebuild the admin course tables.

```
docker exec -it webwork2_app_1 bash

# Inside container
mysql -u webworkWrite -p -h db webwork
# give password (from .env)

mysql> drop admin_user;
mysql> exit

exit
docker-compose down
docker-compose up
```

I see in the middle of the output:
```
app_1  | check admin course and admin tables
app_1  | Waiting for database to become available...
db_1   | 2021-07-07 16:11:52 8 [Warning] Aborted connection 8 to db: 'unconnected' user: 'unauthenticated' host: '172.20.0.4' (This connection closed normally without authentication)
app_1  | using pre-existing admin course and admin tables
```
namely that the proposed version of `docker-compose.yml` is not triggering the admin course table recreation.

```
docker-compose down
```

Change the line in docker-config/docker-entrypoint.sh (diff below):
```
-    elif [ ! $ADMIN_TABLE_EXISTS ]; then
+    elif [ $ADMIN_TABLE_EXISTS == 0 ]; then
```

Rebuild the container (to get the modified version of `docker-entrypoint.sh` into the image). (Externally mounting via `docker-compose.yml` it would also work.)

```
docker-compose build
docker-compose up
```

This time I see:
```
app_1  | check admin course and admin tables
app_1  | Waiting for database to become available...
db_1   | 2021-07-07 16:16:28 8 [Warning] Aborted connection 8 to db: 'unconnected' user: 'unauthenticated' host: '172.21.0.4' (This connection closed normally without authentication)
app_1  | admin course db tables need updating
app_1  | Table user created as admin_user in database.
app_1  | 
app_1  | wwsh - The WeBWorK Shell
app_1  | Available objects: $ce (WeBWorK::CourseEnvironment)
app_1  |                    $db (WeBWorK::DB)
app_1  | Available modules: Data::Dumper
app_1  | 
app_1  | courseID: admin and scriptFile:  /opt/webwork/webwork2/bin/addadmin
app_1  | --------------------------------
app_1  | errors addPassword: password exists (perhaps you meant to use putPassword?) at /opt/webwork/webwork2/bin/addadmin line 4.
app_1  | 
app_1  | admin course tables created with one user: admin   whose password is admin
```

The error about `addPassword` vs `putPassword` is expected, as the `admin_password` table was not removed.

Open `http://localhost:8080/webwork2/admin` and login as `admin` - it works.

mgage:
I fixed the line to 
elif [ $ADMIN_TABLE_EXISTS == 0 ]; then

-- just missed that item when I pushed your other suggestions.
My tests read the same as yours.  

As to an earlier comment to Peter Staab.  docker-compose.yml still needs to be edited to set the root password for mariaDB.  Is there an alternative method that doesn't modify docker-compose.yml? I keep catching myself just before I commit changes in docker-compose.yml and have to go back and fix the password so that it is not revealed.

mgage:
I'm still getting the error
Table 'webwork.OPL_local_statistics' doesn't exist at /opt/webwork/webwork2/lib/WeBWorK/Utils/LibraryStats.pm line 71. 

The line editing the showLibraryLocalStatistics flag failed to include the fact that the original line in localOverrides.conf began with a # since it was commented out.

taniwallach:
> docker-compose.yml still needs to be edited to set the root password for mariaDB. Is there an alternative method that doesn't modify docker-compose.yml? I keep catching myself just before I commit changes in docker-compose.yml and have to go back and fix the password so that it is not revealed.

I cannot think of any reason why it could not be set in the `.env` file and the loaded from the environment in a similar manner to what we already for for the regular DB password.
```
    MYSQL_PASSWORD: ${WEBWORK_DB_PASSWORD}
```
Trying this out requires totally removing the storage volume for the database, as AFAIK only when a new database is first created is the `MYSQL_ROOT_PASSWORD` environment variable used to create the root database user account.

taniwallach:
I think we should probably also add `.env` to `.gitignore` and the only allow forcing changes to be put into that file.

taniwallach:
I just quickly tested a change to `docker-compose.yml` to load the mysql root password from `.env` using the same method as used for other things. After a `docker volume rm webwork2_mysql` and a `docker-compose up` the DB volume is recreated.

The new root SQL password was tested as follows:
```
docker exec -it webwork2_db_1 bash
mysql -u root -p
```

---

I tested the fix to `showLibraryLocalStatistics` from the last commit. It now works as expected. Sorry - I you added code to change it, and did not test that it was OK.

taniwallach (approved):
Other than not liking the startup delay due to `ADD_APT_PACKAGES` all the key changes in this PR now works in my tests.

The startup delay is something I and others can avoid by commenting out the `ADD_APT_PACKAGES` line.

  1. The `admin_user` table is recreated when missing, and a full table schema check is done then.
  2. The `admin` user account get created.
  3. The SQL root password is set in `.env` and gets processed from there on the first creation of the database.
  4. The library browser works properly "out of the box".
I did not test using R, but the setting change makes sense.

PR #1437, opened 2021-07-07
WeBWork 2.16 Release

drgrice1:
This is the official release of WeBWorK 2.16.


Alex-Jordan:
Suppose this is merged into `main`. Does `main` then immediately get merged into `develop`?

drgrice1:
Currently the only difference between the WeBWorK-2.16 branch (this pull request) and the develop branch are the version changes.  Run `git diff origin/WeBWorK-2.16 origin/develop` to see that this is the case.  The same is true for the origin/PG-2.16 and origin/develop branches.  So there should be no need for the merge into develop.  Development can continue on the develop branches as usual.

Alex-Jordan:
OK, I see that. When I look at GitHub Insights, in the Network graph, there appear to be more differences. The commit hashes are not aligned.

I was looking because in #1130, @pschan-gh 's develop-answer-log branch has merged the latest 2.16. It is targeted into develop, and GitHub is showing a ton of differences, including things that are unrelated to the answer log and were part of 2.16 development. Should that be rebased off of develop now?

drgrice1:
For now I targeted #1130 to the WeBWorK-2.16 branch.  That cleaned up the other things.  The best way to fix it to work with the develop branch is to merge the develop branch into that branch.  Rebasing would be a pain as there would be a lot of conflicts to resolve.  Note that git cleans things up and the commits are not doubled.  Although they have different hashes, they are the same.  This messes up the network graph, but works out.

To fix the network graph for the develop branch we probably will want to merge the WeBWorK-2.16 into develop as well, and then remove the version changes.

drgrice1:
Maybe the cleanest way to do it would be to reset develop to main after WeBWorK-2.16 is merged, and then revert the version change information.  This would involve a force push though.  That will properly align everything and fix the mess with develop.  It seems that a merge doesn't do things as cleanly as I hoped.

taniwallach:
> Maybe the cleanest way to do it would be to reset develop to main after WeBWorK-2.16 is merged, and then revert the version change information. This would involve a force push though. That will properly align everything and fix the mess with develop. It seems that a merge doesn't do things as cleanly as I hoped.

I think that any solution which gets develop cleanly realigned with main after the 2.16 merge would be a good thing. The differences have become quite confusing to understand at time. Recreating develop as a copy of the new main seems a very good option.

drgrice1:
Yeah, I think that will be the plan.

PR #1438, opened 2021-07-08
Docker - use main instead of master branch (only to be merged when 2.16 is about to be merged into main)

taniwallach:
Modify the build args section:
  - To use main branches by default.
  - Layout change of the settings.

Add a minor comment regarding ADD_APT_PACKAGES and startup speed.

**Should be merged only shortly before 2.16 is merged into main.**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -77,15 +77,22 @@ services:
         #  context: /Path_To/webwork2/
         #  dockerfile: /Path_To/Dockerfile
         args:
-            # build args specifying the branches for webwork2 and pg used to build the image
-            # To use the master branches of webwork2 and pg 
-            #- ARG WEBWORK2_BRANCH=master
-            #- ARG PG_BRANCH=master
-            # To use the 2.15 branches of webwork2 and pg from the "official" GitHub repositories
+            # Build args to set the source Git repositories. Here we set the main repositories:
             - WEBWORK2_GIT_URL=https://github.com/openwebwork/webwork2.git
-            - WEBWORK2_BRANCH=WeBWorK-2.16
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This generally looks good.  However, should the defaults in the WeBWorK 2.16 branch be for the WeBWorK-2.16 and PG-2.16 branches?  And then switch that to main right before the official merge?

taniwallach:
> However, should the defaults in the WeBWorK 2.16 branch be for the WeBWorK-2.16 and PG-2.16 branches? And then switch that to main right before the official merge?

Yes. My intention was for this to only be merged in just before 2.16 gets merged into main, as which point the 2.16 branch is only of historical interest. That was what I intended by the line at the end of the first post.

docker-compose.yml
```diff
@@ -77,15 +77,22 @@ services:
...
+            # Exactly one pair of settings should be left uncommented and all the others
+            # should be commented out.
+
+            # main branches:
+            - WEBWORK2_BRANCH=main
+            - PG_BRANCH=main
+
+            # WeBWorK/PG 2.16 branches:
+            #- WEBWORK2_BRANCH=WeBWorK-2.16
+            #- PG_BRANCH=PG-2.16
```
drgrice1, line 95: All of the 2.16 references should not be here.  Instead these should be references to the develop branch.  After a while the 2.16 branches will be deleted.

taniwallach: Good point. It was meant as a sample alternate option, but using develop is a better idea.
Done.

drgrice1 (approved):
Looks good.  This will be merged immediately before the WeBWorK-2.16 branch is merged into main.

PR #1445, opened 2021-07-14
Lms restricted navigation (Take 2)

limefrogyank:
This is a redo of a pull request made previously:  https://github.com/openwebwork/webwork2/pull/1224

See that PR for the reasons for this one.  However, there were certain flaws that I hope I have fixed in this version.

## Scenario A
User uses webwork through LMS only.  Login is automatic, so student can't login to webwork directly.
   - Breadcrumbs, most of the sidebar links, the masthead, etc. all disappear.  This is now done partially by editing code in the perl files, but also partially by creating a new template file.  The templateName function is overridden in a few of the relevant files.
   - If viewed through an IFRAME, a go back button shows up on all pages except the ProblemSets page.
   - If viewed directly in the browser, there is no "go back" button.  Users will have to use the browser controls.
 
image

Weakness to this method:  User can type the problem set name manually in the URL if they're not using an IFRAME.

## Scenario B
User has a login/pass for webwork.  
-  All of the above apply
- When student clicks on class, no problem sets will be displayed.  Instead, a warning in a red box shows up to tell them to login via their LMS.
image

Weakness:  Same as above.  Manual URL changes can bypass all of this.

## Improvements?
Aside from forcing students to use iframes, the only other fix to editing the url for navigation I can think of would be to include a unique key for each problem set as a url parameter and check for it.  This key would have to be stored in the database and instructors would have to know to paste a url including this key as links in their LMS.  I didn't want to go that route because it seems too complicated.

I think I've kept the UI styling from becoming too broken.  I had to use templates in order to do that.  I've also added the "Course Configuration" link (if allowed by role) to the menu so that you can test the restrictions and change back using the UI.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -820,6 +820,13 @@ $authen{xmlrpc_module}  = "WeBWorK::Authen::XMLRPC";
 	download_hardcopy_format_pdf => "guest",
 	download_hardcopy_format_tex => "ta",
 	download_hardcopy_change_theme=> "ta",
+
+	# UI modifications for LMS
+	# A permission that is normally set to guest, but can be set to 'login_proctor' so that students don't see navigation links.
+	# This would be very useful when having students navigate to problem sets from an LMS and preventing the student from navigating
+	# to a different problem set from within WeBWorK.  Doing so prevents the LMS from receiving scores for the navigated problem sets. 
+	# Those problem sets must navigated to directly from the LMS for the grade to be sent back to the LMS gradebook. 
+	navigation_allowed => "guest",  # default everyone is allowed
 );
 
 # This is the default permission level given to new students and students with
@@ -1721,6 +1728,11 @@ $ConfigValues = [
 		{ var => 'permissionLevels{can_show_old_answers}',
 		  doc => 'Can show old answers',
 		  doc2 => 'When viewing a problem, WeBWorK usually puts the previously submitted answer in the answer blank.  Below this level, old answers are never shown.  Typically, that is the desired behaviour for guest accounts.',
+		  type => 'permission'
+		},
+		{ var => 'permissionLevels{navigation_allowed}',
+		  doc => 'Allowed to navigate to all areas of WeBWorK. Otherwise, restricted to direct-linked problem sets.',
+		  doc2 => 'If using an LMS with grading for each assignment, the LMS will require navigating to each specific homework set directly from the LMS.  If students finish a homework set and then proceed to another one within WeBWorK without using the LMS links, the second homework assignment\'s grades will not be passed back to the LMS until the student uses the direct link.  Setting this to anything higher than student will prevent the student from navigating to another assignment by removing all links.',
 		  type => 'permission'
 		},
 	],
--- a/htdocs/themes/math4/lms.template
+++ b/htdocs/themes/math4/lms.template
@@ -1,0 +1,336 @@
+<!DOCTYPE html>
+<html <!--#output_course_lang_and_dir-->>
+<!--  Note that the lang and dir attributes are now set during the
+      processing of the template and not hard-coded above.
+-->
+<head>
+<meta charset='utf-8'>
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<!-- This encourages IE to *not* use compatibility view, which breaks math4 -->
+<meta http-equiv="X-UA-Compatible" content="IE=Edge"/>
+
+<!--
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2021 The WeBWorK Project, http://openwebwork.sf.net/
+# 
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+# 
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+-->
+
+<link rel="icon" type="x-image/icon" href="<!--#url type="webwork" name="htdocs"-->/images/favicon.ico"/>
+
+<!-- CSS Loads -->
+<link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="htdocs"-->/js/vendor/bootstrap/css/bootstrap.css"/>
+<link href="<!--#url type="webwork" name="htdocs"-->/js/vendor/bootstrap/css/bootstrap-responsive.css" rel="stylesheet" />
+<link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="htdocs"-->/node_modules/@fortawesome/fontawesome-free/css/all.min.css"/>
+<link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="theme"-->/math4.css"/>
+<!--#if can="output_jquery_ui"-->
+	<link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="htdocs"-->/node_modules/jquery-ui-dist/jquery-ui.min.css"/>
+<!--#endif-->
+<!--#if can="output_achievement_CSS"-->
+<link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="theme"-->/achievements.css"/>
+<!--#endif-->
+<!--#if can="output_CSS"-->
+	<!--#output_CSS-->
+<!--#endif-->
+<!--#if exists="math4-coloring.css"-->
+<link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="theme"-->/math4-coloring.css"/>
+<!--#endif-->
+<!--#if exists="math4-overrides.css"-->
+<link rel="stylesheet" type="text/css" href="<!--#url type="webwork" name="theme"-->/math4-overrides.css"/>
+<!--#endif-->
+
+<!-- JS Loads -->
+<script src="https://polyfill.io/v3/polyfill.min.js?features=es6" defer></script>
+<script type="text/javascript" src="<!--#url type="webwork" name="htdocs"-->/js/apps/MathJaxConfig/mathjax-config.js" defer></script>
+<script type="text/javascript" src="<!--#url type="webwork" name="MathJax"-->" id="MathJax-script" defer></script>
+<script type="text/javascript" src="<!--#url type="webwork" name="htdocs"-->/node_modules/jquery/dist/jquery.min.js"></script>
+<!--#if can="output_jquery_ui"-->
+	<script type="text/javascript" src="<!--#url type="webwork" name="htdocs"-->/node_modules/jquery-ui-dist/jquery-ui.min.js"></script>
+<!--#endif-->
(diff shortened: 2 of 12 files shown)


limefrogyank:
Also, a few of the code review comments said to use `return "" unless ...` instead of just `return ""`.  I couldn't get that to work.  Maybe it's just a syntax problem, but 
```
return "" unless ...;
```
threw errors.

conf/defaults.config
```diff
@@ -820,6 +820,13 @@ $authen{xmlrpc_module}  = "WeBWorK::Authen::XMLRPC";
 	download_hardcopy_format_pdf => "guest",
 	download_hardcopy_format_tex => "ta",
 	download_hardcopy_change_theme=> "ta",
+
+	# UI modifications for LMS
+	# A permission that is normally set to guest, but can be set to 'login_proctor' so that students don't see navigation links.
+	# This would be very useful when having students navigate to problem sets from an LMS and preventing the student from navigating
+	# to a different problem set from within WeBWorK.  Doing so prevents the LMS from receiving scores for the navigated problem sets. 
+	# Those problem sets must navigated to directly from the LMS for the grade to be sent back to the LMS gradebook. 
+	navigation_allowed => "guest",  # default everyone is allowed
```
drgrice1, line 829: The comments here are too long.  Please reduce the length to at most 120 characters.

htdocs/themes/math4/lms.template
```diff
@@ -0,0 +1,336 @@
...
+<body>
+<a href="#page-title" id="stmc-link" class="sr-only sr-only-focusable">Skip to main content</a>
+
+<!-- Bootstrap Fluid Layout Scaffolding start -->
+<div id="bs-container" class="container-fluid">
+<!-- Header -->
+<div id = "masthead" class="row-fluid" role="banner">
+	<div id="loginstatus" class="span4">
+		<!--#loginstatus-->
+	</div>
```
drgrice1, line 83: Why remove the WeBWorK and MAA logos?  The links that those include are harmless, and there really is no reason to do so.  Furthermore, this makes the resulting masthead look quite bad with the span4 being the only thing in the row.  At the very least, an offset class should be added to put the span4 with the loginstatus back to the right.  I don't like that approach though, as the logos really should still be here.

htdocs/themes/math4/math4.js
```diff
@@ -265,4 +265,18 @@ if ($.fn.button.noConflict) $.fn.bootstrapBtn = $.fn.button.noConflict();
...
+		// The page is in an iframe, show the GoBack button if it exists.   By default, it's hidden so set display to block.
+		$('#restrictedNavigationGoBackButton').css({'display':'block'});
+	} else {
+		// The page is not in an iframe, do nothing because GoBack button is automatically hidden.  User will use browser controls for navigation.		
+	}
+
+	//Restricted Navigation in LMS
+	$('#restrictedNavigationNavControls').css({'height':'38px', 'margin-bottom':'10px'});
+	$('#restrictedNavigationGoBackButton').addClass('btn btn-primary').click(function() {
+		window.history.back();
```
drgrice1, line 279: The styles should not be added by javascript.  The css added here should be in math4.css.  The `btn` and `btn-primary` classes should just be added to the `class` when the button is created in Problem.pm, Options.pm, and Hardcopy.pm.  I also would suggest adding the style
```css
#restrictedNavigationGoBackButton {
    height: 38px;
}
```
to math4.css to improve the appearance of the button relative to the menu toggle button.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -785,6 +785,22 @@ sub links {
 	print CGI::start_li({class => "nav-header"});
 	print CGI::h2($r->maketext("Main Menu"));
 	print CGI::end_li();
+
+	unless ($authz->hasPermissions($userID, "navigation_allowed")){
```
drgrice1, line 789: This has problems.  There are several cases where there is no `$userID`.  For example on the course list page and the login page.  On the course list page this causes the `Courses` link to be omitted from the sidebar even if the site is not using the new option.  Furthermore, if a user signs in with an invalid username or password (and so there is an username, but it is invalid), then instead of getting the message that authentication failed, the user gets an error page.  This should be changed to:
```suggestion
	if (defined $userID && $r->db->existsUser($userID) && !$authz->hasPermissions($userID, "navigation_allowed"))
```

lib/WeBWorK/ContentGenerator/Feedback.pm
```diff
@@ -70,6 +70,21 @@ use constant MP2 => ( exists $ENV{MOD_PERL_API_VERSION} and $ENV{MOD_PERL_API_VE
...
+sub templateName {
+	my $self = shift;
+	my $r = $self->r;
+	my $authz = $r->authz;
+	my $userID = $r->param('user');
+	# override templateName with lms template
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+		$self->{templateName} = "lms";
+		return "lms";
+	}
```
drgrice1, line 82: There is no need for the `$authz` and `$userID` variables.  Those values are only used once, so just use them.  So change this to
```suggestion
	# override templateName with lms template
	unless ($r->authz->hasPermissions($r->param('user'), "navigation_allowed")) {
		$self->{templateName} = "lms";
		return "lms";
	}
```

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -112,6 +112,49 @@ our @HC_FORMAT_DISPLAY_ORDER = ('tex', 'pdf');
...
+	# If regular navigation is restricted, we'll need a "go back" button in case students are using an iframe window within their LMS.
+	# Using browser navigation will only navigate away from the entire webwork iframe in that case.
+	if (defined $courseID) {
+		unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+			print CGI::button(
+				-id=>'restrictedNavigationGoBackButton',
+				-class=>'nav_button',
+				-value=>'Go Back',
+				-style=>'display:none;');
+		}
```
drgrice1, line 137: Why do you think it is a good idea to only have the back button if the page is in an iframe?  That doesn't make sense, and it only makes your implementation convoluted.  Remove the `display:none` style, and delete the javascript in math4.js that adds the `display:block` style.

Also remember to add the `btn btn-primary` classes to the class here.

This code could also use clean up.  There are several variables defined here that don't need to be.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -112,6 +112,49 @@ our @HC_FORMAT_DISPLAY_ORDER = ('tex', 'pdf');
...
+sub templateName {
+	my $self = shift;
+	my $r = $self->r;
+	my $authz = $r->authz;
+	my $userID = $r->param('user');
+	# override templateName with lms template
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+		$self->{templateName} = "lms";
+		return "lms";
+	}
```
drgrice1, line 152: Same clean up needed as mentioned in Feedback.pm.

lib/WeBWorK/ContentGenerator/Options.pm
```diff
@@ -30,7 +30,48 @@ use WeBWorK::CGI;
...
+	my $userID = $r->param('user');
+	# override templateName with lms template
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+		$self->{templateName} = "lms";
+		return "lms";
+	}
+	my $templateName = $r->param('templateName')//'system';
+	$self->{templateName}= $templateName;
+	$templateName;
+}
```
drgrice1, line 74: All of the comments for the file Hardcopy.pm apply here as well.

lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -1227,6 +1243,15 @@ sub path {
...
 	}
 
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+		print CGI::button(
+			-id=>'restrictedNavigationGoBackButton',
+			-class=>'nav_button',
+			-value=>'Go Back',
+			-style=>'display:none;');
+		return "";
+	}
```
drgrice1, line 1253: Same changes needed here as in previous files.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -166,6 +174,15 @@ sub body {
 	
 	my $hardcopyPage = $urlpath->newFromModule("WeBWorK::ContentGenerator::Hardcopy",  $r, courseID => $courseName);
 	my $actionURL = $self->systemLink($hardcopyPage, authen => 0); # no authen info for form action
+
+	# prevent showing anything in the body and instead display message that you must view assignments from the LMS links
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+		print CGI::start_div({id=>'warningRestrictedProblemSetsView'});
+		print CGI::h3({},"You must view your problems through your LMS (i.e. Blackboard, D2L, Moodle, etc).");
```
drgrice1, line 181: The hash argument is optional, and it should be omitted if it is empty.  So change this to
```suggestion
		print CGI::h3("You must view your problems through your LMS (i.e. Blackboard, D2L, Moodle, etc).");
```

Also, the $userID is used only once, and so the definition is not needed above.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -150,6 +149,13 @@ sub title {
 sub templateName {
 	my $self = shift;
 	my $r = $self->r;
+	my $authz = $r->authz;
+	my $userID = $r->param('user');
+	# override templateName with lms template
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+		$self->{templateName} = "lms";
+		return "lms";
+	}
```
drgrice1, line 158: Clean up as before.

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -126,6 +126,13 @@ sub help {   # non-standard help, since the file path includes the course name
 sub templateName {
 	my $self = shift;
 	my $r = $self->r;
+	my $authz = $r->authz;
+	my $userID = $r->param('user');
+	# override templateName with lms template
+	unless ($authz->hasPermissions($userID, "navigation_allowed")) {
+		$self->{templateName} = "lms";
+		return "lms";
+	}
```
drgrice1, line 135: Clean up as before.

htdocs/themes/math4/math4.js
drgrice1, line 279: A cleaner and no jQuery implemenation would be better.  For example:
```suggestion
	document.getElementById('restrictedNavigationGoBackButton')?.addEventListener('click', () => window.history.back());
```

drgrice1 (changes requested):
This needs some more work.  See my comments.

Also, this does not address the issue that @Alex-Jordan commented on in the previous pull request about the achievements page.  If achievements are enabled for the course, then students should be able to go to that page.

Also, please do not create another pull request again.  Make additional commits to the branch this pull request is based on and push them.  Those will be added to this pull request.  That keeps the conversation here, and doesn't make reviewers need to look at another closed pull request for reference.  That is the proper way to work with GitHub pull requests.

drgrice1:
One other thing I forgot to comment on in my review... Why was the empty htdocs/thems/math4/math4-coloring.css file added?  That file should be removed from this pull request.

taniwallach:
@drgrice1 Can this be closed. It seems that https://github.com/openwebwork/webwork2/pull/1588 replaces what this PR was intended to do.

drgrice1:
Yeah, this can be closed.

PR #1446, opened 2021-07-15
force modelCourse to be listed as template source

drdrew42:
recent changes to CourseAdmin.pm `listCourses` has stopped listing `modelCourse` under the available sources for template files during course creation.

This PR forcibly adds `modelCourse` to the list of available template sources. It should be merged into main as a hotfix for 2.16

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -485,6 +485,7 @@ sub add_course_form {
 	#});
 	
 	my @existingCourses = listCourses($ce);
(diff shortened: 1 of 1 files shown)


drgrice1:
Good catch.  I will go ahead and merge this and get this into main as well.

pstaabp:
I was actually going to suggest that modelCourse is placed first in the list.  With a lot of courses, it may be difficult to scroll to the bottom.

taniwallach:
For future consideration, we could create a config variable which could make it possible to control what gets added in this manner.
In Slack I also suggested prepending... sort of what @pstaabp just wrote.

PR #1448, opened 2021-07-15
Configurable modelcourses list for admin add course

taniwallach:
This is intended to supplement what was just done as a hot fix in https://github.com/openwebwork/webwork2/pull/1447 but to make it possible to configure a list of model courses which get prepended to the provided list.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -682,6 +682,10 @@ $courseFiles{logs}{activity_log} = '';
 # to the new course being created.
 $siteDefaults{default_templates_course} ="modelCourse";
 
+# Provide a list of model courses which are not real courses, but from which
+# the templates for a new course can be copied.
+$modelCoursesForCopy = [ "modelCourse" ];
+
 ################################################################################
 # Authentication system
 ################################################################################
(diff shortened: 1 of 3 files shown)


drgrice1:
I am not sure that this qualifies for a hotfix.  This adds a new feature that was not present before.  Generally that is something that should not be added after the release.

taniwallach:
Yes and no. In the past - just about any directory in the courses directory could be used as a the source of the templates directory, as `listCourses` included them (so long as they avoided some regular expression conditions). This provides an option to optionally define selected directories (other than the fixed `modelCourse`) to be included in the list. So it reenables more of the prior functionality, but in a controlled manner.

It can certainly be put in develop instead (or dropped). Whatever the team prefers.

pstaabp (approved):
I think this looks good.  Technically not a hot fix, but functionality nearly equivalent and more flexible.

drgrice1:
I see.  I suppose that could be hotfix worthy if that is something that was commonly utilized.  What do others thing on this?

drdrew42 (approved):
This does give an alternative for the options that were available before listCourses was modified.

drgrice1:
Okay.  A hotfix it can be then.

drgrice1 (approved):
I will merge and also add it to develop.

PR #1456, opened 2021-07-21
Bugfix: JITAR problem sets 'save' not respecting nested problems

drdrew42:
* nestedSortable changed their object structure, and this broke our Hmwk Sets Editor script
* Also, fixing a persistent typo

Thanks to Monica VanDieren for pointing out this issue

### To test:
**BEFORE PATCH:**
1. create a new set of blank problems, or use an existing set with at least two problems
2. convert the set to just-in-time and save
3. drag a problem to the right to make it a child of the previous problem
4. attempt to save and notice that the parent has gone missing, and the child has taken its place.
5. also note that 'top-level' problems display "Counts For Parent" configuration setting

**AFTER PATCH:**

  1-3. Same
  4. attempt to save and see that you've kept the parent, and it has a child (you might need to click to expand).
  5. note that top-level problems do not display "Counts For Parent" (as it should be)

### DIFF
--- a/htdocs/js/apps/ProblemSetDetail/problemsetdetail.js
+++ b/htdocs/js/apps/ProblemSetDetail/problemsetdetail.js
@@ -101,9 +101,9 @@
 	// This uses recursion to set the #prob_num_id fields to the
 	// new position in the tree whenever the tree is updated or
 	// the renumber button is clicked
-	var recurse_on_heirarchy = function (heirarchy,array) {
-		for (var i=0; i < heirarchy.length; i++) {
-			var id = heirarchy[i].id;
+	var recurse_on_hierarchy = function (hierarchy,array) {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Good catch.  I approve this for a hotfix.

drgrice1:
I am pretty sure the heirarchy misspelling didn't have a role in the problem though.  I don't know why I didn't see that when I worked on that file last.

drdrew42:
This was a tricky one to find -- exactly because no one's been working on problemsetdetail this version...

drgrice1:
There were the changes to the rendering of problems.  That involved quite a bit of change to this javascript file.

Of course it was the jQuery update that triggered the need for an update to the nestedSortable javascript that caused this issue.

pstaabp (approved):
"i before e" except after h? 

Looks good.  Nice catch.

PR #1461, opened 2021-07-27
New utility script: bin/crypt_passwords_in_classlist.pl

taniwallach:
Add `bin/crypt_passwords_in_classlist.pl` to provide a reasonable manner of preparing a classlist file with properly crypted passwords.

The script was coded with a copy of the 2 needed subroutines from `lib/WeBWorK/Utils.pm` so it can easily be used on a machine with Perl but without WeBWorK.

This is meant to help address the fact that https://webwork.maa.org/wiki/Classlist_Files#Password_details had long outdated instructions on how to prepare `crypt`-ed passwords for classlist files.

### DIFF
--- a/bin/crypt_passwords_in_classlist.pl
+++ b/bin/crypt_passwords_in_classlist.pl
@@ -1,0 +1,54 @@
+#!/usr/bin/perl
+
+use open IO => ':encoding(UTF-8)';
+
+# ==================================================================
+
+# 2 subroutines copied from lib/WeBWorK/Utils.pm from WW 2.16 version
+
+sub cryptPassword($) {
+	my ($clearPassword) = @_;
+	#Use an SHA512 salt with 16 digits
+	my $salt = '$6$';
+	for (my $i=0; $i<16; $i++) {
+		$salt .= ('.','/','0'..'9','A'..'Z','a'..'z')[rand 64];
+	}
+
+	my $cryptPassword = crypt(trim_spaces($clearPassword), $salt);
+	return $cryptPassword;
+}
+
+## Utility function to trim whitespace off the start and end of its input
+sub trim_spaces {
+	my $in = shift;
+	return '' unless $in;  # skip blank spaces
+	$in =~ s/^\s*|\s*$//g;
+	return($in);
+}
+
+# ==================================================================
+my $inputfile = shift;
+my $outfile = "crypted_" . $inputfile;
+
+if ( -e $inputfile && -r $inputfile ) {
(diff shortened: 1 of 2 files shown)


taniwallach:
The changes which "require" the changes were in:
  - https://github.com/openwebwork/webwork2/pull/473
  - https://github.com/openwebwork/webwork2/commit/f7aa5d7c95e3f72ead91820a006cea3a1b752610
  - https://github.com/openwebwork/webwork2/commit/5226b4d9c4644de47cc549429eb627517857707b
and those seem to have been in WeBWorK 2.10 (or so).

The discussion seems indicate that the old approach of generating a crypted password to put in the classlist file should still work, but it did not work for me. In any case, it seems unwise to continue to recommend an approach which will create the old and less secure crypted form to be used in modern sites/courses.

bin/crypt_passwords_in_classlist.pl
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/perl
+
+use open IO => ':encoding(UTF-8)';
+
+# ==================================================================
+
+# 2 subroutines copied from lib/WeBWorK/Utils.pm from WW 2.16 version
+
+sub cryptPassword($) {
```
pstaabp, line 9: Why copy these over?  Instead just 
```
use WeBWorK::Utils qw/cryptPassword trim_spaces/;
```

taniwallach: I want this script to be something which people who need to prepare classlist files can use off the WeBWorK server and without needing some sort of local installation of the webwork2 codebase.

That should let non-WeBWorK staff from an institution (ex. one being hosted by someone else) provide a classlist file in which initial encrypted passwords are provided to the teachers or IT staff handling the WeBWorK course/server in a secure manner.

The current use case is an institution which cannot use LTI, is requesting hosting for a trial course, and which intends to use LDAP authentication for that specific course (but to allow system support staff to use the local authentication as a fallback). Loading a classlist file without prepared encrypted passwords would default to using `student_id` as the password, which is not particularly secure.

pstaabp: This is okay. I know you have a comment to make sure that the two subroutines are identical, but that seems like it requires us (developers) to do the right thing.

pstaabp (approved):
Works as advertised and definitely a better solution that current.

drgrice1:
This is going in.  It adds a script and doesn't affect anything else.  @pstaabp approved it.

PR #1471, opened 2021-08-09
Update README.md for WW 2.16

taniwallach:
Update `README.md` to refer to WW 2.16 as the current version, add link to openwebwork.org and to the installation manuals.
Some other small changes.

More changes can be added via PR to my branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,9 @@
                                    Version 2.*
                         Branch:  github.com/openwebwork 
 
-	          http://webwork.maa.org/wiki/Release_notes_for_WeBWorK_2.14
-                    Copyright 2000-2019, The WeBWorK Project
+             https://webwork.maa.org/wiki/Release_notes_for_WeBWorK_2.16
+                    Copyright 2000-2021, The WeBWorK Project
+                             https://openwebwork.org/
                              http://webwork.maa.org
                               All rights reserved.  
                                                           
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Good catch.

Although, perhaps we should not specifically refer to the version in the README.md file.  We could instead just link to https://webwork.maa.org/wiki/Category:Release_Notes for the release notes.  The versions would not need to be specified in the information for downloading.  The download instructions for the latest version will be at the top of the Instation_Manuals page on the wiki.  Perhaps that page could be made more clear as to what the latest release is.  I am not sure there is a need to refer to the copyright in the readme either.

Just a thought on possible simplification to the release process.

taniwallach:
@drgrice1 Sounds good to me, I suggest that a new PR be created with suitable changes, and then I will close this. Some of what I changed is probably worth keeping.

drgrice1:
For now we can just go with this.  For future releases we can consider the alternative approach that I discussed.

drgrice1:
This pull request also adds the new link to openwebwork.org which is needed.

pstaabp (approved):
I think this is sufficient for now.  If the one in develop is not updated to this then we should do this too.

taniwallach:
> I think this is sufficient for now. If the one in develop is not updated to this then we should do this too.

I'm not sure what method we decided is the best one to keep develop in sync with hotfixes to main. 

If I try to make a PR from the source branch of this PR to develop  https://github.com/openwebwork/webwork2/compare/develop...taniwallach:fix-readme-2.16?expand=1 it collects several older commits (and merge commits) made to main already, but most real commits seem to have been added to develop in a different manner.

@drgrice1 did we come up with a good approach?

drgrice1:
We ended up just going with the hotfix approach used before.  A separate pull request to develop is made with the same branch.  There doesn't really seem to be a better way to keep two branches in GitHub in sync.

I can take care of that when this is merged.

drgrice1:
I think this needs to be merged.

PR #1474, opened 2021-08-11
Optionally remove student id from feedback

taniwallach:
This PR adds a new control switch `blockStudentIDinFeedback` to allow keeping the `student_id` out of feedback messages.

It is intended to be enabled where the `student_id` is "personal information" under local privacy laws/regulations, so should be protected.

It is not really necessary to include the `student_id` in the email message, as an instructor can, when necessary, find the data in the class-list editor.

The default is not to remove it (set in `defaults.config`), and an explanatory comment and sample line to make use of the new feature were added to `localOverrides.conf.dist` but the setting could also be made via a `course.conf` file.

I also noticed a typo "feeback" in several places, some of which get into the translation tables. The typo was fixed everywhere in the second commit of the PR.

Reviewers may wish to inspect the changes from the two  commits separately.


pstaabp:
Overall, the code looks like it will work.  I can't check it right now in that I still haven't set up a SMTP server on my development machine.  I will try to do this.  

@taniwallach I'm guessing you've thought about this, but are other places we need to scrub this kind of data?

taniwallach:
I have not really looked hard for where else there may be problems. The biggest problems I am aware of are:
  1. The ID in emails - which this PR provides a mechanism to disable. The problem is that the emails get sent in an insecure manner, so should only have necessary information (login name, but preferably not an ID number, when that is "personal" information.)
  2. The fact that PDF files are available to whoever guesses the URL (and not only to the user they belong to) and on servers which do not protect the indices can be found by web crawlers. That can contain ID, email, and name so should not be found in the public." Eventually we should fix that. It will probably require changing where the PDF is created (not in `htdocs/tmp`), and serving the file via the Perl code and only to an authorized and authenticated  user who should have access.
  3. The grades files contain lots of data - and we probably should put up a reminder that the people who download it should handle the data with suitable care. Same applies to course archives.

From some reading, many countries have data privacy regulations which would apparently require implementing mechanisms of tracking who has access to the data, when access is used, etc. Meeting such standards is probably far beyond what it reasonable for us to implement by patching webwork2. I think for now, the assumption should just be that all "staff" accounts have access to all the personal data in the course, but that other than downloads (PDF, grades, course archive) we should try not to allow any "personal information" to be accessed except via the web interface by people with sufficient permissions. I think that is more or less the best we can do for now.

PR #1475, opened 2021-08-12
html2xml add hideAuthData switch (for man-in-the-middle use)

taniwallach:
Adds a `hideAuthData` control setting which when enabled will replace the values of the hidden input fields of authentication data in `html2xml` generated problems with "hidden".

Such a change is only feasible when the problems are being requested by a man-in-the-middle system which can add the authentication data when processing a request/submission, so that it need not be provided to the end user's browser (and thus essentially made public).

The goal at present is to protect the daemon `course_password`, etc. when problems are rendered by `html2xml` for the WeBWorK XBlock:       https://github.com/Technion-WeBWorK/xblock-webwork

Note: Submitting a problem loaded into an HTML page using `html2xml` when this setting is enables will **not** allow submissions of answers. That is intentional, and this new setting should **not** be used when plain HTML embedding is required.

### DIFF
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -131,6 +131,15 @@ sub formatRenderedProblem {
 	my $displayMode = $self->{inputs_ref}{displayMode};
 	my $hideWasNotRecordedMessage = $ce->{hideWasNotRecordedMessage} // 0;
 
+	# Provide an option to hide the authentication data when ht2ml2xml is being used
+	# by a man-in-the-middle system.
+	if ( $ce->{hideAuthData} ) {
+		$course_password = "hidden";
+		$session_key = "hidden";
+		$userID = "hidden";
+		$courseID = "hidden";
+	}
+
 	# HTML document language settings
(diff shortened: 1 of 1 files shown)


drgrice1:
@taniwallach:  Do you still want to get this into the next release?  I don't see any real problem with it, except that it provides an undocumented course environment option.

taniwallach:
The use case is for the webwork XBlock for edX (https://github.com/Technion-WeBWorK/xblock-webwork) but for the planned production use we intend to use "standalone renderers" rather than `html2xml` from regular webwork servers. For testing of the XBlock using `html2xml` I added the changes here locally.

Others who want to try out the XBlock may find it convenient to test against a regular existing webwork server, and can add this patch locally if they need it, as it will protect the authentication data from being in the problem content available in the embedded problems.

As such, I would be happy to see it merged, but it is certainly not something critical or urgent.

taniwallach:
I don't think this is of much interest to anyone except me, as the Standalone renderer is a better option as a rendering backend for a man-in-the-middle system. I'm closing this. It can be reconsidered if someone else has a similar use case where the feature would be helpful.

PR #1478, opened 2021-08-15
Error with utf8 characters when exporting classlist.

jutrembBDEB:
There was an error in the file .lst when exporting the users with utf8 characters. I corrected the bug. Can you test that it's working on your end.

Thank you!

### DIFF
--- a/lib/WeBWorK/File/Classlist.pm
+++ b/lib/WeBWorK/File/Classlist.pm
@@ -91,7 +91,7 @@ sub parse_classlist($) {
 sub write_classlist($@) {
 	my ($file, @records) = @_;
 	
-	my $fh = new IO::File($file, ">")
+	my $fh = new IO::File($file, '>:encoding(UTF-8)')
 		or die "Failed to open classist '$file' for writing: $!\n";
 	
 	my $csv = Text::CSV->new({ binary => 1});


taniwallach:
Until this change, UTF-8 characters in the data fields would trigger a warning:
```
Wide character in print at /opt/webwork/webwork2/lib/WeBWorK/File/Classlist.pm line 117.
```
but in my experience the exported class-list files were generated correctly as UTF-8 files and could be imported.

I suspect that was a result of the CSV converter being created using `binary` mode:
```
my $csv = Text::CSV->new({ binary => 1});
```

After this patch, no warnings are issued.

At least for my test case, the class-list file generated before and after the change are identical.

taniwallach (approved):
Fixes the "wide character warning" being issued on exporting a classlist file when there are "wide" UTF-8 characters in the data.

Since the actual exported data seems fine even before the fix, I am not certain is this should be accepted as a hotfix to main, or just fixed in develop. @drgrice1 - what do you think?

drgrice1:
It would be nice to get rid of warnings.  So I think it could be a hotfix.

jutrembBDEB:
It’s more than a warning to me. When I download the classlist file and open it in a text editor, all the strings with UTF-8 characters are all wrong. And if I upload the same file in an another course, the name of the users are all wrong also. It would be a big problem for the instructors in my institution.

drgrice1 (approved):

taniwallach:
> It’s more than a warning to me. When I download the classlist file and open it in a text editor, all the strings with UTF-8 characters are all wrong. And if I upload the same file in an another course, the name of the users are all wrong also. It would be a big problem for the instructors in my institution.

Ouch. Then this should certainly be a hotfix.

I'm not really sure why it works on my systems (WW 2.16 inside Docker) and exported class list files with Hebrew work properly. Since the files work for me, I did not take the warning as something to fix urgently. I'm guess that the difference is probably "low level" (Perl version or OS settings effecting Perl's default behavior). It could be that my having a setting for `ADD_LOCALES` in the `docker-compose.yml` has some effect on this, but I don't think so.

drgrice1:
I am going to merge this.  We have two approvals, and it is a minor change.

Issue #1477, opened 2021-08-15
WW 2.16 Missing the msgid "Import" in the translation files for the label on the set editor / user editor import tab


taniwallach:
The `msgid "Import"` was moved in the PO translation files into a comment at the bottom of the file. It used to provide the translation for the "label" on the Import tab of the set editor.

This occurred in WW 2.16, but the string was available for translation in WW 2.15.

It seems that this probably occurred as part of the changes in https://github.com/openwebwork/webwork2/pull/1371 . That PR certainly added a section with 3 `contant` labels under the comment line
```
# capture names for maketext
```
and it seems that they are needed as the text displayed seems to be set by https://github.com/openwebwork/webwork2/blob/b9b4e45b771b5e250eb275f3aa2c88337203d879/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm#L595 which uses `$r->maketext(ucfirst(WeBWorK::split_cap($actionID)))`  but that an additional one
```
use constant mk_publish => x('Import');
```
will be needed.

@drgrice1 - Do you agree that this is the cause of the problem and the proper approach to fixing it?

taniwallach:
The same issue effects also the user editor page.

drgrice1:
The reason that the string "Import" was in the po/pot files for WeBWorK-2.15 is due to the fact that maketext was called on the original (now removed) ProblemSetList.pm file on line 1238.  The 3 lines under `# capture names for maketext` were not added by the pull request, but were moved from another place in the ProblemSetList2.pm file (where they shoudn't have been).

I agree that adding `use constant mk_import => x('Import');` will fix the problem, and that is probably good enough for now.

However, I do not agree that this is the correct way to fix the problem.  In fact, this instance is a case in point that this is not the correct way to do this.  Anytime that you are calling maketext on something that is not a literal string, you are probably doing it incorrectly.  The problem is that this technique of calling maketext does not ensure that the strings coming from something like `ucfirst(WeBWorK::split_cap($actionID))` are all properly captured as there is not a direct correlation to the possible strings that could be translated.  I suspect there are other strings that are not translated in this case also.  For example, I don't see the strings "Cancel Edit", "Save Edit", "Save Export", or "Cancel Export" in the po/pot files either.  Those were also not there in WeBWorK-2.15.

[...]

PR #1479, opened 2021-08-16
Fix for issue 1477 - uses a revised method of getting "tab titles" in Classlist editor / Homework set editor /Achievement editor into the translation files

taniwallach:
Fixes https://github.com/openwebwork/webwork2/issues/1477 - that several strings used by these editors are not available for translation.

Reviews should please focus their attention on the first commit which follows the approach suggested by @drgrice1  in https://github.com/openwebwork/webwork2/issues/1477#issuecomment-899478389 and used already in `lib/WeBWorK/ContentGenerator/Instructor/AchievementEditor.pm` and
`lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm` which seems simpler that the other proposal.

That approach groups the "tab titles" into a single constant hash in a single block of code near the top of each file.

The second commit updates all the translation files, adding the missing strings used by these editors. Due to many other small changes since the last update - there are very many line number changes in the revised files.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
@@ -51,7 +51,7 @@ use warnings;
 #use CGI qw(-nosticky );
 use WeBWorK::CGI;
 use WeBWorK::Debug;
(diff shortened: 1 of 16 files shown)


drgrice1 (approved):
Looks good.  Thanks for taking care of this.

drgrice1:
This needs to be merged.

PR #1484, opened 2021-08-24
update location of mathquill files

drdrew42:
This pair of urls didn't get updated when we moved the js resources around...


drgrice1:
Good catch.  

I didn't realize that we ever had the mathquill files in vendor.

drdrew42:
Same. So these URLs have probably been broken for a lot longer...

drgrice1:
I think these pull request can be directly merged as they are an obvious problem with an obvious fix.  So I will do so.

PR #1485, opened 2021-08-24
Bugfix/update mathquill url

drdrew42:
matching PR to update develop


drgrice1:
I think you will need to rebase into develop.

drdrew42:
LOL, yep, forgot that this wouldn't be that easy ;P

Issue #1488, opened 2021-08-26
 Checkbox for Open in new window in Problem Editor is not functioning correctly


jutrembBDEB:
In Problem Editor, the is two checkboxes for "Open in new window", in the View and Save tabs, and both are given the ID "newWindow". This creates a problem because in the JavaScript, when it asks to see if the checkbox is checked in the Save tab, it is always looking at the checkbox in the View tab instead. So, in the Save tab, the checkbox is not behaving correctly. Here's the related code in pgproblemeditor.js.

```
if ((actionView && actionView.classList.contains('active')) || (actionSave && actionSave.classList.contains('active'))) {
			target = document.getElementById("newWindow").checked ? "WW_View" : "pg_editor_frame";
```

In the precedent versions of WeBWorK, there was only on checkbox for "Open in new window", that why it didn't have this problem. I think the solution is to give both checkbox different ID, like newViewWindow and newSaveWindow. Here's what I would do 

```
if((actionView && actionView.classList.contains('active'))){
	target = document.getElementById("newViewWindow").checked ? "WW_View" : "pg_editor_frame";
} else if ((actionSave && actionSave.classList.contains('active'))) {
	target = document.getElementById("newSaveWindow").checked ? "WW_View" : "pg_editor_frame";
}
```

PR #1489, opened 2021-08-26
Fix the "Open in new window" checkbox behavior for the problem editor.

drgrice1:
This fixes issue #1488.  Thanks for pointing this out @jutrembBDEB.


mgage:
This fix seems fine to me.  I'll merge it.

PR #1490, opened 2021-08-26
Fix the "Open in new window" checkbox behavior for the problem editor (for develop).

drgrice1:
This fixes issue #1488.  Thanks for pointing this out @jutrembBDEB.

### DIFF
--- a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
@@ -41,9 +41,10 @@
 		var actionSave = document.getElementById('action_save');
 
 		var target = "_self";
-		if ((actionView && actionView.classList.contains('active')) || (actionSave && actionSave.classList.contains('active'))) {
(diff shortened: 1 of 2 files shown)


pstaabp:
@drgrice1 First, a little nudge to remind to clear the cache when testing would be helpful on PRs that require such.  

I'm also not exactly sure what this is supposed to do.  I was expecting two separate new windows, one for each under "View" and "Update" when the checkbox is selected under both.  I'm not getting that.  I'm only getting a single new window.

drgrice1:
What should happen is that only the checkbox in the active tab will have effect.

drgrice1:
Further clarification...

Currently if you check "Open new window" when the "View" tab is active, and then change to the "Update" tab, leave the now visible checkbox unchecked, and click take action a new window is opened even the the checkbox visible on the page to open in a new window is not checked.  This is unexpected behavior.

With this pull request that does not happen.

pstaabp (approved):
I now follow the problem this resolves.  Looks good.

mgage:
I already merged the PR to main (#1489).  I'll merge this one as well.

Issue #1378, opened 2021-05-24
langage from the course is not rendered in simple template


jutrembBDEB:
I’m working on a project of translating some pretext book in french, like Active Calculus. We have our own WeBWorK server and I succeded in converting the book in html. Everything works fine, except the interface (buttons, message, etc) of the webwork exercices  is always in english, even though my course configuration is set in french (fr-CA)

What i think I understand is that in WeBWorK, the FormatRenderedProblem.pm is suppose to rendered the WeBWorK problem in the simple format, but localization to french is not working. 

$formLanguage = ($self->{inputs_ref}->{language})//'en';

This $formLanguage variable is suppose to contain the language, but is set automatically to « en » because language is empty. 

My server is 2.15, but @Alex-Jordan 
just tested with a 2.16 server and he have the same behavior.

drgrice1:
To set the language you will need to add the language argument to the url.  So the url should look something like `https://.../webwork2/...?sourceFilePath=...&language=fr-CA`.

Alex-Jordan:
OK, @jutrembBDEB I can make it so that PreTeXt attaches that to the URLs we use. Funny, I tried this earlier today, just shooting in the dark, but I tried `lang` instead of `language`.

mgage:
Did this resolve the problem?  supposedly maketext is being used to translate the buttons in FormatRenderedProblem.pm  (see lines 243 and following) https://github.com/openwebwork/webwork2/blob/WeBWorK-2.16/lib/FormatRenderedProblem.pm#L246

Alex-Jordan:
@drgrice's answer resolves @jutrembBDEB's question.

For the use case, I need to add a feature to PreTeXt, but that's not for this forum.

One thing though, the French localization file needs updating. If I use:
https://webwork-ptx.aimath.org/webwork2/html2xml?&amp;problemSeed=1&amp;answersSubmitted=0&amp;sourceFilePath=Contrib/PCC/BasicAlgebra/QuadraticEquations/solveQuad90.pg&amp;displayMode=MathJax&amp;courseID=anonymous&amp;userID=anonymous&amp;course_password=anonymous&amp;outputformat=simple&language=fr

(or with `fr-FR` or `fr-CA`)

then "Preview My Answers" and "Show Correct Answers" are still in English. @jutrembBDEB, would you be up for updating this? It is at:
https://github.com/openwebwork/webwork2/blob/WeBWorK-2.16/lib/WeBWorK/Localize/fr.po

[...]

PR #1491, opened 2021-08-26
Set the language for problems that are rendered in the library browser or in the homework sets editor.

drgrice1:
See the discussion in issue #1378.

This is a potential candidate for a hotfix, but I will leave that up for discussion at this point.

### DIFF
--- a/htdocs/js/apps/ProblemSetDetail/problemsetdetail.js
+++ b/htdocs/js/apps/ProblemSetDetail/problemsetdetail.js
@@ -338,6 +338,7 @@
 				'html{overflow-y:hidden;}body{padding:1px;background:#f5f5f5;}.container-fluid{padding:0px;}' +
 				'</style>';
 			if (window.location.port) ro.forcePortNumber = window.location.port;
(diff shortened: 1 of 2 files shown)


taniwallach:
I just ran into this PR.

This code run into an issue for Hebrew. The value expected by the translation code of WeBWorK is that which appears in the name of the PO file. For unknown historical reasons the Hebrew file was named `heb.po` while HTML uses `he`.

It might also if HTML gets a language code 

Most likely, we should align all the PO file names to be the same as the HTML language code to avoid problems, and which would make the code in this PR work well in all languages. In particular, we would need to change the name of the Hebrew po file (and change the option in `lib/WeBWorK/Localize.pm`).

Unfortunately, the change to the name of the PO file is likely to break many Hebrew courses, as `simple.conf` will also need to be updated to request `$language = 'he';` instead of `$language = 'heb';` as is done at present. Such a change would also be needed in any other place proving the language code as `heb` (ex. URL parameters for `html2xml`).

drgrice1:
I have added this to #1619, but done in a better way that will also work for Hebrew.  So I am withdrawing this pull request.

Issue #1427, opened 2021-06-28
Display mode: "None" is not respected in the Library Browser


dlglin:
In the Library Browser if you set the display mode to "None", and then click "View Problems", the problems are rendered using MathJax.  I believe the intent of the "None" option was to simply list the problems that match the search criteria without actually rendering them.

drgrice1:
This was fixed by #1494.

PR #1493, opened 2021-08-28
Do not render problems if the display mode is set to "None" in the library browser.

drgrice1:
This fixes issue #1427.

This may not be hotfix worthy.  What do you think?


mgage:
I think it can wait.  I suggest targeting it to develop

drgrice1:
I already made a matching pull request to the develop branch.  So I will just close this one then.

PR #1494, opened 2021-08-28
Do not render problems if the display mode is set to "None" in the library browser.

drgrice1:
This fixes issue #1427.


pstaabp (approved):
Looks good.

drgrice1:
This needs more work, so it has been converted to a draft pull request.  Although it basically works there are several issues.

If you select display mode `None`, load problems (which are not rendered), and then click on the re-randomize button, then the problem renders, and has errors because the display mode `None` is invalid for pg.  So I need to disable that button in this case.

Also, if have display mode `MathJax` set when the page first loads and load problems, and then as the problems are still loading select display mode `None`, the remaining problems have invalid display mode issue.  So either the display mode needs to be used at page load time all the way through, or the page should stop loading problems if this happens.

drgrice1:
Okay, this is ready to be tested again.  The issues described are fixed simply by moving the display mode check into the render function.

drdrew42 (approved):

PR #1498, opened 2021-08-31
Javascript component for Drag-N-Drop macros.

pschan-gh:
This pull request is in tandem with another one to be submitted to PG/develop, for updated versions of draggableProof-type macros.


drgrice1:
This will need to be cleaned up.  It looks like your branch was based off of main, and so this pull request is bringing in the hotfix commit history from that branch that is duplicated in develop.  This is a common issue with our method of synchronizing the main and develop branches.

One way to fix this is to rebase onto the develop branch.  That will involve some merge conflict resolution in this case, and so is not entirely trivial.

Alternately, you could create a diff from the develop branch to your branch, and then recreate this pull request from that diff.  You will lose your commit history, but looking at your commit history we will probably squash it anyway.

Something similar will need to be done with your corresponding pg pull request.

pschan-gh:
Since only a few files are modified, I will just create a new branch from develop and replace each file one by one. Thanks!

PR #1505, opened 2021-09-18
LTI tab in Course Config

Alex-Jordan:
This follows up a proposal I made in #1504. If you uncomment this code in authen_LTI.conf, Course Config will give access to some of the LTI-related configuration variables.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -376,14 +376,14 @@ $ConfigValues = [@$ConfigValues,
 		},
 		{
 			var => 'external_auth',
-			doc => 'Require students to log in through the LMS',
+			doc => 'Require users to log in through the LMS',
 			doc2 => 'If this is set, all users (including the instructor) must enter the WeBWorK course through the LMS. If a user reaches the regular WeBWorK login screen, they receive a message directing them back to the LMS.',
 			type => 'boolean'
 		},
 		{
 			var => 'LTIGradeMode',
 			doc => 'Grade passback mode',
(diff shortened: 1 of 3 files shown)


drgrice1:
If you copy authen_LTI.conf.dist to authen_LTI.conf, and then uncomment the `$configValues` that you added, then you can not save anything from the Course Configuration page without setting defaults for all of the variables in the `$configValues` that you added.  The `LMS_url`, `external_auth`, and `LTIGradeMode` variables do not have default values set in the current file.

If you try to change any setting from the Course Configuration page (or even not change any setting), then click "Save", you get the message "Use of uninitialized value $cmpoldval in string eq at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/Config.pm".  So the LTI config tab is completely useless until defaults are set in authen_LTI.conf.

Perhaps you should also add default values for those variables in this pull request.

Alex-Jordan:
Ah, I knew about that but apparently forgot to commit it with this PR. Our production server has the defaults set. Will update, thanks for looking at it.

Alex-Jordan:
I added defaults to variables. Not just the ones that are in that config tab, but to all of them. If it is a string and it previously did not have a default, its default is now `''`. If it is boolean and it previously did not have a default, its default is now `0`.

I changed how that big block of code is commented to use `=begin comment ... =cut` so it's easier to uncomment.

I edited the code in other places in ways that I think clean it up, although I'm not sure what our coding standards are for some things. If you see something that should be done differently, let me know and I'll change it.

drgrice1 (approved):
This looks good.

What you have done in terms of clean up is good.  You could go a bit further and run perltidy on the file with the .perltidyrc file in @pstaapb's pull request openwebwork/pg#618 (also in openwebwork/webwork3).  I did that, and it doesn't change a whole lot though.  It just aligns things nicely.

Alex-Jordan:
If you'd like to, please go ahead and add a commit following using perltidy. I'm out of time for WeBWorK stuff today.

drgrice1:
This is fine as it is.  There is no need to perltidy it.  Maybe at some point we will perltidy the entire codebase.  That will definitely be done with pg.  As to webwork2, that depends on how much longer we continue with webwork2 versus webwork3.

pstaabp:
Since the config settings are in authen_LTI.conf, if any changes are made in future releases, this would mean copying over again.  This might be better in defaults.config and then have a switch in authen_LTI.conf to show/hide the configuration?

drgrice1:
I am going to merge this.  In another pull request that I am working on I will move this to defaults.conf as @pstaabp suggested with another flag to enable/disable it in authen_LTI.conf.  Not only for the reason that @pstaabp suggested as to it being in a file that will need to be copied over, but I am also not sure that this should really be in a file that is edited directly (even if it is by the system administrator).

PR #1506, opened 2021-09-19
Hotfix? Update for WordPress Q&A plugin

drdrew42:
CityTech is releasing a [Q&A plugin for WordPress](https://github.com/openlab-at-city-tech/webworkqa).

The design is intended to accommodate more than just WeBWorK, so the courseID needs to be included in the submitted form-data, rather than parsed from the URL.

The release date is coming up very quickly, and this *should* be the final update necessary before release -- otherwise, this would be directed at develop.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Problem.pm
+++ b/lib/WeBWorK/ContentGenerator/Problem.pm
@@ -2257,6 +2257,8 @@ sub output_past_answer_button{
 sub output_email_instructor{
 	my $self = shift;
 	my $r = $self->r;
+	my $urlpath    = $r->urlpath;
+	my $courseName = $urlpath->arg("courseID");
 	my $problem = $self->{problem};
 	my %will = %{ $self->{will} };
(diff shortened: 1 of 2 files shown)


lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -2257,6 +2257,8 @@ sub output_past_answer_button{
 sub output_email_instructor{
 	my $self = shift;
 	my $r = $self->r;
+	my $urlpath    = $r->urlpath;
+	my $courseName = $urlpath->arg("courseID");
```
drgrice1, line 2261: There is no need to create the `$urlpath` variable since it is not used anywhere else.  This should be changed to
```suggestion
	my $courseName = $r->urlpath->arg("courseID");
```

drdrew42: Thanks, you're right.

drgrice1 (approved):
Looks good.

Issue #1512, opened 2021-10-12
Possible issue with nonce checking


dlglin:
This is based on https://webwork.maa.org/moodle/mod/forum/discuss.php?d=6367#p17322

I don't fully understand the code, but shouldn't https://github.com/openwebwork/webwork2/blob/cfb690ade3d3c55527d522ed1a8f1defd5e19f6a/lib/WeBWorK/Authen/LTIAdvanced.pm#L811 be `$db->putKey($Key);`?

Is the reason that no one else has reported this behaviour because that block of code is almost never invoked?

taniwallach:
I agree this is a bug. There is no `put()` method to call and that is the reason for the error message:
```
Can't locate object method "put" via package "WeBWorK::DB" at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 811.
```

Looking at `lib/WeBWorK/DB.pm` it certainly seems that `putKey()` is what is needed here.

The code here is meant to trigger when the provided nonce shows up in the database, which should not _normally_ occur as nonces are not supposed to get reused "too soon". So yes, it should not be happening often, which is probably why it took so long for the issue to get reported.

Note: The condition under which is triggered was somewhat changed relatively recently (https://github.com/openwebwork/webwork2/commit/9966a0d78f27ac2026664b85c77af2df8eb330c0) dropping the test  `$Key->timestamp <  $self->{"timestamp"}`. I had assumed when I dropped that test that the `$self->{"timestamp"}` should essentially always be the larger value when the code got there. However, as the timestamp used is actually coming from the `$self->{oauth_timestamp}` provided by the remote LTI consumer, that is not a valid assumption.

I just made https://github.com/openwebwork/webwork2/pull/1519 to fix the bug and reinstate the conditional for when the nonce database record gets updated.

drgrice1:
This was fixed by #1519.

PR #1519, opened 2021-11-01
LTI bug fix + reimplement test of when to update the nonce in the DB

taniwallach:
**This has been rebased to be a PR to develop and force pushed.**

Fix "$db->put($Key)" bug.

See: https://github.com/openwebwork/webwork2/issues/1512

Put back the test condition which was removed in https://github.com/openwebwork/webwork2/pull/1199 for when the nonce should be updated in the DB. The value being used as an update comes from the `oauth_timestamp` value provided by the LMS (remote LTI consumer) so should not be assumed to be larger than the value currently stored in the database.

I'm not sure how we can trigger the code, as this code only runs when a remote LMS sends an LTI connection using a nonce which is still stored in the database.

### DIFF
--- a/bin/ww_purge_old_nonces
+++ b/bin/ww_purge_old_nonces
@@ -57,6 +57,8 @@ use lib "$pg_dir/lib";
 use WeBWorK::CourseEnvironment;
 use WeBWorK::DB;
 
+use constant NONCE_LIFETIME => 21600; # 6 hours
+
 sub usage {
 	print STDERR "usage: $0  course \n";
 	exit 1;
@@ -79,11 +81,13 @@ my @listKeys = $db -> listKeys();
 
 foreach my $user_id (@listKeys) {
 	my $Key;
-	eval { $Key = $db -> getKey($user_id);};
+	eval { $Key = $db->getKey($user_id);};
 	if ($@) { push @errors, "$user_id: ". $@ ;}
 	else {
-		if ($Key -> key eq "nonce" && $Key -> timestamp +10 < time()) {
-			eval {$db -> deleteKey($user_id);};
+		if ($Key->key eq "nonce" &&
+			(time()-$Key->timestamp > NONCE_LIFETIME)
+		) {
+			eval {$db->deleteKey($user_id);};
 			if ($@) { push @errors, "$user_id: ". $@ ;}
 		}
 	}
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -806,9 +806,12 @@ sub ok {
     return 1;
   } else {
     # The nonce is in the database - so was used "recently" so should NOT be allowed
-    # Update timestamp - so deletion will be delayed from now and not from original timestamp.
-    $Key->timestamp($self->{"timestamp"});
-    $db->put($Key);
+    if ( $Key->timestamp <  $self->{"timestamp"} ) {
+      # Update timestamp - so deletion will be delayed from the most recent value
+      # of oauth_timestamp sent by the LTI consumer and not from an earlier timestamp.
+      $Key->timestamp($self->{"timestamp"});
+      $db->putKey($Key);
+    }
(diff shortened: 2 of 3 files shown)


drgrice1 (approved):
After looking at this closely, this is certainly correct.

First, there is no `put` method, and obviously it should be `putKey`.  It seems that bug goes back to the creation of this file.  Note that this same issue is on line 870 of `LTIBasic.pm`.  Could you also fix that while you are at it.  Most likely @goehle copied this from there, and in that file this dates back to 2012 when it was created.

Also obvious is that the timestamp in the database should not be allowed to be set to an earlier timestamp, and so the check that the previous timestamp is less than the timestamp in this request is needed.  This check is not in `LTIBasic.pm` and should be, so could you also add it there.

drgrice1:
By the way, I am able to recreate the issue here by using a local setup on my test server (my local computer).  I set the NonceLifeTime to a long time (3 minutes was enough), and set my "fake" lms to send a repeated nonce, or send a repeated nonce with an old timestamp, and I can see that this behaves in the correct way with this pull request.  That is authentication fails with the repeated nonce, and the timestamp is only updated if it is newer than the timestamp in the database.  Without this pull request, all of those cases just give the error `Can't locate method "put"`.

drgrice1:
As to this being a hotfix, I am not sure that it is that critical.  An LMS should not be sending repeated nonces.  I doubt that happens very frequently.  Furthermore, this just prevents an error from being shown in that case, and makes the correct "Your authentication has failed" message be shown instead.  The student still will not be able to log in with that nonce.

This is perhaps more likely to occur due to someone trying to hack the account, than an LMS using a repeated nonce.  (Although even that seems unlikely -- it seems that it would be difficult to find the exact nonce that was last used???)

drgrice1:
Thanks for re-targeting @taniwallach.

If anyone disagrees, and thinks this should be a hotfix, please speak up.

taniwallach:
> Note that this same issue is on line 870 of `LTIBasic.pm`. Could you also fix that while you are at it.

Done in the revised branch.

> Also obvious is that the timestamp in the database should not be allowed to be set to an earlier timestamp, and so the check that the previous timestamp is less than the timestamp in this request is needed. This check is not in `LTIBasic.pm` and should be, so could you also add it there.

I changed the logic in `LTIBasic.pm` to match what is now in `LTIAdvanced.pm` in terms of when the stored timestamp is revised, and also to reject connection attempts using an "old" nonce.

I noticed that `LTIBasic.pm` is pretty broken on its own. It depends on a cron job being run to purge old nonces. The file in `bin/ww_purge_old_nonces` would delete nonces after they are 10 second old. That is really fast, and probably makes replay attacks feasible. I modified it to use a 6 hour duration, and set the value using the line
```
use constant NONCE_LIFETIME => 21600; # 6 hours
```
to be as similar to  `LTIAdvanced.pm` as possible.

Maybe we should drop `LTIBasic.pm`? Is there any real reason to continue to maintain and support it?

Back in the WW 2.13 release notes it was indicated that `LTIBasic` "should be considered legacy code."

We could probably make it reject all connections and "warn" that the connection attempt was made using a mechanism no longer supported.

> Thanks for re-targeting @taniwallach.
> 
> If anyone disagrees, and thinks this should be a hotfix, please speak up.

No one complained or begged for the past few months, so I think anyone who really needs the patch for WW 2.16 can just manually install it.

drgrice1:
Looks good. Dropping LTIBasic.pm sounds good to me.

dlglin:
I also think dropping LTIBasic.pm is the right way to go.  AFAIK it is completely superseded by LTIAdvanced.pm.

Alex-Jordan (approved):
FWIW, I eyed through this code and of course I don't see anything bad or you would have caught it already. I see one unnecessary space character being added :)

I doubt I will have bandwidth to actually test this any time soon. I trust @taniwallach and @drgrice1 have this sorted out well.

drgrice1:
I forgot about this.  I am merging this one.

PR #1520, opened 2021-11-04
Fix Correct/Incorrect Select element UI

jamesweber7:
Gives elements with the _select_ tagname and _correct/incorrect_ classname UI to reflect whether the answer is correct or incorrect. UI addition matches preexisting style for elements with _correct/incorrect_ classname, _input_ tagname, and _text_ type (lines 767-790).

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -799,6 +799,30 @@ input.changed[type="text"] { /* orange */
   color: inherit;
 }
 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Thanks for this. For new development, we use the `develop` branch and view `main` as the locked version 2.16 (at present). Can you rebase the new commit off `develop` and then target a pull request to `develop` here?

PR #1521, opened 2021-11-05
Fix Correct/Incorrect Select element UI

jamesweber7:
Gives elements with the _select_ tagname and _correct/incorrect_ classname UI to reflect whether the answer is correct or incorrect. UI addition matches preexisting style for elements with _correct/incorrect_ classname, _input_ tagname, and _text_ type (lines 767-790).

### DIFF
--- a/htdocs/themes/math4/math4.css
+++ b/htdocs/themes/math4/math4.css
@@ -799,6 +799,30 @@ input.changed[type="text"] { /* orange */
   color: inherit;
 }
 
+select.correct { /* green */
+    border-color: rgba(81,153,81, 0.8);
+    outline: 0;
+    outline: thin dotted \9;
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
These rules for the `select` element are identical to those for the `input[type="text"]` elements.  So instead of duplicating them, just add them to the list that those rules apply to.  Like
```css
input[type="text"].correct, select.correct {
    ....
]
```

Although, this is good for now in any case.  I have an upcoming pull request that changes most of the css.  The css that is for problems like this will be moved into a separate file that will eventually be moved into github.com/openwebwork/pg, and used by webwork2, webwork3, and the standalone renderer.  I will make sure that this is incorporated into those changes.

PR #1522, opened 2021-11-11
Switch to having PG insert mathquill responses

drgrice1:
A new translation option is passed to PG that notifies PG that MathQuill is enabled.  Webwork2 (or other front end) just needs to tell PG that mathquill is in use via this translation option.  This is also used on the PG side to modify parsing behavior to work better with MathQuill.  This parsing is only in effect when parsing a student answer (and of course only when MathQuill is enabled).

The mathquill latex answers are now stored in KEPT_EXTRA_ANSWERS instead of as an additional response appended to the answer, so the ProblemUtil.pm create_ans_str_from_responses method no longer needs to worry about skipping these for past answers.

Install mathquill via npm from my github branch of my mathquill fork that contains the changes needed to make the new parsing work.  Note that before this is merged, that fork should be merged into the main branch of openwebwork/mathquill and the htdocs/package.json install updated to use that branch instead.

Note that the package-lock.json file is upgraded to lockfile version 2.  This is due to having a newew version of npm installed (as needed for ww3 development).  If this is a problem I will see what I can do about reverting to lockfile version 1.  There may also be problems with the `npm install` or `npm ci` step as mathquill is actually built when that runs and older versions of npm may not work.  So hopefully someone with an older version (for example the default version on Ubuntu) can test this.

Edit:  The package-lock.json file is no longer lockfile version 2.  I switched that back to lockfile version 1, and also made some changes to https://github.com/openwebwork/mathquill/pull/1 (also switched that to lockfile version 1 and another minor change).  Otherwise things seemed to fail with the versions of node and npm in Ubuntu 20.04.


pstaabp (changes requested):
A suggestion would be to pull in openwebwork/mathquill#1 first and then change line 17 of package.json to pull from openwebwork instead.  Eventually that should be done and doing it now would be better.

drgrice1:
@pstaabp:. I agree, and that's was the plan all along.  I said so in my original comments.

pstaabp:
@drgrice1 I was just suggesting we do this before merging this in.

drgrice1:
> Install mathquill via npm from my github branch of my mathquill fork that contains the changes needed to make the new parsing work. **Note that before this is merged**, that fork should be merged into the main branch of openwebwork/mathquill and the htdocs/package.json install updated to use that branch instead.

Yeah, that is what I meant with this.

pstaabp:
I think we're both on the same page.  All I'm saying is that we actually need to do these things before merging. 

Would you like me to merge openwebwork/mathquill#1?  Then we can make that change, double check that mathquill is being pulled from openwebwork instead of your branch.  We can still have others test as well.

drgrice1:
Yeah, I have done that too.  Not reading the pull request messages closely enough that is.  Yeah, I am good with merging openwebwork/mathquill#1.  As soon as that is done, then I will make the change here.  I want to leave it as is here for testing purposes until that is done though.

pstaabp:
Just merged that in.  When you make the change to `package.json`, I'll retest.

drgrice1:
@pstaabp:  I updated the package.json file to point to the openwebwork/mathquill main branch.

pstaabp (approved):
Looks good now.

pstaabp:
Looking at something related, I notice there is some `$user->useMathQuill` in `Problem.pm` and `GatewayQuiz.pm`.  Is there a way for a user to switch this?  I didn't see anything in UserSettings.

drgrice1:
Yes, a user can change this from the user settings page.  It shows up if MathQuill is enabled for the course.  It doesn't say MathQuill though.  It says something about "live equation rendering".

pstaabp:
Yep.  Got it.

drdrew42:
With the removal of options like WIRIS, it's now more reasonable to change the text in course config and user settings to reference MathQuill specifically. The verbiage before was attempting to classify all equation editors/answer assist options together... is it worth changing?

drgrice1:
Well, the WIRIS option has not yet been removed.  So we probably should wait until that is actually done, but when that is done it would be good to make that verbiage change.

drdrew42 (approved):

PR #1523, opened 2021-11-16
Improve the MathQuill editor toolbar

drgrice1:
Much of this is just the changes that have been made to the MathQuill editor that are implemented for ww3 and the standalone renderer.  This is code that will eventually work for webwork2, ww3, and the standalone renderer directly.  Although for that to happen, webwork2 will need to be updated to use bootstrap 5.

Clean up the way that the MathQuill toolbar is added and removed when a MathQuill input is focused and blurred.

Make the MathQuill toolbar responsive to the window height.  This makes the toolbar work better for short windows, for example on a mobile device or the case of a short problem in an iframe.

Make the button separated and give the container a border and background to make the appearance more uniform and clean.

Change the toolbar buttons to a dark color.

Switch to using bootstrap tooltips instead of jquery-ui tooltips.  This takes an annoying work around as bootstrap 2 has a bug with tooltips that are positioned to the left.

I recommend viewing the changes ignoring white space changes.  That helps a little at least.

### DIFF
--- a/htdocs/js/apps/MathQuill/mqeditor.css
+++ b/htdocs/js/apps/MathQuill/mqeditor.css
@@ -1,139 +1,114 @@
-span[id^=mq-answer].correct /* green */
-{
-    border-color: rgba(81,153,81, 0.8);
-    outline: 0;
-    outline: thin dotted \9;
-    /* IE6-9 */
-    -webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(81,153,81, ,.6);
-    -moz-box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(81,153,81,.6);
-    box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(81,153,81,.6);
-    color:inherit;
+span[id^="mq-answer"].correct {
+	border-color: rgba(81, 153, 81, 0.8);
+	outline: 0;
+	outline: thin dotted \9;
+	box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 8px rgba(81, 153, 81, .6);
+	color:inherit;
 }
 
-span[id^=mq-answer].incorrect /* red */
-{
-    border-color: rgba(191, 84, 84, 0.8);
-    outline: 0;
-    outline: thin dotted \9;
-    /* IE6-9 */
-    -webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(191,84,84,.6);
-    -moz-box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(191,84,84,.6);
-    box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(191,84,84,.6);
-    color: inherit;
+span[id^="mq-answer"].incorrect {
+	border-color: rgba(191, 84, 84, 0.8);
+	outline: 0;
+	outline: thin dotted \9;
+	box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 8px rgba(191, 84, 84, .6);
+	color: inherit;
 }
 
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
This looks great.  I think you showed us this last week (or similar responsiveness).

drgrice1:
Yeah, I merged this pull request into my bootstrap update work.

Alex-Jordan:
I'm having some issues with this. I moved to develop to test this, and I haven't been on develop for a while. I saw that I needed to update the apache config file before I could restart apache. Should I be looking at anything else to update?

I ran `npm install` just in case that matters. And I did clear my cache.

What I experience is:

1. Load a problem with MQ enabled.
2. Click in an answer blank. The palette appears. I can see its new styling.
3. Resize window height until the text tool moves into a new column. (I think this is necessary to demonstrate the issue I am seeing.)
4. Click a tool button (say the square root button). 

Now undesirable things happen. I can't reliably get the same undesirable thing to happen each time, but it's always been one of the following.

- The cursor leaves the input field and nothing is printed there. The palette tool remains.
- The cursor leaves the input field and nothing is printed there. The palette tool disappears.
- Something is printed (say the square root radical) and the cursor stays there. I click the same button one more time, and then one of the above two things happens.
- Or like the last thing, but it takes two additional uses of the tool for one of the first things to happen.

I think this is probably related: if I just let my mouse hover over a button in Chrome, there is flickering. It looks like some tooltip briefly appears. If I hover over a button in FF, it's less of a flicker but I still see a constant toggling of some sort. I do not get this flickering or toggling if it's still only one column wide.

Can anyone verify the same happens for you? I am using Firefox on a Mac. I have these issues with Firefox and with Chrome, although I seem to get a few more clicks in with Chrome before I lose the palette.

Also worth noting: the styling differs on the second column between FF and Chrome. With FF, I get the items in the scond column vertically centered. With Chrome, they are top-aligned.

Another styling difference, hardly worth mentioning. In Chrome, the right absolute value bar is almost touching the blue box. But in FF, the blue box is better centered.

drgrice1:
@Alex-Jordan:  The issues you describe should be fixed with my latest commit.  There were two things.

The main thing that was causing the problems you describe was the tooltip positioning hack.  It was not taking the width of the toolbar into account.  So if your mouse was over a button that was not in the right most column of the toolbar, then your mouse was over the tooltip.  This caused the tooltip to open, after which the mouse is now hovering over the tooltip and not the button, and so the tooltip would be removed, after which the mouse is again hovering over the tooltip, and so the tooltip would be opened again, etc.

That hack is simply deleted in the bootstrap 5 update because bootstrap 5 doesn't have the position left tooltip bug.

The other issue that may have been contributing is a possible race condition between the window resize event and the toolbar focusout event in which the resize event could possible occur after the toolbar is deleted.  I am not sure exactly how as the resize event is disconnected before the toolbar is deleted, but I saw this occur and cause console errors a few times.  It is really hard to make this happen though.  In any case, the resize event now checks to make sure the toolbar is defined before trying to resize it.

drgrice1:
The issue of alignment in Chrome versus Firefox is unfortunately an issue with Chrome not honoring the css `justify-content: center` style for flex columns when the css `height` is not set, while Firefox does.  If consistent appearance between the two browsers is desired, then we can just remove the `justify-content: center` style, and then both (all) browsers will display the toolbar in the same way that Chrome does.

Alex-Jordan:
There is perhaps still an issue with the tooltip. I only discovered it by clicking fast to check that things were working now.

If there are two columns, and the mouse is hovering over a button in the right column, a tooltip appears to the left, over a button in the left column. Suppose I want to click that button in the left column, and it's only sloppy mousework that has put me over in the right column for now. So I move over to the button in the left column to click it. If I am too fast, the tooltip is still there and clicking dismisses the whole palette.

Can the tooltip go away immediately once the mouse is no longer hovering over the corresponding button?

drgrice1:
The tooltip does go away immediately as soon as the mouse leaves the button.  The tooltip is set with a hide delay of 0.  You really can't do any better.  I have to say, you really do have to do some fast clicking to achieve what you are saying.

PR #1526, opened 2021-11-29
FIX docker just exits w no err due to hardcoded db

ionparticle:
When checking if the admin table exists, the mysql command used the
hardcoded db host "db" when it should've used the variable
$WEBWORK_DB_HOST like the wait_for_db function.

The lack of error messages was due to stderr being blackholed into
/dev/null. Removing the blackhole, this message shows up in docker logs:

ERROR 2005 (HY000): Unknown MySQL server host 'db' (-2)


taniwallach:
The correction makes sense. I don't have time to try to set up an environment to test it at present.

drgrice1 (approved):
Good catch.  Thanks for pointing this out.

PR #1529, opened 2021-12-01
on problem sets page, show quiz start date using local timezone/locale

Alex-Jordan:
This addresses item 4 on #1528. Prior to this commit, a quiz version start time is displayed using the server's local time. This changes it to use the course-local timezone locale, the same as other dates are displayed like the close date or open date of a problem set.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSets.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSets.pm
@@ -669,8 +669,11 @@ sub setListRow {
 	  
 	  if ( defined( $set->assignment_type() ) && 
 	       $set->assignment_type() =~ /gateway/ && $gwtype == 1 ) {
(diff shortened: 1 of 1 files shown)


taniwallach:
This certainly makes sense, and makes things consistent for the end users.

Just for record:
  - When using `$self->formatDateTime()` the `undef` as the second argument forces a fallback to using `$ce->{siteDefaults}{timezone}` as the value of `$display_tz`.
  - That value can be set in a course's `simple.conf` to override the value from `site.conf`. The value set there is what is meant by a course specific time zone.

drgrice1 (approved):
This looks good.   I like the locale formatted time better even if it were not in the local timezone.  Of course it should be in the configured timezone though.

pstaabp (approved):
Agreed.  Looks good.

drgrice1:
I will go ahead and merge this now that we have two approvals.

PR #1499, opened 2021-08-31
Another attempt of a pull request for Drag-N-Drop

pschan-gh:
Paired with openwebwork/pg#602.

Javacript and config files in support of the Drag-N-Drop module DragNDrop.pm and associated  macros draggableProof.pl, draggableSubsets.pl.

PR #1532, opened 2021-12-30
Update to Bootstrap version 5 and rewrite the UI

drgrice1:
Bootstrap version 2 has been removed from the vendor directory, and version 5 is installed via npm.  Make sure to run `npm install` or `npm ci` to test this pull request.  Note that css classes are inserted by perl directly rather than adding them later with javascript.  This prevents the flash of unstyled content when the page loads.

Theming is now done quite differently.  Instead of modifying the override files (which still exist and are not updated), copy an existing theme directory and modify the files _theme-colors.scss and _theme-overrides.scss.  Then run `npm run generate-css` from the htdocs directory.  Note that the override files may still be used if desired, but that approach will take a lot more work to get a nice theme.  Note that running `npm install` will automatically execute the `generate-css` script, and so no further action is required to use the existing themes.

The existing themes are updated, and math4-yellow theme is added.  The point of adding the yellow theme is to demonstrate how to implement a theme with a light primary color.  (The colors are the MWSU colors though, and this is the theme I will be using!)

You can also run `npm run sass-watch` from the htdocs directory to automatically compile the css files anytime one of the scss files changes.  This is convenient when creating/modifying a theme.  Note that `npm run generate-css` will generate compressed css files with no source maps for production use, while `npm run sass-watch` will not generate compressed css and will generate source maps which are more convenient for development, but not intended for production use.

There are a lot of changes to the UI that may need to be changed further, and lots of testing and discussion will be needed.  Perltidy was used on many of the edited portions of code with the rules agreed on for webwork 3 code, and unfortunately that changes things even more from the previous code.  Although the resulting code is much more readable when not comparing to the previous code.

A paired pull request (openwebwork/pg#623) updates scaffolds for Bootstrap 5.  Everything in this pull request will work without that one, with the exception of scaffolds in problems.


drgrice1:
Note that this also includes pull request #1523.  So I guess this depends on that one.

drgrice1:
In addition to testing on desktop/laptop computers, test on mobile devices.  Things should look much better there than before.

pstaabp:
Had to try it out.  Noticed on the front page:

image

Did you intend to include the bullets?

drgrice1:
You may need to clear your browser cache.  I am assuming that you ran `npm install` and restarted the server.

pstaabp:
Forgot to restart apache.  Duh.  Too much ww3 development work these days without apache restarts.  All set.

pstaabp:
Suggestion:  match the buttons in the library browser problems (upper right corner of each problem)
image

to similar to those in the Problem Set view: 

image

There's plenty of room on the top row of the library browser problem view.

drgrice1:
Sounds good to me.  I though about doing that, but left it.  I will add that.

pstaabp:
Another suggestion.  When there is information in the red or green box, like:
image

To make a bootstrap alert: https://getbootstrap.com/docs/5.1/components/alerts/
That should be easy (I hope).  Looks like it is a `ResultsWithoutError` class and there is probably a similar one for errors.

drgrice1:
The bootstrap alerts seem a bit obtrusive to me for this.  They add a lot of padding.  I could reduce that perhaps, but it will take some care as the ResultsWithoutError class is used for a lot of things.

I used bootstrap alerts for the warnings for errors in problems, so you can see what they look like there.  Of course bootstrap 2 alerts were used for this before, but the class name is different (alert-error doesn't exist anymore, it is now alert-danger).

drgrice1:
@pstaabp: I added a couple of commits that do the things you suggested here.  I actually had implemented something like what you suggested with bootstrap alerts before, but I took that out.  I added it back in, but also tweaked the problem styles some.

pstaabp:
Just checking out the new commits.  I'm getting the pi symbol at full scale in both Safari and in Chrome and it appears that it is 1700 x 1700 and doesn't even fit on my screen.  Looking at the html and css, I don't see any rescaling in that code.

drgrice1:
Make sure that you run `npm install`, `npm ci`, or just `npm run generate-css` again.  There is css that takes care of this, but you need to update your build.

pstaabp:
Yep.  and the new icon looks very good.

pstaabp:
Playing with the new theme, you stated above that running `npm run gen-theme-css` updates the css for the theme.  It appears this should be `npm run generate-css`.  I didn't see a `gan-theme-css` in the `package.json` file.  

And the watch command should be `npm run sass-watch`. Or rename to `Sasquatch`? ;)

drgrice1:
You are correct.  I changed the name of the script, and forgot.  It should be `npm run generate-css`.  When I first set things up the only css that was generated was in the theme directory.  But then I moved the problem.scss file from the themes into js/apps/Problem and made the script also generate that css file.  Then the gen-theme-css name didn't make sense anymore.

pstaabp:
Looks like the date picker is still a jQuery-ui element.  Correct?  I guess bootstrap doesn't have one and there doesn't seem to be a simple way to do what we have currently.  Maybe we'll continue with jQuery a bit longer?

drgrice1:
Yeah, the datepicker is still jquery-ui.  But there are other things that use jquery-ui also.  jquery.mjs.nestedSortable.js that is used in ProblemSetDetail.pm and jquery.nestable.js that is used by the draggableProof.pl macro (and loaded from cloudflare) depend on it.  I may try to replace those things before the release yet.  We'll see what I get to.

pstaabp:
Looking at the problem grader and I'm getting the following error on save:
```
problemgrader.js:100 Uncaught TypeError: Cannot read properties of null (reading 'value')
    at Object.success (problemgrader.js:100)
    at c (jquery.min.js:2)
    at Object.fireWith [as resolveWith] (jquery.min.js:2)
    at l (jquery.min.js:2)
    at XMLHttpRequest.<anonymous> (jquery.min.js:2)
```

Tried this on a few different problems.

drgrice1:
I am not seeing the error that you are seeing, but the score isn't saving.  I will look into it.

drgrice1:
@pstaabp:  The problem grader issue is fixed.  Thanks for all of your testing.  That is extremely useful.  Hopefully others help with this too.

pstaabp:
This fixed the error I was getting and also saving the score correct.

pstaabp:
I'm getting errors related to DragNDrop.pm missing.  Did #1499 get merged into this or maybe needs to be?

drgrice1:
I have merged the current develop branch into both this pull request and the corresponding pg pull request.  I am not getting that error.

pstaabp:
Found a strange css error.  In safari, on the Homework Sets editor (but the same for every select object), it looks like:
image

It's taking the native select appearance and overlaying the chevron.  However on the bootstrap site, a select looks like:
image

Drilling down into the css using the browser, the getbootstrap.com site seems to be using bootstrap.min.css and here we are using bootstrap.css.  Maybe there's a discrepancy between the two.  Possibly a bug. I'll investigate more.

drgrice1:
We are using the custom css file htdocs/themes/current_theme/bootstrap.css.  It is generated when you run `npm install`.  Check that that file is up to date by running that again.  Or just run `npm run generate-css`.

drgrice1:
If the generated bootstrap.css file is up to date and this still happens, then use the DOM inspector and see if you see the bootstrap styles being applied in the css rules.  You could also try using `npx sass themes` from the htdocs directory instead to get source maps and get more information in the DOM inspector.

drgrice1:
By the way you can compare the files htdocs/themes/math4/bootstrap.scss and htdocs/node_modules/bootstrap/scss/bootstrap.scss to see the difference between what bootstraps defaults are and what my custom file uses.  I use the recommended approach for this on the bootstrap website.  They recommend not using their default file (the one you get from getbootstrap.com) and customizing with your organizations colors and such.  See Option B https://getbootstrap.com/docs/5.0/customize/sass/.  This is the same approach used for the bootstrap.scss used for webwork3.  Although there we need to omit some parts of bootstrap that conflict with quasar.  I also leave out several other things.  In that case I was trying to only pull in the things that are needed for pg.

pstaabp:
Got the dragNdrop fixed.  Hadn't done a pull on the PG side of things.  

Looks like in `node_modules/bootstrap/dist/css/bootstrap.css`, there is a `-webkit-appearance: none` line (line 2415), whereas the corresponding one that gets generated is in `node_modules/bootstrap/scss/forms/_form-select.scss` has an `appearance: none`, however, my browser says that's not valid.  Hmmm...  Seems like either a safari or bootstrap bug.

pstaabp:
Aah... GitHub issues has solved it.  https://github.com/twbs/bootstrap/issues/35614

It seems that we need to run some postcss.  Bootstrap recommends https://github.com/postcss/autoprefixer.  We may need to install something else and run that.

pstaabp:
Just put in a PR to your branch that fixed this for me.

pstaabp:
When inputting via WIRIS, there is an icon (on the develop branch) inside an input element.  It's missing on this branch.  

Note: it doesn't look like this is completely working on develop.  I've got a matrix as an input and the icon is only showing up on the upper left input of a matrix.   Do people use this?  It's much more clumsy than mathquill.  Perhaps we can deprecate this for 2.17?

drgrice1:
Note that if you are testing this on Ubuntu 18.04 (or any distribution that has a version of node prior to verison 10), then you will need to install a newer version of node that is not provided by your distribution.  You can install node version 16 on Ubuntu 18.04 (or 20.04) with the commands:
```bash
curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash -
sudo apt-get install -y nodejs
```
For other distributions see https://github.com/nodesource/distributions/blob/master/README.md

drgrice1:
There should be a lot of css rules that are invalid.  Invalid css rules are ignored.  That is the way css works.  Those are there for the browsers for which they are valid.  Firefox shows almost all of the webkit prefixes as invalid.

drgrice1:
> When inputting via WIRIS, there is an icon (on the develop branch) inside an input element. It's missing on this branch.
> 
> Note: it doesn't look like this is completely working on develop. I've got a matrix as an input and the icon is only showing up on the upper left input of a matrix. Do people use this? It's much more clumsy than mathquill. Perhaps we can deprecate this for 2.17?

Wiris doesn't work.  It never has.  That should be removed.  It was never fully implemented, and relies on proprietary software to work.

drgrice1:
I fixed the wiris styles so that you can see the icon again.  I still thing that wiris should be removed.  It has lots of problems, and is a bad idea conceptually to begin with.

pstaabp:
I would agree to remove WIRIS.  I don't think it's necessary--MathQuill is much better.  Removing it may be another new feature of 2.17.    :-)

pstaabp:
> There should be a lot of css rules that are invalid. Invalid css rules are ignored. That is the way css works. Those are there for the browsers for which they are valid. Firefox shows almost all of the webkit prefixes as invalid.

What Safari was showing though was that `appearance: none` is invalid, whereas `-webkit-appearance: none` is fine.  I think MDN says that Safari should accept that after Safari 3.  It's on 15 now.  Strange.

drgrice1:
Yeah, well the autoprefixer is fixing it for now.

If the developers that introduced wiris want to maintain it, then we can leave it, but otherwise I think it should go.

pstaabp:
Thought about the Math Objects button in a problem (in the OPL):  Right now, clicking opens a new tab.  Since the total amount of info is fairly small, how about a popover for this instead and maybe add a link to the MathObject documentation in the wiki.

drgrice1:
Sounds good to me.

lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
```diff
@@ -752,91 +751,154 @@ sub displaySets {
...
+	#####################################################################################
+	print CGI::p($r->maketext(
+		'A period (.) indicates a problem has not been attempted, and a number from 0 to 100 '
+			. 'indicates the grade earned. The number on the second line gives the number of incorrect attempts.'
+	));
+	print CGI::p($r->maketext(
+		"Click on a student's name to see the student's version of the homework set. "
+			. "Click a heading to sort the table."
+	));
+	print CGI::p(
```
pstaabp: Should these strings on this and next two lines be wrapped in `maketext` ?

pstaabp (approved):
Overall, I think this is a solid update.  Since it touch nearly every part of webwork, reviewing everything will take time, but I would recommend we pull in sooner than later, so we can build around this.  If minor issue come up, we can fix them as they come.

htdocs/js/apps/Problem/problem.scss
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	input[type=radio] {
+		margin-right: 0.25rem;
+	}
+
+	select {
+		cursor: pointer;
+
+		&[multipele], &[size] {
```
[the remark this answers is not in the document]
drgrice1: This is fixed.  Thanks for pointing that out.

lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
drgrice1: Yes.

drgrice1: I think that you are asking if it is okay to do this sort of splitting of strings in a maketext call.  The answer is yes.  When gettext runs it merges those strings.  The resulting po files are the same whether the string is one lone string or a concatenated string as is done here.

drgrice1: This is what ends up in the po files either way:
```
#: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm:875
msgid ""
"Click on a student's name to see the student's version of the homework set. "
"Click a heading to sort the table."
msgstr ""
```
and this is what ends up in the webwork2.pot file either way:
```
#: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm:875
msgid "Click on a student's name to see the student's version of the homework set. Click a heading to sort the table."
msgstr ""
```
The perl `xgettext.pl` generates the pot file first, and it merges the strings.  Then `msgmerge` generates the po files from that, and it puts the strings onto separate lines to make it easier for translators to read (I think).

PR #1533, opened 2021-12-30
Fix for WWSafe not found in scripts

pstaabp:
This is a fix for a WWSafe not found error when running scripts in the `WEBWORK_ROOT/bin` directory.


drgrice1 (commented):
Thanks for dealing with this. This looks good.

I tested several of the scripts, but not all of them.  All of them are much the same, so they should all work as well as before at least.

drgrice1 (approved):

PR #1534, opened 2021-12-31
remove 'Inactive' from 'Hide Inactive Courses'

Alex-Jordan:
This just changes "Hide Inactive Courses" to "Hide Courses" on the admin page. I often have reasons to hide a course that is still active.


taniwallach (approved):
Looks fine to me and makes sense.

drgrice1 (approved):
I also hide active courses.  Particularly when external authentication is disabled for a course with LMS authentication.

Looks good.

Issue #1530, opened 2021-12-20
machine rounding error with conditional release


Alex-Jordan:
A machine rounding error has popped up, and I'm unsure what is the best way to address it.  An instructor had conditional release set to 80%. Meaning if a student has reached at least 80% on set A, they get access to start set B. For this instructor, set A has 5 exercises. Students who get 4 correct should reach 80% and gain access to set B. But they are not getting access.

I think that the score 4/5 is becoming 0.79995422... due to machine rounding error. And then that does not pass the test (because it is less than 0.8).

The comparison happens at line 1232 of lib/WeBWorK/Utils.pm. Instead of comparing `$r_score < $restriction`, it could become `$r_score + 0.001 < $restriction` or something like that. That should work in practice, but it feels like a hack to work around the machine rounding error. Can anyone imagine a better solution?

dlglin:
I don't see a better way to handle this than to add a fudge factor as you suggest.

drgrice1:
Is your 5 exercise example hypothetical, or is that what the instructor actually has?

I just tested this by creating a homework set with 5 problems, and then making another homework set conditional on that one with the 80% access requirement.  The result was successful.  After working four of the five in the first homework set correctly, the restricted set was opened.  Checking, the `$r_score` and `$restriction` values in the `sub is_restricted` method of `lib/WeBWorK/Utils.pm` on line 1232, I see that both are exactly 0.8 with no rounding.

Is there something else going on that is causing the rounding in your situation?  I left the question weights all at 1 in my first test.  Then tried again with all weights set to 2.  Both worked, but perhaps the instructor in your situation used different equal weights?  Or is this perhaps the problem?  I.e., the weights are not all the same, and so four questions may not be 80%?

Alex-Jordan:
It's not hypothetical, but it may have been an 8/10 situation. This was a few months back, and she indicated that changing the setting to 70% was not a good solution. That may indicate that it was indeed 8/10, not 4/5. I added a 75% option for her and that met her needs. I'll see if I can track down if it was really a 10-question set.

Alex-Jordan:
I looked at this instructors quizzes. The first one has a release condition for five sets: "L03,L04,L05,L06,L07". Those sets have problem counts: 5, 6, 7, 10, 5. There is no weighting in those sets (each problem uses weight 1).

I'll ask the instructor for permission to mess around in her course and try to reproduce it.

drgrice1:
Sounds good.

[...]

PR #1535, opened 2021-12-31
round values before doing conditional release comparison

Alex-Jordan:
This addresses #1530. The discussion there includes a way to reproduce the issue.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -1232,7 +1232,7 @@ sub is_restricted {
 	  	  my $restrictor =  $db->getGlobalSet($_);
 		  my $r_score = grade_set($db,$restrictor,$_, $studentName,0);
 		  # round to evade machine rounding error
-		  $r_score = sprintf("%.2f", "$r_score");
+		  $r_score = sprintf("%.2f", $r_score);
 		  if($r_score < $restriction) {
 	  	    push @needed,$_;
 		  }
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -1221,7 +1221,9 @@ sub is_restricted {
 	my @needed;
 	if ($set->restricted_release ) {
 	        my @proposed_sets = split(/\s*,\s*/,$set->restricted_release);
-		my $restriction =  $set->restricted_status  ||  0;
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/Utils.pm
```diff
@@ -1221,14 +1221,18 @@ sub is_restricted {
...
+		$restriction = sprintf("%.2f", $restriction);
 		my @good_sets;
 		foreach(@proposed_sets) {
 		  push @good_sets,$_ if $db->existsGlobalSet($_);
 		}
 		foreach(@good_sets) {
 	  	  my $restrictor =  $db->getGlobalSet($_);
 		  my $r_score = grade_set($db,$restrictor,$_, $studentName,0);
+		  # round to evade machine rounding error
+		  $r_score = sprintf("%.2f", "$r_score");
```
drgrice1, line 1235: There is a minor issue here.  The `$r_score` inside the `sprintf` should not be quoted.  That doesn't do anything useful and merely adds the inefficiency of perl needing to interpolate the string.

Alex-Jordan:
I removed those quotes.

BTW, at first I tried doing the rounding in the same line that initiates `$r_score`:

```
 my $r_score = sprintf("%.2f", grade_set($db,$restrictor,$_, $studentName,0));
```
I had an example where the student score was 0.75. But apache logging was showing that `$r_score` was coming out as 7.5 instead with the above line of code. I cannot explain this. 

I reconfigured it over two lines and that appears to work as intended. (The logs showed 0.75.) Those quotes were in there as part of my probing to understand how 7.5 could come out and I forgot to remove them.

drgrice1:
The reason that the single line code
```
my $r_score = sprintf("%.2f", grade_set($db,$restrictor,$_, $studentName,0));
```
did not work is because as when you put it as the second argument (or really the list of remaining arguments) for sprintf it is in list context.  The grade_set returns a list with first element the total right and the second element the total when it is called in list context.  If you instead use
```
my $r_score = sprintf("%.2f", scalar(grade_set($db,$restrictor,$_, $studentName,0)));
```
it would work.

drgrice1:
Note that `my $restriction` could also be directly assigned the sprintf value.  There you don't have to worry about the `scalar` usage as it is directly a scalar to begin with (no wantarray usage).

PR #1536, opened 2022-01-01
Add a pg htdocs location, and move the pg related files in htdocs there.

drgrice1:
This works by adding an apache2 rewrite.  When a file in the `$webwork_htdocs_url` (`/webwork2_files`) is requested, then the server first checks for the file in `$webwork_htdocs_dir`.  If the file is there, then that file is served.  If not it tries to serve the file from `$pg_dir/htdocs`.  The url for that location is `/pg_files`.

This is done in conf/webwork.apache2.4-config.dist, so make sure to update your copy of that file.

The relevant files in the webwork2 files used purely by pg problems have been moved to the pg/htdocs directory.  They are deleted in this pull request, and added in a corresponding pg pull request (openwebwork/pg#624).  This includes many of the javascript and css files in htdocs/js/apps, and the html files in htdocs/helpFiles that are used for student help.

The ADD_JS_FILE and ADD_CSS_FILE parsing methods in Problem.pm, GatewayQuiz.pm, and FormatRenderedProblem.pm have been updated to check for the existence of files in the pg htdocs directory as well.  Furthermore, the ADD_CSS_FILE parsing method has been updated to preserve the order that css files are added in the same way that javascript order is preserved.

Note that the 3rd party jsxgraph library is now installed via npm (on the pg side), and the files for that deleted from this repository.  Make sure to run `npm install` in the pg/htdocs directory for this.

Also the knowl.js and knowlstyle.css files are moved to htdocs/js/apps/Knowls in the pg htdocs directory and the locations
updated in the files here.

Note that the htdocs/package-lock.json file has been removed from the git repository.  It has become a pain to manage that file with varying versions of npm, and we don't really need that file for webwork2's usage.  As such the instructions to use `npm ci` will need to be replaced with `npm install` in the installation instructions.

Note that the file `htdocs/js/apps/Problem/problem.scss` has been removed and is now on the pg side.  Make sure to delete `htdocs/js/apps/Problem/problem.css` if that file exists to test this properly.  Of course also run `npm install` in `pg/htdocs` to generate the css file there.


pstaabp:
Just a couple of other notes: 

- Make sure that the `mod_rewrite.so` http module is installed.  This wasn't by default on my server. 
- `npm install` needs to be done in both `webwork2/htdocs` and `pg/htdocs`

drgrice1:
Actually, there is no need to run `npm install` in `pg/htdocs`.  That is added for future usage by the standalone renderer.

pstaabp:
I got an error related to JSX when loading a GraphTool problem without running `npm install` in `pg/htdocs` (I think).

drgrice1:
You shouldn't.  If you run `npm install` in `webwork2/htdocs` it installs the same thing.

pstaabp:
I guess the `npm install` in the webwork2 directory fixed that error.  Isn't the purpose of this to serve js/css off the PG side?  Why is jsxgraph still on the webwork2 side? 

Anyway, looks solid. 
Tested the following: 

- mathquill
- mathview
- wiris
- help pages for input (via knowls). Didn't test every help page...
- graphtool

Anything else to test?

Also, is any of the code in `htdocs/legacy` and `htdocs/vendor` used anymore?  Or maybe after the switch to bootstrap 5?

drgrice1:
Well, jsxgraph is on both sides, but really only in the package.json file, and installed on both if you run `npm install`.  The point was to avoid another installation step.  For webwork2 it really doesn't matter which side the file is on.  Apache2 checks webwork2/htdocs for anything requested, and if not found there serves it from pg/htdocs.  The standalone renderer will just serve these files from the pg location, and so it will use the jsxgraph installed via `npm install` in pg/htdocs.

Although it probably is a moot point as my bootstrap update adds the htdocs/js/apps/Problem/problem.scss file that will also need to be moved to the pg/htdocs location.  That will need to be compiled.  So I will remove jsxgraph from the webwork2/htdocs package.json, and add the new installation step.

drgrice1:
So @pstaabp will be correct.  Make sure to run `npm install` in `pg/htdocs` to test this.

drgrice1:
You could test the new draggable proofs macro now too.  I moved the js and css for that to the pg side as well.

I have another branch that cleans up the `htdocs/js/vendor` and `htdocs/js/legacy` directories (and guts htdocs in general).  It is pending on the bootstrap pull request though, and so I was waiting to put that in.

drdrew42 (approved):
requires: `sudo a2enmod rewrite`
tested:  
- [x] MathQuill
- [x] ImageView
- [x] DraggableProof
- [x] GeoGebra

drgrice1:
@pstaabp:  You merged the pg pull request that this is paired with.  This should be merged also.

taniwallach:
@drgrice1 Nice work. It makes lots of sense to move the dependencies needed for PG into pg.

I'm not sure why you decided to use the Apache rewrite approach. If we are checking that the files exist in many cases by checking 2 potential locations and coding the location by hand in others, why don't we try to make sure everything uses the real location?

---

For release notes: Sites which use `lighttpd` to server static file:
  - https://webwork.maa.org/wiki/Serving_static_files_with_lighttpd
  - https://webwork.maa.org/wiki/Installation_Manual_for_2.16_on_Ubuntu#Implement_Option_B_.28lighttpd.29
will also need to create a rewrite rule for `lighttpd`.

From a quick look, this will require some changes to the `Dockerfile` in order for Docker to work after these changes.
  - It seems to require adding the extra `npm install` step, as well as `a2enmod rewrite`.
There may be more.

drgrice1:
@taniwallach:  The rewrite rule approach was chosen because it does not require changing all of the links to these files that are generated everywhere in the webwork2 and pg code.  That is invasive and really not needed.

Instructions in the release notes can be added for those using lighttpd to serve static files.

If you can figure out what is needed for Docker, that would be great.

drgrice1:
@taniwallach:  What do you mean with we are "coding the location by hand in others"?  Where are we coding the location by hand?

As to the Docker changes, it seems that all that is needed is the `npm install` step in pg/htdocs.  `a2enmod rewrite` is already there.

Issue #1262, opened 2021-03-18
Update/validate external links in WW code


dlglin:
There are a number of links in WW to external sites.  A few examples:

- In the problem editor, there are links to the wiki and to PGlabs, among others
- The "report bugs" link to bugzilla

Some of these are currently broken.  There are a couple of things that should be done:

- Check that links still work.  If they don't, then either remove or update them.
- Consider changing these to aliases so we can change the destination without people having to update WW

Alex-Jordan:
I did a search for URLs to see what we are up against. One that occurs in almost every file is `http://openwebwork.sf.net/` in the copyright blurb. I think we should either just remove all instances of this, or if TWP has an official landing page, use that instead.

somiaj:
Closing this, all links mentioned no longer exist. There are a few external links pointing to `openwebwork.org` which we control.

PR #1537, opened 2022-01-05
fixes broken links.

pstaabp:
The ProblemEditor has many broken links.  This fixes two of them and hides a third.  This is a fix for #1262. 

In short, the demo course at Rochester isn't working and I don't think will be reliable in the future. I removed the PGLab link.  Is there somewhere else that is hosted?  The other two are pointing to documentation at webwork.maa.org.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -356,8 +356,8 @@ $webworkURLs{SiteMap}       = "https://webwork.maa.org/wiki/WeBWorK_Sites";
 $webworkURLs{problemTechniquesHelpURL}='https://webwork.maa.org/wiki/Category:Problem_Techniques';
 $webworkURLs{MathObjectsHelpURL}      ='https://webwork.maa.org/wiki/Category:MathObjects';
 $webworkURLs{PODHelpURL}              ='https://webwork.maa.org/pod/';
-$webworkURLs{PGLabHelpURL}            ='https://demo.webwork.rochester.edu/webwork2/wikiExamples/?login_practice_user=true';
-$webworkURLs{PGMLHelpURL}             ='https://demo.webwork.rochester.edu/webwork2/cervone_course/PGML/2?login_practice_user=true;
+$webworkURLs{PGLabHelpURL}            ='https://demo.webwork.rochester.edu/webwork2/wikiExamples/MathObjectsLabs2/2?login_practice_user=true';
+$webworkURLs{PGMLHelpURL}             ='https://demo.webwork.rochester.edu/webwork2/cervone_course/PGML/1?login_practice_user=true;
 $webworkURLs{AuthorHelpURL}           ='https://webwork.maa.org/wiki/Category:Authors';
 
 # Location of CSS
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
It is too bad that the lab links don't work.  Those were useful.  Is there any way that we could copy those courses to webwork.maa.org and host them there?  Who has access to those servers that could get archives of those courses?

The pod link should go to webwork.maa.org anyway.

dlglin:
Mike should have access to any of the Rochester demo sites.

There currently isn't a WW installation on webwork.maa.org.  If we install webwork there then someone will have to maintain it.

If we are going to point those links to a new server, I recommend registering a new domain, and having it resolve to wherever these things are going to live.  We could register demo.openwebwork.org, and then point it at whatever demo server we are using.  This way we can change the destination without people having to upgrade their WW installations.

drgrice1:
Yeah, there is the maintenance issue.  For the most part though, once it is installed the maintenance should be relatively minimal since it just needs to allow guest access to use the PG and PGML labs.  I could probably deal with this.

Also, the PGML lab course is not on the Rochester demo site.  At least the link in webwork doesn't point there.  The link in webwork points to courses1.webwork.maa.org.  So does someone have an archive of that course?

taniwallach:
The "PointAndGraph AppletDemos" assignment at https://demo.webwork.rochester.edu/webwork2/wikiExamples/PointAndGraph_AppletDemos/ should probably be removed from whatever new site eventually hosts this course. It seems to have been used to demonstrate Flash and Java applets, which are no longer supported.

pstaabp:
Good point.  @mgage can you hide this assignment in the `wikiExamples` course on demo?

conf/defaults.config
```diff
@@ -355,9 +355,9 @@ $webworkURLs{SiteMap}       = "https://webwork.maa.org/wiki/WeBWorK_Sites";
 #URL for help buttons on the PGProblemEditor pages:
 $webworkURLs{problemTechniquesHelpURL}='https://webwork.maa.org/wiki/Category:Problem_Techniques';
 $webworkURLs{MathObjectsHelpURL}      ='https://webwork.maa.org/wiki/Category:MathObjects';
-$webworkURLs{PODHelpURL}              ='https://demo.webwork.rochester.edu/wwdocs/';
-$webworkURLs{PGLabHelpURL}            ='https://demo.webwork.rochester.edu/webwork2/wikiExamples/MathObjectsLabs2/2/?login_practice_user=true';
-$webworkURLs{PGMLHelpURL}             ='https://courses1.webwork.maa.org/webwork2/cervone_course/PGML/1/?login_practice_user=true';
+$webworkURLs{PODHelpURL}              ='https://webwork.maa.org/pod/';
+$webworkURLs{PGLabHelpURL}            ='https://demo.webwork.rochester.edu/webwork2/wikiExamples/?login_practice_user=true';
+$webworkURLs{PGMLHelpURL}             ='https://demo.webwork.rochester.edu/webwork2/cervone_course/PGML/2';
```
drdrew42, line 360: Any reason to remove the `?login_practice_user=true` from this url?

drdrew42 (approved):

conf/defaults.config
drgrice1: Shouldn't this also still be problem 1 in the problem set?

Also shouldn't the PGLabHelpURL still point to the MathObjectsLabs2/2 set and problem?

pstaabp: Updated these.

PR #1541, opened 2022-01-14
Modifications to PDF hardcopy -- comments + score overrides

drdrew42:
As the title suggests, this PR enables the user to include instructor feedback when generating a hardcopy for any set.

When generating a hardcopy, a "Comments" checkbox has been added (between Student Answers and Correct Answers -- since that is where it appears in the hardcopy). Selecting this checkbox adds the instructor feedback to the PDF.

Additionally, when generating a PDF that includes the students' responses, now provides both the pg-result of the student response as well as the recorded score (in case the pg result was over-written).

And finally, swap out `verbatim` mode for `lstlisting` with student responses. Text does not wrap in verbatim-mode, and over-long student responses (or instructor feedback) will overflow the two-column (and in extreme cases, one-column) layouts. Since the `listings` package is necessary for wrapping comments, I figured I might as well resolve this long-standing student answer overflow issue.

Lingering questions.
* should the recorded score omitted unless it differs from the pg result? 
* should the comments be positioned elsewhere?


lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -1358,26 +1363,58 @@ sub write_problem_tex {
...
 			my $stuAnswers = "\\par{\\small{\\it ".
 			  $r->maketext("Answer(s) submitted:").
 			  "}\n" .
 			"\\vspace{-\\parskip}\\begin{itemize}\n";
 		for my $ansName ( @ans_entry_order ) {
 			my $stuAns = $pg->{answers}->{$ansName}->{original_student_ans} // "";
-			$stuAnswers .= "\\item\\begin{verbatim}$stuAns\\end{verbatim}\n";
+			# alternate itemize bullets disappear without extra newline here...
+			# newline also required after \begin{lstlisting}
+			$stuAnswers .= "\\item\\begin{lstlisting}\n$stuAns\\end{verbatim}\n\n";
```
drgrice1, line 1394: Something is wrong here.  There is a `\begin{lstlisting}` that is ended with a `\end{verbatim}`.  This causes the latex compilation to fail.  Also, you need yet another newline.  This should be:
```suggestion
			$stuAnswers .= "\\item\\begin{lstlisting}\n$stuAns\n\\end{lstlisting}\n\n";
```
With this change this compiles for me.

drgrice1 (commented):
I think that if the student answers are switched to using lstlisting, the correct answers should also be switched to that.  With this the student and correct answers look different.  The lstlisting uses a different font than the verbatim.  Here are a couple of screenshots made from a pdf generated (after making my suggested changes):
Problem1
Problem3

On another note, I would rather remove the verbatim in a different way.  I would rather switch to using `$pg->{answers}{$ansName}{preview_latex_string}` instead of `$pg->{answers}->{$ansName}->{original_student_ans}`, and `$pg->{answers}{$ansName}{correct_ans_latex_string}` instead of `$pg->{answers}->{$ansName}->{correct_ans}` and show the answers in LaTeX.  The answers just look better this way, and this would also fix the issue with certain types of answers for which not all of the answers are displayed, like MultiAnswers with singleResult or Matrices via ans_array.  Of course the answers that have non_tex_preview set would need to be avoided.

I think that comments would be better off in latex as well, although the backtick latex formatting that comments allow would need to be dealt with.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
drdrew42: <facepalm.gif>

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -1358,26 +1363,58 @@ sub write_problem_tex {
...
 			my $stuAnswers = "\\par{\\small{\\it ".
 			  $r->maketext("Answer(s) submitted:").
 			  "}\n" .
 			"\\vspace{-\\parskip}\\begin{itemize}\n";
 		for my $ansName ( @ans_entry_order ) {
 			my $stuAns = $pg->{answers}->{$ansName}->{original_student_ans} // "";
-			$stuAnswers .= "\\item\\begin{verbatim}$stuAns\\end{verbatim}\n";
+			# alternate itemize bullets disappear without extra newline here...
+			# newline also required after \begin{lstlisting}
+			$stuAnswers .= "\\item\\begin{lstlisting}\n$stuAns\\end{lstlisting}\n\n";
```
drgrice1, line 1394: The additional newline after `$stuAns` is still needed.  Either that or a check to see if `$stuAns` is the empty string.  The problem occurs if you try to generate a hardcopy and check to print student answers (it is checked by default), and there are no answers for a problem.  In that case latex gives an error about a missing `\item`.  The addition of the newline fixes this.  Latex is okay with an empty line there.

drdrew42: I eliminated the newline after `$stuAns` because it created a literal newline inside the listing environment and I didn't like the resulting spacing. I think I'd prefer to use a default 'no response' for `$stuAns` in the case that it is empty, rather than leaving it as a blank `\item`.

drgrice1:
That is fine.  I don't really like the bullet with nothing.  So change it that way instead.

pstaabp:
If you're including the `listings` package, we should add that to the `bin/check_latex.tex` file.

drgrice1:
Maybe the `bin/check_latex.tex` file should just include the file `conf/snippets/hardcopyThemes/common/packages.tex` (and perhaps the other files there as well).  That way things stay in sync.

drdrew42:
Okay, so the check_latex.tex file now pulls in packages.tex (there is now redundancy on packages, I'm sure).

I've switched student responses and correct answers over to latex_strings, with correct answers having a fall-back to `correct_ans` in text-mode if `correct_ans_latex_string`  is not present. I've also opted for `\displaystyle` since things get a little cramped without it...

I've checked this PR with vectors, matrices, multianswers, multiple select (checkboxes), and the usual fraction, decimal, and equation mathobjects -- but of course this should be confirmed.

Lingering questions:
- Is there any chance that a student response would not have a `preview_latex_string`?

drgrice1:
If the student enters something that fails to parse, then there will not be a `preview_latex_string`.  Falling back to `$pg->{answers}{$ansName}{original_student_ans}` is probably the best that can be done in this case.  It would be nice to be able to fall back to `$pg->{answers}{$ansName}{student_ans}`, but that would contain html markup in this case.

There is a case in which there will be a `preview_latex_string` that should definitely not be used here.  That is the case that the `$pg->{answers}{$ansName}{non_tex_preview}` flag is set.  Currently the graphtool macro is the only one that uses this flag, and for that macro the `preview_latex_string` will be javascript.

drdrew42:
> There is a case in which there will be a `preview_latex_string` that should definitely not be used here. That is the case that the `$pg->{answers}{$ansName}{non_tex_preview}` flag is set. Currently the graphtool macro is the only one that uses this flag, and for that macro the `preview_latex_string` will be javascript.

What should be used in this situation?

drgrice1:
Also, the `preview_latex_string` could be "0", and that could be a valid answer.  In this case perl will treat that as false, and the usage of `||` will cause "no response" to be displayed.  So that should be `//` instead.

Switching to showing latex answers will take some care.  I should have said more when I suggested this.

drdrew42:
Something like this, then:
```perl
my $stuAns;
if (defined $pg->{answers}{$ansName}{preview_latex_string} && $pg->{answers}{$ansName}{preview_latex_string} ne '') {
	$stuAns = $pg->{answers}{$ansName}{preview_latex_string};
} elsif (defined $pg->{answers}{$ansName}{original_student_ans} && $pg->{answers}{$ansName}{original_student_ans} ne '') {
	$stuAns = "\\text{".$pg->{answers}{$ansName}{original_student_ans}."}";
} else {
	$stuAns = "\\text{no response}";
}
```

I'm pretty sure the `defined` check is not necessary, since the key is certain to exist in the hash...

drgrice1:
Yeah, I think that looks right.  The first conditional should also contain `!$pg->{answers}{$ansName}{non_tex_preview}`.  So
```perl
my $stuAns;
if (!$pg->{answers}{$ansName}{non_tex_preview} && defined $pg->{answers}{$ansName}{preview_latex_string}
	&& $pg->{answers}{$ansName}{preview_latex_string} ne '') {
	$stuAns = $pg->{answers}{$ansName}{preview_latex_string};
} elsif (defined $pg->{answers}{$ansName}{original_student_ans} && $pg->{answers}{$ansName}{original_student_ans} ne '') {
	$stuAns = "\\text{".$pg->{answers}{$ansName}{original_student_ans}."}";
} else {
	$stuAns = "\\text{no response}";
}
```

drdrew42:
So, in the case of graphtool, there will be relevant content in the `original_student_ans`? Or should the logic first branch on `non_tex_preview` alone, in order to provide a message like 'this answer is a graphtool image'?

drgrice1:
For the graphtool that will either be the empty string or it will be filled with something that will always succeed in being parsed into a MathObject list.  Javascript fills that with data from what the student graphs.  In fact the `student_ans` would always work here.  No failed parsing.

Also, the correct answer will need the same check.  The `correct_ans_latex_string` will not be valid latex either (it will be javascript), and so the `correct_ans` should be used for the `non_tex_preview` correct answer.

drgrice1:
If the `original_student_ans` and `correct_ans` are used for the graph tool macro then you will see something like
`{circle,solid,(1,4),(2,2)}` meaning the student graphed a solid circle centered at (1,4) and passing through the point (2,2), or that is the correct answer for `correct_ans` of course.

Or for a more complex answer `{circle, solid, (-4,-3), (1,1)}, {line, solid, (-5,-1), (3,1)}, {parabola, solid, horizontal, (-2,-2), (4,3)}`.

drgrice1:
Scratch everything I said about the special handling for the `non_tex_preview` case.

The correct answer will just work as it is.  The graphtool macro only changes the `correct_ans_latex_string` to javascript when the display mode is not `TeX`.

Currently the student `preview_latex_string` is always changed to javascript, but the only reason that I did that was because that was never used in print.  So we will instead modify the parserGraphTool.pl macro to do the same as for the correct answers.

drgrice1:
I added a pull request to your branch with a suggestion on the `check_latex.tex` thing.  There are some issues with just adding `\input{packages.tex}` that the pull request resolves in one way.  You can do something else if you prefer.

I note this here because some seem to have not noticed pull requests to their branches.  I am not sure if GitHub sends notifications or not.

drdrew42:
> Scratch everything I said about the special handling for the `non_tex_preview` case.
> 
> The correct answer will just work as it is. The graphtool macro only changes the `correct_ans_latex_string` to javascript when the display mode is not `TeX`.
> 
> Currently the student `preview_latex_string` is always changed to javascript, but the only reason that I did that was because that was never used in print. So we will instead modify the parserGraphTool.pl macro to do the same as for the correct answers.

I think I'll still make the change to check the hash-values just for the sake of thoroughness, but it sounds like I can leave out the check for `non_tex_preview`.

Thanks for the update on the check_latex script. I didn't notice any issue when testing it out before, but it's quite possible I was overlooking something, or that system configuration might make a difference here?

drgrice1:
@drdrew42:  I added another pull request to your branch that fixes the merge conflicts.

pstaabp:
Overall this looks good.   I don't use the hardcopy feature very often.   Seems like the comments doesn't need to be lstlisting environment, that is in monospace font.   Is there formatting that this should have for monospace? 

Comment that may have been discussed some long time before now,  but it might be more intuitive to have the hardcopy pdf open in a new tab--that seems to be common when pdf's are clicked on.  I know this is unrelated to this PR, but my two cents...

drdrew42:
At first, the comments were inserted as `verbatim`, but that mode does not allow for text wrapping. This causes major overflow issues with any comments of substantive length and two-column mode. I was not aiming to support LaTeX-in-comments, though that is supported via HTML... instead my aim was to produce an archival copy of student exams (with comments and grade overrides). I don't know if students would use the PDF hardcopy feature frequently enough (with comments) to justify the headache of regex-ing comments for hardcopy LaTeX support.

drgrice1:
Yeah, I think it is easiest to just use the lstlisting package for comments, and I agree that I don't think students use the hardcopy feature much.

@pstaapb:  It actually is not the standard to have a pdf open in a new tab/window.  It may be common on the pages you happen to go to, but that is largely the choice of the page creator as to if the pdf opens in a new tab/window or is offered for download.  I actually think for this purpose, that offering the pdf for download is better.

PR #1543, opened 2022-01-17
Achievement features

somiaj:
Here are some modifications I have made to the Achievements system I use that I thought were worth sharing. This request contains the following:

+ Add a $userAchievements hash to allow achievement evaluators to use the status of other achievements.
   Included an example of earn N achievements achievement.

+ Only display current achievement in achievement chains (do 10,20,... of something) and hide future achievements
   until the previous ones have been earned.

+ Show achievement points and display level progress as a delta starting from the previous level instead of 0.
   I use a course.conf variable for this as I found that easier than looping through the level achievements each time.

+ Add some achievement images for others to use.


lib/WeBWorK/ContentGenerator/Achievements.pm
```diff
@@ -284,6 +292,10 @@ sub body {
 	
 		@achievements = sortAchievements(@achievements);
 		my $previousCategory = $achievements[0]->category;
+		my $previousNumber = $achievements[0]->number;
+		(my $chainName = $achievements[0]->achievement_id) =~ s/^([^_]*)_.*$/$1/;
```
drgrice1, line 296: It is better to use the `r` substitution flag than declaring the variable inside parenthesis.  The `r` flag means don't modify the input and return the result after performing the substitution.  So change this to
```suggestion
		my $chainName = $achievements[0]->achievement_id =~ s/^([^_]*)_.*$/$1/r;
```

lib/WeBWorK/ContentGenerator/Achievements.pm
```diff
@@ -296,13 +308,32 @@ sub body {
...
+
+			#setup up chain achievements
+			my $isChain = 1;
+			if ($previousCategory ne $achievement->category ||
+				$previousNumber + 1 != $achievement->number ||
+				$achievement->achievement_id !~ /^$chainName/ )
+			{
+				$isChain = 0;
+				$chainCount = 0;
+				($chainName = $achievement->achievement_id) =~ s/^([^_]*)_.*$/$1/;
```
drgrice1, line 320: Same thing here.

lib/WeBWorK/ContentGenerator/Achievements.pm
somiaj: Thanks, updated my commit. I also fixed a achievement numbering issue in the default achievements so this works.

pstaabp:
This will need to be updated to fix the conflicts.  Merge in openwebwork/develop.

somiaj:
@pstaabp Thanks for pointing this out. Updated.

pstaabp:
Overall, this seems good. I don't usually use achievements, so am having a bit of trouble evaluating this. The new images look good and the ideas seem solid.  Anything in particular to check?

somiaj:
It is working as described in my current class fwiw. I can't think of anything in particular that needs checking outside of loading and assigning all the default achievements, earn a few by doing some problems then go take a look and see if anything sticks out.

drgrice1:
One thing that I don't like here is the addition to courses.distt/modelCourse/course.conf.  I must have missed the related addition to that file in #1316, or decided to let that slide for that last minute addition to the 2.16 release.  These things really shouldn't be here.  The settings should be added to the course settings page, probably in the "Optional Modules" part (or perhaps achievements need their own configuration tab).  The reason that adding these things to course.conf is bad is because generally instructors do not have access to edit that file (or so the File Manager would have you believe).  Even though technically there is a back door through which instructors can edit that file, it is just simply that editing a configuration file isn't a way that settings like these should be handled.

Alex-Jordan:
Is the course.conf file in courses.dist/modelCourse/course.conf even used? (I mean other than directly for that course?) I suppose it might be inspected as something to model after.

If a new course is generated, its course.conf file is written by [CourseManagement.pm](https://github.com/openwebwork/webwork2/blob/main/lib/WeBWorK/Utils/CourseManagement.pm), which is not affected by this PR.

It would make sense to have these set in `defaults.config` and then allow Course Config to overwrite them in `simple.conf`. Since many people do not use achievements, having them in their own tab that only appears after achievements are enabled would be nice.

drgrice1:
I don't think that courses.dist/modelCourse/course.conf is actually used.

I agree that an achievements tab in config should only be visible if achievements are enabled.

drgrice1:
The fact that courses.dist/modelCourse/course.conf is not used is part of my point though.  These features are not being created in a way that is transparent to the end user in any way.  The settings are stowed away in an obscure location that no instructor sees, unless they happen to also be a system administrator and are delving into the files in course.conf.

somiaj:
I would be interested in a better solution for this particular case (outside of updates to the configuration system). This particular variable is configured elsewhere in the actual achievements. I just didn't think it was worth looping through all the achievements each and every time just to get mostly static values stored inside of them.

The other option I had in mind was to add another column to the global_user_achievements table, which would not only store next_level_points but also store prev_level_points. I just didn't want to modify the database with this, but in hind site that is probably the better solution here?

Alex-Jordan:
In a PR, at this line: https://github.com/openwebwork/webwork2/pull/1505/files#diff-5b1cffd14071ff573baa4f7ca41ec48810fcbbdb731fbeafef8ee7ebdc2703beL383

I put in some commented out code that could be uncommented to add an LTI tab to the Course Config page. This is reasonable, since if a site is going to use LTI, they will be setting up the details of their authen_LTI.conf file.

If there is a reasonable collection of Achievements-related configuration settings, the distribution could include a config file that is specific to Achievement settings. Among other things it could enable achievements globally, and `include()`-ing this config file at some level could become the way to enable Achievements, like is done with LTI.

And then this config file could include similar code to what I linked to, that adds a tab for Achievement specific settings. Except I think there would be no need to comment it out. (The LTI tab is commented out because it's reasonable to have LTI enabled, but you don't want to encourage instructors to mess with the settings.)

I'm not necessarily recommending this. I'm not sure it's a good idea for every optional module to come with its own config file. But it's at least an option to consider.

somiaj:
@Alex-Jordan the only issue with having it enabled by default would be it may disagree with how a particular course has changed their level achievements, so a course may need to override this to fit their desired levels anyways. I left it commented out because in that state it reverts to the current behavior of achievements.

In general it is probably best to rethink how to store this data. Maybe if a configuration tab for achievements was enabled have it set there. Though first the numbers could be moved out of the achievements themselves and put into a central configuration variable. Or just add another column to the database to store the data I am storing via this variable (it is how levels are done right now anyways).

Quite frankly, I'm unsure the best/preferred method to do this.

drgrice1:
@somiaj:  Since the file courses.dist/modelCourse/course.conf is not used, it really doesn't matter if it is commented out or not.  Note that the settings available on the course options page are not stored in the database.  They are stored in the course's simple.conf file.  That file is loaded after the course's course.conf file, and override anything in that file if set.  So adding a tab on the options page that appears when achievements are enabled would be the equivalent to adding to course.conf, but better.  The settings are only saved to simple.conf if they change the defaults.  The default values should be set in defaults.config.  Those default values should be set to how things currently are, and the code adapted to use those if these are replacing currently hardcoded values.

somiaj:
The levels are not hard coded, but instead are set using the level achievement files when they are evaluated. That data is then stored in the global user achievement table. Since other data is already stored there, I was thinking this could be a column along side the other data, but I overlooked another option.

I have updated the patch to store the data I need, prev_level_points, inside the global user achievement frozen_hash, which is meant to store data related to achievements. I initially overlooked using this. The recent update no longer adds or use anything in course.conf.

Alex-Jordan:
I'm looking over the edits, possibly ahead of doing some testing this coming weekend. I'm curious why the numbering for challenge achievements is changing from 601, 602, 603, etc to 601, 603, 605, etc.

somiaj:
@Alex-Jordan a side effect of hiding the chain achievements patch, 'earn X of something'. Since all the challenge achievements start with the same string `challenge_` and are numbered in order, they get flagged by the logic that determines if this is an achievement chain, thus they will be hidden from view. But in this case they are individual achievements, not a chain. Changing the numbers makes the logic not flag them.

somiaj:
@Alex-Jordan I just thought of a cleaner way to do this. I also check if achievements have a `max_counter` to determine if they are a chain. This way the challenge achievements do not need to be renumbered to avoid the chain logic. I have also rebased this branch against develop so it is more up to date.

drgrice1 (commented):
As far as I can tell this looks good.   See my comment on the one change that is needed.

drgrice1:
There is another possible issue that I see here.  You are adding to the data stored in the frozen hash.  That hash is already causing problems because the data being stored in it can very easily become to large and exceed the allowed database size (`VARCHAR(1024)`).  This can be reproduced for example by using the `still_not_right` achievement which stores the students last answer in that hash, and having problems that store lots of data in the last answer column of the user problem (like Geogebra applet problems which store the applet state there).  Of course this only exacerbates this issue, and is not a new issue.  Perhaps it is time with the next release to change the type of the frozen hash column to something that will allow more data without causing errors.

drgrice1:
Geogebra applets do not store the data in the frozen hash.  They store the data in the last answer column of the user problem.  The `still_not_right` achievement copies that into the frozen hash.  Yes, that would be for the user achievement, and not the global user achievement.

Don't deal with the size issue of that hash in this pull request.  We should talk about it in our meetings, and then we can deal with that in another pull request.

lib/WeBWorK/AchievementEvaluator.pm
```diff
@@ -138,6 +139,14 @@ sub checkForAchievements {
 		$globalData = thaw_base64($globalUserAchievement->frozen_hash);
     }
 
+    #Generate hash of user achievements:
+    foreach my $achievement (@achievements) {
+        next unless $achievement->enabled;
+        next unless ($db->existsUserAchievement($user_id,$achievement->achievement_id));
+        my $userAchievement = $db->getUserAchievement($user_id,$achievement->achievement_id);
+        $userAchievements->{$achievement->achievement_id} = $userAchievement->earned;
```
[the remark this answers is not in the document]
somiaj: Thanks.

pstaabp:
@somiaj I was going to look at this again, but we have some conflicts now with this.  Can you fix these?

somiaj:
@pstaabp the commits are now rebased on the current develop branch.

pstaabp:
I'm getting missing icons for all the images inside the achievement editor.  This is because they are all in the modelCourse and I would need to recopy everything over for an existing course.  This would also be a problem for new courses that have been generated from old courses.  I see the only way to handle this would be to have an admin copy all new image files over to each needed courses.  

I wonder if we should move all of the images from individual courses to `webwork2/htdocs/achievements` instead?  And maybe have a mechanism for allowing individual courses to pull custom achievements from the `courses/course_name/templates` directory.

drdrew42:
> I wonder if we should move all of the images from individual courses to `webwork2/htdocs/achievements` instead?

So long as WeBWorK checks the `course_name/htdocs/` location first, this would be okay... we have courses that use different 'themed' badges for certain achievements (e.g. reaching level two in precalculus is visually different than reaching level two in calculus 1) and we would want to override the consolidated 'default' image.

somiaj:
> I'm getting missing icons for all the images inside the achievement editor.

@pstaabp What about the achievements themselves? Did you just copy the achievements over but not the images?

Anyways, to me I don't see how you would have the new achievements that request the new images without copying those over from the modelCorse/templates directory anyways, so seems to me that admins just need to know to copy over both the achievements and the images. Though I'm not against having a mechanism look in `webwork2/htdocs/achievements` if the image is not found in the course directory, I think this issue is just inherit in the fact that all achievement info is in the course directory.

pstaabp:
> What about the achievements themselves? Did you just copy the achievements over but not the images?

None of the icons are working for me and I guess I'm not seeing the new achievement either.  I'm not sure how I ended up with any achievements.  

In that light, let's keep everything in the modelCourse and have admins copy everything over as needed.

somiaj:
I rebased this off develop and added another commit that fixes an accessibility issue that the h2 achievement header was not printed if achievement items were disabled causing the headers to skip from h1 to h3.

Another possible issue I noticed, but can't think of the best way to fix it, is that the h1 header and then following h2 header are both called 'Achievements'. This might not be a problem, but having different headers could help.

pstaabp:
> Another possible issue I noticed, but can't think of the best way to fix it, is that the h1 header and then following h2 header are both called 'Achievements'. 

We have discussed for more than a year to rename the achievements, especially to distinguish between a badge (yeah I did it) and a reward (yeah I get something for this!!).  I think we all agreed and just haven't gotten around to it.  Renaming may fix this problem.

somiaj:
@pstaabp I just ran across that issue today and was thinking to adopt the names here. Besides just updating the headers (and as mentioned the docs), is there much else that needs to be changed? This is something where we could start by just calling them rewards and badges for students without much issue I think.

pstaabp:
> This is something where we could start by just calling them rewards and badges for students without much issue I think.

@somiaj I think that would be great.  

We could change the names internally to be consistent with badges/rewards, but I'm not sure that's necessary as long we document/comment correctly.

somiaj:
@pstaabp I agree that it probably isn't worth changing the internal code. I have updated the headers on the student page to just be 'Rewards' and 'Badges', also updated a few things to say stuff like 'Use Reward', 'You don\'t have any rewards!', etc. I even updated the displayed name on the course configuration page to say 'Enable Achievement Rewards'. Outside of the documentation I don't see anywhere else that this needs to updated.

drgrice1 (approved):
This looks good.  I think it is ready to go.

drgrice1:
If it is okay with others I will merge this.  It will conflict with at least one of my pull requests, so I would rather merge it before those so @somiaj doesn't need to deal with the merge conflicts if mine are merged first.

drdrew42:
I'm good with merging this. I have not tested it, but I have read through the changes. So glad that the progress bar will now accurately reflect the actual span from level to level.

drgrice1:
Okay.  I will merge this.  If there are issues missed, we will fix them.

Issue #1544, opened 2022-01-17
webwork2_wsdl is broken because of POD missing


jutrembBDEB:
We have a lot of teachers in Quebec who use the plugins wwassigment/wwlink for Moodle, and since the new version 2.16, the SOAP connection is broken.

I found the reason is because of the push [#1306](https://github.com/openwebwork/webwork2/pull/1306)

All the lines with POD wsdl have been deleted in WebworkSOAP and in all the files generating the XML file (wsdl). So the webwork2_wsdl is incomplete and the connection with Moodle doesn’t work. 

In my server, I put back all those lines that were deleted in the SOAP files and it resolved the problem. But, maybe you erased those lines for a reason and you have a better solution. 

I just know that it need to be resolved quickly because a lot of institutions use Moodle with WeBWorK where I’m from. 

Thank you!

drgrice1:
That doesn't make any sense.  The lines that were deleted were perl POD (plain old documentation).  Those will have no effect at all on the exececution of the code. In that file.  Furthermore, it was invalid POD, and as such caused POD generation on the file to emit errors.

In addition you should note that the WeBWorK SOAP endpoint is a severe security vulnerability, and should not actually be used.  It is possible for a student to use that endpoint to change all of their homework grades to perfect scores without any need to even authenticate.  More disastrous things can also be accomplished with the SOAP endpoint.  For example, deleting all users in a course.

drgrice1:
Furthermore, recent versions of Moodle support LTI authentication, and WeBWorK provides this functionality.  That is secure.

drgrice1:
I see.  It uses the Pod::WSDL library.  I will add that back.  It seems that the POD can be fixed so that it does not generate errors.  The problem with that was the invalid POD markup that was used.

You should note that I figured this out because the exploit scripts were no longer working on my test server.  I would never enable WeBWorK SOAP on my production server due to the reasons that I mentioned.  The exploit scripts that I have can do the things I mentioned before.  You should at least make sure that if this is enabled on your server that you change line 37 of /opt/webwork/webwork2/lib/WebworkSOAP.pm to something that is more secure and at the very least not what is publicly visible in the openwebwork/webwork2 github repository.  And don't share that with anyone.

jutrembBDEB:
> Furthermore, recent versions of Moodle support LTI authentication, and WeBWorK provides this functionality. That is secure.

[...]

PR #1545, opened 2022-01-17
Fix the WebworkSOAP endpoint.

drgrice1:
The POD is used in an invalid way by the Pod::WSDL library to make this work.  That library has bad idea written all over it.

I was at least able to fix the POD so that it does not emit errors when POD generation occurs.

Note that if the WebworkSOAP endpoint is used that the hardcoded soap_authen_key on line 37 of lib/WebworkSOAP.pm should be changed to something more secure.  That should probably be something read from a configuration file instead of hardcoded there.

This fixes issue #1544.

### DIFF
--- a/conf/webwork.apache2.4-config.dist
+++ b/conf/webwork.apache2.4-config.dist
@@ -195,6 +195,11 @@ if ($webwork_url eq "") {
 	$Location{$webwork_htdocs_url} = { SetHandler => "none" };
 }
 
+# Uncomment the line below, and the WebworkSOAP and WSDL handlers sections below
+# that to use the WebworkSOAP handlers.  Make sure that you change the value to
+# something more secure.
+#$WeBWorK::SeedCE{soap_authen_key} = "123456789123456789";
+
 </Perl>
 
 ####################################################################
--- a/lib/WebworkSOAP.pm
+++ b/lib/WebworkSOAP.pm
@@ -37,7 +37,6 @@ our %SeedCE;
 $WebworkSOAP::SeedCE{soap_authen_key} = "123456789123456789";
 #$WebworkSOAP::SeedCE{webwork_dir} = $ENV{WEBWORK_ROOT}|| warn "\$ENV{WEBWORK_ROOT} is undefined -- check your httpd configuration. Error caught ";
 
-
 sub new {
     my($self,$authenKey,$courseName) = @_;
     $self = {};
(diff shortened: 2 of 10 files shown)


drgrice1:
This should be reviewed and merged relatively quickly, and I think this is a valid hot fix.

drgrice1:
Attached is a script that can be used to test this.  It requires python 3 and the python 3 suds library to run.  You will need to change the server `url` in the script appropriately.
[list_users.py.txt](https://github.com/openwebwork/webwork2/files/7885234/list_users.py.txt)
This script will list all users in all courses on the server.

drgrice1:
Also you will need to uncomment the `WebworkSOAP handlers` and `WebworkSOAP WSDL HANDLER` sections in `conf/webwork.apache2.4-config` and restart the apache2 server.

drgrice1:
I also removed the hard coded `soap_authen_key` from the lib/WebworkSOAP file, and set that instead in webwork.apache2.4-config.dist.  So make sure to make the appropriate changes to your copy of that file to test this.

taniwallach:
I'm sorry that I'm not free to help with testing at present.
If changes are being made to how a site enables this interface, it probably makes sense to put some explanation of the necessary changes in whatever Wiki pages explains how to use this.

drgrice1:
@taniwallach:  It seems that would be here: https://webwork.maa.org/wiki/Moodle_Assignment

Very few people use this, and it shouldn't be used anymore.  LTI authentication is far more secure.  This is already documented for Moodle.  See https://webwork.maa.org/wiki/LTI-Advanced_Authentication#Moodle.

pstaabp:
I'm not able to test, but seems like if this works for @jutrembBDEB maybe we should push through.  As I can tell, it doesn't change functionality, just whether or not the POD is generated correctly.

drgrice1:
It does actually change functionality.  The Pod::WSDL module uses the content of the `=begin WSDL` ... `=end WSDL` POD to make its methods available.  When I removed that it broke the SOAP/WSDL endpoints.  This makes it work again.

drgrice1:
I guess since the one known instance of this being an issue @jutrembBDEB already knows about this, and has it fixed locally, this would not need to be a hotfix.  If there are others we can direct them to this pull request.

What do you think?

drgrice1:
Not hotfixing.

PR #1546, opened 2022-01-18
Fix the WebworkSOAP endpoint - develop.

drgrice1:
The POD is used in an invalid way by the Pod::WSDL library to make this work.  That library has bad idea written all over it.

I was at least able to fix the POD so that it does not emit errors when POD generation occurs.

Note that if the WebworkSOAP endpoint is used that the hardcoded soap_authen_key on line 37 of lib/WebworkSOAP.pm should be changed to something more secure.  That should probably be something read from a configuration file instead of hardcoded there.

This fixes issue #1544.


drgrice1:
This pull request has been rebuilt with a force push.  Somehow this was built on top of another pull request by mistake.

PR #1548, opened 2022-01-19
Replace the jquery-ui-timepicker-addon with flatpickr.

drgrice1:
The flatpickr library was chosen because it is somewhat similar to the jquery-ui-timepicker-addon that we were using, and its usage is simple.  Instead of sliders for hours and minutes it uses numerical inputs.

Also it is an actively maintained library.  Its homepage is https://flatpickr.js.org/ and its source located at https://github.com/flatpickr/flatpickr.

I chose the dark theme that it provides, but there are other themes that could be chosen.

Note that the library also provides localization support, although this pull request does not implement that.

Make sure to run `npm install` in `htdocs` to install the flatpickr package to test this.


pstaabp:
Overall, I think it looks good.  I think the light theme may go better with the rest of the UI.

pstaabp (approved):
Looks good.  Also, just tested on a phone and the picker looks a bit different but easier to use with fingers.

drgrice1:
Yes, this is definitely easier to use on touch screens than the jquery-ui datetime picker was.

drgrice1:
Darn, I forgot that the knowl rewrite makes the javscript `defer`ed so that it will also work with webwork3.  So there will still need to be a paired webwork2 pull request.

drdrew42 (approved):
This looks great -- good riddance to time sliders.

PR #1549, opened 2022-01-19
Replace the jquery ui nestedSortable with sortablejs

drgrice1:
This removes all jquery and jquery-ui usage in htdocs/js/apps/ProblemSetDetail/problemsetdetail.js.

The sortablejs is an actively maintained library that seems to generally work well.  See https://github.com/SortableJS/Sortable.  Although finding a good drag and drop library is extremely hard.  It turns out there really aren't any, particularly when it comes to nesting and fully cross platform support.  They all take work to use effectively.  (Even the vue-draggable-next library is not very good -- @pstaabp this will make things difficult for ww3).

This will need testing on various browsers.  I have tested this on Firefox and Google Chrome on Ubuntu 20.04, and Google Chrome on Android.  I do not have access to Safari on any device, so that is one case that particularly needs testing.  @pstaabp  or @drdrew42: I believe you both have Safari, so if you could test this and help diagnose issues that would be appreciated.

With PR #1548 this removes all jquery-ui usage directly in WeBWorK.  Most of the macros that were using jquery-ui no longer are.  In fact I believe that only @mgage's pg/macros/quickMatrixEntry.pl macro still uses it. So we are very close to not needing jquery-ui anymore.  We may still need to make it available for problems for a while in case it is still used by locally written problems though.


drgrice1:
Of course, make sure to run `npm install` to install the sortablejs library.  In fact get in the habit of doing this anytime you change branches or are testing a pull request.

pstaabp:
I'm getting problem reordering working on this on safari and chrome except for JITAR problem sets on safari where I am trying to change the level of a problem (to and from a child).  It's also a bit clunky on chrome.  This might be the package however.  I just looked at examples and their nestable ones seem to be visually different. 

I wonder if we need some different UI to make an element a parent (like some extra empty space at the bottom) that has a drag target to make something a child.

drgrice1:
On Safari can you pick anything up, or does that not even work?

To nest you need to drag up to right from below.  Note that I have it set up not to nest unless you drag far enough to the right (the position of the mouse cursor needs to be to the right of the move icon of the item you are dragging over).

It will take some getting used to as to how the dragging works, and I think that is what the "clunkiness" that you are talking about really is.  To be honest, the jquery ui nested sortable was also clunky, and if you were not used to that you actually might not find this as clunky.  Can you be more specific as what you mean by clunky?

Also what "package" are you talking about?  Do you mean the sortablejs javascript library that is used?  Note that the way it works it really assumes that to nest a sub list already exists.  We don't have that in the markup we generate so I did some work to make it so that a sub list is dynamically created during the drag.

When there is no nesting, I see nothing clunky at all for either Firefox or Chrome, and things work way better than they did with the jquery ui nested sortable.  Scrolling is sometimes not the best, but scrolling was quite bad with the jquery ui nested sortable also.

pstaabp:
Both browsers work correctly.  It seems like as soon as you describe the *drag up to right from below*, that showed the dark gray bar indicating the indent/nesting level.  

The clunkiness I would describe if that the dark gray bar doesn't appear, then it is hard to determine what happens.  Sometimes the indent worked, other times not.  

You can ignore the rest of my comment now that I have figured it out.  In fact, I can't reproduce what I was seeing earlier when the dark gray bar wouldn't appear.

drgrice1:
One thing that seems odd to me in your comment is the reference to this "dark gray bar".  With this pull request there should be no dark gray bar.  That is what things looked like with the jquery-ui nested sortable.  With this pull request you should see a translucent duplicate of the object being dragged.

pstaabp:
So I must have been running the old code.  I've done npm install, restarted apache, cleared the cache.  No more dark gray bar.

> With this pull request you should see a translucent duplicate of the object being dragged.

I do see that now.  The behavior in Safari and Chrome seems to be similar.  I do find it difficult to switch a JITAR set problem to change levels (become a parent or child).  If I drag around enough it does switch, but not obvious how to get it to change.  Reordering within a level works fine. 

The other issue I'm seeing is that if I drag around and the mouse ends on an edit or view button, that is activated.   If needed, I can make a screencast to post what I'm seeing.

drgrice1:
Here is a quick video of what I see with Google Chrome.  What I see works pretty well.  If you can post what you see that might help.

I am not sure what you are talking about with the edit or view button.  What do you mean by an activated edit or view button also?

https://user-images.githubusercontent.com/47527406/150409080-b7d8cc70-9109-43e3-ade4-7b4d4edc8859.mp4

drgrice1:
The video gets cut of at the end because I thought it was to big.  Then I had to re-encode it to get it to work here and it got smaller.  So I didn't need to after all.

drgrice1:
At the end I was pointing out that with the jquery-ui nested sortable I observed that if you drop on an edit or view icon the tooltip does not close even after you move the mouse away.  It gets stuck open, and won't go away.  So this is better for that.

pstaabp (approved):
This is working fairly well now.

drdrew42 (approved):
Tested on Chrome and Safari on MacOS and iOS -- reordering problems is easier and looks better than before. 👍

PR #1551, opened 2022-01-20
Change the way that coloring of inputs is done for correct and incorrect answers.

drgrice1:
Instead of creating a global javascript variable, use data attributes.  This pull request makes the changes to the AttemptsTable and loads the javascript defered.

This is a pull request is paired with the pg pulled request openwebwork/pg#640.  That pull request makes the changes to the javascript.


pstaabp:
I went to a quiz that was submitted and over time.  I clicked "Check Test" and only the first problem (out of 3) colored the input.  The other two problems had no green/red around the inputs.  This was the same using both mathquill and with it turned off.

drgrice1:
@pstaabp: Did you check out the corresponding pg pull request also?  Both are needed for this to work.  Of course this is a javscript change, so browser caching could also be an issue.

drdrew42 (approved):

PR #1552, opened 2022-01-20
Fix the instructor help for the problem sets page.

drgrice1:
The override method was not working, and not needed.

This is easy to test.  Without this pull request clicking on the `?` link in the site nav on the Homework Sets page opens to "There is no help ...".  With this pull request clicking on the `?` link opens to "Course Home Help Page ...".

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSets.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSets.pm
@@ -119,14 +119,7 @@ sub info {
 		return "";
 	}
 }
-sub help {   # non-standard help, since the file path includes the course name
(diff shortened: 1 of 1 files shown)


pstaabp:
Just to be clear, you are talking about the ? icon not the "Help" link. 
image

correct?

drgrice1:
Yes, I meant the `?` icon.

pstaabp (approved):
Works as advertised.

drdrew42 (approved):
> A fun case where simply deleting code makes things work right!

Whoa, oh, it's ma-a-gic ... you know-ow-ow

PR #1553, opened 2022-01-20
htdocs clean up

drgrice1:
Remove the unused files in htdocs/js/vendor, htdocs/css, htdocs/images, and all files in htdocs/applets, htdocs/html_templates, and htdocs/codemirror2.

The file htdocs/js/vendor/jquery/modules/jstorage.js was actually used by htdocs/js/apps/LocalStorage/localstorage.js, but that file has been updated to use native javascript, and so no longer uses it.

The files htdocs/css/jquery-ui-timepicker-addon.css and htdocs/js/apps/DatePicker/jquery-ui-timepicker-addon.js have been removed, and that is now installed via npm.  This will conflict with another pull request of mine, but it is easy enough to resolve.

Note that several of the other files are used by pg/macros/LiveGraphics3d.pl, but that macro heavily depends on flash and java, and as such will be deleted.  The javascript for that macro is also deleted here.

Also remove the unused, unmaintained, and unlinked to files
lib/WeBWorK/ContentGenerator/Instructor/Compare.pm,
lib/WeBWorK/ContentGenerator/Instructor/GetTargetSetProblems.pm,
lib/WeBWorK/ContentGenerator/Instructor/GetLibrarySetProblems.pm,
lib/WeBWorK/ContentGenerator/Instructor/Preflight.pm and
their corresponding paths in lib/WeBWorK/URLPath.pm.

If anyone sees a reason to keep a file that was deleted here, make sure to speak up.


pstaabp:
An initial pull of this and going through all of the pages seems to not generate any errors--which I think is the big point of this.

drdrew42 (approved):
🥳 

Should this wait until after we merge the flatpicker? or does it matter?

drgrice1:
That shouldn't matter.  I can fix the merge conflict.

drgrice1:
Either way, one of the pull requests will have a merge conflict anyway.

Issue #1511, opened 2021-10-12
wwsh script - '.' is no longer in @INC


chrisdukey:
With perl 5 version 30 the `/opt/webwork/webwork2/bin/wwsh` script throws an error when you try and add an admin user (e.g. `./wwsh myCourse addadmin`). The error is `do "addadmin" failed, '.' is no longer in @INC;`

This can be fixed by adding `use lib ".";` into the wwsh script

pstaabp:
@chrisdukey Which webwork version are you running?

I'm trying this on the develop branch and running perl 5.30 and don't see this, but some changes have been made with perl libraries.

chrisdukey:
@pstaabp We're on version 2.16

drdrew42:
Is it any different if you use `wwsh myCourse ./addadmin`? i.e. provide a path to the script file?

pstaabp:
I see another issue.   Running as `wwsh` or `./wwsh` runs perl as `/usr/bin/perl` (first line of the file.).  That's not the perl that I use (I'm using perlbrew).  The top line should be `#!/usr/bin/env perl`.  Once I change it, I see @chrisdukey error.  

I generally run scripts starting with perl, like `perl bin/wwsh myCourse addadmin`, 

And yes, `use lib ".";` fixes it.   I think adding the proper top line and this line should fix this.

drgrice1:
@pstaabp:  Looking at the git history for this file, I see why `#!/usr/bin/perl` was used.  It used to be `#!/usr/bin/perl -d`.  Flags don't work with `#!/usr/bin/env perl`.  Of course, now that the flag has been removed it could be changed.  @mgage removed the flag in 2015.

drgrice1:
@pstaabp:  Also, have you tried running the script as @drdrew42 suggested with the relative (or absolute) path to the `addadmin` file?  So for your usage try `perl bin/wwsh myCourse bin/addadmin` without adding `use lib "."`, and see if that works.

drgrice1:
If I run `perl bin/wwsh myCourse bin/addadmin` it fails with the error that @chrisdukey shows.  But if I run `perl bin/wwsh myCourse ./bin/addadmin` it works.

pstaabp:
I can get it to run and I can get it to fail depending on how it is called and in what directory.

If we want it to run as an executable I would say we should call the user's version of perl (not necessarily `/usr/bin/perl`). 

[...]

PR #1554, opened 2022-01-20
fix for wwsh

pstaabp:
This fixes issue #1511 in that recent perl versions (at least 30) would not run a script using `wwsh`. 

Note: this should now run in the following forms:
```
wwsh myCourse addAdmin      # if WEBWORK_HOME is in PATH
./wwsh myCourse addAdmin    # inside bin directory
perl wwsh myCourse addAdmin  # inside bin directory
perl bin/wwsh myCourse addAdmin # inside $WEBWORK_HOME
```

### DIFF
--- a/bin/wwsh
+++ b/bin/wwsh
@@ -30,10 +30,6 @@ BEGIN {
 	die "WEBWORK_ROOT not found in environment.\n" unless exists $ENV{WEBWORK_ROOT};
 	$pg_dir = $ENV{PG_ROOT} // "$ENV{WEBWORK_ROOT}/../pg";
 	die "The pg directory must be defined in PG_ROOT" unless (-e $pg_dir);
-
-	use File::Basename qw/dirname/;
-	use Cwd qw/abs_path/;
-	$bin_dir = abs_path(dirname(__FILE__));
 }
 
 our $ce;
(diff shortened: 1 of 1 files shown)


drgrice1 (changes requested):
This won't work.  The only thing here that is okay is the change of the first line of the file as discussed.  Your change to line 71 enforces that the script must be in the bin directory.  That is wrong.  Read my comment https://github.com/openwebwork/webwork2/issues/1511#issuecomment-1016953500.  I mentioned the way that this needs to be done.

pstaabp:
> Your change to line 71 enforces that the script must be in the bin directory.

My thought is that all the scripts that are run using `wwsh` are in that directory.  I guess thinking a bit more, a user can write their own script which shouldn't be in that directory.

drgrice1:
Yes, a script could be anywhere.  That is my point.

It sort of feels like you are thinking of the argument passed to the script as a command rather than a file.  A file is something that should have a relative or absolute file.   The usage of a command with a script would be more like what you are thinking with the command `wwsh myCourse addadmin`.   However, for a file (in this case a script file), the path should be given to tell the `wwsh` script where that file is.  So that means it should be `wwsh myCourse /path/to/addadmin`.

So basically, what I am trying to say is that this whole idea of making it so that you don't need to give the path is inherently incorrect.  I understand the desire to "make things easier", but that is just misguided here.

drdrew42:
I have a bunch of wwsh scripts that are *not* in the webwork2/bin directory. They're custom for my institution and are not intended to be shared or monitored by git.

pstaabp:
It seems like I'll just update the first line, but maybe build in some documentation on how to use since it isn't exactly clear how to use the script.  Should I use the `Getopt::Long` package like we've been doing, or just some custom help stuff?

drdrew42 (approved):

drgrice1 (approved):
I also approve!

drgrice1:
I am going to merge this.  It has two approvals, and isn't that extensive.

PR #1558, opened 2022-01-22
Add the capability for students to add attachments to feedback emails.

drgrice1:
This switches to using the perl Email::Stuffer module, instead of directly using the Email::Simple module.  The Email::Stuffer is much easier to work with for adding attachments.  It uses Email::Simple and Email::MIME under the hood.

This also adds the option `$mail{maxAttachmentSize}` in site.conf which is the maximum allowed attachment file size in megabytes.  The default value for this is 10 megabytes.

The allowed file types are images with the extension jpg, jpeg, png, or gif, pdf files, zip files, txt files, and csv files.  This is hardcoded for now.  Should videos be allowed?


drdrew42:
I needed to update `Net::SMTP` after encountering the warning: 
`Argument "3.08_01" isn't numeric in numeric lt (<) at /perl/path/Email/Sender/Transport/SMTP.pm line 297.`

Do we have anything like a package.lock for perl modules? This will be something that is potentially encountered during an upgrade.

drdrew42 (approved):

drgrice1:
Yeah, a `package.lock` would be useful.  On Ubuntu I always try to work with the perl modules in the Ubuntu repositories.  That covers probably most servers, since that is probably the most commonly used distribution for WeBWorK.

Also, I don't think that error is caused by this pull request though.  I seem to recall that happening for the version of Net::SMTP on Ubuntu 18.04 after some change in WeBWorK 2.16.

Alex-Jordan:
Is it in some release notes somewhere to check perl modules?  In particular I needed to add Email::Stuffer.

Alex-Jordan:
Yes, I mean for the next release. Taht place you linked is good. I just mean for a place to jot down something that should be mentioned before we collectively forget about it.

drgrice1:
The value should never be set to 0.  It was not intended to make adding attachments optional.  Just the max attachment size limitation, and that should be in site.conf.  It is a server configuration like the other email settings, and belongs together with those.  It is not something that should be done per course as it is really to be set to match the limitations of the email server used, and not some willy-nilly decision of a course instructor.

Alex-Jordan:
OK, sounds good.

Alex-Jordan (approved):

drgrice1:
I will add more comments in site.conf about the value of that setting.

drgrice1:
I added a bit more in the comments in site.conf.  If it is desired that allowing attachments be optional, then that can be implemented.  It was just not in my design.  What do you think?

Alex-Jordan:
I don't have an opinion about making it optional. I would defer to people who know more about mail service. I can't think of a legitimate reason to turn it off at the course level. So at the site level, is there any reason an admin would not want this feature? It sounds like no.

Alex-Jordan:
I'm testing a lot after putting 2.17 on a new Oracle server and transferring our production database and courses over. This feature is very nice. While testing, I initiated a message and chose a file to attach. I don't see a way to change my mind and "detach" the file without exiting the message. Is that expected?

drgrice1:
I wouldn't say expected.  It perhaps would be nice to be able to do this, but isn't implemented.  This is the way that the html file input works.  Notice that the same behavior is exhibited on the instructor file manager page with the file upload.  I am sure it is possible to implement a "clear" approach.

PR #1559, opened 2022-01-23
Remove the last (unused) uses of CGI_labeled_input and the method itself

drgrice1:
Most of the uses of that method were removed in my bootstrap 5 update.  In fact all of them that were actually still in use.  The remaining occurrences are in obsolete code.

Furthermore, that method is unwieldy and not really useful.  It is actually harder to use that method than to use the CGI methods directly.  Also, it is not versatile enough to handle modern html mark up.


pstaabp (approved):
It looks like the two main functions that you removed `duplicate_form` and `duplicate_handler` were no longer being used.  The duplicate set still works and everything else seems fine.

drdrew42 (approved):
🧹

PR #1560, opened 2022-01-24
Remove the gateway confirmation dialog, and replace it with a gateway confirmation page

drgrice1:
This is really part of the GateQuiz.pm module, and so a student entering the quiz via the link (for example via LTI authentication) will be shown this page and asked to confirm the start of the quiz.  The page contents are basically the same as what the dialog had before.

Note that the page will not be displayed if the quiz is not a timed quiz, if an instructor is acting as a student and has permissions to start a new version when acting (another page is already shown then), or if the quiz is a proctored gateway quiz (then you get the proctor login page).


drgrice1:
@drdrew42 and @mgage:  I typically keep webwork2 in one browser tab and pg in another, and I accidentally switched browser tabs and requested the review on the wrong pull request.  I meant to request your reviews on openwebwork/pg#642.  Although you can review this one too if you like.

drgrice1:
Note that you should make sure to update the theme css by running `npm run generate-css` (or just `npm install`) to test this.

pstaabp:
I see the new confirmation page when going directly to a quiz--I just copied a link and pasted into a new tab to get this. 

> Note that the page will not be displayed if the quiz is not a timed quiz, if an instructor is acting as a student and has permissions to start a new version when acting (another page is already shown then), or if the quiz is a proctored gateway quiz (then you get the proctor login page).

I don't see the new confirmation page if I access this directly from the  Homework Sets page.  Should I?  If not, why not.  It would seem helpful.

drgrice1:
You should see the confirmation page if you go directly from the homework sets page.  If not, I would also like to know why not, unless it is one of the cases I listed.

Also, this pull request is probably going to be dropped.  #1561 is a better idea, and will supersede this.

pstaabp:
> You should see the confirmation page if you go directly from the homework sets page. If not, I would also like to know why not, unless it is one of the cases I listed.

I had a timed quiz and was a student -- not just acting as one.  I assumed that I should see the screen.  

Even with #1561, I think a better confirmation to start the quiz is helpful, so maybe after that is pulled in, we can tweak the start of quiz.

drgrice1:
I will withdraw this now.  This won't be needed with the new quiz version page in ProblemSet.pm.

PR #1561, opened 2022-01-24
Create ProblemSet.pm page for gateway sets.

somiaj:
This is a draft of using the problemSet page for gateway quizzes to act as a quiz information page that provides information about the quiz, includes a start new version button, and lists the quiz versions. As a part of this, gateway quizzes can now use set headers to display information about the set.

This removes quiz versions from the problemSets list, as they are now all accessible via the quiz problemSet page.

The problemSets list looks like:

problem-setlist

And the problemSet details for a quiz look like:

problem-quizdetails


somiaj:
Additional Info:

I am using this on my live server with LTIAdvanced. All my tests so far have been good. I can now link the LTI assignment directly to this info page, and quiz versions won't automatically spawn when students access the quiz from the LTI. As for my test student the 'homework' LTI grade link is working. This week I will see how it holds up when my students take their next quiz.

This was my main motivation, to provide a nicer way to link my LTI with quizzes, by providing a landing page that can then be used to start a version or access old versions, and all my tests seem to work (I only use 1 version of non proctored quizzes so they are limited).

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -386,81 +386,207 @@ sub body {
...
+			CGI::th({ -scope => 'col'}, 'Attempts'),
+			CGI::th({ -scope => 'col'}, 'Status'),
+			CGI::th({ -scope => 'col'}, 'Score'),
+			CGI::th({ -scope => 'col'}, 'Start Date'),
+			CGI::th({ -scope => 'col'}, 'Close Date'),
+			CGI::th(
+				{ -scope => 'col', class => 'hardcopy'},
+				CGI::i(
+					{
+						class => 'icon far fa-arrow-alt-circle-down',
```
drgrice1, line 456: This also needs the css clas `fa-lg`.  That makes the icon a little bigger and is consistent with the appearance of this icon on the problem sets page (See line 260 of ProblemSets.pm).

dlglin:
Continuing the discussion from https://webwork.maa.org/moodle/mod/forum/discuss.php?d=5571

I have concerns about replacing "version" with "attempt".  While @drgrice1 is probably correct that students (and possibly instructors) might not understand what a "version" of a quiz is, I think that calling them "attempts" could lead to a different confusion in cases where multiple submissions are allowed on a version.

e.g. (using the current terminology): Create a quiz that allows three versions, and two submissions per version.  In such a scenario when I see "attempt" I would equate that with a submission, and not a new version.  Thus I would interpret this quiz as allowing six "attempts".

At my institution we use the term "version" in exactly this sense for MC (bubble sheet) exams: Different versions contain different randomizations of the questions, with a different answer key, so this terminology is used in at least some contexts.  Perhaps @glarose has thoughts, since he initiated this terminology.

Some of this can be helped by the page itself.  What about changing the "Start New Attempt" button to "Generate a New Quiz"?  I think "Generate" makes it clearer that this will be a different randomization of the quiz.  Also, including the number of used/available submissions next to each version would help to distinguish between versions and submissions.

drgrice1:
I disagree with you still @dlglin.  You just have multiple submissions for an attempt.  It still makes perfect sense to say you have multiple submissions per attempt.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -386,81 +386,207 @@ sub body {
...
+			my @problemRecords = $db->getAllProblemVersions($effectiveUser, $set->set_id, $ver);
+			my $score = 0;
+			my $open = $self->formatDateTime($verSet->open_date(), undef, $ce->{studentDateDisplayFormat});
+			my $close = $self->formatDateTime($verSet->due_date(), undef, $ce->{studentDateDisplayFormat});
+			my $hardcopy = '';
+			my $status = ''; ## Future use for 'will open on'/'open'/'reduced scoring period'/'closed'
+
+			my $interactiveURL = $self->systemLink(
+				$urlpath->newFromModule($urlModule, $r,
+					courseID => $courseID, setID => $verSet->set_id.',v'.$ver )); # Why doesn't $verSet->set_id include version?
```
drgrice1, line 477: The reason that $verSet->set_id does not include the version is due to the way that the database works for versioned sets.  Adding the ",v%" suffix was a way to store the versioned sets in the database in the same way as other sets without adding another column to the database, and the getSetVersion was implemented in this way to make it look like the version is a separate column.  So the id matches the template set and the version is in a separate field.

Now that you know the answer, please remove this comment from the code.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -386,81 +386,207 @@ sub body {
...
+					my $pval = $pRec->value() ? $pRec->value() : 1;
+					if (defined($pRec) && $score ne 'undef') {
+						$score += $pRec->status()*$pval;
+					} else {
+						$score = 'udef';
+					}
+					$possible += $pval;
+				}
+				$score = wwRound(2,$score);
+				$score = "$score/$possible";
```
drgrice1, line 502: This can be greatly simplified.  The code in ProblemSets.pm that you copied here was not done right to begin with.  There is a grade_set method in lib/WeBWorK/Utils.pm that can do this for you.  Using that method also ensures consistently with the other places in WeBWorK that grades are computed.  Most of them already use this method.  Change the code here to:
```suggestion
				my ($total, $possible) = grade_set($db, $verSet, $verSet->set_id, $effectiveUser, 1);
				$total = wwRound(2, $total);
				$score = "$total/$possible";
```
Of course you will also need to add `grade_set` to the list of things that are imported from the WeBWorK::Utils module on line 32 of this file, and you should delete line 468 above as that is no longer needed.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -386,81 +386,207 @@ sub body {
...
+		} else {
+			print CGI::p($r->maketext("You can no longer start new attempts."));
+		}
 
-			if ($problem->flags =~ /essay/)  {
-				$canScoreProblems = 1;
-				$gradeableProblems[$problemID] = 1;
+		print CGI::start_table({
+				class    => 'problem_set_table table table-sm caption-top font-sm',
+				-summary =>$r->maketext(
```
drgrice1, line 438: Please remove the `-` prefix on `summary`.  I removed most of these in my bootstrap 5 pull request and missed this one (where you copied this from).   The `-` prefix actually makes perl CGI try to maintain order of the attributes, which isn't needed and is inefficient.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -386,81 +386,207 @@ sub body {
...
+					'This table lists the current attempts for this quiz for this class, '
```
drgrice1, line 439: Remove "for this class" from this.  The fact that the quiz is for this class is rather clear.

drgrice1 (commented):
This looks like a good start.  See my comments for a few things I observed on my first initial perusal.

The checkboxes for hardcopy printing are still needed in the version list on the problem set list page.  This will take some care to make this work with the current setup since usual homework sets will not have these checkboxes.

If a current quiz is in progress, then the "Start New Attempt" button should not say that anymore.  Instead it should say "Resume Current Attempt" (or "Version" depending on the outcome of the Version/Attempt discussion).  That is what happens anyway as the GatewayQuiz.pm modules detects the in progress attempt and reopens it.

This will replace my current gateway quiz confirmation page pull request #1560, but I would suggest that you integrate the page part of that into this pull request for the top part of the new page that shows the time (if timed) and start new attempt button.  That is lines 1352-1433 of GatewayQuiz.pm in that pull request.  I can add this later if you don't want to do this now.

dlglin:
> I disagree with you still @dlglin. You just have multiple submissions for an attempt. It still makes perfect sense to say you have multiple submissions per attempt.

@drgrice1 this assumes that a user intuitively understands the difference between submission and attempt, and I'm not sure that this is going to be everyone's default interpretation.  Our understanding of the term "submission" may come from having already used it in this context.  I'm curious how someone without that background would read this.

dlglin:
Some suggestions:

- Include the Set Description somewhere on this page
- Include the number of submissions allowed per version along with the time ("You may submit each version n times")
- Show the number of submissions made/remaining for each version
- In your screenshot the Status column is empty.  Thinking of possible values, I've come up with "Still open", "Completed", and "Completed; can be reviewed after ...".  I can't see a scenario where a version will be in a "Will open..." state.
- Perhaps instead of having the version names be links themselves, have a button next to a version labelled "Continue" or "Review".  Right now there is no distinction between opening a completed quiz and a still open one.

somiaj:
> Some suggestions:
> 
>     * Include the Set Description somewhere on this page
> 
>     * Include the number of submissions allowed per version along with the time ("You may submit each version n times")
> 

My vision here is a patch on top of this which will add a 'set details' table near the top. I am not a fan of the reduced scoring date/due date being part of the header for the page. Instead I think that and other info should be included in this details table which contains stuff like open/reduced scoring/closed/answers available/description/time limit/versions allowed/ and so forth. And this table should be dynamic enough to only show the details which are important to the set in question.

drgrice1:
@somiaj:  The status will need to be in the table, because it is different for each attempt.  An attempt may be "Open", "Completed", or "Over time, closed".  This was displayed before, and will still need to be displayed.

somiaj:
@drgrice1 Rodger, that is what the place keeper in the table is for. Since I use quizzes in such a limited way, I haven't fully wrapped my mind around all possible status and how to check for them, that is also on my todo list.

drgrice1:
> @drgrice1 this assumes that a user intuitively understands the difference between submission and attempt, and I'm not sure that this is going to be everyone's default interpretation. Our understanding of the term "submission" may come from having already used it in this context. I'm curious how someone without that background would read this.

To be quite frank, you have it backwards.  The confusion is caused by using the wording "version" and not "attempt".  Everyone is confused by this that sees WeBWorK's settings in the problem set details the first time.  They always have to ask how to give students multiple attempts.  If the settings in the problem set details said "Number of Attempts per Time Interval", "Number of Graded Submissions per Test", and "Time Interval for a New Set of Attempts", instead of "Number of Tests per Time Interval", and "Number of Graded Submissions per Test", and "Time Interval for New Test Versions", things would be much clearer.

dlglin:
> To be quite frank, you have it backwards. The confusion is caused by using the wording "version" and not "attempt". Everyone is confused by this that sees WeBWorK's settings in the problem set details the first time. They always have to ask how to give students multiple attempts. If the settings in the problem set details said "Number of Attempts per Time Interval", "Number of Graded Submissions per Test", and "Time Interval for a New Set of Attempts", instead of "Number of Tests per Time Interval", and "Number of Graded Submissions per Test", and "Time Interval for New Test Versions", things would be much clearer.

I agree that the current problem set detail page is confusing, but I'm not convinced that changing "version" to "attempt" solves that.  If I read "New Set of Attempts", I would interpret that to mean a set of attempts on one version.  How about "Number of Tests per Time Interval", "Number of Graded Submissions per Test", and "Time Interval for a New Set of Tests"?

Another option would be to make the time interval an advanced feature (hidden by default).  Then the wording on the main page could be simplified to something like "Number of New Tests a Student can Generate" and "Number of Graded Submissions per Test".

This would also benefit from some embedded help, either as inline descriptions of the settings, or a link to a page with some examples.

drgrice1:
My alternate wording was supposed to read:
"Number of Attempts per Time Interval", "Number of Graded Submissions per Attempt", and "Time Interval for a New Set of Attempts"

In any case:
"Number of Test Versions per Time Interval", "Number of Graded Submissions per Test Version", and "Time Interval for a New Set of Test Versions"
would be better than what is currently stated.

somiaj:
Updates:

I now add a status based on if the test is open/closed, how many submits that remain, and if answers are available. I have changed the columns to start/end, which will be the last time they hit the submit button (if they did). If they haven't hit the submit button it will let them know the time that remains or overtime (though this is static, and maybe misleading if a student sits a a page for a while?)

The continue current version of the gateway quizzes (when not specifying a version) seems to only work in some conditions (though most likely due to my strange versions with spawning them then changing settings/spawning again), anyways, I had an open version with no submits and the GatewayQuiz page told me I couldn't create new versions. So instead if there are any open versions that are not submitted yet, I provide a button directly to that version. I also provide a start new test button as well and both could be visible, since it could be possible a student can continue a current version or start a new one.

I have added the checkboxes for hardcopy generation.

I removed the time limit statement (plan to move this to a set details table with other info, but that I think should be another patch).

I switched back to calling them versions. Mostly for consistency, as this is what they are called elsewhere, and it is beyond the scope of this to rename things. I do agree that some cleaning up on language and naming would be nice, but not something I want to do here. That should probably be another pull request.

To go with the 'Preview Test' and 'Grade Test' buttons in the actual test, I tried to be more consistent at just calling them Tests.

I addressed @drgrice1 code comments, but then rewrote a good chunk of it (to hopefully clean it up and generate the data for the table at the start before printing).

Current screenshot:

myscrot

somiaj:
> You are incorrect in showing both the button to start a new test and the button to continue an already started test. The reason is that the gateway quiz module does not allow starting a new test while an existing test is in progress. The result is that both buttons actually do the same thing.

That is not my experience and both buttons do different things here. I am able to keep spawning versions even if I have open version that I have not answered any submits. The only time heading to a non-versioned link for the quiz took me to currently open version is once I ran out of versions in the current time interval.

I was just working with the behavior I saw, could it be the gateway module isn't functioning as intended here?

drgrice1:
I don't know why that would be happening for you, but the GatewayQuiz.pm module only creates a new version when there are no versions in progress.

somiaj:
This has been my experience across multiple WeBWorK installs (I also found it a bit annoying, but once I set the version cap to 1, it hasn't been an issue since).

Anyways, I was just able to reload the main quiz_mode link, and it created my student account a new version until it hit the test versions per time limit interval. At that time it stopped creating versions, and returned me to the last open version. So to me the module isn't working as you described. I kinda thought this was by design, but now it sounds like it might be an issue? No one else has experienced this with tests that allow more than 1 version in a given time interval?

I do agree that both buttons shouldn't be there, I just thought the design was to allow for such a thing. last updates fix your comments and now only one button appears or a message about when a new test might be available.

somiaj:
Update, checkboxes only shown for users who can download multiple sets, otherwise and arrow is shown for students.

somiaj:
Here are my quiz settings for my latest test:

myscrot

After staring a new test I just hit reload, and it generated 5 versions, and finally after generating all 5 versions would take me to the last running version. Here is it is seen on the updates quiz page.

myscrot

drgrice1:
There are three instances on lines 380, 385, and 389.

somiaj:
Updated the classes for you. Noticed an issue where grades were being shown before any submits were made. Now grades require at least one submit to show.

somiaj:
Update: Only display answer download link (arrow icon) if answers are available to non multiSet users.

drgrice1:
@somiaj:  I converted this to a draft pull request for now, so that it is not accidentally merged.  When you are ready you can change it back to a regular pull request.

drgrice1:
@somiaj:  So it is the "Number of Graded Submissions per Test" setting that does it.  If that is set to 0 (for an infinite number of submissions per test), then every time you use the "unversioned" link a new test is created.  If that is set to anything else, the "unversioned" link opens the last in progress test.   That is a bug.

somiaj:
@drgrice1 Okay, I just thought that was the way it was meant to act.

Just to double check the logic, does it resume an open version only if it has zero submits (logic I'm using here)? Thus one doesn't have to use all available submits per version?

pstaabp:
Finally taking a look at this.  First of all, I think this clarifies quizzes a lot and make this much more consistent with problem sets in that clicking on a Quiz, shows more details instead of starting a quiz.  I think this and a combination of #1560 will help even more. 

@drgrice1  mentioned:
> I don't know why that would be happening for you, but the GatewayQuiz.pm module only creates a new version when there are no versions in progress.

I'm able to get many current versions running at the same with the develop branch.  This new PR seems to eliminate that. 

image

I haven't messed around with any of the setting that might prevent this, but default settings should probably prevent this. 

In terms of wording, if a quiz is still open, then under the "End" column, how about "Quiz is open.  XXXX mins remaining." or "Quiz not yet submitted. XXX mins remaining."

drgrice1:
@somiaj: No.  It has nothing to do with that.  Currently it creates a new version any time you have that setting set to 0.  It shouldn't.  If there is one submission, then a new version should be created, but not if there are no submissions.  One should not be required to use all submissions.

@pstaabp:  I see now that the reason you were not seeing the quiz confirmation page is because you had the quiz set for an infinite number of submissions per test.  That is what @somiaj and I were just discussing here.  There is a bug in the Gateway quiz logic for when a new version should be created that causes that.

somiaj:
@pstaabp Suggestions added:

myscrot

somiaj:
@drgrice1 Is there already a macro to say turn time in seconds to 'XXX min', 'XXX hours', 'XXX days', etc? `4844 min` might be better displayed as `3 days`.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,386 @@ sub body {
...
+			$data->{id} = $set->set_id.',v'.$ver;
+			$data->{version} = $ver;
+			$data->{start} = $self->formatDateTime($verSet->version_creation_time, undef, $ce->{studentDateDisplayFormat});
+
+			my $timeLeft = int(($verSet->due_date - $timeNow)/60);
+			if (defined($verSet->version_last_attempt_time) && $verSet->version_last_attempt_time > 0) {
+				if ($timeNow < $verSet->due_date && ($maxSubmits <= 0 ||
+						($maxSubmits > 0 && $verSubmits <= $maxSubmits))
+				) {
+					$data->{end} = "Additional submissions avilable. $timeLeft min remain.";
```
drgrice1, line 445: This string should be translated (via $r->maketext).

drgrice1:
Not that I know of.  I think you can safely assume that no one will actually set a time limit that is that long.  You should look at the code I have in #1560 for this.  It does hours, minutes, and seconds, and that should be sufficient.

somiaj:
Added a details table that list information about the set and removed this from the set header. Here is the preview.

myscrot

and for homework

myscrot

Though ran into a strange issue when the set description is NULL in the database.

myscrot

I don't plan to give time in seconds, just looking for initial feedback as I work on printing better units.

drgrice1:
I don't like the change making this overly detailed table at the top.  Student's don't want to see something like that.  It should be kept much simpler.  I recommend putting it back to how you had it before that.

Why do you think that message that you showed is due to a NULL value in the database?  It looks like that is caused by a missing string in a po file.  Do you have translations enabled?  I recommend that you do not have them enabled at this point unless you are specifically testing them.  Right now there are lots of strings that are not in the pot/po files.  Before the next release we will update them, and that will fix those issues.  If you want to update the pot/po files, then run bin/update_localization_files.  If you do that make sure that you do not commit the changes though.

drgrice1:
Also, I recommend that you stop force pushing your changes, and start adding commits.  This has gone far enough that changes may need to be reverted.  Adding commits makes it easier to track what has been done, and go back if needed.

drgrice1:
Note that the `Set Info` show to the right is where the information you have added in the table should be.  It is up to the instructor to put whatever information they want there.  So please remove the table at the top.

drgrice1:
I see the issue with NULL in the database that you are talking about.  Not all sets have descriptions.  So you will have to check if that is defined and non-empty, and not show the description it isn't defined or is empty.

somiaj:
The set details is a separate commit, so I can revert that. Would putting that info in a collapse so it isn't taking up much space but still be available be useful? Or moving it to the bottom? Either way, if it isn't wanted I will revert.

Thanks for the pointer there, that was what I was missing was checking for it was also not empty, I was only checking for defined.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -355,14 +336,70 @@ sub body {
 	}
 
 	my $isJitarSet = ($set->assignment_type eq 'jitar');
+	my $isGateway = ($set->assignment_type =~ /gateway/);
 
+	# Hash for set details table
+	my @setDetails = ();
+	if (defined($set->description)) {
```
drgrice1, line 343: Change this to 
```suggestion
	if ($set->description) {
```
and it will fix the NULL issue that you are seeing.

drgrice1:
I think it shouldn't be there at all.  Again, that is what the `Set Info` is for.  You could add the set description as a sub header, but the rest is generally to much.  It is also redundant as that information is on the ProblemSets page.  It isn't needed again here.

drgrice1:
I am thinking about reworking the problem sets page.  I mentioned this to the others at the development meeting last week.  I would like to remove the table, and instead have each homework set/quiz in its own bootstrap `card`.  Then we could reorganize the information that is shown, and make it more clear.  Once the student enters the homework set or quiz they should not need all of that again, and can focus on doing the work.  The `Set Info` can add anything else the instructor wants to say.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
somiaj: Updated to translate this and other strings in the `$data` hash.

somiaj:
Okay, since the set details table wasn't wanted, unless there are issues I need to address, I'll leave it alone now.

I do look forward to seeing a more configurable ProblemSet page so I could put `Set Info` front and center. I would also like the due date/reduced scoring date to be removed from the page header.

@drgrice1 Thanks again for all your help and feedback.

drgrice1:
Hopefully you still have that table that you removed.  Hang on to that work and see what others say about that.  If others think that is a good idea, then you can add it back.  Perhaps if the LMS restricted navigation pull request (#1445) is ever finished and added that might be desired.  At least in that case where students will not see the problem sets page at all.

somiaj:
Oh wait, I misunderstood, you are talking about updates to the problem sets page.

I do still have that details table commit, I put it another local branch if others end up wanting it.

drgrice1:
Yes, I was talking about the ProblemSets.pm page, not the ProblemSet.pm page.

What you are asking, to make the page configurable so that `Set Info` is front and center, is not a good idea.  The set info shouldn't be "front and center".  It should be where it is.  What you are saying is that you disagree with the layout.  That is fine, although I disagree with you, but not something that should be configurable.

So, why do you want the set info front and center as you put it?  It works well to the right.  You can put the info you want there, and it is perfectly visible to students.  Also note that on narrow screens (like mobile phones), the set info is at the top in the center.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,388 @@ sub body {
...
+				-method => 'POST',
+				-action => $hardcopyURL
+			),
+			$self->hidden_authen_fields;
+		}
 
-	if (@problemNumbers) {
-		# This table contains a summary, a caption, and scope variables for the columns.
-		print CGI::div({ class => 'table-responsive' });
 		print CGI::start_table({
```
drgrice1, line 568: Talking about the `Set Info` and noting its behavior on narrow screens reminded me that I had better check this pull request for its behavior on narrow screens.  As such I see another thing needed here.  This table should be wrapped in a `div` with the `table-responsive` class.  So change this to
```suggestion
		print CGI::start_div({ class => 'table-responsive' });
		print CGI::start_table({
```
and then add `print CGI::end_div();` after line 654.

drgrice1:
Another thing that I just noticed is that the "Test Versions" table is still there even when there are no existing versions.  When there are no existing versions that table should not be shown.

drgrice1:
When a homework set has no problems the page shows "This homework set contains no problems." where the table would be.  For quizzes it should probably say something like "No versions of this test have been taken."

somiaj:
Unless anyone notices any issues with the current state of things, I think this is now ready to go.

I just gave a quiz today using a similar (older version) modification to my live server, and everything has worked well, and the LTI integration also is working when linking the assignment directly to this new quiz ProblemSet page (Note I only use one version, but I assume if one version links correctly, so will multiple versions).

If the set details table (pictured above) is something that others might be interested in, I think that should be done in a new pull request so it can be discussed more. I will only create a new pull request if others would actually want that feature, but as @drgrice1 mentioned, all of that info can be put in the `Set Info` header now, and this allows instructors to provide and have way more control of any additional info that should be included with the quiz.

drgrice1:
There is some cleanup that will be needed before this can be merged.  I noticed this a while ago, and keep forgetting to mention that here.  You are removing the usage of the quiz confirmation dialog that I quickly added before the last release.  So you should delete the file `htdocs/js/apps/ProblemSets/problemsets.js`, and remove the `output_JS` method from the end of `ProblemSets.pm` (lines 559-567).

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,398 @@ sub body {
...
+			my $interactive = $r->maketext('Version #[_1]', $ver->{version});
+			if ($authz->hasPermissions($user, 'view_hidden_work') ||
+				$ver->{status} =~ /Open/ || $ver->{status} =~ /Answers/)
+			{
+				my $interactiveURL = $self->systemLink(
+					$urlpath->newFromModule($urlModule, $r,
+						courseID => $courseID, setID => $ver->{id} ));
+				$interactive = CGI::a(
+					{
+						title => '',
```
drgrice1, line 644: I just noticed another thing as well.  Previously the gateway template link (the main link) and all versions had a tooltip that showed the set description.  The template link on the ProblemSets.pm page still has that, but the versions should have that in the ProblemSet.pm page now.

Also, this should not set an empty title on these anchors.

So this should be changed to
```suggestion
						class             => 'set-id-tooltip',
						data_bs_toggle    => 'tooltip',
						data_bs_placement => 'right',
						data_bs_title     => $set->description(),

```

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -407,64 +332,48 @@ sub body {
...
+		$display_name = CGI::i(
+			{
+				class => "icon far fa-clock",
+				-title => $r->maketext("Test/quiz with time limit."),
+				-data_alt => $r->maketext("Test/quiz with time limit.")
+			},
+			'') .
+			' ' .
+			CGI::span(
+				{ class => 'gwquiz-info', data_time_limit => $set->{version_time_limit} },
```
drgrice1, line 372: This line should be deleted.  This was all here for the now obsolete quiz confirmation dialog, and is not used for anything else.

drgrice1: Of course then you can condense this all to `CGI::span($display_name)`.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,398 @@ sub body {
...
+		print CGI::end_form();
+	} elsif (! $isGateway) {
+		print CGI::div(
+			{ class => "problem_set_options mb-2" },
+			CGI::a(
+				{ href => $hardcopyURL, class => 'btn btn-primary' },
+				$r->maketext("Download PDF or TeX Hardcopy for Current Set")
+			)
+		);
 	}
```
drgrice1, line 770: You copied the incorrect `p` usage that I apparently didn't remove in my bootstrap 5 update pull request.  The `p` tag should only ever be used for literal paragraph type content.  Certainly not for buttons.  This should be changed to
```perl
	if ($isGateway && $multiSet) {
		print CGI::div({ class => 'mb-3' },
			CGI::reset({ id => 'clear', value => $r->maketext('Clear'), class => 'btn btn-primary' }));
		print CGI::div(
			{ class => 'mb-3' },
			CGI::submit({
				id    => 'hardcopy',
				name  => 'hardcopy',
				value => $r->maketext('Download PDF or TeX Hardcopy for Selected Tests'),
				class => 'btn btn-primary'
			})
		);
		print CGI::end_form();
	} elsif (! $isGateway) {
		print CGI::div(
			{ class => "mb-3" },
			CGI::a(
				{ href => $hardcopyURL, class => 'btn btn-primary' },
				$r->maketext("Download PDF or TeX Hardcopy for Current Set")
			)
		);
	}
```

Note also, that the code I pasted here is perltidied!

Github won't let me use the suggestion button here.  It says you can't apply suggestions to deleted lines, but I don't see any deleted lines here.  Hmm...

drgrice1: Also, note that the `problem_set_options` class is obsolete, and can be deleted everywhere.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,398 @@ sub body {
...
+			my $interactiveURL = $self->systemLink(
+				$urlpath->newFromModule($urlModule, $r,
+					courseID => $courseID, setID => $set->set_id.',v'.$continueVersion)
+			);
+			print CGI::start_div({-class => 'problem_set_options'});
+			print CGI::a(
+				{ href => $interactiveURL, class => 'btn btn-primary' },
+				$r->maketext('Continue Open Test')
+			);
+			print CGI::end_div();
```
drgrice1, line 508: This would be much better written as
```suggestion
			print CGI::start_div(CGI::a(
				{ href => $interactiveURL, class => 'btn btn-primary' }, $r->maketext('Continue Open Test')));
```
rather than three separate `print` statements.  Note the removal of the obsolete `problem_set_options` class.

Also, anytime you are passing arguments in a hash to a CGI method, drop the `-` prefix on the hash keys.  That isn't needed.  It is needed if the arguments are not inside a hash.

drgrice1: Note there are numerous other places that you use the `start_div`/`end_div` construction in this way.  Please consolidate  all of those.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,398 @@ sub body {
...
+			print CGI::br();
```
drgrice1, line 509: This would be much better written as
```suggestion
			print CGI::start({ class => 'mb-3' },
				CGI::a({ href => $interactiveURL, class => 'btn btn-primary' }, $r->maketext('Continue Open Test')));
```
rather than three separate `print` statements.  Note the removal of the obsolete `problem_set_options` class.

Also, anytime you are passing arguments in a hash to a CGI method, drop the `-` prefix on the hash keys.  That isn't needed.  It is needed if the arguments are not inside a hash.

drgrice1: Sorry, edited this.  I noticed the `<br>`.  Don't use those at all.  Use the bootstrap 5 margin utility classes.  In this case `mb-3`.  `<br>`'s don't work well in modern html markup.

I deleted my previous comments, but I am not sure it went through right.  The intent is that all of lines 503-509 be changed in that suggestion.

Furthermore, check your other markup.  There are several other places where things should be cleaned up like this.  I worked hard to remove a lot of this crap in the bootstrap 5 pull request, and I don't want to see it added back.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,398 @@ sub body {
...
-			if ($problem->flags =~ /essay/)  {
-				$canScoreProblems = 1;
-				$gradeableProblems[$problemID] = 1;
+			# Is it past due date?
+			if ($timeNow >= $set->due_date) {
+				$msg = $r->maketext('Test is closed.');
 			}
+
+			print CGI::div(CGI::p(CGI::strong($msg)));
+			print CGI::br();
```
drgrice1, line 555: There is excessive white space here due to the combination of a `p` tag and the `br`.  Bootstrap adds a bottom margin of `1rem` to `p` tags, so the `br` is unneeded.  Delete that.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,97 +377,398 @@ sub body {
...
+				$urlpath->newFromModule($urlModule, $r,
+					courseID => $courseID, setID => $set->set_id)
+			);
+			print CGI::start_div({-class=>"problem_set_options"});
+			print CGI::a(
+				{ href => $interactiveURL, class => 'btn btn-primary' },
+				$r->maketext('Start New Test')
+			);
+			print CGI::end_div();
+			print CGI::br();
```
drgrice1, line 534: Consolidate this as suggested in the other place, and remove the `br`.

drgrice1:
There is a little style tweak that improves the tables on both the problem sets page and the versions table on the set page.  That is delete line 476 of htdocs/themes/math4.scss (the `font-size: 3px` on the `.problem_set_table td a` elements).  That increases the font size of the `a` tags in these tables by a tiny amount.  The problem is that it makes the hardcopy icons larger in the table than in the header row.  Furthermore, the increased font size for the set names isn't needed.  In fact, I think that removing that increase makes the table fit better.

drgrice1:
I see where you are getting these large minute numbers from now.  Those come in when the gateway quiz is an untimed test.  In that case, the table should not show time remaining.  Those are open until the close date, and that date is already on the page.  So either put the close date in the table again, or just omit it.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
somiaj: I assumed you meant `CGI::div` up there not `CGI::start`.

somiaj:
@drgrice1 Split it up into multiple commits so you can better see the changes at this stage. Let me know if you want me to squish them together before final merge.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
drgrice1: Yeah, I accidentally had start_div still, and then I deleted the wrong part of that.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,115 +377,411 @@ sub body {
...
+		my $totalVersions = scalar @setVers;
+		my $timeLimit = $set->version_time_limit() / 60 || 0;
+
+		# Compute how many versions have been launched within timeInterval
+		#     to determine if a new version can be created, if a version
+		#     can be continued, and the date a next version can be started.
+		#     If there is an open version with no submits, add button to
+		#     continue the first such version found.
+		#     Build a data hash for each version that is used to create the
+		#     quiz versions table.
```
drgrice1, line 409: You don't need to change this, but I hate this kind of style for comments.   The only place that I have ever seen this kind of ugly commenting style is in WeBWorK code.  Why indent all of the latter lines in a comment?  When you indent a paragraph you indent the first line, not all of the rest of the lines.  This kind of formatting in no way improves readability.

drgrice1: To your credit, at least you capitalize, and use punctuation.

somiaj: I mostly just mimic the style I see around me, must have recently saw that comment style and went with it. I'll take note of this in the future.

drgrice1:
One thing I noticed just now is that if a timed quiz has not been opened yet, there is no indication anywhere of how much time the student will have to complete the exam.  The quiz confirmation dialog before gave that in hours, minutes, and seconds.  My quiz confirmation page (#1560) that was intended to replace that also gave this information.  I recommend that you add a method that does that conversion (look at my code for a translatable way of doing this), and display that above the `Start New Test` button.

You could then also use that method to show the time remaining in a better format in the table.  Only showing the remaining time for less than 90 minutes for a timed quiz is probably not enough.  My point was to not show the remaining time for un-timed quizzes.  Untimed quizzes should always show the close date in the table.  In progress timed quizzes should show the remaining time for anything less than 1 day.

drgrice1:
Here is the code for hours, minutes, and seconds from my confirmation page pull request converted into a method:
```perl
sub {
	my $self = shift;
	my $seconds = shift;

	my $r = $self->r;

	my $hours = int($seconds / 3600);
	$seconds %= 3600;
	my $minutes = int($seconds / 60);

	if ($hours && $minutes) {
		return $r->maketext('You will have [quant,_1,hour] and [quant,_2,minute] to complete the quiz.',
			$hours, $minutes);
	} else {
		# Translation Note: In this case only one of hours or minutes is non-zero, so the zero case of
		# the "quant" will be used for the other two.
		return $r->maketext('You will have [quant,_1,hour,hours,][quant,_2,minute,minutes,] to complete the quiz.',
			$hours, $minutes, $seconds);
	}
}
```
Note that I simplified it by removing seconds from the displayed time.  Seconds are probably not needed here. 

You could put this before the `sub body` method, and use it to both display the remaining time in the table, and the total time for a new version.

drgrice1:
I guess the wording in that method won't work in both places.  Hmm...

somiaj:
In your code why do you have `[quant,_1,hour,hours,]` in one location while only `[quant,_1,hour]` in other locations. Wouldn't the plural need to be provided in other places too? Is this you need that blank final item in the list to get no output for the zero?

somiaj:
I ran into the same issue with the wording, and didn't see a nice way. Also in your original code, the time limit is set in terms of minutes on the configuration page, probably reasonable to assume seconds aren't needed there.

drgrice1:
> In your code why do you have `[quant,_1,hour,hours,]` in one location while only `[quant,_1,hour]` in other locations. Wouldn't the plural need to be provided in other places too? Is this you need that blank final item in the list to get no output for the zero?

`[quant,_1,hour,hours,]` will result in `1 hour` if the quantity is 1, `2 hours`, if the quantity is 2, and the empty string if the quantity is 0.

`[quant,_1,hour]` will result in `1 hour` if the quantity is 1, `2 hours`, if the quantity is 2, and `0 hours` if the quantity is 0.

Basically, as you know in English the plural form of most words is the same, but with an "s" added, and so the plural form is optional and if that second optional argument is not provided the result with be the singular form with an "s" added.  However, the third optional argument is the zero form.  If the zero form is not provided then it is assumed that you use the plural form for zero as is also common in English.  So in the `[quant,_1,hour,hours,]` case I am providing the optional second argument, really so that I can provide the optional third argument, which is the empty string.  So I am saying drop the quantity from the resulting string if the quantity is zero.

drgrice1:
So there is a minor issue since we dropped seconds from the display.  If the student is in the last minute of a test, and for some reason gets signed out and has to go back to the problems sets page to get back into the test, then the page displays "You have remaining to complete the test."  This is probably not to big of a deal, because this probably won't happen often.

drgrice1 (approved):
I think this pull request is looking good, and I think this is ready for merge consideration.

Thanks for working with me with all of my tedious requests for changes.  I think we both want the same thing though.  Something that looks good, works well, and is intuitive for students.

Good work!

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,115 +377,487 @@ sub body {
...
+				if ($verSubmits > 0) {
+					$data->{status} = $r->maketext('Completed.');
+				} else {
+					$data->{status} = $r->maketext('Closed.');
+				}
+			}
+			if (($verSubmits > 0 || $timeNow >= $verSet->due_date) && ($set->hide_work eq 'N' ||
+					($set->hide_work eq 'BeforeAnswerDate' && $timeNow >= $set->answer_date))) {
+				$data->{status} .= ' ' . $r->maketext('Answers Available.');
+			}
```
drgrice1, line 507: The "Answers Available" message in the status column is not being showed at the correct time.  Currently, there are some cases where this is showing that answers are available, when they are in fact not available.  The correct logic for this is in the `can_showCorrectAnswers` method of the `GatewayQuiz.pm` module (starting on line 82 of that file).  Note that the `$attemptsUsed` in that method will be the `$verSubmits`, and you won't need to worry about the `$addOne` there.

The main thing that I see that this is missing right now is that you are only checking if the number of submissions is greater than zero, when you should be checking if that is greater than the maximum number of attempts allowed, and if that maximum number of attempts allowed is non-zero (answers are never available for infinite number of submissions).  It looks like there might be some other cases that are missed as well.

In general, I think there is going to be a need for careful checking of the logic for everything in this pull request.

somiaj: This was my biggest worry when attempting this was to make sure to be consistent with how things currently work. Though outside of showing the score, everything is then double checked by the gateway module, so it will be at most an inconsistency (which is why I felt I could attempt this).

In double checking things, here I just was using the same logic as show score, which of course I shouldn't have.

In testing this a bit more, I did run across something that I was getting `$maxSubmits` from the `$Problem->max_attempts()` which seems to be used for something else, I needed `$verSet->attempts_per_version`, so updating that has fixed some things.

One thing I found a bit surprising when looking at the `can_showCorrectAnswers` method was that their is additional check that the `due_date` and `answer_date` have to be the same to see answers before the `answer_date`. I guess this makes sense, if someone set that date they want to delay answers, but this seems to disagree with the setting of 'show answers on completed tests' to 'yes'. If I set yes, show answers on completed tests (as opposed to after answer date), this additional check seems to give me unexpected behavior.

Fixing the above issues then updating the code to include this additional check.

drgrice1:
Note that the due_date and answer_date for the student's versioned problem is not the same as the due_data and answer_date of the template, and is changed by the GatewayQuiz.pm module when a student submits their final attempt.

somiaj:
Though here I don't think the issue is that they aren't the same, it is that they are not equal (and if they aren't equal in the template they will still be not equal in the student version?)

Anyways in testing this, I had a student who had completed the quiz and couldn't see the answers because the answer_date and due_date were different even though show answers on completed tests was set to yes, but as soon as I made them equal to each other in the versioned test, they could see answers.

Another possible issue, should I be using `$db->getMergedSetVersion` here, that way if something like hide_work isn't set in the student version I don't have to additionally check the template set?

```
my $verSet = $db->getSetVersion($effectiveUser, $set->set_id, $ver);
```

lib/WeBWorK/ContentGenerator/ProblemSet.pm
somiaj: I think I see what is going on now, this will allow it to be possible that students can see their work but not the answers. Makes sense, updating logic so links can still be active even if answers aren't in this case.

somiaj:
> Thanks for working with me with all of my tedious requests for changes. I think we both want the same thing though. Something that looks good, works well, and is intuitive for students.

Thanks for all the help and looking at the code so closely. This is something I have been wanting for a while but wasn't sure if I knew enough about perl and webwork to pull it off, it was really helpful. Glad I could contribute.

I have tried to clean up the logic subtly a bit, I'm hoping my understanding of the `getMergedSetVersion` is correct, but with poking around with a few more tests things seem to be more consistent with current behavior of the gateway module. I do suspect there are some subtle edge cases in the logic, but some testing will hopefully help track them down.

Do you want me to squish these commits together so not so many get merged into the main devlop branch?

drgrice1:
I think that you can leave the commits.  No need to squash them.  We can always do that when we merge the pull request if we deem it appropriate.  I doubt that we will in this case though.  You kept things pretty well organized.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -371,115 +377,493 @@ sub body {
...
-	# Check permissions and see if any of the problems have are gradeable
-	my $canScoreProblems = 0;
-	if ($authz->hasPermissions($user, "access_instructor_tools") &&
-		$authz->hasPermissions($user, "score_sets")) {
+			if (defined($verSet->version_last_attempt_time) && $verSet->version_last_attempt_time > 0) {
+				if ($timeNow < $verSet->due_date && ($maxSubmits <= 0 ||
+						($maxSubmits > 0 && $verSubmits < $maxSubmits))
+				) {
+					if ($verSubmits > 0) {
+						$data->{end} = $r->maketext('Additional submissions avilable.') . " $timeLeftText";
```
pstaabp, line 467: typo: avilable should be available.

pstaabp:
Decided to look at this with a fresh new quiz and if I have one I get:
image

I'm not sure what the Clear Button is supposed to do--figured out what it is (comments below) and probably we should remove it.   Also, since there are no quizzes taken yet, it doesn't make sense to have the Hardcopy button there.  

I figured out later that the Clear button, clears the download checkboxes. I guess this is from the original ProblemSets page--which isn't completely clear (pun intended). I realize this isn't something you added.  Just to clarify, maybe we should change this to "Clear Selected Sets" on the overall problem sets page and maybe "Clear Selected Quiz Versions" on the new page.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
somiaj: Thanks, squished the typofix into the previous commit.

somiaj:
@pstaabp maybe remove the clear language, how does 'Deselect All Sets' and 'Deselect All Test Versions' sound to better clarify what it does? Though I guess here the more standard approach would be to have a checkbox in the table header the user can use to select all/deselect all sets, and maybe a reset button isn't needed here.

pstaabp:
I think that language sounds good.  I think we talked about having a checkbox in the table header sometime recently, but people preferred the down arrow to indicate "downloading".  Maybe someone can remind me about this discussion.

pstaabp:
@somiaj  This looks great now.  I'll merge it in. Again, I know there were a lot of back and forth and changes here, but this significantly helps out a bunch of problems we've had with quizzes.

drgrice1:
@somiaj:  You mentioned that you want to have the set info at the top instead of to the right.  There is a way that you can do that using the math4-overrides.js file.  Add the following to that file in the directory of the theme that you are using and you will get that.
```javascript
const infoPanel = document.getElementById('info-panel-right');
infoPanel?.classList.remove('order-md-last', 'col-md-4');
infoPanel?.classList.add('col-md-8');
infoPanel?.parentNode.classList.add('flex-column');
```
Note that this will also put the site info and course info on the top.  If you only want to do it on the problem set page, it would take more care.  You would have to find something that is unique for that page and wrap the above in an `if (unique thing) {}`.

somiaj:
@drgrice1 Thanks. I was looking at similar ideas as well, and ran into the same issue with it would also affect the site info on the main page. Though as I thought about it more, maybe you are correct here. I personally like to have info about each assignment (especially quizzes) that students 'cannot miss' to help them out or set the standards for the assignment/quiz. I found that off to the side some students would 'miss' or more likely just ignore it. Though the students who 'miss' it sitting off to the side are the same ones who will probably just scroll right past it, so making it slightly inconvenient for other students to get to the homework problems isn't worth trying to make information more front and center for some students who seem to 'miss' the info I have put there.

Thanks again for the suggestion.

somiaj:
@pstaabp @drgrice1 Thanks for all the help to get this to the point it can be merged. I know there were lots of comments on the code and desired info that is displayed on this new page. I'm still quite a novice, so it was really helpful to have constant feedback to take my very crude idea and make it a reality. Thanks for all you do with WeBWorK.

PR #1563, opened 2022-01-25
Gateway quizzes should not always create a new version when attempts_per_version is 0.

drgrice1:
Currently when attempts_per_version (Number of Graded Submissions per Test) is 0 and you click on the gateway template in the problem sets page, it always creates a new version even if there is a version in progress with no submissions.  In this case, the in progress version should be opened instead, the same as it does for any value greater than zero.

The current logic checks if the current number of submissions is greater than or equal to the max number of submissions or if the time has expired.  If that happens then a new version is created.  So if the max number of submissions is greater than zero then it does not create a new version until all submissions are used up, which is how it presumably was intended.

However, if the number of submissions is meant to be infinite, and so the setting is 0, then the current number of submissions is always greater than or equal to the setting of 0, and so it always creates a new version even if no submissions have been made to an in progress version.

This changes the logic so that if attempts_per_version is 0 (so an infinite number of submissions is allowed), then at least one submission is required before a new version is created.


dlglin:
This would lead to some inconsistent behaviour.  Consider the following two scenarios:
Quiz A: Unlimited submissions allowed:
Student has submitted the quiz once.  They click on "Take Quiz A test".  They get a new version.
Quiz B: 5 submissions allowed:
Student has submitted the quiz once.  They click on "Take Quiz B test".  They get sent back to the same version.

From a student's perspective, they do the same thing, but get a different result.

My understanding of the original design was that there was never intended to be more than one version "live" at the same time, which doesn't admit the scenario of unlimited submissions, no time limit and multiple versions allowed, since there would be no condition that would allow the student to start a new version.

If #1561 is merged, then this probably becomes moot, since that page can be designed to only show links that make sense.

drgrice1:
You do realize that it was inconsistent behavior before don't you?

drgrice1:
Just to summarize the behavior without this pull request:
Quiz A: Unlimited submissions allowed:
Student has opened a quiz but not submitted the quiz once. They click on "Take Quiz A test". They get a new version.
Quiz B: 5 submissions allowed:
Student has submitted the quiz once. They click on "Take Quiz B test". They get sent back to the same version.

drgrice1:
To add to that in the Quiz A unlimited submission case if the student has submitted the quiz once, goes back to the problem page, and clicks on Quiz A again, they again get a new version.  Always a new version.

drgrice1:
> My understanding of the original design was that there was never intended to be more than one version "live" at the same time, which doesn't admit the scenario of unlimited submissions, no time limit and multiple versions allowed, since there would be no condition that would allow the student to start a new version.

If there was never intended to be more than one version "live" at the same time, it was not working with unlimited submissions.  

Note that if #1561 is merged, there will still be more than one "live" version at a time.  With the way that @somiaj is setting it up it will change the existing behavior so that anytime there is more than one submission allowed, a new test version can be created as soon as one submission has occurred.  If you don't want that to happen, then you better comment on that pull request.

dlglin:
> > My understanding of the original design was that there was never intended to be more than one version "live" at the same time, which doesn't admit the scenario of unlimited submissions, no time limit and multiple versions allowed, since there would be no condition that would allow the student to start a new version.
> 
> If there was never intended to be more than one version "live" at the same time, it was not working with unlimited submissions.
> 
This is pretty much what I said in the paragraph you quoted.

The fundamental problem is that the "Take ... test" link is trying to guess what the student wants to do: resume an existing test or start a new one.  If #1561 is done right, then it fixes this by offering the student only explicit options, where the student knows exactly what they're getting.  In that scenario there is no issue with having multiple versions "live", since the student can properly choose what they want to do.

drgrice1:
So what is your problem here?  You still haven't made any real point.  This changes the guess that is made to a better one than it was making before.

drgrice1:
Are you saying you have a problem with this pull request?  That isn't even clear from what you have said.

drgrice1:
Furthermore, this pull request does not conflict with #1561, and is consistent with the behavior that is currently being implemented there, and even makes it easier to implement that behavior as planned there.

drdrew42:
My understanding is that #1561 provides a button to start a new version of a gateway assignment. This goes a long way towards alleviating the _accidental_ creation of overlapping (open at the same time) versions.

Whether or not we want to block simultaneous versions of a gateway assignment should be handled in issues (IMO) -- right now it is not blocked (in certain cases); and if the consensus is to block, that should be implemented in a separate PR after resolving these two open PRs.

drdrew42 (approved):
Works as advertised. For gateway assignments with unlimited versions, a student must have submitted their current version at least once before a new version will be created.

pstaabp:
Finally got to this.  Looks good.

pstaabp (approved):

PR #1564, opened 2022-01-26
Prevent problem randomization from saving global defaults to a user problem.

drgrice1:
When problem randomization is enable and the problem randomization count is saved to the database, save a "pure" problem with the updated count. This prevents all of the other database values from being overridden by the values in the global problem.


drgrice1:
This issue has come up several times in the forums.  The latest instance is discussed in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=6611.

drdrew42:
I was a bit confused by your description of this fix. Granted, what's going on is confusing already 😝 

In case it helps anyone else checking this out:
1. enable problem randomization
2. assign a problem set to a user with at least one problem that has a maximum attempts
3. have that user attempt (or even just view) a problem in that set
4. change the max attempts for said problem
5. confirm that the max attempts for the user did / did not (with this PR / without this PR, resp) update

Does that sound right? I assume the same could be tested for changing the weight of a problem instead of max attempts?

drgrice1:
Yes, that summarizes it.  You can also look at the values et in the problem_user table in the database.  Without this pull request you will see that all of the students previously NULL values are set after a student opens a problem.  With this pull request you will only see the prCount set.  The problem_seed is technically set, but should already be non-null is set to what it was before, so you won't see a change (unless requesting a new seed).

drdrew42 (approved):
👍

Issue #1540, opened 2022-01-14
issue with carriage return character at end of line in classlist file


Alex-Jordan:
An instructor created a classlist file that appeared to be formatted fine, but when we attempted to import it, there was an error message:

Wide character in crypt at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 1026. 

Eventually I sussed out that the file had carriage return characters (in addition to newline characters). I removed those and the file imports successfully.

Given instructors using Windows and other things, we should probably support the kind of file this instructor tried to use.

drdrew42:
This also happens when saving as any of the alternate CSV formats provided by excel.

pstaabp:
@Alex-Jordan can you provide a sanitized example of this?

pstaabp:
@Alex-Jordan I remember we talked about this last week.  I've been exporting classlists from excel in other formats.  I tried Macintosh Comma Separated and MSDOS separated and couldn't repeat this error.  I wonder if this is an export from Windows version of excel.

Alex-Jordan:
Here is one test file. I took the actual classlist file that an instructor provided. I opened it in vim and deleted all but 2 lines. I cleaned up things. I left the first 3 characters of the student ID as is, because the theory is that there is something hidden right at the start of each line that is the real problem (and not so much the carriage returns). I experience the issue when I try to import this classlist.

I haven't tried making something from Excel yet. However, in a separate incident, a different instructor raised the same issue. I told him about what Drew has pointed out, to use the very bare bones CSV output option in Excel. Then he reported that things worked. Could be Windows-specific.

I had to rename the file with .csv for upload to GitHub.
[testlist.csv](https://github.com/openwebwork/webwork2/files/7954182/testlist.csv)

drgrice1:
This seems to be your problem:  https://www.freecodecamp.org/news/a-quick-tale-about-feff-the-invisible-character-cd25cd4630e7/

I opened the file in binary mode with `vim -b` and I see this character at the beginning of your file.  I can delete that character in vim, and then webwork imports the students in the file fine.

I am not sure what to do about it though.

drgrice1:
[...]

PR #1566, opened 2022-01-27
Remove the byte order mark from classlist import files

drgrice1:
Remove the byte order mark that Excel seems to put at the front of csv files that it creates for some systems when importing a classlist file.

See https://en.wikipedia.org/wiki/Byte_order_mark and https://www.freecodecamp.org/news/a-quick-tale-about-feff-the-invisible-character-cd25cd4630e7/ for more information on this.

This fixes issue #1540.

### DIFF
--- a/lib/WeBWorK/File/Classlist.pm
+++ b/lib/WeBWorK/File/Classlist.pm
@@ -1,13 +1,12 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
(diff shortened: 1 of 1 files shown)


pstaabp:
This works for me.  I realized that there was no error if "Add which users?" is "no users".  I figured it would still read the file and throw an error.  It doesn't though.  I needed to actually add any users to get the error to throw from develop.

pstaabp (approved):

drdrew42 (approved):
Very grateful for this. Good riddance to complaints about 'wide crypt' when instructors upload their own classlist files.

PR #1567, opened 2022-01-28
Show rendering warnings in the library browser and homework set editor.

drgrice1:
To make this work those warnings needed to be added to the json data that is sent by formatRendereredProblem.

I also improve the appearance of these warnings with a bootstrap alert.

I removed the hardcoded warning header in the debug_format.pl.  The warnings that are output already have headers, so that isn't really needed.


drgrice1:
Note that this also adds the `.AuthorComment` style back to the math4.scss file.  That was moved to the pg/htdocs/Problem/problem.scss file which is not loaded on these pages.

pstaabp (approved):
This looks fantastic and clearly much easier than both creating a deprecation flag in the OPL and then assigning problem with that flag.

drdrew42 (approved):
Tested this with some other PRs that throw warnings -- looks good!

PR #1570, opened 2022-01-31
Update the javascript for the gateway quiz module.

drgrice1:
This utilizes data attributes to pass in the timing inforamtion, rather than a defunct form with hidden inputs.  The timer is placed much more appropriately.

Instead of using a native alert for notifications near the end of a test, use bootstrap toasts.  The native alerts interupt a student's
work, forcing the student to close them.  Boostrap toasts sit quietly at the top, and the student can continue to enter answers while the toast is still visible.  The toasts are set to stay open until the next alert message, although the student can close the toast earlier.

Add a quiz auto submission at the end of the quiz that occurs approximately 10 seconds before the end of the grace period.

### DIFF
--- a/htdocs/js/apps/GatewayQuiz/gateway.js
+++ b/htdocs/js/apps/GatewayQuiz/gateway.js
@@ -25,7 +25,7 @@
 	const alertToast = (message, delay = 5000) => {
 		const toastContainer = document.createElement('div');
 		toastContainer.classList.add(
-			'toast-container', 'position-fixed', 'top-0', 'start-50',  'translate-middle-x', 'p-3');
+			'gwAlert', 'toast-container', 'position-fixed', 'top-0', 'start-50',  'translate-middle-x', 'p-3');
 		toastContainer.innerHTML =
 			'<div class="toast bg-white" role="alert" aria-live="assertive" aria-atomic="true">' +
 			'<div class="toast-header">' +
--- a/htdocs/js/apps/GatewayQuiz/gateway.js
+++ b/htdocs/js/apps/GatewayQuiz/gateway.js
@@ -1,114 +1,133 @@
-/***********************************************************
- *
- * Javascript for gateway tests.
- *
- * This file includes the routines allowing navigation
- * within gateway tests, manages the timer, and posts
- * alerts when test time is winding up.
- *
- * The code here relies heavily on the existence of form elements
- * created by GatewayQuiz.pm
- *
- ***********************************************************/
+// Javascript for gateway tests.
+//
+// This file includes the routines allowing navigation within gateway tests, manages the timer, and posts alerts when
+// test time is winding up.
+//
+// The timer code relies on the existence of data attributes for the gwTimer div created by GatewayQuiz.pm.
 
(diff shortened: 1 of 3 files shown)


pstaabp:
This is what I'm seeing and I'm not sure if the functionality has changed recently.  I don't think I've tested quizzes at the end.  I set up a quiz with a time limit of 1 minute. I have unlimited graded submissions.  Others settings are default.

1. I get the nice banner at the top alerting there is 90 second, then 45 seconds left. There is also a banner at the upper right. 
2. I took the quiz, clicked "Grade Test" and the page refreshed with an update on which questions were wrong and right.
3. The timer was down to zero.
4. I was informed that I should click "grade test" now.  (_If I were a student, I would have said "huh", just clicked it._)
5. After about a minute the page refreshed to the "view results" of the quiz. 

My impressions (as someone who doesn't use quizzes that often--perhaps because they are confusing and perhaps have not been paying attention to all the details.)
- The banner and countdown clock are improvements.  I agree the alert is too intrusive.
- I would be very confused as a student at the end of the quiz.  

I think this is the intended functionality to ensure that the scores get submitted.  
- If the student can still submit answers-- and I think this is true, we should have some language on the page that there is still time to submit/change grades. 
- If this isn't the intention, perhaps jump to the "view results" part of the page instead.

drgrice1:
@pstaabp:  I am guessing that you are setting the quiz up to allow multiple submissions (the default probably of an infinite number?).  This and the original javascript are really designed for the single submission case.  I will see if I can improve things a bit for the multiple submission case.

I am betting that by far the most common usage is for one submission.  One common thing I have heard done is two allow for two or three submissions to allow for the "accidental" submission.

As to your comment about a student still being allowed to still submit answers, that depends on the settings.  Furthermore, there is language on the page that states if submissions are still available.

drgrice1:
@pstaabp:  I changed the wording of the dialog that appears after time runs out in the case that multiple submissions are allowed.  This wording should make things less confusing for those types of tests.  It now says `Press "Grade Test" now if you have unsaved changes!`.  If there are not multiple attempts, then it just says `Press "Grade Test" now!` as it did before.  This wording is also used for what is shown in the page as well in the two cases.  I also put the message about the number of attempts remaining in an info alert to make it stand out more.

drgrice1:
@pstaabp:  There is no "view results" part of the page.  So I am not sure what you mean with that comment.  The results are spread throughout the page below each problem when you click "Grade Test" or "Check Test".  The result for the entire test is at the top of the page, so there is no need to "jump" to that.

drgrice1:
I changed this again for multiple submission quizzes.  Now if there is one submission already, then it says nothing about pressing the "Grade Test" button.  It just says you are out of time.  The same is true of the 45 second alert.  It only says "You have less than 45 seconds left!"

I also fixed the $attemptNumber count.  It was not counting the current submission.

pstaabp (approved):
Looks good now.

drdrew42 (approved):
This is great -- tested with time expiring and no submissions. The auto submit worked as described. Other changes look great.

PR #1572, opened 2022-01-31
Update some dependency data (develop pre-2.17)

taniwallach:
When trying to get Docker to build the current develop (which will become WW 2.17 eventually) I ran into some minor issues.

This is intended to help with those which are "dependency" issues not in the Docker files themselves.

1. `npm install` gave a warning about needing to install `popperjs/core@^2.10.2`:
  - @drgrice1 If this should be done differently - please explain how.
  - The recommended solution on https://getbootstrap.com/docs/5.1/getting-started/parcel/#install-bootstrap is to run `npm install @popperjs/core` which essentially adds the line I put in `package.json`.
  - **Update**: @drgrice1 explained that this can be ignored. So the change to add `@popperjs/core` to `package.json` was removed.
```
npm WARN bootstrap@5.1.3 requires a peer of @popperjs/core@^2.10.2 but none is installed. You must install peer dependencies yourself.
```
2. After I got a Docker image built, on startup I got an error message and the startup failed due to a missing module `YAML:XS`, so that is being added to `bin/check_modules.pl`.
  - The install instructions should mention this additional module. It is being added to the Dockerfile to be included in the image.
```
[...]

PR #1571, opened 2022-01-31
Update the docker build for recent changes.

drgrice1:
Run `npm install` from the `pg/htdocs` location.  Also both of the `npm install` commands need to be run with `--unsafe-perm` so that the `prepare` script will run.

Don't try to change ownership and permissions on the now non-existent applets directory.

Fix the bin/addcourse script for the new WWSafe.pm location.  This script was missed in #1533.

Add the Ubuntu package libyaml-libyaml-perl which is needed for the new PGEnvironment.pm module.

Note that to test this you will need to change line 81 of docker-compose.yml to `- WEBWORK2_GIT_URL=https://github.com/drgrice1/webwork2.git` and line 95 to `- WEBWORK2_BRANCH=fix-docker`.

PR #1573, opened 2022-01-31
Docker updates early 2022

taniwallach:
Updates for Docker for develop branch and later WW 2.17.
Depends on the changes in https://github.com/openwebwork/webwork2/pull/1572 which are also included here.

---
To test before merging it is necessary to have the Docker build run using the `Dockerfile` (or the 2 stage files) from the source branch of this PR.

One could start:
```
git remote add tani https://github.com/taniwallach/webwork2.git
git fetch tani
git checkout docker-updates-early-2022 
```
or get the files from the branch in some other manner.

You may want to update the images our build process depends on:
```
docker pull alpine/git
docker pull ubuntu:20.04
```

You then need 2 temporary local changes in `docker-compose.yml`:
Set:
  1. `WEBWORK2_GIT_URL=https://github.com/taniwallach/webwork2.git`
  2. `WEBWORK2_BRANCH=docker-updates-early-2022`
so the build will run with the updated `htdocs/package.json` file.

I recommend testing using the 2 stage build.
```
docker build --tag webwork-base:forWW217 -f DockerfileStage1 .
docker-compose build
```

---

Summary of what was changed:
1. Installs two new core dependencies from Ubuntu.
2. `htdocs/applets` no longer exists, and the use in the `Dockerfile` triggered an error and broke the build process.
3. Use `npm install --unsafe-perm` as in Docker `npm install` is run as root, and without that it would report:
```npm WARN lifecycle webwork.javascript_package_manager@~prepare: cannot run in wd webwork.javascript_package_manager@ npm run generate-css && npm run css-prefix (wd=/opt/webwork/webwork2/htdocs)
```
and
```
npm WARN lifecycle pg.javascript_package_manager@~prepare: cannot run in wd pg.javascript_package_manager@ npm run generate-css (wd=/opt/webwork/pg/htdocs)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -181,6 +181,8 @@ RUN apt-get update \
 	libjson-xs-perl \
 	libjson-maybexs-perl \
 	libcpanel-json-xs-perl \
+	libyaml-libyaml-perl \
+	sass \
 	make \
 	netpbm \
 	patch \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -181,6 +181,8 @@ RUN apt-get update \
 	libjson-xs-perl \
 	libjson-maybexs-perl \
 	libcpanel-json-xs-perl \
+	libyaml-libyaml-perl \
+	sass \
```
drgrice1, line 185: Sass should not be installed here. Sass is installed via npm.

taniwallach: OK

DockerfileStage1
```diff
@@ -100,6 +100,8 @@ RUN apt-get update \
 	libjson-xs-perl \
 	libjson-maybexs-perl \
 	libcpanel-json-xs-perl \
+	libyaml-libyaml-perl \
+	sass \
```
drgrice1, line 104: Sass is also not needed here.

taniwallach: OK.

taniwallach:
@drgrice1 You did mostly the same things in https://github.com/openwebwork/webwork2/pull/1571
As you fixed other things there. I recommend we close this and polish that PR.
I do think we need something like what is in https://github.com/openwebwork/webwork2/pull/1572

drgrice1:
Yeah, I will do that.

Issue #1574, opened 2022-01-31
Check for accessibility issues in the new UI (for WW 2.17)


taniwallach:
We should make an accessibility review of the new UI in develop for obvious accessibility issues which can be fixed before a release.
I'm not a big expert, but the WAVE plugin can find some issues.
@Alex-Jordan may also have some advice as he seems very knowledgeable about the requirements and what really matters to users with disabilities.

taniwallach:
1. The bread crumb of "open" item (ex. "Hmwk Sets Editor" in http://localhost:8080/webwork2/admin/instructor/sets/) does not meet contrast ratio requirements. This seems to also affect such "last" bread-crumb items students can see.
2. WAVE reports some other issues on the set editor page. but as they are staff-only fixing them is probably less critical.
   - ex: The red `*` after "Match on what? (separate multiple IDs with commas)" also has a contrast ratio failure.(Staff only - less critical).

drgrice1:
Note that Firefox has a built in accessibility tool in its developer tools, and I have been using that to look for things already.  Thanks for pointing out the breadcrumb issue, I hadn't noticed that one.  Bootstrap 5 expects the breadcrumb text to be on a lighter background than the color we use here.  I will fix that, and probably the other issues you mentioned.

pstaabp:
This is also addressed in #1603

taniwallach:
It looks like quite a bit was done on this. Thanks @drgrice1 and @pstaabp.
I think we should close this issue.
Maybe a collection of remaining issues to address in the future can be collected in a new issue.

pstaabp:
Since this is a WeBWorK issue, let's close this.  We can open a PG issue for accessibility instead.

PR #1575, opened 2022-01-31
Fix a few accessibility issues.

drgrice1:
These are mainly foreground/background contrast ratios issues.

Disabling gradients fixes some issues with contrast of buttons.  That is probably why bootstrap disables this by default.

See #1574.  This fixes all of the issues currently listed there.  The Firefox accessibility tool shows no other issues on the student pages (other than the check boxes on the new quiz page fixed in another pull request).  On the other hand, there are quite a few on instructor only pages.

### DIFF
--- a/htdocs/themes/math4/bootstrap.scss
+++ b/htdocs/themes/math4/bootstrap.scss
@@ -19,7 +19,6 @@
 
 // Enable shadows and gradients.  These are disabled by default.
 $enable-shadows: true;
-$enable-gradients: true;
 
 // Use a smaller grid gutter width.  The default is 1.5rem.
 $grid-gutter-width: 1rem;
@@ -31,6 +30,10 @@ $headings-font-weight: 600;
 // Links
 $link-decoration: none;
 $link-hover-decoration: underline;
+
+// Make breadcrumb dividers and active items a bit darker.
+$breadcrumb-divider-color: #495057;
+$breadcrumb-active-color: #495057;
 
 @import "./theme-colors";
 
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -92,7 +92,7 @@
 
 	// Turn help boxes into popovers
 	document.querySelectorAll('.help-popup').forEach((popover) => {
-		new bootstrap.Popover(popover, {trigger: 'hover'});
+		new bootstrap.Popover(popover, {trigger: 'hover focus'});
 	});
 
 	// Problem page popovers
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -33,7 +33,7 @@ table caption {
 }
 
 .required-field {
-	color: red;
+	color: #dc3545;
 }
 
 /* Screen Reader */
@@ -89,7 +89,7 @@ table caption {
 	display: flex;
 	align-items: center;
 	padding: 5px 0;
-	background-color: var(--ww-logo-background-color, #104aad);
+	background-color: var(--ww-logo-background-color, #104aac);
 
 	img {
 		max-height: 57px;
@@ -186,7 +186,7 @@ table caption {
 		color: black;
 		font-weight: bold;
 		font-size: 0.9rem;
-		opacity: 0.5;
+		opacity: 0.6;
 	}
 
 	.info-box {
@@ -372,6 +372,7 @@ h2.page-title {
 .Warnings {
 	code {
 		white-space: normal;
+		color: inherit;
 	}
 }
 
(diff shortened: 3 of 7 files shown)


pstaabp (approved):
Overall, looks good. However, I didn't run any accessibility assessment tools on this and perhaps we should do a thorough job before 2.17 ships.  I'd also like to hear from @Alex-Jordan about this thoughts.

Alex-Jordan:
I skimmed through this.

A third of it is something I immediately understood and is good.
A third of it is something I immediately understood and is neutral. (Changing color #104aad to #104aac)?
A third is something I don't immediately understand, but I trust Glenn that it's good.

I gather from one of these recent threads that you are using automated accessibility tools to check on things. That will cover a lot. Some things that could be missed by automated testing are color contrast issues and issues relating to click events. But it's clear you are aware of such things.

drgrice1:
@Alex-Jordan:  The webwork logo color is actually set in the _theme_colors.scss file, and is set to be 14% darker than the info color, which is `#1a67ea`.  The value that you are looking at in math4.scss is only a fallback value that is used if the `--ww-logo-background-color` is not set.  Since that will be set, it really isn't used, but I like to keep it in sync with the computed value.  Before I disabled gradients I attempted to fix the contrast issue in another way by changing the info color and the percentage that the logo color was darker than that.  After that change it changed the computed value to `#104aac`.  However, I forgot to change that back to `#104aad` when I put the computed value back.

drgrice1:
I updated the fallback value, so the computed and fallback values are the same again, and they are what they were before.

Also, I can explain every change, and what it fixes for accessibility.  Just specify the ones that don't make sense if you want to know.

drgrice1:
In addition to using Firefox's accessibility checker, I also installed the Wave extension on Google Chrome as suggested by @taniwallach.  Furthermore, I consulted some online documentation on things that I have found.  One advantage of using Bootstrap 5 is that their documentation is quite good on what is needed to make the usage of their library accessible.

Alex-Jordan (approved):
Looks good.

Prior to this PR, it seems there was at least one change to HTML structure in the side navigation. We should plan to do some HTML validation and accessibility checking at some point. (I know that a lot has been checked along the way with  the Wave extension and other things.)

I dropped a page into an HTML validator just now and I get two errors and many warnings. They were probably already there in 2.16. But we can also look at cleaning those up.

drgrice1:
What are the errors and warnings that you are seeing?

Alex-Jordan:
I used a Homework Sets page. There are 3 errors. One is the `a` inside the `ul` that we already talked about and this PR changes. The other two are indeed old things. One is:

```
A document must not include both a meta element with an http-equiv attribute whose value is content-type, and a meta element with a charset attribute.
```

The other is saying that `summary` attributes on tables are obsolete. Not sure why it's an error instead of a warning.

Then there are warnings, mostly about not needing attributes `charset="UTF-8"` and `type="text/css"` on some things. One about not needing `role="navigation"` on a `nav`.

drgrice1:
The meta tags probably need to be reviewed.

As to the table summaries, that is good news to me.  I always wanted to eliminate those summaries.  As to deprecated attributes on tables, PG adds many of them in problems also.  cellpadding, width, border, etc.  All deprecated.  Those are things that are to be done with css.

I have removed most of the `type="text/css"` attributes.  Where did I miss some?  Or perhaps those are still in pull requests to come.

I will look into the `role="navigation"` on a nav warning.  Not sure where that is.  I don't think I added that, but maybe I changed a div to a nav and forgot to remove the role.

drgrice1:
I think this pull request should be merged so that these things are in.  The other things can be added later, and I think some of them are already dealt with in other pull requests (like the link type attributes, same goes for script type attributes).

Alex-Jordan:
It effectively has 3 approvals and I will merge it.

Alex-Jordan:
>  Or perhaps those are still in pull requests to come.

That could be.

For HTML validation, I just saved a page to a .html file and then uploaded into:
https://validator.w3.org/nu/

There is this, which may be worth trying out:
https://addons.mozilla.org/en-US/firefox/addon/html-validator/

drgrice1:
> ```
> A document must not include both a meta element with an http-equiv attribute whose value is content-type, and a meta element with a charset attribute.
> ```

I used the https://validator.w3.org site to validate numerous pages in webwork2, and I don't get this.  Also, it seems that we do have the meta element with a charset attribute in both the templates and the formats used by html2xml.  We have a meta element with the http-equiv attribute in the templates (but not in the formats), but the value needs to be updated.  The `X-UA-Compatible` value for that isn't valid anymore, and the `content="IE=Edge"` attribute needs to be deleted or changed.  Although https://validator.w3.org/nu does not show anything for this.

Where exactly are you getting this?

PR #1576, opened 2022-02-01
Rework the path that is displayed in the breadcrumb for gateway quizzes.

drgrice1:
Instead of showing 'Home / courseId / setId,v?' in the breadcrumb, show 'Home / courseId / setId / v?'.  The `setId` in this is a link back to the new gateway quiz problem set page that lists the quiz versions and such.

Also, replace all underscores with spaces in the names displayed in the breadcrumb and don't convert those names to lowercase.  This just looks better.

For all pages, replace underscores in a path when it is used for a page title.  This used to be done this way, but for some reason it was commented out.  There is no situation where an underscore needs to be displayed in the page title.


pstaabp (approved):
Looks good and as advertised.

drdrew42 (approved):
Works as advertised 👍

Issue #1539, opened 2022-01-11
Feature Request: Enable spellcheck in CodeMirror


somiaj:
CodeMirror can use the browser's spell checker in the code editor for problems and achievements. The following patch has enabled spell check for me here and I haven't noticed any issues so far.

```diff
diff --git a/htdocs/js/apps/AchievementEditor/achievementeditor.js b/htdocs/js/apps/AchievementEditor/achievementeditor.js
--- a/htdocs/js/apps/AchievementEditor/achievementeditor.js
+++ b/htdocs/js/apps/AchievementEditor/achievementeditor.js
@@ -10,6 +10,8 @@
                                {Tab: function(cm) {cm.execCommand('insertSoftTab')}},
                                highlightSelectionMatches: true,
                                matchBrackets: true,
+                               inputStyle: 'contenteditable',
+                               spellcheck: true,
 
                        });
                cm.setSize("100%", 400);
diff --git a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
--- a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
@@ -11,6 +11,8 @@
                                {Tab: function(cm) {cm.execCommand('insertSoftTab')}},
                                highlightSelectionMatches: true,
                                matchBrackets: true,
+                               inputStyle: 'contenteditable',
+                               spellcheck: true,
 
                        });
                cm.setSize("100%", 400);
```

pstaabp:
@somiaj Have you tried enabling this?  This seems like a good idea, however since this is mainly for problem editing, I'm guessing it can't handle code from text blocks, and spell checking in code can be annoying.

somiaj:
@pstaabp I am currently using this and it works well. I like the additional check for problem text and just ignore the underlines in rest of the code. I don't find it annoying, but understand if others do. In firefox and chromium (maybe the same for other browsers) the right click in the editor div provides an enable/disable spell check option, so the browser spell checker can be disabled if someone finds it annoying. So once this is enabled in codemirror, the user can turn it on/off in their browser as desired.

pstaabp:
This is also being addressed in #1608

PR #1577, opened 2022-02-04
Codemirror editor improvements.

somiaj:
The codemirror editor has addons to extend functionality, such as providing find and replace. These were meant to be enabled, but due to using local copies (and missing loading the corresponding css files) were only partly functional.

This removes the local copy of the addons, and uses the addons from the codemirror source directly and loads the required CSS files for the addons to fully function. It also creates a `htdocs/css/codemirror.css` file needed for some addons that don't include the required CSS.

In addition this enables the ability to allow browsers to spellcheck the contents of the codemirror editor as mentioned in #1539.

This was initiated due to trying to improve syntax highlighting as mentioned in <https://github.com/openwebwork/pg/issues/649>. It turns out that the syntax highlighter was already configured to show trailing white spaces at the end of lines in PGML blocks, but due to the barely functioning addons, just wasn't working. This now underlines trailing whitespace in PGML blocks without the need to load the trailing whitespace module mentioned in the issue.


drgrice1 (changes requested):
I haven't looked closely at this at all yet, but I see one change that should be made already.  Move `htdocs/css/codemirror.css` to `htdocs/js/apps/PGCodeMirror`.

somiaj:
Note: See https://codemirror.net/demo/search.html for the key bindings for the search and find and replace features.

somiaj:
@drgrice1 done. I just wasn't sure if css should go in the js directory or not. Thanks.

drgrice1:
Sorry, I should explain the expected layout and the reason for my request.  The `htdocs/css` directory is for generic css that is not associated with javascript.  The css that is associated with javascript should go with it in the `htdocs/js/apps` directory.  That keeps the associated code and style together, and makes it easier to find.

Thanks for moving it.

drgrice1:
Actually, I want to delete the css directory.  All css should either be associated with javascript or be part of the theme.  The `rtl.css` was added by @taniwallach, and needs to be deleted.  Right-to-left styles should be done correctly, and should be in math4.scss using the `[dir="rtl"]` selector.

somiaj:
Note the commit comment doesn't list the actual path, a very minor inconstancy I can try to update if you like.

drgrice1:
That is fine.  No need.

somiaj:
Opps. I had done some local editing in testing the trailing whitespace highlight and overlooked that the required CSS was missing to highlight trailing whitespace in PGML blocks. Added in previous commit.

somiaj:
Note: The ability to resize the codemirror div (last commit) doesn't work with safari in my testing, It appears to be something with the way safari handles scroll bars, and you can't click on the resize icon in the bottom right corner which is under the scroll bar.

The resize option works well in firefox and chrome, is supported by codemirror, and I think is worth enabling even if safari users can't use it. All other aspects of the editor seem to work in safari and this doesn't cause any issues I have found.

pstaabp:
A few comments here:
- I am also seeing the same issue in Safari about resizing, but not that big a deal. I don't see it as an issue on Codemirror github. 
- The search/replace works great.
- Is there a way to get jump to line on MacOS? I tried to test that.
- I'm not seeing in spellchecking on Safari.  It's working in Chrome.

somiaj:
@pstaabp The jump to line is another addon. This wasn't one of the addons that was originally meant to be enabled, so I didn't add it. If you think it is worthwhile I can also load that addon (that is why it wasn't working).

pstaabp:
> The jump to line is another addon.

I guess I was playing with a code mirror demo that had both enabled.  Most pg files are small enough that I don't think this is needed. 

About Safari, it's fine.  If we can verify that spell checking is not available in Safari, we can make a note in the release notes.

pstaabp (approved):
Looks like a nice addition.  We should make sure this get documentation/in the release notes.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -2077,8 +2077,19 @@ sub output_JS{
...
 		print CGI::start_script({src=>"$site_url/node_modules/codemirror/lib/codemirror.js"}), CGI::end_script();
-		print CGI::start_script({src=>"$site_url/js/apps/PGCodeMirror/PGaddons.js"}), CGI::end_script();
+		print CGI::start_script({src=>"$site_url/node_modules/codemirror/addon/dialog/dialog.js"}), CGI::end_script();
+		print CGI::start_script({src=>"$site_url/node_modules/codemirror/addon/search/search.js"}), CGI::end_script();
+		print CGI::start_script({src=>"$site_url/node_modules/codemirror/addon/search/searchcursor.js"}), CGI::end_script();
+		print CGI::start_script({src=>"$site_url/node_modules/codemirror/addon/search/matchesonscrollbar.js"}), CGI::end_script();
+		print CGI::start_script({src=>"$site_url/node_modules/codemirror/addon/search/match-highlighter.js"}), CGI::end_script();
+		print CGI::start_script({src=>"$site_url/node_modules/codemirror/addon/scroll/annotatescrollbar.js"}), CGI::end_script();
+		print CGI::start_script({src=>"$site_url/node_modules/codemirror/addon/edit/matchbrackets.js"}), CGI::end_script();
+
```
drgrice1, line 2092: Both here and in lib/WeBWorK/ContentGenerator/Instructor/AchievementEditor.pm there is some clean up needed.

First, do not add `type="text/css"` to css links.  That is not needed in modern html, and its usage is discouraged.

Now for css links, use `CGI::Link` instead of using `qq` and piecing this together.  I don't think that the previous WeBWorK developers knew about the `CGI::Link` method (I just recently found it myself), and that is why it was not used.  See the merge conflicts that this branch has to see the correct usage of that.

For the script tags, don't use `print CGI::start_script({ ... }), CGI::end_script();`.  Instead use `print CGI::script({ ... }, '');`.  It is much more succinct.  Also see the merge conflicts this branch has for the correct usage here.

If you can run perltidy on a code section in a file, then please do that to the `output_JS` methods here and in the other file.

drgrice1 (commented):
There is a CodeMirror css style override in `htdocs/themes/math4/math4.scss`.  Either add that to your new `htdocs/js/apps/PGCodeMirror/codemirror.css` file, or instead of adding that file put those styles into `htdocs/themes/math4/math4.scss`.

Adding the spellcheck option is **not** a good idea.  You may be accustomed to ignoring its suggestions, but if there is a lot of code the suggestions are quite pervasive.  Don't assume that others are accustomed to this.  Please remove this option.  One way that this could be done is to add a checkbox below the CodeMirror textarea that when checked activates spell checking, and is not checked by default.  It could also be made so that its status is saved, and anytime the page loads it has the status it had before (check if it was, unchecked if it wasn't).  There are other options that could also be done this way.

drgrice1 (approved):
This looks good now.

The spell checking option would be nice.  I think that there are other things that could be made optional too.  For example, one could be allowed to change the Codemirror theme.  Many prefer dark backgrounds, rather than white for example.  But these things can wait for another pull request.

PR #1578, opened 2022-02-05
Rework how reducedScoring is computed and stored.

somiaj:
This is paired with removing reducedScoring from PG. Instead webwork should handle taking the raw score from PG and adjusting the score based on reducedScoring options.

This adds a method `apply_reduced_scoring_adjustment` to ProblemUtils. This method needs to be run after the raw score is returned from PG, but before writing it to the database to apply any reduced scoring adjustments to the raw score. The method updates the current problem right before writing it to the database.

This also changes how the columns `status` and `sub_status` in the `problem_user` table are used. `status` still contains the actual adjusted grade used by webwork, while `sub_status` now stores the raw non adjusted grade. Provided the reducedScoringValue has not changed, it is possible to use the previous `status` and `sub_status` to recover the portion of the problem that was done before the reduced scoring date and the portion that was done after, so only the part of the problem done after the reduced scoring date is adjusted.

This needs to be merged with <https://github.com/openwebwork/pg/pull/650> and stems from the discussion <https://github.com/openwebwork/pg/issues/648>.

Now that `sub_status` contains the raw score, it is now possible to address some of the concerns mentioned by @Alex-Jordan for places that should check the raw score and not the adjusted score when determining what to due (such as conditional release or indicating that a problem has been completed).

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
@@ -134,13 +134,16 @@ sub process_and_log_answer{
 
 			# store state in DB if it makes sense
 			if ($will{recordAnswers}) {
-				$problem->status($pg->{state}->{recorded_score});
-				$problem->sub_status($pg->{state}->{sub_recorded_score});
+				# Reduced scoring adjustments
+				my $new_score = compute_reduced_score($self, $problem, $set, $pg->{state}->{recorded_score});
+
+				$problem->status($new_score);
+				$problem->sub_status($pg->{state}->{recorded_score});
 				$problem->attempted(1);
 				$problem->num_correct($pg->{state}->{num_of_correct_ans});
 				$problem->num_incorrect($pg->{state}->{num_of_incorrect_ans});
-				$pureProblem->status($pg->{state}->{recorded_score});
-				$pureProblem->sub_status($pg->{state}->{sub_recorded_score});
+				$pureProblem->status($new_score);
+				$pureProblem->sub_status($pg->{state}->{recorded_score});
 				$pureProblem->attempted(1);
 				$pureProblem->num_correct($pg->{state}->{num_of_correct_ans});
 				$pureProblem->num_incorrect($pg->{state}->{num_of_incorrect_ans});
@@ -673,6 +676,36 @@ Comment:    $comment
 
 
     return "";
+}
+
+# Determines if a set is in the reduced scoring period and applies
+# the reduced scoring adjustment to the portion of the score earned
+# during the reduced scoring period.
+sub compute_reduced_score {
+	my ($self, $problem, $set, $new_status) = @_;
+
+	my $r = $self->r;
+	my $reducedScoringEnabled = $r->{ce}->{pg}{ansEvalDefaults}{enableReducedScoring};
+	my $reducedScoringValue = $r->{ce}->{pg}{ansEvalDefaults}{reducedScoringValue};
+
+	# If no adjustments need to be applied, return the full score.
+	if (! $reducedScoringEnabled
+			|| ! $set->enable_reduced_scoring
+			|| ! $set->reduced_scoring_date
+			|| $set->reduced_scoring_date == $set->due_date
+			|| time < $set->reduced_scoring_date
+			|| $reducedScoringValue == 1
+	) {
+		return $new_status;
+	}
+
+	# Invert the reduced scoring computation to get the portion of the
+	# grade earned before the reduced scoring date, then return the
+	# adjusted score by applying the reduced scoring adjustment to the
+	# portion done after the reduced scoring date.
+	my $pre_reduced_score = $problem->sub_status - ($problem->sub_status
+		- $problem->status) / (1 - $reducedScoringValue);
+	return $pre_reduced_score + $reducedScoringValue * ($new_status - $pre_reduced_score);
 }
 
 1;
--- a/lib/WeBWorK/AchievementEvaluator.pm
+++ b/lib/WeBWorK/AchievementEvaluator.pm
@@ -105,7 +105,6 @@ sub checkForAchievements {
     # $pg->{result} reflects the current submission, $pg->{state} holds the best result
     # close the unlimited achievement points loophole by only using the current result!
     $problem->status($pg->{result}->{score});
-    $problem->sub_status($pg->{state}->{sub_recorded_score});
     $problem->attempted(1);
     $problem->num_correct($pg->{state}->{num_of_correct_ans});
     $problem->num_incorrect($pg->{state}->{num_of_incorrect_ans});
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1456,11 +1456,17 @@ sub body {
 
 			# next, store the state in the database if that makes sense
 			if ($submitAnswers && $will{recordAnswers}) {
-				$problems[$i]->status(wwRound(2,$pg_results[$i]->{state}->{recorded_score}));
+				# Reduced scoring adjustments
+				my $new_score = WeBWorK::ContentGenerator::ProblemUtil::ProblemUtil::compute_reduced_score(
+						$self, $pureProblem, $set, $pg_results[$i]->{state}->{recorded_score});
+
+				$problems[$i]->status(wwRound(2,$new_score));
+				$problems[$i]->sub_status($pg_results[$i]->{state}->{recorded_score});
 				$problems[$i]->attempted(1);
 				$problems[$i]->num_correct($pg_results[$i]->{state}->{num_of_correct_ans});
 				$problems[$i]->num_incorrect($pg_results[$i]->{state}->{num_of_incorrect_ans});
-				$pureProblem->status(wwRound(2,$pg_results[$i]->{state}->{recorded_score}));
+				$pureProblem->status(wwRound(2,$new_score));
+				$pureProblem->sub_status($pg_results[$i]->{state}->{recorded_score});
 				$pureProblem->attempted(1);
 				$pureProblem->num_correct($pg_results[$i]->{state}->{num_of_correct_ans});
 				$pureProblem->num_incorrect($pg_results[$i]->{state}->{num_of_incorrect_ans});
@@ -1746,8 +1752,9 @@ sub body {
 			my $pScore = 0;
 			my $numParts = 0;
 			if (ref($pg)) {  # then we have a pg object
-				###
-				$pScore = $pg->{state}->{recorded_score};
+				# Reduced scoring adjustments
+				$pScore = WeBWorK::ContentGenerator::ProblemUtil::ProblemUtil::compute_reduced_score(
+						$self, $problems[$i], $set, $pg->{state}->{recorded_score});
 				$probStatus[$i] = $pScore;
 				$numParts = 1;
 				###
--- a/lib/WeBWorK/DB/Record/UserProblem.pm
+++ b/lib/WeBWorK/DB/Record/UserProblem.pm
@@ -43,14 +43,14 @@ BEGIN {
 		# periodic re-randomization number of attempts for the current seed
 		prCount => {type => "INT"},
 		problem_seed  => { type=>"INT" },
-		status        => { type=>"FLOAT" },
+		status        => { type=>"FLOAT" },   # The adjusted problem score
 		attempted     => { type=>"INT" },
 		last_answer   => { type=>"TEXT" },
 		num_correct   => { type=>"INT" },
 		num_incorrect => { type=>"INT" },
 		att_to_open_children => { type=>"INT" },
 		counts_parent_grade => { type=>"INT" },
-		sub_status    => { type=>"FLOAT" },    # A subsidiary status used to implement the reduced scoring period
+		sub_status    => { type=>"FLOAT" },    # The raw problem score before any adjustments
 		# a field for flags which need to be set
 		flags => { type=>"TEXT" },
 	);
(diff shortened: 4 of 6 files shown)


somiaj:
With changing how `sub_status` is used in the database, there could be a possible upgrade issue from 2.16 for any current courses. If `sub_status < status`, it should be safe to assume this is an upgrade, so we can recreate the new `sub_status` from this (provided the instructor hasn't changed the `reducedScoringValue`):

```
sub_status += (status - sub_status) / reducedScoringValue
```

If `sub_status == status` then this means the problem was done outside of the reduced scoring period, and these two fields will be the same both before the upgrade and after the upgrade, so there is nothing to be done.

Thus to ensure a safe upgrade from <= 2.16, we should check if `sub_status < status` and then update `sub_status` if needed. I think this should be done in the database module when the data is retrieved from the database.

somiaj:
I was thinking it is possible at the time an instructor changes the `reducedScoringValue` to loop through the `problem_user` table and update the `status` field based on `sub_status`, the old value, and the new value. This is probably a good idea to avoid any strange effects of changing this value in the math used to keep track of the full portion and reduced portion of the grade. Though it will be slow, as that table could be fairly big. I suspect this won't happen that much and then it would be like the weight, if you change it, it effects all assignments.

As for what to do if an instructor enables/disables reduced scoring. I don't see any way to nicely deal with being able to turn this on/off/on and keep track of the data. My initial thought is leave it alone, and in this case students could go increase their grade by earning full credit on any open assignments, but any past assignments will remain as they are. Another option is to update the `status` field if it goes from on to off, but this is one way, there would be no way to get the data back after that. Yet another possibility could be leave the table alone, but make the various places webwork reports the grade use the `sub_status` field instead of `status` if reduced scoring is disabled. This way if it is enabled again, all the data is still there. This still has the same issue is any open assignments students are working on will loose any past reduced scoring data once a new grade is submitted and the `status` field gets updated.

somiaj:
In testing this just noticed another issue with the manual problem grader. It is using the raw scores from PG, so when you first submit a quiz in the reduced scoring date, everything looks like it is 100%, but if you leave the quiz and come back, the reduced scoring has been correctly applied.

drgrice1:
This is certainly going to need careful scrutiny before it goes in.

I am not sure that the way that the `apply_reduced_scoring_adjustment` method changes the database records is a good approach.

I haven't looked particularly close at this yet though.

somiaj:
@drgrice1 I think it will be useful to save the raw grade in the database, which currently isn't done. To try to clarify the change.

Current method: `status` = adjusted grade, `sub_status` = grade earned before the reduced scoring date. This was so the reduced scoring only applied to anything earned after the reduced scoring date.

Suggested change: `status` = adjusted grade (stays the same), `sub_status` = raw score. I then recover the grade earned before the reduced scoring date (vs storing it). Works provided reducedScoringValue doesn't change.

So the change just stores the raw score so it can be used other places which will be helpful to determine conditional release or when a problem is completed. It could be possible to compute the raw score from the curent `sub_status` and `reducedScoringValue`, I just think overall having the raw score being stored will be more beneficial than scoring just the portion earned before the reduced scoring date.

drgrice1:
This pull request will need to be completely redone.  The current formula that you have implemented is in no way correct.  Basically, it is possible with this formula for a student to always increase their grade as long as they have submissions available.  The limit of that possible grade is 1 with this formula.  That limit can not be achieved, but in just a few iterations it can be made pretty close to that, and can easily exceed the reduced credit the student is currently allowed.

For a simple test to see that this is the case, create an assignment that only has one problem.  To make this easy use the attached problem [AlwaysCorrect.pg.txt](https://github.com/openwebwork/webwork2/files/8041760/AlwaysCorrect.pg.txt).  This problem is always correct not matter what you enter (even if you enter nothing).

Now enable reduced scoring for the course and the set, and set the open date and reduced scoring date to a date before today, and the due date to a date after today.  So the students will all have a status and sub_status of 0, and will be in the reduced scoring period from the start.  Also set the reduced scoring value to 0.75 (the default).

Now go to this problem in the assignment and submit any answer.
Your implementation will set the status to 0.75 and the sub_status to 1.
Now submit any answer again.  Now the status is 0.91667 and the sub_status of course is still 1 and it will never change again.
Submit again.  Now the status is 0.972222.
Submit again.  Status is 0.990741.
Submit again.  Status is 0.996914.
Note that these are the values seen in the database for this user, but you can see the rounded results on the grades page (or elsewhere).

Of course if you start this test all over with the current develop branch and the current reduced scoring implementation, you see that after the first attempt the status is set to 0.75 and never changes again, and the sub_status never changes from 0.

somiaj:
Thanks for catching this. I was testing with 50% reduced scoring, and it all worked, I needed to divide by (1 - reducedScoringValue), not reducedScoringValue, but at 50% it didn't matter. I'm not able to increase the score now by just submitting again and again.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
```diff
@@ -675,4 +675,81 @@ Comment:    $comment
...
+	# Return now to avoid division by zero.
+	if ($reducedScoringValue == 1) {
+		$problem->sub_status($problem->status);
+		return;
+	}
+
+	# Split old_status into the full part earned before the reduced scoring
+	# date and the adjusted part earned after the reduced scoring date.
+	my $full_part = $oldProblem->sub_status - ($oldProblem->sub_status
+		- $oldProblem->status) / (1 - $reducedScoringValue);
```
drgrice1, line 730: Use more meaningful variable names, and make your comments also more meaningful.  `$full_part` is not only oxymoronic, but has no meaning.  Use something like `$pre_reduced_period_score` or for something shorter just `$pre_reduced_score`.

I have verified that this formula is correct now.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
```diff
@@ -164,6 +162,8 @@ sub process_and_log_answer{
 				    $db->putGlobalProblem($globalProblem);
 				}
 
+				# Apply reduced scoring adjustements.
+				apply_reduced_scoring_adjustment($self, $pureProblem, $set, 0);
```
drgrice1, line 166: Move this up to right before line 137 and completely rework the `apply_reduced_scoring_adjustment` method.  There is no reason at all for that method to access the database.

Let me outline the behavior for you as you have it.  First the `$pureProblem` is obtained from the database on line 79.  That has the previous `status` and `sub_status` in it.  Then on line 137 the previous status is overwritten by the result from the current attempt.  Then this method is called and it again queries the database to get that same previous `status` and `sub_status` because those were just overwritten.

If you move this method up to before line 137 as I said then pass in the `$pg->{state}->{recorded_score}` then you don't need a database query at all.  Of course you will need to do the same reordering in the GatewayQuiz.pm module.  With this you also won't need to worry about if the problem is versioned or not.  The method will need to be redesigned to not actually modify the problem at all (and probably not even pass it in, just pass in only the status and sub_status), and then the method should return the new values for those.  After that the `$problem` and `$pureProblem` values would be set from the return values.  This also makes the usage on line 1762 more natural.

Note that you are also incorrectly using `getMergedProblem` to re-obtain this information.  The result of `getUserProblem` has the `status` and `sub_status` and those are never in global set values.  You only need to use `getMergedProblem` if you need to obtain the problem where unset columns in the user problem are filled in with the global values.  So that is needless overhead in addition to not needing to query the database anyway.

drgrice1 (commented):
I have left some comments.  This is just a start on what will need to be done here to get this ready to go in.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
```diff
@@ -675,4 +675,81 @@ Comment:    $comment
...
+
+	# Split old_status into the full part earned before the reduced scoring
+	# date and the adjusted part earned after the reduced scoring date.
+	my $full_part = $oldProblem->sub_status - ($oldProblem->sub_status
+		- $oldProblem->status) / (1 - $reducedScoringValue);
+	
+	# If the reducedScoringValue has changed, the above computation will not
+	# give the correct value. Be as lenient as possible in this case by giving
+	# full credit if the value is unreasonable.
+	$full_part = $oldProblem->sub_status if ($full_part > $oldProblem->sub_status || $full_part < 0);
```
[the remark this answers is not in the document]
somiaj: This isn't protection against a bad reducedScoringValue, it is protection against the value being changed while there are sets in the reduced scoring period.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
```diff
@@ -675,4 +675,81 @@ Comment:    $comment
...
+			|| $set->reduced_scoring_date == $set->due_date
+			|| time < $set->reduced_scoring_date
+	) {
+		# There are already checks before this, but to be safe, use
+		# the greater of the two scores: status = new, sub_status = old.
+		if ($problem->status > $problem->sub_status) {
+			$problem->sub_status($problem->status);
+		} else {
+			$problem->status($problem->sub_status);
+		}
```
drgrice1, line 711: This comment above is completely incorrect.  There are no previous checks for this, and if this code isn't done the `sub_status` would never be saved.  Furthermore this is convoluted.  Prior to the reduced scoring period or if reduced scoring is not enabled bot the `sub_status` and the `status` should be set to the new pg result score.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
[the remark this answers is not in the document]
drgrice1: Ah, right.  Let me amend my comment.  This is just messed up.  It should not be possible unless your code is not working.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
somiaj: I see. In doing some more tests with my much revised version, these tests should be done elsewhere, and I was over complicating this trying to deal with malformed values here.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
somiaj: I got too focused on getting this to work I overlooked applying this reduction sooner. I didn't like the database calls, but at the time I couldn't think of another way to do it (since I was focused on where I was applying the reduction). I have made an updated version which I think cleans a lot of stuff up. Thanks again for your feedback.

lib/WeBWorK/ContentGenerator/ProblemUtil/ProblemUtil.pm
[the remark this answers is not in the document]
somiaj: This maybe the bane of this approach. I think it would be helpful to be able to save the raw score in the database in such a way that won't require a new column. Though it is only possible to invert the reduced scoring calculation to get the correct grade earned before the reduced scoring period if the reduced scoring value isn't changed by an instructor. As I thought about it more, if the grade is manually changed with the manual grader it will also have this same problem that it won't be possible to correctly invert the computation. Yet again I was over thinking it and trying to prepare for every possibility.

somiaj:
I have drastically cleaned things up and changed how and when the reduced scoring is applied thanks to @drgrice1's comments. My tests seem to be positive so far, but I do think the major down side of this approach is that the reduced scoring computation may not be correctly inverted if the reduced scoring value is changed or grades are manually changed.

So this leaves a few key questions:

Is it important to save the raw score in the database? There have been various reasons to support this mentioned in <https://github.com/openwebwork/pg/issues/648#issuecomment-1029327132>.

If it is important to save the raw score, should another column be added to the database? This approach doesn't add another column, but there will be some edge cases that the before/after reduced scoring portion can no longer be correctly computed.

There might also be other options I have not thought of to store the raw score in the database without adding more columns.

At this point is won't take long to change this to how reduced scoring was originally computed, I was just trying to explore the possibility of using `sub_status` for the raw score.

drgrice1:
Okay, so here is your next rework.  This approach of saving the pg score in the database is just not going to reliably work.  Adding another column to the database simply isn't needed either.

Here is what is needed.  This should be redone to simply move the pg code that performed the reduced scoring to the webwork2 side, but using the same formulas as that code and using the same meaning of the values in the database.  The thing is that that will actually work to fix all issues that were brought up in #648.  It will take a bit more care in some cases, but is the only reliable way to have the score before the reduced scoring period begins.  Furthermore, manual grading of problems will not mess it up.  Instead of unreliably trying to recover the score before the reduced scoring period begins from the currently saved pg result score and reduced score, you recover the last pg result score from the currently save reduced score.  That is much more reliable, and failure is not as critical as the places where this is needed as requested by @Alex-Jordan in #648 are only UI display things.  This is also much simpler for those issues because you really don't need the actual pg result score, but you just need to know how the saved score compares to the maximum possible score in the reduced scoring period, and this is easily computed with the current information in the database.

Perhaps you should have heeded what @pstaabp said in https://github.com/openwebwork/pg/issues/648#issuecomment-1030060985.  He told me to hold off on this, and you probably should have too.

drgrice1:
I think I need to reword the beginning of that last comment.  Instead of calling it your next rework, I think what needs to be said here is put this pull request on hold a bit until further discussion can be had.

Alex-Jordan:
Some references above are pointing to an old PR in webwork2. The pg issue is here:
https://github.com/openwebwork/pg/issues/648

I am about to go add an item to the list.

drgrice1:
Thanks @Alex-Jordan.  I forgot to prefix those with the pg repository.  That works if the issue is from the same repository, but that one is not.

somiaj:
This approach won't work in the long run. Closing.

PR #1579, opened 2022-02-06
Better consistancy and configurablity of gateway quizzes answer date.

somiaj:
When a test version is created, its answer date is set to be equal to the close date plus the difference between the templates close date and answer date. So when configuring if and when scores and answers are available, this can depend on either the template answer date or the version answer date. Which date was used was inconsistent (GatewayQuiz module used the template date while Grades module used the version date), plus there was some additional logic for when the answer date was the same as the due date.

This adds a new option to hide_work and hide_score, 'BeforeVersionAnswerDate', so it can be configured which answer date to use. Removes the condition of if the answer date and due date are equal. And makes use of the template or version answer date consistent with the quiz settings.

The current and desired behavior was discussed in #740 and at the time no consensus as to what the behavior should be was made. Overall this doesn't try to change the behavior in a way that should require a consensus (I hope), but instead makes the current behavior more consistent and configurable.

This does slightly alter the database by adding another ENUM option, which makes it not work so nicely with current courses, but it works fine with any newly created courses. I am unsure of any policy on how changes to the database can take place, and my assumption is this is probably why the strange behavior with set answer date = close date was being used. In this case I think it is worth adding another ENUM option to help both clarify behavior and to make it easier to configure than the voodoo date logic.

Also adds some more help text to the instructor configuration page to help clarify how the settings affect quiz behavior.


drgrice1 (changes requested):
Instead of creating a new enum value that has the behavior of an old enum value, and changing the behavior of an old enum, create a new enum value with the new behavior, and leave the behavior of the old enum value the same.  So create a new enum value named something like `BeforeGlobalAnswerDate`, and make that do what you now have `BeforeAnswerDate` doing, and leave `BeforeAnswerDate` doing what it was.  The enum names in the database don't have to be as meaningful as those are under the hood sort of things.  Then you change the names and explanation of things in the ui.  That way, previously created quizzes in old courses will still work as expected.

somiaj:
@dgrice1 I can change that. Though since the current use is inconsistent, what is the current behavior? Right now the GatewayQuizzes.pm module is using  `BeforeAnswerDate` with the template answer date, which is how this change still works. It was only the Grades.pm module that used this with the version answer date. So I thought I was keeping the behavior of `BeforeAnswerDate` the same according the the GatewayQuizzes.pm module.

drgrice1:
The current behavior is what is in the code now without your pull request.  You are certainly not keeping that behavior the same in the code for this pull request.

somiaj:
The current behavior is not the same across modules, and so to make things consistent it cannot be kept the same everywhere. I'm using the GatewayQuiz.pm `can_` modules to determine what the current behavior should be:

```diff
	($Set->hide_work() eq 'N' ||
-			($Set->hide_work() eq 'BeforeAnswerDate' && time > $tmplSet->answer_date)));
+			($Set->hide_work() eq 'BeforeAnswerDate' && time > $tmplSet->answer_date) ||
+			($Set->hide_work() eq 'BeforeVersionAnswerDate' && time > $Set->answer_date)));
}
```

The old option is still pared with `$tmplSet->answer_date`, keeping it the same (this was similar in all `can_` modules). Elsewhere the behavior is different (It is currently possible to go to the quiz module and be told you can see your answers at a date in the past because it is printing the version answer date, but is actually using the template answer date as shown above). So in either case the behavior will change somewhere. This to me is kinda the crutch of the consensus of #740, what should the current behavior be, and it has mostly be left alone in this inconsistent case from what I can tell.

I can switch it up as you suggest, in that case the `can_` modules will no longer function the same.

drgrice1:
This will need some consideration.  This pull request should definitely not be merged until after #1569, #1563, and #1576.  Then we will need to look at this closely.

Be warned that if this changes current behavior, this pull request may see the same fate as #740.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -85,11 +86,12 @@ sub can_showCorrectAnswers {
 	my $authz = $self->r->authz;
 
 	# gateway change here to allow correct answers to be viewed after all attempts
-	#   at a version are exhausted as well as if it's after the answer date
+	#   at a version are exhausted as well as if it's after the due date
 	# $addOne allows us to count the current submission
 	my $addOne = defined($submitAnswers) ? $submitAnswers : 0;
 	my $maxAttempts = $Set->attempts_per_version() || 0;
 	my $attemptsUsed = $Problem->num_correct + $Problem->num_incorrect + $addOne || 0;
+	my $setCompleted = after($Set->due_date) || ($attemptsUsed >= $maxAttempts && $maxAttempts != 0);
```
drgrice1, line 94: Changing this to using the due_date instead of the answer_date definitely changes behavior in an unacceptable way.  This is the meaning of the answer_date.  There are those (myself included) that want this behavior.  I want my students to be able to see the answers to the quiz they have just completed, as soon as it is done. 

Also, there is no need for the `$setCompleted` variable.  The value of this variable is only used once, so just put its definition where it is used.  In addition, this will make the changes to code more clear as that is how it was (correctly done) before.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -104,13 +106,12 @@ sub can_showCorrectAnswers {
...
-			($Set->hide_score eq 'BeforeAnswerDate' && after($tmplSet->answer_date)));
+			($Set->hide_score eq 'BeforeAnswerDate' && after($tmplSet->answer_date)) ||
+			($Set->hide_score eq 'BeforeVersionAnswerDate' && after($Set->answer_date)));
 
-	return (((after($Set->answer_date) ||
-				($attemptsUsed >= $maxAttempts && $maxAttempts != 0 &&
-					$Set->due_date() == $Set->answer_date())) ||
-			$authz->hasPermissions($User->user_id, "show_correct_answers_before_answer_date")) &&
-		($authz->hasPermissions($User->user_id, "view_hidden_work") || $canShowScores));
+
```
drgrice1, line 112: You added an extra empty line here.  Never more than one consecutive empty line.

drgrice1 (changes requested):
This definitely changes behavior in an unacceptable way.  Switching to allowing answers to only be shown after the due date instead of the answer date is not correct.

If you can implement the new option so that the old behavior is not changed, and so that existing sets continue to work, then that is fine.

drgrice1:
I think that we need to have a discussion about how we want the GatewayQuiz.pm module to behave.  I am a bit loathe to add yet another option to the configuration of gateway quizzes, when we already have so many options and these options confuse many already.  I realized this doesn't affect my usage as I use the "Y" options for both showing problems and showing scores, but this change may not work for others.  I

somiaj:
Agreed, the options are confusing, and to me the most confusing part was the template answer date vs the version answer date which is dynamically generated. It is also very hard to adjust the version answer date for multiple students/versions without access to the database. So my attempt was to make it a lot clearer there are two answer dates and to allow the instructor to toggle which one they want to use. There was also the undocumented (at least not that I found) condition of what to do when the answer date = close date that I thought I could clean up with this as well.

But in trying to make this more configurable, I did overlook one case that I didn't want to break. I confused myself and was treating show answers the same as `version completed && show score && show work`, and this works for almost all cases except the case where an instructor wants a delay between showing the score and showing the answers while still letting students see the problems. In this case I don't see a simple way to allow the instructor to control which answer date to use (template vs version).

I have added the logic back about if answer date = close date and the version is completed (no more attempts remain), that they can see the answers right away, and also that the version answer date will be used in the case the instructor wants a delay between showing the score and showing the answers. This of course means they have to use the dynamically created date and cannot set a single date for which all students can see the answers (without also delaying when they can see the score and/or problems).

It could be that there is no nice way to clean this up, make it more understandable to instructors, trying not to break current behavior, without some redesign of how the GatewayQuiz module works.

This was meant to be a draft to attempt to try to clean up these confusing options. Let's see if anyone else wants to chime in on what they think. I'm fine with dropping this if no one else thinks this is a wanted change.

I do think that if this gets dropped, the second patch of adding some help to the gateway quiz options and a new patch to clean up the consistency in other modules should be done.

somiaj:
Closing, as this pull request isn't going anywhere.

PR #1581, opened 2022-02-07
Remove the direct adding of pg javascript and css from the problem rendering on the webwork2 side.

drgrice1:
Now pg requests those files via ADD_JS_FILES and ADD_CSS_FILES.  This is done in openwebwork/pg#652.

To make this fully work useMathView and useWirisEditor are added to what is passed to pg (along with useMathQuill), and pg requests the necessary javascript and css for those.

Note it is expected that the front end (webwork2 in this case) add the the javascript and css for bootstrap, mathjax, and jquery (and jquery-ui).


drdrew42 (approved):
Tested with MQ, MV, and WIRIS 👍

pstaabp (approved):
This looks good.

PR #1584, opened 2022-02-08
Fix the importClassList.pl and upgrade_admin_db.pl scripts for the move of WWSafe.pm to the pg side.

drgrice1:
These scripts were missed when these things were fixed before apparently.  The upgrade_admin_db.pl script is a rather critical one.

### DIFF
--- a/bin/importClassList.pl
+++ b/bin/importClassList.pl
@@ -1,29 +1,36 @@
 #!/usr/bin/env perl
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2019 The WeBWorK Project, http://openwebwork.sf.net/
-# 
+# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+#
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
 # Free Software Foundation; either version 2, or (at your option) any later
 # version, or (b) the "Artistic License" which comes with this package.
-# 
+#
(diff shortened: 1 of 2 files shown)


bin/importClassList.pl
```diff
@@ -56,7 +63,7 @@ BEGIN
 my @replaceList =();          # Empty list
 my (@replaced, @added, @skipped);
 
-# This was copied with MINOR changes from lib/WeBWorK/ContentGenerator/Instructor/UserList2.pm 
+# This was copied with MINOR changes from lib/WeBWorK/ContentGenerator/Instructor/UserList2.pm
```
pstaabp, line 66: `UserList2.pm` doesn't exist anymore.  If this is relevant for `UserList.pm`, else delete the comment?

drgrice1: It is probably still relevant for `UserList.pm`.  Recall that `UserList2.pm` was moved to `UserList.pm` and the old `UserList.pm` was removed.  I will change the comment.

pstaabp:
Can you also switch the top line of the `update_admin_db.pl` file to `#!/usr/bin/env perl`?

pstaabp:
We should eventually change the tops of all of the scripts to this.

drgrice1:
I changed that shebang as requested.

pstaabp:
Not sure what I just did.  I seem to somehow merged in the previous commit.

drgrice1:
Why is this pull request still showing up?  What did you do @pstaabp?

drgrice1:
Did you push this on the command line?

PR #1585, opened 2022-02-08
Improve database usage by consolidating and removing database calls

drgrice1:
Most of the DBFIXME's scattered throughout the code have been dealt with.  In some cases this was accomplished by doing exactly what was said.  In some cases better approaches were used to consolidate calls and cache data.

In other cases the DBFIXME's were just deleted.  The primary case of this is those asking for the usage of database iterators.  Those are really not correct.  The amount of data contained in the database queries is generally not substantial enough to justify their usage.  Furthermore, the correct usage of database iteration (not implemented in webwork) would accumulate data into chunks of a more significant size rather than handle each row separately.

`WHERE` clauses are used to obtain records, rather than listing id's and then filtering on those id's.  `ORDER BY` clauses are used to do sorting in the database, rather than sorting in perl.

This usage has been tested on a course that I have that has more than 40,000 users.  Significant speed improvements are seen particularly on the pages that utilize the data for all users, like the classlist editor, statistics, and student progress pages.

Note that now the set level proctor users are not visible anywhere in the user interface.  Previously they were hidden on the classlist editor page, but showed up many places where they should not have.  There isn't anywhere that they should be shown.

Also note that several invalid `e` flags on regex expressions have been removed in Grades.pm and SendMail.pm.  There was a FIXME in Grades.pm stating that this flag is not needed for variable interpolation, and it isn't.  I have seen also seen errors due to these flags being used.

I recommend using the "Hide Whitespace" option to view this (if you haven't become accustomed to this already).

I also added a .perltidyrc file and ran perltidy on all of the files in the directory lib/WeBWorK/DB.  There are no operational changes to those files, so just ignore them in reviewing this pull request.  The pull request is currently in two commits.  The first has all of the real changes, and the second the perltidy stuff.  (Note that the .perltidyrc file is the same as the webwork3 file except I added the `-xci` option.  We may want to add that to webwork3's file and the pg one also.  That does a much better job of indentation with nested structures.)

This is #1583 again.  @pstaabp be more careful!

### DIFF
--- a/.perltidyrc
+++ b/.perltidyrc
@@ -1,0 +1,22 @@
+# PBP .perltidyrc file
+-l=120  # Max line width is 120 cols
+-et=4   # Use tabs instead of spaces.
+-i=4    # Indent level is 4 cols
+-ci=4   # Continuation indent is 4 cols
+-b      # Write the file inline and create a .bak file
+-vt=0   # Minimal vertical tightness
+-cti=0  # No extra indentation for closing brackets
+-pt=2   # Maximum parenthesis tightness
+-bt=1   # Medium brace tightness
+-sbt=1  # Medium square bracket tightness
+-bbt=1  # Medium block brace tightness
+-nsfs   # No space before semicolons
+-nolq   # Don't outdent long quoted strings
+-mbl=1  # Do not allow multiple empty lines
+-ce     # Cuddled else
+-cb     # Cuddled blocks
+-nbbc   # Do not add blank lines before full length comments
+-nbot   # No line break on ternary
+-nlop   # No logical padding (this causes mixed tabs and spaces)
+-wn     # Weld nested containers
+-xci    # Extended continuation indentation
--- a/bin/importClassList.pl
+++ b/bin/importClassList.pl
@@ -69,7 +69,7 @@ my (@replaced, @added, @skipped);
 sub importUsersFromCSV {
 	my ($fileName, $createNew, $replaceExisting, @replaceList) = @_;
 
-	my @allUserIDs = grep {$_ !~ /^set_id:/} $db->listUsers;
+	my @allUserIDs = $db->listUsers;
 	my %allUserIDs = map { $_ => 1 } @allUserIDs;
 
 	my %replaceOK;
--- a/lib/WeBWorK/AchievementEvaluator.pm
+++ b/lib/WeBWorK/AchievementEvaluator.pm
@@ -33,324 +33,309 @@ use WWSafe;
 
 sub checkForAchievements {
 
-    our $problem = shift;
-    my $pg = shift;
-    my $r = shift;
-    my %options = @_;
-    my $db = $r->db;
-    my $ce = $r->ce;
+	our $problem = shift;
+	my $pg      = shift;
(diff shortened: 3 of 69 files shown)


drdrew42:
I'm getting an error on the Email page (SendMail.pm L#138-47):
`DBD::mysql::st execute failed: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AND ( `user_id` NOT LIKE 'set_id:%' AND `user_id` NOT LIKE 'practice%' ) ) ) ORD' at line 1 at /opt/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm line 881.`

Which is somewhat strange, given that a similar WHERE clause is used elsewhere (e.g. StudentProgress) and I don't see errors there.

drgrice1:
It isn't odd.  You don't have viewable_sections or viewable_recitations arrays defined, and I forgot to add the protection about that here.  I added it on line 350 of StudentProgress.pm, but forgot it here.

drgrice1:
Thanks for testing and finding that.  It is fixed now.

drdrew42:
I think you committed some other changes ... ceed876 modifies UserDetail.pm

drgrice1:
Oops.  Did not want `git commit -a`.

drgrice1:
Ahh, I didn't use `git commit -a`.  I just added the wrong file.

drgrice1:
Also, I force pushed that fix.

drdrew42:
That worked, the email page now loads without error. 👍

pstaabp:
Overall, this looks good.  I'm trying to test everything, since this seems to touch nearly every part of webwork. 

Since you mention speed up, I've noticed that the Homework Sets Editor still loads slowly compared to others and I only really notice it when I have caching turned off.  I haven't dug into what might be causing it.  Was there anything in here that should speed this up?

drgrice1:
I have noticed that as well.  I am still looking into that.  The thing is that that page loads more data from the database than any other.  It doesn't do so very efficiently, but to make it so will require quite the rewrite.

pstaabp:
This still seems a bit slow compared to other pages, but tolerable--especially with caching on.  Personally, I would spend too much time on this.  It's been like this for a long time and I think I (and others) are just used to it.

pstaabp:
Looking at some timing, Homework Sets Editor seems 2x to 3x slower than Classlist editor (ignoring loading of js/css).

pstaabp (approved):
Overall, good improvement.

dlglin:
> Since you mention speed up, I've noticed that the Homework Sets Editor still loads slowly compared to others and I only really notice it when I have caching turned off. I haven't dug into what might be causing it. Was there anything in here that should speed this up?

Every time the Homework Sets Editor loads it combs through the entire templates directory tree looking for def files, including the OPL and Contrib.  Depending on how fast your storage is this can greatly slow down the page load.

drgrice1:
@dlglin:  I just did some timing of things, and came to exactly that conclusion.  It is not a database query, it is line 1269 of ProblemSetList.pm that is causing the slow down, and that is the line that does what you say.

PR #1586, opened 2022-02-09
New feature: Add the capability of deleting a specific quiz version.

drgrice1:
This has been requested in the forums (see for instance https://webwork.maa.org/moodle/mod/forum/discuss.php?d=2400), and is something that I would also find useful.  In my classes I have quizzes that allow three versions with one submission per version, and show only one problem per page.  Students frequently will start one version, and then accidentally hit "Grade Test" after only working one problem.  This wastes one of the versions for this student.  Prior to this pull request I could do this, but had to do so manually by deleting the version and problems in the database.  Note that doing so also does not cause problems anywhere else.  Note that prior to pull request #1585 there was a minor glitch caused on the user detail page if a version was deleted that was not the last version that exists.  (This frequently happens because not only do students accidentally hit submit, but then they go start another version before contacting me to fix the problem.)  The glitch was caused by the assumption that the version_id stored in the database coincided with array indices (offset by one) in the array of versioned user sets obtained by listing all versions in the database. That of course is not the case with #1585.  I have been deleting versions manually for years with no issues other than this one.

This adds a checkbox to the quiz versions on the user detail page (which you can get to by clicking on the assigned sets for a user in the class list editor).  If this checkbox is unchecked, then that specific version of the gateway quiz will be deleted for that user.  Note that javascript is used to make the server side behavior clear to the user.  If the template set is unchecked, then all of the versions are unchecked as in addition to the template set being unassigned, the versions will all be deleted.  If a version is checked (after previously being unchecked) then it is ensured that the template set is also checked.

Since this adds a javascript file for the UserDetail.pm module, all of the inline javascript in that file is also moved into the javascript file (the change, blur, and keyup handlers for the date inputs, and the click handler for the "Assign All Sets to Current User" button).

This clearly depends on #1585, and should not be merged before that one (merging this pull request would also merge that one).

### DIFF
--- a/.perltidyrc
+++ b/.perltidyrc
@@ -1,0 +1,22 @@
+# PBP .perltidyrc file
+-l=120  # Max line width is 120 cols
+-et=4   # Use tabs instead of spaces.
+-i=4    # Indent level is 4 cols
+-ci=4   # Continuation indent is 4 cols
+-b      # Write the file inline and create a .bak file
+-vt=0   # Minimal vertical tightness
+-cti=0  # No extra indentation for closing brackets
+-pt=2   # Maximum parenthesis tightness
+-bt=1   # Medium brace tightness
+-sbt=1  # Medium square bracket tightness
+-bbt=1  # Medium block brace tightness
+-nsfs   # No space before semicolons
+-nolq   # Don't outdent long quoted strings
+-mbl=1  # Do not allow multiple empty lines
+-ce     # Cuddled else
+-cb     # Cuddled blocks
+-nbbc   # Do not add blank lines before full length comments
+-nbot   # No line break on ternary
+-nlop   # No logical padding (this causes mixed tabs and spaces)
+-wn     # Weld nested containers
+-xci    # Extended continuation indentation
--- a/bin/importClassList.pl
+++ b/bin/importClassList.pl
@@ -69,7 +69,7 @@ my (@replaced, @added, @skipped);
 sub importUsersFromCSV {
 	my ($fileName, $createNew, $replaceExisting, @replaceList) = @_;
 
-	my @allUserIDs = grep {$_ !~ /^set_id:/} $db->listUsers;
+	my @allUserIDs = $db->listUsers;
 	my %allUserIDs = map { $_ => 1 } @allUserIDs;
 
 	my %replaceOK;
--- a/htdocs/js/apps/UserDetail/userdetail.js
+++ b/htdocs/js/apps/UserDetail/userdetail.js
@@ -1,0 +1,43 @@
+(() => {
+	// Check or uncheck assignment checkboxes for versioned sets and their template to make it clear to the user what
+	// the backend will do with their selections.
+	document.querySelectorAll('input[type="checkbox"][name^="set."][name$=".assignment"]').forEach((checkbox) => {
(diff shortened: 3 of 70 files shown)


drgrice1:
Note there is one thing that would be nice with this that I am open to suggestions as to how to deal with.  That is that now there are checkboxes by all sets in the list on the user detail page.  It would be nice if the versioned set checkboxes would be offset in some nice way to make it clearer that those are dependent on the template checkbox.  Javascript makes the behavior clear by checking/unchecking things appropriately, but the display doesn't.  One easy way of doing this is to add a left margin to those checkboxes.  That works, but doesn't really look the best to me.

drgrice1:
I am going to need specific details on this.  What are the quiz parameters?  What exact steps need to be taken to reproduce this.  I can't make it happen.

somiaj:
I'm am also having trouble trying to reproduce this in a fresh assignment too and was gonna credit this to the fact I was using an old quiz with lots of versions with zero submits. But in my final test it appears to deal with quiz versions  >= 10. I am not sure on exact steps (as I initially missed when this happened), but once I had more than 10 quiz versions, and deleted version 11, and then created a new version 11 I noticed the following:

image

Quiz 10 times just do not match up, the dates/time for quiz 10 are now the exact same as quiz version 1, but version 10 was created after version 9, so I don't see why it would have times before it. I then went and created a new version 12, then deleted version 12, and now I see the following:

image

So now both quiz 11 and quiz 10 have had their dates changed. So something strange is going on once I had more than 10 versions. I'll keep doing some more tests to see if I can figure out exact steps to get this to happen.

somiaj:
And here are my quiz parameters:

image

somiaj:
Okay, some more testing, and it does appear it has to deal with >= 10 versions, and is not related to deleting of versions at all. I quickly created 11 versions by create version -> grade test -> create version -> create test -> etc. In order to track the times of the versions, I then when and edited all the times of the versions so they all had different times (so I could tell the times apart). And then just save, and that was enough to get this strange behavior.

Original times of the 11 versions:

image

My changes to the 11 versions so I can tell them apart:

image

Click Save Changes:

image

As you can see, the values I entered were not the values that got set. I wonder if this is due to the versions being ordered alphabetically: 1, 10, 11, 2, 3, 4, 5, 6, 7, 8, 9

This might just be some underlying bug that I just happened to stumble upon in testing this.

drgrice1:
I already know what is going on. Please stop posting screenshots. I will fix it later when I have time.

drgrice1:
The sorting issue is fixed.  Although, not in this pull request, but in #1585 where the problem was caused.  This pull request is rebased onto the changes there.

drgrice1:
> Since deleting versions is permanent, in addition to all the warnings about this, would a popup that lists the sets/versions that are being unassigned/deleted then click OK to proceeded be helpful to avoid accidentally deleting the wrong thing? Though such a popup could be used elsewhere and wouldn't be a part of this pull request.

I think that a popup dialog with the list of things that will be changed might be nice, but is not essential.  I doubt that your idea of this being used elsewhere would work though.  Popup dialogs are javascript, and making that modular in such a way that it would work elsewhere would be extremely difficult.  Particularly because there isn't anywhere else that really has this same type of data.

PR #1587, opened 2022-02-09
New approach to restricted navigation for use with LTI authentication.

drgrice1:
This adds a new permission `navigation_allowed`.  Users that do not have this permission will not be allowed to view the course home page (i.e., the problems sets list page in ProblemSets.pm).  These users will also be restricted to only viewing one assignment set.  That is the set that the user initially signs in to view.

This does not remove the breadcrumb and site navigation elements.  Instead the elements in the breadcrumb and site navigation that would normally be links that take the user to these restricted pages are not links for these users.  Furthermore, the set that the user initially signs in to view is always present in the site navigation and is of course a link that the user can use to get back to that set.  So if the user goes to the `User Settings`, `Email Instructor`, `Download Hardcopy`, `Achievements` (if enabled), or any other unrestricted page, the user can still get back to the set the user signed in to view.  Note that the pages that are considered unrestricted are flagged as such in URLPath.pm.  Note that the Grades page is not one of these due to the
links to other problems sets that are provided there.

This works by making the `setID` persistent in various ways so that the `links` method in ContentGenerator.pm can always get this value.  First, if a link in the page has the `setID` path parameter then the `links` method gets it from there (this is not a change it has always done that).  Second, any link in the page that does not have the `setID` path parameter has the `setID` url parameter added (only for these users though).  Third, a hidden input is added to any form (via hidden_authen_fields) with name `setID`.  Of course both of these latter approaches will then be contained in the `$r->param` hash so that the `links` method can get it.

Also, if a user tries to modify the location address in the browser to get to the ProblemSets.pm page, a message will be displayed telling the user to access assignments from their course management system.

If LTIGradeMode is set to homework and the user modifies the location address in the browser to point to another set for the class that the student has not accessed yet via the learning management system it will give the error `The selected problem set (setID) is not a valid set for user_id.  You must user your LMS to access this set.` (not exact wording).  This of course is already what happens in this case.


pstaabp (approved):
Looks like this works.  I don't have an LTI to test from, but after changing the navigation_allowed to above student, I get the `You must access assignments from your Course Management System ().` error when navigating in the URL bar.

drgrice1:
Closed in favor of #1588.

PR #1588, opened 2022-02-09
New approach to restricted navigation for use with LTI authentication (alternate new approach).

drgrice1:
This adds a new permission `navigation_allowed`.  Users that do not have this permission will not be allowed to view the course home page (i.e., the problems sets list page in ProblemSets.pm).  These users will also be restricted to only viewing one assignment set.  That is the set that the user initially signs in to view.

This does not remove the breadcrumb and site navigation elements. Instead the elements in the breadcrumb and site navigation that would normally be links that take the user to these restricted pages are not links for these users.  Furthermore, the set that the user initially signs in to view is always present in the site navigation and is of course a link that the user can use to get back to that set.  So if the user goes to the `User Settings`, `Email Instructor`, `Download Hardcopy`, `Achievements` (if enabled), or any other unrestricted page, the user can still get back to the set the user signed in to view.  Note that the pages that are considered unrestricted are flagged as such in URLPath.pm.  Note that the Grades page is not one of these due to the
links to other problems sets that are provided there.

This works by saving the `setID` in the session so that the `links` method in ContentGenerator.pm can always get this value.  This value is saved in the session key in the database if session_management_via is set to key, and is saved in the session cookie (and actually the session key in the database as well) if session_management_via is set to session_cookie.  **The downside of this approach is that it requires a new column in the database key table.**  The upside of this approach is that it gives a perhaps more reliable method of keeping the setID active for the session.

Also, if a user tries to modify the location address in the browser to get to the ProblemSets.pm page, a message will be displayed telling the user to access assignments from their course management system.

If LTIGradeMode is set to homework and the user modifies the location address in the browser to point to another set for the class that the student has not accessed yet via the learning management system it will give the error `The selected problem set (setID) is not a valid set for user_id.  You must user your LMS to access this set.` (not exact wording).  This of course is already what happens in this case.

Of course only of of #1587 or this pull request should be merged.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -750,6 +750,7 @@ $authen{xmlrpc_module}  = "WeBWorK::Authen::XMLRPC";
 #
 %permissionLevels = (
 	login                          => "guest",
+	navigation_allowed             => "guest",
 	report_bugs                    => "ta",
 	submit_feedback                => "student",
 	change_password                => "student",
@@ -1948,6 +1949,15 @@ $ConfigValues = [
 					. 'guest accounts.'
 			),
 			type => 'permission'
+		},
+		{ var => 'permissionLevels{navigation_allowed}',
+		  doc => 'Allowed to view course home page',
+		  doc2 => 'If a user does not have this permission, then the user will not be allowed to navigate to the '
+			. 'course home page, i.e., the Homework Sets page.  This should only be used for a course when LTI '
+			. 'authentication is used, and is most useful when LTIGradeMode is set to homework.  In this case the '
+			. 'Homework Sets page is not useful and can even be confusing to students.  To use this feature set '
+			. 'this permission to "login_proctor".',
+		  type => 'permission'
 		},
 	],
 	[
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -96,7 +96,7 @@ table caption {
 		max-width: 100%;
 	}
 
-	a {
+	a, span {
 		display: block;
 		margin-left: 0.5rem;
 		margin-right: 0.5rem;
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -574,7 +574,7 @@ sub maybe_send_cookie {
 	my $r = $self->{r};
 	my $ce = $r -> {ce};
 
-	my ($cookie_user, $cookie_key, $cookie_timestamp) = $self->fetchCookie;
+	my ($cookie_user, $cookie_key, $cookie_timestamp, $setID) = $self->fetchCookie;
 
 	# we send a cookie if any of these conditions are met:
 
@@ -601,7 +601,7 @@ sub maybe_send_cookie {
 
 	if ($used_cookie or $unused_valid_cookie or $user_requests_cookie or $session_management_via_cookies) {
 		#debug("Authen::maybe_send_cookie is sending a cookie");
-		$self->sendCookie($self->{user_id}, $self->{session_key});
+		$self->sendCookie($self->{user_id}, $self->{session_key}, $setID);
 	} else {
 		$self->killCookie;
 	}
@@ -738,8 +738,9 @@ sub unexpired_session_exists {
 # the key is returned
 sub create_session {
 	my ($self, $userID, $newKey) = @_;
-	my $ce = $self->{r}->ce;
-	my $db = $self->{r}->db;
+	my $r = $self->{r};
+	my $ce = $r->ce;
+	my $db = $r->db;
 
 	my $timestamp = time;
 	unless ($newKey) {
@@ -750,7 +751,12 @@ sub create_session {
 		$newKey = join ("", @chars[map rand(@chars), 1 .. $length]);
 	}
 
-	my $Key = $db->newKey(user_id => $userID, key => $newKey, timestamp => $timestamp);
+	my $setID =
+		$r->param('user')
+		&& !$r->authz->hasPermissions($r->param('user'), 'navigation_allowed') ? $r->urlpath->arg("setID") : '';
+
+	my $Key = $db->newKey(user_id => $userID, key => $newKey, timestamp => $timestamp, set_id => $setID);
+
 	# DBFIXME this should be a REPLACE
 	eval { $db->deleteKey($userID) };
 	eval { $db->addKey($Key) };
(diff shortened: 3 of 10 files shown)


drgrice1:
After discussing this we believe that this is the way to go.  Pull request #1587 is really an unreliable hack to make this work.  If anyone disagrees say something please.

pstaabp:
To test this, you need to upgrade the database.  To do this
1. run `bin/upgrade_admin.pl` from the webwork2 directory.
2. log in to the admin course
3. Under "Upgrade Course", follow the steps to upgrade the course.

drgrice1:
Thanks @pstaabp. I forgot to mention that.

drgrice1:
In order to facilitate testing of this pull request I wrote a small Mojolicious app that emulates an LMS and perfoms LTI authentication into webwork.  The app is in the attached zip file.
[lms-test-app.zip](https://github.com/openwebwork/webwork2/files/8502026/lms-test-app.zip)

To use the app unzip the attached file's contents, and copy `l_m_s.dist.yml` to `l_m_s.yml`.  You can set the `webwork_url` (full url to homework assignment on server including http[s]://) and the `consumer_secret` in that file for convenience, or just enter them in the form in the app UI.  Run `morbo script/LMS`, and open your browser to `localhost:3000`.  The usage should be pretty straightforward from there.

drdrew42:
@drgrice1 absolutely terrific! Thanks so much for setting up this app for testing -- so easy, and works (though the iframe login is not persistent if is sent to the set instead of an individual problem, or attempt is made to proceed to the 'next problem')

drdrew42 (approved):

drgrice1:
@drdrew42:  What you are seeing with iframe login not being persistent depends on your cookie settings for WeBWorK, and is not a problem with the app but it something that also occurs with actual LMS's with these settings as well.  This will happen if you have the default settings of `$CookieSameSite="Lax"` and `$session_management_via="session_cookie", and your site is secured with SSL certificates.  If you change the cookie setting to `$CookieSameSite="None"` this will not happen.

drgrice1:
That is why I included the "iframe" option in the app.  So you can test these things.  The app really does the same thing that an LMS does.

Also, grade passback works with the app!

drdrew42:
Yeah, I saw those routes, so I figured there was more to it -- I had to have missed something. I'll try changing the `$CookieSameSite` now

drdrew42:
Okay, so your advice fixed the iframe persistent login issue. However, I wasn't able to get grade passback working (expected since I'm sending local LMS to remote WW2, with no route back to the LMS). However, the debug warning message for grade passback seems to indicate the wrong score?

Screen Shot 2022-04-19 at 5 01 50 PM

drdrew42:
Also looks like there's a missing newline between the first two `scoreRecordedMessage` sentences.

drgrice1:
Yeah, if you are on a local computer with remote webwork and no route back, then grade passback won't work.

As to the incorrect grade thing, that should be unrelated to this pull request.  Try it with the develop branch of webwork and see if the same thing happens.  This pull request doesn't change anything with grade passback.

In which file are you talking about with the `scoreRecordedMessage` sentences?

drdrew42:
> As to the incorrect grade thing, that should be unrelated to this pull request. Try it with the develop branch of webwork and see if the same thing happens. This pull request doesn't change anything with grade passback.

Yep, still getting the wrong score reported in the warning/debug message.

drgrice1:
I see.  It seems there needs to be a `<br>` injected.  I will see if I can fix that and put that in a separate pull request, since it is unrelated to this one.

As to the grade sent, which grade passback mode do you have set in `authen_LTI.conf?   Also, what grade are you expecting to have been sent?  Note that the grade that is passed back is either the grade for all assigned sets (if "course" mode is set) or for the whole set (if "homework" mode is set).  So although that problem has 100%, the course grade or set grade probably won't be 1.

drdrew42:
> So although that problem has 100%, the course grade or set grade probably won't be 1.

That's it. Mismatched expectations -- mode is 'homework' but for some reason I was expecting 'problem' ... d'oh!

pstaabp:
I think I'm having trouble setting up the LTI parameters correctly.  (Dragging me kicking and screaming into LTI land....). Here's the parameters/error that I see trying to log in via the Mojolicious service as a student:
```
===== parameters received =======
user_id => homer
oauth_version => 1.0
lis_person_contact_email_primary => homer@msn.com
oauth_signature_method => HMAC-SHA1
oauth_consumer_key => webwork
lis_person_name_given => Homer
lis_outcome_service_url => http://localhost:3000/grade-passback
oauth_nonce => 039c6fb70dca6a986a07bc3b6e348941
roles => Student
oauth_signature => aHxn0zDYm/pql7MKPwmDrFoBNdE=
lis_person_name_family => Simpson
lis_result_sourcedid => 1000
oauth_callback => about:blank
lis_person_sourcedid => homer
oauth_timestamp => 1650539811
============================
===== timestamp info =======
oauth_nonce = 1650539811
WW_server_time = 1650539811
diff(server-oauth) = 0 seconds (0 minutes)
============================ at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 142.
=========== summary ============ at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 261.
User id is |homer| (obtained from lis_person_sourcedid which was preferred_source_of_username)
User mail address is |homer@msn.com|
strip_address_from_email is |0|
Student id is ||
preferred_source_of_username is |lis_person_sourcedid|
fallback_source_of_username is |lis_person_sourcedid|
preferred_source_of_student_id is ||
================================
The following path was reconstructed by WeBWorK. It should match the path in the LMS: at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 444.
http://localhost/webwork2/staab_course/set_1/ at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 445.
```

drgrice1:
Step 1:  Make sure that the `$LTIBasicConsumerSecret` in `webwork2/conf/authen_LTI.conf` is exactly what you enter into the "Consumer Secret" field in the test app.

Step 2:  The "WeBWorK Url" must begin with "$server_root_url$webwork_url" where the variables in that string are replaced with the values you have in `webwork2/conf/site.conf`.  The rest of the url should be the course id and then the set id.

If those things are set up right and you are using a student user ("User Role" = "Student"), then it should work.

pstaabp:
It was just that the parameters were being shown for debugging. (LTI config page).  Should this be true for default?

drgrice1:
Also, note that you must provide the "User Email Address" with the default setting of `$preferred_source_of_username="list_person_contact_email_primary"` that is set in `webwork2/conf/authen_LTI.conf`.  Either that or change that setting to `$preferred_source_of_username="lis_person_sourcedid"`.  Maybe I should have made the "User Email Address" field required in the app UI.

Although, it looks like you did provide a value in that field from your debug message, so that isn't your issue.

drgrice1:
I see.  Yeah, I find that confusing too.  That is the default value that has been there for a long time (it looks like since 2016 in the WeBWorK 2.14 release).

pstaabp (approved):
After navigating the LTI waters, this looks good...

pstaabp:
We'll need to add info about updating the database in the 2.17 release notes as well.

taniwallach:
> I see. Yeah, I find that confusing too. That is the default value that has been there for a long time (it looks like since 2016 in the WeBWorK 2.14 release).

@mgage felt (see https://github.com/openwebwork/webwork2/pull/857#issuecomment-526681087 ) that having the debug default to on was helpful for sites first setting up LTI, and that sites would disable it once they got the connection working. It is something which can be reconsidered.

pstaabp:
BTW, I just added the steps to update the database in the release notes.

drgrice1:
@pstaabp: Thanks for taking care of that.

PR #1589, opened 2022-02-10
Fix errors when processing unassigned quizzes.

somiaj:
When linking directly to a gateway quiz from an LMS (or typing in the url with quiz_mode), GatewayQuiz.pm would error out if the quiz was not assigned to the user. This provides a fix to avoid trying to access undefined variables in the case authz->checkSet has failed, and updates the class for the alert div that returns the message about trying to access an invalid set to match what is done in ProblemSet.pm.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -414,6 +414,9 @@ sub get_instructor_comment {
 sub pre_header_initialize {
 	my ($self)     = @_;
 
+	# if authz->checkSet has failed, this set is invalid, and no need to proceeded.
+	return if $self->{invalidSet};
+
 	my $r = $self->r;
 	my $ce = $r->ce;
 	my $db = $r->db;
@@ -1290,7 +1293,7 @@ sub body {
 
 	if ($self->{invalidSet} || $self->{invalidProblem}) {
 		# delete any proctor keys that are floating around
-		if ($self->{'assignment_type'} eq 'proctored_gateway') {
+		if (defined($self->{'assignment_type'}) && $self->{'assignment_type'} eq 'proctored_gateway') {
 			my $proctorID = $r->param('proctor_user');
 			if ($proctorID) {
 				eval{ $db->deleteKey("$effectiveUser,$proctorID"); };


lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -1314,7 +1317,7 @@ sub body {
 		}
 
 		return CGI::div(
-			{ class => 'ResultsWithError mb-2' },
+			{ class => 'alert alert-danger mb-2' },
```
drgrice1, line 1320: This should stay the way it was.  The GatewayQuiz.pm module uses the `ResultsWithError` class from pg/htdocs/js/apps/Problem/problem.scss.

somiaj: Okay, I can revert that. That class just gives a white background with a shadow under the bottom/right border, which made me think maybe this class wasn't updated with the bootstrap 5 change, so I made it match what happens in the ProblemSet.pm case that puts the output in a red div.

drgrice1:
It sounds like you need to run `npm install` in the `pg/htdocs` directory to update `pt/htdocs/js/apps/Problem/problem/css`.

drgrice1 (approved):
This pull request looks good.  Although this is fixing an outlier case that really shouldn't occur unless someone is messing around with the url, or an instructor is providing invalid links for students to use in the LMS.

drgrice1:
Note that the `ResultsWithError` class is changed in pull request #1581.  With that pull request the `problem.css` file will only be loaded on a page if a problem is loaded.  This particular instance of the class is only displayed when there is no problem shown.  So it won't work after that pull request.  At this point the `problem.css` file is always loaded for the gateway quiz pages (it is in the template).  Note all of the other instances of this class in GatewayQuiz.pm and Problem.pm are intentional, and should not be changed.  Those messages will only be displayed in the cases that a problem is loaded.

pstaabp (approved):
I'm still setting up a LTI to start testing these cases, but the code looks fine.

somiaj:
@pstaabp you can just type the url directly into the address bar, the LMS was the mechanism that one could get a link directly to a quiz without being assigned to it.

pstaabp:
> can just type the url directly into the address bar,

Yep.  Got the error in develop.  This does a nice job catching it.

PR #1590, opened 2022-02-11
Allow student access to existing Gateway versions even before the open date.

drgrice1:
If a specific version of a gateway quiz is requested, then use its open date rather than the template set's open date to determine if it is open.  Furthermore, if a gateway quiz version already exists then a link is provided on the Homework Sets page to the quiz version list page (in ProblemSet.pm) and student's are allowed to view that page even if it is before the set open date.  That page still shows that the set is closed and no new versions can be opened.

This allows for things like having a gateway quiz that is open for a period of time, then closed for a period, and then subsequetly open again.  With this pull request, students that had taken the quiz during the first period that the quiz was open can still view their past attempts during the period in the middle that the quiz is closed.

This was requested by @glarose for a situation that came up for him.

This also includes a small rework of how conditional release is implemented.  Instead of calling `die` in the content generator sub-modules when those pages are loaded, work the conditional release check into `checkSet` in lib/WeBWorK/Authz.pm.  This is how it should have been done in the first place.  Furthermore, there should never be a `die` call in the content generator.  There are probably still many of these though ... things done wrong.

Also included is the removal of some redundant database calls in ProblemSet.pm and the addition of doing ordering in the database.  The gateway quiz version list versus problem list portion of the code is also reorganized in a much more logical way.

### DIFF
--- a/lib/WeBWorK/Authz.pm
+++ b/lib/WeBWorK/Authz.pm
@@ -399,6 +399,11 @@ sub checkSet {
 			return $r->maketext("Requested set '[_1]' could not be found in the database for user [_2].",
 				$setName, $effectiveUserName);
 		}
+		# Don't allow versioned sets to be viewed from the problem-list page.
+		if ($node_name eq 'problem_list') {
+			return $r->maketext("Requested version ([_1]) of set '[_2]' can not be directly accessed.", $verNum,
+				$setName);
+		}
 	} else {
 		if ($set && $set->set_id eq $setName && $set->user_id eq $effectiveUserName) {
 			# If we have all of this, then we can just use this set and skip the rest.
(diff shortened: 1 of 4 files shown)


drgrice1:
I should have tested @glarose's code before creating this pull request.  It isn't working.  I will see if this can be made to work.

drgrice1:
It seems that the ProblemSet.pm page will also need something.

@glarose: Is this worth it for this odd scenario?  Is this something that frequently occurs for you?  I am guessing not, since this is the first time it has come up since you wrote the GatewayQuiz.pm module.  I will see what can be worked up in any case.

drgrice1:
Okay, this is all working now and is ready to be reviewed.

drgrice1:
Just in case there is concern, a student still will not be given access to a quiz version before the open date if the quiz is set up so that problems on finished tests are not allowed to be viewed.  Furthermore, scores will not be shown if the quiz is set up so that those are not allowed to be seen.

pstaabp:
First, trying to reproduce the bug posted in #1595.
1.  I have a quiz that was open earlier and a user took at least one version. 
2. Changed the test open date to after the date the user took the quiz. 

I should see that the user (tested both with a student and a professor)  cannot view the previously-taken version of the quiz.  I'm not seeing this.  Just want to clarify that this is the problem.

drgrice1:
Yes, with the current develop branch (or any previous version of WeBWorK), the student cannot view a previously taken quiz if the open date of the gateway quiz is set to a later date.  Here are the steps to properly test this.
1. Create a gateway quiz, set the open date to a date in the past, and assign it to a student in the course.
2. Login as that student and open and grade the quiz.
3. As the instructor, change the open date to a date in the future.
4. Login as the student again.

Without this pull request, after step 4 the student will no longer be able to access the quiz version.  On the homework sets page, the quiz will not be a link.  If you manually type in the link to access the version listing page (http://localhost/webwork2/courseID/setID) it will say the selected problem set is not yet open.  If you manually type the quiz_mode url (https://localhost/webwork2/courseID/quiz_mode/setID,v1) it say you do not have permission to view unopened sets.

With this pull request, the gateway quiz will be a link on the problem sets page, and all of the links are then accessible via the web interface and work.

pstaabp (approved):
Looks good.  I have verified that the above comments work.

PR #1593, opened 2022-02-13
Fix the tag widget.

drgrice1:
I broke this when I removed underscore from the vendor js files.  I forgot about this one completely unnecessary usage of that javascript library.  That usage has been removed.  Furthermore, this removes all of the related inline javascript.  The contents of the tagging-taxonomy.json file are now fetched by the javascript, so this also removes an instance of disk access in the perl code.

Also remove some javascript and css that is being loaded in ProblemSetDetail.pm and SetMaker.pm that isn't needed with the current rendering methods.

### DIFF
--- a/htdocs/js/apps/TagWidget/tagwidget.js
+++ b/htdocs/js/apps/TagWidget/tagwidget.js
@@ -1,268 +1,282 @@
 // Setup for ajax
-var basicWebserviceURL = "/webwork2/instructorXMLHandler";
-var basicRequestObject = {
-    "xml_command":"listLib",
-    "course_password":'change-me',
-    "session_key":'change-me',
-    "user":"user-needs-to-be-defined",
-    "library_name":"Library",
-    "courseID":'change-me',
-    "set":"set0",
-    "new_set_name":"new set",
-    "command":"searchLib"
-};
+(async () => {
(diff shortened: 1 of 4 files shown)


pstaabp:
I don't see the tag widget/editor at all now.  I'm logged in as admin and in the library browser, here's the top of a problem:
image

Before it had the select items, not styled with no options.  

Is there a setting I have to change now?

drgrice1:
It is not called via xmlrpc.  Have you restarted apache2 and dealt with browser caching issues?

drgrice1:
Note that if you are using the library browser, Problem.pm is not used at all for that.

drgrice1:
If you are using the library browser, then SetMaker.pm outputs the containing div for the tag widget.  The actual widget is injected by javascript.  None of that has changed.  The only thing that has changed is the way that the javascript does this injection.

drgrice1:
With this pull request the code in SetMaker.pm starts at line 1346.

pstaabp:
I see them now.  Probably a caching problem.  I thought it was disabled.  This is what I'm seeing now:
image

Is it supposed to be styled with bootstrap?

pstaabp (approved):
This looks good now.

PR #1594, opened 2022-02-14
Add a new permission `change_pg_display_settings`.

drgrice1:
If a user does not have this permission, then the user will not be able to change the display settings on the `User Settings` page.

I don't want my students changing these settings.  There really is no need for them to do so.  Also, changing the MathQuill setting from `Yes` to `No` can cause problems.  Basically, if the student has that set to `No`, and the professor has it set to `Yes`, then when the instructor acts as that student and views the problems for that student, no answers appear in the MathQuill answer inputs.  The answers are in the now hidden text inputs only.  Perhaps it should be made so that when an instructor acts as a student, the student's user settings for display are used.

Furthermore, change the display on the `User Settings` page.  If a user does not have a permission to change a setting, don't show "You do not have the permission to change this setting."  Instead just don't show the setting or its header at all.  There are three permissions for this: `change_password`, `change_email_address`, and the new `change_pg_display_settings`.

Additionally, if a user doesn't have any of those permissions, don't even add the `User Settings` link in the site nav (MAIN MENU).

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -754,6 +754,7 @@ $authen{xmlrpc_module}  = "WeBWorK::Authen::XMLRPC";
 	submit_feedback                => "student",
 	change_password                => "student",
 	change_email_address           => "student",
+	change_pg_display_settings     => "student",
 
 	proctor_quiz_login             => "login_proctor",
 	proctor_quiz_grade             => "grade_proctor",
@@ -1718,6 +1719,11 @@ $ConfigValues = [
 		{ var => 'permissionLevels{change_email_address}',
 		  doc => 'Allowed to change their e-mail address',
 		  doc2 => 'Users at this level and higher are allowed to change their e-mail address. Normally guest users are not allowed to change the e-mail address since it does not make sense to send e-mail to anonymous accounts.',
+		  type => 'permission'
+		},
+		{ var => 'permissionLevels{change_pg_display_settings}',
+		  doc => 'Allowed to change display settings used in pg problems',
+		  doc2 => 'Users at this level and higher are allowed to change display settings used in pg problems.',
 		  type => 'permission'
 		},
 		{ var => 'permissionLevels{view_answers}',
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -795,7 +795,10 @@ sub links {
 
 
 			print CGI::li({ class => 'nav-item' },
-				&$makelink("${pfx}Options", urlpath_args => { %args }, systemlink_args => \%systemlink_args));
+				&$makelink("${pfx}Options", urlpath_args => {%args}, systemlink_args => \%systemlink_args))
+				if ($authz->hasPermissions($userID, 'change_password')
+					|| $authz->hasPermissions($userID, 'change_email_address')
+					|| $authz->hasPermissions($userID, 'change_pg_display_settings'));
 
 			print CGI::li({ class => 'nav-item' },
 				&$makelink("${pfx}Grades", urlpath_args => { %args }, systemlink_args => \%systemlink_args));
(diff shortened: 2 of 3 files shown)


Alex-Jordan:
Some of the history is that back when "images" was more common for the course's default mode of rendering math, it was important for accessibility to allow students to switch to MathJax. So around version 2.13, this was added to User Settings.

Does any of that original reasoning still hold? Now the default is to use MathJax, and it's hard to see a reason to even offer `images` int he first place.

But with MathQuill rendering, is it important to allow a blind user to disable it?

BTW, while looking at this, I changed the setting to No, and I got the response message "Your display options have been saved" twice:
Screen Shot 2022-02-17 at 3 19 27 PM

One more thing. I think I understand that this User Settings option lets people turn off MathQuill rending in their answer blanks. But when I first came to this, I did not understand its purpose. I was thinking of "equation rendering" as the thing that renders equations that WW is displaying to me, and I couldn't make sense of it. Is there a way to phrase this so that it is clear that it is about rendering user input? Maybe "live rendering of input equations"?

drgrice1:
@Alex-Jordan: I suspect that you need to update your system.template file.  That is most likely why you got that message twice.

drgrice1:
"Images" are something that are being kept around largely for historical significance.  @mgage wants them to be kept.

If you have a blind student in your course, then you probably would need to not use this permission.  It is an option, and is configurable at the course level.  Although, out of curiosity what is the issue with a blind student for MathQuill?  Is that something that could be fixed so that MathQuill is accessible?  In any case, since it is an option, I do not see this as a stopper for this pull request.

As to the wording of the setting, that has already been discussed.  I agree with you on that.  I didn't pick that wording.

drgrice1:
Actually, if your system.template were not updated you would probably have worse failures, and that still wouldn't make that message come up twice.  I don't know how you got that.  I can't reproduce it.

Alex-Jordan:
> what is the issue with a blind student for MathQuill? 

I don't know if there is an issue. It just seems plausible that there could be one. If a blind user types some input in a regular input field, it is easy for them to review what they have typed to look for typos, for example. I am unsure if MathQuill affects their ability to review what they have entered before they hit submit.

I haven't updated system.template directly. Would that not come along with the git pull?

drgrice1:
I was thinking you were using a custom theme with a custom system.template file.  Yeah, the git pull would bring in the correct one if you are using one of the built in them (math4-red I presume).

drgrice1:
Also, on the disabling MathQuill thing, as I noted in my initial comment for this pull request, there are problems with disabling MathQuill.  That is really the main thing that led to this pull request.

It would be best to figure out if MathQuill can be made more accessible if there are issues there.

Alex-Jordan:
I couldn't reproduce it either, but then I hit on it. Try changing both the MathJax/images setting and the live rendering option at the same time. I must have changed two things on that run, although I cannot remember why.

drgrice1:
Are you doing that with this pull request, or are you getting that with the develop branch?

drgrice1:
I can reproduce what you are describing with the develop branch and with WeBWorK 2.16, but I can not reproduce what you are showing with this pull request.  In fact, that is one of the things that I fixed here.

Alex-Jordan:
Sorry, that's it then. I was only considering the question of allowing the disabling. So I went to User Settings in develop to understand what would be disabled. I did not pull this branch.

pstaabp:
Just to clarify, to test this, in course settings, you need to go to the new setting (in the Permissions tab) to 
image

and switch the option to something above student.

drgrice1:
Nope.  That is not what perltidy does.  It only adds spaces when there are nested delimiters.

pstaabp (approved):
Looks good.  Tested both 1) that using the new setting suppressed the ability to change display settings and 2) if a user does not have permissions to change display setting, change email or change password, then the User Settings is not shown in the navigation menu.

PR #1596, opened 2022-02-14
Fix show me another for problems with resources added via PGresource.

drgrice1:
The problem is that any resource (like an image) added in this way adds the url with unique id alias (from PGalias.pm) of the resource to the generated html (the pg body_text).  This will always be different if the seed is different.  So the check in ShowMeAnother.pm will always determine that the body_text of the new problem and the original problem
are different regardless of if there are any actual differences elsewhere.

So this pull request strips those urls from both the original problem and the new problem before comparison.

I have attached two test problems.
[TestProblems.zip](https://github.com/openwebwork/webwork2/files/8064171/TestProblems.zip)

The problem tikz_tex_package_test.pg has no random parameters, but has an image.  So the show me another problem should always fail to give a new problem.  However, with the current develop branch (and before) it will still give a new problem (that is exactly the same).  With this pull request it will not.

The problem PointCoordinates.pg has one random parameter with two possible values.  With the current development branch, again it will always give a new problem, and 50% of the time it will actually be the same problem as the original.  With this pull request it will either give the other possible problem, or (in rare cases) it will fail to generate a new problem.  Of course it is possible that in five tries (the default max number of tries) it will always get the original random parameter value.  Note that this will always fail to generate a new problem for a student before solutions are available because the only actual difference is in the solution.  Unfortunately, this is the down side.  Problems whose only difference is the actual image contents will never give a new problem.  That really can't be helped though.  Note the full problem I actually use here also asks for the quadrant/axis the point is in/on in the graph, and that provides a valid difference.  So problems should always be designed to have something different other than image contents if one expects show me another to work in a valid way.

Also add the showMeAnotherBox style back.  That was accidentally removed in the bootstrap update pull request.


pstaabp (approved):
Works as advertised.

PR #1598, opened 2022-02-16
Add permission for ignoring feedback.

tyearke:
Users with this permission will be excluded from receiving feedback emails that they would otherwise get by having the `receive_feedback` permission.

Our use case for this permission is allowing our system administrators to assist faculty with their WeBWorK courses without receiving feedback emails. This pull request does not modify the current behavior by default, but by changing the `ignore_feedback` permission to `admin` in `conf/defaults.config`, it works as desired for this use case.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -770,6 +770,7 @@ $authen{xmlrpc_module}  = "WeBWorK::Authen::XMLRPC";
 	score_sets                     => "professor",
 	send_mail                      => "professor",
 	receive_feedback               => "ta",
+	ignore_feedback                => "nobody",
 
 	create_and_delete_problem_sets => "professor",
 	assign_problem_sets            => "professor",
@@ -1832,6 +1833,11 @@ $ConfigValues = [
 		{ var => 'permissionLevels{receive_feedback}',
 		  doc => 'E-mail feedback from students automatically sent to this permission level and higher:',
 		  doc2 => 'Users with this permssion level or greater will automatically be sent feedback from students (generated when they use the "Contact instructor" button on any problem page).  In addition the feedback message will be sent to addresses listed below.  To send ONLY to addresses listed below set permission level to "nobody".',
+		  type => 'permission'
+		},
+		{ var => 'permissionLevels{ignore_feedback}',
+		  doc => 'Ignore feedback if this permission level and higher:',
+		  doc2 => 'Users with this permission level or greater will be excluded from receiving feedback from students (generated when they use the "Contact instructor" button on any problem page).',
 		  type => 'permission'
 		},
 		{ var => 'mail{feedbackRecipients}',
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -367,9 +367,14 @@ sub getFeedbackRecipients {
 
 	my @recipients;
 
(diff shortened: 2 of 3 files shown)


dlglin:
My concern with this is the scenario when someone sets `ignore_feedback` to a lower permission level than `receive_feedback`.  Which one should take precedence?

One idea that was suggested for a more robust way to handle this would be to modify the `hasPermissions` subroutine to accept either a single permission level from the `%permissionlevels` hash or an array of two permission levels.  If a single permission level is supplied, then it is assumed to be the minimum permission level required for that action, but if two permission levels are supplied, then they will be the minimum and the maximum.  This would be backward compatible with the current `%permissionlevels` hash, while allowing more flexibility.

My workaround with the current code is to simply delete the email addresses of any administrators from their profiles in WeBWorK courses.

Alex-Jordan:
There could be another issue with this in practice. What about the admins who are also instructors? It sounds like if the site-wide setting prevents admins from receiving feedback emails, then an admin who is also an instructor will need to edit their course.conf.

Does WW handle this through permissions to avoid having a database column for this user attribute?

dlglin:
> Does WW handle this through permissions to avoid having a database column for this user attribute?

Which user attribute?  Whether or not the user should receive feedback?

The admin permission level was added long after the `receive_feedback` permission was implemented, so I don't know that anyone thought through the scenarios around administrators receiving student email.

If the ability to receive student emails was configured as a user attribute (and stored in the DB), then it would have to be enabled for the right users when they were added to a course, which means that either the person adding the user or the WW logic would have to make this decision.  The advantage of having it as a permission is that the instructor/administrator does not need to explicitly determine who should receive emails.  The disadvantage is that there is basically no flexibility to include/exclude users.

I do think that being able to exclude administrators from receiving feedback is a useful feature.  There are a few ways to implement it, such as

1. This PR.  The downside being that this introduces opposing permissions, and can lead to the conflict I described earlier.
2. Adding a min/max permission system as I described, which is a limited improvement to the versatility of the permissions system, but could be implemented without a huge amount of work.
3. Deleting the administrator's email address from the course, which already works, but involves manually adding or deleting email addresses from users, and could be problematic if user data is synced with an external system (e.g. using LTI).
4. Adding a black list of emails that should not receive feedback.

None of these automatically address the issue of an administrator who is also an instructor in some courses, though they would all allow it to be handled by editing course.conf.  All the ways that I can come up with to do this more gracefully involve a more significant overhaul of the permissions system (e.g. either allowing a user to have more than one role in a course, or having the permission levels not be hierarchical), which would be a good project, but probably won't happen for WeBWorK 2.

Alex-Jordan:
One more idea to your list: at the moment a course is created and admin accounts are being created in the new course, the system could leave their emails out. They get to keep email addresses in the admin course, without automatically having them appear in the regular course.

> Which user attribute? Whether or not the user should receive feedback?

Yes, that is what I meant.

somiaj:
There are two other ways that can also be used to keep feedback from going to particular admins/profs/etc. The one I use here is (which can be configured under course configuration), is to disable all emails and explicitly enter the emails you want feedback to go to.

```
$permissionLevels{receive_feedback} = nobody;
$mail{feedbackRecipients} = ['email1', 'email2'];
```

Another option is to set `$feedback_by_section=1`, then put admins in different sections than the students are in.

pstaabp:
An alternative of this is #2117.

PR #1599, opened 2022-02-16
Add an asset build system for javascript and css files.

drgrice1:
All javascript and scss/css files in htdocs/themes and htdocs/js/apps are compiled to minimized js or cs files.  The filename of the generated file contains a hash of the contents.  The generated file names are saved to the file htdocs/static-assets.json identified by the unminimized file name.  Webwork then loads this file (and a same named file in pg/htdocs), and any time that a javascript or css file is needed a new method `getAssetsURL` in lib/WeBWorK/Utils.pm is called to get the minimized file name from these lists.  Note the template files call the `url` method of ContentGenerator.pm, which in turn calls `getAssetsURL`.  The file name is now an argument to the `url` method, instead of being appended to the return value of this method.

Most important here is that since the generated file names contain the content hash, this solves the browser caching problem.  Any time the contents of a javascript or css file change, the hash will change.  Since the file name is now different the browser will not look in its cache for the updated file, and will download the new one.

Note that files in webwork2/htdocs and pg/htdocs are directed to their respective actual server urls (/webwork2_files or /pg_files respectively).  So the need for the apache2 rewrite rule is lessened.  It is still left in place as a fallback.

In addition, there is an added file `htdocs/third-party-assets.json` which contains a mapping of files we use in `htdocs/node_modules` (installed via `npm install`) to a corresponding CDN location (on cdn.jsdelivr.net).

When `npm install` is run, the `prepare` script is run, which in turn executes the `generate-assets` script, and that runs the node script `htdocs/generate-assets.js`.  One can also separately run that script with `./generate-assets.js` or `node generate-assets.js`.  If the option `--useCDN` (or `-c` for short) is passed to that script the CDN locations will be used instead of the local files for third party libraries.  Note that the old `generate-css` and `css-prefix` npm scripts have been removed.  Those will no longer work with this scheme, and the `generate-assets` script does everything they did and more.

Developers working on the javascript or css files can also pass the flag `--enable-sourcemaps` (or `-s` for short) to the `generate-assets` script to generate css and js files with inline source maps in them.  Additionally, one can pass the flag `--watch-files` (or `-w` for short) to continue to watch files for changes, and automatically regenerate an updated minimized file.  Of course those flags can be used together.   Finally, by passing the flag `--clean` (or `-d` for short), all of the minimized files from previous builds will be deleted.

Note that all references to the math4-coloring.css file have been removed.  That file is now obsolete.  The _theme-colors.scss file should be used instead.  In addition, the `generate-assets.js` script will create a math4-overrides.css and math4-overrides.js file in each theme directory if one does no exist already.  So this solves the console warnings about those files not being found when the html2xml rendering method is used.  The partial fix for this in the docker-entrypoint.sh file has been removed.  That will be more properly dealt with when `npm install` is run in the docker build.

The `.min` infix is used on the generated js and css files so that these can be ignored in `.gitignore`.

This is paired with the pg pull request openwebwork/pg#657.  However, it will work without that one, except the unminimized and unhashed pg files will be used.


drgrice1:
I should not for those that have already been working with the develop branch since switching to bootstrap 5 that you will probably need to delete the generated files from before.  That is htdocs/themes/*/bootstrap.css, htdocs/themes/*/achievements.css, and htdocs/themes/*/gateway.css, and on the pg side pg/htdocs/js/apps/Problem/problem.css.

Also, if you already have a custom theme, then you will need to rework that a bit.  Instead of links from the derived themes to the css files in math4, there are now links to the scss files in math4.  The css files for the theme are generated from those.

pstaabp:
With your comment above, I decided to delete my custom theme directory (`math4-purple`), then recopy the math4-green directory, then update the `_theme-colors.scss` file.  I reran `npm install` and the related messages is:
```
Proccessing themes/math4-purple/achievements.scss
Proccessing themes/math4-purple/bootstrap.css
Removing old version themes/math4-purple/bootstrap.0a2818d3.min.css.
Error: ENOENT: no such file or directory, unlink '/opt/webwork/webwork2/htdocs/themes/math4-purple/bootstrap.0a2818d3.min.css'
    at Object.unlinkSync (node:fs:1722:3)
    at processDir (/opt/webwork/webwork2/htdocs/generate-assets.js:108:9)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async processDir (/opt/webwork/webwork2/htdocs/generate-assets.js:40:4)
    at async Promise.all (index 0)
    at async build (/opt/webwork/webwork2/htdocs/generate-assets.js:129:3) {
  errno: -2,
  syscall: 'unlink',
  code: 'ENOENT',
  path: '/opt/webwork/webwork2/htdocs/themes/math4-purple/bootstrap.0a2818d3.min.css'
}
```

It looks like it removed the old `.min.css` file then complained that it wasn't there.

pstaabp:
Also, the new theme had the copied colors (green).  A second run of `npm install` has no errors and fixed the theme colors.

drgrice1:
I assume that you ran the script first before deleting your old theme directory.  Then that added the `math4-purple/bootstrap.0a2818d3.min.css` file to the `static-assets.json` file.  Then you recreated the `math4-purple` directory, and ran the script again.  Then the script tried to delete the file in the `static-assets.json` file when it created the new `math4-purple/bootstrap.*.min.css` file.  However, that file doesn't exist, and so it gives an error.  I will add a check for existence before trying to delete.

What do you mean the new theme had the copied colors?

drgrice1:
I added the check for existence before deleting.

pstaabp:
> What do you mean the new theme had the copied colors?

I had copied the math4-green directory to a new directory called math4-purple and changed colors.  I ran `npm install` got the error above.  Navigating through WW keep the green colors.  A second run of `npm install` fixed the problem.

pstaabp:
I'll go through all of the steps again to double check.

drgrice1:
That is because the error prevented the execution of the script from succeeding.

pstaabp:
Another issue: I'm viewing a version of quiz and I get the following:

Warning Message:
```
Use of uninitialized value $file in hash element at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 1861. Use of uninitialized value $file in hash element at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 1869.
```

Error Message:
```
Can't use string ("js/apps/Problem/problem.css") as a HASH ref while "strict refs" in use at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 1878.
```

Call Stack:
```
in WeBWorK::Utils::getAssetURL called at line 2443 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
in WeBWorK::ContentGenerator::GatewayQuiz::output_JS called at line 155 of /opt/webwork/webwork2/lib/WeBWorK/Template.pm
in WeBWorK::Template::template called at line 620 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::content called at line 233 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator.pm
in WeBWorK::ContentGenerator::go called at line 386 of /opt/webwork/webwork2/lib/WeBWorK.pm
```

I reran `npm install` in both ww and pg and restarted the server.  I still see the error.

drgrice1:
That is fixed.  Forgot to add the course environment as the first argument for the call to `getAssetURL` on line 2443 of `GatewayQuiz.pm`.

drgrice1:
I changed the script so that instead of trying to delete files from previous builds from the previous static-assets.json file, it just deletes all files that match `/.[a-z0-9]{8}.min.(css|js)$/` in the `js/apps` and `themes` directory.  So hashed file names that end in min.css or min.js.  This is much more reliable.

pstaabp:
retested creating a new theme and that works fine.  Also no errors on gateway quiz now.

drgrice1:
I made some nice changes to the `generate-assets.js` script.  See the paragraph about development that I edited in the initial comment for this pull request.

htdocs/helpFiles/instructor_links.html
```diff
@@ -23,38 +22,38 @@
 <head>
 <meta http-equiv="content-type" content="text/html; charset=utf-8" />
 <title>Instructor Links Help Page</title>
-<link rel="stylesheet" type="text/css" href="/webwork2_files/node_modules/@fortawesome/fontawesome-free/css/all.min.css"/>
+<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@5.15.2/css/all.min.css"/>
```
pstaabp, line 25: Why is this switched to a cdn?  I know that since this is an HTML, we can't use the getAssets that you just created (which I think handles either CDN or local files). Is this because you don't know where this is, so always use a CDN?

drgrice1:
Yes, exactly.  This is a static html file, so you can't call `getAssetURL`.  Part of the point of this pull request is also to eliminate the incorrect hard coding of the `/webwork2_files` into links.  That is configurable in defaults.config/localOverrides.conf.  So if someone changes that setting, then this file would be broken.  Using a CDN works around that.

drgrice1:
There of course are still several other places where the `webwork2_files` is hardcoded still.  So the problem is still not entirely fixed.

pstaabp (approved):
More overall testing.  Looks great now.

PR #1603, opened 2022-02-18
UI improvements and html validation fixes.

drgrice1:
Here is a detailed list of the changes made:

Add the table-responsive wrapper to the table on the add users page.

Fix the comparison of the request parameter that determines if the "Import from where?" select on the homework sets editor page is multiple or not on page load.  That needs to be a string comparison.  Also, change that to checking that the value is `ne '1'` instead of `eq '8'` in case that in the future the multiple size is changed.  The single size probably won't.

Make the state of the checkboxes and set id select on `Scoring Tools` page (Scoring.pm) persistent when the form is submitted.  Also make it so that the "You must select one or more sets for scoring!" message is only displayed after clicking on the "Score selected set(s) and save to:" button, and not when the page initially loads.

Fix a typo in Achievements.pm.  aria_labe -> aria_label.  Also prefix the id's of the hidden authen inputs in the forms for using achievement items inside the modal dialogs that are embedded in the page when a user has such items available for use.  Really, these hidden forms shouldn't even have id's.  However, at this point some javascript uses that to find these forms.  That can be fixed by searching by name instead, but will take care to make sure all case where these id's are used are found.

Remove the duplicate id "problem_body" in the system.template file.

Remove the `role="navigation"` attribute from the breadcrumb `nav` in the system.template and gateway.template files.  `nav` tags should not have that role.

Remove the `<meta http-equiv="X-UA-Compatible" content="IE=Edge"/>` tag from the system.template and gateway.template files.  Those tags are neither valid nor needed anymore.

Use the `$setID` with underscores for the html id of the set siblings on the ProblemSet page.  The `$pretty_set_id` has spaces which is invalid for an html id.  Do these anchors even need an id?

Move the progress bar out of the siblings `ul` in Problem.pm.  It is invalid html to have that in the list.  Also correct the html markup for the divs that make up the progress bar.  Several were self closing, and then also followed by an end div tag.

Make the `problem-sub-header` `div` in the page title of Problem.pm a `span` instead (but give it the css style display block).  It is invalid html for a `div` to be inside an `h1`.

Remove the underscore from the target attribute for the show me another link in Problem.pm.  Targets should not have the underscore prefix unless they are reserved word targets like `_self` or `_blank`.  Also change that target name to be more consistent with the other target names used by webwork.

Fix the number of columns for the gateway quiz rows that have no versions, or that do not allow the display of scores for the set in Grades.pm.

### DIFF
--- a/htdocs/js/apps/UserList/userlist.js
+++ b/htdocs/js/apps/UserList/userlist.js
@@ -1,23 +1,28 @@
 (() => {
-	const filter_select = document.getElementById('filter_select')
-	const classlist_add_filter_elements = () => {
-		const filter_elements = document.getElementById('filter_elements');
+	const filter_select = document.getElementById('filter_select');
+	if (filter_select) {
+		const classlist_add_filter_elements = () => {
+			const filter_elements = document.getElementById('filter_elements');
 
-		if (filter_select?.selectedIndex === 3) filter_elements.style.display = 'block';
-		else filter_elements.style.display = 'none';
-	};
+			if (filter_select.selectedIndex === 3) filter_elements.style.display = 'block';
+			else filter_elements.style.display = 'none';
+		};
 
-	filter_select?.addEventListener('change', classlist_add_filter_elements);
-	classlist_add_filter_elements();
+		filter_select.addEventListener('change', classlist_add_filter_elements);
+		classlist_add_filter_elements();
+	}
 
 	const export_select_target = document.getElementById('export_select_target');
-	const classlist_add_export_elements = () => {
-		const export_elements = document.getElementById('export_elements');
+	if (export_select_target) {
+		const classlist_add_export_elements = () => {
+			const export_elements = document.getElementById('export_elements');
+			if (!export_elements) return;
 
-		if (export_select_target?.selectedIndex === 0) export_elements.style.display = 'block';
-		else export_elements.style.display = 'none';
-	};
+			if (export_select_target.selectedIndex === 0) export_elements.style.display = 'block';
+			else export_elements.style.display = 'none';
+		};
 
-	export_select_target?.addEventListener('change', classlist_add_export_elements);
-	classlist_add_export_elements();
+		export_select_target.addEventListener('change', classlist_add_export_elements);
+		classlist_add_export_elements();
+	}
 })();
--- a/htdocs/themes/math4/gateway.template
+++ b/htdocs/themes/math4/gateway.template
@@ -6,8 +6,6 @@
 <head>
 <meta charset='utf-8'>
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<!-- This encourages IE to *not* use compatibility view, which breaks math4 -->
-<meta http-equiv="X-UA-Compatible" content="IE=Edge"/>
 
 <!--
 ################################################################################
@@ -73,7 +71,7 @@
 <!-- Breadcrumb -->
 <div id="breadcrumb-row" class="row my-2">
 	<div class="col-12 d-flex align-items-center">
-		<nav id="breadcrumb-navigation" role="navigation" aria-label="breadcrumb navigation" class="w-100">
+		<nav id="breadcrumb-navigation" aria-label="breadcrumb navigation" class="w-100">
 			<ol class="breadcrumb">
 				<!--#path style="bootstrap" text=""-->
 			</ol>
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -708,7 +708,7 @@ input.changed[type=text] { /* orange */
 /* Text colors for Auditing, Current, and Dropped students */
 .Audit { font-style: normal; color: purple; background-color: inherit; }
 .Enrolled { font-weight: normal; color: black; background-color: inherit; }
-.Drop { font-style: italic; color: gray; background-color: inherit; }
+.Drop { font-style: italic; color: #555; background-color: inherit; }
 
 /* Styles for the PGProblemEditor Page */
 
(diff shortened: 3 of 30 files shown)


drgrice1:
This fixes all of the issues that I could find on the student pages.  I fear delving into the instructor and course admin pages.

drgrice1:
I also left the table summaries for now as I am not entirely certain what should be done there.  I believe they should just be deleted.  They describe what the table headers already give information about, as well as telling about the buttons below.  The button descriptions certainly aren't needed.  A table summary should talk about things that aren't even in the table.  Furthermore, when students view the page those buttons are not all there, but the description still is.

drgrice1:
I have also added some fixes for things on the course admin page and in the instructor pages.  There is a lot more work to be done in the instructor pages though.

somiaj:
I was doing some improvements on the StudentProgress table and wasn't sure what to do with the summary either, and was just following what was done elsewhere. Due to your comments, I went with using `aria-describedby` option to use the descriptive text that was above the table as a summary.

My changes are here, https://github.com/somiaj/webwork2/commit/003efb5efb92b146a2e7babce62be7b99846d7ba if you have any suggestions. Also feel free to include them in this if you think it is appropriate.

drgrice1:
I don't think that aria-describedby is correct.  If you want to put a div above containing the description, you could do that, but I don't think that table should have the aria-describedby attribute.  The correct answer is to structure the table in such a way that no description is necessary.

pstaabp:
Anything in particular you'd like us to test for? This mainly seems like some code/html cleanup with little affect on the output.  I have run through most of the pages--everything looks good.

drgrice1:
That is basically it.  If you want to and know how you could run html validation on the pages, and also check accessibility.

pstaabp:
I just tried doing validation with the node package `html-validate` and with default settings, there are a number of standard errors for html validation.  None seem that crucial.  For example, it's looking for a self-close tag on elements without children like `<input>` and expecting double quotes on html attributes--items like that. 

How did you validate?  If there a standard set that we want to use, I can rerun if we want, but I'm not sure this is worth the time. 

I also just found a plugin for accessibility for VSCode, which I have been using for development.  I can grab the page source and create a new html file and evaluate it.  I just ran this on the homework sets page and found a few accessibility issues.  For example, there are three inputs (like Email the Instructor) that is missing an aria-label.

drgrice1:
I validated using two methods.  I used https://validator.w3.org/nu and the Wave extension.

Are you seeing those issues with this pull request or without?  This pull request should fix the self closing tags on `<input>`s.  The validators I used didn't give issues about double quotes on html attributes, and that one is not a valid issue either.  Whichever validator you are using is giving incorrect information on that.  HTML specifications do allow single quotes on attributes.  See https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2.

drgrice1:
As I have said, most likely the table summaries just need to be deleted.  Particularly the one on the homework sets page.  The structure of the table is clear, and should need no further explanation.  Furthermore, the current summary is not even valid for what students see, but that same summary is shown in all cases.

What are you using for checking accessibility?  The https://validator.w3.org/nu site only does html validation.  Also, you should test this for a student page instead of an instructor page.  For that you should not see anything other than the table summary.  Whatever you are using that is giving the logo issue is giving an invalid error.  Everything is correct there.

pstaabp:
In terms of accessibility, looks like Chrome has this new tool built-in called Lighthouse that does checking.  Most of the pages seem to be fine. Here's a few things that came up using that tool:

- On tables with the select all checkbox, there is the accessibility error that there is no label associated with it. (classlist, homework sets, achievements)
- On the user Grades page , it doesn't like the fact that the Grades is in an h1 and the User name is in an h3. Says to not skip heading levels.  
- Same on the Statistics and Student Progress page.
- On Scoring tools, the input box for the file name is missing a label. Can the button be the label?
-  On Email, the Save As: field at the bottom is missing a label.  This seems to be the same as the previous.
- On File manager, the filename on the upload input is missing a label.
- seems like for my chosen theme, a few places has said "insufficient contrast" on some of the buttons. Maybe we make a comment in the release notes about themes. 
- On Set Details: The expand/collapse chevrons on problems has the error: "Buttons do not have an accessible name".
- On Set Details: the restricted release input and the set description is missing a label.

drgrice1:
Lighthouse isn't that new.  I have used that also.  It is actually not as good as the built in accessibility checker of Firefox.  I forget why right now though.  I know that when I compared the two there was something about Lighthouse that was not as good.

>     * On tables with the select all checkbox, there is the accessibility error that there is no label associated with it. (classlist, homework sets, achievements)

Yes, but this is minor.  These checkboxes only appear for instructors, and not for students, and the missing label only applies to the `select all` checkbox itself, and not all of the other checkboxes which are now labeled with this pull request.  I could make the corresponding column header the label for the select all checkbox.

>     * On the user Grades page , it doesn't like the fact that the Grades is in an h1 and the User name is in an h3. Says to not skip heading levels.

I changed this to an h2.

>     * Same on the Statistics and Student Progress page.
>     * On Scoring tools, the input box for the file name is missing a label. Can the button be the label?
>     * On Email, the Save As: field at the bottom is missing a label.  This seems to be the same as the previous.
>     * On File manager, the filename on the upload input is missing a label.
>     * On Set Details: The expand/collapse chevrons on problems has the error: "Buttons do not have an accessible name".
>     * On Set Details: the restricted release input and the set description is missing a label.

These are all issues on instructor pages, and so are not as signification.  These issues were not fixed because they are more difficult to fix without messing up display formatting and other things.

>     * seems like for my chosen theme, a few places has said "insufficient contrast" on some of the buttons. Maybe we make a comment in the release notes about themes.

Is this a custom theme of yours, or one of the built in themes?  If it is a built in theme then I will fix it.  If it is your custom theme, then you need to fix it.  Also, which buttons?  Some disabled buttons give this warning, but ignore that warning for now.

In short, you haven't found anything that I didn't already know about.

PR #1605, opened 2022-02-28
add height and width to CGI::img command for pi.svg

eltenedor:
if not added, the pi svg spans as much space as allowed by the sourrounding container. Not sure if this is the best approach in terms of responsiveness.

Bases of the changes introduced in commit 8a9e3e3cf594c2376c8985a046805fe39956edb1

Makes the symbol bar in the library browser look like

image

instead of

image

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
@@ -1392,7 +1392,9 @@ sub make_data_row {
 			},
 			CGI::img({
 				src => $r->ce->{webworkURLs}{htdocs} . '/images/pi.svg',
-				alt => $r->maketext('Uses Math Objects')
+				alt => $r->maketext('Uses Math Objects'),
+                                width => '10',
+                                height => '10'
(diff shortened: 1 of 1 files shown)


drgrice1:
This is not correct and not needed.  The reason you are seeing this is probably because your css is not up to date or because your browser has cached css that is being loaded.  Run `npm install` in htdocs and clear your browser cache.

drgrice1:
It looks like you have some other issues on your system too.  The word "edit" should not appear beside the pencil icon.

drgrice1:
Note that the pertinent style setting is in `htdocs/themes/math4/math4.scss` at line 627.  Any `img` tag in the `.lb-problem-icons` div has a max width and height of `15px`.

drgrice1:
I see why the word "edit" is appearing for you.  This is also because you need to run `npm install` in the htdocs directory.  You are missing the `.sr-only-glyphicon` style.

eltenedor:
Dear @drgrice1 thank you for the hints and your quick reply on my request. Indeed, running `npm install` fixed all mentioned issues for me.

Issue #1240, opened 2021-03-08
Feature requests for manual problem grader


Alex-Jordan:
Here are some feature requests passed up to me from local faculty using the new manual problem grader. Recording them here so they are not lost.

> 1) It would be helpful if the "Previous Test", Student, and "Next Test" buttons were also at the bottom of the screen since you usually work down and end at the last question and ready to move on.

> 2) I wish it would retain the options that are checked at the bottom of the quiz as you move to the next student (i.e. "Show correct answers", "Show solutions", "Show Problem Graders")

drgrice1:
It does retain the "Show Problem Graders" option. Note that that option implies the "Show correct answers" option.  It does not retain "Show solutions" at this point.

drgrice1:
Note that request number 2 is fully implemented now.

Request 1 is more difficult, and at this point not entirely possible to implement with the current setup.  The nav is set up by a call from the templates.  The template could call that again, but would cause rather severe database thrashing.  The nav has to load information about all users that are assigned to the current set from the database.  Currently there is no way to just put that nav on the page twice.  There would probably also be issues with html id clashing.

Alex-Jordan:
Would it work to give that `div.user-nav` sticky CSS so that it stays attached to the top of the page as you scroll down?

Alex-Jordan:
Sorry, not requesting that you try that unless you happen to be active in the vicinity. I have to catch up on grading today and this weekend, and I'm still not really free for better WW contributions until mid June.

somiaj:
I've been using the manual grader a while and like it. Thanks for the feature. I am teaching a course with multiple instructors and sections so it would be nice to filter the students listed in the manual grader by section. So have a second drop down menu next to the student list with `all` followed by a list of sections. Selecting a section only shows students in that section and prev/next cycle through that list.

drgrice1:
@somiaj :  I don't think that adding another dropdown to the stick nav is a good idea.  That would mess with the symmetry of things in the sticky nav as well as start to make it crowded, particularly on narrow screens.  A better approach would probably be to put another dropdown somewhere else on the page.  That dropdown does not need to be in the sticky nav as that would be something that would not be changed for every problem or student.

somiaj:
@drgrice1, I'm okay with having the option to filter students by section (or maybe even other ways) in another location, I just don't know the layout enough to have an idea where it would best be to put those things (maybe a hidden div you can open with advanced options?)

[...]

PR #1607, opened 2022-03-02
Section/recitation filter for the sticky student nav

drgrice1:
If a course has sections or recitations,  then add a dropdown to the right of the previous dropdown in the student nav that appears when an instructor is acting as another user and is viewing a problem in a problem set or viewing a gateway quiz version.

This dropdown lists all sections and recitations for users assigned to the set on the problem page and for taken test version on a gateway quiz page.  If the section or recitation is clicked on, then only those users/tests will be shown.  There is also an item added in the dropdown when a filter is active that if clicked will remove the filter.  The active filter will be persistent when navigating between problems and students, or when a problem or gateway quiz submit button is pressed.

This is a feature that was requested in #1240.

In addition the order of the things in the sticky nav is changed on the problem page.  Now the "Previous Problem", "Problem List", and "Next Problem" buttons are on the left (or top on narrow screens), and the student nav is on the right (or bottom on narrow screens).  This makes the "Previous Problem", "Problem List", and "Next Problem" buttons appear in the same place that they were before the student nav was added.

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -380,6 +380,7 @@ h2.page-title {
 .sticky-nav {
 	display: flex;
 	flex-wrap: wrap;
+	gap: 0.5rem;
 	align-content: space-between;
 	justify-content: space-between;
 	z-index: 1;
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1233,94 +1233,134 @@ sub nav {
 		my $maxTestIndex = $minTestIndex + 200 < $#userRecords ? $minTestIndex + 200 : $#userRecords;
 
 		# Set up the student nav.
-		print join(
-			'',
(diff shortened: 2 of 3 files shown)


pstaabp:
The functionality works great.  It's a bit of a hidden feature (which may be intended) in that the right-most dropdown is there:
image

but not clear what it does until clicking on it.  Can we add a label?

I'm guessing you were thinking that it might clutter the top to include, but maybe even a label like "Section:" would be clearer but not too intrusive?

drgrice1:
I wanted to add a tooltip, but unfortunately a dropdown button can not also have a tooltip.  The whole point here was to make this as unobtrusive as possible.  A label makes things to wide and causes display issues.

drgrice1:
@pstaabp:  I switch to the chevron left/right button group and separate drop down for the filter as we discussed.

drgrice1:
I also added a vertical gap between the problem selector part and the student selector part of the sticky nav on the problem page.

pstaabp (approved):
This looks good now.  I think the chevrons are clear and it responds well to small screens.

PR #1608, opened 2022-03-02
Improve the PG Problem Editor and Achievement Editor with CodeMirror options

drgrice1:
Add an option to turn spell checking on or off, an option to select the CodeMirror theme, and an option to change the keymap.
    
These options are saved in local storage, and so are persistent as long as you are using the same browser and device.

The javascript or css for the keymap or theme is loaded dynamically.  The initial approach was to blindly load all of it.
    
Also, the limitation on the width of the codemirror editor has been increased.  The size will now be as large as it can up to 1000px.
Edit:  On second thought, I removed the width restriction.  The codemirror editor will now expand to fill the available space.  Since I also added the `resize: vertical` css restriction which does not allow horizontal resizing the width restriction is annoying if you want to make the viewing area wider.  Without the restriction you can enlarge your browser or close the site navigation on the left to make the viewer wider.  You can always use the handy dandy "Enter" button to not make such long lines.
    
Note that the code that was duplicated for the Achievement Editor and the PG Problem Editor is now moved into a separate file that is called by both of those modules.


drgrice1:
Note that since this pull request depends on #1599, you will also need openwebwork/pg#657 to test it.

pstaabp (approved):
This looks good.  Safari seems to be hit or miss with this. I think I saw this issue before with #1577.   It seems like something out of our hands.  I might put an issue on CodeMirror, but maybe we should note this in the release notes.

drgrice1:
Which part of this is hit or miss for Safari? Is it the spell check, the themes, or the key mappings? Or is it all of the above?

somiaj:
Tested this out and everything works as advertised. Two small things: (1) the `until` keyword isn't highlighted (not something new, just something I have noticed lately when editing problems) (2) A help (?) popup that states what the key bindings are or a link to the CodeMirror docs could be useful to users.

drgrice1:
1. I didn't change any of the syntax highlighting, and that is not part of this pull request.
2. If anyone does not know what those keybindings are, they should not use them.  They should stick with the default.  CodeMirror does not provide documentation as to what those are either.  Nor should they.  CodeMirror did not create those editors.  I don't really think a help popup would be as useful as you think.

drgrice1:
@somiaj:  I added `until` to the syntax highlighting list.  I decided that was an easy thing to do.

somiaj:
Thanks. I only mentioned the `until` issue as it was very minor and thought something that didn't need its own pull request.

You maybe correct that adding some help with even the default key bindings wouldn't be that useful. It was just an idea I thought some users may find useful. There is also `helpFiles/InstructorPGProblemEditor.html` as a possible place to add some info about the default keybindings. Though at this point probably not worth adding in this pull request.

The highlight of `until` works fine in my tests.

drgrice1:
@somiaj:  I added some help for the new options to the help file.  I didn't think about adding that there.  It does not go into detail about what the key maps are, but gives a brief description of the default key map and that it is basically the browser text area.  It is way beyond even the scope of the help file to go into a lot of detail on that, and certainly to go into detail on emacs, sublime, or vim.

pstaabp:
This all still looks good.  Safari is still a bit wonky with spell checking, but not sure there's anything to do about that.

pstaabp:
I think this is ready to go.

PR #1609, opened 2022-03-04
use best version score for conditional release

drdrew42:
When using the conditional release feature, dependencies on _non_-versioned problem sets have not worked as expected.

This PR changes the logic of `is_restricted` in WeBWorK::Utils to loop through all versions of any versioned set dependency in determining whether or not the required score has been attained.

### Replication
1. Create two sets: a gateway set (GW) and a homework (HW) and add 1 'blankProblem.pg' to each.
2. Enable conditional release in Course Configuration
3. Modify HW to require GW as a conditional release dependency, and require 100% score
4. It helps if GW has only one graded attempt per version, with unlimited versions available
5. Assign both sets to a student user (proper dates etc)
6. Score zero points on version 1 of GW, then generate a new version and score 100% for version 2
7. See that HW is (without this PR) locked even though required score is achieved
8. See that HW is (with this PR) available since required score has been achieved

@dlglin I await your feedback ;P

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -1234,10 +1234,11 @@ sub is_restricted {
 			my $restrictor_set = $db->getGlobalSet($restrictor);
 
 			if ($restrictor_set->assignment_type =~ /gateway/) {
-				my @versions = $db->listSetVersions($studentName, $restrictor);
+				my @versions =
+					$db->getSetVersionsWhere({ user_id => $studentName, set_id => { like => $restrictor . ',v%' } });
 				foreach (@versions) {
-					my $restrictor_set_version = $db->getSetVersion($studentName, $restrictor, $_);
-					my $v_score = grade_set($db, $restrictor_set_version, $restrictor, $studentName,1);
+					my $v_score = grade_set($db, $_, $restrictor, $studentName, 1);
+
 					$r_score = $v_score if ($v_score > $r_score);
 				}
 			} else {
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -1212,30 +1212,43 @@ sub has_aux_files ($) { #  determine whether a question has auxiliary files
 }
 
 sub is_restricted {
-        my ($db, $set, $studentName) = @_;
+    my ($db, $set, $studentName) = @_;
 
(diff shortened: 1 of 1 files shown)


drgrice1 (commented):
Looks good.

One minor code cleanup suggestion though.  Can you add spaces after the commas on lines 1237, 1239, 1240, and 1244?

In testing this I noticed an issue on the Grades page.  This is not something created by this pull request, and happens for regular homework sets dependent on other regular sets also.  On the grades page there is a link into homework sets, and that link is presented even if a student has not satisfied the prerequisites for the set.  The link does not work and gives a "WeBWorK error" page with the error message "You do not have permission to view unopened sets ...", but it should be the same as on the homework sets page and not have the link in this case.

drgrice1 (approved):

lib/WeBWorK/Utils.pm
```diff
@@ -1212,30 +1212,43 @@ sub has_aux_files ($) { #  determine whether a question has auxiliary files
...
+
+		foreach my $restrictor (@good_sets) {
+			my $r_score = 0;
+			my $restrictor_set = $db->getGlobalSet($restrictor);
+
+			if ($restrictor_set->assignment_type =~ /gateway/) {
+				my @versions = $db->listSetVersions($studentName, $restrictor);
+				foreach (@versions) {
+					my $restrictor_set_version = $db->getSetVersion($studentName, $restrictor, $_);
+					my $v_score = grade_set($db, $restrictor_set_version, $restrictor, $studentName,1);
```
drgrice1, line 1240: I missed this before.  This can be done with one database query.  Instead of listing and then retrieving each set separately, just do:
```suggestion
				my @versions =
					$db->getSetVersionsWhere({ user_id => $studentName, set_id => { like => $restrictor . ',v%' } });
				foreach (@versions) {
					my $v_score = grade_set($db, $_, $restrictor, $studentName, 1);

```

drgrice1 (commented):
I missed some extra database calls that aren't needed.

lib/WeBWorK/Utils.pm
drdrew42: Excellent, I was not thrilled about all the queries. I actually forgot that your updates were available now!

pstaabp (approved):
Looks great!

PR #1610, opened 2022-03-13
Fix the instructor tools buttons for large selections of users or sets.

drgrice1:
See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=6639#p17921.

The problem is that the form on that page was using the `GET` method, and then redirecting to these pages.  A `GET` request uses url parameters to submit data, and the length of a url is limited to 8190 bytes.  This limitation can easily be exceeded if enough users or sets are selected.  So this pull request switches the form to using the `POST` method.  The data in a `POST` request can hold approximately 2 GBs which should be more than enough for even a large class.

The submit buttons that use multiple students or sets are switched to a direct `POST` instead of using a redirect.  The parameters are adjusted via javascript if needed.

For all of the submit buttons the form is checked for the requisite data needed for that request and does not allow a submission if it is not present.  The error messages are still displayed as before, they are just added immediately via javascript without a submission.  Note that these error messages are also now translated.

There is also a minor fix to the javscript in htdocs/js/apps/UserList/userlist.js.  When editing users the elements with the filter_select and export_select_target ids are not present. That was already checked for, but in that case the elements with the ids filter_elements and the export_elements are also not present which was not checked.

### DIFF
--- a/htdocs/js/apps/InstructorTools/instructortools.js
+++ b/htdocs/js/apps/InstructorTools/instructortools.js
@@ -1,0 +1,108 @@
+(() => {
+	const form = document.forms['instructor-tools-form'];
+
+	form.addEventListener('submit', (e) => {
+		const selectedUsers = Array.from(document.querySelector('select[name=selected_users]')?.options ?? [])
+			.filter((option) => option.selected);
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
I didn't try anything with a large set, but everything works fine.

PR #1613, opened 2022-03-14
Add a gateway quiz 'Grade Test' confirm dialog

drgrice1:
Add a confirmation dialog that is shown when a student clicks on 'Grade Test' in a gateway quiz.  The dialog is not shown for gateway quizzes that allow an infinite number of submissions per version, and is not shown in the last 90 seconds of a timed quiz (that is when the other alerts are shown telling the student to submit).  The dialog asks the student if he/she really does want to submit the test, and informs the student of the number of remaining submissions, and consequences of submitting the test.

Also remove the remaining inline javascript in the GatewayQuiz.pm module (for the page change links and per problem preview buttons).

### DIFF
--- a/htdocs/js/apps/GatewayQuiz/gateway.js
+++ b/htdocs/js/apps/GatewayQuiz/gateway.js
@@ -110,6 +110,15 @@
 		}
 	};
 
+	// Set up the preview buttons.
+	document.querySelectorAll('.gateway-preview-btn').forEach((btn) => {
+		btn.addEventListener('click', (evt) => {
+			// Prevent the link from being followed.
+			evt.preventDefault();
+			if (btn.dataset.currentPage) document.gwquiz.newPage.value = btn.dataset.currentPage;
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
This looks great.

somiaj:
Tested and works as advertised. Good addition.

drgrice1:
You should see the dialog if you hit "Grade Test" with more than 90 seconds remaining.  After that it won't show.

You also won't see the dialog at all if there are an unlimited number of submissions allowed per version.

drgrice1:
@pstaabp:  At one point I was testing something and changed the check to "less than 20" seconds for the dialog.  It seems that I forgot to change it back, and that is why you were still seeing the dialog within 90 seconds of the test ending.  That has been changed back now.

pstaabp (approved):
The alert at 90 seconds is working now as advertised.

drdrew42 (approved):

PR #1614, opened 2022-03-17
Add right-to-left language support to the bootstrap update.

drgrice1:
Bootstrap does not use the same css for right-to-left languages that it uses for left-to-right languages.  Furthermore, some of our local css does not work correctly for right-to-left languages (at least since the bootstrap update).

This adds generation of right-to-left css to the build system.  When scss and css files are processed by generate-assets.js, in addition to generating a minimized and prefixed css file it also generates a rtl variant.  When the lib/WeBWorK/Utils.pm method is called to get a static asset location, it now also checks for a rtl variant when a right-to-left language is enabled.

To make this work css minification is no longer done by sass.  For the right-to-left css we first need to run rtlcss, then autoprefixer, and then minification is done by cssnano.  All of that is done with postcss. Of course for rtl css the rtlcss step is skipped.

openwebwork/pg#663 adds rtl styles for problems.

### DIFF
--- a/htdocs/generate-assets.js
+++ b/htdocs/generate-assets.js
@@ -11,6 +11,8 @@ const crypto = require('crypto');
 const sass = require('sass');
 const autoprefixer = require('autoprefixer');
 const postcss = require('postcss');
+const rtlcss = require('rtlcss');
+const cssMinify = require('cssnano');
 const thirdPartyAssets = require('./third-party-assets.json');
 
 const argv = yargs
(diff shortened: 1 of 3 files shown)


pstaabp (approved):
This and the corresponding PG PR both look good.  Tested in Hebrew.  There are a number of places that we need translation.

drgrice1:
Yeah, at the end of this week I will run the `bin/update_localization_files` script and make a pull request with the result.  We need to automate that and set that up to work with Transifex (or another such service).

PR #1615, opened 2022-03-17
Use time strings formatted for the locale for homework set dates.

drgrice1:
Instead of formatting the unix timestamp for display in the inputs for dates in the instructor tools, just use the unix timestamp in those inputs.  The flatpickr javascript now hides those inputs and adds an input that the user actually sees.  That input contains a localized time string that is formatted for the locale (language) of the course via native javascript's Intl.DateTimeFormat.  The flatpickr javascript takes care of updating the value of the hidden input with a unix timestamp as the user changes the visible input.

To make this work well the dates that are only used for display are formatted with a different approach by Perl.  The DateTime format_cldr is used for this instead of strftime.  This allows for the use of the DateTime::Locale methods which provides uniform formatting for different languages.  A special case is added to the lib/WeBWorK/Utils.pm formatDateTime function to handle this.

To be precise, javascript uses Intl.DateTimeFormat with the options `{ dateStyle: 'short', timeStyle: 'short' }`, and perl uses the formatting string returned by `$dt->locale->datetime_format_short`.

Edit:   Now the javascript instead uses the `luxon` date time library with carefully chosen formats that match the perl formats.

This also makes things work cleaner in the way that the server handles dates because you don't need to worry about parsing a date string back to a unix timestamp.

For most languages the perl DateTime::Locale formatting and javascript formatting are the same.  Unfortunately, for a few languages Perl's DateTime::Locale formatting is not correct.  This can be seen when editing a problem set values for a user.  The user values are formatted by javascript, and the class values by perl.  The languages that seem to have problems here are any of the Asian languages.  In Korean javascript gives `22. 3. 16. 오전 12:00` and perl gives `22. 3. 16. AM 12:00`.  Clearly perl is not properly translating the AM/PM part.  For zh_CN javascript gives `2022/3/16 00:00` and perl gives `2022/3/16 上午12:00`.  Javscript gives a 24 hour time, and perl uses a 12 hour time.  For zh_hk javascript gives `16/3/2022 上午12:00` and perl gives `2022/3/16 上午12:00`.  So perl does not seem to use the correct ordering of the year, month, and day for that locale.

Edit:  With the recent changes to this pull request, only the Korean language still has differences between the perl and javascript.

Of course this means that the old strftime format `%m/%d/%Y at %I:%M%P` is not used for set dates anymore.  For English the CLDR format returned by `datetime_format_short` that is now used is `M/d/yy, h:mm a`.  This means instead of `03/16/2022 at 12:00am` you get `3/16/22, 12:00 AM`.  I would prefer `03/16/2022, 12:00 AM`, but that is what our locale gives consistently for javascript and perl.  So it is probably best to go with the standard.

Note that the old format is still used for set export and import files. In fact that is the only place that the parseDateTime method is still used.  It is used there to ensure that the correct timezone is set for the saved dates.

Note that the `useDateTimePicker` option has been removed.  That could be re-implemented if desired, but it would be a bit tedious to do so. The problem is that the date/time picker does not format and parse dates, but if that option is disabled then you would need to.  I don't really see the need for that option anymore though.

There is a minor change in the lib/WeBWorK/Utils/LanguageAndDirection.pm file.  That was using the po filenames directly for the html `lang` attribute.  Those filenames use underscores which are not valid for the `lang` attribute.  So the underscores are translated to dashes.

Edit:  The above comment is no longer valid, and was removed from the code.

Note that this pull request replaces #1333, and it what I meant in the discussion there as to the correct way to do this.


pstaabp:
This seems to work as much as I can tell.  I would love to see someone who uses this in a non-English language test it out though.

robert-marik:
Hello, thank you for working on this issue. I will check the patch. However, I do not have last WeBWorK version running and after upgrade my apache2 did not start. Perhaps some missing dependencies. Need a couple of days to resolve. I will let you know.

drgrice1:
@robert-marik:  Yeah, there are several changes that need to be accounted for to get things up and running.  We need to get that documented.  Note that there are some changes to `webwork2/conf/webwork.apache2.4-config` that you will need to make (see the dist file).  You will need to run `npm install` in both `webwork2/htdocs` and `pg/htdocs`.  You will need to make sure the perl `Email::Stuffer` package is installed (`sudo apt install libemail-stuffer-perl` will do this).

taniwallach:
I'm still new to the new "asset build system".
In order to test this, I found that I needed to manually run `./generate-assets.js` in the `/opt/webwork/webwork2/htdocs`.
For some reason, running `npm install` did not seem to do that.

---

When I first tested this, I had not set the timezone for the course in the test course I was using nor in `docker-compose.yml` in Docker, and as a result the code was using a different time-zone in the "date picker tool" and in what was being shown in the input box. The tool was apparently using my local time zone, and converting into the default `America/New_York` timezone (7 hours earlier at present) for the value shown in the input box.

After setting `$siteDefaults{timezone}` for my timezone, the date-picker seems to work properly.

However, intentionally setting a course to a different timezone than the local timezone, which is a use case my server supports - it becomes very confusing to use the date-picker, as the tool is certainly using the local timezone, while the input box/stored setting is using the course time-zone.

drgrice1:
I am not sure what you are saying.  It seems to be working correctly for me.  If the course default is the site default in `$siteDefaults{timezone}`, the I see the time in that time zone.  If the course default is something different, then I see the time in that different time zone.

drgrice1:
Also, you should not need to separately run `./generate-assets.js`.  It should run when you run `npm install`.  What versions of npm and node are you using?  Versions of npm and node prior to the ones on Ubuntu 20.04 may not do this, but I don't think they will work for other things as well.  If you are using Ubuntu 18.04 you will need to install newer versions of npm and node than those in the Ubuntu repositories.  See the "unpublished" release notes at https://webwork.maa.org/wiki/Release_notes_for_WeBWorK_2.17.

taniwallach (changes requested):
I am not sure what to recommend to fix this, but this code behaves **very problematically** if a user needs to change a date in a course whose timezone is **different** from his **local** \(the browsers) timezone.

For example, set a course to a time-zone different that your and then edit a date by changing the values of the minutes by a small shift. After saving the change - the hour will be changed. (Pressing tab has essentially the same effect.) It seems that this is a result of the "text value" being parsed in the browser's local timezone and then converted into the course time-zone.

Apparently, flatpickr is known to be quite problematic in terms of time-zone support.
See:
  - https://github.com/flatpickr/flatpickr/issues/930
  - https://github.com/flatpickr/flatpickr/issues/975
  - https://github.com/flatpickr/flatpickr/issues/1532

taniwallach:
> Also, you should not need to separately run `./generate-assets.js`. It should run when you run `npm install`. What versions of npm and node are you using? Versions of npm and node prior to the ones on Ubuntu 20.04 may not do this, but I don't think they will work for other things as well. If you are using Ubuntu 18.04 you will need to install newer versions of npm and node than those in the Ubuntu repositories. See the "unpublished" release notes at https://webwork.maa.org/wiki/Release_notes_for_WeBWorK_2.17.

I was running `npm install` inside the Docker container (Ubuntu 20.04). It did not seem to run the code to generate the assets.
```
root@myhost:/opt/webwork/webwork2/htdocs# npm --version
6.14.4
root@myhost:/opt/webwork/webwork2/htdocs# npm install
npm WARN lifecycle webwork.javascript_package_manager@~prepare: cannot run in wd webwork.javascript_package_manager@ npm run generate-assets (wd=/opt/webwork/webwork2/htdocs)
npm WARN bootstrap@5.1.3 requires a peer of @popperjs/core@^2.10.2 but none is installed. You must install peer dependencies yourself.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.3.2 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

audited 128 packages in 5.075s

27 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities


```

taniwallach:
Here is a movie of the unusual behavior when the browser timezone and course timezone are different.

https://user-images.githubusercontent.com/39089094/161444790-02dab4c9-3d77-49a3-9e25-f0286ab7ca33.mp4

drgrice1:
I can't see your video.  It seems to be corrupt.  Although, now I see your problem.  If you open the time picker and select a time, that time is not what ends up in the input.  It gets adjusted by the time zone.  I will see what I can do about that.

drgrice1:
By the way, it isn't flatpickr that is the problem.  flatpickr is already in use in the develop branch, and this does not happen there.  It is the locale time formatting that I added in this pull request that is the problem.

taniwallach:
> I am not sure what you are saying. It seems to be working correctly for me. If the course default is the site default in `$siteDefaults{timezone}`, then I see the time in that time zone. If the course default is something different, then I see the time in that different time zone.

The format of the time for the local seems fine. When things first load - the times are being displayed in a format which the local seems to have set.

I am running into strange behavior when making changes.
1. If the browser timezone is different from the course timezone - the flatpickr times and input box times are not in agreement. See the first video below. Watch the hour in both locations.
2. With the course (and my local) timezone as `Asia/Jerusalem` I am seeing date/month flipping roles when making changes by  hand. See the second video below. Pay attention to the 3.5 and 5.3 flipping.

https://user-images.githubusercontent.com/39089094/161445340-529100f6-a592-4516-84e8-35e67711bf1d.mp4

https://user-images.githubusercontent.com/39089094/161445347-b3795ad9-37bf-4820-aeb1-159097f77b04.mp4

drgrice1:
Your videos are all corrupt.  I can't see any of them.

taniwallach:
> Your videos are all corrupt. I can't see any of them.

Sorry. They work for me in Chrome and when downloaded from the file. Firefox reports for me that they are corrupt.

I'm sending a few screen-shots of some of the problems - from a course set to English. The timezone of course (`America/New_York`) is not the "browser" timezone (Israel), and that different triggers for the "hour" to be different differ between the chooser and the textbox, and to jump on changes based on the "shift" between the browser timezone and the course timezone.

The course timezone is seen in: Timezone-01

Just after requesting to edit:  Timezone-02-before-change

After changing the time by hand (before tab or save): 
Timezone-03-after-edit-before-tab

After hitting tab **note that the day of month and hour both changed**: Timezone-04-after-hitting-tab

After save: Timezone-05-after-save

https://user-images.githubusercontent.com/39089094/161445544-51692bfe-f924-42e7-8943-b5fb6fc36a47.mp4

drgrice1:
You can stop posting videos and pictures.  Did you see https://github.com/openwebwork/webwork2/pull/1615#issuecomment-1086934947 after your first video?  I already said I see your problem.  I didn't need any of your videos or pictures.

drgrice1:
Also, did you see https://github.com/openwebwork/webwork2/pull/1615#issuecomment-1086935295?  On the develop branch this is not a problem, and that uses flatpickr also.

taniwallach:
Chooser vs. input box difference: 
Timezone-06-chooser-vs-input-box

The course time-zone (`America/New_York`) is 7 hours earlier than the system/browser timezone (`Asia/Jerusalem`) at present, which explains the 7 hour difference.

When I manually edit the minutes and hit tab, the hours gets decreased by 7, as apparently the "new time" is treated as being given in the "browser time zone" and converted to the "course time zone" for the updated display (when tab is pressed or save is done). I assume that the conversion is first to the hidden Unix timestamp value (second since the epoch) and then converted back to an updated "display string".

drgrice1:
@taniwallach:  Are you completely ignoring what I say?

taniwallach:
> @taniwallach: Are you completely ignoring what I say?

Not intentionally. Half asleep and was trying to send the images since I saw you said the video was broken.

drgrice1:
@taniwallach:  The timezone issue is fixed now.  Please test again.  Note, I am not able to reproduce the month/day flipping issue that you show in your video.  Although I do see some odd behavior in the input.  If I try to select the shown date, it seems to do odd things.  Does "dir=ltr" need to be set on the input for right-to-left languages?

drgrice1:
There is another issue that I am working on.  The Date.parse method isn't working as I had thought.

taniwallach:
> @taniwallach: The timezone issue is fixed now. Please test again. 

I no longer see the time shifting or a difference between the hour in the tool and in the input box. That fix seems to work. Thanks!

> Note, I am not able to reproduce the month/day flipping issue that you show in your video. 

I still see it, but only when the course language is set using `$language = 'heb';` or `$language = 'he-IL';`.

Maybe the flipping issue is related to the matter with Date.parse you said you are working on.

> Although I do see some odd behavior in the input. If I try to select the shown date, it seems to do odd things. Does "dir=ltr" need to be set on the input for right-to-left languages?

Bidirectional behavior is always quite strange and annoying. The dir atttribute will set the default text direction, and that has an influence of how the arrow keys behave, etc. Internally, the Unicode bidi algorithm will influence how things are actually displayed.

For boxes where the contexts is certainly primarily "left to right" text forcing `dir="ltr"` usually improves things. For example, I keep getting annoyed at the the input box for the problem path in the set editor, which should be forced to LTR, as the contents are always in "english" (latin) letters and the current inherited RTL primary direction is very annoying.

For an input box displaying/setting a date and time - I'm really not certain what the best setting is for a RTL language. From what I see for "he-IL" Unicode bidi will still format the "date" portion and "time" portion individually as LTR, so the difference is on which side of the date will the time be displayed. I think the end users (the teaching staff) will handle whichever setting is in use. I do not think the HTML `dir` attribute would be related to the "month-day flipping" I have seen.

taniwallach:
I tried setting `$language = 'ar';` to see how the formatting of the dates in Arabic, also a right-to-left language. I'm not familiar with what would be the desired display, just wanted to see if the date/time flipping was also happening in Arabic and to what extent other behavior seems consistent.

I see a totally different visible format in the list of all sets in the homework editor page and when trying to edit a date/time. For some reason the year is being places onscreen in 2 different places. However, cut-and-paste of the date+time into vim gives the same string result in both cases. I have no idea what causes this.

Arabic-1

Arabic-2

taniwallach:
@drgrice1 I really appreciate the time and effort you are putting into trying to get the date/time display/editing behavior to be more friendly to international users of WeBWorK.

---

> There is another issue that I am working on. The Date.parse method isn't working as I had thought.

It seems that JavaScript's `Date.parse` is considered relatively unreliable due to lack of cross-browser consistency for many formats, and due to deviations between the ECMA and ISO standards on what it should do for certain formats. See:
  - https://stackoverflow.com/a/20463521/18664912
  - https://alanstorm.com/javascript-date-string-parsing/

I am skeptical that it is worth the effort to find some approach which fixes the proposed code to work well for all locales if it cannot be easily provided using some existing library.

---

The idea of using Unix time as the main back-end data/time format inside the JavaScript code seems to make lots of sense. 

I think it is critical that making changes by hand to the date+time string be a simple and dependable manner of editing the datetimes, whether by making small edits by hand or by pasting in a validly formatted value from a different input box (or elsewhere).

If formatting/parsing locale formatted strings is getting too messy and complicated, it could be that settling for a for more limited set of formatting/parsing options for the date/time editing features might be a more feasible goal.

Even a simple choice between `MM/DD/YYYY` and `DD/MM/YYYY` (without any support for locale based customization) during formatting/parsing would already be a significant improvement for international users who find the current `MM/DD/YYYY` format in the editors to be confusing. (Maybe that could be done using an on-screen select element.) The setting or default setting could be set in either the `simple.conf` or `course.conf` file to override the system default value.

Alternately, modifying the code in this PR to provide support via a new setting to force the use of one of a given set of "trusted" formats would help for any locale where the current code does not behave well.
  - It would be nice if both DD/MM and MM/DD formats could be supported as trusted formats when such a special setting is being used. That would make things less confusing for many international users.
  - A special value of the setting could be used to indicate that the behavior of the "primary" code from this PR should be used (even if potentially inconsistent).
  - For course languages (such as Hebrew) where problems are occurring with the current code - such a setting can be forced if no setting was provided by the regular chain of configuration files.

I think it might be helpful if any case where a "parse" error occurs triggered some visible "alert" so that power users did not miss noticing whatever fallback behavior occurs on a parsing failure.

Another potential approach would be to make the "locale formatted" string read-only, and to provide a second editable box which uses a "safe" format which can be consistently generated/parsed. Having both a read-write box with an "ugly" but safe format and a read-only box with the (hopefully) nicely formatted locale controlled format is probably quite an ugly approach visually but may be relatively easy to implement.

---

Here are ideas for two small change which would help international / cross-time-zone staff users.

1. Provide an eye-catching on-screen reminder of the manner in which the dates are being formatted and parsed.
   - At present, we just expect users to know what format (US style) is being used in the instructor tools.
   - In the future, if the location of the month digits can differ from course to course - there can be ambiguity in understanding the dates displayed (unless the flatpickr GUI tool is opened).
   - Reminding the users where the month digits are being placed will help prevent confusion.
2. Provide an on-screen reminder of the course time-zone on instructor pages where date/times can be edited.
   - Maybe even elsewhere.

drgrice1:
This is fixed and working quite nicely now.  The times are displayed in the course timezone and change appropriately when using the popup time picker or when typing in a date.  The javascript language formats now all match the perl DateTime::Locale formats.  With the exception of the Korean AM/PM (perl should translate those, javascript does).

This adds luxon as a javascript dependency to deal with parsing and formatting the localized time strings.

Note that you need to make sure to run `npm install` again (and `generate-assets.js` if it doesn't run when you do that).

@taniwallach:  I will look into why `npm install` is not executing the prepare script when run in a docker container.  I suspect it has to do with running `npm` as root, which always seems to be problematic.

drgrice1:
@taniwallach:  You never answered my question about the `dir=ltr` setting on the date/time inputs.  Is that needed?  It seems that behavior is quite odd as it is.

drgrice1:
@taniwallach:  As to running `npm install` inside of a docker container, make sure that you add the `--unsafe-perm` flag.  Then it will run the prepare script.  Note that when the image is built `npm install --unsafe-perm` is run, and the prepare script is executed, so the static assets are generated at that time.  Also note that with later versions of npm the `--unsafe-perm` option has been removed, and will not be needed.

taniwallach:
> @taniwallach: You never answered my question about the `dir=ltr` setting on the date/time inputs. Is that needed? It seems that behavior is quite odd as it is.

I sent a response inside https://github.com/openwebwork/webwork2/pull/1615#issuecomment-1087205329 .
I'm not sure what is the best thing for a date+time field. Bidirectional behavior is often inconvenient in form fields. I have no objections to adding a `dir=ltr` setting here. In the Hebrew locale - no Hebrew is actually being used and `dir=ltr` would probably behave better. It might not be the right thing for Arabic or some other RTL languages, but I think end users will get used to whatever there is.

drgrice1:
@taniwallach:  Sorry, I missed that.  Thanks.  I will add the `dir=ltr` setting then.

drgrice1:
You should note that this only supports the languages that we have translations for.  So this does not support Arabic at this point.  Each language that we support will now need a format added to datepicker.js and problemsetlist.js.  All of our current languages already have formats in those files.  The formats should be chosen to be consistent with the formats that perl's DateTime::Locale pacakge uses.

drgrice1:
@taniwallach:  So have you had a chance to test this again since I fixed everything?

taniwallach (approved):
In terms of the functionality of the new code to modify dates - everything I tested works properly now. I tested in Hebrew for the various places where a set deadline can be edited for all users or a specific user.

The change to use `dir="ltr"` certainly makes the date+time input boxes behave more conveniently for Hebrew courses, but the formatting is not fully consistent. I think that for the sake of consistency the places in the instructor tools put the date+time as a fixed string (ex. the table in the list of sets in the set editor) should also have `dir="ltr"` applied to them.  This is already the case in the table of sets assigned to a user ("UserDetail-date-table") for the class-wide fixed date.

If you do make that change - would you also try to force the "Set name" fields to also use `dir="ltr"` as when there are spaces in the set name, and a RTL language is used the name gets garbled by the RTL direction.

drgrice1:
I see.  I missed adding the `dir="ltr"` on the dates in the ProblemSetList main page (the Hmwk Sets Editor main page).

Yeah, the underscore to space substitutions aren't needed anymore.  This pull request came before the one that changed the locale file names, and so that was needed.  I will remove that.

I can look into adding `dir="ltr"` to the set name inputs.  Where exactly is that needed?

drgrice1:
I will deal with the set names in a separate pull request.  That is unrelated to this one.  Although, I would like to talk to you in general about this.  Is there any possibility that you could meet via zoom sometime today?  I can meet anytime.  I was looking into this, and it may be that instead of adding `dir="ltr"` we need instead to use the css `unicode-bidi`.

pstaabp:
Looks like @robert-marik approved this.  @drgrice1 let's merge after the conflicts.

drgrice1:
Working on it now.

PR #1616, opened 2022-03-19
Improvements to student progress table.

somiaj:
+ Adds a table header and table body to the student progress table.
  + Wraps the table description in a div.
  + Add email as a sortable option when viewing gateway quizzes.
  + Add email address to students who haven't started any test versions.
  + Only display statement about the problems column if it is shown.
  + Set the bottom margin to zero on problems column pre element.
  + Add $r->maketext to more output text for translation.
  + Make student links open up the appropriate problem set instead of
    the homework set list page to match description. For gateway quizzes
    the version number links to the respective test version.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
@@ -307,14 +307,14 @@ sub displaySets {
 
 	};
 	my %display_sort_method_name = (
-		last_name => 'last name',
-		first_name => 'first name',
-		email_address => 'email address',
-		score => 'score',
-		index => 'success indicator',
-		section => 'section',
-		recitation => 'recitation',
-		user_id => 'login name',
+		last_name     => $r->maketext('last name'),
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This improves the appearance of the table on the student progress page.  I think it is a good start for the page in general.

It seems this page still needs a lot of work though.  There are some invalid html issues.

The `align` attribute used on the `th` tag on lines 744 and 869 is obsolete.  This is easy to fix, since it can simply be deleted and nothing will change.

A harder problem to fix is the `pre` tags that are used.   A `pre` tag inside of a `span` is invalid.  This page should be updated to use something like is done on the `Grades` page instead or on this page when viewing progress for a single student.

I think we will merge this for now though.  The rest can be added later.  If you want to work on those things and continue your work from this pull request you can do this in another pull request.  Sound good?

somiaj:
Sounds good, I'll take a look at the issues you mentioned and if I come up with something make a new pull request.

drgrice1:
Okay.  I will merge this then.

PR #1619, opened 2022-03-21
Update the javascript for SetMaker (the library browser).

drgrice1:
In addition to modernizing the setmaker.js file, this fixes several bugs.

One issue is one that I caused when I switched from using the `onclick` attributes and inline javascript to using `addEventListener` and external javascript.  I missed a case where the "More/Less like this" handling changed the `onclick` attribute.  That has now been converted to change the appropriate data attribute instead.

Another issue was that the "Target Set" select was directly using its values for labels.  In particular it was using the "Select a Set from this Course" value which is translated.  The javascript looks for that value, and if it is translated it fails.  So now proper labels and values are used.  Furthermore the labels for sets have the underscore translated to spaces.  In addition, all of the "All ..." options for the subjects, chapters, sections, etc., are now translated in labels.  Of course the values are not translated.

There was also an instance of a jquery-ui dialog in setmaker.js.  Of course that is a problem as jquery-ui is no longer loaded on that page.  That was replaced with a bootstrap modal.  Since the contents of that modal are webwork errors, the styles of those errors were tweaked a bit to make them appear well both on a regular page and in those dialogs.  Also, all native alerts were replaced with bootstrap toasts.

In problemsetdetail.js there is a minor issue fixed in that the executor of a promise should not be an async function.

Translations are also added for some of the dialogs on the library browser page (the ones that will typically appear in day-to-day use). These are passed to the javascript via data attributes.

Some obsolete and already commented out code in lib/WeBWorK/Utils/ListingsDB.pm was deleted.  This is to make it clearer as to why some code was also deleted in SetMaker.pm.

### DIFF
--- a/htdocs/js/apps/ProblemSetDetail/problemsetdetail.js
+++ b/htdocs/js/apps/ProblemSetDetail/problemsetdetail.js
@@ -296,124 +296,119 @@
 	// Send a request to the webwork webservice and render a problem.
 	const basicWebserviceURL = '/webwork2/html2xml';
 
-	const render = async (id) => {
-		return new Promise(async (resolve) => {
-			const renderArea = document.getElementById(`psr_render_area_${id}`);
+	const render = (id) => new Promise((resolve) => {
+		const renderArea = document.getElementById(`psr_render_area_${id}`);
 
-			const ro = {
-				userID: document.getElementById('hidden_user')?.value,
-				courseID: document.getElementById('hidden_course_id')?.value,
-				session_key: document.getElementById('hidden_key')?.value
-			};
+		const ro = {
+			userID: document.getElementById('hidden_user')?.value,
+			courseID: document.getElementById('hidden_course_id')?.value,
+			session_key: document.getElementById('hidden_key')?.value
+		};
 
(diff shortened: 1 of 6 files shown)


drgrice1:
I think you are referring to #1615 regarding the hyphen versus underscore naming.  It seems that Transifex also uses underscores for its locale names.  However, HTML and javascript both want a hyphen.  That may be able to be changed though.  The perl DateTime::Locale library accepts either.

There is also a slight issue with the `zh_hk` and `en_us` locales.  Those should be `zh_HK` (or `zh-HK`) and `en_US` (or `en-US`).  The perl DateTime::Locale library accepts any case.  Javascript, html, and Transifex don't like the lower case region subtag.

pstaabp:
The setmaker (LibraryBrowswer) looks good.  I saw all of the code changes from native alerts to alertToasts, but couldn't trigger them all. 

 I'm not sure if there is anything on the problem set details to notice other than code rewrite.

pstaabp (approved):
Looks good.

drdrew42 (approved):

PR #1620, opened 2022-03-21
Allow instructors to use spaces in set names (rework of #1412)

drgrice1:
This updates pull request #1412 to the current develop branch.

Furthermore, this adds two methods in lib/WeBWorK/Utils.pm for formating set names.  `format_set_name_internal` is used to reformat set names when input by an instructor into a text input, and replaces spaces with underscores.  `format_set_name_display` goes the other way, and replaces underscores with spaces anytime a set is displayed in the ui.  The methods are in lib/WeBWorK/Utils.pm instead of in lib/WeBWorK/ContentGenerator/Instructor.pm, because the methods are also used outside of the instructor tools (at least `format_set_name_display` is).  Note that the only place that this is not done is in the Library Browser.  That is because this is done in pull request #1619.  Although it does not use the `format_set_name_display` method as that method is not available in that pull request.  Once either of these pull requests are merged, I will update the other accordingly to do so.

@Alex-Jordan:  I hope you don't mind me continuing your work on this.

### DIFF
--- a/lib/WeBWorK/AchievementItems.pm
+++ b/lib/WeBWorK/AchievementItems.pm
@@ -85,30 +85,31 @@ sub UserItems {
 # The id, label_text, and values are required parameters.
 sub form_popup_menu_row {
 	my %params = (
-		id => "",
-	   	label_text => "",
-	   	label_attr => {},
-		values => [],
-		menu_attr => {},
+		id                  => '',
+		label_text          => '',
+		label_attr          => {},
+		values              => [],
+		labels              => {},
+		menu_attr           => {},
 		menu_container_attr => {},
-		add_container => 1,
-	   	@_
+		add_container       => 1,
+		@_
 	);
 
-	$params{label_attr}{for} = $params{id};
-	$params{label_attr}{class} = 'col-4 col-form-label' unless defined $params{label_attr}{class};
-	$params{menu_attr}{values} = $params{values};
-	$params{menu_attr}{id} = $params{id};
-	$params{menu_attr}{name} = $params{id};
-	$params{menu_attr}{class} = 'form-select' unless defined $params{menu_attr}{class};
-	$params{menu_container_attr}{class} = 'col-8' unless defined $params{menu_container_attr}{class};
+	$params{label_attr}{for}            = $params{id};
+	$params{label_attr}{class}          = 'col-4 col-form-label' unless defined $params{label_attr}{class};
+	$params{menu_attr}{values}          = $params{values};
+	$params{menu_attr}{labels}          = $params{labels};
+	$params{menu_attr}{id}              = $params{id};
+	$params{menu_attr}{name}            = $params{id};
+	$params{menu_attr}{class}           = 'form-select' unless defined $params{menu_attr}{class};
+	$params{menu_container_attr}{class} = 'col-8'       unless defined $params{menu_container_attr}{class};
 
 	return join('',
 		$params{add_container} ? CGI::start_div({ class => 'row mb-3' }) : '',
 		CGI::label($params{label_attr}, $params{label_text}),
 		CGI::div($params{menu_container_attr}, CGI::popup_menu($params{menu_attr})),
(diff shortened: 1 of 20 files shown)


drgrice1:
Note that set names are still saved to scoring files with underscores.  I am not sure if it would be a good idea to save those with spaces instead.  I can change that if the opinion is that would be better.

somiaj:
I have tested this out. It appears to allow spaces and underscores to be used interchangeably when creating new sets with the homework set editor. If I already had a set created with spaces, I cannot create a similar set with underscores instead (as expected). Overall this works as advertised and being able to create sets with spaces in set names without throwing an error seems useful.

When looking at various pages such as grades, statistics, student progress, underscores are displayed as spaces in set names. The only place I saw underscores used instead of spaces was the problem editor.

I tested some achievements that use the set_id and underscores are required here as expected since the set_id uses underscores and the achievements evaluators use the string `eq` comparison to determine if a problem is in the correct set. I think due to this it is probably best for consistency to only use underscores in the database and not spaces.

I also tested putting spaces in the url, and get the error, `invalid characters in 'setID' field: 'Set With Spaces'`. Although I don't think user should be using spaces in urls, just pointing out there is an error if they manually type in the url.

Overall this works well and I don't think the above needs to be fixed, just pointing out what I've tested and ran into.

drgrice1:
This does not make it so that sets can be created with actual spaces in their names.  That can not be done with webwork's current structure.  This makes it so that for all practical purposes it appears that sets are being created with spaces in their names.  You can also use an underscore, and it will appear as a space everywhere.  In the database there will be no spaces in set ids.

If a set id is used in an achievement, the author will need to know that set ids actually do not have spaces.

Of course url's will still need underscores.  Spaces are invalid in url's in any case. 

I fixed the instance of the set id being displayed with underscores in the PGProblemEditor.  Thanks for pointing that out.

pstaabp:
I just created a set called `HW 1` and then another one called `HW 1 ` which of course translates it to `HW_1_`, a valid name, however in the set list, they look identical.  I'm not too worried about this, since the instructor shouldn't do this, and could have done this with underscores before.

drdrew42:
Trim leading/trailing whitespace, but not underscores?

drgrice1:
Now leading and trailing spaces are trimmed.  Underscores are not.  So you can ask to create `HW 1 `, and it will be `HW_1` in the database.  If you create `HW_1_`, it will be `HW_1_`.

drgrice1:
Now leading and trailing spaces are trimmed.  Underscores are not.  So you can ask to create `HW 1 `, and it will be `HW_1` in the database.  If you create `HW_1_`, it will be `HW_1_`.

somiaj:
I tested the trailing/leading underscores, and this works as described. But if you have both sets `HW_1` and `HW_1_`, they appear indistinguishable in the set editor, set lists, and so forth since the underscores are not shown in the set name. Since leading/trailing underscores are being treated different than spaces, would it also be reasonable to display them in the visible set name?

Could these utils functions also be applied when creating a new course? That way admins can create courses with spaces, and they are turned into underscores automatically. On this note I also noticed that if you use underscores in a course name, the underscores are visible on the main courses page, `/webwork2/`, but they aren't visible elsewhere.

drgrice1:
Yes, that is what happens. If you intentionally pus an underscore at the end, it is assumed you intended it to be that way, and yes, it will be visibly indistinguishable from a set that does not have that at the end but otherwise has the same name.  This is how it is intended to work.

At this point, we are not going to mess with the course name creation.  Admins are assumed to know a little more about the workings of webwork.

I have also noticed that the underscores are displayed on the main courses page.  That should be easy to fix.

drgrice1:
Now course names will be displayed with spaces instead of underscores on the main home page and on the login page.

drgrice1:
Also for course names the admin page has clear instructions as to what characters are allowed.  This is not the case for the places where sets are created.

pstaabp:
This looks good now.  Unless anyone wants to do more testing, I'll merge in.

Issue #1075, opened 2020-02-11
Temporary files are accessible without a login


dlglin:
On Feb 10, 2020, at 3:53 PM, Glenn Rice <grice1@missouriwestern.edu> wrote:

Danny seems to hit on the real problem here.  The TTU server has directory browsing enabled.  By default apache2 does this, but for a secure server it should be disabled.  To disable directory browsing for apache2 edit /etc/apache2/apache2.conf and remove "Indexes" from the root directory settings.

The default settings look something like the following:

<Directory />
    Options Indexes FollowSymLinks
    AllowOverride None
    Require all denied
</Directory>

Perhaps this should be added to the installation documentation for webwork.  Of course this still doesn't solve the issue if someone the exact path to a file. 

Glenn

On 2/10/20 4:43 PM, Danny Glin wrote:
A couple of points:

It looks like there is a configuration issue with the TTU server which allows temporary directory contents to be viewed from the web.  For example if I visit https://webwork.math.ttu.edu/wwtmp/ I am able to navigate through any of the tmp directories.  When I try this on my server, I get a landing page telling me that I shouldn’t be there, and if I try to view a subdirectory I get a 403 Forbidden error.  This configuration difference is likely why google was able to index the tmp files on the TTU server.  Adding a robots.txt file as Mike suggests would be another layer of protection to try to prevent these files from showing up on google.  Even with these fixes, someone who knows the exact path to a hardcopy file would still be able to access it.  Having a cron job periodically delete these mitigates, but does not solve the issue.

I had the same initial reaction as Tani regarding temporary files containing PPI;  I took a quick look at the files served from the htdocs/tmp directory, and it looks like it is only image files and pdf hardcopies of homework sets.  I can’t think of a scenario where an image related to a webwork exercise would contain PPI.  If that is the case, then it is just the hardcopy pdfs that are at issue.  Tani’s solution is probably the appropriate one.  

Danny

On Feb 9, 2020, at 3:06 PM, Nathan Wallach <taniwallach@gmail.com> wrote:

1. "We" actually just leaked this persons name + email address + some of his personal situation (from the request) not only to the recipients of this list but via the public archive at:  http://webwork.maa.org/pipermail/webwork-devel/ . I edited the older mails below so this message does not repeat that. Can someone manually edit the email archive to clean up the prior postst?

[...]

PR #1621, opened 2022-03-22
When hardcopy files are generated, stop generating them in a publicly accessible location.

drgrice1:
This solves the problem of these hardcopy files being made publicly available that may contain secure information.  See issue #1075.

The way this works is that the files are generated in a temporary subdirectory of /opt/webwork/webwork2/tmp/courseId/hardcopy/userID. Of course since that is not a publicly accessible, this means that the files can also not be directly downloaded via a link.  Instead the files are served via the content generator method `reply_with_file`.

When hardcopy generation succeeds the file is sent immediately, and is deleted from the server.  Note that file may be a pdf or zip file depending on the hardcopy format chosen.

If hardcopy generation fails, then the temporary directory is not deleted (same as before), but the files are served by adding a url parameter to the hardcopy module path whose value is the file name.  The hardcopy module then serves these files via `reply_with_file`.

Note that users will only be able to download files that are in the subdirectory with their own userID.  Furthermore users that do not have the 'download_hardcopy_multiuser' permission will not be able to download any of these files unless the user id that they are signed in with is in the file name between periods.  For example if `tguy` is signed in, then the user will be able to download `courseId.tguy.setId.pdf`, but not the temporary files `work.aq8jglRJ/pdflatex.stderr`, `work.aq8jglRJ/hardcopy.tex`, etc. (the file must also be in the subdirectory named for the active user id).

Note that when acting as another user the files are still created in a subdirectory named for your user id, and you can still download them.  So you technically do not have access to files generated by your students via the ui.  Of course you can still act as a student and get essentially the same files.

Also note that what your browser does when these files are served will probably change depending on how you have your browser set.  If pdf files opened in the browser before, they may not now (although this can still be configured by your browser if it has such configuration).  Text files will probably not open in your browser and will need to be downloaded.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -60,21 +60,29 @@ our $PreserveTempFiles = 0 unless defined $PreserveTempFiles;
 
 our $HC_DEFAULT_FORMAT = "pdf"; # problems if this is not an allowed format for the user...
 our %HC_FORMATS = (
-	tex => { name => x("TeX Source"), subr => "generate_hardcopy_tex" },
-	pdf => { name => x("Adobe PDF"),  subr => "generate_hardcopy_pdf" },
+	tex => { name => x("TeX Source"), subr => "generate_hardcopy_tex", file_type => 'application/zip' },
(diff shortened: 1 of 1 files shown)


somiaj:
Tested this out and it works as described. I can generate and download hardcopies normally. I had a test case where the build had an error, but as described I can still download the .pdf, hardcopy.tex, etc via the provided links. When working as a student I can only download the .pdf in this test case.

Overall looks good and works as described.

PR #1622, opened 2022-03-22
Hebrew translations updates - from WW 2.16 in the Technion

taniwallach:
Updates to the Hebrew translation dictionary (WW 2.16 version) taken from the Technion's production server. 

The file has translations for many strings not translated before, and changes many poor choices of phrasing/terminology from the first large-scale translation effort for Hebrew.

The current file may still have some typos or issues, but in my opinion if far better and more complete than what was in the original `heb.po` file from WW 2.16.

I would appreciate this being merged before the script to update the translations files for WW 2.17 is run, so these translations are carried into the new version.

### DIFF
--- a/lib/WeBWorK/Localize/heb.po
+++ b/lib/WeBWorK/Localize/heb.po
@@ -23,7 +23,7 @@ msgstr ""
 
 #: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm:683
 msgid "# of active students"
-msgstr "מספר תלמידים פעילים"
+msgstr "מספר סטודנטים פעילים"
 
 #: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm:498
 msgid "#corr"
@@ -35,7 +35,7 @@ msgstr "#שגוי"
 
 #: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm:2032
 msgid "% Score:"
-msgstr ""
+msgstr "% ציון:"
 
 #: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm:667
 msgid "% correct"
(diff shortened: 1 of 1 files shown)


drgrice1:
You asked me for a review, but I think the only qualified reviewer that we have for this is you!  So that is why I just merged it.

drgrice1:
On a related note, do you have access to the Transifex webwork project?  Could you add me as either an administrator or a project maintainer to the organization?  I want to work on setting up integration with github for that.  I believe that the project maintainer role is sufficient for that.

mgage:
I’ve invited you to be project manager for webwork on Transifex.  Let me know if you need something else from me.  I think I can upgrade you to administrator once you’ve joined.

Take care,

Mike

On Mar 22, 2022, at 11:24 AM, Glenn Rice ***@***.******@***.***>> wrote:



On a related note, do you have access to the Transifex webwork project? Could you add me as either an administrator or a project maintainer to the organization? I want to work on setting up integration with github for that. I believe that the project maintainer role is sufficient for that.

taniwallach:
> You asked me for a review, but I think the only qualified reviewer that we have for this is you! So that is why I just merged it.

Thanks. I really just wanted to make sure it got merged before you ran the script...

About Transifex - the stuff there is almost certainly very old (the main list of phrases is almost certainly from before WW 2.14 and was not updated for a very long time unless someone tried to update things in the last few years), and I do not know to what extent there are translations in Transifex which did not get into GitHub. Some new translations may have been startered since, but using an old collection of phrases...

Before WW 2.14 was released I tried to get everything which seemed "reasonable" which was in Transifex into the files on GitHub.

I would tend to say that the best approach might be to download all the translation files from Transifex, dump them in some other location on GitHub (so that the data is available if someone wants to try to look at it), and then try to import the current files from GitHub into Transifex.

drgrice1:
@taniwallach:  Sounds good.

It looks like I will need admin access to set up github integrations.  Currently I get 403 forbidden when I try to access the organization settings page where it is set up.

PR #1623, opened 2022-03-24
More student progress table improvements.

somiaj:
Remove the pre elements and use a nested table to display the
  problems scores for each set as mentioned in #1616.

  This is done by removing the custom grade_set method and creating
  a new generate_problem_table method, which creates a two row table
  listing the problem status and number of incorrect attempts.

  Switches to using the WeBWorK::Utils::grade_set method to grade
  each set instead of the custom grade_set method.

  Removes the success indicator index which was no longer being
  shown on the student progress table.

  Removes the two instances of using obsolete align attribute
  in the table headers.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
@@ -28,7 +28,7 @@ use warnings;
 use WeBWorK::CGI;
 use WeBWorK::Debug;
 use WeBWorK::ContentGenerator::Grades;
-use WeBWorK::Utils qw(jitar_id_to_seq jitar_problem_adjusted_status wwRound);
+use WeBWorK::Utils qw(jitar_id_to_seq jitar_problem_adjusted_status wwRound grade_set);
 #use WeBWorK::Utils qw(readDirectory list2hash max sortByName);
 use WeBWorK::Utils::Grades qw/list_set_versions/;
 use WeBWorK::DB::Record::UserSet;  #FIXME -- this is only used in one spot.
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
```diff
@@ -975,225 +909,104 @@ sub displaySets {
...
-			   $twoString,
-			   $totalRight,
-			   $total,
-			   $num_of_attempts,
-			   $num_of_problems
-		);
+	return $problemTable;
 }
 #################################
 # Utility function NOT a method
```
drgrice1, line 1012: You should delete the `fourSpaceFill` sub below here since it is no longer used.

drgrice1 (commented):
Now that I actually see this in the context of the rest of the table and page it doesn't look as good as I thought it would.  It is possible that with some css it could be made to look a little better, but still not the best.  The issue is that the width styles that you added on lines 541 and 1004 don't work that well.  When there are enough problems to make the percent width small the scores (particularly a 100) can exceed the given width, and then the problem scores tables below do not align correctly with the header table above.

I think the only way that alignment will actually work out is to do it the same as on the grades page and on the student progress page for a single user.  It should actually be easier to do that here than on those pages.  Here you have the advantage of knowing the number of problems, and that will be the same for all students.

drgrice1:
Here is a screenshot to show you what I am talking about with the alignment issue.
misaligned

This would get considerably worse if there were more nonzero scores.

somiaj:
The class I was using as a test case only has like 8-12 problems per set, so I overlooked what happens with larger numbers. I think I've done a lot of the setup with removing some of that old code, so I will work on integrating the table vs using nested tables.

drgrice1 (approved):
This looks much better now.  It is also more consistent in appearance with the other pages that do this sort of thing.

somiaj:
I just caught a issue in computing the number of columns to span when 'no tests taken' is shown and the problems are not visible on a gateway quiz progress. Update fixed that.

pstaabp (commented):
This looks good.  Nice improvement.

drgrice1:
I will count @pstaabp's comment as an approval and merge this.

PR #1624, opened 2022-03-24
Add gateway quizzes to set list in ProblemSet

somiaj:
Now that gateway quizzes can be seen on the ProblemSet page, there is no longer a need to remove them from the Sets navigation menu on the ProblemSet page.


drgrice1 (approved):
This looks good.

PR #1625, opened 2022-03-25
add German language file

fabiangabel:
derived from es.po

### DIFF
--- a/lib/WeBWorK/Localize/de.po
+++ b/lib/WeBWorK/Localize/de.po
@@ -2117,11 +2117,11 @@ msgstr ""
 
 #: /opt/webwork/webwork2/lib/WeBWorK/Localize.pm:413
 msgid "E-Mail"
-msgstr "E-Mail versenden"
+msgstr "EMail versenden"
 
 #: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Feedback.pm:344
 msgid "E-mail Instructor"
-msgstr "E-Mail an Dozenten schicken."
+msgstr "Email an Dozenten"
 
 #: /opt/webwork/webwork2/lib/WeBWorK/Localize.pm:431
 msgid "E-mail addresses which can receive e-mail from a pg problem"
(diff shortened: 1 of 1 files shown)


drgrice1:
@fabiangabel:  It looks like you are using a de.po file created from the webwork.pot file that is currently in the develop branch.  I recommend that instead of doing that, you run the `update_localization_files` script with your current de.po file.  This will update the file for all of the strings currently in the code.  If you don't do this, then you will end up wasting a lot of effort now for strings that have changed, and you will also be missing many strings that have been added.  We need to update the pot and po files in the develop branch, and will probably do that this week, but in the mean time you can already use that.  Note that running the `update_localization_files` will also tell you about errors in your translation file.  There are a few that you will need to fix.  The file won't update with the script until you fix those errors.

fabiangabel:
@drgrice1 thank you for pointing out `update_localization_files` script, I just fixed the errors and will push the new version of the po file now. It seems like the "deprecated strings" appear in the last section of the .po-file?

fabiangabel:
@drgrice1 Thanks for adding me to the transifex team. I have also uploaded the current translation there. What is the recommended workflow? I would have assumed to work locally and keep updating the transifex entries via upload?

drgrice1:
@fabiangabel:  The plan is that all translations will be done in Transifex from now on.  You can either download the file from there, edit it, and upload it again.  You can also edit the strings directly via the Transifex web interface.  Once pull request #1632 is merged this will become the official way to do translations.  After that, as soon as any language is 100% translated, Transifex will automatically make a pull request here with the updated translation strings.  We can also manually trigger the generation of pull requests at lower percentages if needed.

fabiangabel:
@drgrice1 Alright, I think that most of the student interface (excluding the Achievements) is translated now (45% according to Transifex).

> Once pull request https://github.com/openwebwork/webwork2/pull/1632 is merged this will become the official way to do translations.

Should we close/merge this PR then? Reaching 100% of the translation seems to be out of reach at the moment.

drgrice1:
It would be best to leave this pull request open for now to keep this in mind.  Don't worry about reaching 100%.  As I mentioned, we can also manually trigger a pull request from Transifex.  The plan is to do that at release time, even if a translation is not 100% translated.

PR #1626, opened 2022-03-25
Fix an error that occurs when viewing a problem from the PGProblemEditor

drgrice1:
This occurs when editing an unattached problem.  In this case the set assignment_type is not defined.  This causes an error about the use of an uninitialized value used in string comparison on line 881 of Problem.pm.


pstaabp (approved):
Works great.

PR #1627, opened 2022-03-25
Remove obsolete pg modules and add Types::Serialiser

drgrice1:
Remove the FlashApplet, JavaApplet, and CanvasApplet pg modules that no longer exist.

Add the `Types::Serialiser` module.  This is needed for AskSage to prevent the "Can't locate package Types::Serialiser::BooleanBase for @JSON::PP::Boolean::ISA at [PG]/lib/WeBWorK/PG/IO.pm line 475" warnings.

Also remove some obsolete comments.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1344,7 +1344,7 @@ ${pg}{modules} = [
 	[qw(AlgParser AlgParserWithImplicitExpand Expr ExprWithImplicitExpand utf8)],
 	[qw(AnswerHash AnswerEvaluator)],
 	[qw(LaTeXImage)],
-	[qw(WWPlot)], # required by Circle (and others)
+	[qw(WWPlot)],    # required by Circle (and others)
 	[qw(Circle)],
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Looks straightforward.

taniwallach:
Regarding the canvas applet - there are certainly some leftovers from that old tool in the webwork2 codebase.

See: `htdocs/js/legacy/sketchgraphhtml5b/SketchGraph.pjs` and some related files in the same directory.


If this tool is no longer viable - some mention of it being removed should be in the release notes and if necessary put on the Wiki.

---

Note: It looks like there may be a dependency on jQuery 1.4.2, as that is loaded by `htdocs/js/legacy/sketchgraphhtml5b/SketchGraph.html`

drgrice1:
Yeah, I was leaving those files for @mgage. But they aren't working anymore.  I would like to delete them and eliminate the last of the legacy js.

drdrew42 (approved):

PR #1628, opened 2022-03-27
Beef up the grade_set method and fix some issues with gateway quizzes on the student progress page.

drgrice1:
Improve the grades_set method so that it does everything that is needed by lib/WeBWorK/ContentGenerator/Grades.pm and
lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm, and have both of those modules use it.  A new parameter for grade_set is added to the end of the existing parameters to flag if the additional information that those modules need is desired.  If that flag is not set the additional processing needed for that information is not done.

The grade_set method sorts problems when the additional processing flag is set, so this also fixes the order of problems for random order gateway quizzes on the student progress page.

Also add a bit of styling to the problem scores in those tables so that uniform width is obtained.  Also center the data.  It just looks better that way.

Also fix the gateway quiz version numbers on the student progress page. Gateway quiz versions are not necessarily consecutive anymore.


somiaj:
Tested this out. All looks good and didn't notice any issues.

pstaabp:
Just to be clear, no changes to grade calculations for this.  Seems like consolidation of the code into a single place. Correct?

pstaabp (approved):
Looks good then.

PR #1629, opened 2022-03-28
Fix numerous display and html validation issues on the instructor stats page.

drgrice1:
Generally clean Stats.pm up.  That file was a considerable mess.  Many variables were defined and never used, and generally a lot of code was unneeded.

Note that the borders on the svg image that are now appearing were in the code before, but were not appearing because of an error in the style attribute.  That was fixed, and the styles were also converted to proper svg attributes.

For gateway quizzes, the links to problems now work.  Note that there is no valid link to the actual problem in the gateway quiz.  So instead the Problem.pm editMode is used.  This at least gives the instructor a link that works and shows the problem which is really what is important here.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
@@ -1294,15 +1294,21 @@ sub view_handler {
 	# redirect to Problem.pm or GatewayQuiz.pm
 	if ($file_type eq 'problem' or $file_type eq 'source_path_for_problem_file') {
 		# we need to know if the set is a gateway set to determine the redirect
-		my $globalSet = $self->r->db->getGlobalSet( $setName );
+		my $globalSet = $r->db->getGlobalSet($setName);
 
 		my $problemPage;
-		if ( defined($globalSet) && $globalSet->assignment_type =~ /gateway/ ) {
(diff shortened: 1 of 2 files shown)


somiaj:
Tested this out and noticed the following:

+ The problem links in the graph open a new tab/window, while the problem links elsewhere in the two tables don't open a new tab/window.
+ The student stats page and student student-progress page are the same with different titles. Are two pages needed for this?
+ When opening a gateway quiz problem in an undefined set it always says problem 1. This is just how it works, but I was initially confused that I opened the wrong problem since the problem number didn't match. (Not an issue just something that initially confused me).

drgrice1:
Your first two comments are just the way that it is.  I didn't change those.  I can make the links all open in a new tab/window.  The student stats and student progress pages are the same, and I don't plan to change that.

I will see if I can make the problem number be consistent with the problem number in the gateway quiz.  I am not sure that can be done with the current parameters though.

drgrice1:
@somiaj:  The problem number is now the same on the undefined set problem page as on the stats page.

drgrice1:
Also, now all problems open in a new tab/window.

pstaabp:
looks good to me.  One thing that just jumped out at me--although this is the way it was done before--is that the breadcrumbs end in Statistics / Statistics if you are at level of set or a user.  Would it make sense to end the breadcrumb with either set_id or user_id?  Again, not a big deal though.

PR #1630, opened 2022-03-28
Update the translation files for current strings in the code.

drgrice1:
This is simply the result of running the `update_localizatio_files` script on the current develop branch.  This will need to be done again once all pull requests are in as there are further changes to translation strings in them.


drgrice1:
Converted this to a draft for now.  This will most likely be preempted by #1632.

PR #1631, opened 2022-03-28
Delete the fix_copyright.sh script, and instead create an update-copyright script in bin/dev_scripts.

drgrice1:
This file should be run immediately before each release to update the copyright in all files, and then a pull request made with the results.  Much the same as what needs to be done with the update-localization-files script.  This could probably all be done in one pull request.

Also move the update-localization-files into the bin/dev_scripts location, and add a README.md to that location that states that the files in that directory are for development use only, and not for those just installing webwork.


pstaabp:
What's the format on the script for this.  Tried `bin/dev_scripts/update-copyright` and ` bin/dev_scripts/update-copyright lib/**/*.pm` and got the following error:

```
bin/dev_scripts/update-copyright: line 34: shopt: globstar: invalid shell option name
```

drgrice1:
The script is executed by running `./bin/dev_scripts/update-copyright`.  The script does not take any options.  Are you running the script on Ubuntu or on OSX?  Although that shouldn't matter.  This should work on both.  Don't you use the `globstar` option when you run `perltidy` on the webwork3 code?

pstaabp:
Looks like the bash version on OSX is 3.2 and globstar was added in 4.0. However, I just installed a newer version of bash (5.1) and still have the same error.  Grrrr.  

Also zsh is default on MacOS now and got the same error there.

drgrice1:
@pstaabp:  It should be the case that even if you are running the zsh, the script will still run with bash due to the shebang in the file.  Although generally speaking, webwork2 is still pretty heavily tied to linux.  So I think it should be generally okay if this does not run on OSX. In addition, this is a developer script only intended for development use.  So it doesn't really need to run on all operating systems.

drgrice1:
> > the script will still run with bash due to the shebang in the file
> 
> This is why even upgrading bash doesn't work. `/bin/bash` is the built-in bash 3.2. My upgraded bash is at `/usr/local/bin/bash`.

I switched the shebang to `#!/bin/env bash` so that it will work in this case.

> ```
> Can't do inplace edit on ./htdocs/tmp/staab_course/images/fd51cace-79b2-393d-b480-a95d050dde6e___63c8ad47-f4d2-3556-bc28-08a6003172b9.png: Cannot make temp name: Permission denied.
> ```
> 
> I can fixed the permission denied, but should we just be updating the copyright on `.pm` and `.pl` (and maybe `.t`) files?

Certainly not.  This script is intended to work for much more than perl files.  It also works for js, css, scss, html, and others.  Some (like some of the shell scripts in bin) do not have extensions either.  So it can not even be restricted by extension.  You will just have to clean out your tmp directories to use this script, or even better move your html tmp location somewhere else (highly recommended for a developer to do).

pstaabp (approved):
Had trouble with the #!/bin/env bash, but changing that line to my local version of bash works for me now.

drdrew42 (approved):

pstaabp:
Two approvals.  Will merge.

PR #1632, opened 2022-03-28
Add Transifex support and rename po files to html/js standards.

drgrice1:
Rename translation po files to use proper language codes that are consistent with naming conventions of html and javascript.  In other words with a hyphen instead of an underscore, lower case two letter language code, and upper case two letter region code.

@taniwallach:  Note the change of heb.po to he-IL.po as we discussed.  Also note that if a course has one of the old file names in its settings, then that will be detected and converted to the new naming scheme.  This is done in CourseEnvironment.pm when the course environment is loaded.  There may be changes needed when rendering via the html2xml endpoint and using the `language` parameter.  At this point, no conversion is done there.  Also, there may be an issue in `get_problem_lang_and_dir` of lib/WeBWorK/Utils/LanguageAndDirection.pm.  That method converts language codes to lower case.
    
Also add Transifex configuration files.  The file .tx/config is used by the local tx client.  The transifex.yml file is used directly by Transifex via the GitHub integrations.

This also updates the po files for current strings in the code.

This pull request must come first before Transifex/GitHub integration can be set up.  The `transifex.yml` file is needed in the develop branch for that.  There may also be further adjustments needed to that file.  Further work will be needed to get this all set up right.  So I am hoping to merge this quickly so that this can be set up.


taniwallach (approved):
I think this can be merged.

I confirm that the fixups for the language setting provided by `simple.conf` work properly, and that translations are loaded from the renamed files for regular course content. (Tested `heb`, `en-us`, `fr_CA`) with the course login page, and a bit more with Hebrew.

I did not look into how `html2xml` is behaving after this change. I would assume that the language parameter being send by such calls will need to be fixed to the new standard. I do not see that as something to hold back merging this commit.

As to whether the code in `lib/WeBWorK/Utils/LanguageAndDirection.pm` should be modified not to change to the case of the country code - I do not think it is critical, but am not certain. AFAIK browsers are not strict about the case there.

I don't know anything about the Transifex config files, but any fixes they need can be addressed later.

drgrice1:
If the `language` parameter` is changed, then the `html2xml` will still work as it always has.  However, if `heb` is used for that parameter, for example, then it probably won't work as expected.

We can iron out the details of the `lib/WeBWorK/Utils/LanguageAndDirection.pm` in the next few months before the final release.  It probably is not a big deal though, as you said.

I have tested the transifex config files now.  You probably saw the pull requests yesterday.  Those were manually triggered.  When a language is 100% translated (apparently you can't change that), then transifex will automatically put in a pull request to update that language.

PR #1633, opened 2022-03-29
Drop support for mod_perl 1 and versions of apache prior to 2.4

drgrice1:
This is overdue.

All of the old apache config files are removed, and the old wwapache2ctl scripts as well.

On an unrelated note the startDancer script is removed.


pstaabp:
> On an unrelated note the startDancer script is removed.

I don't think this is unrelated.  I'm not sure how that ended up in develop ever.  But wouldn't be surprised if it was traced to me, since that's was the backend for the old ww3 code. 

Searching for `MP2` and `mod_perl` resulted in some code in 
* `t/grab_course_environment.pl` Line 50

Not a big deal because it is in test code.

Overall, code runs great.

drgrice1:
Yeah, I saw that in `t/grab_course_environment.pl`.  I didn't mess with that file because it is broken anyway, and I don't have a valid way to test changes.  I will go ahead and remove the mod perl though.

pstaabp (approved):
This looks good now.

drdrew42 (approved):

PR #1640, opened 2022-03-30
More accessibility and html validation fixes as well as translation improvements.

drgrice1:
Add the role "figure" to the progress bar divs in lib/WeBWorK/ContentGenrator/Problem.pm and lib/WeBWorK/ContentGenrator/Achievements.pm.  It is invalid for a standard html element like a div to have an aria label without such a role.  I am not sure that this is the correct role for this though.

Fix a `div` that should have been a `start_div` in lib/WeBWorK/ContentGenerator/ProbemSet.pm.

Remove the obsolete summary attributes on the tables in lib/WeBWorK/ContentGenerator/ProbemSet.pm and lib/WeBWorK/ContentGenerator/ProbemSets.pm.  Those tables are structured in such a way that the summaries are not needed.  The summaries were basically just saying what is already in the captions and headers.  The summary in ProblemSets.pm was also mentioning the buttons below the table which was not correct to do to begin with.

The obsolete summary attributes in lib/WeBWorK/ContentGenrator/Instructor/UserList.pm and lib/WeBWorK/ContentGenrator/Instructor/ProblemSetList.pm have also been removed.  Some of their contents were added to the "Site Description" instead.

Make the "Login Name" column header contents the label for the select all checkbox in lib/WeBWorK/ContentGenrator/Instructor/UserList.pm, lib/WeBWorK/ContentGenrator/Instructor/ProblemSetList.pm, and lib/WeBWorK/ContentGenrator/Instructor/AchievementList.pm.  Also add an aria label to the select all checkbox itself which serves as the text for the column header (the WAVE tool does not accept this, but the Firefox accessibility tool and Chrome Lighthouse do, and it is better than nothing).

In lib/WeBWorK/ContentGenrator/Instructor/UserList.pm and lib/WeBWorK/ContentGenrator/Instructor/ProblemSetList.pm make the column headers the labels for the inputs in the column using the aria_labelledby attribute.  Of course this means that ids had to be added to the columnn headers.

Fix potentially duplicated ids in the achievement item forms inside the modal dialogs in lib/WeBWorK/ContentGenrator/Achievements.pm.  These will be duplicated if a student has more than one item available.

Remove the config strings from lib/WeBWorK/Localize.pm, and directly mark them in conf/defaults.config.  The bin/update_localization_files script is updated to pick up on these.  Having the strings duplicated in Localize.pm was not a good idea.

Also remove the translations for site descriptions (and removed table summaries) from the lexicon and put them directly into the one place that each is used.  This is better for translators if they want to look in source files for context.  The lexicon is meant to be used for translations that are used in multiple locations and need to be kept in sync.

The config values for the LTI tab that were in conf/authen_LTI.conf.dist have been moved to conf/LTIConfigValues.config.  Instead the @LTIConfigVariables has been added to conf/authen_LTI.conf.dist.  If any of the values in that array are uncommented, then the LTI tab will be shown with the uncommented values appearing.  The conf/LTIConfigValues.config file is included in conf/authen_LTI.conf after the definition of @LTIConfigVariables so that it can act on that setting.

There is still a lot of work to do to get the instructor pages up to speed for both accessibility and html validation.  In particular the achievement pages are total disaster areas for both of those things.
Edit:  Well, maybe not disaster areas.  It seems it is just a lot of work on input labels that is needed.  Html validation is pretty good.
Edit:  After additional work, the achievement editor pages are looking pretty good now.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -709,10 +709,13 @@ sub FieldHTML {
 				)
 			);
 		} else {
+			my $value = $r->param("$recordType.$recordID.$field") || ($forUsers ? $userValue : $globalValue);
+			$value = format_set_name_display($value =~ s/\s*,\s*/,/gr) if $field eq 'restricted_release';
+
 			$inputType = CGI::textfield({
 				name          => "$recordType.$recordID.$field",
 				id            => "$recordType.$recordID.${field}_id",
-				value         => $r->param("$recordType.$recordID.$field") || ($forUsers ? $userValue : $globalValue),
+				value         => $value,
 				data_override => "$recordType.$recordID.$field.override_id",
 				class         => 'form-control form-control-sm',
 				$forUsers && $check ? (aria_labelledby => "$recordType.$recordID.$field.label") : (),
@@ -753,6 +756,7 @@ sub FieldHTML {
 		&& defined($properties{labels}->{$globalValue})
 		? $r->maketext($properties{labels}->{$globalValue})
 		: $globalValue;
+	$gDisplVal = format_set_name_display($gDisplVal) if $field eq 'restricted_release';
 
 	my @return;
 


somiaj:
Any particular things you would like tested. I'm running through the various instructor pages using site-improve. Here are some things that I notice with accessibility issues flagged by site-improve:

+ On the Homework Set editor page I get the following:
    + 'Button without a text alternative' for the Collapse/Expand problem details up and down arrows on each homework problem.
    + 'Headings are not structured' for the 'Set Description' which jumps to an h4.
    + 'Form field is not labeled' for a lot of the form drop down menus and so fort in the homework/gateway quiz settings.
+ On the Problem page I get 'Form field is not labeled' for the drop down menus for selecting problem tags (I also get this for any answer boxes in the problem but this is a PG issue).
+ On the stats and student progress main page I get 'Headings are not structured' for the h3 headings view by set and view by student (these should probably be h2 headings).
+ On the email page I get a 'Container Element is empty'. There is a drop down menu to 'Open' a message file that is empty. I have never created any of these files so that is why. Maybe not show this drop down menu if there are no elements.
+ On the file manager page I get 'Form field is not labeled' for the drop down menu for the directory list, and the main list of files.

Anyways things look good, and outside of what site-improve flagged above I didn't notice anything in my first round of tests.

drgrice1:
What drop down menus are you talking about on the problem page?  If you are talking about the tag widget, you can forget about that.  That is only there for admin users, and no one cares about that thing.  Obviously you should be ignoring anything inside the problem container in regards to this pull request.

On the email page most users will see the `welcome.msg` file that is in the model course, and I am not particularly concerned about that being empty.

I know of the form fields on the file manager page.  I couldn't think of a good way to deal with that.

In general there is still some more to be done, but this takes quite a bit further.  On the homework set editor page I am seeing a lot more issues than you listed.  That is one page that is going to take a lot of work.

somiaj:
Yes, the tag widget on the problem page. I didn't think it was used that often, just stating what sort of things I found in my first pass at looking at the various pages. Overall it looks good and this is a good improvement.

drgrice1:
I switched the headings on the student progress page to `h2`s.  Note that this is done on the stats page in #1629.

drgrice1:
I added fixes for most html validation and accessibility issues on the ProblemSetDetail.pm page now.

drgrice1:
@pstaapb:  The h2 on the stats page is fixed in #1629, and I did mention the form elements on the file manager page.

pstaabp:
Great.  let's merge this then.

PR #1642, opened 2022-04-01
Remove some more redundant database calls in ProblemSet.pm and add sorting in the database.

drgrice1:
I have had this branch sitting here for a while, somewhat forgotten about.  This was not included in #1585 because I didn't want to cause merge conflicts with #1569 (both now merged).

The file is also cleaned up a bit and the gateway versions listing and homework problem listing sections separated in a more logical and readable way.  Those sections were perltidied.


somiaj:
Tested this out. Quizzes and homework sets both work in ProblemSet page. Outside of possible missing translations mentioned above, I didn't notice any other issues.

PR #1644, opened 2022-04-01
Fix a couple of missed instances where set names are still displayed with underscores.

drgrice1:
Also, one of those instances was not translated.


pstaabp:
I'm going to go ahead and just merge.

drgrice1:
Yeah, trivial one.

Issue #1480, opened 2021-08-19
ProblemSetList inconsistent use of "visible sets" (2 different meaning)


taniwallach:
In the file `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm` in the pull-down menu options the option "visible sets" seems to have 2 different meanings:

  - In the "filter" menu - the meaning is that the set has been "published" (made visible) to the students.
  - In the "edit" and "export" menus - the meaning is that the set is currently displayed in the table.

Filtering to a small group of sets, and using these options seems to be the best way to confirm what they actually mean/do.

This inconsistent usage is both confusing for users and makes proper translation challenging.

Maybe the options could better be called
  - Filter: "published sets" or "sets visible to students"
  - Edit/Export: "listed sets" or "sets displayed in the table" or something similar which makes it clear that it refers to what appears on screen and not to the "published / not-published" status of sets.

I suspect that we do not want to rename the variable `visibleSetIDs` which seems to control what is displayed and appears in many places.

@drgrice1 @mgage @pstaabp - Comments / suggestions?

taniwallach:
Being fixed by https://github.com/openwebwork/webwork2/pull/1646 which is essentially ready to merge.

PR #1646, opened 2022-04-05
Change the wording in the Filter, Edit, and Export selects as suggested in issue #1480.

drgrice1:
See the referenced issue #1480.

This now also adds a new option to the "Publish" form that if selected allows the user to publish all sets listed on the page.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
@@ -43,7 +43,7 @@ Sort sets by:
 	- visibility to students
 
 Switch from view mode to edit mode:
-	- showing visible sets
+	- showing listed sets
 	- showing selected sets
 Switch from edit mode to view and save changes
 Switch from edit mode to view and abandon changes
@@ -652,8 +652,8 @@ sub filter_form {
 						all       => $r->maketext('all sets'),
 						none      => $r->maketext('no sets'),
 						selected  => $r->maketext('selected sets'),
-						visible   => $r->maketext('visible sets'),
-						unvisible => $r->maketext('hidden sets'),
+						visible   => $r->maketext('sets visible to students'),
+						unvisible => $r->maketext('sets hidden from students'),
 						match_ids => $r->maketext('enter matching set IDs below'),
 					}
 				})
@@ -713,10 +713,10 @@ sub filter_handler {
 		my @setIDs     = grep {/$regexTerms/i} @{ $self->{allSetIDs} };
 		$self->{visibleSetIDs} = \@setIDs;
 	} elsif ($scope eq "visible") {
-		$result = $r->maketext("showing visible sets");
+		$result = $r->maketext("showing sets that are visible to students");
 		$self->{visibleSetIDs} = [ $db->listGlobalSetsWhere({ visible => 1 }) ];
 	} elsif ($scope eq "unvisible") {
-		$result = $r->maketext("showing hidden sets");
+		$result = $r->maketext("showing sets that are hidden from students");
 		$self->{visibleSetIDs} = [ $db->listGlobalSetsWhere({ visible => 0 }) ];
 	}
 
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
```diff
@@ -927,13 +927,6 @@ sub publish_handler {
 			? CGI::div({ class => 'alert alert-success p-1 mb-0' },
 				$r->maketext("All sets made visible for all students"))
 			: CGI::div({ class => 'alert alert-success p-1 mb-0' }, $r->maketext("All sets hidden from all students"));
-	} elsif ($scope eq "visible") {
```
taniwallach, line 930: I recommend leaving this block of code, but fixing the messages. @drgrice1 dropped it here as it cannot be reached at present.

Instead I recommend adding `visible` into the list in line 871 (the `values` line near the start of `publish_form`) which determines what options are provided by the select element for the "publish tab". That will allow using the filter feature to select sets which contain a given sub-string (the `match_ids` filter) and then to publish/unpublish the "listed" (matching) sets. 

The same end result can be attained by clicking the "select all" checkbox after using the `match_ids` filter is used and before changing to the "Publish tab" and then using the "selected" option from the pull-down menu of the "Publish" tab.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
```diff
@@ -927,13 +927,6 @@ sub publish_handler {
...
-		@setIDs = @{ $self->{visibleSetIDs} };
-		$result = $value
-			? CGI::div({ class => 'alert alert-success p-1 mb-0' },
-				$r->maketext("All visible sets made visible for all students"))
```
taniwallach, line 934: $r->maketext("All listed sets were made visible for all the students"))

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
```diff
@@ -927,13 +927,6 @@ sub publish_handler {
...
-			: CGI::div({ class => 'alert alert-success p-1 mb-0' },
-				$r->maketext("All visible hidden from all students"));
```
taniwallach, line 936: `$r->maketext("All listed sets were hidden from all the students"));`

taniwallach (changes requested):
All the "real" changes are to "phrasing" of menu items (and will influence the strings to be translated). That is certainly a change for the better and addresses https://github.com/openwebwork/webwork2/issues/1480.

I recommend letting `publish_handler` process the "listed sets", which can be useful after a `match_ids` filter.

taniwallach:
Note: When testing this I discovered that there is currently a problem with the "filter" for "hidden" or "visible" sets.
When such a filter is used, the code which generates the hidden `visible_sets` input fields is broken, and creates things like
```
<input type="hidden" name="visible_sets" value="ARRAY(0x561d1425d2b8)">
```
where the values should really be the names of the sets which should remain "listed" in the tables.

I showed this to @drgrice1 and he will work on fixing that issue, probably introduced in https://github.com/openwebwork/webwork2/pull/1585 . Until that is fixed, testing of making changes to "listed sets" after filtering by "hidden" or "visible" sets will fail.

drgrice1:
> Note: When testing this I discovered that there is currently a problem with the "filter" for "hidden" or "visible" sets. When such a filter is used, the code which generates the hidden `visible_sets` input fields is broken, and creates things like
> 
> ```
> <input type="hidden" name="visible_sets" value="ARRAY(0x561d1425d2b8)">
> ```
> 
> where the values should really be the names of the sets which should remain "listed" in the tables.
> 
> I showed this to @drgrice1 and he will work on fixing that issue, probably introduced in #1585 . Until that is fixed, testing of making changes to "listed sets" after filtering by "hidden" or "visible" sets will fail.

This is fixed in #1650.

drgrice1:
Note, that this pull request currently only changes wording and does not change any functionality.  The deleted block of code was deleted because the "visible" value did not exist, making that code unreachable as @taniwallach said.  If I merge @taniwallach's pull request to this, then there will be a functionality change.  That is adding a new option to actually export visible (visible as in visible to students) sets.

drgrice1:
@taniwallach:  I merged your pull request.

taniwallach (approved):
I think this is ready to merge.
Next person who has a look should probably just merge it unless they see something to fix.

PR #1647, opened 2022-04-06
Login status show name instead of user id

drgrice1:
Make the login status next to the login button show the user's first and last name, instead of the user id.

This is also done for the effective user when an instructor is acting as a student, except in this case the user id is shown in parentheses after the full name.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -1029,8 +1029,7 @@ sub loginstatus {
 		my $signOutIcon = CGI::i({ class=> "icon fas fa-sign-out-alt", aria_hidden => "true", data_alt => "signout" }, "");
 
 		my $user = $r->db->getUser($userID);
-		my $prettyUserName =
-			($user->first_name || $user->last_name) ? join(' ', $user->first_name, $user->last_name) : $user->user_id;
+		my $prettyUserName = $user->full_name || $user->user_id;
 
 		if ($eUserID eq $userID) {
 			print $r->maketext("Logged in as [_1].", HTML::Entities::encode_entities($prettyUserName))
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/ContentGenerator.pm
```diff
@@ -1028,16 +1028,26 @@ sub loginstatus {
...
-			print $r->maketext("Logged in as [_1].", HTML::Entities::encode_entities($userID))
+			print $r->maketext("Logged in as [_1].", HTML::Entities::encode_entities($prettyUserName))
 				. CGI::a({ href => $logoutURL, class => "btn btn-light btn-sm ms-2" },
 				$r->maketext("Log Out") . " " . $signOutIcon);
 		} else {
-			print $r->maketext("Logged in as [_1].", HTML::Entities::encode_entities($userID))
+			my $eUser = $r->db->getUser($eUserID);
+			my $prettyEUserName =
+				($eUser->first_name || $eUser->last_name)
+				? join(' ', $eUser->first_name, $eUser->last_name)
```
taniwallach, line 1043: I would add the `$eUser->user_id` into the `join` to avoid potential uncertainly if multiple users had the same name infomation.

taniwallach (changes requested):
Student names are not necessarily unique.

As a result, I think that `$prettyEUserName` for the "Acting as" setting should include the `$eUser->user_id` so it provides something which cannot be ambiguous.

lib/WeBWorK/ContentGenerator.pm
drgrice1: Good point.  I will add the user id in parentheses for the effective user.

drgrice1:
@taniwallach:  The user_id is now shown in parentheses after the student name for the effective user.

taniwallach (approved):

pstaabp:
Looks good.

PR #1648, opened 2022-04-06
Force the set name entry inputs to be left-to-right.

drgrice1:
This makes set name inputs behave better for right-to-left languages.

@taniwallach:  Is this all that is needed?  Or did you want all places that set names are displayed to also be forced to left-to-right?

### DIFF
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -121,6 +121,11 @@
 		(el) => {if (el.dataset.bsTitle) new bootstrap.Tooltip(el, {trigger: 'hover', fallbackPlacements: []});}
 	);
 
+	// Hardcopy tooltips shown on the Problem Sets page.
+	document.querySelectorAll('.hardcopy-tooltip').forEach(
+		(el) => new bootstrap.Tooltip(el, { trigger: 'hover', fallbackPlacements: [], html: true })
+	);
+
 	// PG Problem Editor
 	document.querySelectorAll('.reference-link').forEach((el) => new bootstrap.Tooltip(el));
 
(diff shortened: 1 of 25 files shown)


drgrice1:
@taniwallach:  I have added a start to displaying set names and file names ltr.  This is just a scratch on the surface though.  There is a lot more to do.

drgrice1:
@taniwallach:  Could you check over what I have done so far to see that I am on the right track with this.

drgrice1:
@taniwallach:  I think this is ready to go now.  I think I got all of the set names 'ltr'ed.  File names will take some more work probably, but this is a good start.  Please check it as soon as you can.

drgrice1:
Note that the set names that appear in the "siblings" part of the site nav are "ltr", but not aligned left.  That can be done, but will take some effort with styling to make it look good.  I will let you decide if that effort is needed, or if it will work this way.

drgrice1:
@pstaabp:  I requested your review for testing English mainly (although you can do what you can with testing Hebrew - probably as good as I can do).  Basically, everything should appear the same for English with or without this pull request.

pstaabp (approved):
As far as I can tell, this looks good.  I did check with English and set names look right.  I also switched to Hebrew and the set names all look LTR as well.

taniwallach:
@drgrice1 I'm sorry I did not get to this sooner. I took a look at the result of these changes in the WeBWorK-2.17 branch, and it certainly helps a lot with the handling of set names in a Hebrew course. Thanks!

pstaabp:
@taniwallach No worries on not getting to it, but (and I know you will) let us know if there are any changes needed.

Issue #1442, opened 2021-07-14
Remove checker for sufficiently new directory_tree.json


pstaabp:
The CourseAdmin.pm file checks that if the `directory_tree.json` file is sufficiently new.  (line 3441)  That file is no longer used in the current code and needs to be removed.

taniwallach:
I just ran into https://webwork.maa.org/moodle/mod/forum/discuss.php?d=5599 which lead me to find this issue.

It looks like `tagging-taxonomy.json` does get updated by `OPL-update` so maybe checking the time of that file could be a suitable test instead of checking `directory_tree.json`.

pstaabp:
This is no longer valid.  I can't find where it was removed.

drgrice1:
@pstaabp:  I think this is actually still there.  It is now on line 3984.

drgrice1:
This seems to be your problem:  https://www.freecodecamp.org/news/a-quick-tale-about-feff-the-invisible-character-cd25cd4630e7/

I opened the file in binary mode with `vim -b` and I see this character at the beginning of your file.  I am not sure what to do about it though.

drgrice1:
Also, if I delete that `<feff>` character that shows up, webwork can successfully import the students in the file.

drgrice1:
I am closing this again.  Now I think it is resolved.

PR #1649, opened 2022-04-06
Remove the upgrade check for the library-directory-tree.json file.

drgrice1:
This is not used by webwork, and so there is no reason to give warnings if it is missing or not up to date.  See issue #1442.


drdrew42 (approved):

pstaabp (approved):
Agreed.  Looks good.

PR #1650, opened 2022-04-06
Fix the filter on the ProblemSetList.pm for showing visible or hidden sets.

drgrice1:
This was a regression introduced in #1585.  I missed a place where the result of a list...Where method was used and needed to be mapped to the actual field value.  I double checked, and this is the only place this was forgotten.

Thanks @taniwallach for pointing this out.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
@@ -713,10 +713,10 @@ sub filter_handler {
 		$self->{visibleSetIDs} = \@setIDs;
 	} elsif ($scope eq "visible") {
 		$result = $r->maketext("showing sets that are visible to students");
-		$self->{visibleSetIDs} = [ $db->listGlobalSetsWhere({ visible => 1 }) ];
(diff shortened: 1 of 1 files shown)


taniwallach (approved):
Fixes the bug.

Testing method:
1. Filter by "hidden sets".
2. Check in the developer tools what is shown for the values of the hidden input elements with `name="visible_sets"`.
3. Check that Export for "visible sets" will export those sets.
4. Repeat 2 and 3 after filtering by "visible sets".

pstaabp:
Trying to see the bug, before applying the fix.  Is the problem if there is the existence of a hidden set that exporting visible sets fails?  I think that's what I'm seeing, although no errors. 

If so, this does seem to fix this.

drgrice1:
This occurs for either the "visible sets" or "hidden sets" filter.  On the Hmwk Sets Editor page select either of those filters and click "Take Action".  Then click on Export and then click on Take Action.  After doing so, no sets will be listed even if you have visible or hidden sets without this pull request.  With this pull request of course visible or hidden sets will be listed if you have any.

drgrice1:
Sorry, that wasn't quite correct.  When you select the Export tab, you need to make sure that "listed sets" is selected before you Take Action (I think it is by default).

pstaabp:
I see the previous bug and this looks fixed.

PR #1653, opened 2022-04-07
Load OPL set definitions from files instead of searching the OPL directory.

drgrice1:
This is done in the getDefList method of lib/WeBWorK/ContentGenerator/Instructor.pm.  The only time that method is called is when the homework sets editor is loaded.  Note that the course's templates directory is still searched, but the Library and Contrib subdirectories are pruned and so not look into.

A new script `bin/generate-OPL-set-def-lists.pl` is created that generates the needed data files by searching the OPL directory.  Note that this script is executed by `bin/OPL-update`.  For now, this script is executed in docker-entrypoint.sh when the docker container is started.  However, these files should be added to the OPL release built by webwork2/bin/OPL_releases/release.sh, and dealt with in the same way as the other files in webwork2/htdocs/DATA.  Of course, that whole process needs to be updated and removed from the personal fork it is currently in.  The release there is now more than two years old.

The defaults.config option $option{useOPLdefFiles} still is checked to determine if OPL set defintion files will be loaded and listed, but note that the $options{setDefSearchDepth} is ignored for this.  The new defaults.config option $option{useContribDefFiles} determines if set definition files from the Contrib directory will be loaded and listed.  This also ignores depth.  So the search depth option only applies to local directories for the course.


bin/generate-OPL-set-def-lists.pl
```diff
@@ -0,0 +1,107 @@
...
+		@$file_list[ sort { $file_depths[$a] <=> $file_depths[$b] || $uc_file_names[$a] cmp $uc_file_names[$b] }
+		0 .. $#$file_list ];
+}
+
+# Sort the files first by depth then by path.
+depth_then_iname_sort(\@opl_set_defs);
+depth_then_iname_sort(\@contrib_set_defs);
+
+# Write the lists to the files in htdocs/DATA.
+if ($ce->{options}{useOPLdefFiles}) {
```
taniwallach, line 99: I do not think the creation of the JSON files should depend on these variables.
Some courses could override the default set in `defaults.config` so the files should always be available on the server.

taniwallach (commented):
It works for me, and should speed up the loading of the set editor.

I'm not sure the location where the file is being saved should be in `htdocs` as no remote use of the files are apparently being planned. The data is just loaded by code in `lib/WeBWorK/ContentGenerator/Instructor.pm`.

drgrice1:
The file is saved in `webwork2/htdocs/DATA` the same as the other library JSON files because that is convenient and keeps it all together.  It makes it easier to deal with in the docker-entrypoint.sh file.  I don't see any real reason to move it elsewhere even if it doesn't need to be there.

Yeah, I thought of the issue of a course overriding the default.  I will remove that check.

drdrew42 (approved):
Works great!

pstaabp (approved):
This sped up the loading of the HW sets editor for me by an order of magnitude.!!! Yeah.

PR #1654, opened 2022-04-07
When viewing student progress list students instead of sets in sliblings.

somiaj:
When viewing student progress on the student progress or stats page, use the siblings links to list the students in the course instead of sets, so when viewing a student progress one can view progress of other students without having to return to the main student progress page or stats page each time.

In addition, make the breadcrumbs on the student progress and stats page show the userid or the set name that is being viewed.


drgrice1 (approved):
Looks good and makes sense.  Showing sets in the siblings when viewing the progress/stats for a user isn't very useful as the sets are already all shown on the page.

drgrice1:
Also, the breadcrumb change is nice.  Good observation.

pstaabp (approved):
Also think this looks great.

Issue #1458, opened 2021-07-23
OPL-update doesn't respect site.conf for htdocs/DATA/*.json


ametts:
In site.conf, I've pointed my htdocs and courses directories to a shared filesystem in a multi-server configuration.  But if I set my htdocs directory in site.conf as follows...

```
# URL and path to htdocs directory.
$webwork_htdocs_url  = "/webwork2_files";
$webwork_htdocs_dir  = <someplace-that-has-nothing-to-do-with-$WEBWORK_ROOT> ;
```
Running OPL-update will write tagging-taxonomy.json into $WEBWORK_ROOT/htdocs/DATA regardless of the site.conf settings.  This leads to "Could not load the OPL Taxonomy from the server" errors in the application.

Line ~~530~~ 546 of OPL-update seems to decide for itself where the webwork htdocs directory is:
```
#### Save the official taxonomy in json format
my $webwork_htdocs = $ce->{webwork_dir}."/htdocs";
my $file = "$webwork_htdocs/DATA/tagging-taxonomy.json";
```

In fact, all of the output JSON files for OPL-update behave in this manner, as does (apparently) the WebWork UI.   If the JSON files exist in my shared filesystem referenced as htdocs in site.conf, but not in the htdocs subdirectory of WEBWORK_ROOT (which is /opt/webwork/webwork2/htdocs on individual servers), they won't be found when using the UI.  In order to avoid running the lengthy OPL-update process on each server, I've had to set up symlinks of htdocs/DATA directory from each server to the shared file system..

drgrice1:
Your statement that "all of the WeBWorK UI" behaves in the same manner is not true.  I just checked, and every instance where the UI uses the htdocs directory correctly utilizes the `$ce->{webworkDirs}{htdocs}` variable which is set from `$webwork_htdocs_dir` set in site.conf.  You are correct that the OPL-update script should also use that variable.

pstaabp:
Fixed in #1655

PR #1655, opened 2022-04-07
Fix incorrect directories in OPL-update

drgrice1:
Fix several of the scripts (most importantly the OPL-update script) to use the correct directories as set in site.conf and defaults.config, instead of assuming that those directories are subdirectories of the webwork root location.  This makes these locations consistent with the locations used by the UI.

This fixes issue #1458.

### DIFF
--- a/bin/OPL-update
+++ b/bin/OPL-update
@@ -550,7 +550,7 @@ if($canopenfile) {
 use JSON;
 
 #### Save the official taxonomy in json format
-my $webwork_htdocs = $ce->{webwork_dir}."/htdocs";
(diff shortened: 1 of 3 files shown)


pstaabp:
I didn't get a chance to run this, but the code looks good.

drdrew42:
Testing this PR, I'll be back in half an hour after OPL-update runs... 😂

drdrew42 (approved):
Finally! 👍

pstaabp:
Checking now.  Off to grab a  🍺.  Repost when it's done....

pstaabp (approved):
Looks good.

PR #1656, opened 2022-04-09
Add missing translations to ProblemSet.pm

somiaj:
The table headers for the gateway quizzes on the ProblemSet page were not translated. This fixes that.


drgrice1 (approved):
Good find.

PR #1657, opened 2022-04-11
Add a script that can be used to download the latest OPL metadata release.

drgrice1:
Also setup docker to use the releases in the openwebwork OPL repository instead of those in heiderich's personal fork.

This all depends on openwebwork/webwork-open-problem-library#924 that adds a workflow to create releases, and that workflow being executed for the first time.

### DIFF
--- a/bin/OPL_releases/Makefile
+++ b/bin/OPL_releases/Makefile
@@ -1,5 +1,5 @@
 RELEASE_FILENAME=webwork-open-problem-library-METADATA.tar.gz
-GITHUB_USERNAME=heiderich
+GITHUB_USERNAME=openwebwork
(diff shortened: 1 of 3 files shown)


pstaabp:
I just looked at openwebwork/webwork-open-problem-library#924 and seems like we should merge that in, tag a release and then test this PR.

drgrice1:
Sounds good.  @drdrew42:  Any thoughts?

drdrew42:
openwebwork/webwork-open-problem-library#924 is now merged and this is ready to go.

drdrew42 (approved):

pstaabp (approved):

PR #1658, opened 2022-04-13
Add button to randomize view seed in PGProblemEditor.

somiaj:
Adds a button to randomize the view seed for the PG Problem Editor. Although one can manually change the seed, this allows for a quicker way to randomize the seed to test out different random versions. This is a minor feature I find useful and seeing if others may find it useful as well.

Also removes two occurrences of an old `pg_editor_action_div` CSS class which is no longer being used.

### DIFF
--- a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
@@ -46,4 +46,8 @@
 			document.body.appendChild(busyIndicator);
 		}
 	});
+
+	document.getElementById('randomize_view_seed_id')?.addEventListener('click', () => {
+		document.getElementById('action_view_seed_id').value = Math.ceil(Math.random()*9999);
+	});
 })();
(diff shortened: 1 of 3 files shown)


pstaabp:
First, I think we should not make it the "primary" color.  That should be for the "Take Action!" button.  I'm also leaning to putting it on the same line as the seed (to the right of the input element).  

Other's thoughts?

somiaj:
@pstaabp There are your suggestions, that look better?

pstaabp (approved):
Looks good now.

drgrice1 (approved):
This looks good, but I would suggest one minor display tweak.  Move the `mb-2` class that is currently on the div started on line 1208 onto each of its children.  So add it to the class of the label on line 1210, the div that contains the text input on line 1214, and the div that contains the new button on line 1223, and remove it from the div that contains all of that on line 1208.

This won't change the appearance when the randomize seed button is displayed on one line for wider screens, but on narrow screens the button will wrap and be below the seed label and input.  In that case the margin will give better separation and look better.  Before there wasn't any wrapping even on narrow screens so the margin only on the containing div was enough.

pstaabp:
updates made.

Issue #1419, opened 2021-06-15
Upgrading courses "check" phase has potential to alter database.


Alex-Jordan:
At #1391, @dlglin  describes an issue with the code for upgrading courses. Essentially, code that is supposed to only check on database structure might alter it. The alteration is supposed to happen in a second phase.

> The code to delete the published column is inside the CheckTableFields routine, which is only intended to identify issues with DB columns, which are then fixed by the UpdateTableFields routine. This means that in the admin course, as soon as you click on "Upgrade Courses" link it will delete the extra column, rather than initially reporting what changes it will make, and then not making the changes until you click the "Upgrade Courses" confirmation button at the bottom.

Alex-Jordan:
Fixed by #1659.

PR #1659, opened 2022-04-14
Delete extra tables and fields when a course database is upgraded.

drgrice1:
If a course has tables or fields that are not present in the database schema, then a checkbox will be displayed below the table or field asking to delete the table or field when the course is upgraded.

In order to make it easy to test this, the obsolete `facebooker` column is deleted from the global_user_achievement table (and all related code removed).  That does not work.  I personally don't think that integrating with a social media platform is a good idea for WeBWorK in any case.

To test table deletion, you will probably need to manually create a table in the database.  You can do this with `create table courseID_bad_table (name int);` for example.

Note that the code previously added to delete the old "published" field has been removed.  If you have that column, then you can now upgrade the course, and check to delete that column.  This is a more correct way to deal with this.  I believe this fixes issue #1419.

There are also some html validation issues that were fixed, and all of the inline styles for color:red and color:green were change to use bootstraps text-danger and text-success classes in the admin tools.

### DIFF
--- a/htdocs/js/apps/SelectAll/selectall.js
+++ b/htdocs/js/apps/SelectAll/selectall.js
@@ -6,9 +6,19 @@
 		const checks = document.querySelectorAll(`input[name$=${selectAll.dataset.selectGroup}]`);
 
 		if (selectAll.type.toLowerCase() === 'checkbox') {
-			selectAll.addEventListener('click', () => checks.forEach(check => check.checked = selectAll.checked));
+			// Find additional select alls in the same group if any.
+			const pairedSelectAlls =
+				document.querySelectorAll(`.select-all[data-select-group="${selectAll.dataset.selectGroup}"]`);
+
+			selectAll.addEventListener('click', () => {
+				checks.forEach(check => check.checked = selectAll.checked);
+
+				// Also check/uncheck any select alls in the same group.
+				pairedSelectAlls.forEach((check) => check.checked = selectAll.checked);
+			});
+
 			checks.forEach(check => check.addEventListener('click',
-				() => { if (!check.checked) selectAll.checked = false }));
+				() => { if (!check.checked) selectAll.checked = false; }));
 
 			// If all checks in the group are checked when the page loads, then also check the select all check.
 			// This can happen in AchievementList.pm.
--- a/htdocs/js/apps/SelectAll/selectall.js
+++ b/htdocs/js/apps/SelectAll/selectall.js
@@ -1,12 +1,27 @@
 // Select all checkbox handling
-// Used in UserList.pm, AchievementList.pm, and ProblemSetList.pm.
+// Used in UserList.pm, AchievementList.pm, ProblemSetList.pm, and CourseAdmin.pm.
 
 (() => {
-	const selectAll = document.getElementById('select-all');
-	if (selectAll) {
-		const checks = document.querySelectorAll(`input[name=${selectAll.dataset.selectGroup}]`);
-		selectAll.addEventListener('click', () => checks.forEach(check => check.checked = selectAll.checked));
-		checks.forEach(check => check.addEventListener('click',
-			() => { if (!check.checked) selectAll.checked = false }));
+	for (const selectAll of document.querySelectorAll('.select-all')) {
+		const checks = document.querySelectorAll(`input[name$=${selectAll.dataset.selectGroup}]`);
+
+		if (selectAll.type.toLowerCase() === 'checkbox') {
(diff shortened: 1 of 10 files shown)


pstaabp:
Accessibility check for Admin tools via Lighthouse (chrome):
 - on Rename course: I'm seeing "Form elements do not have associated labels" on the bottom three input elements on the page. 
 - 

The database deletion works great.  BTW if you upgrade courses with PR #1588 then you will get the new field `set_id` in the `key` table that you can delete with this PR.  

I only had a few courses.  I wonder if it would be beneficial to have a "select all deletable keys" toggle checkbox for selecting/deselecting all if this needs to happen for a large number of courses.

drgrice1:
It is a little off the track of this pull request, but I made the checkbox labels corresponding to those inputs pull double duty as the labels for the inputs via the aria-labelledby attribute.

I had also thought about mentioning #1588 as you noticed.

I also thought someone would ask for a "select all" toggle.  I will see what I can do about that.

drgrice1:
I added "select all" check boxes for the missing tables and fields now.

drgrice1:
I moved the checkboxes to the end of the line, instead of being below the field or table.  This makes them more visible on the page.

pstaabp:
If others are testing the column deletion, here's how to add a column to an existing table (the COURSE_NAME_key table):
```
 alter table COURSE_NAME_set add column new_column varchar(255);
```

and this script will remove it.

pstaabp (approved):
Looks good and works well.

drdrew42:
Used this on a dev server with many courses -- I was selecting the extra column to delete for each individual course before I decided that was enough and scrolled all the way to the bottom to take action. It was only then that I saw the 'select all' option.

Perhaps the select all option should be positioned atop the list of courses and their tables?

drdrew42:
Well, a combination of individually selected columns to delete _and_ the 'select all' option did not result in all extra columns being dropped -- only those courses where I explicitly selected the columns for deletion were modified. Is this intentional?

drgrice1:
I am not sure exactly what you mean.  Can you clarify exactly what you did?

drdrew42:
* visit 'upgrade courses' tab
* all courses with possible upgrades are selected automatically when the page loads so just 'take action'
* now all courses are shown with expanded detail and 'facebooker' column has the unselected checkbox 'delete this field when upgrading'
* check this box for some courses
* scroll to the bottom, leaving some courses with the facebooker column unselected for deletion
* check the box right at the end: 'Select/unselect all fields missing in schema for deletion'
* take action

result: only the individually selected courses had the facebooker field deleted, so the final 'select all' did not seem to take effect.

drgrice1:
Odd.  I will look into it.  It should delete all.

drdrew42:
I've checked, and it's irrelevant whether or not individual course fields are selected for 'delete on upgrade' -- the select all at the end doesn't seem to work regardless.

drgrice1:
Hmm.  This did not happen for me.  I checked the 'delete this field' checks for two out of many courses.  When to the bottom, and checked 'Select/unselect all fields...' and take action, and the fields were deleted from all courses.

drgrice1:
Did you run `npm install` (or at least generate-assets.js) to test?

drdrew42:
`<picard_facepalm.gif>`

drdrew42 (approved):

drdrew42:
Though, I would still suggest moving the 'select all' to the top of the list? If there are very many courses, one has to know it's down there...

drgrice1:
Sounds good to me.  Perhaps the check boxes should be both at the top and bottom?

However it will be take some work to put the check boxes at the top.  The current structure of the code is such that all of the course database tables (and directories) are checked in a loop, and as that is done the reports are printed.  During that loop it determines if there are extra tables or fields for any of the courses.  Then if there are extra tables or fields the check boxes are displayed after the loop.  So to put the check boxes at the top, I will have to make it delay printing.

drdrew42:
The delay would be necessary in order to avoid displaying a 'select all' option if/when there are no available deletions?

drgrice1:
Exactly.

drgrice1:
I won't take that much work really.  I am on it.

drgrice1:
There.  That should do it.

drgrice1:
Crap.  Now that the upgrade confirmation page has the nice structure with borders around the course reports, the other pages with the database/directory reports are going to need it to or there will be inconsistency.

drdrew42:
I was just gonna say that the borders look nice

Issue #1660, opened 2022-04-14
Breadcrumb links on problem_detail page


somiaj:
I noticed that on the problem_detail page the breadcrumbs have a minor issue that the final problem number is a link, when it shouldn't be. I also noticed that when using show me another, the breadcrumbs aren't showing the name '/ Show Me Another' after the problem number in the breadcrumb link, which I think it should be.

pstaabp:
fixed in #1661

PR #1661, opened 2022-04-14
Breadcrumb link tweaks (issue #1660)

drgrice1:
In the breadcrumb make the problem number not a link on the problem page.  That is a link to the page the user is currently on, and so should not be a link.  This is also consistent with the breadcrumb for other pages.

Also add "Show Me Another" to the breadcrumb on the ShowMeAnother page.


somiaj:
Tested and can confirm the breadcrumbs now work as described.

pstaabp:
Agreed.  Looks good.

PR #1662, opened 2022-04-15
Fix checkboxes on the AchievementUserEditor.pm page.

drgrice1:
When I converted the CGI::checkbox to a CGI::label I forgot to convert the way the checked attribute is handled.  The conversion was necessary to remove the invalid empty label that CGI::checkbox adds.


drdrew42 (approved):

pstaabp (approved):
Fixed the error.  Nice catch.

Issue #1664, opened 2022-04-15
Show scores is set to override when editing a user's quiz by default.


somiaj:
When editing the user details for a gateway quiz, the checkbox to override the set's option to 'Show Scores on Finished Assignments?' is selected by default. I don't think this should be selected by default, and I find I am unclicking this checkbox with the quizzes I need to change the start/due times for an individual student.

drgrice1:
I disagree.  I always use that option as is, and so if the option were the other way, then I would always have to change it back.

drgrice1:
In any case, This is not technically an issue.  You just disagree with the default.  We can have a discussion about what the default should be though.

somiaj:
The reason I brought this up as I found it odd this one field is treated different than the others. For my use case I would prefer it was unchecked and consistent with how the other fields work, but if this default is useful to most, I can keep manually unchecking it myself.

drgrice1:
After looking at this closer I see that I was misinterpreting what you are talking about.  I thought you were talking about the default values that are set when a gateway quiz is created.  You are instead talking about the checkbox for determining if the user value overrides the global value.  You are correct, and this definitely is an issue.  There is a mistake in the logic for determining when the override should be checked.  I will put in a pull request soon to fix it.

pstaabp:
Fixed by #1665

PR #1665, opened 2022-04-16
Fix override check logic in ProblemSetDetail.pm

drgrice1:
Fix the logic for determining if a user value is set to override the global set value on the problem set detail page.

Note the only actual change to ProblemSetDetail.pm is on line 762.  The rest is general clean up I did while sorting out the issue.

See isue #1664 for details.

There is also a fix for an invalid value for the checked attribute in ProblemSetDetail.pm.


somiaj (approved):
The override checkbox is no longer selected by default. This fixes the issue for me.

drdrew42 (approved):
Confirmed fix.

pstaabp (approved):

PR #1666, opened 2022-04-16
Remove the inline javascript in AchievementItems.pm

drgrice1:
Also remove the jquery usage in that javascript.

The easiest way to test this is to reset achievements (assign all achievements and overwrite data), then manually set your users achievement_points to a large number (I used 20000) in the database with `update courseID_global_user_achievement set achievement_points=20000 where user_id='your_user_id';`, and then work a problem in a homework set (you don't have to get it right).  All of the achievements will then flash on the screen, and you probably won't have time to read them all.  Then all of the achievement items will be shown when you go to the achievements page.  On any of the items that have both sets and problems, you should see the correct number of problems shown in the problem select when you change sets in the set select.


drdrew42:
If you don't want to manually edit the database entries, this PR may also be tested by setting the achievement points earned per problem to >= 20000 in course config, and then answering one problem correctly.

Screen Shot 2022-04-19 at 2 17 26 PM

drdrew42 (approved):
Tested 👍

drgrice1:
@drdrew42:  Nice.  That is easier than manually editing the database.

pstaabp:
I'm not sure I saw all of them (seems like all of the level ones though).

drgrice1:
All of the items that are affected by this pull request are received with the levels.  So you should see all that matters.

pstaabp (approved):

Issue #1667, opened 2022-04-16
Email to/from headers showing encoded string


somiaj:
I was testing sending emails and the email headers I got looked like:

```
from: =?UTF-8?B?SmFpbW9zIFNrcmlsZXR6?= <name@domain.com>
to: =?UTF-8?B?SmFpbW9zIFNrcmlsZXR6?= <name@domain.com>
```

Looking into this it appears that Email::Stuffer uses `header_str_set` to set the header strings, which will encode strings that contain control or 8-bit characters.

https://metacpan.org/pod/Email::MIME#header_str_set

I believe this means that the `rfc822_mailbox` method in `lib/WeBWorK/DB/Record/User.pm` no longer needs to encode `$full_name` since Email::Stuffer will do the encoding if needed.

drgrice1:
I have a branch that already does this.  When I switched to using Email::Stuffer I planned on  doing this.  In addition to the minor issue on the email page that you observed there is a more serious issue.  It actually raises the spam score of emails to unnecessarily base64 encode an address.  An email sent from webwork comes in with something like this in it when it goes through the pipeline of my universities email server:
```
X-Barracuda-Spam-Report: Code version 3.2, rules version 3.2.3.97391
	Rule breakdown below
	 pts rule name              description
	---- ---------------------- --------------------------------------------------
	0.14 MISSING_MID            Missing Message-Id: header
	0.01 FROM_EXCESS_BASE64     From: base64 encoded unnecessarily
	1.05 FROM_EXCESS_BASE64_2   From: base64 encoded unnecessarily
```
The missing message id is something I haven't figured out what to do about yet, but the other two are caused by this issue.

somiaj:
@drgrice1 The message ID might be an issue of the SMTP server being used and not WeBWorK. I have tried to look into this and found a stackexchange post that talks about how the server should be adding message IDs to the emails it sends out. At my university I use a mail relay, and when I look at the headers of the emails sent by WeBWorK, they correctly have a Message-Id included in the email header from the SMTP relay I am using.

drgrice1:
That probably is the case.  So that is not WeBWorK's problem.

pstaabp:
This was fixed in #1668.

PR #1668, opened 2022-04-16
Remove the MIME-Header/Base64 encoding of the full name in the email to and from addresses

drgrice1:
It is incorrect to do this for all names regardless of if the name contains characters which neccessitate such encoding and raises the spam score of the email address in cases where the name does not contain such characters.
    
Email::Stuffer automatically encodes names that do have such characters, and does not do so if the name lacks such characters, so this encoding by webwork is no longer neccessary.


somiaj:
Shouldn't the include `use Encode qw(encode);` on line 27 of User.pm also be removed since this function is no longer being used?

drgrice1:
Yeah, good catch.

somiaj (approved):
I can confirm this fixes the issue with the encoding of the $full_name in #1667.

drdrew42 (approved):
I can feel the spam score improving already...

pstaabp (approved):

pstaabp:
Note: this fixes the issue in #1667

PR #1672, opened 2022-04-21
Fix typo in file_type of PDF in %HC_FORMATS

taniwallach:
The typo gets into the `Content-Type:` response header, so may interfere with browser handling of the file for some browsers.


drgrice1:
I will go ahead and merge.  This is an obvious mistake.

PR #1673, opened 2022-04-22
Remove the apache2 rewrite on the webwork2/htdocs location.

drgrice1:
That is not needed anymore with the js/css build system setup.


drdrew42 (approved):
💀

drgrice1:
It seems the rewrite is still needed for some things.

PR #1677, opened 2022-04-27
Update OPL statistics upload target

drdrew42:
The community will benefit from reviving the collection of OPL statistics.

This functionality has been dead for several years, and should be resurrected.

The sharing of OPL statistics is completely voluntary, and should pose no issue with rules concerning the sharing of student PII.

Statistics collected:
----------------------
| field | description |
|----|----|
| source_file | path to problem source (only paths =~ `/^Library\//`) |
| students_attempted | *number* of students who attempted the problem |
| average_attempts | average number of attempts per student |
| average_status | average completion of problem per student |
---------------------

There should be little concern over data sharing, as absolutely zero student PII will be shared. The data is entirely in aggregate as far as individual students are concerned. The data to be shared can be confirmed by checking your server's OPL_local_statistics table.

### DIFF
--- a/bin/upload-OPL-statistics.pl
+++ b/bin/upload-OPL-statistics.pl
@@ -177,7 +177,7 @@ print "Zipping files\n";
 `tar -czf $tar_file $output_file $desc_file`;
 
 print "Uploading file\n";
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This adds the new external program `sshpass`.  This should probably be added to the `check_modules.pl` script.  Note that this program is not installed by default (at least on Ubuntu), and so we will also need to add this to the list of things to install in the installation instructions and release note upgrade instructions.

drdrew42:
I must have installed sshpass at some point then. Since it's not installed by default, I figured out how to create an anonymous account for uploads.

PR #1678, opened 2022-04-28
Use useragent to get client ip address in logs.

somiaj:
When running a server behind a proxy, client_addr returns the
  ip address of the proxy. This switches to using useragent_addr
  which can return the ip address of the original client.
  useragent_addr is already being used to determine the restricted
  ip address, this now uses the useragent_addr ip address in logs
  and emails.


drdrew42:
@somiaj this should be re-targeted to the 2.17 branch so it gets into the upcoming release. 🙏

drdrew42 (approved):

dlglin:
When I tried this I used `$r->useragent_ip` rather than `$r->useragent_addr->ip_get`.  Is there a reason to use one rather than the other?

dlglin (approved):

drdrew42:
[Documentation](https://httpd.apache.org/docs/2.4/developer/new_api_2_4.html)

Search for `remote_ip`:
> When you require the IP address of the user agent, which might be connected directly to the server, or might optionally be separated from the server by a transparent load balancer or proxy, use request_rec->useragent_ip and request_rec->useragent_addr.

Looks like `$req->useragent_ip` is interchangeable with `$req->useragent_addr->ip_get`?

somiaj:
@drdrew42 Should I make a second pull request for the 2.17 branch so it gets into both places, or just retarget this one?

It appears there is a small difference between `$r->useragent_ip` and `$r->useragent_addr->ip_get`, that may have a small efficiency difference, but the end result is the same. I went with the second since I was replacing a call to `$r->connection->client_addr->ip_get` which could have also been `$r->connection->client_ip`. I found this from a mailing list:

> 'useragent_ip' belongs to request_rec, from which you get the information about 
> the HTTP request from the real user end point (the Web browser) on the 
> application layer. You can do: `$r->useragent_ip()`
>
> It seems that '$r->useragent_addr' is mapped to 'apr_sockaddr_t', which means 
> you can do: `$r->useragent_addr->ip_get()` and `$r->useragent_addr->port()`

One issue I thought about initially was that the useragent can be spoofed, so I almost considered making this a `site.conf` configuration option for those who use a proxy, and use `$r->connection->client_add->ip_get` for those who aren't. The reason I didn't go this route was the main place this would matter most is for restricted_ip assignments which was using the useragent_ip anyways, so didn't think it was considered that much of an issue for the logs.

drdrew42:
Just re-target this one, there have been several PRs that have merged into 2.17 and not develop (yet). Once the RC is 'finalized', we'll merge 2.17 back into develop.

PR #1681, opened 2022-05-04
Update docker-entrypoint.sh

mgage:
replaced master with main in docker-config/docker-entrypoint.sh

### DIFF
--- a/docker-config/docker-entrypoint.sh
+++ b/docker-config/docker-entrypoint.sh
@@ -57,7 +57,7 @@ fi
 if [ ! -d "$APP_ROOT/libraries/webwork-open-problem-library/OpenProblemLibrary" ]; then
   echo "Cloning the OPL - This takes time - please be patient."
   cd $APP_ROOT/libraries/
-  /usr/bin/git clone -v --progress --single-branch --branch master https://github.com/openwebwork/webwork-open-problem-library.git
+  /usr/bin/git clone -v --progress --single-branch --branch main https://github.com/openwebwork/webwork-open-problem-library.git
 
   # The next line forces the system to run OPL-update or load saved OPL tables below, as we just installed it
   touch "$APP_ROOT/libraries/Restore_or_build_OPL_tables"


drgrice1:
I am not sure that it is a good idea to put this in as a hot fix at this point with the release candidate already out, and the release only a few of months out.  This will be needed in the release though.

taniwallach (approved):
This probably qualifies for a hotfix since the OPL branch name has been changed.
We should probably post instructions somewhere about how to fix the branch name on existing OPL installations.

drgrice1:
So are you also going to hotfix WeBWorK 2.14 and 2.15?  Are we going to rebase the main branch so that can be done?

This is just highlighting the incorrectness of having the docker build as part of the webwork2 github repository code.  It should be in a separate repository entirely.

taniwallach:
> I am not sure that it is a good idea to put this in as a hot fix at this point with the release candidate already out, and the release only a few of months out. This will be needed in the release though.

Anyone installing the current main branch of WW and using Docker will run into problems with the OPL installation until the new release is finished. That is why I think this might justify a hotfix. If we do not make the hotfix - then I would recommend adding a "Gotcha" note about working around this for now in the Docker install instructions on the GitHub wiki.

drgrice1 (approved):
I will approve this as a hot fix for now to ease those using the main branch at this point.  However, this does need to be changed.  It has never made sense to me that you need to download the entire webwork2 repository, and really are using none of it for the docker build other than the docker files, and then the docker build downloads the entire webwork2 repository again where it actually uses the webwork2 code.  If the docker build were in a separate repository it would fix that redundancy as well as make it easier to deal with issues like this.

mgage:
I'm less expert than you guys, but I'd be in favor of having a separate docker build repository -- it's conceptually cleaner to have docker building something outside of itself.  On my installation I would have directories    webwork/webwork2, webwork/pg, webwork/docker, ww-docker-data (built by docker, or alternatively a directory courses since at the moment ww-docker-data doesn't hold anything except courses)

taniwallach:
@drgrice1 Thanks for accepting/merging this as a hot fix.

We certainly have limits to how much we can try to maintain. So long as the Docker files are in the webwork2 repository, I do not think we can reasonably maintain the Docker files for prior releases once a new release is made, but I do think that the current release should function "out of the box" until a new release is made.

I opened a discussion at https://github.com/openwebwork/webwork2/discussions/1683 about the idea of moving the Docker files to a separate repository. I can see certain benefits to that approach. It was considered in the past but not implemented, as there was some objection at the time from people with more Docker experience. I think we can reconsider the matter.

In particular, it would be nice to have functioning Docker files for the "prior" release and I have an idea of how that could work. They key seems to be to adding tags for updated "commit heads" of the releases to use as targets for the `git clone`, and for `docker-compose.yml` for the "prior release" to use the tag instead of a branch name.

Keeping the Docker files for the prior releases easily available and somewhat easy to maintain would be far easier if the files were maintained in branches of a different repository.

PR #1686, opened 2022-05-06
Update check_modules to add node and check its version as well as rem…

pstaabp:
Update check_modules to add node and check its version as well as remove apache1 modules


bin/check_modules.pl
```diff
@@ -196,20 +181,30 @@ =head1 DESCRIPTION
 	check_modules(@modulesList);
 }
 
+use Data::Dumper;
```
drgrice1, line 184: I think you forgot this!

drgrice1 (approved):
Looks good.

bin/check_modules.pl
pstaabp: removed it!

drgrice1: Thanks.

drgrice1:

PR #1687, opened 2022-05-10
Fix an issue with rendering a student problem in the Hmwk Sets Editor.

drgrice1:
The issue can be seen by going to the Hmwk Sets Editor, and then editing
a set for a student when an override source file is not set (which is
probably most of the time).  Click the "Render Problem" button to render
the student's problem.  Then click "Check Answers".  The problem will
fail to render, and errors are reported.  The issue was a mistake in the
javascript that checked the existence of the source file input instead
of checking to see if the value was non empty before falling back to the
default source file for the problem.


drdrew42 (approved):
tested + fixes the issue with 'check answers'

pstaabp (approved):

PR #1690, opened 2022-05-13
Add option to force lowercase usernames with LTI.

somiaj:
This option will make the username source lowercase when logging in via an LTI.


conf/authen_LTI.conf.dist
```diff
@@ -154,6 +154,11 @@ $strip_address_from_email = 0;
 #$preferred_source_of_username = "lis_person_sourcedid";
 #$fallback_source_of_username = "lis_person_contact_email_primary";
 
+# Depending on the username source, capitalization may vary between students,
+# such as when using an email address to get the username. Turing this option
```
drgrice1, line 158: I will make sure to "tur" this option on! Is that the correct base form of that verb. 😋

drgrice1 (approved):
Looks good.

PR #1691, opened 2022-05-13
Hide LMS notice for users who can access the assignment anyways in ProblemSets.pm

somiaj:
When a set must be accessed via an LMS, add a link to the
  LMS_name and only show message about accessing sets via the
  LMS to users who cannot access the set.


pstaabp (approved):
I didn't test--don't have a LMS setup right now.  Looks good though.

somiaj:
You can test it without an LMS setup, just define `$LMS_name='somename';`, `$LMS_url='https://someurl';`, and `$LTIGradeMode='homework';` in your course.conf, and you'll see this notice for students and not for profs once you apply the patch.

drgrice1:
@pstaapb:  You can also use the Mojolicious app that I posted on #1588 to test this (see https://github.com/openwebwork/webwork2/pull/1588#issuecomment-1100912621).

PR #1693, opened 2022-05-16
Add Contrib option to Library Browser

drdrew42:
Now that Contrib is almost as large as the OPL itself, it seems time to support the searching of Contrib alongside the OPL in the Library Browser. 

Testing this PR requires the OPL release created by drdrew42/webwork-open-problem-library@c2eb5be26b7910167fb1b5afb58da0f5879a9d7f

To test:

Load up the OPL release [here](https://github.com/drdrew42/webwork-open-problem-library/releases/tag/vLib-test-5)

Go to Library Browser, select a DBsubject/chapter/section, and check the "Include Contrib" box. Problems from Contrib will appear at the end of search, so you might have to go through multiple pages (or increase the number of results shown per page).


drgrice1 (approved):
Looks good.

Would it be a good idea to add another checkbox to include OPL problems also, that is checked by default?  If unchecked, and `Include Contrib` is checked, then only Contrib problems are shown.  I guess if both are unchecked you get nothing!  I am just throwing this out there, because it might be nice to see what the inclusion of Contrib is adding for those that have already culled the OPL problems.

Another thing that might be nice is if the count were updated to also count the Contrib problems that will be shown when the checkbox is checked or unchecked.  At this point it is a bit misleading when the page shows `There are 120 matching problems` while `Include Contrib` is unchecked, and when you check the box there is no change.  Yet going to the end I do see that Contrib problems are now listed.

pstaabp:
Does this require to rerun OPL-update? or does fetching from the OPL handle this instead?

drdrew42:
use `download-OPL-metadata-release` with a modification to use the `drdrew42` repo instead of `openwebwork` (line 32?)

drgrice1:
Yeah, that is probably simpler than what I did!

pstaabp:
I ran `OPL-update` and got a fatal error 
```
Invalid subject Particle and Quantum Physics
File /opt/webwork/libraries/webwork-open-problem-library/Contrib/BCIT/OpenHomeworkSystemsProjectBCcampus/BCIT_uwlm.pg
DBD::MariaDB::db do failed: Incorrect integer value: '' for column `webwork`.`opl_chapter`.`textbook_id` at row 1 at bin/OPL-update line 820, <IN> line 5323776.
```

I'll try the `download-OPL-metadata-release` as well.

pstaabp:
It appears that even though it failed, I can down select contrib problems!!

drdrew42:
I had to update `File::Fetch` on my dev server... definitely should be mentioned in the upgrade instructions.

bin/update-OPL-statistics.pl
```diff
@@ -168,6 +168,8 @@ BEGIN
   ON `${courseID}_problem_user`.set_id = `${courseID}_set`.set_id
 WHERE (`${courseID}_problem`.source_file LIKE 'Library/%'
   OR `${courseID}_problem`.source_file LIKE 'local/Library/%')
+  OR `${courseID}_problem`.source_file LIKE 'Contrib/%')
+  OR `${courseID}_problem`.source_file LIKE 'local/Contrib/%')
```
drgrice1, line 172: You have some extra ending parentheses here.  These cause failure due to an error in the SQL syntax.

drgrice1:
Starting the lengthy OPL-update process to test this now.  Back in 10 to 15 minutes...

drgrice1:
Wait, its done already?

drdrew42:
Don't merge yet -- still needs some tweaks. e.g. if a course does not have a link to Contrib, then searching Contrib can lead to adding problems that are 'unreachable'

pstaabp:
OPL-update now just downloads the metadata.  Did you run `OPL-update-legacy`?

drgrice1:
@pstaabp:  That was a joke.

drdrew42:
legacy now runs successfully, albeit with a TON of warnings about tags.

drgrice1:
@pstaabp:  Can you give this another try now.  I think we got this all working.  Then we can merge!

pstaabp (approved):
Looks great now.  the OPL-update-legacy ran and ran and ran and ran and finished!!

taniwallach:
This had some new `maketext` calls. The localization files will need to be updated again before the release.

drgrice1:
@taniwallach:  Yeah, another update is needed.  There was another change a while back that isn't in Transifex as well.

drgrice1:
@taniwallach:  All of the current translations strings are now in Transifex.  The final step will be to push those back to here before the final release.  I made an announcement on Transifex a while back about this for translators.  Hopefully we can have some complete translation files!

PR #1695, opened 2022-05-17
Change calls to hasPermissions to use $userID rather than $r->param('user')

dlglin:
For the restricted navigation hasPermissions was checking `$r->param('user')` rather than the `$userID` passed to the function.  This PR reverses that.

I am using a custom LDAP authentication module which sets the user ID for the session to be something different from the username that the student entered on the login page, so in my case for the code to work I need the hasPermissions check to be done on the userID, and not the username from the request.

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+# Copyright &copy; 2000-2020 The WeBWorK Project, https://openwebworkorg.wordpress.com/
 #
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
@@ -752,8 +752,7 @@ sub create_session {
 	}
 
 	my $setID =
-		$r->param('user')
-		&& !$r->authz->hasPermissions($r->param('user'), 'navigation_allowed') ? $r->urlpath->arg("setID") : '';
+		!$r->authz->hasPermissions($userID, 'navigation_allowed') ? $r->urlpath->arg("setID") : '';
 
 	my $Key = $db->newKey(user_id => $userID, key => $newKey, timestamp => $timestamp, set_id => $setID);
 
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/Authen.pm
```diff
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+# Copyright &copy; 2000-2020 The WeBWorK Project, https://openwebworkorg.wordpress.com/
```
drgrice1, line 3: You seem to have accidentally reverted the copyright date and project url.

dlglin: Oops.  Fixed now.

drgrice1 (approved):
I am not sure why I used `$r->param('user')` instead of the passed in `$userID` parameter.  If I had a reason, then there may be a problem with this, but since I can't recall a reason it may just be that I just copied the code from other places where I had to use the `$r->param('user')` for this, and missed that the user id is passed in as a parameter in this case.

pstaabp (approved):

drgrice1:
@pstaabp:  You merged this a bit to soon.

I found the problem here.  To see the problem, first take a proctored gateway quiz as a student, but don't grade the quiz.  Then while that quiz is in progress go view the quiz as a professor.  You will need to enter a valid proctor username and password to enter the quiz.  After you do that go to the end of the page, and you will see an error.  It says `User 'studentID,proctorID' has no PermissionLevel record -- assuming no permission.`  Using some warn statements shows that this is coming from the call to hasPermissions from create_session in Authen.pm.  So the `$userID` that is passed to that method is a string consisting of the student user id and proctor user id separated by a comma.

This is why I had used the `$r->param('user')`.  So more care will be needed here to make this work.

drgrice1:
There is a simpler way to test this, and a more serious manifestation of the issue.  The bug already happens when the proctor username and password is entered for the student.

drgrice1:
It also occurs when the proctor username and password is entered to grade the test for the student.

drgrice1:
My original test method that I outlined also revealed another unrelated bug.  Testing shows this bug is also present in ww 2.16 (and who knows how far back it goes).  If you do as I said, and start a proctored gateway quiz as a student, leave it open, and then view the gateway quiz as a professor (entering the proctor username and password to do that), the user set is changed from a proctored gateway to a regular gateway.  That should not happen until the proctored gateway quiz is graded.

PR #1697, opened 2022-05-18
Graded proctored gateway quizzes are no longer proctored

drgrice1:
On the ProblemSet.pm page for a proctored gateway quiz if there is a version that has been submitted and graded, it should no longer use the proctored gateway quiz module.  It should use the usual gateway quiz module.  The mistake was in the assumption that all versions are of the same type as the template.  This is an incorrect assumption.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSet.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSet.pm
@@ -380,6 +380,7 @@ sub body {
 			$data->{version} = $verSet->version_id;
 			$data->{start} =
 				$self->formatDateTime($verSet->version_creation_time, undef, $ce->{studentDateDisplayFormat});
+			$data->{proctored} = $verSet->assignment_type =~ /proctored/;
 
 			# Display close date if this is not a timed test.
 			my $closeText = '';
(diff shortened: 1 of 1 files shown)


drgrice1:
To test this, create a proctored gateway quiz (or use an existing one if you have one), and take the quiz and submit it for a grade, entering the proctor username and password as needed.  Then go back to the problem set page for the quiz and try to enter the quiz again.  As it is currently you will need to enter the proctor username and password again.  This pull request fixes that. You should not need to enter the proctor username and password to review a completed quiz.

pstaabp:
Just tried this.  I had run out of time, then submitted, needed a proctor password and after entering, I got the following on the bottom of the first page: 
```
User 'homer,peter,g' has no PermissionLevel record -- assuming no permission. at /opt/webwork/webwork2/lib/WeBWorK/Authz.pm line 247.
```

homer is my student, `peter` was the instructor.

pstaabp:
This is what you referenced in #1698.

pstaabp (approved):

drgrice1:
Yeah, that is caused by @dlglin's #1695, and is fixed in #1698.

PR #1698, opened 2022-05-19
Fix the hasPermission calls for the navigation_allowed permission in Authen.pm

drgrice1:
When the proctored gateway quiz module is used and proctor authentication is obtained, the userID string that is passed into create_session is the student user id and the proctor user id separated by a comma.  In addition, when the gateway quiz is graded and proctor authentication is obtained in that case ",g" is appended to the end ofthat.  So those will not work in the hasPermission call.  This strips off the proctor user id and ",g" to pass as the user id to the hasPermission call.

This was introduced in #1695.  See the discussion there.

This does not fix the other unrelated bug revealed in testing this issue, and that I mentioned in #1695.

### DIFF
--- a/lib/WeBWorK/Authen/Proctor.pm
+++ b/lib/WeBWorK/Authen/Proctor.pm
@@ -49,7 +49,7 @@ sub get_credentials {
 
 	my $urlpath = $r->urlpath;
 	my ($set_id, $version_id) = grok_vsetID( $urlpath->arg('setID') );
-	
+
 	# at least the user ID is available in request parameters
 	if (defined $r->param("proctor_user")) {
 		my $student_user_id = $r->param("effectiveUser");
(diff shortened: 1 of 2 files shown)


drgrice1:
This needs even more work.  Unfortunately, a user id can have a comma in it.  So this will strip of to much.

drgrice1:
I think that the Proctor.pm create_session method is just going to have to be rewritten to do everything that the create_session method of Authen.pm does, instead of calling that.

drgrice1:
Okay.  This approach should work.  Instead of stripping of the proctor user id and possible ",g", the true user id is passed in from Proctor.pm via a new parameter.

lib/WeBWorK/Authen/Proctor.pm
```diff
@@ -152,26 +152,26 @@ sub set_params {
 # and then call the default create_session method.
 sub create_session {
 	my ($self, $userID, $newKey) = @_;
-	
-	return $self->SUPER::create_session($self->proctor_key_id($userID), $newKey);
+
+	return $self->SUPER::create_session($self->proctor_key_id($userID), $newKey, $userID);
 }
 
 # rewrite the userID to include bith the proctor's and the student's user ID
```
pstaabp, line 159: typo: bith -> both

Issue #1699, opened 2022-05-20
Quiz Problem Navigation is broken when using problemPanic.pl


pstaabp:
Strange one here.  Steps to reproduce
1. Create a quiz with a few problems. Make sure that the quiz has a single problem per page. 
2. Add the attached problem to the quiz and put it as question #2.
3. Take the quiz.
4. After navigating to problem 2, the navigation menu is no longer functioning and there is the javascript error:
```
undefined is not an object (evaluating 'a?.dataset.confirmDialogMessage')
```
in the gateway.js file.

One of our instructors found this and narrowed it down to the `problemPanic.pl` file.  

[broken_quiz_nav.pg.txt](https://github.com/openwebwork/webwork2/files/8739920/broken_quiz_nav.pg.txt)

drgrice1:
Funny.  Problem number 2 has something to do with Shibboleth Authentication and a bug fix for archiving!

drgrice1:
The problem here is that the problemPanic.pl macro adds a new button with the name submitAnswers.  Which is the same name as the gateway quiz 'Grade Test' button.  The javascript assumes there is only one of these for a gateway quiz, and furthermore, that this button has several data attributes that it uses for the confirmation dialog when the quiz is graded.  Since the added panic button is earlier in the DOM, the gateway.js javascript gets that button instead, and since it does not have the correct dataset information, it causes the error.

This can be fixed, but I doubt that the problemPanic.pl macro will work right in a gateway quiz in any case.

drgrice1:
Note that using the panic button injected by the problemPanic.pl macro will result in the quiz being graded.  That is the problem with it in a gateway quiz.

pstaabp:
Seems like it shouldn't be used in the quiz then.  I'll see what the instructor who found this thinks. 

However, I wonder if there is something we should do for this, like show a warning, to alert the instructor.

PR #1700, opened 2022-05-20
In the gateway quiz javascript, choose the last button name submitAnswers.

drgrice1:
If there are more than one submitAnswers button on a gateway quiz page, choose the last one to use for the confirmation dialog and other things done in the gateway.js file.  This fixes a javascript error caused by things like the problemPanic.pl macro adding another submitAnswers button.  It is highly unlikely that the problemPanic.pl macro will work as expected in a gateway quiz.  This doesn't fix that.

See issue #1699.


pstaabp (approved):

drdrew42 (approved):
👍

PR #1701, opened 2022-05-22
Fix when a proctored gateway quiz is changed to a regular gateway quiz.

drgrice1:
There were two manifestations of the issue here.
    
First, if a professor views an in progress proctored gateway quiz when acting as a student and does not have the permission to record_answers_when_acting_as_student then the quiz should obviously remain a proctored gateway quiz.  Currently a simply the act of viewing a proctored quiz when acting as a student will change a proctored gateway quiz into an unproctored one.  This is clearly really bad.  This means that the student may now continue to work the quiz without needing proctor authentication to grade the quiz.
    
Second, if a professor does have the permission to record_answers_when_acting_as_student, grades the quiz, and uses the last attempt at a quiz in doing so, then the proctored quiz should be changed into an unproctored quiz.  Currently, that is not done.  This means that the student will not be able to continue to work the quiz, but also will need to obtain proctor authentication to review the quiz.
    
Note that both of these bugs were present in previous versions of webwork.
    
Also, note that this includes both #1698 and #1697.  This is because without both of those you can not properly test this pull request as they both interfere with proper behavior of a proctored gateway quiz.

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -736,8 +736,10 @@ sub unexpired_session_exists {
 # clobbers any existing session for this $userID
 # if $newKey is not specified, a random key is generated
 # the key is returned
+# When this is called in Proctor.pm, the actual user id is passed in via $trueUserID.
+# The $userID is modified in that case and will not work in the hasPermissions call.
 sub create_session {
-	my ($self, $userID, $newKey) = @_;
+	my ($self, $userID, $newKey, $trueUserID) = @_;
 	my $r = $self->{r};
 	my $ce = $r->ce;
 	my $db = $r->db;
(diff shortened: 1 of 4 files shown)


pstaabp (approved):
I think I followed both situations laid out here and this does seem to fix everything.  

Since this already has the previous two PRs, perhaps we close those and just review this one?

drgrice1:
Yeah, you can just consider them as a whole here.

drgrice1:
Note that if this pull request is merged, it should automatically close the other two.

drdrew42 (approved):
Tested both, proctor requirements remain unchanged after viewing as student (with and without submit_as_student), and requirements are removed after submitting as student on the final attempt.

PR #1702, opened 2022-05-24
Remove problem link under Homework Sets for quiz

somiaj:
Removes the link in the Main Menu under Homework Sets for the problem when editing a problem for a gateway quiz (the link is broken since it links to the wrong module).

To test this, edit a problem in a gateway quiz from the homework set editor, and you'll see a link to the problem you are editing in the main menu. This removes that broken link.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -826,12 +826,14 @@ sub links {
 					print CGI::start_li({ class => 'nav-item' });
 					print CGI::start_ul({ class => 'nav flex-column' });
 					print CGI::start_li({ class => 'nav-item' });          # $problemID
-					print &$makelink(
-						"${pfx}Problem",
-						text            => $r->maketext("Problem [_1]", $prettyProblemID),
-						urlpath_args    => { %args, setID => $setID, problemID => $problemID },
-						systemlink_args => \%systemlink_args
-					);
+					print $setRecord->assignment_type =~ /gateway/
+						? CGI::a({ class => 'nav-link' }, $r->maketext('Problem [_1]', $prettyProblemID))
+						: &$makelink(
+							"${pfx}Problem",
+							text            => $r->maketext("Problem [_1]", $prettyProblemID),
+							urlpath_args    => { %args, setID => $setID, problemID => $problemID },
+							systemlink_args => \%systemlink_args
+						);
 					print CGI::end_li();                                   # end $problemID
 					print CGI::end_ul();
 					print CGI::end_li();                                   # end $setID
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks good.

Another approach is to instead make the problem id not a link in the case of a problem from a gateway quiz.  This can be done by changing lines 829-834 to
```perl
print $setRecord->assignment_type =~ /gateway/
	? CGI::a({ class => 'nav-link' }, $r->maketext('Problem [_1]', $prettyProblemID))
	: &$makelink(
		"${pfx}Problem",
		text            => $r->maketext('Problem [_1]', $prettyProblemID),
		urlpath_args    => { %args, setID => $setID, problemID => $problemID },
		systemlink_args => \%systemlink_args
	);
```

somiaj:
I think not having anything there is better than the 'Problem N' item there that isn't a link.

Anyone else have any thoughts, would you prefer that we just remove the 'Problem N' link altogether for gateway quizzes, or prefer we keep it there but make it not a link. I can update the pull request depending on what others prefer.

drgrice1:
I disagree with you on that.  The 'Problem n' being there gives an indication of where you are, and what problem you are working on.  I think it is better to have the place marker, than to have nothing.

drgrice1:
Another option is to do what is done in the PGProblemEditor.pm file for problems in gateway quizzes, and make the link work but in an undefined gateway quiz set.

somiaj:
Updated to just make link unclickable. I also prefixed the link under the homework set editor for the problem number with 'Problem ' instead of it just being the number.

I don't have a strong opinion one way or the other, so I updated it (easy enough to revert depending on what others say). Note the menu also tells us where we are at under the Hmwk Sets Editor, and I prefixed that link with 'Problem ' as well so it is no longer just a number.

drgrice1:
Just in case we want to go this way instead, here is code to make the `Problem n` under the set id in the site nav an active link that acts much like the problem editor when "Open in new window" is checked:
```perl
print $setRecord->assignment_type =~ /gateway/
	? &$makelink(
		'WeBWorK::ContentGenerator::GatewayQuiz',
		text            => $r->maketext("Problem [_1]", $prettyProblemID),
		urlpath_args    => { %args, setID => 'Undefined_Set' },
		systemlink_args => {
			params => {
				%{ $systemlink_args{params} // {} },
				editMode       => 'savedFile',
				sourceFilePath => $db->getMergedProblem($eUserID, $setID, $problemID)->source_file
			}
		},
		target => "WW_View"
	)
	: &$makelink(
		"${pfx}Problem",
		text            => $r->maketext("Problem [_1]", $prettyProblemID),
		urlpath_args    => { %args, setID => $setID, problemID => $problemID },
		systemlink_args => \%systemlink_args
	);
```

pstaabp:
First, I noticed that the problem number in the breadcrumbs is also a broken link.  This is a good place to fix that as well. 

I would say that having the "Problem N" as placeholder text is helpful for understanding where this problem fits rather than eliminating it.

drgrice1:
@pstaabp:  I added a separate pull request that fixes the problem number in the breadcrumb.

drdrew42 (approved):
I could go either way on including the "Problem n" text in the main menu while editing a problem here... I suppose, for behavioral consistency, leaving the text and eliminating the link behavior is probably better than removing the text altogether.

This PR keeps the main menu looking consistent, I'm good with it as is.

Issue #1703, opened 2022-05-24
Spans with ltr appear in titles on some pages.


somiaj:
Some of the page titles include the `<span dir="ltr">` in them.

![image](https://user-images.githubusercontent.com/18082412/169942866-006ae2b0-0360-45ba-b5fd-2eefd19f82f3.png)

I tried to track down where this is being set, but wasn't able to. The pages I notice this on are: Problem Set, Problem Set Details, Stats/Student progress. Seems to be any page that has the problem setID as part of the title.

pstaabp:
Fixed in #1704

PR #1704, opened 2022-05-24
Remove HTML from page title

drgrice1:
The URLPath name method can not contain a span with the dir="ltr" attribute when it is used in the page title. However, it needs that markup when it is displayed as a page heading for left-to-right languages like Hebrew.  So a new $displayHTML argument is added to the method that tells it if that is desired.

This argument is then passed for the ContentGenerator.pm title method which is used for the system.template page-title that displayed as a page heading.

The ContentGenerator.pm `title` method is used for what is called "page-title" in the system.template, but is not the "page title".  The ContentGenerator.pm `path` method is use for the page title.  Both call the URLPath name method.

This fixes issue #1703.


somiaj (approved):
This fixes the issue of span tags appearing in the title for me.

pstaabp:
Looks good.  Will merge.

PR #1707, opened 2022-05-26
Change Home to WeBWorK in gateway path.

somiaj:
Use WeBWorK instead of Home for the top path name in gateway quizzes
  (used in breadcrubs and title) to be consistent with other pages.


drgrice1 (approved):
Looks good.  I hadn't noticed that, but that is inconsistent.

drgrice1:
This is simple enough.  I will go ahead and merge.

PR #1709, opened 2022-05-26
Manual problem grader update scores in quizzes

somiaj:
Add the functionality for the manual problem grader to update both problem
  score tables (above and below the problems) and update the hidden score field
  when saving a new score in gateway quizzes.

### DIFF
--- a/htdocs/js/apps/ProblemGrader/problemgrader.js
+++ b/htdocs/js/apps/ProblemGrader/problemgrader.js
@@ -77,6 +77,16 @@
 				return;
 			}
 
+			// Update the hidden problem status fields and score table for gateway quizzes
+			if (saveData.versionId !== '0') {
+				var scoreCells = document.querySelectorAll('table td.score');
+				const numberProblems = scoreCells.length / 2;
+				scoreCells[parseInt(saveData.problemId) - 1].textContent =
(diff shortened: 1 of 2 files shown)


drgrice1 (changes requested):
The code that you added should be in the `success` callback of the first post request.  So after what is currently line 107 with this pull request (after line 97 in the WeBWorK-2.17 branch).  This should not occur if the saving of the score fails.

This code fails to take into account problem randomization.  If I enter a quiz that has problem randomization enabled, and then change the score for a problem using the problem grader, I see the score on some other problem number change.  Not the one that I changed the score for.  I think the only way to do this and reliably obtain the correct score for the problem is to add an `id` or identifying data attribute to the `td.score` elements in the GatewayQuiz.pm file.

drgrice1:
@somiaj:  I added a pull request to this one that makes my suggested changes.

somiaj:
@drgrice1 thanks for the updates. Would you prefer I merged the two commits into a single commit before the final merge?

drgrice1:
For this it is best if they are left as two.

drgrice1 (approved):
This looks good now.  Thanks.

pstaabp (approved):

PR #1710, opened 2022-05-27
Make the problem number in the breadcrumb for a problem in a gateway quiz not a link.

drgrice1:
This was requested by @pstaabp in #1702.  I was going to put it in as a pull request to that one, but realized the code is independent enough that it may as well be its own pull request.


pstaabp (approved):
Fixes the issue.

drdrew42 (approved):
Tested 👍

PR #1714, opened 2022-05-27
Nav menu updates

somiaj:
On siblings link menu, make the active link unclickable since it is a
  link to the current page (similar breadcrubs). Update siblings link
  menu to show the active link on ProblemSet, Stats, and StudentProgress
  pages. Make the Archive this Course link in the file manager not active.

  Remove the nav menu on the Home page since the only link in the
  nav menu is to the current home page.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -660,12 +660,13 @@ sub links {
 	#return "" unless $authen->was_verified;
 
 	# grab some interesting data from the request
-	my $courseID = $urlpath->arg("courseID");
-	my $userID = $r->param('user');
-	my $eUserID   = $r->param("effectiveUser");
-	my $setID     = $urlpath->arg("setID");
-	my $problemID = $urlpath->arg("problemID");
-	my $achievementID = $urlpath->arg("achievementID");
+	my $courseID      = $urlpath->arg('courseID');
+	my $userID        = $r->param('user');
+	my $urlUserID     = $urlpath->arg('userID');
+	my $eUserID       = $r->param('effectiveUser');
+	my $setID         = $urlpath->arg('setID');
+	my $problemID     = $urlpath->arg('problemID');
+	my $achievementID = $urlpath->arg('achievementID');
 
 	# Determine if navigation is restricted for this user.
 	my $restricted_navigation = $authen->was_verified && !$authz->hasPermissions($userID, 'navigation_allowed');
@@ -914,14 +915,26 @@ sub links {
 
 				print CGI::start_li({ class => 'nav-item' }); # Stats
 				print &$makelink("${pfx}Stats", urlpath_args=>{%args}, systemlink_args=>\%systemlink_args);
-				if ($userID ne $eUserID or defined $setID) {
+				if ($userID ne $eUserID or defined $setID or defined $urlUserID) {
 					print CGI::start_ul({ class => 'nav flex-column' });
-					if ($userID ne $eUserID) {
+					if (defined $urlUserID) {
 						print CGI::li({ class => 'nav-item' },
 							&$makelink("${pfx}Stats",
-							   	text => "$eUserID",
-							   	urlpath_args => { %args,statType => "student", userID => $eUserID },
-							   	systemlink_args => \%systemlink_args));
+								text => $urlUserID,
+								urlpath_args => { %args, statType => "student", userID => $urlUserID },
+								systemlink_args => \%systemlink_args
+							)
+						);
+					}
+					if ($userID ne $eUserID && (!defined $urlUserID || $urlUserID ne $eUserID)) {
+						print CGI::li({ class => 'nav-item' },
+							&$makelink("${pfx}Stats",
+								text => $eUserID,
+								urlpath_args => { %args, statType => "student", userID => $eUserID },
+								systemlink_args => \%systemlink_args,
+								active => $urlpath->type eq 'instructor_user_statistics' && !defined $urlUserID
+							)
+						);
 					}
 					if (defined $setID) {
 						# make sure we don't try to send a versioned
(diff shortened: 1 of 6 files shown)


drgrice1 (commented):
After thinking about this more, I think it is not a good idea to make the active page not a link.  At least not in all cases, maybe not ever.  The thing is that data can change on the server while you are viewing a page.

For example, if you are viewing the statistics for a homework set, and students are currently working on that set and thus changing the data.  You can then click on the set in the siblings menu to obtain updated information from the server, and monitor progress.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -1030,7 +1030,8 @@ sub links {
 						&$makelink("${pfx}FileManager",
 							text => $r->maketext("Archive this Course"),
 							urlpath_args => { %args },
-						   	systemlink_args => \%augmentedSystemLinks));
+							systemlink_args => \%augmentedSystemLinks,
+							active => 0));
```
drgrice1, line 1034: Why did you add `active => 0` here?

somiaj: The 'Archive this Course' link is active otherwise, so when you are on the file manager link, both 'File Manager' and 'Archive this Course' are flagged as active, and since this is a link to the FileManager with `?archiveCourse=1`, the makelink subroutine flags it as active without adding this.

drgrice1: I see now.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -131,10 +131,23 @@ sub siblings {
...
+				);
+			} else {
+				print CGI::li(
+					{ class => 'nav-item' },
+					CGI::a(
+						{ href => $self->systemLink($userStatisticsPage), class => 'nav-link' },
+						"$last_name, $first_name  ($user_id)"
+					)
+				);
+			}
```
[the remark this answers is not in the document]
somiaj: When viewing the stats for a homework set, there is a usable link in the main navigation menu, so you don't need to also use the siblings link to do what you describe. The only case this would not work is when viewing the stats/student progress of a student, since in that case there is no link put in the main menu page. I too use this method to view progress when students are working on a quiz/set, but haven't used this for watching single student on all sets.

There is one put there when acting as a student, but it is for the student you are acting as, not the one you are viewing the stats/progress for. Though this one case may be reason enough to just keep all active links in the siblings menu as usable links.

The other option is to add a link to the student being viewed via stats/student progress similar to what is done when viewing a homework set. This would mean that there would be two students in the main menu when acting as a student and viewing the progress of another student.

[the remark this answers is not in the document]
drgrice1: I was actually looking at stats for a student, and started writing my comment that way, but then changed it to a set instead but didn't actually look.  Yeah, I guess there is a link that will work for a set.  It is useful for refreshing for a student as well though.  I have used it that way.  Both for stats and progress.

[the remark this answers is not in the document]
somiaj: @drgrice1 I updated the logic to put the conditional inside the hash as you suggested in all the places I was working on.

I added the name of the student you are viewing the stats/progress of to the main menu, this way you can refresh that page easily, and this is more consistent with how sets are treated. If you are acting as a student you should see both the student you are viewing and the acting as student in the main menu. One tiny side effect was when editing the sets for a student, `instructor/users/userID` or `instructor/users/userID/sets`, the links for that student also appear in the main menu. I didn't see any other locations that `userID` is parsed from the urlpath.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -914,14 +915,26 @@ sub links {
...
+					if ($userID ne $eUserID && (!defined $urlUserID || $urlUserID ne $eUserID)) {
 						print CGI::li({ class => 'nav-item' },
 							&$makelink("${pfx}Stats",
-							   	text => "$eUserID",
-							   	urlpath_args => { %args,statType => "student", userID => $eUserID },
-							   	systemlink_args => \%systemlink_args));
+								text => $eUserID,
+								urlpath_args => { %args, statType => "student", userID => $eUserID },
+								systemlink_args => \%systemlink_args,
+								active => !defined $urlUserID
```
drgrice1, line 935: With this logic the student that is being acted as is highlighted anytime that you are not on the stats or progress pages for a student.  This should be
```perl
active => $urlpath->type eq 'instructor_use_statistics' && !defined $urlUserID
```

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -944,14 +957,26 @@ sub links {
...
+							)
+						);
+					}
+					if ($userID ne $eUserID && (!defined $urlUserID || $urlUserID ne $eUserID)) {
+						print CGI::li({ class => 'nav-item' },
+							&$makelink("${pfx}StudentProgress",
+								text => $eUserID,
+								urlpath_args => { %args, statType => "student", userID => $eUserID },
+								systemlink_args => \%systemlink_args,
+								active => !defined $urlUserID
```
drgrice1, line 977: This should be
```perl
active => $urlpath->type eq 'instructor_use_progress' && !defined $urlUserID
```

drgrice1 (changes requested):
This is getting closer, but there is an issue when acting as a student.  The link for that student under stats and progress is always marked as active anytime you are not on a stat or progress page for a student.  It should only be active when on the stat or progress page for that student.  The code to fix this is in the comments.

lib/WeBWorK/ContentGenerator.pm
drgrice1: Another way to do this that is more efficient is to instead change this line (and the corresponding line for the progress link) to:
```perl
defined $urlUserID ? (active => 0) : ()
```

drgrice1: Take your pick of the approaches.  Now that I think about it more, the first is actually more efficient because the `makelink` method for determining when a nav link should be marked active is not at all efficient.  With the second suggestion, the `makelink` method will be used when the `$urlUserID` is not defined.

The `makelink` method really needs to be improved.  Before comparing path args it should compare path types for the current path and the link path.  It should then check that the number of path args are the same for those paths.  Then it should compare the args.  Don't try to fix that method for now though.

lib/WeBWorK/ContentGenerator.pm
somiaj: Thanks, in my testing I was focused on testing the stats/progress page and other pages where urlUserID would have been defined, so over looked this case. Updated.

pstaabp:
If (not acting as a student), select Stats/Student Progress and select a student.  The username now appears as a submenu item in both Stats/Student Progress (nice).  However, if you go to HwkSetEditor and select a homework set, that set is no longer a submenu item in HwkSetEditor/Stats/Student Progress.  It is if you get there via HomeworkSets link.  Was this intentional?

pstaabp:
Hold on.  I can't reproduce this now.

somiaj:
@pstaabp This should not have changed the behavior with homework sets at all. The homework set links are shown any time the setID is extracted from the URL. The only thing I can think of is you clicked on the edit pencil when 'selecting a homework set', which doesn't link to the set editor, doesn't put the setID in the url, and thus won't show the links. I'm glad you are no longer able to reproduce the issue.

drgrice1 (approved):
This looks good now.

somiaj:
@pstaabp I think I figured out the possible issue you ran across (as I just ran across a similar thing). If you are not assigned to a set, the links to the set don't appear in the Main Menu as sub menu items (unless you are acting as a student who is assigned to the set). This is an issue outside the scope of this pull request, but could be fixed if you think this is an issue (vs working as intended). And since it is possible to get to sets you are not assigned to via the HwmkSetEditor, this seems possible to what you found.

The code that does this is around lines 684-692 in `lib//WeBWorK/ContentGenerator.pm`:

```perl
	# It's possible that the setID and the problemID are invalid, since they're just taken from the URL path info.
	if ($authen->was_verified) {
		if (defined $setID && $db->existsUserSet($eUserID, $setID)) {
			$problemID = undef unless (defined $problemID && $db->existsUserProblem($eUserID, $setID, $problemID));
		} else {
			$setID     = undef;
			$problemID = undef;
		}
	}
```
Add some different logic if the user can see some instructor tools. Will also have to do something similar to what is done on gateway quizzes to make the top links to the Set/Problem unclickable since they won't work as the user isn't assigned to the set.

Though I think there are other issues with professors when they aren't assigned to a set, so probably best to leave this alone and deal with it if/when someone wants to track all those additional issues down.

pstaabp (approved):
I think I'm fine with it without additional change.  Will approve and merge.

PR #1717, opened 2022-05-30
Fix issue #1716.

### LINKED ISSUE #1716: Problem Editor doesn't return to original problem when creating a new version for a file that exists.
Use the problem editor to edit any problem that is not the last problem in an assignment set.

Create a new version of the problem you are working on and append it to the end of the current set.

Use a file name that already exists. The problem editor doesn't create a new problem due to the file already existing, and states, `The text box now contains the source of the original problem. You can recover lost edits by using the Back button on your browser.`, but instead of containing the text of the original problem, it contains the text of the last problem in the homework set.

Seems that in this case of appending to the end of the set, the problem editor goes to the last problem number, instead of the problem number that was being worked on.

drgrice1:
If a file is not saved when saving as a new version, then do not change the problem number of the redirect.  This makes the message that is displayed correct that the "text box now contains the source of the original problem."

Fixes #1716.


drdrew42 (approved):
RIP 'redirectoed' 🦶 💀

pstaabp (approved):

PR #1722, opened 2022-06-10
Docker updates for WW 2.17: ADDITIONAL_BASE_IMAGE_PACKAGES and updated OPL metadata download approach

taniwallach:
1. Modify the approach Docker uses to get the OPL metadata - to use the new approach.
2. Add a new feature to make it possible to use a new build time argument to add additional packages to the stage 1 (OS) image so which is more efficient that adding them each time a container is started.
```
docker build --tag webwork-base:forWW217 --build-arg ADDITIONAL_BASE_IMAGE_PACKAGES="nano less" -f DockerfileStage1 .
```

### DIFF
--- a/docker-config/docker-entrypoint.sh
+++ b/docker-config/docker-entrypoint.sh
@@ -59,7 +59,7 @@ if [ ! -d "$APP_ROOT/libraries/webwork-open-problem-library/OpenProblemLibrary" 
   cd $APP_ROOT/libraries/
   /usr/bin/git clone -v --progress --single-branch --branch main https://github.com/openwebwork/webwork-open-problem-library.git
 
-  # The next line forces the system to run OPL-update or load saved OPL tables below, as we just installed it
+  # The next line forces the system to download and install the OPL metadata later
   touch "$APP_ROOT/libraries/Restore_or_build_OPL_tables"
 fi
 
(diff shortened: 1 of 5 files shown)


taniwallach:
@mgage - The idea of the `ADDITIONAL_BASE_IMAGE_PACKAGES` is intended to make it easy to have persistent local additional packages in the image itself. I did not find a manner where this can be done without the extra packages being set on the `docker build` command line. I do not like the startup delay caused when the packages are installed at container startup time using `ADD_APT_PACKAGES`.

drgrice1:
I haven't tested this yet, but the general concept of the things done here looks good.  I will try to do some testing this weekend.

taniwallach:
Forcing an existing container to download the metadata from GitHub (on the next startup) can be done by starting a shell in the running container `docker container exec -it webwork2_app_1 bash` and then running:
```
touch $APP_ROOT/libraries/Restore_or_build_OPL_tables
rm $APP_ROOT/libraries/webwork-open-problem-library/TABLE-DUMP/OPL-tables.sql
```
so that the next `docker-compose up` will need to download the metadata.

When `htdocs/DATA` is not persistent, on a second restart it should not download the metadata again from GitHub but just reload it and copy the JSON file back to where they are needed.

docker-config/docker-entrypoint.sh
```diff
@@ -159,32 +159,22 @@ if [ "$1" = 'apache2' ]; then
...
-        echo "Checking out tag $OPL_TAG"
+        # Download the metadata and install it
+        export SKIP_UPLOAD_OPL_statistics=1
+        if [ ! -d $APP_ROOT/libraries/webwork-open-problem-library/JSON-SAVED ]; then
+          mkdir $APP_ROOT/libraries/webwork-open-problem-library/JSON-SAVED
+        fi
+        wait_for_db
+        $WEBWORK_ROOT/bin/OPL-update
+        cd $APP_ROOT/libraries/webwork-open-problem-library
         git checkout $OPL_TAG
```
drgrice1, line 176: This line (line 176) should be deleted.  The `bin/download-OPL-metadata-release.pl` script already has checked out the branch corresponding to the metadata release at this point.  That script is run by `bin/OPL-update`.  Of course then line 175 before it isn't needed either.

I tested the build without those two lines, and it works as promised.  Running `docker container exec -it webwork2_app_1 git -C /opt/webwork/libraries/webwork-open-problem-library branch` shows that the branch corresponding to the current metadata release (currently v2022-05-30) is checked out.

drgrice1 (commented):
This looks good.  One minor detail with a couple of unneeded lines.

docker-config/docker-entrypoint.sh
taniwallach: Thanks. Lines deleted.

mgage (approved):
I was able to build the webwork2 container from scratch including adding vim using the build-arg.

After a couple of tries I was able to rebuild the library tables.  I found that starting from docker-compose build and then using docker-compose up was the easiest way for me to make sure that I was using the latest OPL-update script

drgrice1:
I will merge this now.

PR #1726, opened 2022-06-20
Make raw format identify version

Alex-Jordan:
This will allow PreTeXt's `pretext-webwork.js` to identify the WW version and make certain decisions accordingly.


drgrice1 (approved):
This looks good.  Although, PreTeXt will not have this information until PG version 2.17.  So PreTeXt will not be able to identify the version of anything prior to that.

Alex-Jordan:
Yeah, that should be OK. The logic will be like if this is not there, then handle it the 2.16- way.

drgrice1:
Yeah.  Not much else can be done I guess.

pstaabp (approved):

drgrice1:
I am going to merge this.  It has two approvals!

Issue #1727, opened 2022-06-22
check_latex artifacts


Alex-Jordan:
Running `check_latex` is leaving `check_latex.aux` and `check_latex.log` in the `bin` folder. Is that desired? If not, we could delete them at the end. Or run the whole thing in a temporary folder that gets trashed.

drgrice1:
That is not desired.  Either of your suggestions are appropriate.  Probably using a temporary directory is best.  In fact that is what is already implemented in the script.  However, I forgot to add a `cd $TEMP_DIR` after creating the temporary directory and before running pdflatex.  Thanks for finding this error.

pstaabp:
fixed in #1729

PR #1729, opened 2022-06-22
Fix the temporary directory usage in the check_latex script

drgrice1:
The script was creating a temporary directory, but the only thing it put
into the directory was the command line output of the pdflatex
execution.

Now pdflatex is run in the temporary directory so that all of the files
created by pdflatex are created there, and subsequently deleted.

See #1727


Alex-Jordan (approved):
Looks good to me. I haven't actually ran this, but yeah, it looks good.

pstaabp (approved):
I did run it, but since it worked, didn't check what happens if it fails.  I think it looks good.

PR #1730, opened 2022-06-22
Add the extra css and js files to the json object that is output for the raw format.

drgrice1:
PreTeXt needs this for its interactive problems.

### DIFF
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -172,13 +172,17 @@ sub formatRenderedProblem {
 		push @cssFiles, @{ $rh_result->{flags}{extra_css_files} };
 	}
 	my %cssFilesAdded;    # Used to avoid duplicates
+	my @extra_css_files;
 	for (@cssFiles) {
 		next if $cssFilesAdded{ $_->{file} };
 		$cssFilesAdded{ $_->{file} } = 1;
 		if ($_->{external}) {
+			push(@extra_css_files, $_);
 			$problemHeadText .= CGI::Link({ href => $_->{file}, rel => 'stylesheet' });
 		} else {
-			$problemHeadText .= CGI::Link({ href => getAssetURL($ce, $_->{file}), rel => 'stylesheet' });
+			my $url = getAssetURL($ce, $_->{file});
+			push(@extra_css_files, { file => $url, external => 0 });
+			$problemHeadText .= CGI::Link({ href => $url, rel => 'stylesheet' });
 		}
 	}
 
@@ -204,6 +208,7 @@ sub formatRenderedProblem {
 		if ($self->{inputs_ref}{outputformat} // '') eq 'sticky';
 
 	# Add JS files requested by problems via ADD_JS_FILE() in the PG file.
+	my @extra_js_files;
 	if (ref($rh_result->{flags}{extra_js_files}) eq 'ARRAY') {
 		my %jsFiles;
 		for (@{ $rh_result->{flags}{extra_js_files} }) {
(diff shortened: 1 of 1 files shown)


pstaabp:
This looks fairly straightforward.  Not sure how I could test though.  Thoughts to do this? Or I could just give a thumbs up.

drgrice1:
This basically does the same thing that the standalone renderer does for webwork3.  To test this you could load a problem using the raw output and see that the output contains the list of js and css files with hashed names.  You can also open a problem in the homework sets editor or library browser to see that the other formats still work correctly.

Alex-Jordan:
I have this on a branch. Plus Glenn's corresponding edits to the PreTeXt repo file pretext-webwork.js,  plus a few edits of my own. And it's basically working as here:

https://spot.pcc.edu/~ajordan/temp/webwork.html

There's some cosmetic stuff on that page to continue working through. For example I think we will drop the "Correct" and "Incorrect" messages in favor of something more subtle. But it looks like it's functional.

In related news, it's working on Runestone. So if that page is part of a book in a Runestone course, the reader can do the exercises as they read and the score, feedback etc gets recorded by Runestone. Brad and I just tested that literally two minutes ago.

drgrice1:
I see that you found a better icon for the message button than the speech bubble we used before!

One thing that isn't working right yet is the Correct/Incorrect messages in scaffolds.  If the scaffold is closed, then the popover messages are still there floating over the closed scaffold.  Although, if you are going to replace those popovers, then that will fix the issue.

taniwallach:
Maybe the raw output mode could also include either the data collected in `@CSSLoads` and `@JSLoads` which have the hashed names of the standard dependencies or a JSON structure similar to that added in the PR but for those files which get data about them in `@CSSLoads` and `@JSLoads`?

drgrice1:
@taniwallach:  I am not sure what you mean exactly.

taniwallach:
> @taniwallach: I am not sure what you mean exactly.

The new asset build systems changes the names of the "standard" JS/CSS files included by all problems when a new build is made. As a result, a remote system cannot know what the correct URLs for those files will be.

To handle that change, in https://github.com/openwebwork/webwork2/pull/1599/ you moved the JS/CSS load lines in the `lib/WebworkClient/simple_format.pl` (and other "regular" format files) to lines which are added to `$problemHeadText` by code in `lib/FormatRenderedProblem.pm` which depends on the `@CSSLoads` and `@JSLoads`. That approach provides the current URLs for the "regular" formats of `html2xml`. The `json` format gets both those arrays provided as part of the provided JSON structure, but it does not seem that data was added anywhere in the `raw` format.

My proposal is to also add the data in those arrays (or something similar) to the `raw` format. Ideally, the structure being used would be similar to what is being done in this PR for JS/CSS files requested by the PG code.

drgrice1:
I am confused.  This pull request adds that data to the raw format.

drgrice1:
I see.  You are specifically talking about the @CSSLoads and @JSLoads.  Really, you shouldn't need that.  All of those are specific to webwork2 theming and or third party inclusions.  In most cases anyone using the raw format should either provide anything there that is needed from other sources (served elsewhere like a CDN).

drgrice1:
This is the approach at this point at least.  This is related to the comments on line 172 and 215 of pg/macros/PG.pl.  The point is that the files mentioned in the comments there are third party files.  In most cases those can be used from a CDN.  In some cases the "requestor" as I call it in the comments there may even need to use custom versions (for example the bootstrap.css and bootstrap.js files).  This is also the approach the standalone renderer is going to use for these.  Webwork3 needs to use a custom bootstrap.css file that does not conflict with Quasar.

The other files added to the @CSSLoads and @JSLoads lists in FormatRenderedProblem.pm are for theming.  Note that the math4 files no longer contain anything for rendering problems.  That has all been moved to pg/htdocs/js/apps/Problem/problem.scss, and that is included in the @extra_css_files.

Alex-Jordan:
My previous link should no longer work. But here is where I intend to keep samples that are up to date:

https://spot.pcc.edu/~ajordan/temp/pretext-webwork

Alex-Jordan:
> Although, if you are going to replace those popovers, then that will fix the issue.

I plan to make it so that the "Correct!", "Incorrect", and "90% correct" (etc) titles go away. Instead, inside the input field there will be an indicator like when you are filling a form and you try to submit some  input that does not validate.

So then there will still be popups but only for any extended feedback. It'll reduce the popup presence a lot and I'll check out the scaffold issue then.

taniwallach:
@Alex-Jordan - Is this ready for merging in your opinion?

Alex-Jordan (approved):
This has been working in my experiments/development of PreTeXt WeBWorK integration for 2.17.

Alex-Jordan:
@taniwallach Yes, I think so.

drgrice1:
I have added the other css and js files to the raw output mode as well as @taniwallach requested.  PreTeXt does not need those, but others using the raw format might.

taniwallach (approved):
@drgrice1 Thanks for data on the adding the third party css and javascript to the raw format output.

taniwallach:
This had @Alex-Jordan's OK as providing the changes needed for PreTeXt with WeBWorK 2.17. With the final release impending - I'm merging this.

PR #1731, opened 2022-06-23
Fix the handling of moving/removing the temporary course archive file.

drgrice1:
In the process of creating a course archive, first a temporary tar file is created in the courses directory.  Then that file is moved to the requested location.

There were several issues with the coding of how the temporary file was moved to the requested location.

First the perl `rename` method was used.  If the requested location is not on the same file system as the courses directory that will fail.  So this switches to using the `move` method in the `File::Copy` module. Note that webwork already uses this module elsewher, so this is not a new dependency.

Second, in the case that the `rename` method fails, the error was never checked.  The result was that the temporary archive file is left in the courses directory, and no errors are reported even though the requested archive file does not exist.

Third, in the case that a previous archive file exists in the courses directory, the `croak` call occured before the `unlink` call to delete the temporary archive file (in the clean up according to the comments). Thus the temporary archive file is not actually deleted.

This also removes the space at the beginning of the temporary file name. That worked, but why have it?

Note:  Hide spaces to review without seeing the removal of the space at the end of lines (done automatically by my text editor).

### DIFF
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -28,6 +28,7 @@ use Carp;
 use DBI;
 use WeBWorK::Debug;
 use File::Path qw(rmtree);
+use File::Copy qw(move);
 use File::Spec;
 use String::ShellQuote;
 use WeBWorK::CourseEnvironment;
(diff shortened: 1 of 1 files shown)


pstaabp:
This works for me. I haven't tested if it fails, which seems to be a big part of this.  Ideas on how to test these?

drgrice1:
Some parts of this pull request are hard to test.  Some of the failures this fixes were not actually possible.  There is redundancy in the code, that this pull request doesn't go far enough with to clean up.

To test the important part, that is the rename/move part, you would need to have the /opt/webwork/courses directory on a different partition from the final resting place of the course archive.  This doesn't happen when archiving a course from the web ui.  So you would need to use a script.

This pull request isn't too important though, so don't spend a lot of time unless I do and give specific testing instructions.

PR #1737, opened 2022-07-11
Update bin/dev_scripts/update-localization-files so that it will

taniwallach:
At present the recently revised `bin/dev_scripts/update-localization-files` pulls in too much from the `webwork2/conf` directory.
At present it was getting a string from `/opt/webwork/webwork2/conf/snippets/hardcopyThemes/common/CAPA.tex` which should not be included, but potentially other junk would get pulled in accidentally.
Fix this by explicitly listing what files in `webwork2/conf` should be processed.

### DIFF
--- a/bin/dev_scripts/update-localization-files
+++ b/bin/dev_scripts/update-localization-files
@@ -55,7 +55,7 @@ cd $LOCDIR
 
 echo "Updating $WEBWORK_ROOT/webwork2.pot"
 
-xgettext.pl -o webwork2.pot -D $WEBWORK_ROOT/lib -D $WEBWORK_ROOT/conf -D $PG_ROOT/lib -D $PG_ROOT/macros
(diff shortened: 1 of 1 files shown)


drgrice1:
Good catch.  This is odd though.  I see that it is pulling in the string from CAPA.tex, but why?  It isn't marked by `maketext` or `x`.

drgrice1 (approved):

taniwallach:
> This is odd though. I see that it is pulling in the string from CAPA.tex, but why? It isn't marked by `maketext` or `x`.
I don't know. I'm guess that the underlying programs use different methods for different types of files, and it is detecting that file as being LaTeX or some non-Perl file type. I did not think there was any reason to investigate further, and as there are lots of other files under `webwork2/conf` there seems little point in scanning those files and waiting for the next surprise.

BTW Thanks for all the work to make the script ready for the Transifex oriented workflow instead of the old one.

drgrice1:
Yeah, probably best for now to just list the two files we want translated.

pstaabp (approved):

PR #1738, opened 2022-07-11
Revisions to controls of SendMail.pm and some changes to what should be translated

taniwallach:
In `lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm`:
  1. The value `n/a` was being translated, but later a test against the plain value `n/a` is used. That will not work. I do not think that there is any need to translate this special string value. If I am not mistaken only instructors should see it.
  2. The value `none` was being translated. It is a one work term, so could require different translations in different contexts. It is used for when there was no test date/time - so using the plain `n/a` seems suitable here also.

In the other file replace two ambiguous short strings used as labels and set to be translated with slightly longer and explicit strings.

If this is merged - the localization file `webwork.pot` should be updated and synced to Transifex.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
@@ -696,8 +696,8 @@ sub print_form {
 								name   => 'send_to',
 								values => [ 'all_students', 'studentID' ],
 								labels => {
-									all_students => $r->maketext('all'),
-									studentID    => $r->maketext('selected')
+									all_students => $r->maketext('all students'),
+									studentID    => $r->maketext('selected students')
 								},
 								default         => $r->param('send_to') // 'studentID',
 								linebreak       => 0,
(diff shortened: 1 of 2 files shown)


drgrice1:
If you are going to change the strings in `Email.pm`, then more needs to be done.

Without this pull request it reads `Send to: | [ ] all | [ ] selected | students`.  This is where `|` is an input group separator, and `[ ]` is a radio button.

With this pull request is reads `Send to: | [ ] all students | [ ] selected students | students`, which is not right.  Basically if `all students` is checked, then it reads `Send to all students students`.

So with this pull request the `students` label at the end of the input group should be removed.

drgrice1:
@taniwallach:  I added a pull request to your branch for this pull request that changes the radio checkboxes on the SendMail.pm page in what I believe is a much better way for translations.  Please consider it.

taniwallach:
@drgrice1 Thanks - your revised code was merged. I only tested the behavior is the browser without sending mails out as the testing was in a local docker based test platform.

drgrice1 (approved):
Looks good.

drgrice1:
I am going to merge this in a few hours unless there are objections.  I want to deal with the Transifex translations, and that depends on this.

taniwallach:
I had hoped that someone else would test the `SendMail.pm` changes in addition to us, but unless someone else on the release team has a chance to test soon, I would say that you should go ahead and merge it.

PR #1742, opened 2022-07-14
Set a few form/display fields to force left-to-right direction.

taniwallach:
Force a few more things to be left-to-right where in Hebrew courses the default would have been right-to-left which seems less appropriate. The changes should not have an effect on courses in left-to-right languages.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -314,6 +314,7 @@ EOF
 					values   => $dirs,
 					labels   => $dirlabels,
 					class    => 'form-select',
+					dir      => 'ltr',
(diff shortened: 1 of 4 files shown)


drgrice1 (approved):
There is probably a lot more of this needed isn't there.  How many places in the interface do we display a file name, email address, or username?

drgrice1:
This is trivial enough, that I will go ahead and merge.

taniwallach:
> There is probably a lot more of this needed isn't there. How many places in the interface do we display a file name, email address, or username?

There probably are more - but the `FileManager.pm` has been annoying in Hebrew courses for a long time, but not that so much else is fixed - was more glaringly noticable. The other places were places I realized there would be filenames/paths+filenames quickly and use often.

Whatever else comes up will wait for the future.

PR #1744, opened 2022-07-15
Move some docker files to be dist files (and out of main directory)

taniwallach:
Move some docker files so that updates to the standard files will not interfere with locally modified versions.

Initial use requires copying them to the regular location, and on updates - local changes should be merged with the changes from the updated `dist` file with suitable care.

1. `docker-compose.yml` moved to `docker-config/docker-compose.yml.dist`
2. `.env` moved to `docker-config/env.dist`


mgage (approved):
This looks fine to me.  We just need to remember to remind new users to compare to the dist files when they update from the github repository.  It we can find a way to automate comparison of the files  that might be useful, but it can wait.

drgrice1:
@mgage:  That is true of any of the `.dist` files.  That is why we always add the reminder to the release notes and upgrade instructions to check config files.

drgrice1 (approved):

drgrice1:
This is ready to merge.

PR #1761, opened 2022-07-19
Add the package-lock.json file back.

drgrice1:
It seems that bootstrap version 5.2.0 was recently released and is not entirely compatible with version 5.1.3.  This causes the bootstrap.scss file to fail to compile.

This adds a package-lock.json file from before the version update occured to lock us in to 5.1.3.

Another way to fix this is to import the `node_modules/bootstrap/scss/maps` file in our bootstrap.scss file. However, that may only be a temporary fix as bootstrap may update again and change something else that breaks our build.

Now that we are also insisting on everyone using node 16 this should not cause the problem that we had before.

Sigh.  Hot fix to webwork 2.17 after one day.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ htdocs/themes/*/images/*
 !htdocs/themes/*/images/webwork_logo.svg
 !htdocs/themes/math4/images/webwork_square.svg
 htdocs/themes/*/*.css
-htdocs/package-lock.json
 htdocs/themes/*
 !htdocs/themes/math4
 !htdocs/themes/math4-red
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -1,0 +1,2836 @@
+{
+    "name": "webwork.javascript_package_manager",
+    "lockfileVersion": 2,
+    "requires": true,
+    "packages": {
+        "": {
+            "name": "webwork.javascript_package_manager",
+            "license": "GPL-2.0+",
+            "dependencies": {
+                "@fortawesome/fontawesome-free": "^6.0.0",
+                "bootstrap": "~5.1.3",
+                "codemirror": "^5.65.2",
+                "flatpickr": "^4.6.9",
+                "iframe-resizer": "^4.3.2",
+                "jquery": "^3.6.0",
+                "jquery-ui-dist": "^1.13.1",
+                "luxon": "^2.3.1",
+                "mathjax": "^3.2.0",
+                "sortablejs": "^1.14.0"
+            },
+            "devDependencies": {
+                "autoprefixer": "^10.4.1",
+                "chokidar": "^3.5.3",
+                "cssnano": "^5.1.4",
+                "postcss": "^8.4.5",
+                "rtlcss": "^3.5.0",
+                "sass": "^1.45.2",
+                "terser": "^5.10.0",
+                "yargs": "^16.2.0"
+            }
+        },
+        "node_modules/@fortawesome/fontawesome-free": {
+            "version": "6.1.1",
+            "resolved": "https://registry.npmjs.org/@fortawesome/fontawesome-free/-/fontawesome-free-6.1.1.tgz",
+            "integrity": "sha512-J/3yg2AIXc9wznaVqpHVX3Wa5jwKovVF0AMYSnbmcXTiL3PpRPfF58pzWucCwEiCJBp+hCNRLWClTomD8SseKg==",
+            "hasInstallScript": true,
+            "engines": {
+                "node": ">=6"
+            }
+        },
+        "node_modules/@jridgewell/gen-mapping": {
+            "version": "0.3.2",
+            "resolved": "https://registry.npmjs.org/@jridgewell/gen-mapping/-/gen-mapping-0.3.2.tgz",
+            "integrity": "sha512-mh65xKQAzI6iBcFzwv28KVWSmCkdRBWoOh+bYQGW3+6OZvbbN3TqMGo5hqYxQniRcH9F2VZIoJCm4pa3BPDK/A==",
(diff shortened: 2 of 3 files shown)


drdrew42 (approved):
This is why we can't have nice things.

drgrice1:
To test and see why this pull request is needed, delete your package-lock.json file and node_modules directory, and run `npm install`.  There will be errors and the bootstrap.css files for the themes will not be created.  With this pull request it will work.

drgrice1:
For future reference (when updating bootstrap) to fix this the `bootstrap.scss` file needs to import `node_modules/bootstrap/scss/maps` before it imports `node_modules/bootstrap/scss/utilities`.  The `_maps.scss` file was added to bootstrap version 5.2.0, and did not exist for version 5.1.3.

taniwallach:
I confirm that the problem exists before the `package-lock.json` file from the PR was added, and is fixed after it is added.

However, with the `package-lock.json` provided, npm reports there being "1 moderate severity vulnerability", and I am not certain that it is the best idea to force versions to this level of detail for WeBWorK sites.

Would it not be possible to solve the same problem at present by instead modifying `package.json` to have
```
        "bootstrap": "~5.1.3",
```
instead of
```
        "bootstrap": "^5.1.3",
```
so that the update to 5.2.0 would not be allowed yet? An `npm install` after that change also seems to avoid any problem in the `generate-assets.js` process.

---

About the reported security vulnerability - it is not clear to me if the recommended update to the `terser` package will cause problems for the WeBWorK javascript code. If we are confident that it will not, and decide to apply this hotfix to provide `package-lock.json` then it probably makes sense to have the `package-lock.json` refer to version 5.14.2, the version after the security fix.

---

`npm audit` gives:
```
# npm audit report

terser  5.0.0 - 5.14.1
Severity: moderate
Terser insecure use of regular expressions before v4.8.1 and v5.14.2 leads to ReDoS - https://github.com/advisories/GHSA-4wf5-vphf-c2xc
fix available via `npm audit fix`
node_modules/terser

1 moderate severity vulnerability

To address all issues, run:
  npm audit fix
```
After running `cp package-lock.json package-lock.json.saved ; npm audit fix` I ran `diff package-lock.json package-lock.json.saved` and got
```
1567,1569c1567,1569
<             "version": "5.14.2",
<             "resolved": "https://registry.npmjs.org/terser/-/terser-5.14.2.tgz",
<             "integrity": "sha512-oL0rGeM/WFQCUd0y2QrWxYnq7tfSuKBiqTjRPWrRgB46WD/kiwHwF8T23z78H6Q6kGCuuHcPB+KULHRdxvVGQA==",
---
>             "version": "5.14.1",
>             "resolved": "https://registry.npmjs.org/terser/-/terser-5.14.1.tgz",
>             "integrity": "sha512-+ahUAE+iheqBTDxXhTisdA8hgvbEG1hHOQ9xmNjeUJSoi6DU/gMrKNcfZjHkyY6Alnuyc+ikYJaxxfHkT3+WuQ==",
2741,2743c2741,2743
<             "version": "5.14.2",
<             "resolved": "https://registry.npmjs.org/terser/-/terser-5.14.2.tgz",
<             "integrity": "sha512-oL0rGeM/WFQCUd0y2QrWxYnq7tfSuKBiqTjRPWrRgB46WD/kiwHwF8T23z78H6Q6kGCuuHcPB+KULHRdxvVGQA==",
---
>             "version": "5.14.1",
>             "resolved": "https://registry.npmjs.org/terser/-/terser-5.14.1.tgz",
>             "integrity": "sha512-+ahUAE+iheqBTDxXhTisdA8hgvbEG1hHOQ9xmNjeUJSoi6DU/gMrKNcfZjHkyY6Alnuyc+ikYJaxxfHkT3+WuQ==",
```

taniwallach (changes requested):
As noted in the discussion, I think a small change to `package.json` may suffice instead of adding a full `package-lock.json` locking down the versions in such a strict manner. However, I am not qualified to really provide an opinion of the pros and cons of the two options. If the consensus is that the `package-lock.json` option is better, I will approve.

If a `package-lock.json` file is added, then the security update for `terser` should be considered as something to maybe use in the `package-lock.json` file being provided. A fresh `npm install` using only the modified `package.json` gets quite a few other package version differences. Which of those to update in the provided file can also be checked.

drgrice1:
@taniwallach:  Let me just cut your lengthy response short.  Simply put, this is the way it is actually supposed to be done.

pstaabp (approved):
Verified that 1) there is a compile error 2) this branch fixes it. 

Although I did test that @taniwallach works, I think the package-lock is the way to go (@drgrice1 just mentioned this), because there could be other packages that update themselves that we would need to hot fix.

drgrice1:
Also note that any security vulnerabilities for dev dependencies like terser are extremely minor.  That is only used to minify the files when the `generate-assets.js` script is run (that is done when you run `npm install`).  This particular vulnerability is not even one of real concern for our usage.

taniwallach (approved):
> @taniwallach: Sorry for the snappy response...
> 
> In any case, I changed the package.json file to only accept patch releases compatible with bootstrap 5.1 as you suggested. The package-lock.json file is still needed.

No need to apologize. I appreciate both the explanation and your expertise.

Let's get the revised hotfix merged ASAP.

drgrice1:
@taniwallach:  I forgot to mention that I also changed the `package-lock.json` file from what I originally committed to address the security vulnerability that you listed.

taniwallach:
> @taniwallach: I forgot to mention that I also changed the `package-lock.json` file from what I originally committed to address the security vulnerability that you listed.

Yes - I saw that, and that it also updated various other dependencies. Thanks.

PR #1762, opened 2022-07-19
Package lock is back develop

drgrice1:
#1761 for develop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ htdocs/themes/*/images/*
 !htdocs/themes/*/images/webwork_logo.svg
 !htdocs/themes/math4/images/webwork_square.svg
 htdocs/themes/*/*.css
-htdocs/package-lock.json
 htdocs/themes/*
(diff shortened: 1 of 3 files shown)


taniwallach:
In https://github.com/openwebwork/webwork2/pull/1761#issuecomment-1189549124 you provided a note on what change is needed to `bootstrap.scss` so that it will work with bootstrap 5.2.0. Would it not make more sense to make that change to develop, unless there is some expectation that the bootstrap update will break functionality in develop for now?

drgrice1:
I agree that the change should be made to develop to fix it for bootstrap 5.2.0.  This pull request will still be needed to keep the branches in sync, assuming the corresponding hot fix pull request is merged.

taniwallach (approved):
To merge when https://github.com/openwebwork/webwork2/pull/1761 gets merged.

drdrew42 (approved):

pstaabp (approved):

Issue #1524, opened 2021-11-17
MathJax disappearing overline and left/right abs bug effects some users on my WW 2.16 server


taniwallach:
MathJax 3.1.2 (used in WW 2.16) seems to have an issue with `\overline{}` disappearing from the screen for some browsers (apparently Chrome and Firefox) in certain circumstances.

---

Edit  (04-Apr-2022): Issues of disappearing `\left\abs` and `\right\abs` have also been encountered in both Chrome and Firefox.

Edit (04-Apr-2022): Using the SVG output mode of instead of `chtml` MathJax seems to avoid these issues.

---

For example some of students using my system have reported that a `\overline{}` was not being shown in problems with the following equations:
```
\( \displaystyle \frac{ z_{1}^{3} - (\overline{z_1})^3 \, }{z_2} = \)
```
and
```
\( \left| \, \left( z_1 \cdot z_2 - \overline{z_3} \, \right)^3 \, \right| = \)
```

The issue seems to be specific to certain browsers and zoom levels.

Sometimes the line appears/disappears after a page refresh or problem submission according to the reports. (I was sent a video of the disappearing `\overline` after a submit.)

I don't seem to be able to trigger the issue on my machine (Debian 11 with Chrome version 95.0.4638.69) except at 50% zoom on Chrome which is so small it cannot be read easily anyway. 

Firefox 78.15.0esr (Debian's packaged version) does not seem to hit it on my machine for any standard zoom level. 

---

Edit (04-Apr-2022): More recent versions of Firefox (94-98) certainly do encounter these issues.

---

The problem seems to be a "quirk" with clipping behavior, and @dvpc seems to be trying to have MathJax work around the issue. It is addressed in:
  - https://github.com/mathjax/MathJax/issues/2701   (refers to chtml output, as WeBWorK uses)
  - https://github.com/mathjax/MathJax-src/pull/726
The reports indicate that `\underline{}`is also effected.

There are quite a few OPL questions which use `\overline{}` so trying to avoid the issue probably justifies a hotfix for WW 2.16 once a reasonable solution is found.  https://github.com/openwebwork/webwork-open-problem-library/search?q=overline

Upgrading to MathJax 3.2.0 may fix the problem, but that is not certain and may be more of a change than we want in a hotifx. 

https://github.com/mathjax/MathJax/issues/2795
reports the problem still occurring for Firefox for some zoom levels and OS versions.

[...]

PR #1763, opened 2022-07-20
Hotfix: Set MathJax to SVG output

taniwallach:
Change to using MathJax's SVG output, as Firefox had a bug which caused some thin lines to disappear when it renders from MathJax's CHTML output at certain Zoom levels.

WW 2.17 no longer allows changing the MathJax mode by modifying a config file, so this hotfix changes the hard-wired default.

See:
  - https://github.com/openwebwork/webwork2/issues/1524
  - https://bugzilla.mozilla.org/show_bug.cgi?id=1741887

### DIFF
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -194,7 +194,7 @@ sub formatRenderedProblem {
 		[ 'node_modules/jquery-ui-dist/jquery-ui.min.js',                      0, {} ],
 		[ 'node_modules/iframe-resizer/js/iframeResizer.contentWindow.min.js', 0, {} ],
 		[ 'js/apps/MathJaxConfig/mathjax-config.js',                0, { defer => undef } ],
-		[ 'node_modules/mathjax/es5/tex-chtml.js',                  0, { defer => undef, id => 'MathJax-script' } ],
+		[ 'node_modules/mathjax/es5/tex-svg.js',                    0, { defer => undef, id => 'MathJax-script' } ],
 		[ 'node_modules/bootstrap/dist/js/bootstrap.bundle.min.js', 0, { defer => undef } ],
 		[ 'js/apps/Problem/problem.js',                             0, { defer => undef } ],
 		[ 'math4.js',                                               1, { defer => undef } ],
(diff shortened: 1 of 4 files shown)


lib/FormatRenderedProblem.pm
```diff
@@ -194,7 +194,7 @@ sub formatRenderedProblem {
 		[ 'node_modules/jquery-ui-dist/jquery-ui.min.js',                      0, {} ],
 		[ 'node_modules/iframe-resizer/js/iframeResizer.contentWindow.min.js', 0, {} ],
 		[ 'js/apps/MathJaxConfig/mathjax-config.js',                0, { defer => undef } ],
-		[ 'node_modules/mathjax/es5/tex-chtml.js',                  0, { defer => undef, id => 'MathJax-script' } ],
+		[ 'node_modules/mathjax/es5/tex-svg.js',                    0, { defer => undef, id => 'MathJax-script' } ],
```
drgrice1, line 197: There seems to be white space at the end of the line here.

drgrice1 (approved):
Looks good other than a minor quibble on white space.

pstaabp (approved):
Looks good.

lib/FormatRenderedProblem.pm
taniwallach: I do not see any space at the end of the line. It is space padding and the 2 additional spaces characters were added to align the `0` with the lines above/below. It is one of a sequence of 6 lines where such formatting was done, so I retained the style.

taniwallach:
Note: the display issue reported by @maharper in https://github.com/openwebwork/webwork2/discussions/1641#discussioncomment-2517758 when using Firefox to display in a problem in the editor's "in screen" viewer does occur. Far less damaging than the missing math for students.

lib/FormatRenderedProblem.pm
drgrice1: I see now.  It is the way that this wrapped in github that confused me.  Sorry.

drgrice1:
Yeah, that doesn't look very nice.  It is the usual problem with the in window pg editor viewer that is caused by the page not being displayed until the iframe load event occurs.

taniwallach:
Locally changing to CHTML or using images mode for the math display work should around that issue for any users who encounter it.

PR #1764, opened 2022-07-20
MathJax to SVG output (as in PR#1763) - to develop

taniwallach:
These are the same changes as in https://github.com/openwebwork/webwork2/pull/1763/ but targeted to develop.
Hope this is the best manner to handle the change of the hotfix getting into develop also.


drgrice1:
Probably we need to discuss a different approach to releases other than using branches.  I think that it would be best to change to using github releases instead.  We don't need an actual git clone anymore.  That was primarily used for the upgrade notifier.  Since that doesn't even work anymore due to git's security changes, that isn't relevant.

drgrice1 (approved):

pstaabp (approved):

PR #1765, opened 2022-07-20
Move the "busy indicator" of the pg problem editor into the modal and show the modal immediately

drgrice1:
Currently the "busy indicator" is shown, and then the modal is shown once the iframe content has loaded.

To make this look better, the "busy indicator" is now contained in the modal body.

This solves several issues with content not being rendered correctly due to the page not having size yet (for example MathJax in SVG mode).

I put this in to develop since it is just a display issue, but if it is desired as a hotfix, that could also be done.

### DIFF
--- a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
@@ -23,7 +23,6 @@
 		frame.contentWindow.addEventListener('resize',
 			() => frame.contentDocument.getElementById('content').classList.remove('col-md-10')
 		);
-		bsModal.show();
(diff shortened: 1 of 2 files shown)


taniwallach (approved):
This fixes the bug which was occuring in Firefox.
I think since WeBWorK 2.17 is so new, and we have a simple fix for this issue - it should also be a hot-fix.

somiaj:
Due to the change to svg for MathJax output, I had to apply this patch to my 2.17 server, because without it the math would not render correctly in the editor unless I opened the problem in a new tab. To me this display issue made the modal view unusable. I think this should probably be a hot fix to go with the svg change.

PR #1767, opened 2022-07-22
Hotfix (change from PR #1765): Move the "busy indicator" of the pg problem editor into the modal and show the modal immediately

taniwallach:
Cherry-pick the commit from https://github.com/openwebwork/webwork2/pull/1765 as a hotfix to main.

---

Copied from https://github.com/openwebwork/webwork2/pull/1765 :

Move the "busy indicator" of the pg problem editor into the modal and show the modal immediately, rather than showing the "busy indicator", and then showing the modal after the page loads.

This solves several issues with content not being rendered correctly due to the page not having size yet (for example MathJax in SVG mode).

### DIFF
--- a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
@@ -23,7 +23,6 @@
 		frame.contentWindow.addEventListener('resize',
 			() => frame.contentDocument.getElementById('content').classList.remove('col-md-10')
 		);
-		bsModal.show();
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
So a hotfix is desired then?

taniwallach:
> So a hotfix is desired then?

I think so - it fixes broken functionality (for Firefox users) in the new release, so there seems little need to leave it broken.
@somiaj also agreed in the discussion thread about the patch to develop.

pstaabp (approved):

PR #1770, opened 2022-08-02
On the grades page only hide scores for gateway sets.

drgrice1:
The code that checks to see if scores for a gateway quiz should be hidden on the grades page was not checking to see if the set was actually a gateway quiz.  However, it is possible to make a homework set have hide_score set to 'Y'.  To make this happen create a homework set, change it to a gateway/quiz, and then set it to not show scores on finished assignments and save.  Then change the assignment type back to homework and save (only once).  Note that if you click save one more time it will reset the hide_score value in the database back to NULL.

If a student is assigned one of these homework sets and tries to view the grades page it causes an error.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8092.


pstaabp:
Does this branch fix the problem you mentioned above that the `hide_score` value shouldn't go to NULL, or does this fix the error on the student side?  That is if it is null it won't throw an error.

I was able to replicate the way to get NULL for `hide_score` but this doesn't seem to change that.

drgrice1:
Actually, that is not a problem.  Homework sets do not use the `hide_score` setting at all, and really it should be NULL.  The technical problem here is that when a gateway set is changed to a homework set the setting is not set to NULL at that time.  Although, it really isn't a problem, because it homework sets just don't use that setting al all.

drgrice1:
This just fixes the error when students view the grades and the setting is not NULL.

pstaabp (approved):
I now see the error and this PR fixed it.

drdrew42 (approved):
👍

Issue #1774, opened 2022-08-05
Issues with essay questions in 2.17


dlglin:
I'm new to essay questions in WW, so it's possible that some of this is just me not understanding how things work.

- In 2.16 if the entry assist method is set to MathView then essay boxes have a tool beside them to insert math expressions using the MathView editor.  I don't see this in 2.17.  Can the same sort of tool be created for MathQuill?
- In 2.16 if a question has essay problems, then when the instructor views the Problems page there is a "Grade Problem" link to take you to the manual grader.  This link isn't showing up for me in 2.17.  It is still accessible through the other links (on the Set Detail or Statistics page).

drgrice1:
I will look into the MathView editor thing.  I don't remember that being there for essay questions for 2.16 though.  It may be possible to do something like that with MathQuill.

It seems that the essay flag is not being set for newly created essay problems.  I will look into it.

drgrice1:
Wait, the flag is getting set.  It just isn't set until after a student answers a question.  So the "Grade Problem" link isn't showing up for some other reason.

I noticed that problems in existing sets created with 2.16 do get the "Grade Problem" link though.

drgrice1:
#1775 fixes the grade problem link on the problem set page.

Now to see about the MathView thing.

drgrice1:
The MathView problem is actually a pg issue! 

Unfortunately, adding a MathQuill editor for essay problems is something to look into for the future, but certainly can't be hotfix to 2.17.

Alex-Jordan:
This seems to all be addressed, at least as of 2.19 if not earlier.

PR #1775, opened 2022-08-05
Fix the "Grade Problem" link for essay problems on the problem set page.

drgrice1:
This fixes part of issue #1774.

The issue is that the merged problem does not hold the essay flag.  The global problem does.  So those have to be fetched from the database for this.


dlglin (approved):
Works as expected for me.

drdrew42 (approved):
👍

PR #1780, opened 2022-08-11
provide errors when called scripts fail

drdrew42:
We've seen several occurrences now ([1](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8094), [2](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8104)) where the new OPL-update runs other scripts that may fail -- but any errors are not displayed, giving the impression that the script successfully ran as expected.

This PR modifies the handling of sub-scripts to capture any error messages they produce.

### DIFF
--- a/bin/Helper.pm
+++ b/bin/Helper.pm
@@ -1,0 +1,17 @@
+package Helper;
+
+use warnings;
+use strict;
+use base 'Exporter';
+
+our @EXPORT_OK = 'runScript';
+
+sub runScript {
+    my $script_path = shift;
+    unless ( do $script_path ) {
+        warn "Execution of $script_path failed:\n";
+        die $@ if $@;
+    }
+}
+
+1;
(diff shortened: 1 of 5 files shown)


drgrice1 (approved):
Looks good.

drgrice1:
Oops.  I missed something.  There is a warning when the download-OPL-metadata-release.pl script runs: "Subroutine runScript redefined at /opt/webwork/webwork2/bin/download-OPL-metadata-release.pl line 98."  It seems that the do statement runs in the scope of the calling script.

You may need to put the runScript method into a separate .pm file in the bin directory.  OPL-update already calls `use lib "$ENV{WEBWORK_ROOT}/bin";`.  That would then need to be added in download-OPL-metadata-release.pl as well.

drgrice1:
I mean the warning for the download-OPL-metadata-release.pl script is when the OPL-update script runs it.

drgrice1:
@drdrew42:  Can you add File::Fetch to check_modules.pl in this pull request.  If not, I will put in a separate pull request, but I want to reduce the number of hotfixes.

This is included in the the Ubuntu installation by default, but not on all distributions (e.g. CentOS 7).
    
See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8104#p19604

drdrew42:
Okay, I've shifted the script wrapper function to a module called `Helper` and added `File::Fetch` to the check_modules.pl script.

drgrice1:
Looks good.

drgrice1:
This needs to go in.  So I am merging.

Issue #1786, opened 2022-08-16
Shifted headers in /instructor/sets/


robert-marik:
Hello. I can see a shift in the /instructor/sets/ table. See the picture. There should be empty th above checkboxes and all other th tags should move right. Can anybody check and confirm this issue? Thank you.

![Screenshot from 2022-08-16 11-28-10](https://user-images.githubusercontent.com/2342862/184847593-e72ecccd-e00c-47b2-ab4a-75f6d96baab5.png)

taniwallach:
I don't see any such shift on WeBWork 2.17 either with the standard `math4` theme or with `math4-green`. Everything seems to display properly for me.

![set_editor](https://user-images.githubusercontent.com/39089094/184853456-edc953d5-6c23-4182-bd80-2f42bde7112e.png)

When I "inspect" I see another `<th>` before the one which has the "Edit Set Data" in it. The `<th>` you are apparently missing has the checkbox in it.

Are you running the current version of the code?

taniwallach:
Yes - that does trigger the display bug. Sorry, I should have noticed in your image that you were in the second stage of the export process.
@drgrice1 seems to have made a patch already while I was trying to understand what was going wrong.

drgrice1:
This is fixed by #1787.

PR #1787, opened 2022-08-16
Add the select all checkbox to the Hmwk Sets Editor export sets page.

drgrice1:
See issue #1786.

Note that this is different than WeBWorK 2.16 and before.  Prior versions of WeBWorK had an empty table heading above the select set checkbox.  This adds the select all checkbox instead.

I can't exactly figure out how the empty header was added by looking at the WeBWorK 2.16 code.  So I took the approach of adding the select all checkbox instead, since that was clearly easy to implement.  In my opinion this is better anyway.

This is not a hotfix pull request at this point.  If that is desired then please say so.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
@@ -2723,7 +2723,7 @@ sub printTableHTML {
 
 	my @tableHeadings = map { $fieldHeaders{$_} } @realFieldNames;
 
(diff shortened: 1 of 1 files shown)


taniwallach (approved):
I checked that this fixes the bug reported in https://github.com/openwebwork/webwork2/issues/1786 .

Given the many other pending changes to the current release, I think this can also be a hotfix, as the display bug might confuse some people.

taniwallach:
BTW I agree that the select / unselect all checkbox is certainly more useful than the prior empty header was.

PR #1797, opened 2022-09-02
Also filter Contrib when viewing local problems in the Library Browser

dlglin:
When viewing local problems in the library browser, it is hard-coded to skip any subfolders of Library, but by default all subfolders of Contrib are shown.

This hard-codes skipping of Contrib as well.  I don't know enough about the perl grep syntax to know if this could be done in a single command.

It would also be an improvement to have the list of folders to skip be a configuration variable rather than hard-code them.

I have targeted this to main, but if it's deemed not to be hotfix-worthy I can re-target to develop.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
@@ -120,8 +120,7 @@ sub get_library_sets {
 	my @dirs = grep {!$ignoredir{$_} and -d "$dir/$_"} @lis;
 	if ($top == 1) {@dirs = grep {!$problib{$_}} @dirs}
 	# Never include Library or Contrib at the top level
-	if ($top == 1) {@dirs = grep {$_ ne 'Library'} @dirs}
-        if ($top == 1) {@dirs = grep {$_ ne 'Contrib'} @dirs}
+	if ($top == 1) {@Dirs = grep {$_ ne 'Library' && $_ ne 'Contrib'} @Dirs}
 	foreach my $subdir (@dirs) {
 		my @results = get_library_sets(0, "$dir/$subdir");
 		$pgcount += shift @results; push(@pgdirs,@results);
--- a/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
@@ -120,7 +120,7 @@ sub get_library_sets {
 	my @dirs = grep {!$ignoredir{$_} and -d "$dir/$_"} @lis;
 	if ($top == 1) {@dirs = grep {!$problib{$_}} @dirs}
 	# Never include Library or Contrib at the top level
-	if ($top == 1) {@Dirs = grep {$_ ne 'Library' && $_ ne 'Contrib'} @Dirs}
+	if ($top == 1) {@dirs = grep {$_ ne 'Library' && $_ ne 'Contrib'} @Dirs}
 	foreach my $subdir (@dirs) {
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
```diff
@@ -119,8 +119,9 @@ sub get_library_sets {
 	my @pgdirs;
 	my @dirs = grep {!$ignoredir{$_} and -d "$dir/$_"} @lis;
 	if ($top == 1) {@dirs = grep {!$problib{$_}} @dirs}
-	# Never include Library at the top level
+	# Never include Library or Contrib at the top level
 	if ($top == 1) {@dirs = grep {$_ ne 'Library'} @dirs}
+        if ($top == 1) {@dirs = grep {$_ ne 'Contrib'} @dirs}
```
drdrew42, line 124:
```suggestion
	# Never include Library or Contrib at the top level
	if ($top == 1) {@dirs = grep {$_ ne 'Library' && $_ ne 'Contrib'} @dirs}
```

can be done in a single line 🤓

lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
```diff
@@ -119,8 +119,8 @@ sub get_library_sets {
 	my @pgdirs;
 	my @dirs = grep {!$ignoredir{$_} and -d "$dir/$_"} @lis;
 	if ($top == 1) {@dirs = grep {!$problib{$_}} @dirs}
-	# Never include Library at the top level
-	if ($top == 1) {@dirs = grep {$_ ne 'Library'} @dirs}
+	# Never include Library or Contrib at the top level
+	if ($top == 1) {@Dirs = grep {$_ ne 'Library' && $_ ne 'Contrib'} @Dirs}
```
drgrice1, line 123: The case of `@Dirs` somehow was switched.  This causes an undefined variable error.

dlglin: Weird.  All I did was accept the changes proposed by @drdrew42.  It's fixed now.

drgrice1: Yeah, the mistake with the case was in @drdrew42's changes.

drdrew42: weird, no idea how that happened... I blame autocorrect

drgrice1 (approved):
This looks good.  Although, how exactly do you make a Contrib directory show up in the library browser?

I don't mind if this is a hotfix, although it doesn't seem very significant of an issue.

drdrew42 (approved):
👍

dlglin:
> This looks good. Although, how exactly do you make a Contrib directory show up in the library browser?
> 

https://github.com/openwebwork/webwork2/blob/0c150af2f57844fb56d4f7c5a1663cfbc18dd01d/conf/defaults.config#L525-L531
If Contrib is linked in a given course there should be a button to browse the Contrib directory.

> I don't mind if this is a hotfix, although it doesn't seem very significant of an issue.

I already patched my prod server, so I don't feel strongly about this being a hotfix.

drgrice1:
I know how to make the Contrib button appear.  That wasn't what I meant.  What I meant is really, what do you need to do to see the effect of this pull request.  What is the undesirable behavior before, and what is the corrected behavior after.

dlglin:
Gotcha.  Before, go to the Library Browser and click on Local Problems.  Before the patch, you will see all of the Contrib folders in the dropdown.  After the patch they will be filtered out.

PR #1798, opened 2022-09-05
Code mirror RTL hotfix (#1796 for develop)

drgrice1:


taniwallach:
@drgrice1 Thanks for setting up the PR for develop.

PR #1800, opened 2022-09-09
change GW quizzes to default to 1 question per page

Alex-Jordan:
This was discussed in a recent meeting.


drgrice1 (approved):
I approve.

Wait, wasn't the vote 50/50 on this? :wink:

drdrew42 (approved):
I think you're thinking of the vote on whether or not to hold a vote on changing the default to 1 per page. 😝

dlglin (approved):
If we don't have a supermajority can I filibuster this change?

somiaj (approved):
If the vote was 50/50, I'll tip it to yes!!

pstaabp (approved):
I want to vote too!!

PR #1801, opened 2022-09-12
put sessionKeyTimeout in localOverrides.conf.dist

Alex-Jordan:
This seems useful as a site-wide setting to override. It is present in `simple.conf`, but faculty do not always understand its importance. Especially when they give timed quizzes. Changing it for the institution is helpful, for example because you might decide to set it to the standard duration of a class at your institution.


drgrice1 (approved):
Looks fine.

PR #1802, opened 2022-09-12
Fix single character directories in templates for Local Problems in the Library Browser

drgrice1:
See https://github.com/openwebwork/pg/issues/725 for details.


drdrew42 (approved):

Alex-Jordan (approved):
It looks like this bug goes back 18+ years!

drgrice1:
It seems it was added on May 7, 2004 according to git blame.  So it is an old one.

PR #1808, opened 2022-09-29
Missing merge file locks Email page

drdrew42:
### To replicate
1. Set a merge file on the email page and *perform any submit action* (preview message will suffice)
2. Go into File Manager and change the name of the merge file (or remove it)
3. Attempt to return to email page -- WeBWorK 'die's because merge file is not found
4. Email page is unrecoverable until merge file is restored

### Background
For persistence on the Email page, WeBWorK stores the last-used merge file in the database. If the file is (re)moved, the DB has no knowledge of the change. As a result, the Email page attempts to restore the last-used settings and dies un-gracefully, leaving the user with no way to recover -- except by restoring a file with the original name.

The fix, as implemented here, is checking for the existence of the merge file when the filename is retrieved from the database. I'm open to feedback if anyone has of a better way of handling this.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
@@ -85,6 +85,7 @@ sub initialize {
 	  $db->setSettingValue("${user}_mergefile",$mergefile);
 	} elsif ($db->settingExists("${user}_mergefile")) {
 	  $mergefile = $db->getSettingValue("${user}_mergefile");
+	  $mergefile = undef unless (-e "$ce->{courseDirs}{scoring}/$mergefile");
 	}
 
 	# Figure out action from submit data


drgrice1 (approved):
This looks fine to me.  I don't know that there is a better way to do it other than reworking how merge files are dealt with entirely.

pstaabp:
Haven't tested, but code looks good.  

However, what if the file gets deleted?  It seems that the only way to recover is to delete the row in the database with the file name. 

A way to prevent a user from messing with the file is to store the file in a location where the user doesn't have access (perhaps in the tmp directory).

drdrew42:
If the file is deleted, then the only recovery available to the user is to restore a file (with any contents, even empty) with the same name. It only needs to pass a check that the file exists.

The 'scoring/' directory is definitely the "right place" for these merge files to come from -- and limiting the user's ability to modify the contents there is not appealing. My change here just ensures that we fail more gracefully, since it isn't catastrophic if the merge file goes missing.

PR #1809, opened 2022-09-30
Quick hack to allow the gateway quiz page to be viewed with restricted IP settings.

drgrice1:
This allows the new gateway quiz summary page to be viewed even if IP restrictions don't allow access to the gateway quiz itself.  The gateway quiz module prevents the quiz itself to be entered still.  This is a somewhat ugly hack though.  It checks the return value of checkSet to see if it matches /^Client ip address .* is not in the list of addresses/ and allows gateway quiz sets through if it is.

This also fixes a call in Hardcopy.pm to the now non-existent Apache2::Connection remote_ip method.  That needs to be the Apache2::Request useragent_ip method now.  Apparently this has been broken since the Apache 2.4 methods were added.  In Authz.pm this change was made, but it must have been missed here.

This fixes the issue in https://webwork.maa.org/moodle/mod/forum/post.php?reply=19793.  Should this be a hotfix?


pstaabp:
What I read from the forum post is that students wish to download their quiz results (hardcopy) from a computer outside the restricted ip range and are unable to do so. 

I'm trying to figure out how to test this on my develop machine.  Any ideas? 

Since this seems to be used in the wild, I would say a hot fix would be good.

drgrice1:
To test this you can create and assign a gateway quiz, and then take it with a student user.  Then modify the quiz so that it has a restricted ip address that is not that of your local machine.  After that the student should still be able to open the quiz page, but should not be able to start a new quiz (or continue an open quiz).  Now change the answer date so that today is after that answer date.  Then the student should still be able to open the quiz pageand should be able to print completed quiz versions.

pstaabp:
I opened a GW quiz with no restrictions.  Assigned it to a student. the student took the quiz, then set the IP restrictions, changed the answer date so current date is after this and still seeing the warning
```
Use of uninitialized value in numeric eq (==) at /opt/webwork/webwork2/lib/WeBWorK/Authz.pm line 550.
```
It doesn't look like `$rIP->overlaps($clientIP)` has a value on line 550. 

Also on my development machine, seems to be using IPv6--not sure if that's a problem.  Here's a dump of the `$clientIP` variable:
```
bless( {
'is_prefix' => 1,
'last_ip' => '0000:0000:0000:0000:0000:0000:0000:0001',
'ipversion' => '6',
'last_bin' => '00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001',
'ip' => '0000:0000:0000:0000:0000:0000:0000:0001',
'binmask' => '11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111',
'binip' => '00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001',
'prefixlen' => '128'
}, 'Net::IP' );
```

drgrice1:
@pstaabp:  Yeah, I was going to point that out for testing.  This doesn't work on `localhost`.  I used a dns so that I would have a proper IP to test this.

drgrice1:
@pstaabp:  Note that once webwork2 is switched to Mojolicious, the issue with the IP for localhost will be resolved.  modperl2's `useragent_ip` method does not provide a IPV4 address for localhost, but Mojolicious via `$c->tx->remote_address` does.

pstaabp:
I have the impression that with IP restrictions on that a student should be able to view the overall quiz page (with versions) at url `webwork2/test/quiz_name`.  Correct?  I'm still seeing "Client IP address XX.XX.XX.XX is not allowed to work..."  This looks exactly the same on develop and this branch.   BTW, running this on a ubuntu machine setup with nginx, so to double check that the server was resetting, added a `warn` near the new code which popped up.

somiaj:
This works as described in my test. With this patch I can access the quiz summary page with the message my ip address is not in the list the assignment can be reached.

If I have a test version that is completed and allows access, there are links to the quiz version, but when used the test module doesn't allow access due to the ip restriction. Unsure if this is the desired result or not, but if it is, should the links on the quiz summary page to version be removed?

drgrice1:
Oops.  Clicked wrong button.

pstaabp (approved):
I finally was able to test this and all looks good.

PR #1810, opened 2022-10-04
Show set header in proctored login.

somiaj:
Use the set header for the page info-box when at a login prompt for a proctored gateway quiz instead of showing the site info or login info.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/LoginProctor.pm
+++ b/lib/WeBWorK/ContentGenerator/LoginProctor.pm
@@ -25,9 +25,10 @@ GatewayQuiz proctored tests.
 
 use strict;
 use warnings;
-use CGI qw(-nosticky );
-use WeBWorK::Utils qw(readFile dequote);
-use WeBWorK::DB::Utils qw(grok_vsetID);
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
The code looks good (although could perhaps be perltidied!).

I think this makes more sense than showing the site info and login info on this page.  However, since this was implemented this way, it may be that there are those that think otherwise.

drgrice1:
Actually, thinking about it more, I forgot that set headers were not used for gateway quizzes before.  So that is probably why the site info was shown.  Now that the set header is used for gateway quizzes this makes more sense.

somiaj:
Removed an unused variable in my code, and ran perltidy on the whole file.

pstaabp (approved):

drgrice1:
@somiaj:  Note that this pull request will need some changes now that the #1792 has been merged.  See line 242 of lib/WeBWorK/ContentGenerator/ProblemSet.pm for the new correct way to call `WeBWorK::PG->new`.

drdrew42 (approved):

PR #1811, opened 2022-10-08
Add option to skip grade proctor auth.

somiaj:
Adds the option `$skip_grade_proctor_auth` to allow skipping authentication when grading a proctored gateway quiz. This can configured to never skip grade proctor auth (default), only skip grade proctor auth for quizzes with a set password to start, or always skip the grade proctor auth.

This is done by converting the versioned proctored_gateway assignment to a normal gateway assignment when a new version is created after authentication has succeeded to start the proctored gateway quiz.

Such a feature was mentioned in the forms here, <https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4835>.

It might be preferred to make this a set configuration option instead of a course configuration option. I initially went with a course configuration option as this doesn't require a possible change to the database, and shows one way such an option can be implemented. I also think most use cases can be controlled via this course configuration option vs modifying the database. This can be changed to be a set configuration option if desired.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -317,7 +317,7 @@ use constant FIELD_PROPERTIES => {
 		},
 	},
 	use_grade_proctor => {
-		name     => x('Require Grade Proctor'),
+		name     => x('Require Grade Authorization'),
 		type     => 'choose',
 		override => 'any',
 		choices  => [qw(Yes No)],
@@ -327,8 +327,9 @@ use constant FIELD_PROPERTIES => {
 		},
 		default   => 'Yes',
 		help_text => x(
-			'Proctored tests always require login proctor authorization to start the test. Configure if '
-				. 'proctored tests also require grade proctor authorization to grade the test at the end.'
+			'Proctored tests always require authorization to start the test. Set to "Yes" to require login '
+				. 'proctor authorization to start and grade proctor authorization to grade. Set to "No" to '
+				. 'require grade proctor authorization to start and no authorization to grade.'
 		),
 	},
 	restrict_prob_progression => {
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -18,7 +18,8 @@ use base qw(WeBWorK::ContentGenerator::Instructor);
 
 =head1 NAME
 
-WeBWorK::ContentGenerator::Instructor::ProblemSetDetail - Edit general set and specific user/set information as well as problem information
+WeBWorK::ContentGenerator::Instructor::ProblemSetDetail - Edit general set and specific user/set information as well
+as problem information
 
 =cut
 
@@ -37,7 +38,11 @@ use WeBWorK::Debug;
 
 # these constants determine which fields belong to what type of record
 use constant SET_FIELDS => [
-	qw(set_header hardcopy_header open_date reduced_scoring_date due_date answer_date visible description enable_reduced_scoring  restricted_release restricted_status restrict_ip relax_restrict_ip assignment_type use_grade_proctor attempts_per_version version_time_limit time_limit_cap versions_per_interval time_interval problem_randorder problems_per_page hide_score:hide_score_by_problem hide_work hide_hint restrict_prob_progression email_instructor)
+	qw(set_header hardcopy_header open_date reduced_scoring_date due_date answer_date visible description
+		enable_reduced_scoring  restricted_release restricted_status restrict_ip relax_restrict_ip
+		assignment_type use_grade_proctor attempts_per_version version_time_limit time_limit_cap
+		versions_per_interval time_interval problem_randorder problems_per_page
+		hide_score:hide_score_by_problem hide_work hide_hint restrict_prob_progression email_instructor)
 ];
 use constant PROBLEM_FIELDS =>
 	[qw(source_file value max_attempts showMeAnother showHintsAfter prPeriod att_to_open_children counts_parent_grade)];
@@ -46,14 +51,16 @@ use constant USER_PROBLEM_FIELDS => [qw(problem_seed status num_correct num_inco
 # these constants determine what order those fields should be displayed in
 use constant HEADER_ORDER => [qw(set_header hardcopy_header)];
 use constant PROBLEM_FIELD_ORDER => [
-	qw(problem_seed status value max_attempts showMeAnother showHintsAfter prPeriod attempted last_answer num_correct num_incorrect)
+	qw(problem_seed status value max_attempts showMeAnother showHintsAfter prPeriod attempted last_answer num_correct
+		num_incorrect)
 ];
 # for gateway sets, we don't want to allow users to change max_attempts on a per
 #    problem basis, as that's nothing but confusing.
 use constant GATEWAY_PROBLEM_FIELD_ORDER =>
 	[qw(problem_seed status value attempted last_answer num_correct num_incorrect)];
 use constant JITAR_PROBLEM_FIELD_ORDER => [
-	qw(problem_seed status value max_attempts showMeAnother showHintsAfter prPeriod att_to_open_children counts_parent_grade attempted last_answer num_correct num_incorrect)
+	qw(problem_seed status value max_attempts showMeAnother showHintsAfter prPeriod att_to_open_children
+		counts_parent_grade attempted last_answer num_correct num_incorrect)
 ];
 
 # we exclude the gateway set fields from the set field order, because they
@@ -64,11 +71,12 @@ use constant JITAR_PROBLEM_FIELD_ORDER => [
 # FIXME: set for non-gateway assignments.  right now (11/30/06) they are
 # FIXME: only used for gateways
 use constant SET_FIELD_ORDER => [
-	qw(open_date reduced_scoring_date due_date answer_date visible enable_reduced_scoring restricted_release restricted_status restrict_ip relax_restrict_ip hide_hint assignment_type)
+	qw(open_date reduced_scoring_date due_date answer_date visible enable_reduced_scoring restricted_release
+		restricted_status restrict_ip relax_restrict_ip hide_hint assignment_type)
 ];
-# use constant GATEWAY_SET_FIELD_ORDER => [qw(attempts_per_version version_time_limit time_interval versions_per_interval problem_randorder problems_per_page hide_score hide_work)];
 use constant GATEWAY_SET_FIELD_ORDER => [
-	qw(version_time_limit time_limit_cap attempts_per_version time_interval versions_per_interval problem_randorder problems_per_page hide_score:hide_score_by_problem hide_work)
+	qw(version_time_limit time_limit_cap attempts_per_version time_interval versions_per_interval problem_randorder
+		problems_per_page hide_score:hide_score_by_problem hide_work)
 ];
 use constant JITAR_SET_FIELD_ORDER => [qw(restrict_prob_progression email_instructor)];
 
@@ -78,17 +86,17 @@ use constant JITAR_SET_FIELD_ORDER => [qw(restrict_prob_progression email_instru
 #
 #	All but name are optional
 #	some_field => {
-#		name      => "Some Field",
-#		type      => "edit",		# edit, choose, hidden, view - defines how the data is displayed
-#		size      => "50",		# size of the edit box (if any)
-#		override  => "none",		# none, one, any, all - defines for whom this data can/must be overidden
-#		module    => "problem_list",	# WeBWorK module
+#		name      => 'Some Field',
+#		type      => 'edit',		# edit, choose, hidden, view - defines how the data is displayed
+#		size      => '50',		# size of the edit box (if any)
+#		override  => 'none',		# none, one, any, all - defines for whom this data can/must be overidden
+#		module    => 'problem_list',	# WeBWorK module
 #		default   => 0			# if a field cannot default to undefined/empty what should it default to
 #		labels    => {			# special values can be hashed to display labels
-#				1 => "Yes",
-#				0 => "No",
+#			1 => x('Yes'),
+#			0 => x('No'),
 #		},
-#               convertby => 60,                # divide incoming database field values by this, and multiply when saving
+#		convertby => 60,		# divide incoming database field values by this, and multiply when saving
 
 use constant BLANKPROBLEM => 'blankProblem.pg';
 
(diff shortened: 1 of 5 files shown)


dlglin:
On first glance I think that this would be better as a set-level configuration option.  Having it in the course/server config seems like it would be a little bit opaque to the instructor.  They could create the same quiz in two different courses and have it behave differently based on a setting whose source is not apparent from the set detail page.

Maybe a compromise is to add some text on the set detail page telling the instructor what will happen based on the current settings.  e.g. "this quiz will/will not require a proctor login for students to submit their completed quiz".  On further thought, that text may be good regardless.  I've seen several instructors try to set a set-level password for a quiz, not realizing that this forces a proctor login for students to submit.

somiaj:
Should this be something that can be overridden in for a user's set? This way you could make exceptions for certain students/makeups? I'll look into making this a set-level configuration option. Outside of that, does this approach seem reasonable?

drgrice1:
It probably should be something that can be overridden for a user's set.  Of course that is something that you couldn't do with the course level setting.

somiaj:
Updated to be a set level configuration option.

@drgrice1 I added a column to the Set and UserSet tables and by my tests this seems to work and can be overridden on a per user basis. Upgrading the database also worked. I am still not quite sure on everything that needs to be done to update the database, so checking if I missed anything.

drgrice1:
It should work with the usual upgrade process.  No further changes should be needed.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -281,6 +281,15 @@ use constant FIELD_PROPERTIES => {
...
 		labels    => { 'N' => "Yes", 'Y' =>"No", 'BeforeAnswerDate' =>x('Only after set answer date') },
 	},
+	use_grade_proctor => {
+		name      => x('Require Grade Proctor'),
+		type      => 'choose',
+		override  => 'any',
+		choices   => [qw(Yes No)],
+		labels    => { Yes => 'Yes', No => 'No' },
+		default   => 'Yes',
+		help_text => x('Proctored tests require login proctor authorization to start the test. Configure if proctored tests also require grade proctor authorization to grade the test.'),
```
pstaabp, line 291: This seems a bit confusing.  Perhaps: Proctored tests require login proctor authorization to start the test. For grading, set to 'Yes' to require proctor authorization and 'No' to skip proctor authorization.

drgrice1: @pstaabp:  However, the statement still needs to distinguish between a "login proctor" and "grade proctor".  Those different permission levels in webwork.  The "login proctor" starts the test, the "grade proctor" ends it.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -36,7 +36,7 @@ use WeBWorK::Debug;
 # 	but they are functionally and semantically different
 
 # these constants determine which fields belong to what type of record
-use constant SET_FIELDS => [qw(set_header hardcopy_header open_date reduced_scoring_date due_date answer_date visible description enable_reduced_scoring  restricted_release restricted_status restrict_ip relax_restrict_ip assignment_type attempts_per_version version_time_limit time_limit_cap versions_per_interval time_interval problem_randorder problems_per_page hide_score:hide_score_by_problem hide_work hide_hint restrict_prob_progression email_instructor)];
+use constant SET_FIELDS => [qw(set_header hardcopy_header open_date reduced_scoring_date due_date answer_date visible description enable_reduced_scoring restricted_release restricted_status restrict_ip relax_restrict_ip assignment_type use_grade_proctor attempts_per_version version_time_limit time_limit_cap versions_per_interval time_interval problem_randorder problems_per_page hide_score:hide_score_by_problem hide_work hide_hint restrict_prob_progression email_instructor)];
```
drgrice1, line 39: Please wrap this line.  It is way to long.  Fix the mess that others made!

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
drgrice1, line 291: Please wrap this line as well.  120 chars max including white space.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -993,10 +1004,7 @@ sub proctoredFieldHTML {
 		CGI::a(
 			{
 				class           => 'help-popup',
-				data_bs_content => $r->maketext(
-					"Proctored tests require proctor authorization to start and to grade.  "
-						. "Provide a password to have a single password for all students to start a proctored test."
-				),
+				data_bs_content => $r->maketext('Provide a password to have a single password for all students to start a proctored test.'),
```
drgrice1, line 1007: Another line too long.

drgrice1 (approved):
This looks good and is a good change.  There is some minor code clean up needed.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
drgrice1: I think better wording is "Proctored tests always require login proctor authorization to start the test. Configure if proctored tests also require grade proctor authorization to grade the test at the end."  There is no need to include anything about what "yes" or "no" means here.  That is clear given that the input is labeled with "Require Grade Proctor".

drgrice1: By the way, @pstaabp's suggested change brings up another issue that is prevalent in webwork2 code.  Never use single quotes in displayed text as a substitute for double quotes.  That is grammatically incorrect.  Single quotes and double quotes in code are a different story.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
somiaj: @drgrice1 I cleaned up the constants code block. See the second commit and let me know if I missed anything else. I got as much as I could with perltidy.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
somiaj: @pstaabp I went with @drgrice1 suggestion, does that clear it up for you?

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -994,8 +1068,8 @@ sub proctoredFieldHTML {
 			{
 				class           => 'help-popup',
 				data_bs_content => $r->maketext(
-					"Proctored tests require proctor authorization to start and to grade.  "
-						. "Provide a password to have a single password for all students to start a proctored test."
+					'Provide a password to have a single password for all students to start a '
+						. 'proctored test.'
```
drgrice1, line 1072: This is another case where you split a bit too much.  When I ran perltidy on your original change here it gave
```perl
				data_bs_content => $r->maketext(
					'Provide a password to have a single password for all students to start a proctored test.'),
```
for this.  That fits into the 120 character limit nicely.

drgrice1 (approved):
Looks cleaner now.  I added a few notes for consideration in the future, but no need to change now.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -67,239 +76,281 @@ use constant JITAR_SET_FIELD_ORDER => [qw(restrict_prob_progression email_instru
...
-		choices   => [qw(0 1)],
-		labels    => { '0' => 'No', '1' =>'Yes' },
+		name     => x('Cap Test Time at Set Close Date?'),
+		type     => 'choose',
+		override => 'all',
+		choices  => [qw(0 1)],
+		labels   => {
+			'0' => x('No'),
+			'1' => x('Yes')
+		},
```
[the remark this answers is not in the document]
somiaj: Thanks, wasn't sure here, since a few of the places that were formatted nicely already used this, I just went with it.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
somiaj: Totally missed wrapping the whole string, not just the end of it. Thanks.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
pstaabp: I'm good with @drgrice1 's  wording change.

pstaabp (approved):

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
somiaj: @drgrice1, I updated this since I had to clean things rebase against develop. I also figured out my original confusion, vim is defaulting to 8 spaces per tab, while perltidy defaults to 4, hence why I thought this went over 120 characters, but with 4 spaces per tab, it does fit nicely in 120 characters.

drgrice1: I see.  You can change that in `~/.vimrc`.  I have `set tabstop=4` in my `~/.vimrc` file.

So I take it that you are not actually using perltidy then.  Perltidy will format it assuming a tabstop of 4 regardless of what your text editor uses.  In vim I recommend coc.nvim with the coc-perl extension.  Then you can visually select a block of code and type `\f` to run perltidy on the block, or run `:Format` to perltidy the whole file.

somiaj: Yea, I just have defaulted to 8 space tabs for other projects, so it was mostly just habit and not realizing perltidy wanted a different amount. I have been running perltidy on the code after I'm done (when I remember, but getting better), which is why I noticed it was wrapping the code differently and took me a while to figure out why it thought 130 characters was okay, until I figured out the tab issue.

drgrice1:
I was thinking about having this as a setting that can be overridden per user, and I am not sure that it really makes that much sense.  It doesn't hurt for now, and so there is no need to make changes to this pull request at this point.  However, it generally doesn't make sense to say that for most students you need to have proctor authorization to have the quiz graded, but some select set of students does not (or vice versa).

When you asked before, I didn't really put much though into it.

dlglin:
> I was thinking about having this as a setting that can be overridden per user, and I am not sure that it really makes that much sense. It doesn't hurt for now, and so there is no need to make changes to this pull request at this point. However, it generally doesn't make sense to say that for most students you need to have proctor authorization to have the quiz graded, but some select set of students does not (or vice versa).
> 
> When you asked before, I didn't really put much though into it.

I can see a scenario where an instructor has an in-class quiz, but may want to allow certain students to write from somewhere else (an exam centre, home, etc.).  In this case they may want to turn off proctor authorization for those students.

drgrice1:
@dlglin:  What you are suggesting is extremely bad practice.  The only point of having a proctored exam when there are restrictions on resources or tools that are allowed to be used and you want to enforce those restrictions.  Enforcing those restrictions for one group of students, but not for others is unfair.  In the case of a testing center, you should provide a proctor account for the proctor at the testing center.  Not drop proctoring.

In any case, this pull request doesn't even allow what you are suggesting.  The only option that is made configurable per user is the option to skip the grade proctor authorization at the end of the exam.  It does not make it possible to configure a quiz to be proctored for some students but not for others.  So the only thing that you could do with the current status of this pull request is to make it so that some students need the proctor password to be entered at the beginning and end, but other students only need it entered at the beginning.

@somiaj:  One thing that bothers me a bit with this pull request is that it really breaks the permissions model of login proctor and grade proctor in spirit.  The point is that without this option a login proctor can only start the quiz, but a grade proctor (a higher permission level) must finish it.  With this option the one authorizing the start of the quiz really becomes both the login and grade proctor, and so should have the higher permission level.  So if a course has both types of users and quizzes utilizing the different options, then the login proctors will really have elevated privileges for the quizzes with this new option.

somiaj:
@drgrice1 I do see what you mean in terms of the spirit of the two permission levels. For me, this option is to allow proctored quizzes with a lower security level in which the instructor accepts that only a single login level proctor is needed. So in a case a course uses both types of user and quizzes, they should be aware of this subtly.

If we want to enforce the subtly more, one possibility I can think of is to require a grade proctor access to login if this option is used. Though this leads to the issue that the single password option then uses a login proctor level access, and it would not make sense to bump that permission level. Thus, if a single password is used, only login level access is needed, but if an actual proctor is needed, it needs to be a grade proctor level of access. This way there isn't an unintentional elevation of regular login proctor users. This will make the option slightly harder to understand (and to describe in a help tool tip), but will ensure that unless a single password is set, a grade proctor is required.

Alex-Jordan:
Case study. I have a remote section right now. There is one student with mobility disabilities and they simply cannot leave their home just to take proctored exams. The department policy is to have proctored exams in general, but with approval through the official accommodations process, that requirement can be waived. And it is for this student.

So there is more to this claim:
> Enforcing those restrictions for one group of students, but not for others is unfair.

than it may seem at first.

drgrice1:
As to the set level password option, that is an abstract user anyway.  So permissions don't matter at all in that case.  So if a grade proctor is required when this option is used the it is actually consistent.

drgrice1:
> Case study. I have a remote section right now. There is one student with mobility disabilities and they simply cannot leave their home just to take proctored exams. The department policy is to have proctored exams in general, but with approval through the official accommodations process, that requirement can be waived. And it is for this student.
> 
> So there is more to this claim:
> 
> > Enforcing those restrictions for one group of students, but not for others is unfair.
> 
> than it may seem at first.

Even in that case proctoring should still be required in my opinion.  There are remote proctoring options that could be used.

somiaj:
Update, to ensure that a grade proctor level (or higher) is always required with a proctored gateway quiz. Now the option is described as "Require Grade Authorization", and the help text says

> Proctored tests always require authorization to start the test. Set to "Yes" to require login 
> proctor authorization to start and grade proctor authorization to grade. Set to "No" to
> require grade proctor authorization to start and no authorization to grade.

@drgrice1 They are abstract users, but their permissions level is login proctor (from what I can tell), so I had to test for if a single password was being used to get this to work (vs changing their permission to grade proctor). This seems to work, and I made the error messages split up into the now three possible cases when starting a quiz. A grade proctor is required, a login proctor is required, or a single set password is required.

drgrice1:
Sounds good.  I will look it over when I get a chance.

somiaj:
One thing to consider that just occurred to me about the method I'm using to skip grade authorization, is the quiz is transformed from a proctored_gateway to a gateway at the start. Then the student gets full access from that point onward. So they could back out of the quiz and go back in without needing a new login proctor (unlike the traditional method, if the student leaves the quiz, and comes back, they need login authorization again).

This means that a student could just close the quiz, act like they are done, then go somewhere else and get back into it. (Though this can still happen with a traditional proctored quiz if a student is using their own laptop and never backs out of the quiz, just takes their laptop elsewhere -- assuming no grade proctor is also needed).

But there is a use case that they are using a computer in a computer lab, they shouldn't be able to get back into the quiz from another location if they didn't grade it. Thus it might be better to only change the quiz to a normal gateway quiz at the instant they grade it (not start it). Problem is I can't figure out exactly how to do this (do I intercept the Grade Test request some how, or do I just do this in LoginProctor.pm, and then redirect to the GatewayQuiz module after turning to a normal gateway quiz).

Personally I think the one use case this matters (a computer lab) should just require a grade proctor, since even if we did wait until Grade Test to turn into a gateway quiz without authorizations, a clever student could get around this with their own laptop. But if this is considered an issue, point me in the correct direction to transform from a proctored gateway to a gateway right after the Grade Test button is hit.

drgrice1:
@somiaj:  I have put in a pull request to your branch for this pull request with an answer to your latest question.  Instead of changing the    assignment type from 'proctored_gateway' to 'gateway' the first time that proctor authorization has been obtained, it skips obtaining proctor authorization when "Grade Test" is hit.  This means that if the user leaves the quiz for any reason the user will still need proctor authorization again to re-enter the quiz.  However, when the user hits "Grade Test" the test will just be graded without the need of proctor authorization.

somiaj:
Thanks, I see what it is doing and will test it when I get home from work tonight.

drgrice1:
Note that the case of a student using a laptop and going elsewhere to continue the test is a case that there is no way to prevent with this new option.  So that is something that we will need to make clear to those using this option in some way.

Another thing that seems confusing to me is the new name for the new option.  "Require Grade Authorization" seems to not quite say what the option does.  Either setting for the option really does still require grade authorization in some form.  At the end if the option is yes, and at the beginning if the option is no.  In addition the name doesn't seem to connect to the "proctored" quiz type.  The help icon popup clarifies this so I guess it is okay.

somiaj:
@drgrice1 I saw that and will fix.

I was also thinking about what to call the option, something that might not seem so clunky. My original thought was this is a "skip", but for a configuration option I thought the idea of a skip may be more confusing, so went with the require grade auth. After some thought would calling the option, "Require Proctor Authorization to Grade", be a little clear than just "Require Grade Authorization"?

somiaj:
My thought on a label for the option is assuming it is understood that authorization is already required to start (which is clarified in the help) but I can't think of a concise way to summarize this into a short label.

somiaj:
How about "Require Proctor Authorization To", and the options are "Start and Grade" and "Start"?

drgrice1:
> How about "Require Proctor Authorization To", and the options are "Start and Grade" and "Start"?

Yeah, I don't have a better short name to offer at this point.  That sounds okayish!

somiaj:
Update. I changed the name of the option in the data base to better reflect what it does too, so you'll have to upgrade the course database to test.

drgrice1:
I would suggest renaming the new database column to `use_grade_auth_proctor`, `use_proctor_grade_auth`, `proctor_use_grade_auth`, or something that contains `proctor`.  The point is to make it clear that this is an value that is for proctored sets.

drgrice1:
This has two approvals and I think it is ready to go in.  I am merging it now.

PR #1817, opened 2022-10-21
Do not UTF-8 encode email body text.

drgrice1:
The Email::Stuffer package tries to automatically detect if the body needs to be UTF-8 encoded (either that or it just does so by default), and so if you UTF-8 encode the message first, then Email::Stuffer does it again resulting in "Mojibake".

I am putting this in as a hotfix.  I am sure @taniwallach will agree.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -215,7 +215,7 @@ $emailableURL
 		}
 
 		my $email = Email::Stuffer->to(join(',', @recipients))->from($sender)->subject($subject)
-			->text_body(Encode::encode('UTF-8', $msg))->header('X-Remote-Host' => $remote_host);
+			->text_body($msg)->header('X-Remote-Host' => $remote_host);
 
 		# Extra headers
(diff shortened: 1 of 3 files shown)


taniwallach (changes requested):
@drgrice1 Thanks!

Tested and this current patch already fixes the Feedback messages. It also fixes the "report message" from the Email generation tool (such as when the course is in Hebrew so `maketext` will use Hebrew strings in the report message).

However, there is still Mojibake in the emails generated by the "Email" generation tool from the menu when Hebrew text is put in the message box. Fixing line 972 of `lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm` to drop the call the `Encode::encode` seems to fix that. Could you force push the branch with that change?

I did not try to figure out how to test the message from `jitar_send_warning_email` from `lib/WeBWorK/Utils/ProblemProcessing.pm` but the code change certainly seems correct.

Thanks for figuring this out and fix the issue.

drgrice1:
@taniwallach:  I missed the second instance of the encoding in the same file.  Thanks.  That is fixed now.

taniwallach (approved):

pstaabp (approved):
Didn't test (no email on develop machine), but code looks good.

PR #1820, opened 2022-10-22
Keep manual grader open when using sibling links in Problem.pm.

somiaj:
When using the manual grader in a homework set, add showProblemGrader=1 to the sibling's link parameters, so it remains visible (just like the next/previous problem links) when using those links to navigate the problems.


drgrice1 (approved):
This looks good.

pstaabp (approved):

PR #1816, opened 2022-10-21
Replace mod_perl2 with Mojolicious

drgrice1:
It is time to say goodbye to mod_perl2, and unbind webwork2 from apache2.  With Mojolicious there are many more deployment options.

The WeBWorK::Cookie module which was really just an alias for the CGI::Cookie module has been removed.  Instead this just uses the Mojolicious cookie methods.  Note that this does change the way the $CookieLifeTime setting works.  Previously it was a string representation of the life time like "+7d", and now it must be numeric and is always in seconds or the string "session" (in which case the expires and max-age values are not set and the cookie expires with the session).

Note that configuration that was previously in webwork2.apache2.4.dist.conf is now in webwork2.mojolicious.dist.yml. Unless you want to change the settings in that file, you don't need to copy that file to webwork2.mojolicious.yml.  The `dist` file will be used directly if the webwork2.mojolicious.yml file does not exist.

The XMLRPC::Lite and Apache::XMLRPC::Lite modules are not longer used.  The xml endpoints have been replaced with a direct remote procedure call endpoint.  This removes the XMLRPC middle man.

Note that the paths for this have changed.

The html2xml path still exists, but is an alias for the new render_rpc path.  The html2xml name doesn't make sense anymore since it does not use xml at all.  It should be considered deprecated and not used in new code.  It is provided to give those that use it time to make the change.   It still works the same at it did before.

The instructorXMLHandler path is now instructor_rpc.  Note that the response that is returned is now JSON, instead of a string that is parsed into JSON. If there is an error then the return object will contain an "error" key explaining the error.  There is no backwards compatibility path for this.  I don't think that this path was really used by anyone other than webwork2 itself.  If I am mistaken, an alias can be made.

These remote procedure call endpoints now use the usual authentication paths in the WeBWorK.pm dispatch method instead of waiting until the WebworkWebservice.pm module.  For this to work it is neccessary to change to using the usual authentication parameters.  Now use `user` instead of `userID`, `passwd` instead of `course_password`, and `key` instead of `session_key`.  Also, `courseName` is obsolete.  Previously both `courseID` and `courseName` were allowed.  However, if you use the now deprecated /webwork2/html2xml endpoint, you can still use the old parameters.  Those will not work with /webwork2/render_rpc.  You should work to switch to the new /webwork2/render_rpc endpoint, as the /webwork2/html2xml endpoint will be deleted eventually.

Note that the forcePortNumber parameter is removed.  Instead the port should be included in the $server_root_url in site.conf.

[...]

PR #1821, opened 2022-10-23
disable autocapitalization from username input field

Alex-Jordan:
This adds `autocapitalize="none"` to the attributes of the text input for a username. For people logging in with a phone, when they tap the input field, currently the keyboard starts out on capital letters (well, this may vary with OS/browser some, but it happens with mine using Android 13 and Chrome). Occasionally it leads to students entering their username the wrong way, not understanding the case sensitivity.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/LoginProctor.pm
+++ b/lib/WeBWorK/ContentGenerator/LoginProctor.pm
@@ -228,11 +228,13 @@ sub body {
 		print CGI::div(
 			{ class => 'col-xl-5 col-lg-6 col-md-7 col-sm-8 form-floating mb-2' },
 			CGI::textfield({
-				name        => 'proctor_user',
-				id          => 'proctor_user',
-				value       => '',
-				class       => 'form-control',
-				placeholder => ''
+				name           => 'proctor_user',
+				id             => 'proctor_user',
+				value          => '',
+				class          => 'form-control',
+				placeholder    => '',
+				autocapitalize => 'none',
+				spellcheck     => 'false'
 			}),
 			CGI::label({ for => 'proctor_user' }, $r->maketext('Proctor Username'))
 		);


pstaabp (approved):

lib/WeBWorK/ContentGenerator/Login.pm
```diff
@@ -220,24 +220,25 @@ sub body {
...
-				value         => $passwd,
-				aria_required => 'true',
-				class         => 'form-control',
-				placeholder   => ''
+				id             => 'pswd',
+				name           => 'passwd',
+				value          => $passwd,
+				aria_required  => 'true',
+				class          => 'form-control',
+				placeholder    => ''
```
drgrice1, line 241: Why the spacing change here?  Note that eventually (probably quite soon actually) there will be a github action added that will check that files are unchanged when perltidy is executed on them.  This spacing change that you added here will be reverted by perltidy.

drgrice1 (approved):
This looks good, and is something that I have noticed as well.

One minor spacing issue with the other change..

drgrice1:
Disabling spellcheck is another good addition here.  Note that these should probably also be added for the login proctor page.  Although it is not as critical for that page.  It is probably going to be extremely rare that a student will take a proctored quiz on a mobile device.

Alex-Jordan:
I undid the spacing change. I was just attempting to keep the alignment with the nearby parallel block, but I see now that perltidy won't care about that.

I also just disabled spellchecking on the username input field.

drgrice1:
Yeah, perltidy only looks at the immediate lines above and below.  If you put an empty line or a comment line between, it will use different alignment above and below the break or comment also.  The above and below a comment line difference is one that I wish it wouldn't do.

Alex-Jordan:
I put the same things in for the login proctor.

Then I thought there might be a centralized `CGI::username_field` for this. So I went looking [here](https://metacpan.org/pod/CGI) and there's a warning "HTML Generation functions should no longer be used". Does this indicate a big project for us?

somiaj:
Is it also possible to disable safari's password generator in the login proctor? An instructor here recently used a proctored gateway quiz with a single password to give to all students to start on their laptop in class, and the mac users had the mac/safari password generator generate and save a strong password for them, making it hard for them to easily type the password to start the quiz.

somiaj:
Unsure if this will fix the exact issue I was talking about, but might be worth while to set `autocomplete="off"` in the login proctor userID and password, so there won't be prompts to save login info in labs, and hopefully deal with auto fill when using a single password for students.

<https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion>

drgrice1:
@Alex-Jordan:  Yes, the Perl CGI module is deprecated, and yes, it would be a big project to replace it with modern equivalents.  It may never happen for webwork2.  Webwork3 does not use Perl CGI or even Perl at all.  It uses the Vue javascript framework (or really Quasar which is built on top of Vue).  You might note that Mojolicious templates are listed as an alternative to use to replace CGI on the page that you linked, and that is one way to go that we are poised to use with pull request #1816 in mind.

@somiaj:  You can add the `autocomplete="off"` attribute to the username and password field, but it won't work.  Most (if not all) browsers ignored that attribute for password fields, and even force auto completion to be on for the text input before it in the DOM.  The rationale is that it is bad practice to disable auto completion for username and password fields because that usually results in people using easier to remember and less secure passwords.  Browsers store the username and password data in a secure encrypted format, and this promotes using harder to remember more secure (even random) passwords.  Of course, that rationale doesn't apply to proctor login fields, but they don't think of that.  There are some javascript hacks to work around it, although browsers keep finding ways to prevent those from working.  We can try to implement one of those though.

drgrice1:
@somiaj:  In fact, the page you linked even says what I just said.  The page also mentions using `autocomplete="new-password"`, and that might work for some browsers, but that depends on if the browser implements that workaround.  Not all do.  In fact the compat table shows that Safari is unknown on being compatible with that.

somiaj:
Thanks for the info @drgrice1. Wasn't sure if it was fully possible or not. If adding `autocomplete="new-password"` helps on some browsers, might be worth adding as it is a fairly minimal change (and maybe more browsers will support this in the future). This is just a minor issue that I wasn't sure if there was a way to work around it due to password security. Unless the javascript hacks are fairly simple, probably not worth it if they just end up not working in the future.

drgrice1:
Adding `autocomplete="new-password"` is sort of odd though.  It does kind of work, but on Firefox is makes the autocomplete dropdown for the password field show a "Use a securely generated password" option.

The best javascript hack that I have found is to add the `readonly` attribute to the username and password field, and then remove that with javascript when the window load event occurs.   Then the browser doesn't auto fill the field.  Although the dropdowns to select saved credentials will still show up.  This worked in testing on both Firefox and Chrome.

Alex-Jordan:
As someone who has never used Gateway proctoring, can I confirm something basic? Is it really correct for the proctor password input field to be `type="password"` instead of `type="text"`?

I don't have much input on the password thing, but I think it would be OK to put that on a distinct PR, with this one being about usernames.

drgrice1:
I am not sure that I understand.  It seems to me that the proctor password field absolutely should be `type="password"`.  If it is `type="text"`, then the proctor password will be visible to the student when you type it in.

Yes, the handling of the proctor password field is a separate concern from the what this pull request addresses.  So a separate pull request is warranted.

Alex-Jordan:
I don't use that feature, and I wasn't sure how it's used. I've heard of it being used where students are in a lab, and the proctor just verbally tells everyone what password to enter. If that's how it's used, then maybe `type="password"` isn't appropriate (as in who cares if the students can see it; they are told to type it), and the whole issue with browser password management could go away. 

But if it's appropriate to have `type="password"` then of course it should stay that way. I just don't know how this feature is used or supposed to work and I wanted to clear up the basics.

somiaj:
The feature is a bit more complicated than that (and there also a pull request to add more options to it).

Currently a login proctor (username/password) is required for a student to start the proctored quiz, and a second grade proctor (username/password) is required to grade the quiz. There is an option to allow the login proctor to be done via a single password an instructor can tell the whole lab/class, so they all start at the same time, but the grade proctor is still required to enter their username/password for a student to grade the quiz. Hence there are cases where a username/password maybe typed in front of the student in which they should now know what it is.

The feature change is to make it so the grade proctor can be configured to be used or not, and in this case only a login proctor is required to start the quiz, but the grade proctor won't be required to grade (if configured). This is a new option and I don't think it has been merged yet.

drgrice1:
This has two approvals and is small enough.  I will merge it.

@somiaj:  You should create an issue about the login proctor autocomplete problem so it is not forgotten.

Issue #1814, opened 2022-10-19
Gateway quiz error if problems are not numbered 1,2,3,4,5,...


KoleNik:
If you have a gateway quiz such that the problems for example are numbered 1,2,4,6 instead 1,2,3,4 you will get an error. The quiz will start and if you open it a second time you can do the problems, but there is an initial error when you try to open it.
[WeBWorK_gateway_num_error.pdf](https://github.com/openwebwork/webwork2/files/9822179/WeBWorK_gateway_num_error.pdf)

drgrice1:
You must be running an old version of WeBWorK.  This but was fixed two releases ago.

KoleNik:
We are on 2.17 in the department. The server got upgraded this August, Well, I wasn't a part of the setup, but could this have been missed somehow from the 2.17 install?

drgrice1:
This is a valid issue it turns out.  There are still some cases where errors can occur with non-consecutive gateway quizzes.  One definite case is if there is no number 1.  There are some others as well.  A major case of interest is when the answer processing is done.

drgrice1:
This should now be fixed with #1823 which is now merged.  You will have to wait until the next release for it though.

PR #1823, opened 2022-10-28
Fix more issues with the gateway problem numbers.

drgrice1:
There were still cases where non-consecutively ordered gateway quizzes can cause errors.  In fact any time that a gateway quiz does not have a problem number 1 in the set definition, you will get an error.  In addition answer processing had issues that could cause errors. Basically, arrays storing problem data should generally not be accessed by the direct numerical index, but indirectly via the @probOrder array index.  There are some exceptions to this.  One exception is noted.

Note that this also fixes an issue with randomly ordered gateway quizzes.  The issue is perhaps minor, but the data stored in a column of the past_answers table was disjoint.  The past_answers_string and scores did not match up with the problem_id and source_file.  The source_file in that table is not used anywhere by webwork, but it still should match.  The problem_id is used.  In the past answers table, this mismatch leads to some rather unexpected results.  If a student takes multiple versions of the quiz, and you view the answers for all versions for that student, the answers clearly look like they come from different problems.  The same happens if you view answers for multiple students. One downside of this is that now the answers are not in the order that the problems appear for the student.  They are in the order they appear in the set definition.  That is consistent for all students and quiz versions though.

This fixes issue #1814 and the gateway random order/non-consecutive problem number issues ... hopefully all of them now.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -736,7 +736,7 @@ sub pre_header_initialize {
 				my $cleanSet = $db->getSetVersion($effectiveUserName, $setName, $setVersionNumber);
 				$set = $db->getMergedSetVersion($effectiveUserName, $setName, $setVersionNumber);
 
-				$Problem = $db->getMergedProblemVersion($effectiveUserName, $setName, $setVersionNumber, 1);
+				$Problem = $db->getMergedProblemVersion($effectiveUserName, $setName, $setVersionNumber, $setPNum[0]);
 
 				# because we're creating this on the fly,
 				#    it should be visible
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -1663,11 +1666,11 @@ sub body {
 					# we assume that allowed is an even
 					#    number of minutes
 					my $allowed = ($set->due_date - $set->open_date)/60;
-					$scoreRecordedMessage[$i] = $r->maketext("Your score was not recorded because you have " .
-						"exceeded the time limit for this test. (Time taken: [_1] min; allowed: [_2] min.)",
+					$scoreRecordedMessage[ $probOrder[$i] ] = $r->maketext('Your score was not recorded because ' .
+						' you haveexceeded the time limit for this test. (Time taken: [_1] min; allowed: [_2] min.)',
```
pstaabp, line 1670: 'you haveexceeded the'  to 'you have exceeded the ...'

drgrice1: Good catch.  It is fixed.

pstaabp (approved):
I can reproduce the error and this seems to fix it.  I have a quiz with 1 problem per page and the problem numbers are listed sequentially without the skip in numbers.  For example, the quiz is 1,3,4,5 and the problems are listed as 1,2,3,4.  Personally, I think this is the right way to do this (student get confused when numbers are skipped).

drgrice1:
That is actually the way it was before.  If the numbers were 1,3,4,5 in the set definition, then they are listed as 1,2,3,4 in the quiz itself.

To test this try creating a quiz with say 5 problems, and delete number 1.  If you do that with the current develop branch (or main) you will get an error when you try to start a quiz.  With this pull request you won't.

Another thing to look at is if you use randomization of problem order.  Create a quiz with say 5 problems each with a single text input answers and different source files.  Then open the quiz and enter the answers 1, 2, 3, 4, 5 into the problems in the order the problems appear.  Then look in the database at the past answers for the quiz.  You will see that the problem source file does not match the answers and scores.

drdrew42 (approved):
👍

PR #1825, opened 2022-10-28
Add perltidy workflow

drgrice1:
It is time to add this here too.

The first commit in this pull request contains the work flow and a `bin/dev_scripts/run-perltidy` script.

The second commit is the result of running `perltidy` on the code base.


pstaabp:
This looks good.  I assume no code changes.  

I'd recommend getting this in soon and then other branches will need to rebase off of develop.

drgrice1:
There was one perltidy error that forced a code change.  It was in `lib/WebworkWebservice/SetActions.pm`.  In that file there are two `getUserSets` methods defined that do slightly different things.  The first gets all sets for a given user, and the second gets all user sets for a given set.  So the second one was renamed to `getSetUserSets`.  I had already made the same name change in the mojolicious rewrite branch.  Since these aren't methods that are actually used by anything or anyone it isn't a big deal.

PR #1826, opened 2022-11-16
Stats page overhaul

somiaj:
This includes a major rewrite of the stats page. I was wanting to update it to include stats I find useful to see instead of exporting scores and creating them myself. This is also built upon #1816 so cannot be merged until after that, so making this a draft to get feedback. You will also need to install `libsvg-perl` to use, as I used that to create the svg graphs.

Some changes include:

+ Shorten page header to just state set name.
+ Include a table with info about the set.
+ Only include students who are assigned to the set in total stats, and restrict students to sections / recitations if such restrictions are defined for the user.
+ Don't include non students (guests, proctors, tas, profs, etc) in stats.
+ Crate two sub sections, overall and individual problems.
+ Only use the best version for each student with gateway quizzes.
+ Use libsvg-perl to generate the bar graphs.
+ Include average and standard deviation in overall stats.
+ Create a problem page for stats for individual problems.
+ Make problem links link to the problem stats page.
+ Allow one to render the problem on the problem stats page to view it vs creating a link to the problem.

I am not quite sure what to do with jitar sets, as I don't use them, so unsure what stats would be helpful there. I just mimicked what was already there for jitar sets, and on the problem stats page, top level problems will show both status and adjusted status. It might be preferable to only show top level problems for single problem stats, and include all the sub problem data there, but since I don't use them, unsure what info would be most helpful to users.

I also noted that the manual grader flag for `essay` questions doesn't seem to be working, as when I looked over the database of a course that used some essay questions, I never saw this flag being set. The links the old stats page created also go to the older manual grader page. I left the code there so the links will show up, but I couldn't figure out if/why the `essay` flag was not being set in the courses I looked at.

### DIFF
--- a/htdocs/js/apps/Stats/stats.js
+++ b/htdocs/js/apps/Stats/stats.js
@@ -1,0 +1,15 @@
+(() => {
+	// Hover action to show exact value of each bar.
+	document.querySelectorAll('.bar_graph_bar').forEach((el) => {
+		el.addEventListener('mousemove', (evt) => {
+			const tooltip = document.getElementById('bar_tooltip');
+			tooltip.innerHTML = evt.target.dataset.tooltip;
+			tooltip.style.display = 'block';
+			tooltip.style.left = evt.pageX + 10 + 'px';
+			tooltip.style.top = evt.pageY + 10 + 'px';
+		});
+		el.addEventListener('mouseout', () => {
+			document.getElementById('bar_tooltip').style.display = 'none';
+		});
+	});
+})();
--- a/htdocs/js/apps/Stats/stats.js
+++ b/htdocs/js/apps/Stats/stats.js
@@ -12,4 +12,113 @@
 			document.getElementById('bar_tooltip').style.display = 'none';
 		});
 	});
+
+	// Send a request to the webwork webservice and render a problem.
+	const basicWebserviceURL = `${webworkConfig?.webwork_url ?? '/webwork2'}/render_rpc`;
+
+	const render = () => new Promise((resolve) => {
+		const renderArea = document.getElementById(`psr_render_area`);
+
+		const ro = {
+			user: document.getElementById('hidden_user')?.value,
+			courseID: document.getElementById('hidden_course_id')?.value,
+			key: document.getElementById('hidden_key')?.value,
+			problemSeed: 1,
+			outputformat: 'simple',
+			showAnswerNumbers: 0,
+			set_id: document.getElementById('hidden_set_id')?.value,
+			probNum: document.getElementById('hidden_problem_id')?.value,
+			sourceFilePath: document.getElementById('hidden_source_file')?.value,
+			showHints: 1,
+			showSolutions: 1,
+			permissionLevel: 10,
+			isInstructor: 1,
+			noprepostambles: 1,
+			processAnswer: 0,
+			showFooter: 0,
+			displayMode: document.getElementById('problem_displaymode')?.value ?? 'MathJax',
+			language: document.querySelector('input[name="hidden_language"]')?.value ?? 'en',
+			send_pg_flags: 1,
+			extra_header_text: '<style>' +
+				'html{overflow-y:hidden;}body{padding:1px;background:#f5f5f5;}.container-fluid{padding:0px;}' +
+				'</style>',
+		};
+
+		if (!(ro.user && ro.courseID && ro.key)) {
+			renderArea.innerHTML = '<div class="alert alert-danger p-1 mb-0 fw-bold">'
+				+ 'Missing hidden credentials: user, session_key, courseID</div>';
+			resolve();
+			return;
+		}
+
+		if (ro.sourceFilePath.startsWith('group')) {
+			renderArea.innerHTML = '<div class="alert alert-danger p-1 mb-0" style="font-weight:bold">'
+				+ 'Problem source is drawn from a grouping set.</div>';
+			resolve();
+			return;
+		}
+
+		const controller = new AbortController();
+		const timeoutId = setTimeout(() => controller.abort(), 10000);
+
+		fetch(
+			basicWebserviceURL,
+			{
+				method: 'post',
+				mode: 'same-origin',
+				body: new URLSearchParams(ro),
+				signal: controller.signal
+			}
+		).then((response) => {
+			clearTimeout(timeoutId);
+			return response.json();
+		}).then((data) => {
+			// If the error is set, show that.
+			if (data.error) throw data.error;
+			// This generally shouldn't happen.
+			if (!data.html) throw 'A server error occured.  The response had no content';
+			// Give nicer file not found error
+			if (/this problem file was empty/i.test(data.html)) throw 'No Such File or Directory!';
+			// Give nicer problem rendering error
+			if ((data.pg_flags && data.pg_flags.error_flag) ||
+				/error caught by translator while processing problem/i.test(data.html))
+				throw 'There was an error rendering this problem!';
+
+			const iframe = document.createElement('iframe');
+			iframe.id = 'psr_render_iframe';
+			iframe.style.border = 'none';
+			iframe.srcdoc = data.html;
+			renderArea.innerHTML = '';
+			renderArea.append(iframe);
+
+			if (data.pg_flags && data.pg_flags.comment)
+				iframe.insertAdjacentHTML('afterend', data.pg_flags.comment);
+			if (data.warnings)
+				iframe.insertAdjacentHTML('afterend', data.warnings);
+
+			iFrameResize({ checkOrigin: false, warningTimeout: 20000, scrolling: 'omit' }, iframe);
+			iframe.addEventListener('load', () => resolve());
+		}).catch((err) => {
+			renderArea.innerHTML = `<div class="alert alert-danger p-1 mb-0 fw-bold">${err?.message ?? err}</div>`;
+			resolve();
+		});
+	});
+
+	const hide = () => {
+		const iframe = document.getElementById('psr_render_iframe');
+		if (iframe && iframe.iFrameResizer) iframe.iFrameResizer.close();
+	};
+
+	// Set up the render button.
+	document.getElementById('pdr_render')?.addEventListener('click', () => {
+		const btn = document.getElementById('pdr_render');
+		if (btn.innerHTML == 'Render Problem') {
+			btn.innerHTML = 'Hide Problem';
+			render();
+		} else {
+			btn.innerHTML = 'Render Problem';
+			hide();
+		}
+	}, { passive: true });
+
 })();
--- a/htdocs/js/apps/Stats/stats.js
+++ b/htdocs/js/apps/Stats/stats.js
@@ -1,23 +1,12 @@
 (() => {
 	// Hover action to show exact value of each bar.
-	document.querySelectorAll('.bar_graph_bar').forEach((el) => {
-		el.addEventListener('mousemove', (evt) => {
-			const tooltip = document.getElementById('bar_tooltip');
-			tooltip.innerHTML = evt.target.dataset.tooltip;
-			tooltip.style.display = 'block';
-			tooltip.style.left = evt.pageX + 10 + 'px';
-			tooltip.style.top = evt.pageY + 10 + 'px';
-		});
-		el.addEventListener('mouseout', () => {
-			document.getElementById('bar_tooltip').style.display = 'none';
-		});
-	});
+	document.querySelectorAll('.bar_graph_bar').forEach((el) => new bootstrap.Tooltip(el, { trigger: 'hover' }));
 
 	// Send a request to the webwork webservice and render a problem.
 	const basicWebserviceURL = `${webworkConfig?.webwork_url ?? '/webwork2'}/render_rpc`;
 
 	const render = () => new Promise((resolve) => {
-		const renderArea = document.getElementById(`psr_render_area`);
+		const renderArea = document.getElementById(`problem_render_area`);
 
 		const ro = {
 			user: document.getElementById('hidden_user')?.value,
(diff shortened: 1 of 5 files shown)


somiaj:
Thanks for the tests, seems some cases were missed in my testing.

The first warning I didn't see, but it is now fixed.

The second would happen if a student had a score of zero on a quiz when trying to figure out what the max version , and it would default to quiz version 0. Making the default quiz version 1 seemed to fix that.

The last was you have a quiz which has zero possible total points, put a check in for that before trying to divide by the possible points to figure out quiz grade.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -469,351 +538,856 @@ sub displaySet {
...
+				CGI::a(
+					{
+						class => 'btn btn-primary',
+						href  => $self->systemLink($urlpath->new(
+							type => 'instructor_problem_editor_withset_withproblem',
+							args => { courseID => $courseID, setID => $setName, problemID => $problemID }
+						))
+					},
+					$r->maketext('Edit Problem')
+				)
```
drgrice1, line 1066: For consistency with all of the other places that the problem editor is called, this should have the attribute `target => 'WW_Editor'` added so that it opens in a new window or tab.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -469,351 +538,856 @@ sub displaySet {
...
+			);
+		}
+		my $tag = @{ $opts{barLinks} } ? $svg->anchor(-href => $opts{barLinks}->[$_]) : $svg;
+		$tag->rect(
+			x              => $xPos,
+			y              => $opts{topMargin} + $opts{plotHeight} - $yHeight,
+			width          => $opts{barWidth},
+			height         => $yHeight,
+			fill           => $opts{barFill},
+			'data-tooltip' => $opts{isPercent} ? (100 * $data->[$_]) . '%' : $data->[$_],
```
drgrice1, line 1368: More on the tooltips.  Change this to:
```suggestion
			'data-bs-toggle' => 'tooltip',
			'data-bs-title'  => $opts{isPercent} ? (100 * $data->[$_]) . '%' : $data->[$_],
```

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -469,351 +538,856 @@ sub displaySet {
...
+		my $yHeight = int($opts{plotHeight} * $data->[$_] / $opts{yMax} + 0.5);
+		if ($opts{isJitarSet} && $opts{jitarBars}->[$_] > 0) {
+			my $jHeight = int($opts{plotHeight} * $opts{jitarBars}->[$_] / $opts{yMax} + 0.5);
+			$svg->rect(
+				x              => $xPos,
+				y              => $opts{topMargin} + $opts{plotHeight} - $jHeight,
+				width          => $opts{barWidth} + $opts{barSep},
+				height         => $jHeight,
+				fill           => $opts{jitarFill},
+				'data-tooltip' => $opts{isPercent} ? (100 * $opts{jitarBars}->[$_]) . '%' : $opts{jitarBars}->[$_],
```
drgrice1, line 1357: These are really ugly tooltips and are inconsistent with all of the Bootstrap tooltips used throughout webwork2.  Furthermore, a tooltip that follows the mouse around is certainly not desirable.  Change this to:
```suggestion
				'data-bs-toggle' => 'tooltip',
				'data-bs-title'  => $opts{isPercent} ? (100 * $opts{jitarBars}->[$_]) . '%' : $opts{jitarBars}->[$_],
```
and see my other comments to use much nicer looking Bootstrap tooltips instead.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -469,351 +538,856 @@ sub displaySet {
...
+	my $tooltip = ($id > 1) ? '' : CGI::div(
+		{
+			id    => 'bar_tooltip',
+			style =>
+				'position: absolute; display: none; background: cornsilk; border: 1px solid black; border-radius: 5px; padding: 5px;'
+		},
+		''
+	);
+
+	return $tooltip . CGI::div({ class => 'img-fluid mb-3', style => "max-width: ${imageWidth}px" }, $svg->render);
```
drgrice1, line 1390: More on the tooltips.  Change this to:
```suggestion
	return CGI::div({ class => 'img-fluid mb-3', style => "max-width: ${imageWidth}px" }, $svg->render);
```

htdocs/js/apps/Stats/stats.js
```diff
@@ -0,0 +1,124 @@
...
+			const tooltip = document.getElementById('bar_tooltip');
+			tooltip.innerHTML = evt.target.dataset.tooltip;
+			tooltip.style.display = 'block';
+			tooltip.style.left = evt.pageX + 10 + 'px';
+			tooltip.style.top = evt.pageY + 10 + 'px';
+		});
+		el.addEventListener('mouseout', () => {
+			document.getElementById('bar_tooltip').style.display = 'none';
+		});
+	});
```
drgrice1, line 14: The final step for making the Bootstrap tooltips work.  Change this to:
```suggestion
	// Tooltip to show exact value of each bar.
	document.querySelectorAll('.bar_graph_bar').forEach((el) => new bootstrap.Tooltip(el, { trigger: 'hover' }));
```

htdocs/js/apps/Stats/stats.js
```diff
@@ -0,0 +1,124 @@
...
+				'html{overflow-y:hidden;}body{padding:1px;background:#f5f5f5;}.container-fluid{padding:0px;}' +
+				'</style>',
+		};
+
+		if (!(ro.user && ro.courseID && ro.key)) {
+			renderArea.innerHTML = '<div class="alert alert-danger p-1 mb-0 fw-bold">'
+				+ 'Missing hidden credentials: user, session_key, courseID</div>';
+			resolve();
+			return;
+		}
```
drgrice1, line 52: I plan on removing these checks from the other places that have them and where you copied this from.  Most of the time these credentials will be there, and if not the fetch request will fail.  It really isn't a message that should be shown to the user.  The check is just not needed.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -469,351 +538,856 @@ sub displaySet {
...
+				type  => 'hidden',
+				id    => 'hidden_source_file',
+				name  => 'sourceFilePath',
+				value => $problemRecord->source_file
+			}),
+			CGI::div(
+				CGI::a(
+					{ id => 'pdr_render', class => 'btn btn-primary', role => 'button', tabindex => 0 },
+					$r->maketext('Render Problem')
+				),
```
drgrice1, line 1056: The class `pdr_render` doesn't really make sense here.  "pdr" is short for "problem detail row".  That makes sense on the problem set detail page, but not here.

drgrice1 (commented):
This looks like a nice improvement to the stats page.  I need to do some more testing and such, but here are some initial observations.

drgrice1:
By the way, it looks like you need to either rebase or merge to get Github to fix its file differences.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -969,6 +946,22 @@ sub links {
...
+									"${pfx}Stats",
+									text         => $r->maketext('Problem [_1]', $prettyProblemID),
+									urlpath_args =>
+										{ %args, statType => 'ste', setID => $nvSetID, problemID => $problemID },
+									systemlink_args => \%systemlink_args
+								)
+							);
+							print CGI::end_ul();
+							print CGI::end_li();
+						}
```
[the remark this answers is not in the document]
somiaj: Thanks for the fix.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
somiaj: Thanks for showing me how to use the bootstrap tool tips.

htdocs/js/apps/Stats/statproblem.js
```diff
@@ -0,0 +1,103 @@
...
+	document.getElementById('problem_render_btn')?.addEventListener('click', () => {
+		const btn = document.getElementById('problem_render_btn');
+		if (btn.innerHTML == 'Render Problem') {
+			btn.innerHTML = 'Hide Problem';
+			render();
+		} else {
+			btn.innerHTML = 'Render Problem';
+			hide();
+		}
+	}, { passive: true });
```
drgrice1, line 101: It is rather silly to set the option `passive: true` for a `click` event.  It doesn't do anything.  The `passive` option is for events that might possibly interfere with page scrolling like mouse or touch events.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -469,350 +665,857 @@ sub displaySet {
...
+	my $self = shift;
+	my $r    = $self->r;
+	my $ce   = $r->ce;
+	print CGI::script(
+		{
+			src   => getAssetURL($ce, 'js/apps/Stats/stats.js'),
+			defer => undef,
+		},
+		''
+	);
```
drgrice1, line 1508: This can all be consolidated onto one line.  That should be done to make this consistent with how it is done in the output_JS methods in all other files.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -469,350 +665,857 @@ sub displaySet {
...
+	if ($self->{type} eq 'problem') {
+		print CGI::script(
+			{
+				src   => getAssetURL($ce, 'js/apps/Stats/statproblem.js'),
+				defer => undef,
+			},
+			''
+		);
```
drgrice1, line 1516: Same here.  One line.

drgrice1 (commented):
I don't really like that you split the javascript into two files.  One file should work fine.

htdocs/js/apps/Stats/statproblem.js
```diff
@@ -0,0 +1,103 @@
...
+				'</style>',
+		};
+
+		if (ro.sourceFilePath.startsWith('group')) {
+			renderArea.innerHTML = '<div class="alert alert-danger p-1 mb-0" style="font-weight:bold">'
+				+ 'Problem source is drawn from a grouping set.</div>';
+			resolve();
+			return;
+		}
+
```
drgrice1, line 39: It would be good to add the following line here:
```
		renderArea.innerHTML = '<div class="alert alert-success p-1 mb-0">Loading Please Wait...</div>';
```

Frequently problems take some time to load if the network is slow or the problem requires some processing.  With the current setup it may appear to the user that clicking on the "Render Problem" button does not work in this case.  Adding this line lets the user know that it did.

somiaj:
@drgrice1 Thanks again. I split the javascript up since I thought it would be better to not load it on a page that doesn't have a render area, but I guess since there isn't a div with the corresponding id, it ends up being okay?

drgrice1:
Something seems to be wrong with this with JITAR sets.  No statistics seem to be reported at all.  Attached are screenshots showing what was there before and what is shown now.  For this particular set I have answered the first problem correctly.  The old stats page showed that.  The new does not.

old
new

somiaj:
@drgrice1 is the student that you are seeing the stats for a professor? Those users have been filtered out and only students are being shown. I could add a filter to include all users if you think that is preferable, my thought here is the stats are for students, not professors, tas, proctors, etc. Here is what I get with my test set that has student data.

image

and here is what a single top level problem looks like

image

drgrice1:
I see.  Yeah, I read that in your summary of changes, but I forgot when I was looking at this.

drgrice1:
> I also noted that the manual grader flag for `essay` questions doesn't seem to be working, as when I looked over the database of a course that used some essay questions, I never saw this flag being set. The links the old stats page created also go to the older manual grader page. I left the code there so the links will show up, but I couldn't figure out if/why the `essay` flag was not being set in the courses I looked at.

This seems to be working as it should.  The manual grader link shows up for essay problems for me with this pull request.  Those links should be to the older manual grader page.  There has not been a replacement for that.  The single problem grader works differently.  It is not a replacement for the manual grader page.

Note that the essay flag is not set for an essay problem until at least one student has answered the essay.

drgrice1:
The manual grader never worked for gateway quizzes.  That is one of the reasons that I wrote the new grader!

pstaabp (approved):
Errors I saw earlier are fixed and I think this is a nice improvement.

lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
```diff
@@ -219,6 +250,200 @@ sub body {
...
+	# Filter out all records except for current/auditing students for stats.
+	# Filter out students not in selected section/recitation.
+	my $filterSection = $r->param('filterSection');
+	my %sections;
+	my @outStudents;
+	for my $student (@studentRecords) {
+		# Only include current/auditing students in stats.
+		next
+			unless ($ce->status_abbrev_has_behavior($student->status, 'include_in_stats')
+				&& $db->getPermissionLevel($student->user_id)->permission == $ce->{userRoles}{student});
```
drgrice1, line 370: This needs to be reworked, and I think I was a bit hasty letting this pull request be merged.  This really should not be dependent on permission level at all.  It certainly should not be dependent on the hard coded student role.  It may be that an instructor wants to view status for other permission levels or roles.  A system may have custom permission levels and roles that are a different type of student, and that stats should be displayed for.

The entire point of the statuses set in `defaults.config` is for this purpose, and this change is subverting that intent.  I think that it is time that the status abbreviations evolve.  Instructors and others that are not "enrolled" students should not have the status "enrolled" (abbreviated as C, c, current, or enrolled).  There should be a different status for these users.

Issue #1580, opened 2022-02-06
Problem editor can save and overwrite a pg file to a different problem.


somiaj:
The following will cause the problem editor to overwrite a pg file to a different problem, causing data loss. The order is important here:

+ Have an assignment with multiple problems, choose to edit problem 2.
+ This opens up the editor in a new window/tab.
+ Without closing the editor window/tab, move the problems around (swap problems 1 and 2) and save the set.
+ Go back to the open tab, and try to update/save the original problem 2 pg file.

This says it will save the changes to the original problem 2 pg file, but when the window opens it says it has saved the changes to the new problem 2 pg file, thus overwriting the new problem's pg file.

It seems it is using the pg file located in the database for problem 2 to determine what to save to, instead it should probably get the info about which file to save to from the post request (or have some check to note that the problem has been moved and not overwrite the new problems pg file).

pstaabp:
@drgrice1 Looks like you have a branch that addresses this, but never put in a PR.

drgrice1:
Yeah, I need to finish that.  I did to much in the branch (code clean up stuff), and never got around to fixing it.

PR #1827, opened 2022-11-16
Revamp the PGProblemEditor

drgrice1:
The problem editor now renders the problem (or header file) immediately when the page loads in an iframe directly on the page, instead of the hack of loading a complete webwork page in an iframe in a modal dialog. That led to all of the crud of having to hide the main menu and other page elements that looked odd embedded in a dialog.

When the "View/Reload" tab is selected and "Take Action" is clicked and "Open in new window" is not checked the problem is reloaded with content updated from the CodeMirror editor's current contents.  If a submit button is pressed in the problem the contents are also updated with the CodeMirror editor's current contents.  Also, when "Randomize Seed" is clicked the problem reloads immediately with the new seed.  In any of these cases a temporary file is saved.  A new WebworkWebservice method has been added for this purpose when rendering in the same window.  The javascript uses this to save the temporary file.  The method only allows saving in a course's templates directory and only if the user has the permission to modify_problem_sets.

Note that the "View/Reload" tab is now shown for course info files.  I don't know why that was disabled, it was fully functional.

When editing hardcopy header files, the in page renderer just gives a message that hardcopy header contents can only be viewed in a new window, and the "Open in new window" checkbox is checked by default.

If a pg file is saved as a new version, then auxiliary files are also copied, even if the file basename and directory name are not the same. That really didn't make sense to only copy the auxiliary files in that case.  Usually that will not be the case anyway, and the resulting copied file will fail to work without its resources.

The completely unused options_info file editing capability has been removed.

Add a sanity check to the PG problem editor to check that the file that the user is saving matches the file on record.  This adds a hidden input to the save form that stores the file name that was in the database for the problem when the editor is opened.  When the user attempts to save the file, that path is verified.  If the path does not match what is in the database at that time, then refuse to save the file.  This fixes issue #1580.

The "Report Bugs in this Problem" button is now only shown for problems in the OPL or Contrib.

Convert WebworkClient formats to Mojolicious templates.  Now there are only three template files, default.html.ep, default.json.ep, and ptx.html.ep.  The simple, sticky, and default output formats all use the default.html.ep template.  The ptx output format clearly uses the ptx.html.ep template.  The json format uses the default.json.ep template.  The raw format is still just code in formatRenderedProblem. The useless body_text output format has been removed, as well as the dysfunctional tex output format.  Note the FormatRenderedProblem module no longer uses CGI at all.

Two new options for submitting the problem source are added to the renderProblem method.  The first is raw unnencoded source (the parameter is rawProblemSource). This will not work embedded in html as is often done with the formats for embedding in an iframe, but can be done when form submission is handled by javascript.  The second is uri encoded source (the parameter is uriEncodedProblemSource). One could Use encodeURI in javascript to encode the source.  Server side that is decoded using Mojo::Util url_unescape.  This does work for embedding the source in the html.  Base64 encoding really should never be used, and just needs to go away.

Finally, a new action tab has been added to the editor page.  A "Generate Hardcopy" tab.  This tab is visible when editing a problem or set header.  There are options to select the seed, hardcopy format (pdf or text), and harcopy theme (the themes set in conf files).  Note that if the pdf format is selected and there are errors then a zip file will be offered for download, otherwise of course a pdf file will be offered. If the tex format is selected then a zip file will be sent.  Another new WebworkWebservice option is added for this.  This replaces the dysfunctional tex output format.  It is called hardcopyRenderedProblem (not sold on that name).  It works much like formatRenderedProblem, except it generates a pdf or zip file.  First renderProblem is called and then, if the outputformat is tex or pdf hardcopyRenderedProblem is called.  If a problem results in no body text, there are translator errors, or in the extreme event that neither a pdf or zip can be created, then a text error is rendered instead.

Note that a more reliable method for adding resource files to the zip archive for the tex format is used and is also implemented in Hardcopy.pm.


drgrice1:
> Testing on problems works well.
> 
> Just tested on the course info. There wasn't a course_info.txt file, a temp one was made and then on update, a new one was created as expected. However, a simple text "This is the course info" fails. It seems to be parsing it like a problem file. I have the error:
> 
> ```
> ERRORS from evaluating PG file: Can't locate object method "the" via package "course" (perhaps you forgot to load "course"?) at line 2 of course_info.txt
> ```

This explanation is not very clear.  Can you clarify what exactly you did?

pstaabp:
Also, in the new generate hardcopy option on the set header, the file downloaded was in double quotes for me. Instead of file.pdf it was "file.pdf"

Removing the double quotes looked correct.

drgrice1:
> Also, in the new generate hardcopy option on the set header, the file downloaded was in double quotes for me. Instead of file.pdf it was "file.pdf"
> 
> Removing the double quotes looked correct.

This is fixed.

drgrice1:
@pstaabp:  I figured out what you did with the course header file.  So it turns out that is also what happens currently with set header files.  So it is not a regression with this pull request.  In any case, I will see what I can do to fix that.

drgrice1:
@pstaabp:  I added some changes to fix the handling of course info files.

pstaabp:
Both of the issues I found are now fixed.

drgrice1:
@pstaabp:  The correct dates for the set will now be displayed when viewing set headers.

pstaabp (approved):
Looking good now.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -707,782 +586,681 @@ sub body {
...
-	###########################################################
-	# If things have worked so far determine if the file might be accompanied by auxiliary files
-	# a path ending in    foo/foo.pg  is assumed to contain auxilliary files
-	#
-	my $auxiliaryFilesExist = has_aux_files($outputFilePath);
-
-	if ($auxiliaryFilesExist and not $do_not_save) {
-		my $sourceDirectory = $sourceFilePath || '';
-		my $outputDirectory = $outputFilePath || '';
+	# If the file is being saved as a new file in a new location, and the file is accompanied by auxiliary files
```
Alex-Jordan, line 915: I think something may be wrong with this. Here is one way to see the issue (there may be more). In the current WeBWorK-2.18 branch, go to some set (I went to a version of the "Orientation" set that comes with the distribution.

Click to Edit the set header. Use the "Save As" tab, saving as something like `[TMPT]/BUG/demo.pg`. Note the `BUG` folder should not previously exist. Check the button for "Create unattached problem", and then click the "Save As" button.

You should see the message `Copied auxiliary files from ... to new location at /opt/webwork/courses/.../templates/BUG.` And what did it copy there? Way too much it seems. In addition to `demo.pg`, that folder for me has:

```
course_info__1.txt
course_info_1.txt
course_info.txt
hardcopyPreamble.tex
hardcopySetDivider.tex
login_info.txt
setDemo.def
setOrientation.def
demoCourse.lst
hardcopyPostamble.tex
hardcopyProblemDivider.tex
hardcopyUserDivider.tex
set0.def
setMAAtutorial.def
```

It's like every file from the `templates/` folder that is not a `.pg` file was copied into here.

somiaj: I tested this and it seems that it considered auxiliary files any non `.pg` file in the same directory as the `.pg` file. I think the idea here is to copy over any images or additional files that might be needed for a problem to work.  Since the first file I tested was in a directory with no other non .pg files, so only the .pg file was copied over.

I tested this with a problem I have that has various `.png` images in the same directory, and when I saved a new unattached problem, it copied over the problem and all the `.png` files (which in this case was useful).

It seems that saving `.pg` files in the templates directory is what is causing all those files to be copied.  I don't think this issue was introduced with this PR, but might be something that could be fleshed out more in deciding what an auxiliary file is.

One simple fix might be a checkbox that allows users to copy over auxiliary files which also gives a comment that it will copy over all files in the directory.  This way these files will only be copied if the user wants them, but not by default.

somiaj: I wonder if this can help better identify if a file is an auxiliary file, so instead of just copying everything over, only copy over the filenames that also appear inside the .pg file using grep to test.

drgrice1: This is how it has always worked.  I didn't change this behavior in this pull request.  I removed the WeBWorK::Utils `has_aux_files` method that really just returned true if a file has the `.pg` extension, and false otherwise, and just put that code here.

The assumption is that any file in the directory that does not have the `.pg` extension is an auxiliary file that may be needed for the problem to function.  Thus these files are copied as well.  Unfortunately, there is no reliable way to determine what files really are needed for the problem.

drgrice1: @somiaj:  A grep would miss a lot of things.  Classically problems have done things like have multiple images where one of them is selected with random parameters.  For example, there would be several files names image1.png, image2.png, ..., and the problem would contain the code `$a = random(1, n); $im = "image$a.png"`.  A grep would not find `image1.png` or any of the other `image?.png` files.

Alex-Jordan:
I see. I was definitely confused and not aware that I was copying a file
from templates/. All I really wanted was a quick way to get to the editor,
and then later I was surprised to see all the extra files there.

It sounds like there are several issues with auxiliary files for a problem,
but they are old issues.

On Mon, Jun 19, 2023 at 3:56 AM Glenn Rice ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-3a1dfdd268a78095&q=1&e=b529853a-b424-4c91-a005-a40db2a60042&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fwebwork2%2Fpull%2F1827%23discussion_r1233889502>
> :
>
> >  	}
>
> -	###########################################################
> -	# FIXME if the file is accompanied by auxiliary files transfer them as well
> -	# if the filepath ends in   foobar/foobar.pg  then we assume there are auxiliary files
> -	# copy the contents of the original foobar directory to the new one
> -	#
> -	###########################################################
> -	# If things have worked so far determine if the file might be accompanied by auxiliary files
> -	# a path ending in    foo/foo.pg  is assumed to contain auxilliary files
> -	#
> -	my $auxiliaryFilesExist = has_aux_files($outputFilePath);
> -
> -	if ($auxiliaryFilesExist and not $do_not_save) {
> -		my $sourceDirectory = $sourceFilePath || '';
> -		my $outputDirectory = $outputFilePath || '';
> +	# If the file is being saved as a new file in a new location, and the file is accompanied by auxiliary files
>
> @somiaj
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-a2dbbe6078f1ac06&q=1&e=b529853a-b424-4c91-a005-a40db2a60042&u=https%3A%2F%2Fgithub.com%2Fsomiaj>:
> A grep would miss a lot of things. Classically problems have done things
> like have multiple images where one of them is selected with random
> parameters. For example, there would be several files names image1.png,
> image2.png, ..., and the problem would contain the code $a = random(1,
> n); $im = "image$a.png". A grep would not find image1.png or any of the
> other image?.png files.
>
> —
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-3a1dfdd268a78095&q=1&e=b529853a-b424-4c91-a005-a40db2a60042&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fwebwork2%2Fpull%2F1827%23discussion_r1233889502>,
> or unsubscribe
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-9318df00aa265ef2&q=1&e=b529853a-b424-4c91-a005-a40db2a60042&u=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABEDOADPSTNZDYHLIYTSCSTXMAV37ANCNFSM6AAAAAASCYIL6E>
> .
> You are receiving this because you commented.Message ID:
> ***@***.***>
>


-- 
Alex Jordan
Mathematics Instructor
Portland Community College

Issue #1828, opened 2022-11-17
Network error in Library Browser


Alex-Jordan:
In 2.17, if I

1. visit the Library Browser
2. select "Calculus - single variable" from the dropdown Subject menu
3. click View Problems _quickly, before `There are ____ matching WeBWorK problems` has time to update_

Then Firefox throws an error at me in some kind of dialog box:
`/webwork2/instructorXMLHandler  NetworkError when attempting to fetch resource.`

Is this something that can be fixed? Or sidestepping the issue, is it possible something about my configuration could be improved to speed up the time to update the count of matching exercises?

somiaj:
I double checked that I do get the same issue (though I have to be really quick). It appears that when you click view it sends a request to load a new page, thus interrupting the XLMhandler, and even though the error appears, the page still loads and renders the problems just fine.

How long is it taking your system to fetch the problems? Here it is less than a second.

Alex-Jordan:
I wasn't trying to be fast when I stumbled upon this. I was just trying to check that Library Browser worked on a new installation. And it was consistently happening until I had developer tools open. Having that open slowed me down enough for clicking the show button, and then for a bit I was confused as to why the problem went away "just because developer tools were open". Eventually I realized it had to do with the fetching of the updated count.

Anyway, it's between 1 and 2 seconds for me.

drgrice1:
I was able to reproduce this once, the first time I tried it.  Then I couldn't do it again.  Lost my click speed I guess.  I could only reproduce it on my production server with WeBWorK 2.17, and not on the develop branch with morbo.  That is probably only because there is no way I could possibly beat the network response to and from the local morbo server though.

In any case, the one time I reproduced it, I sat there looking at the toast reporting the error, and it suddenly disappeared and the page loaded the problems anyway.

What is happening is that when "View Problems" button is clicked it submits the form on the page.  Any time a new page request is submitted it aborts any javascript requests in progress that have not yet completed.   Unfortunately, I don't think there is anything that we can do about that.  I doubt there is really anything you could do with your configuration to speed things up either.  It is simply a factor of the speed of the connection between the computer and the server, and the time it takes for the server to process the request.

pstaabp:
This was fixed in #1829.

PR #1829, opened 2022-11-17
Don't show network errors in the library browser when unloading the page.

drgrice1:
This just stops showing error toasts when the window's beforeunload event fires on the library browser page.

This can happen if you select a subject and then rapidly click on "View Problems" before the request for library listing for the selected subject completes.

The problem is that when the form on the page is submitted this causes the browser to navigate to a new page, and all javascript network requests are aborted.  That is done by the browser, and there is nothing that can be done about that.  Usually this isn't a problem since the new page loads, except in this case the new page that is requested could be a slow loading page (for example if there are more than 8000 matching problems), giving the javascript time to react and show the error toast that the request was aborted before the new page actually loads.

Although I must say this is a rather extreme edge case.  To test this I had to slow down the server response to have time to select the subject and then click "View Problems" before the page load occurred.  I added `for (0 .. 100000000) {}` to the beginning of WebworkWebservice::LibraryActions::searchLib.

See https://github.com/openwebwork/webwork2/issues/1828.

### DIFF
--- a/htdocs/js/apps/SetMaker/setmaker.js
+++ b/htdocs/js/apps/SetMaker/setmaker.js
@@ -2,8 +2,13 @@
 	const webworkURL = webworkConfig?.webwork_url ?? '/webwork2';
 	const basicWebserviceURL = `${webworkURL}/instructor_rpc`;
 
+	let unloading = false;
+	window.addEventListener('beforeunload', () => unloading = true);
+
 	// Informational alerts/errors
 	const alertToast = (title, msg, good = false) => {
+		if (unloading) return;
+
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
I couldn't get the error toasts to appear, but the code looks good.

drgrice1:
If you add the `for` loop to the beginning of the `searchLib` method as I mentioned, it makes it rather easy to make the error toasts appear.

pstaabp:
Saw the toast with error now and this fixes it.

drdrew42 (approved):

Issue #1824, opened 2022-10-28
Disable autocompletion of proctor login username/password.


somiaj:
The proctor login username/password on the LoginProctor.pm page are not a standard username/password form that a browser should save into its password database or generate a secure password for since they are either meant to be used by a different user than the one currently logged into webwork, or a single generic password that is provided for all users to use.

This form should ideally not be auto filled with the credentials from the browsers saved logins for the webwork site, and it should not prompt to save the entered password or to generate a secure password. Since this is a fairly nonstandard use of a username/password, this doesn't seem to be something that browsers seem to fully support, so it might require some javascript for this.

This was discussed in #1821.

pstaabp:
This is fixed in #1821.

drgrice1:
Actually, it is not.  This issue was opened as a result of that pull request and the discussion therein.

drgrice1:
This is fixed in #1830!

PR #1830, opened 2022-11-18
Disable browser password autocompletion on the proctor login page.

drgrice1:
This just adds the attribute `autocomplete="new-password"` to the proctor user and password fields.  This is the recommended way to do this according to
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#preventing_autofilling_with_autocompletenew-password.

However, this may not be supported by all browsers.  Safari is listed as support unknown.  So could a Safari user test this and let me know if it works?  If it doesn't, I have another way that should work for all browsers.

This is to address issue #1824 (also see the discussion in pull request #1821).

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Login.pm
+++ b/lib/WeBWorK/ContentGenerator/Login.pm
@@ -261,7 +261,7 @@ sub body {
 			print CGI::end_div();
 		}
 
-		print CGI::submit({ type => "submit", value => $r->maketext("Continue"), class => 'btn btn-primary' });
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
Checked in safari.  It doesn't autofill the password, but does give the option to fill a password from my keychain.

drgrice1:
Yeah, that is how it works on Firefox and Chrome also.  So it seems that it works the same for Safari.  That means this works for what we want.  The important thing is that it doesn't autofill the username and password so that the proctor needs to delete it to enter the proctor username and password.

drdrew42 (approved):

PR #1831, opened 2022-11-18
Fix utf8 characters in warnings and morbo logs to the terminal.

drgrice1:
The 'warnings' stash value is not utf8 encoded.  That was removed with the switch to Mojolicious since there is no need for it.  That means that it should not be utf8 decoded when it is used.

There was also a problem created by the `use open qw(:encoding(UTF-8) :std);` call that was in lib/WeBWorK/File/Classlist.pm.  When `use open` is called in that way it has global scope and affects all standard handles for the entire application.  Most importantly it was setting the encoding for STDERR.  Mojolicious writes already encoded logs to STDERR when morbo is used.  This results in mojibake in the terminal.  Never call `use open` in that way.  Instead use `use open IO => ':encoding(UTF-8)'` to apply encoding only in the lexical scope of where it is called.  Even better, don't call `use open`.  Just apply encoding as needed when opening a file.

Many of the binmode calls also need to be reviewed.  This isn't being done correctly and there are still issues with this.  This is most important for PG code going forward.

You can test this with the following problem:  
[z_squared2.pg.txt](https://github.com/openwebwork/webwork2/files/10044002/z_squared2.pg.txt)

Without this pull request the `warn` statement will make the request die.  With this pull request the problem will render and the warning will be displayed in the usual way.  Furthermore the morbo terminal log will properly display the Hebrew warning.  In production the Hebrew characters will properly appear in the `logs/webwork2.log` log file.

### DIFF
--- a/lib/Mojolicious/WeBWorK/Controller/Handler.pm
+++ b/lib/Mojolicious/WeBWorK/Controller/Handler.pm
@@ -26,7 +26,6 @@ use Mojo::Base 'Mojolicious::Controller', -signatures, -async_await;
 use HTML::Entities;
 use HTML::Scrubber;
 use Date::Format;
-use Encode;
 use JSON::MaybeXS;
 use UUID::Tiny ':std';
 
@@ -56,7 +55,7 @@ async sub handler ($c) {
 			$log->warn("[$uri] $warning");
 		};
 
-		$SIG{__DIE__} = my $exception_handler = sub {
+		$SIG{__DIE__} = sub {
 			@backtrace = backtrace();
 			die @_;
 		};
@@ -77,10 +76,6 @@ async sub handler ($c) {
 	if ($@) {
 		my $exception = $@;
 
-		# Set the binmode for print to utf8 because some language options use utf8 characters.
-		binmode(STDOUT, ':encoding(UTF-8)');
-		my $warnings = Encode::encode('UTF-8', $c->stash->{warnings});
-
 		my $htmlMessage;
 		my $uuid = create_uuid_as_string(UUID_SHA1, UUID_NS_URL, $uri) . "::" . create_uuid_as_string(UUID_TIME);
 		my $time = time2str('%a %b %d %H:%M:%S %Y', time);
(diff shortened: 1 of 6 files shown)


pstaabp (approved):
Checked the login as well.  Looks good.

taniwallach (changes requested):
I tested the login page, import/export of a class list file, and the provided problem which issued a warning with Hebrew characters. All work properly after the patch.
Before the patch, as reported, fatal errors occurred on invalid logins when the course language was `he-IL` and when attempting to view the problem.

I approve the patches, but found one more fix which seems related and necessary.

---

I did a search for where else a `decode` may potentially cause a similar error using:
```
ind lib -type f -exec grep -Hn "decode" {} \;  | grep -Ev "decodeAnswers|JSON->new->decode|decode_utf8_base64|decode_base64"

```

The only things which caught my eye are needed to be fixed is the code in `SteMaker.pm` which used `Encode::decode`. The existing code also makes problems is a local "library" is set up with a name using UTF-8 characters. Ex. create a `Technion` subdirectory in `templates` and add
```
$courseFiles{problibs}->{Technion} = "טכניון";
```
to `course.conf`. Then try to access the library browser. If gives a fatal error
```
Wide character at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm line 1158.
```
It did not work in WeBWorK 2.17 either (also reporting a wide character), but I never tried using such a setting until I found the `Encode::decode` when looking now for anything else which may break for the same reason as the bugs fixed here.
I think I fixed it in the patch at https://github.com/drgrice1/webwork2/pull/8

drgrice1:
I happen to be working on `SetMaker.pm` right now in the process of removing the CGI module usage, and I also noticed those.  I hadn't gotten around to testing it yet though.  I will merge your pull request to this one.

taniwallach:
> I happen to be working on `SetMaker.pm` right now in the process of removing the CGI module usage, and I also noticed those. I hadn't gotten around to testing it yet though. I will merge your pull request to this one.

Sorry - I did not perltidy after the change to `SetMaker.pm`. Sorry - it has not yet become a habit.
It probably makes sense to run the `perltidy` and rebase into one commit.

drgrice1:
No problem.  I perltidied it!

taniwallach (approved):

drgrice1:
Note that there is a difference from the `use open qw( :encoding(UTF-8) :std );` that was in `ClassList.pm` and the `use open IO => ':encoding(UTF-8)';` that is in `Utils.pm`.  The `IO => ...` part restricts the encoding to the calling scope.

You have to read the documentation on this carefully and the wording is not entirely clear on this.  See https://perldoc.perl.org/open.

taniwallach:
> Note that there is a difference from the `use open qw( :encoding(UTF-8) :std );` that was in `ClassList.pm` and the `use open IO => ':encoding(UTF-8)';` that is in `Utils.pm`. The `IO => ...` part restricts the encoding to the calling scope.
> 
> You have to read the documentation on this carefully and the wording is not entirely clear on this. See https://perldoc.perl.org/open.

@drgrice1 Thanks for the additional explanation. I just read the documentation there, but I seem to understand it differently than what you have suggested.

It seems to me that the `IO =>` part is not what is making the difference. From what I understand that just means that the "layers" get applied to both input and output streams (to set defaults) but that should only be for the "lexical scope" where the pragma is applied.

Instead, I think the problem in `lib/WeBWorK/File/Classlist.pm` was that it used `:std` which is supposed to break out of the lexical scope and modify the "layer stack" for the global standard filehandles (`STDIN`, `STDOUT`, `STDERR`).

Certainly when the `use open qw( :encoding(UTF-8) :std );` was put in `ClassList.pm` almost 10 years ago (https://github.com/openwebwork/webwork2/commit/ba90466c1faf487e431aec43ada56bf0380cc7e0) that was overkill, as it certainly seems that all that was intended was to read the specific class list file as being a UTF-8 encoded file.

BTW I'm surprised that the intended support for UTF-8 characters in names worked at all back then, long before the database tables were intentionally supporting UTF-8 encoding in the tables or the database connection. Apparently the database driver was somehow making it work, probably by some hidden encoding/decoding occurring as data was put into and retrieved from the database. I wonder how well the script in `bin/upgrade-database-to-utf8mb4.pl` (https://github.com/openwebwork/webwork2/pull/1309) handles such things.

drgrice1:
Yeah, I think that is right, and I think that is how I also interpreted it the first time I read it when I made the change for this pull request.  Although, I think that documentation is quite pathetic.  The wording is quite bad, and it is entirely unclear what the author is even talking about.

taniwallach (approved):

PR #1832, opened 2022-11-19
CodeMirror syntax highlighting match multiple digits for lists.

somiaj:
This is a small adjustment to match https://github.com/openwebwork/pg/pull/750, so the syntax highlighter will match numeric lists created with multiple digits. Also adds l and L to roman numeral lists, which was missing.


pstaabp (approved):
Looks good.  I'm not sure I want to write a problem with L (50) parts, but ...

drgrice1 (approved):
This looks good.

PR #1833, opened 2022-11-21
Fix the showHintsAfter setting for old style problem list files

drgrice1:
I didn't test this for version 1 type problem list set definition files when I added the showHintsAfter setting, and of course it doesn't work. The showHintsAfter setting is guaranteed to not be in any of these files, and so the setting ends up being set to the empty string.  The old style list format doesn't fill in defaults, and so the empty string is put into the database.  The database interprets this as a null value and croaks since the column is not null.

To fix this the method for parsing the old style lists just needs to use the default value from the conf files.

To test this try to import one of these old style problem set list files (for example `Contrib/Michigan/gateways/calIIentr/setchain_rule.def`).  Without the current develop branch you will get the error "DBD::MariaDB::st execute failed: Column 'showHintsAfter' cannot be null".  With this pull request the set will import correctly.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
@@ -2168,14 +2168,14 @@ sub readSetDef {
 				push(
 					@problemData,
 					{
-						source_file    => $name,
-						value          => $weight,
-						max_attempts   => $attemptLimit,
-						showMeAnother  => $showMeAnother,
-						showHintsAfter => $showHintsAfter,
-						# use default since it's not going to be in the file
-						prPeriod     => $prPeriod_default,
-						continuation => $continueFlag,
+						source_file   => $name,
+						value         => $weight,
+						max_attempts  => $attemptLimit,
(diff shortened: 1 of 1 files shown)


pstaabp:
This PR fixes the database error listed.  

I also tried to load this in the Library Browser and get a separate error:
```
can not open /opt/webwork/courses/staab_course/templates/../../../problembanks/gateways/derivative/topic_chain_rule/prob1.pg: No such file or directory at /opt/webwork/webwork2/lib/WeBWorK/Utils/Tags.pm line 214.
```

It seems like this (and nearly all of the sets inside the `Contrib/Michigan` directory) point to this directory.  These probably shouldn't be in the OPL since it reads a directory outside the OPL. I don't know if there is anything else we should do with this.

pstaabp (approved):

drgrice1:
So that is a separate issue, but I agree that those set definition files that point to problems not in the OPL directory should not be in the OPL.  In my opinion, all of the set definition files in the OPL should be removed.  The OPL should be problems only.

drdrew42 (approved):

PR #1839, opened 2022-11-24
Rename Docker stage 1 image for develop (to become 2.18) to avoid conflict with that from WW 2.17

taniwallach:
Refer to the Docker base image in the two stage Docker build as `webwork-base:forWW218` since the change to Mojolicious requires a very different base image than was needed for Apache+mod_perl.

As a result, building a stage 2 image for WW 2.17 and develop/WW 2.18 requires using the correct base image, and the tag difference allows both the be available at the same time.


drgrice1 (approved):
Looks good.  Then we don't need to worry about changing it for the release either.

pstaabp (approved):

PR #1840, opened 2022-11-24
Use a fixed submitTime in time based checks in Problem.pm and GatewayQuiz.pm

taniwallach:
In `Problem.pm` change the code to use a fixed `submitTime` value during all the processing related to time base conditionals, and in particular when determining if the recorded score/answers should be updated.

The time being tested in `compute_reduced_score` should also depend on the recorded `submitTime` (and not the current time) so is passed to that subroutine as an additional parameter.

For `GatewayQuiz.pm` use the already saved `$timeNow` as the submission time which gets passed to `compute_reduced_score`. I hope that is a correct approach for gateways.

One other conditional test in `GatewayQuiz.pm` was also modified to depend on `$timeNow` (line 1659).

Update: Several more tests in `GatewayQuiz.pm` will use the saved `submitTime` when they would have used a new call to `time()` otherwise.

---

Motivation: I seem to have encountered a case where the `answer.log` file recorded an answer with the time being listed as several seconds after the deadline, but the score itself was apparently not updated. This behavior is not consistent, first because the log record reports the submission as being after the deadline and also in terms of the score not having been updated. The only potential cause I could come up with for the inconsistent records was the fact that in `Problem.pm` the various calls to time based tests (`before`, `after`, and `between`) were not providing a "now" time, so that in potential the time could change enough to cross a boundary during the processing.

~~Note: The time entry at the start of the `answer.log` file lines is **not** dependent on the actual submission time.~~

Updated note: The time entry at the start of the `answer.log` file lines **will now be** the actual submission time. 

Until this PR the  logged `timestamp` value also was not (it was the time just before the call to create the log record). After this PR, the timestamp in the log record and that set in the past answer table will use the determined submitTime and not the time just before the records are created. Hopefully that will prevent records being saved for submissions with a time "just after" a deadline. 

~~However, the formatted time at the beginning of the log record will still be the creation time of the log record and not the submission time.~~

### DIFF
--- a/lib/WeBWorK.pm
+++ b/lib/WeBWorK.pm
@@ -78,7 +78,7 @@ our %SeedCE;
 async sub dispatch {
 	my $controller = shift;
 	my $r          = WeBWorK::Request->new($controller);
-	$r->submitTime(time); # this is Time::HiRes's time, which gives floating point values
+	$r->submitTime(time);    # this is Time::HiRes's time, which gives floating point values
 
 	my $method    = $r->req->method;
 	my $location  = $r->location;
--- a/lib/WeBWorK.pm
+++ b/lib/WeBWorK.pm
@@ -78,6 +78,7 @@ our %SeedCE;
 async sub dispatch {
 	my $controller = shift;
 	my $r          = WeBWorK::Request->new($controller);
+	$r->submitTime(time); # this is Time::HiRes's time, which gives floating point values
 
 	my $method    = $r->req->method;
 	my $location  = $r->location;
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -33,7 +33,7 @@ use WeBWorK::PG;
 use WeBWorK::PG::ImageGenerator;
 use WeBWorK::PG::IO;
 # Use the ContentGenerator formatDateTime, not the version in Utils.
-use WeBWorK::Utils qw(writeLog writeCourseLog encodeAnswers decodeAnswers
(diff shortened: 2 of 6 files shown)


drgrice1:
It seems that a much better way to do this would be to store a timestamp in the request.  This should be set in the `dispatch` method of `lib/WeBWorK.pm`.  Then any of the content generator methods could use that to compare to the time at the start of the request.  An accessor for this could be added in `lib/WeBWorK/Request.pm`.

Also, make sure to run `bin/dev_script/run-perltidy` on this.

taniwallach:
@drgrice1 Thanks. I will try to make those changes soon and then run perl-tidy. Will probably do a force push once there is a new approach.

taniwallach:
@drgrice1 I think I have now implemented it the way you suggested.

I also changed `GateWay.pm` to use the recorded value where it would have used a newly generated `time()` value.

The `version_creation_time` fields set by `GateWay.pm` now share the same value, but in that case a new call to `time()` is made, but just once now.

taniwallach:
I added code so that the answer log will record the request submission time instead of the time when the log record was generated.

@pstaabp I cannot think of any good way to trigger the issue of a deadline passing between time based tests except by hacking the existing code to somehow add a delay between some of the tests. I think it makes more sense to review the code changes and just verify that the code still differentiates between before deadline, after deadline, after answers date. Testing for when a problem is in reduced scoring may also be relevant.

pstaabp (approved):
This looks good and hopefully fixes this edge case.  Couldn't figure out how to test it.  Everything seems to run as it should and the code looks fine.

drgrice1 (approved):
I think this looks good.

drgrice1:
Two approvals.  Time to merge!

Issue #1841, opened 2022-11-26
LTI mass update via Mojo::IOLoop causes server response issues


somiaj:
I have been testing out LTI mass update on a full class with multiple assignments with mojolicious and noticed the following:

+ When using the webwork2-morbo development server, the whole server stops during the update and cannot handle any requests until the update has finished.

+ When using the webwork2.service file which launches via hypnotoad, the very next request after the mass_update is started waits for the update to finish to respond. Though if I send another request it goes through just fine. So it mostly works with hypnotoad, except for that first request after the update has started.

I have tried looking into Mojo::IOLoop to see what is going on, but not finding any useful information as to why that first request after the mass update starts would be waiting for the update to finish, yet any requests after that seem to process just fine.

drgrice1:
With morbo that will happen.  Morbo only runs one worker, and so latter requests have to wait until the first worker finishes.  That is why it is for development only.

With hypnotoad, that should not happen.  I will look into it.

drgrice1:
Testing things with a generic Mojolicious app and a timer that just sleeps for 10 seconds, I see that everything is the same with that.

For morbo, as I said, the process is occupied and until the timer completes no further requests can be processed.  That is to be expected for a development server.

For hypnotoad, when the timer starts it seems that the next request after that from the same client tries to use the same worker.  Sending another request seems to move to another worker, and all is good for that request.  Note that a request from another browser seems to go to another worker anyway and is fine.

In this generic app, I also tested forking the process inside the timer (using Mojo::IOLoop::Subprocess), and that fixes the problem both for morbo and hypnotoad.  Both are able to continue to serve requests even with the same worker.  However, this does not seem to be an option for webwork2 and the LTI mass_update method.  The problem is that the database connection gets interrupted when a new request comes in, and everything explodes.  The mass_update fails with database errors, and the new request dies.  So far worse than the current situation.

For now, we are just going to have to live with how it is I think.  The user that triggers the mass update, will either need to wait, or click on a link a second time.  I suspect that in production with more entropy in the system (more users making requests and such), that this might actually be better as worker processes will be rotated more frequently.

taniwallach:
[...]

PR #1842, opened 2022-11-27
LTI mass_update page to trigger updates.

somiaj:
Adds an LTIUpdate.pm page which allows professors to trigger
  an LTI mass_update for a selected set and/or user. This includes
  moving the logic to determine if a mass_update should run
  from SubmitGrade.pm to ContentGenerator.pm, so mass_update can
  be triggered before the time interval is up. This also extends
  the mass_update function to be able to update only a single
  set or single user as well as all sets and users.

I'm using the current method of Mojo::IOLoop timers to trigger updates, though as discussed in #1841 another method might be helpful for this, but appears this is currently the best option for webwork2.

I would like some way to detect if an update is currently running, so multiple updates cannot be running at once. I have written code for this, but unless there is a way to correctly detect a running update, it won't be functional. One thought I had was to add a setting value to the database that can be used, like a lock file. Though if the update was interrupted and never released the lock setting this would prevent future updates without manually updating the database.

### DIFF
--- a/htdocs/js/apps/LTIUpdate/ltiupdate.js
+++ b/htdocs/js/apps/LTIUpdate/ltiupdate.js
@@ -1,0 +1,33 @@
+(() => {
+	// Store select options in dictionary to rebuild the user/set menus from.
+	const userOptions = {};
+	const setOptions = {};
+	const userMenu = document.getElementById('updateUserID');
+	const setMenu = document.getElementById('updateSetID');
+	for (let i = 1; i < userMenu.length; i++) {
+		userOptions[userMenu[i].value] = userMenu[i];
+	}
+	for (let i = 1; i < setMenu.length; i++) {
+		setOptions[setMenu[i].value] = setMenu[i];
+	}
+
+	// Update user and set drop down menus to only include valid user/set combinations.
+	document.getElementById('updateUserID')?.addEventListener('change', (e) => {
+		const setList = e.target.options[e.target.selectedIndex].dataset.sets.split(':');
+		const selectedSet = setMenu.value;
+		while (setMenu.length > 1) setMenu.lastChild.remove();
+		setList.forEach((set) => {
+			setMenu.append(setOptions[set]);
+		});
+		setMenu.value = selectedSet;
+	});
+	document.getElementById('updateSetID')?.addEventListener('change', (e) => {
+		const userList = e.target.options[e.target.selectedIndex].dataset.users.split(':');
+		const selectedUser = userMenu.value;
+		while (userMenu.length > 1) userMenu.lastChild.remove();
+		userList.forEach((user) => {
+			userMenu.append(userOptions[user]);
+		});
+		userMenu.value = selectedUser;
+	});
+})();
--- a/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
@@ -32,6 +32,7 @@ use Net::OAuth;
 use HTTP::Request;
 use LWP::UserAgent;
 use UUID::Tiny ':std';
+use Mojo::IOLoop;
 
 use Digest::SHA qw(sha1_base64);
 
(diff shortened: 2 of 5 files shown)


drgrice1:
Instead of moving the logic into ContentGenerator.pm, move the timer loop into SubmitGrades.pm.  That is much cleaner.

somiaj:
@drgrice1 I still think the logic is better in ContentGenerator.pm, because that is the point that it should check if the time interval has passed or not, but I moved the Mojo::IOLoop->timer to SubmitGrade.pm. Moving the logic from ContentGenerator.pm would require two different update methods, `all` and `force_all` for example, to allow the new LTIUpdate.pm page to start a full update before the timer has elapsed. Though I guess having this second method would be useful if this function gets called from some future feature.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -158,20 +157,12 @@ async sub go {
...
-			}
-		);
-
-		Mojo::IOLoop->start unless Mojo::IOLoop->is_running;
+		my $lastUpdate     = $r->db->getSettingValue('LTILastUpdate') || 0;
+		my $updateInterval = $ce->{LTIMassUpdateInterval}             || -1;    # Never update
+		if ($updateInterval != -1 && time - $lastUpdate > $updateInterval) {
+			my $grader = WeBWorK::Authen::LTIAdvanced::SubmitGrade->new($r);
+			$grader->mass_update('all');
+		}
```
drgrice1, line 165: What I mean is to move all of this to SubmitGrades.pm.  This would become simply:
```suggestion
			my $grader = WeBWorK::Authen::LTIAdvanced::SubmitGrade->new($r);
			$grader->mass_update('all');
```
The removed logic here would be in the mass_update method.  If it is not time for a mass update the method would just return.

One other note:  Change `ref($r->{db} // '')` to `ref($r->db)`.  If `$r->db` is undefined, then `ref` will return the empty string.  There is no need for the `// ''` fallback.  I know that was that way, but it was wrong.

somiaj: Thanks for the comments, updated.

taniwallach:
@somiaj - I think that this is a great idea.

> I would like some way to detect if an update is currently running, so multiple updates cannot be running at once. I have written code for this, but unless there is a way to correctly detect a running update, it won't be functional. One thought I had was to add a setting value to the database that can be used, like a lock file. Though if the update was interrupted and never released the lock setting this would prevent future updates without manually updating the database.

I'm not sure how feasible it would be to be absolutely sure if there was a running/pending update. In particular, if the development team manages to move the updates into something run in the background outside the main server, doing so would be difficult. The lock record idea seems feasible, and if the lock included an expiration time the problem of requiring manual modification of the database would be avoided or greatly reduced. The default lock duration could be a system default, which could be overridden at the course level. Lock expiration can be handled whenever any check of the locks in the given course was needed.

Allowing the instructor to request an update for a different set or a different user when there is a pending update of that type seems to be something which is desired and which should not be much of a problem, particularly if the tasks  were queued via a job queue and not executed immediately. However, pending updates of those types would seem to justify preventing the request for an "all" (or auto) update to be run at the same time.

Given that there would now be several types of "mass updates" I think you might want different types of locks. 
Maybe for all/auto, set (locking only for that set), and user (locking only for that user).

In order to make it possible to have mass updates run outside of a request, I think it is probably necessary to change the mass_update subroutine (possibly all of `SubmitGrade.pm`) so that it did not depend on a request (`$r`) to provide the course environment and the database "handle" and instead depended only on having a course environment `$ce` as a direct parameter, and then directly requested its "own" Database "handle": `my $db = new WeBWorK::DB($ce->{dbLayout});`.

As far as I can tell, the only part of `SubmitGrade.pm` which really depends on there being a request available is `update_sourcedid` (which updates the database with the LTI parameters needed to later do grade pass-back based on the parameters LTI provides at "connection" time) so maybe that the request can passed in as a parameter when `LTIAdvanced.pm` calls it.

somiaj:
@taniwallach I had also thought of adding a timestamp to the lock setting to auto remove it if for some reason the process didn't manually remove it when done. I don't think there should be a separate lock for 'all' and 'auto'. My thought is just to not have multiple log running jobs that will be doing the same thing at the same time. Though I hadn't considered the idea of still allowing single set/user updates, which is probably fine, as those should be much shorter, even if a full 'all' process is already running. Maybe two locks could be enough, one to lock any full all/auto update, and one to lock shorter updates for single sets/users, this way someone can't get impatient and start a bunch of updates, taking up the time of the workers.

Though this might be something worth adding later. @drgrice1 mentioned in #1841 of implementing a new method to run this job, so it might be worth merging this (pending approvals and any issues anyone notices), so that the new method can be built upon having this lti mass_update trigger page.

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
```diff
@@ -0,0 +1,223 @@
...
+			? CGI::popup_menu({
+				id      => 'updateSetID',
+				name    => 'updateSetID',
+				value   => [ 'All Sets', @sets ],
+				default => 'All Sets',
+				class   => 'form-select mb-3',
+				labels  => {
+					'All Sets' => $r->maketext('All Sets'),
+					map { $_ => format_set_name_display($_) } @sets
+				}
```
drgrice1, line 184: These popup menus should really each be wrapped in a pair of divs (the usual row and column divs).  Not only is this how Bootstrap demonstrates the usage of the `form-select` class, but this makes it possible to do other things.  For example, these popup menus need labels.  Something like "For which users:" and "For which sets:" (or simply "Users:" and "Sets:" or something).  To make that work you need all of the proper markup.  In addition, I don't like that they stretch the entire width of the window.  With the row/column divs you can add a `col-n` class to control the width (this will be needed with the label anyway).  For a simple example of what is needed here see line 615-632 of CourseAdmin.pm.

drgrice1 (commented):
This generally looks good aside from the attempt at checking to see that a grade update is already in progress.  I suppose that can be let slide for now though until that is sorted out.

There are a few changes that are needed though.  The line 541 change to SubmitGrade.pm is clearly essential!  Then this can be merged.

I have done some more work on making the mass update not block on the main thread, and I am rather certain the only way that is going to work is to make a Minion worker queue.  There are a few annoying things going on with the mysql and sqlite backends in the upstream code though, but those can be worked around.  There is an chance that this could be executed in an IOLoop subprocess, but it would take pretty much a duplication of the methods in the SubmitGrade.pm file.  I don't think what I have in mind would be able to share the code that submits a single grade for a student withthe mass update methods.  Mainly because it would require a separate database connection and course environment and can not be connected to a content generator.  @taniwallach:  Note the Mojo::UserAgent doesn't solve the problem even if async and non-blocking.  It turns out it is the database connection that is doing to blocking.

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
```diff
@@ -0,0 +1,223 @@
...
+	my $ce   = $r->ce;
+	my $user = $r->param('user');
+
+	# Check permissions
+	return unless $r->authz->hasPermissions($user, 'score_sets');
+	return unless $ce->{LTIGradeMode};
+
+	# Want to be able to determine if an update is currently running, and not allow
+	# a new update to start in that case, but haven't figured out a good way to do this.
+	$self->{isUpdating} = 0;    #Mojo::IOLoop->is_running;
```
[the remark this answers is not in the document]
somiaj: Thanks. That was what I noticed, Mojo::IOLoop->is_running was always true. I can strip out this code (or just leave it in as a starting point for a future change which will have a way to better check if an update is running).

drgrice1:
There is one thing that I think you have done that I don't like in this pull request.  That is the positioning of the `LTI Grade Update` link in the nav.  I would like that to be above the `Course Configuration` link.  I like `Course Configuration` to be essentially last in the list (other than the help links).  I also don't like that `Archive Course` shows up after the help links when the file manager is open (not your fault of course).

Others may disagree.  Please weigh in with opinions!

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
```diff
@@ -533,55 +534,90 @@ EOS
...
+	# Start update loop.
+	Mojo::IOLoop->timer(
+		1 => sub {
+			# Catch exceptions generated during the sending process.
+			eval {
+				# Determine what needs to be updated.
+				my %updateUsers;
+				if ($update eq 'all') {
+					if ($ce->{LTIGradeMode} eq 'course') {
+						%updateUsers = map { $_ => [] } $db->listUsers;
```
drgrice1, line 587: Instead of using a reference to an empty array to signify a 'course' grade mode update for the user, it would be better to use something like the string 'course' (or anything that is not a reference to an array).  Then the `for` loop that does the updating below would just check the `ref` and ensure that it is equal to 'ARRAY'.  One reason that a reference to an empty array ref is a bad idea is that it might be possible that someone could call this method with a user that is not assigned to any sets when the course grade mode is homework.  In that case this would set `$updateUsers{$name}` to be the empty array reference (below on line 597), and then the `for` loop below would attempt to upgrade the user's course grade, instead of the homework grade.

In testing this, I see that this does cause problems.  From the LTI update page select a user that is not assigned to a set and select that set in the dropdown.  You will see a warning in the output when running morbo.

This also shows another problem with this pull request.  The dropdown should only show the sets that are assigned to the selected user.   That will take a little work to implement, but can be done.  Note that fixing the dropdown issue won't fix the above issue, because someone could still submit a request via other means that could still contain the same invalid data.

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
```diff
@@ -0,0 +1,242 @@
...
+
+	my $gradeMode      = $ce->{LTIGradeMode};
+	my $lastUpdate     = $db->getSettingValue('LTILastUpdate') || 0;
+	my $updateInterval = $ce->{LTIMassUpdateInterval}          || -1;
+
+	# Print status table
+	print CGI::div(
+		{ class => 'table-responsive' },
+		CGI::table(
+			{ class => 'table table-bordered', style => 'width: auto;' },
```
drgrice1, line 124: Please remove the inline `style` attribute.  Instead use the Bootstrap `w-auto` class.

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
```diff
@@ -0,0 +1,242 @@
...
+							map { $_ => $_ } @users
+						}
+					})
+				)
+			),
+			$gradeMode eq 'homework'
+			? CGI::div(
+				{ class => 'row mb-3' },
+				CGI::label(
+					{ for => 'updateSetID', class => 'col-auto col-form-lable fw-bold' },
```
drgrice1, line 187: `col-form-lable` should be `col-form-label`.  This causes the misalignment of the label.

drgrice1 (commented):
This is looking decent, but there are some things that need to be done yet.

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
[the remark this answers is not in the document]
drgrice1: I think you should go ahead and remove the `isRunning` code.  It isn't going to work in this way.  When I add the Minion worker queue this won't be needed at all.  Multiple jobs will be able to be queued.  The queue can be made to run them sequentially.  Most likely I will add a table showing queued jobs and their status to the new LTI Update page.

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
somiaj: @drgrice1 I have updated this, and have added some javascript functions that from my tests update the menus to only allow valid set/user selections.

somiaj: @drgrice1 I have updated how the javascripts works. First it appears that there is no cross-browser way to hide items in a popup menu (select), so hiding the options won't be possible without rebuilding the menu each time an item is selected, but another option was to disable the items. This keeps them in the menu, but greys them out and makes them unselectable.

Using this approach, the updated code now adds a dataset to each option that lists the valid sets/users assigned to the selected item. Then updates the corresponding menu to disable all the invalid options, making it so only the sets a user is assigned to (or the users a set is assigned to) are selectable.

drgrice1: Using `display:none` css styling works reliably cross browser.  We are already using that for achievement rewards, as I mentioned.

htdocs/js/apps/LTIUpdate/ltiupdate.js
```diff
@@ -0,0 +1,18 @@
+(() => {
+	// Update user and set drop down menus to only include valid user/set combinations.
+	document.getElementById('updateUserID')?.addEventListener('change', (e) => {
+		const setMenu = document.getElementById('updateSetID');
+		const setList = e.target.options[e.target.selectedIndex].dataset.sets.split(':');
+		for (var i = 1; i < setMenu.length; i++) {
+			setMenu[i].disabled = !setList.includes(setMenu[i].value);
```
drgrice1, line 7: If you change this to
```
			if (setList.includes(setMenu[i].value)) setMenu[i].classList.remove('d-none');
			else setMenu[i].classList.add('d-none')
```
and delete the `change` event on the `updateSetID` element below, that is all that is needed.

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
somiaj: I found multiple posts which stated that `display:none` it isn't supported by safari for options in a select. I tried your suggestion below, and it worked on firefox, but didn't work on safari on my ipad. So for my tests it appears the posts I found were correct that safari doesn't support hiding select options this way (and maybe the achievement rewards need to be updated because of this).

For this particular case, I like the approach of disabling the items vs hiding them better, that way I do know that all the sets and users are listed, but can only select valid pairs. So my opinion is to just disable the invalid sets/users vs hide them.

htdocs/js/apps/LTIUpdate/ltiupdate.js
somiaj: I don't quite understand your logic as to why the `updateSetID` event needs to be deleted. Both should be needed, the `updateUserID` event updates the Set list so only sets assigned to the selected user can be selected, while the `updateSetID` event updates the user list so only users assigned to the selected set can be selected.

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
drgrice1: I think we should drop support for Safari.  Safari is the new Internet Explorer!  Sad but so true.

drgrice1: I don't like the disabled approach.  It is ugly, and this is not supposed to be an information page as to what sets are available.  You can still get that here by selecting all users.

htdocs/js/apps/LTIUpdate/ltiupdate.js
drgrice1: That was based on being able to hide sets.  In that case, you only need to update the set list.  When a user is selected there are different sets assigned to that user.  If you change the set selection, that user is still selected.  If you now go change the user to a set that is not assigned that set will be hidden and the selection will change to a set that is no longer hidden.  So there is no need to deal with both.

somiaj: Ahh okay, I see the logic now, but this doesn't seem to quite work, just tested. First select a set, then select a user that wasn't assigned to that set. It does hide the set from the list, but the hidden item is still selected, and thus one could submit an invalid combo this way.

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
somiaj: I do agree that safari support is a pain, though I don't know if we can just drop support for it, as users will still be using Safari. So in this case a safari user could submit invalid user/set combos. Though there is a way around this, you can first detatch the options, then rebuild the menu each time from the list of detatched options.

htdocs/js/apps/LTIUpdate/ltiupdate.js
drgrice1: Yeah, testing this on the achievements page, I see that is an issue.  Okay, so you will need to do both lists.

Note that was not actually my design on the achievements page.  I moved the javascript from begin inline to a proper javascript file, but left the same basic design.  It has been that way for a long time.

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
drgrice1: Removing the options and re-adding is probably the better way to go.  It is what a reactive framework like Vue would do anyway.

somiaj: Updated so it rebuilds the menu each time instead of hiding the menu options, to play nicely with Safari.

htdocs/js/apps/LTIUpdate/ltiupdate.js
```diff
@@ -0,0 +1,33 @@
...
+	const userOptions = {};
+	const setOptions = {};
+	const userMenu = document.getElementById('updateUserID');
+	const setMenu = document.getElementById('updateSetID');
+	for (var i = 1; i < userMenu.length; i++) {
+		userOptions[userMenu[i].value] = userMenu[i];
+	}
+	for (var i = 1; i < setMenu.length; i++) {
+		setOptions[setMenu[i].value] = setMenu[i];
+	}
```
drgrice1, line 12: Lines 7 and 10 should be `for (let i = 1; ...) {`.  There is no `var`!

drgrice1 (approved):
One minor code issue, but I think this is basically ready to go other than that.

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
```diff
@@ -0,0 +1,239 @@
...
+
+	return CGI::div({ class => 'alert alert-danger p-1' }, $r->maketext('You are not authorized to update lti scores'))
+		unless $authz->hasPermissions($user, 'score_sets');
+
+	return CGI::div({ class => 'alert alert-danger p-1' },
+		$r->maketext('LTI grade passback is not enabled for this course'))
+		unless $ce->{LTIGradeMode};
+
+	# Update message.
+	print CGI::div({ class => 'alert alert-warning p-1' }, $self->{updateMessage}) if defined($self->{updateMessage});
```
drgrice1, line 99: One minor quibble here.  I don't like that this is an `alert-warning`.  That indicates that something is not quite right, when in fact everything has gone as it should.  This should be an `alert-success`.

drgrice1:
I just noticed a problem with this.  When LTIGradeMode is "course" and a single user is selected on the new "LTI Grade Update" page, the grade update does not occur.  The problem is the logic in lines 51 - 70 of LITUpdate.pm.  That never checks the case that the userID is set and LTIGradeMode is "course".

Issue #1844, opened 2022-11-29
Shibboleth.pm - 025010a should be reverted


jmacdone:
This change from https://github.com/openwebwork/webwork2/commit/025010a5b077b201eda13451877979ebe8875d31
```
- my $user_id = $ENV{$ce->{shibboleth}{mapping}{user_id}};
+ my $user_id = $ce->{shibboleth}{mapping}{user_id};
```
causes `$user_id` to be assigned name of the environment variable, rather than the desired value.

For example, from v2.16:
```
[Tue Nov 29 09:25:38.624653 2022] WeBWorK::Authen::Shibboleth::get_credentials: user_id is user_id
```
and after (manually) reverting, we get
```
[Tue Nov 29 10:32:38.20267 2022] WeBWorK::Authen::Shibboleth::get_credentials: user_id is jmacdone
```

jmacdone:
FWIW, in the above example, the sysadmin is extracting what is typically referred to as `uid` instead as `user_id` and had `localOverrides.conf` set accordingly.

    jmacdone@webwork2:/etc/shibboleth$ grep user_id attribute-map.xml
    <Attribute name="urn:oid:0.9.2342.19200300.100.1.1"  id="user_id" />

and

    jmacdone@webwork2:/opt/webwork/webwork2/conf$ grep "'user_id'" localOverrides.conf
    $shibboleth{mapping}{user_id} = 'user_id';

taniwallach:
@jmacdone - Thanks for reporting this. It certainly looks like a serious bug, and something to fix for WW 2.17 using a hot fix. I just created https://github.com/openwebwork/webwork2/pull/1845

@xcompass @mgage @drdrew42 @drgrice1 Could one of you take a look?

The problematic changes got in to the system via https://github.com/openwebwork/webwork2/pull/1056.

The Apache `mod_shib` documentation https://wiki.sunet.se/display/SWAMID/1.2+Configuring+Apache+Web+Server+to+use+Shibboleth  indicates that the data provided by the Identity Provider is intended to be accessed via "HTTP environment variables", namely via `$ENV` as was done, and you want to fix in one place.

[...]

Issue #1039, opened 2019-10-18
Use of ENV is not needed for getting configuration from localOverride.conf file


SumitLubal:
https://github.com/openwebwork/webwork2/blob/6df44736610e457c1d30ac133e2890451258b49d/lib/WeBWorK/Authen/Shibboleth.pm#L78

SumitLubal:
Here is the PR to fix this issue
https://github.com/openwebwork/webwork2/pull/1044

PR #1845, opened 2022-11-30
Shibboleth fix - hotfix

taniwallach:
Intended to fix https://github.com/openwebwork/webwork2/issues/1844 for WeBWorK 2.17.

Note the second commit is just from running perltidy on the file, so review should focus on the changes in the first commit.

### DIFF
--- a/lib/WeBWorK/Authen/Shibboleth.pm
+++ b/lib/WeBWorK/Authen/Shibboleth.pm
@@ -75,9 +75,9 @@ sub get_credentials {
 		return $self->SUPER::get_credentials( @_ );
 	}
 
-	if ( defined ($ce->{shibboleth}{session_header}) && defined( $ce->{shibboleth}{mapping}{user_id} ) ) {
+	if ( defined ($ENV{$ce->{shibboleth}{session_header}}) && defined( $ENV{$ce->{shibboleth}{mapping}{user_id}} ) ) {
 		debug('Got shib header and user_id');
-		my $user_id = $ce->{shibboleth}{mapping}{user_id};
+		my $user_id = $ENV{$ce->{shibboleth}{mapping}{user_id}};
 		if ( defined ($ce->{shibboleth}{hash_user_id_method}) &&
 		     $ce->{shibboleth}{hash_user_id_method} ne "none" &&
 		     $ce->{shibboleth}{hash_user_id_method} ne "" ) {


drgrice1 (changes requested):
I assume you are doing this to address issue #1844.  I think that this needs to be investigated more carefully.  This was changed bu @SumitLubal  in 2019 for specifically the opposite reason.  I suspect that the usage of the Shibboleth authentication varies and this is actually an issue with different institutions using incompatible setups.  We can't keep switching this back and forth to satisfy the institution that is currently paying attention.  The setup of Shibboleth needs proper documentation.  I highly suspect that this is in fact the wrong way for this change to be made.

jmacdone:
Looking at #1039, if the intent was to change the conditional from "is shibboleth working" to "is shibboleth configured" then most of #1044 was correct, but this extra $ENV removal on line 80 was an oversight and introduces a breaking bug.  https://github.com/openwebwork/webwork2/pull/1044/commits/189ad0a15fd045993a0190feaa8be21d318c8b5d

jmacdone:
Or, more explicitly, instead of a full revert, perhaps just this will address both #1844 and whatever intentions/concerns @SumitLubal had.
```diff
--- a/lib/WeBWorK/Authen/Shibboleth.pm
+++ b/lib/WeBWorK/Authen/Shibboleth.pm
@@ -77,7 +77,7 @@ sub get_credentials {

        if ( defined ($ce->{shibboleth}{session_header}) && defined( $ce->{shibboleth}{mapping}{user_id} ) ) {
                debug('Got shib header and user_id');
-               my $user_id = $ce->{shibboleth}{mapping}{user_id};
+               my $user_id = $ENV{$ce->{shibboleth}{mapping}{user_id}};
                if ( defined ($ce->{shibboleth}{hash_user_id_method}) &&
                     $ce->{shibboleth}{hash_user_id_method} ne "none" &&
                     $ce->{shibboleth}{hash_user_id_method} ne "" ) {
```

taniwallach:
@jmacdone Can you confirm that Shibboleth authentication is working properly for you in WW 2.16 or WW 2.17 after you make the patch you mentioned.

@drgrice1 I don't think the changes from 2019 make sense. @mgage merged https://github.com/openwebwork/webwork2/pull/1056 into develop, but wrote that he could not test it.

Note: The thread at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8176 indicates that Shibboleth was working properly in WW 2.14, and that Kevin de la Cueva worked with some people from his institution (@jmacdone ?) to debug the issue and they also made some updates to the documentation.

The documentation at https://webwork.maa.org/wiki/External_(Shibboleth)_Authentication may be partial, but is all we have to go by unless someone who uses Shibboleth updates it more. @jmacdone made some updates recently: https://webwork.maa.org/mediawiki_new/index.php?title=External_%28Shibboleth%29_Authentication&type=revision&diff=23211&oldid=12816

Reading the code raises the question of how setting the `$user_id` value to be used by WeBWorK to a static value set in the WeBWorK configuration (`$ce->{shibboleth}{mapping}{user_id}`) is supposed to make the authentication work. That issue is the core of what was reported in https://github.com/openwebwork/webwork2/issues/1844 . I don't see how an incompatible Shibboleth configuration would get that change to work, as `mod_shib` should not have any means of writing into some value inside the course environment to provide a valid user_id value for whatever user is trying to authenticate.

As I understand the debug line between the two lines with changes, it pretty clearly indicates that the test was intended to be that the "data" arrived and not that the configuration was made.

If checking that the relevant configuration settings were made is helpful, then those checks should be in addition to the test that `mod_shib` actually sent the expected data and not instead of it. Testing that the relevant configuration was made on each authentication and not that `mod_shib` sent the data expected seems to me to be relatively pointless in production, but having additional such tests is not harmful. However, I suspect that if the configuration settings are not made, then the tests should just fail anyway.

---

The different between institutions mentioned seems only to be in terms of the "field name" to which `attribute-map.xml` for Shibboleth is extracting the data provided by the Identity Provider. The need to extract the relevant data and to set the WW configuration to find it in the correct "field name" seems reasonably well documented in the Wiki page after the recent changes by @jmacdone.
```
# define mapping between shib and webwork
# use the id value of one of attributes being provided/extracted in shibboleth/attribute-map.xml
# eppn, uid, persistant-id are common choices
#$shibboleth{mapping}{user_id} = 'eppn'; 
$shibboleth{mapping}{user_id} = 'uid';
```

jmacdone:
Confirming the patch from https://github.com/openwebwork/webwork2/pull/1845#issuecomment-1332843401 is working in 2.16 (or at least that's what `VERSION` says).  I didn't touch the conditional.

And, yes, Kevin de la Cueva and I are colleagues.  He's sysadmin for webwork and I'm lead for identity management.   The forum post and #1844 are related.

drgrice1:
@taniwallach:. It does not make sense to get the user id from the course environment.  Not does it make sense to get the user id from a system environment variable.  

This should be put on hold until the documentation is updated and proper investigation into the workings of Shibboleth authentication is done.  Clearly the documentation is quite out of date as it refers to running WeBWorK in the root HTML directory which has not worked correctly for years (although that is fixed in the develop branch).

jmacdone:
>Nor does it make sense to get the user id from a system environment variable.

That's how mod_shib does its thing, though.  It presents SAML assertion attributes as environment variables.  (They also support injecting request headers, but that's not recommended)
https://shibboleth.atlassian.net/wiki/spaces/SP3/pages/2065335257/AttributeAccess

drgrice1:
I'm that case mod_shib is an unsecure hack and support for it should be dropped. Passing login information via environment variable is insane.

jmacdone:
IDK, @scantor is pretty fanatical about security.

>The safest mechanism, and the default for servers that allow for it, is the use of server variables. [...] that cannot be manipulated in any way from outside the web server. Specifically, the client has no say in them.

drgrice1:
@jmacdone:  I don't know who that is.  Nor do I care.  Quoting some obscure individual doesn't make any real difference here on the incorrect usage of environment variables.  It also doesn't fix the fact that Shibboleth authentication has been hacked into WeBWorK with no proper documentation, and is an obscure authentication method used by very few institutions.  Making it a prime candidate for removal from the code base.  If you or someone is willing to step in and do the footwork to update the documentation and the code, and provide methods by which this can be tested, then there is hope for the module.

taniwallach:
@jmacdone Thanks for your comments and for the issue report. Without someone reporting that Shibboleth authentication was broken - we simply would not have known, as the current development team does not have the ability to test all the existing features before each release. Please understand that the development team are all volunteers and @drgrice1 in particular has been making very significant contributions to improving WeBWorK. At times he can get a bit upset at all the hacked up parts of the system and the history of merging in code from just about anyone. His feedback has been very valuable to me on work I have done, and I certainly appreciate working with him even if at times I get a tongue lashing. I hope that you and CSUSB might be willing to help keep Shibboleth authentication viable for the next release of WeBWorK.

> > Nor does it make sense to get the user id from a system environment variable.
> 
> That's how mod_shib does its thing, though. It presents SAML assertion attributes as environment variables. (They also support injecting request headers, but that's not recommended) https://shibboleth.atlassian.net/wiki/spaces/SP3/pages/2065335257/AttributeAccess

@drgrice1: Sorry, but my response is also a bit of a rant, and certainly very long. In short - I still think this is a justified hot fix, that you were missing some of the fine details and did not fully consider the background, and that any significant changes/improvements/redesign for `Shibboleth.pm` should be worked on for develop.
  1. In this case - I think you are being a bit too antagonistic.
     - @jmacdone proposed a bug fix to an existing feature (working in WW 2.15 AFAIK but broken in 2.16 and 2.17) which breaks functionality on which his institution depends.
     - The development team accidentally merged code which caused this bug.
     - I took his bug report and proposed fix and set up this hot fix PR.
     - @jmacdone has confirmed that the patch fixes the bug in production.
     - Accepting the patch will allow institutions which want to use Shibboleth to use the current release of WeBWorK without needing to locally patch the bug.
     - All in all, this seems a totally valid case where a hot fix is appropriate.
     - Can we consider addressing the big issues in the future and focus on the small patch at present?
  2. Let's please make @jmacdone on other non-developers feel appreciated for being willing to participate in reporting bugs, proposing fixed and taking part in the discussions.
  3. The `%ENV` hash in `mod_perl` provides access to "regular" environment variables and also to "server variables" set by Apache and Apache modules. `mod_shib` is providing data via "server variables". (See below.) Understanding that distinction seems important, and hopefully will make the code seem at least somewhat less evil than you thought it was.
  4. SAML via Shibboleth seems to be the only modern SSO option available for WeBWorK at present. As such, it does not seem that wise to leave it broken in the current release when only a small fix is needed. It would also be nice if the feature remained feasible for WeBWorK 2.18, but getting there will certainly require some help from institutions with a vested interest in this feature. 
  5. We accepted https://github.com/openwebwork/webwork2/pull/1734 to improve CAS authentication for 2.17 even though it is apparently only used now at UC Berkeley, while Shibboleth certainly has a larger potential audience. Thus, being "an obscure authentication method used by very few institutions" is not an absolute reason to block a PR.
  6. It is probably true that Shibboleth authentication for WeBWorK is apparently not very widely used (no one who upgraded to WW 2.16 or WW 2.17 reported the bug until now). The low popularity is probably because many institutions can make use of LTI which also can handle account creation. However, we really have no idea how many institutions are using Shibboleth authentication with WeBWorK. Here are some that apparently do (or did in the past):
     - UBC: @xcompass seems to still be using it, in addition to LTI from Canvas.
     - NCSU: @jmacdone (https://delta.ncsu.edu/about-delta/people/name/jonathan-champ/) did some cleanup to the code, and they seem to be using it there. Picking a course at https://webwork.wolfware.ncsu.edu/webwork2/ brings up a Shibboleth login page.
     - CSUSB (but on WW 2.13) https://webwork.csusb.edu/webwork2  (Kevin de la Cueva is listed there https://www.csusb.edu/profile/kevin.delacueva but apparently is now involved with the NCSU webwork server)
     - University of Galway: Niall Madden posted (2021) about using it in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4881
     - Ted Boggs also posted (2018) about using it in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4329
  7. You are welcome to read more about SAML and Shibboleth if you want to spend the time on it now, but my impression is that Shibboleth is pretty well accepted, as are the mechanisms of using it with Apache, nginx and other web servers. 
  8. It seems to me that @xcompass followed what were then accepted practices for WW when he coded the module, and later fixed some serious bugs before the code became stable enough and dormant. As such that fixing the existing bug which accidentally got merged in 2019 is not a terrible thing to do.
  9. The existing approach to Shibboleth authentication will certainly **not work** for WeBWorK 2.18 since that will be running in Mojolicious and not directly in Apache. There will not be any access to `%ENV` to get attributes provided by `mod_shib`.
     - The authentication mechanism will apparently need to be split between what happens in a reverse proxy and what gets done in WeBWorK.
     - The mechanisms to do that will need to be designed. implemented and tested.
     - As such, we will need to consider whether we can support Shibboleth for 2.18, and if so how.
     - Hopefully some of the institutions which use Shibboleth can help us with the necessary work and testing for 2.18.
     - If not, then 2.18 may be forced to deprecate the Shibboleth support, until some interested parties take an interest in helping to get it working again.
  10. The `lib/WeBWorK/Authen/Cosign.pm` authentication module also depends on accessing something from `$ENV` (and old notes indicate that it was the model on which Shibboleth.pm was based.) Hence, attacking the current state of `lib/WeBWorK/Authen/Shibboleth.pm` for doing that is something to we can look into fixing in the future, but should not be using to challenge @jmacdone's attempt to get the feature working again. In any case, that mechanism is certainly not relevant for develop/WW 2.18 so will break both these authentication modules.
  11. Like many parts of WeBWorK - the documentation about Shibboleth authentication is not as detailed or well maintained as it could be. As I pointed out above @jmacdone already did made some improvements to the Wiki page in order to somewhat improve the documentation. I'm not sure that much additional work on that documentation now is a really a worthwhile investment of someones time and effort, as anything done will almost certainly become pretty irrelevant for WW 2.18 or at least require very significant revision.

---

There is a pretty nice explanation of the whole SAML authentication process at https://hungreyweasel.wordpress.com/2020/05/10/saml-ssos-shibboleth-sp-reverse-proxy/

WeBWorK would be the Service Provider, and some remote system will be the Identity Provider.

---

About the manner in which Shibboleth integrates with web-servers, that is apparently the accepted practice recommended by a pretty big consortium which is quite interested in security. Their documentation does try to clear emphasize some of the potential problems, in particular the dangers of passing authentication response information to a proxied service via request headers and how to suitable protect that approach.

Apache differentiates between system environment variables and server environment variables. See: https://httpd.apache.org/docs/2.4/env.html  so at least potentially the Shibboleth provided data is somewhat more protected than a regular environment variable. From that I understand from the documentation, when an external CGI script or the like would need to access server environment variables they would need to be converted to regular environment variables. However, `mod_perl` is running inside Apache so should have direct access to them, based on `mod_perl` putting the data into the `%ENV` hash available to the Perl code.

It looks like it should be possible to access the same data via something like `$r->subprocess_env( $key )` from 
according to 
https://perl.apache.org/docs/1.0/api/Apache.html#_r_E_gt_subprocess_env___key____value___
and if you feel that is significantly more secure - we could see if that works.

---

Looking ahead with regards to develop/WW 2.18, there is a post which explains how to set up nginx acting as a reverse proxy to provide the Shibboleth authentication response data to a back-end app via **request headers** (as a reverse proxy cannot really use any more secure mechanism) at https://tcg.stanford.edu/?p=131 . There are also explanations at https://github.com/nginx-shib/nginx-http-shibboleth#configuration and that pretty clearly indicates that reverse proxy setup will need to use the **less secure** request header approach to passing data (and thus first clear ones which matter) as the mechanism to pass the data to the back-end. 

Note also that newer material indicates that more recent versions of nginx support dynamic module loading, which may allow avoiding a need to locally compile nginx to use it as a reverse proxy handling the Shibboleth authentication. https://github.com/nginx-shib/nginx-http-shibboleth/


---

For the future, we could possibly consider replacing Shibboleth with an alternative SAML authentication mechanism, which would be a more generic solution (supporting multiple Identity Providers, not just Shibboleth.

  1. SAML is apparently quite a complicated protocol and has been evolving over time. The Shibboleth consortium have far more interest in maintaining a working free/open implementation of a SAML authentication system than we do.
  2. There is a suitable Perl module https://metacpan.org/pod/Net::SAML2 but we would need to code the full authentication mechanism using the features provided. There is a tutorial at https://github.com/perl-net-saml2/perl-Net-SAML2/blob/master/TUTORIAL.md
  3. The Mojo::SAML module does not seem to be maintained and is marked as highly experimental, so probably not something to consider.

scantor:
> Note that it is not actually less secure to use the request. Where do you thing mod_shib gets the user_id from?
> That doesn't just magically appear in the environment variable on the server. It ultimately comes from the request.

If you're going to insult me, and my work, you probably want to avoid being wrong. The Shibboleth SP module runs inside Apache and it most certainly *does* "magically" (not really, but your words, not mine) put the user identity and anything else into the request, the same as most any other complex Apache authentication module does. It sure as hell does not get it "from the request", ultimately or otherwise.

If you don't understand why using headers vs. system/server variables is a fundamentally different risk posture for an application, you should do some more research.

Anyone who's not a giant ass is welcome to ask anything they'd like on the Shibboleth users mailing list if there are questions about how it works in the context of your project.

drgrice1:
@scantor:  I did not insult your work.  I still don't know who you are.  Someone involved with Shibboleth development apparently.  However, you still are incorrect.  There is an initial request from somewhere that contains the users identifying information.  I did not say headers, I said request.  That is ultimately where the identifying information comes from.  There never is magic.  Magic doesn't exist.  Perhaps you should do some more research.

scantor:
@drgrice1: Yes, you did. I don't need to research how my own code works, and I don't need to really demonstrate to anybody else in this thread that you're being an ass. To the others that need any help, you know where to find it.

drgrice1:
Your name should never have been mentioned in this conversation.  I really did not mean to offend you, and I would appreciate it if you did not jump into a conversation after someone offhandedly mentioning your name, and taking offense for no real reason.  Furthermore injecting insults is just not appropriate.

taniwallach:
> I am fine with patching this in for now after you change this as @jmacdone suggested to only add the `ENV` usage back on the user_id.

@drgrice1 Why? What benefit do you see to testing only on whether the configuration was made?

AFAIK The original code and that proposed here will also fail the test when either the configuration is not made (as it will be essentially trying to access `$ENV{undef}`) or when `mod_shib` failed to provide the expected data. If we want to be more proactive, we could test all 4 conditions.

We do not want to try to set `$user_id` (which will be the identity of the user who gets connected if the authentication succeeds) using `$ENV{ $ce->{shibboleth}{mapping}{user_id} }` if Shibboleth did not send WW the value to use, so why enter the block of code without that value available?

As such, do not think that testing only that the configuration was made is the right thing to do. It is far to clear to me why @SumitLubal requested/submitted such a change.

> Note that it is not actually less secure to use the request. Where do you thing `mod_shib` gets the user_id from? That doesn't just magically appear in the environment variable on the server. It ultimately comes from the request.

If comes from data which an identity provider sent to the client and then on to our server in SAML assertions which were cryptographically signed and that signature gets verified by `mod_shib`. All the work to make the SAML handshake safe and reliable is handled by `mod_shib` and the identity provider. That is about as secure as SSO can be. `mod_shib` needs to provide data to back-end code which needs it, but is supposed to provide it only if the entire process followed the SAML standard. A plain HTTP request header can be added by any remote user with enough savvy.

> As I said, someone that cares about Shibboleth authentication needs to become actively involved in the development of the module if there is any hope for its future. That includes the promise of maintaining and supporting the module. Not just a one and done, hack something in, and disappear as is what is being done for many of these things (like Cosign, Caliper, etc). Most open source projects don't allow things like this to be added without the promise of continued maintenance. We shouldn't either.

That would be very nice, but may be too much to ask for. Maybe we can find someone who will commit to say 18 months of support, so that the code can get deployed and tested in the real world, and improvements and fixes can be made. Once it is "stable" - hopefully some degree of further involvement may continue, but at a lower level of committment.

By the way, I ran into some web page https://incommon.org/case-study-lafayette-college/ where it was mentioned that Shibboleth can integrate with CAS (Central Authentication System) as a back-end. It seems that CAS is still alive outside of UC Berkeley, but is hidden behind other things. From there I discovered that InCommon is maintaining a set of Docker containers of identity management use. See: https://spaces.at.internet2.edu/display/ITAP/InCommon+Trusted+Access+Platform+Release
It could be that those containers could make it possible for us to do some testing against Shibboleth with far less effort than setting up a system from scatch would require.

drgrice1:
I was hoping that @SumitLubal would comment on why the change was made in 2019.  Since that isn't happening, lets just revert the code to what it was before that.  Note that for webwork 2.18 none of this will work.  The Mojolicious app will not have access to these environment variables from the apache process.  So unless someone steps up to maintain the Shibboleth authentication module, it will be dropped.

There is one change that needs to be made to this pull request though, before it is merged.  That is, please remove the perltidy commit.  I know that with the develop branch I would insist on that, and the workflow I added would check that, but the main branch was before all of that.  The perltidy changes will only serve to create merge conflicts later when we merge the develop branch for the next release.  The corresponding change to the develop branch will need the perltidy bit.  There are going to be merge conflicts in any case, but they won't be as extensive without the perltidy part.

taniwallach:
> There is one change that needs to be made to this pull request though, before it is merged. That is, please remove the perltidy commit.

Done. Force pushed to drop the perltidy commit.

drgrice1 (approved):

jmacdone:
> Shibboleth is pretty well accepted, as are the mechanisms of using it with Apache, nginx and other web servers

This is the awareness I was hoping for.  Citing @scantor was deliberate and intended to counter the prejudicial claim of Shibboleth being a "obscure", "insane", "unsecure hack".  Scott is certainly not obscure in security circles and many of us owe our careers to Scott's decades of dedication to the Shibboleth Consortium and authentication in general.  I'm glad he was able to stop in to say "hi".

> Shibboleth authentication has been hacked into WeBWorK

I totally get @drgrice1's point that it's unfun to try to maintain something the project is unequipped to test.  Pragmatically, since webwork isn't using hardly any shibboleth features, I speculate our institution could likely get by using that `Cosign.pm` plugin as [mod_shib is happy enough to set REMOTE_USER if asked](https://shibboleth.atlassian.net/wiki/spaces/SP3/pages/2065335257/AttributeAccess#REMOTE_USER).

> possibly consider replacing Shibboleth with an alternative SAML authentication mechanism

Just to seed a thought: expanding Shibboleth support could provide a funding mechanism.  Have a WeBWorK SaaS that is part of the InCommon Federation and I know our management would more than happy to cut a check.

But, yes, just using the SAML subset does make it more appealing to folks using, say, AD FS.  And, it's arguably easier to containerize apps that don't need the separate shibboleth sp daemon.

> Apache differentiates between system environment variables and server environment variables. See: https://httpd.apache.org/docs/2.4/env.html 

I've received private messages on this as well, highlighting that this is likely the core of the misunderstanding.  @scantor was right (as he always is) to refer to (what I'm accustomed to calling) environment variables as server variables in the documentation.

---------------
tl;dr thanks for taking the time to review the bug report

PR #1846, opened 2022-12-01
Make adding all/multiple users to a given set more efficient

taniwallach:
The updated version of this PR is far more efficient than the earlier versions. Thanks to @drgrice1 for pointing out how to do multiple inserts in a single call to the lower layer which at least gets the prepared statement to be reused.

The improvement based on some very rough bench-marking for an assignment with 25 problems is almost 10 fold, once the inserts for all students are batched into a single group. 

The improvement is almost certainly a function of the size of the assignment as the improvement is mainly due to getting the actual user-problem records to be inserted sequentially using one request of the prepared statement.

---

Add code to allow adding multiple records to the `problem_user` table (for a single set for a single user) as a batch, which allows reducing the number of test calls made to the database. It would be better to be able to add all records in a single database call, but at present the underlying database abstraction layer does not seem to have any support for inserting multiple records in one call.

For a large sample course (862 user accounts) and test assignments with 25 problems each, I observed the "until rendered" time of assigning a set to all users as taking 44-47 seconds with the old code and 38-44 seconds with the new code (on my laptop, using a Docker based setup). That seems to be about a 10% increase in performance. I had hoped for better, as the number of test calls made in the process decreased from 50 per assignment (2 per problem) to 1 per assignment, and there are less calls to subroutines in the process, but clearly that is not that larger bottleneck.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/UsersAssignedToSet.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/UsersAssignedToSet.pm
@@ -25,7 +25,7 @@ users to which sets are assigned.
 
 use WeBWorK::Debug;
 use WeBWorK::Utils qw(format_set_name_display);
-use WeBWorK::Utils::Instructor qw(assignSetToUser assignSetToAllUsers);
+use WeBWorK::Utils::Instructor qw(assignSetToAllUsers assignSetToGivenUsers);
 
 sub initialize ($c) {
 	my $authz = $c->authz;
@@ -70,17 +70,22 @@ sub initialize ($c) {
 		die "Unable to get global set record for $setID " unless $setRecord;
 
 		my %setUsers = map { $_ => 1 } $db->listSetUsers($setID);
+		my @usersToAdd;
 		for my $selectedUser (map { $_->user_id } @{ $c->{user_records} }) {
 			if (exists $selectedUsers{$selectedUser}) {
 				unless ($setUsers{$selectedUser}) {    # skip users already in the set
-					debug("assignSetToUser($selectedUser, ...)");
-					assignSetToUser($db, $selectedUser, $setRecord);
-					debug("done assignSetToUser($selectedUser, ...)");
+					debug("saving $selectedUser to be added to set later");
+					push(@usersToAdd, $selectedUser);
 				}
 			} else {
 				next unless $setUsers{$selectedUser};    # skip users not in the set
 				$db->deleteUserSet($selectedUser, $setID);
 			}
+		}
+		if (@usersToAdd) {
+			debug("assignSetToGivenUsers(...)");
+			assignSetToGivenUsers($db, $c->ce, $setID, 1, $db->getUsers(@usersToAdd));
+			debug("done assignSetToGivenUsers(...)");
 		}
 	}
 
--- a/lib/WeBWorK/DB.pm
+++ b/lib/WeBWorK/DB.pm
@@ -2137,28 +2137,18 @@ sub addUserProblem {
 sub addUserMultipleProblems {
 	my ($self, @problemLists) = @_;
 
-	my $firstUserProblem = $problemLists[0][0];
-	if (!defined($firstUserProblem) || !defined($firstUserProblem->user_id)) {
-		croak "Error in request to addUserMultipleProblems";
-	}
-	my $user_id = $firstUserProblem->user_id;
-	my $set_id  = $firstUserProblem->set_id;
-
-	my ($nv_set_id, $versionNum) = grok_vsetID($set_id);
-	croak "addUserMultipleProblems does not support versioned sets"
-		if defined($versionNum);
-
-	# Possible improvement, if a version set was sent, make calls to addUserProblem for each one.
+	croak "Error in request to addUserMultipleProblems" if ref($problemLists[0]) ne 'ARRAY';
 
 	# Do a single checkArgs call, all records in the list were made in the same manner
-	my @tmp = ($firstUserProblem);
-	$self->checkArgs(\@tmp, qw/VREC:problem_user/);
+	$self->checkArgs([ $problemLists[0][0] ], qw/REC:problem_user/);
+
+	my $set_id = $problemLists[0][0]->set_id;
 
 	# Array in which all records to add are collected
-	my @collectedProblemList = ();
+	my @collectedProblemList;
 
 	for my $problemListRef (@problemLists) {
-		$user_id = $problemListRef->[0]->user_id;
+		my $user_id = $problemListRef->[0]->user_id;
 		croak "addMultipleUserProblems: user set $set_id for user $user_id not found"
 			unless $self->{set_user}->exists($user_id, $set_id);
 
--- a/lib/WeBWorK/DB.pm
+++ b/lib/WeBWorK/DB.pm
@@ -465,6 +465,48 @@ sub restore_tables {
 }
 
 ################################################################################
+# transaction support
+################################################################################
+
+# Any course will have the user table, so that allows getting the database
+# handle.
+
+sub start_transaction {
+	my $self = shift;
+	eval { $self->{user}->dbh->begin_work; };
+	if ($@) {
+		my $msg = "Error in start_transaction: $@";
+		if ($msg =~ /Already in a transaction/) {
+			warn "Aborting active transaction.";
+			$self->{user}->dbh->rollback;
+		}
+		croak $msg;
+	}
+}
+
+sub end_transaction {
+	my $self = shift;
+	eval { $self->{user}->dbh->commit; };
+	if ($@) {
+		my $msg = "Error in end_transaction: $@";
+		$self->abort_transaction;
+		croak $msg;
+	}
+}
+
+sub abort_transaction {
+	my $self = shift;
+	eval {
+		$self->{user}->dbh->{AutoCommit} = 0;
+		$self->{user}->dbh->rollback;
+	};
+	if ($@) {
+		my $msg = "Error in abort_transaction: $@";
+		croak $msg;
+	}
+}
+
+################################################################################
 # user functions
 ################################################################################
 
@@ -1537,6 +1579,41 @@ sub addUserSet {
 	}
 }
 
+# Used to add a list of UserSet records to the DB in less DB calls.
+sub addMultipleUserSets {
+	my ($self, @userSetList) = @_;
+
+	# Do a single checkArgs call, all records in the list were made in the same manner
+	$self->checkArgs([ $userSetList[0] ], qw/REC:set_user/);
+
+	my @badUserIDs;
+	my %badSetIDs;
+	my @toInsert;
+	for my $userSet (@userSetList) {
+		if ($self->{user}->exists($userSet->user_id)) {
+			if ($self->{set}->exists($userSet->set_id)) {
+				# This record is OK
+				push(@toInsert, $userSet);
+			} else {
+				$badSetIDs{ $userSet->set_id } = 1;
+			}
+		} else {
+			push(@badUserIDs, join("", 'user ', $userSet->user_id, " not found, cannot add set ", $userSet->set_id));
+		}
+	}
+	my $badSetIDerrorMsg =
+		keys(%badSetIDs) ? join(" ", 'bad set IDs seen:', keys(%badSetIDs)) : "";
+
+	eval { return $self->{set_user}->insert_records([@toInsert]); };
+	if ($@) {
+		croak join("\n", "addMultipleUserSets errors:\nDB call error: $@", $badSetIDerrorMsg, @badUserIDs);
+	}
+
+	if (@badUserIDs || keys(@badUserIDs)) {
+		croak join("\n", "addMultipleUserSets errors:\n", $badSetIDerrorMsg, @badUserIDs);
+	}
+}
+
 # the code from putUserSet() is duplicated in large part in the following
 # putVersionedUserSet; c.f. that routine
 sub putUserSet {
@@ -2050,6 +2127,54 @@ sub addUserProblem {
 		croak "addUserProblem: user problem exists (perhaps you meant to use putUserProblem?)";
 	} elsif ($@) {
 		die $@;
+	}
+}
+
+# Used to add multiple problems to a SINGLE set with less DB calls.
+# Not for use by versioned sets
+# This expects @problemLists to be a 2 dimensional array, each
+# entry being the problemList array for one user.
+sub addUserMultipleProblems {
+	my ($self, @problemLists) = @_;
+
+	my $firstUserProblem = $problemLists[0][0];
+	if (!defined($firstUserProblem) || !defined($firstUserProblem->user_id)) {
+		croak "Error in request to addUserMultipleProblems";
+	}
+	my $user_id = $firstUserProblem->user_id;
+	my $set_id  = $firstUserProblem->set_id;
+
+	my ($nv_set_id, $versionNum) = grok_vsetID($set_id);
+	croak "addUserMultipleProblems does not support versioned sets"
+		if defined($versionNum);
+
+	# Possible improvement, if a version set was sent, make calls to addUserProblem for each one.
+
+	# Do a single checkArgs call, all records in the list were made in the same manner
+	my @tmp = ($firstUserProblem);
+	$self->checkArgs(\@tmp, qw/VREC:problem_user/);
+
+	# Array in which all records to add are collected
+	my @collectedProblemList = ();
+
+	for my $problemListRef (@problemLists) {
+		$user_id = $problemListRef->[0]->user_id;
+		croak "addMultipleUserProblems: user set $set_id for user $user_id not found"
+			unless $self->{set_user}->exists($user_id, $set_id);
+
+		# Test whether there are already problem records for this user in this set, as in that case
+		# inserting multiple records at once would trigger an error.
+		my $where = [ user_id_eq_set_id_eq => $user_id, $set_id ];
+
+		croak "addMultipleUserProblems cannot be run as set $set_id already contains some problems for user $user_id"
+			if $self->{problem_user}->count_where($where);
+		# If we got here, we can add this list of problems to those to insert into the database
+		push(@collectedProblemList, @{$problemListRef});
+	}
+
+	eval { $self->{problem_user}->insert_records([@collectedProblemList]); };
+	if ($@) {
+		croak "addUserMultipleProblems: $@";
 	}
 }
 
--- a/lib/WeBWorK/Utils/Instructor.pm
+++ b/lib/WeBWorK/Utils/Instructor.pm
@@ -41,8 +41,10 @@ our @EXPORT_OK = qw(
 	assignAllSetsToUser
 	unassignAllSetsFromUser
 	assignSetsToUsers
+	assignSetToGivenUsers
 	unassignSetsFromUsers
 	assignProblemToAllSetUsers
+	assignMultipleProblemsToGivenUsers
 	addProblemToSet
 	read_dir
 	getCSVList
@@ -90,11 +92,22 @@ sub assignSetToUser {
 		}
 	}
 
-	my @GlobalProblems = grep { defined $_ } $db->getAllGlobalProblems($setID);
-	foreach my $GlobalProblem (@GlobalProblems) {
-		my @result = assignProblemToUser($db, $userID, $GlobalProblem);
-		push @results, @result if @result and not $set_assigned;
+	my @globalProblemIDs = $db->listGlobalProblems($setID);
+
+	my $result;
+	# Make the next operation as close to a transaction as possible
+	eval {
+		$db->start_transaction;
+		$result = assignMultipleProblemsToGivenUsers($db, [$userID], $setID, @globalProblemIDs);
+		$db->end_transaction;
+	};
+	if ($@) {
+		my $msg = "assignSetToUser: error during asignMultipleProblemsToGivenUsers: $@";
+		$db->abort_transaction;
+		die $msg;
 	}
+
+	push @results, $result if $result and not $set_assigned;
 
 	return @results;
 }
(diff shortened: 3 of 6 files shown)


lib/WeBWorK/DB.pm
```diff
@@ -2051,6 +2051,53 @@ sub addUserProblem {
...
+	}
+
+	my $msg1 = (@sawRecordExists == 0) ? "" : join("",
+		"addUserMultipleProblems: the following user problems existed before ",
+		join(", ", @sawRecordExists), "\n");
+	if (@otherErrors) {
+		die join("", $msg1, @otherErrors);
+	} elsif (@sawRecordExists) {
+		croak $msg1;
+	}
```
drgrice1, line 2097: This is inconsistent.  Don't mix `die` and `croak`.  Use one or the other.  Although, there is no need for the `if/else` here.  Just do:
```suggestion
	if (@sawRecordExists || @otherErrors) {
		croak join(
			'',
			@sawRecordExists
			? join('',
				'addUserMultipleProblems: the following user problems existed before ',
				join(', ', @sawRecordExists), "\n")
			: '',
			@otherErrors
		);
	}
```

Also note the change from double quotes to single quotes.   Never use a double quote unless it is needed for interpolation or for an escaped character like `"\n"`.  Anytime you use a double quote you tell Perl to try to interpolate it, which adds inefficiency if interpolation is not needed.

lib/WeBWorK/DB.pm
```diff
@@ -2051,6 +2051,53 @@ sub addUserProblem {
...
+	# Just needed once
+	croak "addUserProblem: user set ", $UserProblem->set_id, " for user ", $UserProblem->user_id, " not found"
+		unless $self->{set_user}->exists($UserProblem->user_id, $UserProblem->set_id);
+
+	my @sawRecordExists;
+	my @otherErrors;
+	my @ToReturn;
+	my $result1;
+
+	foreach $UserProblem (@ProblemList) {
```
drgrice1, line 2081: `$UserProblem` needs to be lexical, so add `my` here.  It would probably be best to rename the previous `$UserProblem` variable to something like `$firstUserProblem` to avoid confusion.

lib/WeBWorK/DB.pm
```diff
@@ -2051,6 +2051,53 @@ sub addUserProblem {
 	}
 }
 
+# Adds multiple problems to a user with less database calls.
+sub addUserMultipleProblem {
+	# Intended to be used to add multiple problems to a SINGLE set with less DB calls.
+	# Not for use by versioned sets
```
drgrice1, line 2057: Please move this comment that describes the sub as a whole before the sub.  Also, perlcritic doesn't like this.  The current line 2058 needs to be the very first line of the method.  Even comments aren't allowed before `@_` is unpacked.  This is for good reason too.  It makes it quite clear what the parameters of the method are.  This is actually a huge problem in the webwork2 and pg code.  Not quite as good as signatures, but we'll get there.

drgrice1 (commented):
Here are some initial observations after looking at the code.  I haven't had a chance to test it yet though.

Please stop using `foreach`.  In Perl `for` and `foreach` are aliases for each other and there is no advantage whatsoever to using the longer word.

Also, please don't capitalize the first letter of a variable name.  That is inconsistent with almost all other variable names in the code, and is just not standard for variable names in any language.  Generally that is reserved for package/class names.

See the comments in the code, most of which are remarks about perlcritic and best practices in perl.

lib/WeBWorK/DB.pm
```diff
@@ -2051,6 +2051,53 @@ sub addUserProblem {
...
+	my @tmp = ($UserProblem);
+	$self->checkArgs(\@tmp, qw/VREC:problem_user/);
+
+	# Just needed once
+	croak "addUserProblem: user set ", $UserProblem->set_id, " for user ", $UserProblem->user_id, " not found"
+		unless $self->{set_user}->exists($UserProblem->user_id, $UserProblem->set_id);
+
+	my @sawRecordExists;
+	my @otherErrors;
+	my @ToReturn;
```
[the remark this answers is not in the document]
taniwallach: Yes, I originally thought it would be helpful, as `addUserProblem` was calling `return` but realized that the calling functions would ignore what was sent anyway, so did not use it. Will be removed in the update version.

lib/WeBWorK/DB.pm
```diff
@@ -2051,6 +2051,53 @@ sub addUserProblem {
...
+	my $result1;
```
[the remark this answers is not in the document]
taniwallach: Totally dropping the `$result1`. It was there to originally allow collecting results into the removed `@ToReturn`.

lib/WeBWorK/DB.pm
taniwallach: Thanks. I will need to get used to avoiding unnecessary usage of double quotes. 

I am putting in your version, but copying it manually into the file as it already has other changes. 

The `croak` vs `die` was carried over from the handling of errors in `addUserProblem` and apparently was following the "advice" that `croak` is better when the calling function is to blame for sending a bad subroutine call, and `die` is better when the error occurred in what the subroutine was supposed to do.

lib/WeBWorK/DB.pm
taniwallach: OK, being done.

lib/WeBWorK/ContentGenerator/Instructor.pm
```diff
@@ -155,6 +154,41 @@ sub unassignSetFromUser {
...
+=cut
+
+sub assignMultipleProblemsToUser {
+	my ($self, $userID, @GlobalProblems) = @_;
+	my $db = $self->{db};
+
+	my $GlobalProblem;
+	my $UserProblem;
+	my @records;
+	foreach $GlobalProblem (@GlobalProblems) {
```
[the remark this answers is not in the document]
taniwallach: Done

lib/WeBWorK/DB.pm
taniwallach: Done

taniwallach:
I'm testing an additional small optimization, passing only an array of the the `problem_id` values, as that is the only thing the new `assignMultipleProblemsToUser` loop really needs. There is no need to pass the original array of the problem records and extract the `problem_id` each time.

lib/WeBWorK/DB.pm
drgrice1: Yeah, there is a difference in where `croak` reports the error occurs versus where `die` reports the error.  `croak` reports the line number that the method is called from, and `die` gives the line number where `die` is called.   However, `croak` adds the original line number where it is called to the beginning of the stack trace, and `die` has already given that, and so just continues to there.

So there is a subtle difference that does perhaps justify your original if/else clause.

drgrice1:
So looking at your approach here, I see why this doesn't cause as significant of a speed improvement as you expected.  Really, you just moved the for loop that iterated over the users from the WeBWorK::ContentGenerator::Instructor package to the DB package.  From the perspective of database access, that isn't really anything different.  It is still putting each row into the database separately.  To obtain a real speed improvement, you would need to go deeper and use the `insert_fields` method (or the `insert_fields_i` method for iterator usage -- which might actually help here) of the DB::Schema::NewSQL::Std package.  Well, maybe not that deep, the `insert_records` (or `insert_records_i`) method of that same package would be deep enough.

To test this, naively change lines 2079-2086 that you have currently in DB.pm to `eval { $self->{problem_user}->insert_records([ @problemList ]); };`. Make sure you don't try to include users that already have the set assigned, because obviously there is no error checking here, but you should see a significant improvement.  Getting the error checking to work per problem will take some more investigation, and may require using `insert_records_i` instead.

drgrice1:
Note that my suggestion still doesn't take it down to only one database access.  But it does reuse the prepared SQL::Abstract statement, which does contribute a lot to slowing down things.  To really take it down to one database insertion, methods would need to be added to Std.pm.

drgrice1:
So all of this error checking is for naught. None of it is reported back to the instructor doing the assigning in the case that a problem is actually already assigned to a user.  The only thing that does happen is if an extreme error occurs (probably due to database failure or something), the content generator itself dies.  If a problem is assigned to a user, it quietly fails, and the instructor would never know it.   The return values of `assignSetsToAllUsers` and `assignSetsToUsers` are completely ignored even though the return values have this information.

Clearly, this needs work.  Not this pull request, but how this has been handled.

taniwallach:
@drgrice1 Thanks for pointing out how to insert multiple records "almost" at once. It might not be a single call, but reusing the prepared statement certainly makes a huge difference. Using   `eval { $self->{problem_user}->insert_records([ @problemList ]); };`  makes a huge improvement.

Runtime for the same test cases is now 13-17 seconds. We are about three times faster at this than before.

I put in a count based test to avoid doing the insert if there are already some problem records for the given user and set. There should not be any "reasonable" case where some user did not have the set assigned before the call and had some subset of problems assigned. Yes, the error is just breaking the page and displaying a single error message.

Yes, I was aware that I essentially moved the loop over problems (for a given user) from `lib/WeBWorK/ContentGenerator/Instructor.pm` to `lib/WeBWorK/DB.pm` but that enabled skipping many repeated database calls used as tests, which I hoped would help some.

taniwallach:
@drgrice1 - Iwould appreciate a bit more advice.

There are two more linked ideas which might further improve performance. I am not sure they are a good idea, as it would be grouping DB changes for multiple students into "grouped transactions" (even though at present they are not really transactions) so problems would possible effect several students if something went wrong.

1. I could extended the code of `assignMultipleProblemsToUser` and `addUserMultipleProblems` so that data for multiple students could be handled together one call thereby grouping many more insertions into each `insert_records` call.
2. If doing so, the `addUserSet` calls would be similarly grouped (same group of students).

The downside is that if something goes wrong (ex. process getting killed or some DB error breaking the chain of work in the middle) the database records for a group of students will be messed up. (Ex. The entire group could all get user-set records, but some might not get the problem records, so their assignments would be broken.) In the current approach, damage would likely be restricted to the records for a single student.

What do you think - is the possible speedup enough of a benefit to take the risk of bigger problems when something fails?

drgrice1:
I think it is worth experimenting with at least.   The thing is that if something goes wrong like a DB error or process being killed, that will still all be the same.  Those are extreme failures where you can't stop data corruption if that is a result.  The assignments are going to be broken anyway.

drgrice1:
I think the speed benefits are worth it.  The database failures should be rare in most cases.  Transactions with InnoDB would be nice!

taniwallach:
> I think the speed benefits are worth it. The database failures should be rare in most cases. Transactions with InnoDB would be nice!

Thanks - I will try to work on coding it next week. I'll try to set it up so if transactions are supported it will be done as a transaction.

taniwallach:
> I think the speed benefits are worth it. The database failures should be rare in most cases.

Now coded to handle inserts of problems for multiple students as groups, and also to create the UserSet records as a single batch.

Runtime on the same test case is now 4-6 seconds, typically under 5 seconds. Total performance improvement relative the old code before this PR is close to a factor of 10 better.

> Transactions with InnoDB would be nice!

I tried to code something to add transaction support and group the revised "batches" into transactions.

See https://github.com/taniwallach/webwork2/commit/3f814e447231aed5944b62a3dc751bf73ee99036

Your comments on the approach taken would be appreciated.

It seems to work fine in both a course where the course tables are myISAM (as the current default) and a test course created after I set `$database_storage_engine = 'innodb';` in `conf/site.conf` and restarted the server. (I manually checked that the tables were in fact created as InnoDB tables.)

In the course using InnoDB I tested after importing the class list (again 862 students) and with an assignment with 25 problems, as before.
  - For the old develop code, "assign to all users" took 1.2 minutes (in each of 3 tests).
    - I conclude that InnoDB is clearly significantly slower when many single row inserts are being done.
    - I take that to indicate that we should assume that single row inserts are slower with InnoDB overall.
  - Using the code from this PR and the additional code from commit adding transaction support - "assign set to all users" took 4-6 seconds (in each of 5 tests), similar to that for the MyISAM based course without the additional transaction support, and more the 10 times faster than the old code for the InnoDB tables.
    - Once the inserts are handled a large batches, there is no noticeable different in the run-time of MyISAM and InnoDB tables for these "assign set to all users" operation.

drgrice1:
Yeah, innoDB is not as good when you have many database writes occurring in separate queries.  One of innoDB's weakpoints in general is writing large amounts of data to the database as I understand.  However, if with the approach of adding multiple sets and problems is close in speed between the two engines, then innoDB does have the huge advantage of being able to roll back the changes when something goes wrong.  Potentially preventing database corruption.  With myISAM if a failure occurs part of the way through the job, there is no way rollback, and that could leave the database in an unstable state.

I will try to look at this soon.

pstaabp (approved):
Although I didn't time it exactly, a course with about 500 users and a set of 25 problems, this sped up the assignment about 10 times.

taniwallach:
@pstaab This was not working after the changes to develop.
I have a branch at https://github.com/taniwallach/webwork2/tree/group-problems-adds-v2 which I think is now working.
One of the problems was that creating an empty set was failing before the last commit there.
Can you take a look at it. It is still not finalized.

pstaabp:
I tested the new one and still works.  With a course of about 800 users and a set of 25 problems, this went from 4+mins to 1+mins. 

I wonder if the following will work to get the git stuff straightened out.  Can you delete the local branch of `group-problems-add`, then rename the `v2` one to this name, then force-push?

drgrice1:
@pstaabp:  Your suggested method does work.  I have used that before to change the branch for a pull request.  It is really only the name of the pull request branch that matters and can't be changed.

taniwallach:
This is updated to a single clean commit onto the current develop, and I added the transaction style modifications.
I want to look into updating
  - `lib/WebworkWebservice/SetActions.pm`
  - `lib/WeBWorK/ContentGenerator/Instructor/UsersAssignedToSet.pm`
to make use of the same type of changes.

lib/WebworkWebservice/SetActions.pm
```diff
@@ -276,62 +277,23 @@ sub assignSetToUsers {
...
-		}
-
-		my @GlobalProblems = grep { defined $_ } $db->getAllGlobalProblems($setID);
-		for my $GlobalProblem (@GlobalProblems) {
-			my @result = assignProblemToUser($self, $userID, $GlobalProblem);
-			push @results, @result if @result and not $set_assigned;
-		}
-	}
+	for my $user (@users) {
+		if ($setUsers{$selectedUser}) {
```
drgrice1, line 287: There is something wrong here.  The variable `$selectedUser` is not defined.

lib/WeBWorK/Utils/Instructor.pm
```diff
@@ -90,12 +92,24 @@ sub assignSetToUser {
 		}
 	}
 
-	my @GlobalProblems = grep { defined $_ } $db->getAllGlobalProblems($setID);
-	foreach my $GlobalProblem (@GlobalProblems) {
-		my @result = assignProblemToUser($db, $userID, $GlobalProblem);
-		push @results, @result if @result and not $set_assigned;
+	my @globalProblems   = grep { defined $_ } $db->getAllGlobalProblems($setID);
+	my @globalProblemIDs = map  { $_->problem_id } @globalProblems;
```
drgrice1, line 96: Since you don't need the entire problem record, it would be better just to list the global problems by changing this to
```suggestion
	my @globalProblemIDs = $db->listGlobalProblems($setID);
```

lib/WeBWorK/Utils/Instructor.pm
```diff
@@ -90,12 +92,24 @@ sub assignSetToUser {
...
+
+	my @result;
+	# Make the next operation as close to a transaction as possible
+	eval {
+		$db->start_transaction;
+		@result = assignMultipleProblemsToGivenUsers($db, [$userID], $setID, @globalProblemIDs);
```
drgrice1, line 102: It is a bit odd to have the return value of `assignMultipleProblemsToGivenUsers`.  The return value of that method is either a single error string or undefined.  So the array variable `@result` should be changed to a scalar `$result`, and the documentation above this method modified to state that `If any assignments fail, an error message is returned.`

lib/WeBWorK/Utils/Instructor.pm
```diff
@@ -153,6 +167,47 @@ sub assignSetVersionToUser {
...
+The list of users are sent as an array reference
+If some problem is already assigned to the user, an error string is returned.
+
+=cut
+
+sub assignMultipleProblemsToGivenUsers {
+	my ($db, $userIDsRef, $set_id, @globalProblemIDs) = @_;
+
+	if (!@globalProblemIDs) {    # When the set is empty there is nothing to do
+		return ();
```
drgrice1, line 182: `return ();` is equivalent to `return;`.  So there is no need to add the parentheses.

lib/WeBWorK/Utils/Instructor.pm
```diff
@@ -299,35 +355,65 @@ sub assignSetToAllUsers {
...
+	if ($@) {
+		my $msg =
+			"assignSetToGivenUsers: error during addMultipleUserSets so did not reach assignMultipleProblemsToGivenUsers: $@";
+		$db->abort_transaction;
+		die $msg;
+	}
+	# Now add the problem records - as a batch
+	my @result;
+	eval {
+		@result = assignMultipleProblemsToGivenUsers($db, [@usersToProcess], $setID, @globalProblemIDs);
```
drgrice1, line 408: The same comments for `@result` made to the `assignSetToUser` method apply here.

lib/WeBWorK/Utils/Instructor.pm
```diff
@@ -299,35 +355,65 @@ sub assignSetToAllUsers {
...
-
-	return @results;
+	# Insert them all at once
+	eval {
+		$db->start_transaction;
+		$db->addMultipleUserSets(@userSetsToAdd);
+	};
+	if ($@) {
+		my $msg =
+			"assignSetToGivenUsers: error during addMultipleUserSets so did not reach assignMultipleProblemsToGivenUsers: $@";
```
drgrice1, line 401: This line exceeds the 120 character limit.  I would recommend just removing the "so did not reach assignMultipleProblemsToGivenUsers" part of the message.  That isn't really that useful.

lib/WeBWorK/Utils/Instructor.pm
```diff
@@ -299,35 +355,65 @@ sub assignSetToAllUsers {
...
+
+=cut
+
+sub assignSetToGivenUsers {
+	my ($db, $ce, $setID, @userRecords) = @_;
+
 	debug("$setID: getting problem list");
-	my @GlobalProblems = $db->getAllGlobalProblems($setID);
+	my @globalProblems   = $db->getAllGlobalProblems($setID);
+	my @globalProblemIDs = map { $_->problem_id } @globalProblems;
```
drgrice1, line 375: This should also just be `listGlobalProblems`.

lib/WeBWorK/DB.pm
```diff
@@ -464,6 +464,51 @@ sub restore_tables {
...
+	my $self = shift;
+	eval { $self->{user}->dbh->begin_work; };
+	if ($@) {
+		my $msg = "Error in start_transaction: $@";
+		if ($msg =~ /Already in a transaction/) {
+			warn "Aborting active transaction.";
+			$self->{user}->dbh->rollback;
+		}
+		warn $msg;
+		croak $msg;
```
drgrice1, line 484: Why both a warn and a croak in these methods?  Remove the warn, and just croak.

lib/WeBWorK/DB.pm
```diff
@@ -1537,6 +1582,48 @@ sub addUserSet {
...
+
+	my $firstUserSet = $userSetList[0];
+	my $user_id      = $firstUserSet->user_id;
+	my $set_id       = $firstUserSet->set_id;
+
+	my ($nv_set_id, $versionNum) = grok_vsetID($set_id);
+
+	# Do a single checkArgs call, all records in the list were made in the same manner
+	my @tmp = ($firstUserSet);
+	$self->checkArgs(\@tmp, qw/REC:set_user/);
```
drgrice1, line 1597: There is a lot of unnecessary stuff here.  This could simply be
```suggestion
	# Do a single checkArgs call, all records in the list were made in the same manner
	$self->checkArgs([ $userSetList[0] ], qw/REC:set_user/);
```

lib/WeBWorK/ContentGenerator/Instructor/UsersAssignedToSet.pm
```diff
@@ -25,7 +25,7 @@ users to which sets are assigned.
 
 use WeBWorK::Debug;
 use WeBWorK::Utils qw(format_set_name_display);
-use WeBWorK::Utils::Instructor qw(assignSetToUser assignSetToAllUsers);
+use WeBWorK::Utils::Instructor qw(assignSetToUser assignSetToAllUsers assignSetToGivenUsers);
```
drgrice1, line 28: `assignSetToUser` is no longer used in this module, and so should be removed from this import list.

drgrice1 (commented):
There are a few things that should be changed, but overall this is looking good.

lib/WebworkWebservice/SetActions.pm
taniwallach: Fixed. Thanks.

lib/WeBWorK/Utils/Instructor.pm
taniwallach: Thanks. Fixed in multiple locations.

lib/WeBWorK/Utils/Instructor.pm
taniwallach: Fixed

lib/WeBWorK/Utils/Instructor.pm
taniwallach: Fixed

lib/WeBWorK/Utils/Instructor.pm
taniwallach: Fixed

lib/WeBWorK/Utils/Instructor.pm
taniwallach: Fixed

lib/WeBWorK/DB.pm
taniwallach: Fixed.

lib/WeBWorK/ContentGenerator/Instructor/UsersAssignedToSet.pm
taniwallach: Fixed

drgrice1:
I have encountered a problem that is introduced by this pull request.  This occurs with dropped users on the UsersAssignedToSet.pm page.  If you check one of these users for an assignment on that page, and then click on "Save" it gives the error `assignSetToGivenUsers: error during addMultipleUserSets: argument 1 must be of type WeBWorK::DB::Record::UserSet at /opt/webwork/webwork2/lib/WeBWorK/Utils/Instructor.pm line 395.`

The same happens from the 'Instructor Tools' page (which uses Assigner.pm) if you select a dropped user and a set, and click on "Assign" selected users to selected sets.

On the develop branch this works.

drgrice1:
There is another problem here.  If you go to the UsersAssignedToSet.pm page, and click on 'Assign to All Current Users', and then click on that again you get the error `assignSetToGivenUsers: error during addMultipleUserSets: argument 1 must be of type WeBWorK::DB::Record::UserSet at /opt/webwork/webwork2/lib/WeBWorK/Utils/Instructor.pm line 395.`  The problem is that `@userSetsToAdd` in `assignSetToGivenUsers` is the empty array because the set has already been assigned to all users.  But the code doesn't exit and proceeds on to the `addMultipleUserSets` call anyway, and that fails.

taniwallach:
@drgrice1 reported several bugs:

> I have encountered a problem that is introduced by this pull request. This occurs with dropped users on the UsersAssignedToSet.pm page. If you check one of these users for an assignment on that page, and then click on "Save" it gives the error `assignSetToGivenUsers: error during addMultipleUserSets: argument 1 must be of type WeBWorK::DB::Record::UserSet at /opt/webwork/webwork2/lib/WeBWorK/Utils/Instructor.pm line 395.`
> 
> The same happens from the 'Instructor Tools' page (which uses Assigner.pm) if you select a dropped user and a set, and click on "Assign" selected users to selected sets.

Fixed. The `assignSetToGivenUsers` method now has an additional argument `$alwaysInclude` which is usually true (`1`) but it set to false in the call from `assignSetToAllUsers`. As a result, the test `$ce->status_abbrev_has_behavior($User->status, "include_in_assignment")` which was originally only in `assignSetToAllUsers` will be used to skip dropped users only in the call to `assignSetToGivenUsers` from `assignSetToAllUsers`.

The code will also now exit when `@usersToProcess` is empty to avoid requesting that the database layer handle a request with an empty list of additions.

The suggested changes to `lib/WeBWorK/DB.pm` were merged. Thanks for fixing the bug and improving the tests.

drgrice1 (approved):
Yep.  Those issues are fixed.  Thanks for this.  This looks good now.

pstaabp:
Checked this again.  I'm not any of the errors noted above and the improvement is still about 75%.  Will merge.

PR #1847, opened 2022-12-01
Rename GatewayQuiz as Test.

somiaj:
Rename gateway quizzes as tests in the UI, as discussed in #1838. This also changes the link names from containing `quiz_mode` and `proctored_quiz_mode` to `test_mode` and `proctored_test_mode`.

This changes all the references to `gateway` I could find in the UI (most references were already calling these tests already).

### DIFF
--- a/lib/WeBWorK.pm
+++ b/lib/WeBWorK.pm
@@ -290,7 +290,7 @@ async sub dispatch {
 			# module we double check this, to be sure that someone isn't taking a
 			# proctored quiz but calling the unproctored ContentGenerator
 			my $urlProducedPath = $urlPath->path();
-			if ($urlProducedPath =~ /proctored_quiz_mode/i) {
+			if ($urlProducedPath =~ /proctored_test_mode/i) {
 				my $proctor_authen_module = WeBWorK::Authen::class($ce, "proctor_module");
 				runtime_use $proctor_authen_module;
 				my $authenProctor = $proctor_authen_module->new($r);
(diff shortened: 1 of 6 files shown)


drgrice1:
One thing to consider with the link change is that those still using the `quiz_mode` or `proctored_quiz_mode` links from LTI will now be forced to change all of their links.  Those that have adapted to using the quiz (or now test) page directly won't be affected, but there are probably many that have not made that change yet.  It is sometimes a cumbersome task to change all of those links in the LMS when using "homework" grade passback mode.

somiaj:
@dlglin since you brought up this change, any thoughts on if we should change the url from `quiz_mode` to `test_mode` to be consistent with calling gateway quizzes tests in the ui, or just keep it as `quiz_mode` for backwards compatibility for users who haven't switched to the new links for quizzes/tests? I changed the url for completeness since it is something users may see, but it is a minor thing and probably best to leave alone.

drgrice1:
Another option is to allow both.

dlglin:
It would be nice to have the url match the nomenclature.  If it's easy to allow both urls as Glenn suggests then that might be the way to go, as it maintains backward compatibility but also allows consistent naming going forward.

pstaabp:
> It would be nice to have the url match the nomenclature. If it's easy to allow both urls as Glenn suggests then that might be the way to go, as it maintains backward compatibility but also allows consistent naming going forward.

I agree with this.

PR #1848, opened 2022-12-02
Modify hypnotoad configuration and add more documentation.

taniwallach:
Set clients to 1 as WeBWorK is blocking code, increase the default value of graceful_timeout (to further reduce the risk of trouble from hard kills of workers) and raise the default value of workers to allow handling more requests in parallel.

Remark about the need for production systems to tune the value of workers and spare.

Motivation for this is discussed in https://github.com/openwebwork/webwork2/issues/1841

See also https://github.com/mojolicious/mojo/wiki/Benchmarking for why `clients=1` is the recommendation.

### DIFF
--- a/conf/webwork2.mojolicious.dist.yml
+++ b/conf/webwork2.mojolicious.dist.yml
@@ -74,19 +74,89 @@ hypnotoad:
     # Below is an example of how to use ssl certificates when using the docker build.
     #- https://*:8080?cert=/etc/ssl/local/fullchain.pem&key=/etc/ssl/local/privkey.pem
 
-  accepts: 400
-  workers: 10
-  spare: 5
-  clients: 100
-  requests: 5
-  graceful_timeout: 45
+  # Documentation about the settings being provided below can be found at
+  # https://docs.mojolicious.org/Mojo/Server/Hypnotoad#SETTINGS
+
+  # WeBWorK is essentially all blocking code, so allowing hypnotoad to allow
+  # a worker to accept additional requests when a request is being processed
+  # can cause severe latency/delays when a new request is waiting for a "slow"
+  # request to be completed. Thus, in accordance with the recommendations for
+  # such settings, for example in https://github.com/mojolicious/mojo/wiki/Benchmarking
+  # set clients to 1, and set workers to a larger number.
+
+  # Critical tuning for a production server is to increase workers based on the
+  # system resources available (primarily RAM) and the expected concurrent load.
+  #    workers is the number of hypnotoad processing which are created and
+  #            run at all times to be available to server requests.
+  #    spare is the number of additional workers which can be started as
+  #            replacements for workers which have been signaled to stop
+  #            themselves (gracefully) in order to be replaced by a new worker.
+  # Since spare allows additional workers to be started before old ones have
+  # stopped, the value of workers+spare needs to take into account the
+  # available memory of the system.
+  #
+  # The majority of RAM available for workers should probably be dedicated to
+  # making workers as high as possible, to allow connections from many end
+  # users to be accepted and processed in parallel. However, due to the extensive
+  # amount of time (heartbeat_timeout+graceful_timeout) a worker can remain busy
+  # before being forcefully killed, reserving some RAM to allow a moderate value
+  # of spare is recommended.
+  #
+  # We recommend setting spare to at least 5 on a server with under 8GB of RAM,
+  # and at least 8-10 on a server with over 8GB of RAM. As possible, consider
+  # increasing spare to about 10% of the value of workers on your server, should that
+  # be larger than the setting suggested above.
+  #
+  # Recommendations for tuning of workers and spare based on RAM available:
+  #    Consider using 10-12 workers and one spare for each GB of RAM
+  #    (after setting aside 2-4 GB of RAM for the OS and all other software needed).
+  # On a machine with less than 8GB of RAM, decrease workers and raise spare to make
+  # sure that spare is at least 5.
+  # These are provisional recommendations based on some testing, but without
+  # experience with the Mojolicious webwork system in a production environment.
+  #
+  # It is recommended to monitor memory usage under load to determine what further
+  # tuning of the settings may be appropriate.
+
+  clients: 1
+  workers: 25
+  spare: 8
+
+  # WeBWorK apparently has some memory leaks which lead to workers using up more memory
+  # over time as additional requests are processed.
+  # Requests which process many problem renders in a single call will
+  # trigger faster accumulation of "lost memory". The "accepts" parameter
+  # should be kept relatively low to avoid the "lost memory" from causing
+  # significant problems.
+  accepts: 100
+
+  # Keep-alive related settings. These limit how long an inactive connection
+  # (one with no active request) can remain open, and how many requests (ex.
+  # for additional files) can be sent over the same connection.
+  # In order to prevent a worker from being kept waiting for additional requests
+  # from a browser which in not active, keep_alive_timeout is set relatively low,
+  # but an attempt is made to allow a bit of time before connections will be
+  # closed, to reduce the need for a new connection to be opened if a user
+  # is actively navigating on the site.
+  # To allow all dependencies of a WeBWorK page to be served over a single
+  # connection the value of requests was set to 50.
+  keep_alive_timeout: 15
+  requests: 50
+
+  # The graceful_timeout determines how long hypnotoad will wait before hard
+  # killing a worker which was marked as failing to send heartbeats due to
+  # failing or having "blocked" while doing a very slow operation.
+  # After the graceful_timeout passes, the worker will be killed and whatever
+  # it was doing will get stopped without completing. Potentially, this could
+  # cause inconsistencies in the database or other less than ideal outcomes.
+  # Try to set to provide enough spare time to avoid such problems!
+  graceful_timeout: 120
 
   # The following timeout settings will be sufficient in most cases.  However, if you have large class sizes they may
   # not be enough.  If you are seeing long requests fail then increase these.  This may happen, for example, if you have
   # a class with thousands of users and try to assign a set to all users.  In that case you may need much larger values.
   # Note that if you are serving via a proxy, you may also need to increase the timeouts for the proxy server.
   inactivity_timeout: 60
-  keep_alive_timeout: 60
   heartbeat_timeout: 60
 
   # Make sure that the user the server is being run as has access to this file.


conf/webwork2.mojolicious.dist.yml
```diff
@@ -74,12 +74,32 @@ hypnotoad:
...
-  graceful_timeout: 45
+
+  # The graceful_timeout determines how long hypnotoad will wait before hard
+  # killing a worker which was marked as failing to send heartbeats due to
+  # failing or having "blocked" while doing a very slow operation.
+  # After the graceful_timeout passes, the worker will be killed and whatever
+  # it was doing will get stopped without completing. Potentially, this could
+  # cause inconsistencies in the database or other less than ideal outcomes.
+  # Try to set to provide enough spare time to avoid such problems!
+  graceful_timeout: 60
```
drgrice1, line 103: Note that this is still actually decreasing the Mojolicious default for this.  The default is actually 120 seconds.  See https://docs.mojolicious.org/Mojo/Server/Prefork#graceful_timeout.

conf/webwork2.mojolicious.dist.yml
```diff
@@ -74,12 +74,32 @@ hypnotoad:
...
+  # request to be completed. Thus, in accordance with the recommendations for
+  # such settings, for example in https://github.com/mojolicious/mojo/wiki/Benchmarking
+  # set clients to 1, and set workers to a larger number.
+
+  # Critical tuning for a production server is to increase workers based on the
+  # system resources available (primarily RAM) and the expected concurrent load.
+  # Raising "spare" somewhat can also help.
+
+  clients: 1
+  workers: 25
```
drgrice1, line 89: I am not sure this is a good idea.  The problems here are only with Instructor operations for large classes.  For student connections, this will most likely have a dramatic negative impact on performance.  This is a severe decrease from the default for the `clients` setting which is 1000.  The default for `workers` is 4.  It seems that the workers should be increased quite a bit more to account for such a  dramatic decrease in clients, and again, this is only to handle instructor operations.

drgrice1 (commented):
I think that this should wait.  There is still a lot to be figured out for the next release, and this is one of the last steps.  Furthermore, it is something that is going to greatly depend on the needs for each installation.  So it may be something that should not be the default, but mentioned for the extreme cases where the server will commonly need to handle such requests.

conf/webwork2.mojolicious.dist.yml
taniwallach: You had 45 seconds, and I upped it a bit. I'm perfectly willing to raise it further and return it to the default value of 120. Hard kills of a worker have a real ponential to create DB corruption.

conf/webwork2.mojolicious.dist.yml
taniwallach: The whole point of this PR is to set `workers=1` as is recommended for systems with long blocking operations, and that is the recommendation of the Mojolicious documentation.

Long instructor operations will block and when the server is under moderate demand some incoming requests will get assigned to the blocked worker. Students will experience latency which they will not understand, and typical staff will also now know how to diagnose or debug it. Such latency, but to a less severe extent will occur even if the new request is queued after a "3 second" request. Multiple queued requests will trigger cumulative delays, as at present nothing in WeBWorK is non-blocking.

I really see no safe option for operations which does not set `clients=1` at present.

What default value of `workers` do you recommend which is likely to work for developers and memory constrained systems? Increasing the value is easy, but handling changes if a system starts swapping to disk due to overuse of memory is harder.

taniwallach:
I think this should merged soon, once we agree on reasonable defaults for now. That will make the recommendation to use `clients=1` very clear to anyone who tests develop.

I think that the installation + upgrade instructions and install notes will need to highlight the need to tune this setting based on server resources. At that time, we can consider additional changes to the defaults if we think they are warranted.

conf/webwork2.mojolicious.dist.yml
drgrice1: Note what was there was not intended to be the final result.  There was not a lot of thought put into that at that point.  It was basically just copied from the standalone renderer's config.  I had tweaked that a bit after some observations and added some comments for documentation, but realized that further tweaking and documentation would be needed.

This setting should probably be set to the default.

conf/webwork2.mojolicious.dist.yml
drgrice1: I think you meant `clients=1` as that is the recommendation in the Mojolicious documentation.  I am not sure what to use for `workers` at this point.  The rule of thumb is 2 workers per CPU core for non-blocking app, and more are needed for blocking apps (https://docs.mojolicious.org/Mojo/Server/Hypnotoad#workers).  Note the default is 4.  I don't know how many more should be used for a blocking app.

conf/webwork2.mojolicious.dist.yml
```diff
@@ -74,12 +74,32 @@ hypnotoad:
...
   spare: 5
```
drgrice1, line 90: I think that the `spare` setting should also be increased.  This is the number of spare workers that Mojolicious can spin up if needed.  The default value for this is 2, which is half of the default workers value of 4.  So probably this should be set to half of whatever we settle on for the workers value.

conf/webwork2.mojolicious.dist.yml
taniwallach: OK - being set to 120 in an updated version of the PR to be force pushed soon.

conf/webwork2.mojolicious.dist.yml
taniwallach: Yes, I meant `clients=1`. Fixed in the comment above.

About `workers` - it should almost definitely be significantly higher for any production server, as it is the actual limit of the number of requests which can be processed in parallel. The ideal is that the number of workers should be somewhat larger than the number of simultaneous requests a server is expected to handle in parallel. However, the setting must be limited based on system memory available to avoid major problems which would occur if the system started to swap RAM to disk. Keeping the value high will allow the operating system to handle the multi-tasking of the multiple processes, and is the best option available to us at present to allow divide the CPU resources among many blocking processes.

At present, I don't think any of us have experience with the memory usage of the hypnotoad workers under production type workloads to provide any recommendations about roughly how many workers can be safely started per GB of RAM (after setting aside say 1-3 GB of RAM for the OS and other software running on the system).

Also worth noting is that hypnotoads `accepts` is somewhat similar to Apache `mod_prefork`'s 
`MaxConnectionsPerChild` (was `MaxRequestsPerChild`) and should be limited to avoid memory leaks from causing too much memory to be used by a worker due to the memory leaks of WeBWorK and PG. Based on several forum posts, I am reducing this from `400` to `100`. If the memory leak problem is less severe for hypnotoad, this can be raised more.

That approach is essentially the same approach as used for the Apache/mod_perl setup where the recommendation was to tune the settings in Apache's `mpm_prefork.conf` file to tune `StartServers`, `MinSpareServers`, `MaxSpareServers`, `MaxRequestWorkers` based on available RAM. There were also posts about using `MaxConnectionsPerChild` (old name `MaxRequestsPerChild`) and `Apache2::SizeLimit;` to limit RAM per worker. 
  - https://webwork.maa.org/moodle/mod/forum/discuss.php?d=3228#p8106
  - https://webwork.maa.org/moodle/mod/forum/discuss.php?d=2692#p5887
  - https://webwork.maa.org/moodle/mod/forum/discuss.php?d=5008
  - https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4645

@dlglin - You seem to have handled many of the UsingWW forum discussions where such issues came up. If you have any better advice on what defaults to set for now, or how we should provide instructions on tuning these settings, please post a message.

conf/webwork2.mojolicious.dist.yml
taniwallach: I think that setting it to half of workers is too high when workers is not a small value, and in particular when `clients` is a small number. When workers is a low value and `clients` is larger, as typical for a non-blocking server, it makes far more sense to have spare be a very significant proportion of workers, so that a new worker can be brought online quickly while an old one is being gracefully stopped, as I assume that hypnotoad will not assign new requests to a worker which has been told to stop.

Since WeBWorK will be using `clients=1` while it remains a blocking codebase, we need to keep many workers running in parallel. The key question in tuning `spare` (relative to `workers`) is how many workers are likely to need to be restarted in parallel.
  - Some workers will need restarts due to the limit on the number of connections accepted (`accepts`). Hypnotoad is somewhat flexible and will restart workers earlier if the code thinks that will help. https://docs.mojolicious.org/Mojo/Server/Hypnotoad#accepts
  - The other expected cause of restarts are requests whose processing time exceeds `heartbeat_timeout`.
    - We currently assume that such restarts should be pretty rare.
    - The primary expected types of requests for which this may occur are certain types of instructor actions which effect many students (or many sets) at once.

Since we need to take into consideration the RAM constraints on the total number of workers the system RAM can safely support, and the fact that by allowing a large value of `graceful_timeout`  we are essentially potentially allowing the server to run `workers` + `spare` workers at once for possibly extended amounts of time. I would not want to reserve about half of the RAM potentially available to run workers to be saved only for `spare` workers and as such reduce by about 50% the number of workers being run under regular circumstances.

I'm raising the provided setting to `spare=8` for now (almost 1/3 of the 25 workers) and adding more comments to the file. If we decide that the memory leaks requires significantly reducing `accepts` then it would make sense to increase both the default value of `spare` further as well as the percentage of `workers` we recommend to use as a setting for `spare` on systems with many workers.

drgrice1: Yeah, I was extrapolating a multiplier of 2.  I only had one data point to work with, so that is not expected to be a reliable answer.  You rationale on the spare not being as large makes sense.  Once hypnotoad determines that a worker is to be stopped it certainly will not assign new requests to the worker.  The worker finishes its current requests, and then is killed.  That is how zero downtime is achieved when restarting hypnotoad as a whole also.

conf/webwork2.mojolicious.dist.yml
drgrice1: The memory leak issues of the past should be considerably less with Mojolicious.  I can't say that webwork2 doesn't have any of that now, but at least with regards to PG and rendering of problems it doesn't.  This is really quite simple to access since that is all done in a forked process.  When that process completes its memory usage is wiped out, and it has no affect on the hypnotoad worker process memory consumption at all (even while still active).

taniwallach:
The settings for the number of `workers` and for `spare` must take into account the RAM available on a server, but also the expected memory usage of a worker. However, we are aware of the fact that WeBWorK has memory leaks, suspected mostly to be in PG.

In some quick testing, I saw memory usage increasing by several MB for each problem render I ran on the Mojo based server. 

Given this issue I looked to see if there is something analogous to `Apache2::SizeLimit` for Mojolicious. I found `Mojolicious::Plugin::SizeLimit`.    https://metacpan.org/pod/Mojolicious::Plugin::SizeLimit

Using such a tool would provide a more dependable mechanism to prevent workers from taking up too much memory over time, and is far more accurate in doing so than just limiting `accepts` to try to avoid the problem.

I recommend we look into whether this can be helpful for us, and if so to what extent we can make it easy to configure via settings in some file in the `webwork2/conf` directory.

drgrice1:
As I mentioned in another comment, the rendering of PG problems will not count toward the worker memory usage at all.  So the `Mojolicious::Plugin::SizeLimit` will miss any memory consumed by that as that is in a forked process.  Fortunately, there will be no memory leaks due to that either and it will be completely cleaned up since the entire forked process will be reaped.

In testing I found that the typical memory consumption for a webwork2 page is far less than 100K.  That is not counting the consumption for PG rendering.  PG rendering typically consumes around 5MB of memory per problem.

conf/webwork2.mojolicious.dist.yml
taniwallach: OK - thanks for explaining. That is a very useful change to the system.

I looked far more carefully at the data I had collected and I was totally misinterpreting it. Earlier I had just focued on the "bottom line" with the total of memory used by all the workers and and not on the individual processes.

I was using `smem` as it tried to allocate shared memory amount all the processes, as otherwise `RSS` just charged all the shared memory to each process. I was running `smem -tk --processfilter="opt/webwork/webwork2/bin"` 

I was seeing the totals for both `USS` and `PSS`  increase quite dramatically after I had WW render the same problem (sequentially)  for an addition 10 users when 25 workers were running. For a test with `Library/Hope/Multi1/04-04-Basis-and-functions/Matrix_rep_05.pg` both increased by about 30MB for 10 more renders.

Upon further examination the main change was that an additional worker became active, and its memory needs jumped by roughly the 30MB increase observed.

I tested again with a single worker, which avoids there being an option for more workers to activate. (Data futher below). In that setting the increase in memory used is very low for an additional 10 renders of the problem.

---

Data from the testing with 25 workers, and where the large increase had been observed. The two runs of `smem` were separated by iterating in the browser through 10 more students of the given problem. In retrospect it is obvious that the additional worker becoming active (loading the WW code?) is the main change.

```
root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
984835 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984831 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984832 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984830 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984834 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984829 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984828 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984827 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984826 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984825 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984824 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984842 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.0M 
984841 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984839 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984840 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984836 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984837 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984843 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.0M 
984847 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984845 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984833 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
984846 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.6M   164.1M 
985220 root     /usr/bin/python3 /usr/bin/s        0     9.5M     9.8M    15.3M 
984823 www-data /opt/webwork/webwork2/bin/w        0     7.8M    14.9M   173.7M 
984838 www-data /opt/webwork/webwork2/bin/w        0    32.0M    38.4M   180.5M 
984844 www-data /opt/webwork/webwork2/bin/w        0    32.1M    38.5M   180.5M 
984848 www-data /opt/webwork/webwork2/bin/w        0    32.1M    38.5M   180.6M 
-------------------------------------------------------------------------------
   27 2                                           0   139.7M   305.3M     4.2G 

root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
984831 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984832 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984830 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984834 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984829 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984828 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984827 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984826 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984825 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.4M   164.0M 
984824 www-data /opt/webwork/webwork2/bin/w        0     1.1M     7.5M   164.0M 
984841 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984842 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.0M 
984837 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984839 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984836 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984843 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.0M 
984835 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984833 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984845 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984846 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
984847 www-data /opt/webwork/webwork2/bin/w        0     1.3M     7.7M   164.1M 
985298 root     /usr/bin/python3 /usr/bin/s        0     9.0M     9.3M    14.9M 
984823 www-data /opt/webwork/webwork2/bin/w        0     7.8M    14.7M   173.7M 
984838 www-data /opt/webwork/webwork2/bin/w        0    32.0M    38.2M   180.5M 
984840 www-data /opt/webwork/webwork2/bin/w        0    32.0M    38.2M   180.5M 
984844 www-data /opt/webwork/webwork2/bin/w        0    32.1M    38.2M   180.5M 
984848 www-data /opt/webwork/webwork2/bin/w        0    32.1M    38.2M   180.6M 
-------------------------------------------------------------------------------
   27 2                                           0   170.2M   335.7M     4.3G 
```

---


I tried the same test with `workers=1`.

Just after starting the Docker container:
```
root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
987286 root     /usr/bin/python3 /usr/bin/s        0     8.5M     8.9M    14.6M 
987284 www-data /opt/webwork/webwork2/bin/w        0     1.4M    82.3M   163.9M 
987283 www-data /opt/webwork/webwork2/bin/w        0    10.7M    92.0M   174.0M 
-------------------------------------------------------------------------------
    3 2                                           0    20.6M   183.2M   352.6M 
```

After loading the "student" list of problems in an assignment (when acting as one specific given user):
```
root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
987320 root     /usr/bin/python3 /usr/bin/s        0     8.6M     8.9M    14.3M 
987283 www-data /opt/webwork/webwork2/bin/w        0    24.7M    98.9M   174.0M 
987284 www-data /opt/webwork/webwork2/bin/w        0    25.5M    99.7M   174.6M 
-------------------------------------------------------------------------------
    3 2                                           0    58.8M   207.5M   363.0M 
```

After loading that same "student" list of problems in an assignment a second time.
```
root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
987339 root     /usr/bin/python3 /usr/bin/s        0     8.6M     8.9M    14.4M 
987283 www-data /opt/webwork/webwork2/bin/w        0    24.7M    98.9M   174.0M 
987284 www-data /opt/webwork/webwork2/bin/w        0    26.2M   100.3M   175.2M 
-------------------------------------------------------------------------------
    3 2                                           0    59.5M   208.1M   363.6M 
```

After loading the first problem for that student and then iterating through 9 more students via the ">" next student link.
```
root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
987405 root     /usr/bin/python3 /usr/bin/s        0     8.6M     8.9M    14.5M 
987283 www-data /opt/webwork/webwork2/bin/w        0    26.1M    99.6M   174.0M 
987284 www-data /opt/webwork/webwork2/bin/w        0    33.6M   107.0M   181.2M 
-------------------------------------------------------------------------------
    3 2                                           0    68.4M   215.5M   369.8M 
```

After loading the problem for another 10 users:
```
root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
987588 root     /usr/bin/python3 /usr/bin/s        0     8.6M     8.9M    14.7M 
987283 www-data /opt/webwork/webwork2/bin/w        0    26.2M    99.6M   174.0M 
987284 www-data /opt/webwork/webwork2/bin/w        0    35.6M   109.0M   183.2M 
-------------------------------------------------------------------------------
    3 2                                           0    70.4M   217.6M   371.9M 
```

The change now was only about 2MB for the 10 additional renders.

After another 10 students:
```
root@lxtani:/home/tani# smem -tk --processfilter="opt/webwork/webwork2/bin"
  PID User     Command                         Swap      USS      PSS      RSS 
988053 root     /usr/bin/python3 /usr/bin/s        0     8.6M     8.9M    14.4M 
987283 www-data /opt/webwork/webwork2/bin/w        0    26.2M    99.7M   174.0M 
987284 www-data /opt/webwork/webwork2/bin/w        0    37.5M   110.9M   185.1M 
-------------------------------------------------------------------------------
    3 2                                           0    72.3M   219.5M   373.6M 
```

taniwallach:
I'm referring below to the memory usage terms `smem` used, as they are more fine grained than what `ps` can give.

My testing showed workers as taking about 25 MB of `USS` memory once they needed to do any webwork2 work.This seems to grow to about 31MB once problems are being rendered and then slowly grows further.

Leaving room for `USS` memory usage to double before a worker gets killed (and ignoring the growth in `PSS` / shared memory usage) but reserving an additional 5MB for the PG rendering per worker suggests that about 15 workers (including spares) per GB of system RAM should be probably be an upper upper limit in the recommendations, at least until we have more experience.

I suggest recommending 10-12 workers and one spare per GB of RAM (after setting aside 2-4 GB of RAM for the OS and all other software needed). That is intentionally somewhat conservative. Obviously system administrators should try to monitor memory usage (at least at busy times) to make sure that a reasonable amount of free memory remains available to the system even under heavy load.

What do you think?

I still think there may be benefit in looking into `Mojolicious::Plugin::SizeLimit`. The better we can limit how much "additional" memory a worker uses the less likely servers are to end up in low memory conditions under heavy or unusual loads. With firmer limits it might be possible to recommend 15-20 workers and two spares per GB of RAM.

drgrice1:
That sounds reasonable.   You have done a pretty good analysis of the usage to support that recommendation.

PR #1851, opened 2022-12-15
installing SVG module for docker image

fabiangabel:
Currently `docker-compose up` yields

```
app_1  | Can't load application from file "/opt/webwork/webwork2/bin/webwork2": Can't locate SVG.pm in @INC (you may need to install the SVG module) (@INC contains: /opt/webwork/pg/lib /opt/webwork/webwork2/lib /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.34.0 /usr/local/share/perl/5.34.0 /usr/lib/x86_64-linux-gnu/perl5/5.34 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.34 /usr/share/perl/5.34 /usr/local/lib/site_perl) at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm line 33.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -126,6 +126,7 @@ RUN apt-get update \
 	libsql-abstract-perl \
 	libstring-shellquote-perl \
 	libsub-uplevel-perl \
+        libsvg-perl \
 	libtemplate-perl \
 	libtest-deep-perl \
 	libtest-exception-perl \
--- a/DockerfileStage1
+++ b/DockerfileStage1
@@ -88,6 +88,7 @@ RUN apt-get update \
 	libsql-abstract-perl \
 	libstring-shellquote-perl \
 	libsub-uplevel-perl \
+        libsvg-perl \
 	libtemplate-perl \
 	libtest-deep-perl \
 	libtest-exception-perl \


drgrice1:
The docker build will need to be updated for the new module added in #1826.  Thanks for pointing that out.

taniwallach:
I think the relevant Perl package is available for Ubuntu as `libsvg-perl` so it can hopefully be included using the distribution package rather than from CPAN.
https://packages.ubuntu.com/jammy/libsvg-perl

It should be possible to set `ADD_APT_PACKAGES` in `docker-compose.yml` to install the package at run time, to make sure that it works before doing a rebuild of the images. @fabiangabel - Do you mind testing that and if relevant updating the PR. Whatever change does get made would need to be in both `DockerfileStage1` and the one-stage `Dockerfile`.

taniwallach:
@fabiangabel I was testing a PR which already depended on the `SVG` module, and can confirm that using `libsvg-perl` in `ADD_APT_PACKAGES` allows the Docker image to start up, and that the bar charts in the revised statistics pages are being generated. Rebuilding the Docker images requires longer than I can spare right now, but that is certainly needed to avoid the startup delay incurred in installing the package at container startup.

drgrice1 (approved):
Looks good.  I haven't tested this, but it should work.

fabiangabel:
@taniwallach and @drgrice1 I have added the package to `Dockerfile` and `DockerfileStage1` 
I successfully tested building the Docker images on my machine.

taniwallach (approved):
I have also build a new Docker image with this, and it works and starts up properly.

taniwallach:
@fabiangabel Could you `git rebase -i 1fc677198aba514a84992be6994bd8805dd460ac` and merge the 2 commits into a single commit with a suitable message. Then, please, `gut push --force`. The will cut back on there being 2 commits for this small change.

drgrice1:
We can also just squash it when we merge this.

fabiangabel:
@taniwallach done. It should be only one commit now.

taniwallach:
Thanks. Merging it.

drgrice1:
I think this can also be merged into develop.  I will go ahead and do that.

drgrice1:
Oh, uh.  You did.

PR #1856, opened 2023-01-04
Fixes to LTIUpdate.pm

somiaj:
Fix an issue where a selected user was not being update when in course grade mode (the logic was incorrectly only updating a user for All Sets only in homework grade mode).

Added an errorMessage for if something goes wrong with the post, and changed the updateMessage to use the alert-success class (instead of warning).

These issues were pointed out in the previous pull request #1842, which has been merged.


drgrice1 (approved):
Looks good.  I will go ahead and merge this.

PR #1857, opened 2023-01-05
Change webwork2 templates into Mojolicious templates  and route with the Mojolicious router

drgrice1:
Convert templates to Mojolicious templates and remove all CGI usage.

To make this work all template escape handlers need to be non-async.  Thus all of the set headers that were generated in the template methods need to be generated instead in the initialize method which can be async.

The previous theme template file system.template is now system.html.ep, and is a layout.  The gateway.template file is no more.  All content generator modules use the system.html.ep template.  The special case for Problem.pm and ShowMeAnother.pm in system.template is also removed.  Those modules have their own template file that is the default content just like all the other content generator modules.

The output_JS and output_CSS files are removed, and added in each content generator modules template.

All ContentGenerator modules are actual controllers and the Mojolicous router is used for dispatching requests.  The WeBWorK::Request module is now the WeBWorK::Controller module. It derives from the Mojolicious::Controller package, and is that base class for all WeBWorK Mojolicious controller modules.  The WeBWorK::ContentGenerator module derives from that, and as before all other content generator modules derive from that.  So the package hierarchy is:
            Mojolicious::Controller
            -> WeBWorK::Controller
            -> WeBWorK::ContentGenerator
            -> WeBWorK::ContentGenerator::...

In the ContentGenerator modules the old $r variable (which stood for an Apache::Request) has been replaced with $c (for Mojolicious::Controller or WeBWorK::Controller).  Furthermore, $self is gone.  There is only $c.  The old $self and $r are now both $c.  The current controller module is available in the templates (made available by Mojolicious) as $c, so this naming is consistent bot in the module itself and in the templates.  Yeah, $self is also made available in the templates by Mojolicious and $c == $self, but $c is shorter and this is consistent with the Mojolicious documentation also.
    
Note that signatures have been enabled in all content generator modules (and a few other modules).
    
lib/WeBWorK.pm still initializes all ContentGenerator requests by initializing the course environment and checking authentication, but then it relinquishes control to the appropriate ContentGenerator controller module.  That is except for the case when the user is not authenticate, in which case the Login.pm module is rendered directly, and the case that login proctor authentication is needed, in which case the LoginProctor.pm module is rendered directly.  This is done in the `around_action` hook.
    
Mojolicious routes are initialized in the WeBWork::Utils::Routes module.  That module also has two convenience methods that return webwork specific parameters that are not Mojo::Routes:Route properties.  Those are the route_title which returns the translated display title for a route, and route_navigation_is_restricted, which determines if a route
is allowed for LTI navigation restricted users.  Note that the WeBWorK::URLPath module is no more.  Everything that did can be done directly with the Mojo::Routes or Mojo::Routes::Route modules or the two convenience methods of the WeBWork::Utils::Routes module.
    
The most prominently utilized Mojolicious method that replaces the old WeBWorK::URLPath usage is the `url_for` Mojolicious::Controller method.  This returns a Mojo::URL.  It automatically incorporates stash values (previously urlpath args) for the current route into the generated url.  Furthermore, the Mojo::URL `query` method is used to simplify adding query parameters.  The WeBWorK::ContentGenerator `systemLink` method is rewritten to utilize this.  That modules
`url_args` and `url_authen_args` methods are revised to not directly generate the tail of the url, but to return a hash that can be passed to the `query` method.
    
Note that the WeBWorK::FakeRequest module has been deleted.  That is not only not used, but also won't work anymore.  It could be made to work again, but it would take some effort and we realy don't need it anymore.

Several files were moved from the lib/WeBWorK/ContentGenerator to lib/WeBWorK/HTML to get non-controller modules out of there.  lib/WeBWorK/Utils/AttemptsTable.pm was also moved to lib/WeBWorK/HTML as it seems to fit better there.  All of the modules there are related to adding HTML elements to the page.  Note there were several unused modules in lib/WeBWorK/HTML that were deleted.

Note that the git upgrade notifier has been removed.  I didn't feel like updating the related HMTL for that, and it is broken anyway.

There is probably more that I am forgetting.  I know this is a huge pull request.  There is more work to be done though.  The global STDOUT selection and global `$SIG{__DIE__}` handler have been removed, but the global `$SIG{__WARN__}` handler remains.  This means that it is possible for one user to get the warnings intended for another user if the timing of their requests are correct and their requests are sent to the same worker.  With the global STDOUT selection currently in the develop branch, the entire page of one user can be sent to another user.


drgrice1:
Note that this is built on top of #1827.  I forgot to mention that.

htdocs/js/apps/SetMaker/setmaker.js
```diff
@@ -39,51 +39,42 @@
...
+		requestObject.library_subjects = libSubjects?.value ?? '';
+		requestObject.library_chapters = libChapters?.value ?? '';
+		requestObject.library_sections = libSections?.value ?? '';
+		requestObject.library_textbook = libraryTextbook?.value ?? '';
+		requestObject.library_textchapter = libraryChapter?.value ?? '';
+		requestObject.library_textsection = librarySection?.value ?? '';
+		requestObject.includeOPL = (includeOPL.type === 'checkbox' && includeOPL?.checked) ||
+			(includeOPL.type === 'hidden' && includeOPL.value) ? 1 : 0;
+		requestObject.includeContrib = includeContrib?.checked ? 1 : 0;
+		console.log(requestObject);
```
pstaabp, line 77: Is this just for debugging?

drgrice1: It was, and it is not supposed to still be in the commit.  Thanks for pointing that out.

pstaabp:
Overall, this looks great.  I came across one issue. If the problem source (file path) is empty, previously there would be an immediate error 
```
Can't use an undefined value as a HASH reference at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 1108.
```

With this branch, things are just stuck.  In the morbo console, I get many warnings/errors, the most relevant seems to be:
```
[2023-01-10 06:42:02.09205] [36579] [warn] [ZG2KRzqkaCAl] [/webwork2/staab_course/set_10/4] Unhandled rejected promise: Can't use an undefined value as a HASH reference at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm line 884.
Context:
  879:          total_inprogress => $total_inprogress,
  880:  );
  881: }
  882: 
  883: sub nav ($c, $args) {
  884:  my %can = %{ $c->{can} };
  885: 
  886:  my $db    = $c->db;
  887:  my $ce    = $c->ce;
  888:  my $authz = $c->authz;
  889: 
```

Although there shouldn't be any problems with empty file_path, this seems to pop up once in a while and I think we should bail out more elegantly.

drgrice1:
I fixed the issue with a problem that has an empty source.

There are still other issues when the set id or problem id are not valid.  Neither Problem.pm nor GatewayQuiz.pm are set up well to deal with invalid url path parameters.

drgrice1:
I made a few changes to make a url with a valid course id, but invalid set or problem work.

So if you have an existing course with "courseID" then you can enter `/webwork2/courseID/invalidSet/5` where "invalidSet" does not exist as a set in the course, or `/webwork2/courseID/validSet/100` where "validSet" is a valid set in the course but it doesn't have problem 100, and this will just show "This problem is not valid."

pstaabp:
These changes look good.

PR #1860, opened 2023-01-07
Update third party libraries to their latest versions.

drgrice1:
Well, that is except for codemirror.  Upgrading to version 6 will require a build system.

The upgrade to Bootstrap 5.2.3 requires a few changes to the code.

The changes to regain the dark borders on table thead sections are perhaps not necessary.  Those borders were not there before the upgrade to the previous version of Bootstrap 5.  The previous version forced those borders.  It is a new feature that those borders are opt in.  This has to be done per table.  However, I have become accustomed to those borders, so I opted in.  If you feel differently, let me know.

In order for the accordion buttons on the ProblemSetDetail.pm page to show up the containing div needs to have the accordion class now. Fortunately with that everything works as it should.


dlglin:
Should there be a hotfix to 2.17 for the luxon vulnerability?

drgrice1:
It wouldn't be a bad idea.  I guess I could just open the dependabot's pull request, and merge it.

pstaabp:
Heads up that after checking out this branch, you need to `npm install` the `generate-assets.js` script is not sufficient. 

> Should there be a hotfix to 2.17 for the luxon vulnerability?

This is probably a good idea.

pstaabp (approved):

drgrice1:
Yeah, since this changes the actual libraries that npm installs, you certainly need to run `npm install`.

Issue #1853, opened 2022-12-16
duplicate LTI tab in Config page


Alex-Jordan:
Sometimes `authen_LTI.conf` gets loaded more than once. While that is not ideal, with instructors managing their `course.conf` file and following online guides, it happens.

In the past, no harm done because all it does is set variable values. But now `authen_LTI.conf` loads `LTIConfigValues.config`, which can push a new entry to `@$ConfigValues`. If this happens twice, you end up seeing the LTI tab twice in the config page.

Can we do something to prevent the push from happening twice? Or to only push entries where the label (like "LTI") is not already in the array? Or to overwrite an entry if that label is already in the array? I thought about doing one of these but I'm not sure what is the "right" way to address the issue.

pstaabp:
Since `$ConfigValues` is a arrayref, it'll just push on another value.  One way to handle this would be to switch to a hashref, where the name of the tab is the fieldname of the hashref and if added a second time, it would just replace what was there before.  

This seems like a clean way to handle this, but would require refactoring both the configuration in `defaults.config` and the config page.

drgrice1:
The `$ConfigValues` needs to be an array reference to maintain order.

pstaabp:
> The `$ConfigValues` needs to be an array reference to maintain order.

Good point.  I guess the code in `authen_LTI.conf` could check for the tab first before pushing on would be reasonable and changes contained just to that file.

drgrice1:
What really needs to be done is the `course.conf` and `simple.conf` files need to be made truly and completely not modifiable by non-administrative users.  This would be a big project though.  First, you would need to remove the files from the file manager listing.  Then you also have to make all of the file manager methods careful not to allow downloading of the files, and protect the files from being overwritten by uploading or unarchiving.  Then you would need to make sure that all settings that an instructor should be allowed to modify are configurable via the web interface (probably the hardest part).  I guess you need to shut down all of those online guides instructing users to modify these files also.

dlglin:
If we want to go down the road of preventing instructors from modifying `course.conf` and `simple.conf` one option is to move them outside the course directory.  This will prevent many of the workarounds using the file manager.

pstaabp:
This is now merged.

PR #1861, opened 2023-01-07
Prevent duplicate LTI tabs on the course configuration page.

drgrice1:
Fixes issue #1853.

This just checks the @ConfigValues array for elements whose first sub array element is 'LTI'.  If any are found, then it doesn't add the elements from the @LTIConfigVariables array.

### DIFF
--- a/conf/LTIConfigValues.config
+++ b/conf/LTIConfigValues.config
@@ -93,7 +93,7 @@ $LTIConfigValues = {
 	},
 };
 
-if (@LTIConfigVariables) {
+if (@LTIConfigVariables && !(grep { $_->[0] eq 'LTI' } @$ConfigValues)) {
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Although I haven't seen the second LTI tab that was reported, this produces one LTI tab and the code looks good.

drgrice1:
To make the second tab appear first enable LTI authentication by uncommenting the `include("conf/authen_LTI.conf")` line in localOverrides.conf (line 425 in the dist file).  Then add that same line to a courses `course.conf` file.  Make sure to copy `authen_LTI.conf.dist` to `authen_LTI.conf` and uncomment lines 372-377.

drgrice1:
Yeah, that would work too!

PR #1864, opened 2023-01-11
Add a Minion job queue.

drgrice1:
This is used for LTI mass grade updates and sending instructor emails.

A special warnings method is added to submit grade that sends debugging warnings to the log instead of using a native warn.  During a mass update those warn statements fall on deaf ears in production deployment with hypnotoad.

The code is generally cleaned up in that file.  Replace CGI::escapeHTML with HTML::Entities::encode_entities which does the same thing (for the most part).  Also, replace LWP::UserAgent with Mojo::UserAgent.

For this to work with the Minion::Backend::SQLite module, you will need to install version 3.002 of the Mojo::SQLite package.  Newer versions of that module cause issues.  Note that by default the sqlite database file is "$webwork_dir/DATA/webwork2_job_queue.db".

Other backends that can be used are the Minion::Backend::Pg module with postgres, and the Minion::Backend::mysql module with mysql.  Setting those up takes a little more work though.  The SQLite backend is much simpler.

The usage is documented in conf/README.md.

In short, for development you can run the job queue with `./bin/webwork2 minion worker`,
and for production copy conf/webwork2-job-queue.dist.service to conf/webwork2-job-queue.service and run
`sudo systemctl enable /opt/webwork/webwork2/conf/webwork2-job-queue.service` and
`sudo systemctl start webwork2-job-queue`

This is #1858 which will shortly be closed.  I accidentally pushed to origin.  My bad.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -112,6 +112,8 @@ my @modulesList = qw(
 	Locale::Maketext::Simple
 	LWP::Protocol::https
 	MIME::Base64
+	Minion
+	Minion::Backend::SQLite
 	Mojolicious
 	Mojolicious::Plugin::NotYAMLConfig
 	Net::IP
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,6 +108,8 @@ RUN apt-get update \
 	libmail-sender-perl \
 	libmariadb-dev \
 	libmime-tools-perl \
+	libminion-backend-sqlite-perl \
+	libminion-perl \
 	libmodule-build-perl \
 	libmodule-pluggable-perl \
 	libmojolicious-perl \
@@ -126,7 +128,7 @@ RUN apt-get update \
 	libsql-abstract-perl \
 	libstring-shellquote-perl \
 	libsub-uplevel-perl \
-        libsvg-perl \
+	libsvg-perl \
 	libtemplate-perl \
 	libtest-deep-perl \
 	libtest-exception-perl \
--- a/DockerfileStage1
+++ b/DockerfileStage1
@@ -70,6 +70,8 @@ RUN apt-get update \
 	libmail-sender-perl \
 	libmariadb-dev \
 	libmime-tools-perl \
+	libminion-backend-sqlite-perl \
+	libminion-perl \
 	libmodule-build-perl \
 	libmodule-pluggable-perl \
 	libmojolicious-perl \
@@ -88,7 +90,7 @@ RUN apt-get update \
 	libsql-abstract-perl \
 	libstring-shellquote-perl \
 	libsub-uplevel-perl \
-        libsvg-perl \
+	libsvg-perl \
 	libtemplate-perl \
 	libtest-deep-perl \
 	libtest-exception-perl \
@@ -138,7 +140,7 @@ RUN apt-get update \
 # ==================================================================
 # Phase 3 - Install additional Perl modules from CPAN that are not packaged for Ubuntu or are outdated in Ubuntu.
 
-RUN cpanm install Statistics::R::IO Mojolicious::Plugin::NotYAMLConfig DBD::MariaDB \
+RUN cpanm install -n Statistics::R::IO Mojolicious::Plugin::NotYAMLConfig DBD::MariaDB Mojo::SQLite@3.002 \
 	&& rm -fr ./cpanm /root/.cpanm /tmp/*
 
 # ==================================================================
--- a/conf/README.md
+++ b/conf/README.md
@@ -33,6 +33,8 @@ Server configuration files.
   `webwork2.mojolicious.yml` if you need to change those settings. You usually will need to do this.
 - `webwork2.dist.service` is a systemd configuration file for linux systems that serves the webwork2 app via the
   Mojolicious hypnotoad server. If you need to change it, then copy it to `webwork2.service`.
+- `webwork2-job-queue.dist.service` is a systemd configuration file for linux systems that runs the webwork2 job queue
+  via Minion. If you need to change it, then copy it to `webwork2-job-queue.service`.
 - `webwork2.apache2.4.dist.conf` is only used if you proxy hypnotoad via apache2. Copy this to `webwork2.apache2.4.conf`
   if any changes need to be made.
 - `webwork2.nginx.dist.conf` is only used if you proxy hypnotoad via nginx. Copy this to `webwork2.nginx.conf` if any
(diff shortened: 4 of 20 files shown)


bin/check_modules.pl
```diff
@@ -112,6 +112,8 @@ =head1 DESCRIPTION
 	Locale::Maketext::Simple
 	LWP::Protocol::https
 	MIME::Base64
+	Minion
```
pstaabp, line 115: Since Minion is a dependency of Minion::Backend::SQLite, we can drop the Minion requirement since it will be picked up automatically.

pstaabp:
Trying to run as development and with the `./bin/webwork2 minion worker` I'm getting the following error:

```
Use of uninitialized value in anonymous hash ({}) at /opt/webwork/webwork2/lib/Mojolicious/WeBWorK.pm line 76.
Backend "Minion::Backend::" missing at /Users/pstaab/perl5/perlbrew/perls/perl-5.30.3/lib/site_perl/5.30.3/Mojolicious/Plugin/Minion.pm line 9.
```

All of the needed packages are installed.  Seems like a strange error .

taniwallach:
@drgrice1 Do you know how a Minion queue can be run inside the webwork Docker container? What additional installs will be needed in Docker?

drgrice1:
I forgot to think about the docker install on this one.  I will look into it.

drgrice1:
@taniwallach:  I added a commit that gets the worker queue working in the docker build.  However, it is probably not the best way to do it.  See the commit message for that last commit.

bin/check_modules.pl
drgrice1: Minion is still required in check_modules.pl.  In fact, this is another case where check_modules.pl is limited.  Really, Minion::Backend::SQLite is not required.  It is one of at least three options for backends that could be used.  See the configuration options added in this pull request in site.conf starting at line 322.

This is similar to the issue is DBD::mysql and DBD::MariaDB.  Only one is needed.  We only list DBD::mysql, and many are using DBD::MariaDB (in fact we recommend it).

drgrice1:
It is probably better to test this separately from #1909.  You can use the LTI Grade Update page to test.  You can also use the Mojolicious LMS app that I have posted elsewhere.

pstaabp:
Digging back into this.  I'm running your LMS mojolicious app and everything seems fine--able to log in and have the "LTI Grade Update" menu item up.  When trying to run the LTI Grade Update, I'm being asked to authenticate.  (is this expected.) 

Next, when I asked run the LTI Grade Update here's the log in the LMS mojolicious app: 
```
[2023-03-29 06:17:31.98612] [37708] [warn] submitting all grades for user: homer@msn.com
[2023-03-29 06:17:31.98722] [37708] [warn] lis_source_did is not available for user: homer@msn.com
[2023-03-29 06:17:32.00659] [37708] [warn] submitting all grades for user: admin
[2023-03-29 06:17:32.00709] [37708] [warn] lis_source_did is not available for user: admin
[2023-03-29 06:17:32.68658] [37708] [warn] submitting all grades for user: peter
[2023-03-29 06:17:32.69048] [37708] [warn] mass_update: Submitting grade using sourcedid: 1000 and score: 0.05
[2023-03-29 06:17:32.70322] [37708] [warn] The request was:
 <?xml version = "1.0" encoding = "UTF-8"?>
<imsx_POXEnvelopeRequest xmlns = "http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0">
  <imsx_POXHeader>
    <imsx_POXRequestHeaderInfo>
      <imsx_version>V1.0</imsx_version>
      <imsx_messageIdentifier>999999123</imsx_messageIdentifier>
    </imsx_POXRequestHeaderInfo>
  </imsx_POXHeader>
  <imsx_POXBody>
    <replaceResultRequest>
      <resultRecord>
	<sourcedGUID>
	  <sourcedId>1000</sourcedId>
	</sourcedGUID>
	<result>
	  <resultScore>
	    <language>en</language>
	    <textString>0.05</textString>
	  </resultScore>
	</result>
      </resultRecord>
    </replaceResultRequest>
  </imsx_POXBody>
</imsx_POXEnvelopeRequest>
[2023-03-29 06:17:32.70383] [37708] [warn] The nonce used is d14e4396-9aa0-53aa-9314-c9d6a023143f__eb15dccd-ce1a-11ed-bd06-a0fb2ec1196a-step2
[2023-03-29 06:17:32.70453] [37708] [warn] The response is:
 HTTP/1.1 404 Not Found
Content-Length: 24
Content-Type: text/html;charset=UTF-8
Date: Wed, 29 Mar 2023 10:17:32 GMT
Server: Mojolicious (Perl)

Invalid request endpoint
```

It appears that the grade pass back is working, but the 404 error I'm guessing shouldn't happen.

drgrice1:
I am not sure what you mean by you are being asked to authenticate when you try to run the LTI Grade Update?  Is it taking you to the webwork2 login page?  That would be something unrelated to this pull request.  One thing that just came up for me, and maybe this is what is happening for you is the following.  Did you log in as a student using the LMS app in the same browser that you are trying to use the LTI Grade Update page with as your instructor user?  When you log in as a student, it will invalidate your instructor's session.

The 404 not found is probably caused by your authen_LTI.conf settings not being right.  That is also not really due to this pull request.

This pull request doesn't change anything with authentication, either to webwork2 or for LTI.  All it does is offloads the mass update of grades and sending of instructor emails into a job queue.

somiaj (approved):
In testing this with the LTI 1.3 branch everything works. My tests were all with LTI 1.1 on that branch, but the LTI mass update page and email page both work with the minion job queue in my testing.

I did wonder if there should be any dependency setup between the two services `webwork2.service` and `webwork2-job-queue.service`?

pstaabp:
I'm still seeing errors on this, but again, I'm not running an LMS to test.  When running the minion worker, I'm getting information like: 
```
[2023-04-11 14:34:12.50819] [39394] [warn] submitting all grades for user: c
[2023-04-11 14:34:12.50902] [39394] [warn] lis_source_did is not available for user: c
[2023-04-11 14:34:12.56410] [39394] [warn] submitting all grades for user: admin
[2023-04-11 14:34:12.56521] [39394] [warn] lis_source_did is not available for user: admin
```

 I'm getting 404 errors for some users--those with actual data, that have the 404 error shown above.  Looking at the code for this, seems to be the request on Line 267 of `WeBWorK::Authen::LTIAdvanced::SubmitGrade` and the `request_url` seems to be (line 195) related to the LMS (which I don't have), so perhaps this is the problem.

drgrice1:
@pstaabp:  The first thing that you should do is test that you have things working on the develop branch (not with this pull request).  When you have that working, then test this pull request.

PR #1866, opened 2023-01-20
Fix saving a change of the permission level in the Classlist Editor.

drgrice1:
I change the parameter name of the permission select for edit mode on the Classlist Editor page, but forgot to change the handling of that parameter when "Take Action" is clicked.  As a result permissions can not be changed for users.  This fixes that.

This also fixes another similar issue on the hardcopy generator page.  The checkboxes on that page weren't working.


pstaabp (approved):
Confirmed the bugs and this fixes them.

PR #1869, opened 2023-01-25
Add a "Now" button to the flatpickr date time picker.

drgrice1:
This seems to be in demand.  See
https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8203.

This uses the shortcut-buttons-flatpickr plugin for flatpickr.  It wouldn't be to hard to implement this button without the plugin, but it is quite lightweight, so not really a big deal.

This now also adds a "Today" button.  This button changes the selected date to the current date.  It does not change the time unless the time and date were not set.  In that case it sets the time to 12:00 am and the date is still the current date.

### DIFF
--- a/htdocs/js/apps/DatePicker/datepicker.js
+++ b/htdocs/js/apps/DatePicker/datepicker.js
@@ -74,11 +74,27 @@
 					new ShortcutButtonsPlugin({
 						button: [
 							{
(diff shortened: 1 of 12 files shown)


drdrew42:
👍 for "today" button that ignores the current time.

pstaabp (approved):
Looks great now.  I like that today button!!

Issue #1870, opened 2023-01-26
"Add All" in Library Browser adds only one problem on develop


pstaabp:
As in the subject, "Add All" in Library Browser adds only one problem in the develop branch.

In the Library Browser 
1. create a new set and make it the target set.
2. Bring up a list of some problems (I've tested with both local problems and from the OPL/Contrib)
3. Click "Add All"

In the newly created set, only the first problem has been added.

pstaabp:
fixed in #1871

PR #1871, opened 2023-01-26
Fix the Library Browser add all button.

drgrice1:
The problem is that I forgot to add the `type="button"` attribute to the "Add All" input.  This does better though.  Rather than use an input for something that is handled by javascript use an actual button.  The reason an input (with type button) was used before is because CGI didn't have a real html button.

This fixes issue #1870.


pstaabp (approved):
Looks good now.

PR #1872, opened 2023-01-30
Fix the problems header row on the grades page

drgrice1:
There was a cut and paste issue for the attributes on the header.


pstaabp (approved):
fixes the header row.  Simple enough. I think I'll merge.

drgrice1:
Yeah, its simple enough.  I don't think a lot of review is needed.  Clearly the develop branch is not right on this.

PR #1874, opened 2023-01-31
Use Mojolicious::Plugin::SetUserGroup instead of authbind for serving webwork2 directly

drgrice1:
Instead of using authbind use Mojolicious::Plugin::SetUserGroup.  This will work on POSIX compliant systems.  According to the documentation for the plugin that includes Linux, OSX, and BSD (although authbind does to).  Note that you may need to force install the module as a test is failing (on my system at least).  The module still works without problem though.

This will only take effect in production mode (for instance when run via hypnotoad) and when run by the root user.  It works by switching to the chosen user and group on the next tick after the Mojo::IOLoop starts. So under hypnotoad all of the workers are not run as root, but as the chosen user and group.

Note that Mojolicious::Plugin::SetUserGroup is not a required module, and should not be added to check_modules.  It is only needed if serving directly, and not when proxying via another web server like apache2 or nginx.  The app will run fine without the module installed, and will not give any warnings or errors unless you try to run hypnotoad as root. You shouldn't do that unless you are serving directly and have the module installed anyway.

### DIFF
--- a/conf/README.md
+++ b/conf/README.md
@@ -43,7 +43,7 @@ Server configuration files.
 ## Initial configururation of webwork2
 
 - Copy `site.conf.dist` to `site.conf` and `localOverrides.conf.dist` to `localOverrides.conf`, and adjust the variables
-  in `site.conf` as needed. In particular you will need to set `$server_root_url` to the the server name, and set
+  in `site.conf` as needed. In particular you will need to set `$server_root_url` to the server name, and set
   `$database_password` to the password for the database.
 - Adjust the variables in `localOverrides.conf` to customize your server for your needs.
 - Copy any of the other `.dist` files and adjust the variables in them as needed. Note that those files will need to be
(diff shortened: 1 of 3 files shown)


conf/README.md
```diff
@@ -87,37 +87,29 @@ with Shibboleth authentication.
...
+The Mojolicious hypnotoad server will be started by the root user and the user and group will be switched to what is set
+for `server_user` and `server_group` after the app starts.  It is not advisable to run the Mojolicious hypnotoad server
+as a user that can directly login to the server.  On Ubuntu systems you can use the `www-data` user that is already
+available. If a user is needed you can create the user `webwork`, for example, with `sudo useradd -M webwork`.  Make
+sure that the user has read access to the SSL certificates given in the configuration above if using certificates.
+Usually the user and group will be the same.
 
 Then set up the systemd service:
 
 - Copy `webwork2.dist.service` to `webwork2.servcice`.
```
taniwallach, line 111: I think there is a typo in `webwork2.servcice` which should be `webwork2.service`.

drgrice1: Thanks @taniwallach.  I fixed that.

Alex-Jordan:
Do you need to add `Mojolicious::Plugin::SetUserGroup` to `check-modules`?

Alex-Jordan:
Oops:

> Note that Mojolicious::Plugin::SetUserGroup is not a required module, and should not be added to check_modules. 

:)

Alex-Jordan:
With Oracle 8.7

* Successfully installed `Mojolicious::Plugin::SetUserGroup` (I did not have any tests fail).
* `mv webwork2.mojolicious.yml webwork2.mojolicious.yml.bak`
* `cp webwork2.mojolicious.dist.yml webwork2.mojolicious.yml`
* edit `webwork2.mojolicious.yml` as described in `conf/README.md`. I used `apache` instead of `www-data`.
* `mv webwork2.service webwork2.service.bak`
* Comment out the `User`, `Group`, and `Environment` settings in the copy.
* `sudo systemctl daemon-reload`
* `sudo systemctl stop webwork2`
* `sudo systemctl start webwork2`

After these steps, everything seems fine visiting the site from the web. Is there anything I should do to verify it's working as intended without authbind? Whatever I had previously done on this system to use authbind is still there, minus the shuffling of config files I did outlined above.

drgrice1:
The issue with tests failing when installing modules from CPAN is an Ubuntu 22.04 thing.  There are numerous modules for which the tests fail when installing from CPAN on Ubuntu 22.04.  They have applied some patches to some base perl module somewhere that is causing issues.  I suspect this is probably related to why the libdbd-mariadb-perl package (containing the DBD::MariaDB module) is not building.  I have seen this now on multiple computers with Ubuntu 22.04 installed.

If `authbind --deep` is not in the `webwork2.service` file (you copied anew from the dist file, so it wouldn't be), then authbind will not be in effect.  So if the service started, and you were able to visit the site from the web, then it is working as intended.  The service won't even start if it isn't working.

drgrice1:
Actually, I think that what I said was wrong.  The service will start even if it isn't working right.  To check that it is working correctly run `ps aux | grep webwork2`.  You should see a list of jobs with the command "/opt/webwork/webwork2/bin/webwork2".  The user for all of these jobs should be "apache" except for the first job which will be run under the user "root".

drdrew42 (approved):

dlglin:
Mojolicious::Plugin::SetUserGroup needs to be added to check_modules.pl

Alex-Jordan:
I said the same thing earlier in the thread. Then I read the opening post.

dlglin:
Sorry, my bad.

I'm thinking about how to beef up check_modules to handle things like this, so we may be able to have it tell people about when optional modules would be needed.

Issue #1873, opened 2023-01-30
safari and datepicker issue


Alex-Jordan:
In Safari, if you go to the Hmwk Sets Editor, Edit tab, and click to edit a set, there is a console error from  the datepicker:

```
RangeError: timeZoneName must be "short" or "long"
```

This is leading to the times in the input fields being filled in as epoch timestamps instead of human readable dates. I do not see this error in Fiirefox, Chrome, or Edge for Mac. This was reported by a Runestone hosting user.

drgrice1:
Does it also happen on the Hmwk Sets Editor page when not in edit mode?  Even if you click on the "Import" tab?
Does it also happen on the Set Detail page?
What about on the UserDetail page (the "Sets assigned to ...." page)?

It would be really odd if it only happens on the Hmwk Sets Editor page, and not the pages I asked about above.  They all use the same code, and `timeZoneName` is set to `shortOffset` in this code.  That is a valid value for `timeZoneName` according to the specifications.

Also, can you reproduce this on the develop branch?

dlglin:
I can't reproduce this.  I'm running Safari 16.3 (18614.4.6.1.5).
<img width="1723" alt="Screenshot 2023-01-30 at 11 03 48 AM" src="https://user-images.githubusercontent.com/1794586/215558413-802dd7af-7ee2-4d6c-9cf6-6142d0904fbc.png">

Alex-Jordan:
> Does it also happen on the Hmwk Sets Editor page when not in edit mode? Even if you click on the "Import" tab?

The console error happens. But you wouldn't notice, because nothing is odd about the dates you see. They are human readable.

>  Does it also happen on the Set Detail page?

Yes. Console error, and dates as timestamps.

>  What about on the UserDetail page (the "Sets assigned to ...." page)?

The console error is there. At first, no timestamps to see because the input fields are blank. If I put a date in the filed and try to save, there is an apache warning: `Argument "1/31/23, 11:59 PM" isn't numeric in numeric lt (<) at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm line 1203.` and an apache error `DBD::MariaDB::st execute failed: Data truncated for column 'open_date' at row 1 at /opt/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm line 881.`

> Also, can you reproduce this on the develop branch?

[...]

PR #1875, opened 2023-01-31
Change to a new method for computing the server/browser time zone adjustment

drgrice1:
This method is simpler and just uses Date object methods directly and not the Intl.DateTimeFormat method.

This should fix issue #1873.  Safari users please test.  Most importantly someone using an older version of Safari.

### DIFF
--- a/htdocs/js/apps/DatePicker/datepicker.js
+++ b/htdocs/js/apps/DatePicker/datepicker.js
@@ -46,15 +46,14 @@
 
 			luxon.Settings.defaultLocale = rule.dataset.locale ?? 'en';
 
-			// Compute the time difference between the current browser timezone and the the course timezone.
+			// Compute the time difference between the current browser timezone and the course timezone.
 			// flatpickr gives the time in the browser's timezone, and this is used to adjust to the course timezone.
-			// Note that this is converted to seconds.
-			const timezoneAdjustment =
-				parseInt(Intl.DateTimeFormat('en-US', { timeZoneName: 'shortOffset' })
-					.format(new Date).split(' ')[1].slice(3) || '0') * 3600000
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
The results seem to be the same to me (Running Safari 16.2--not as up to date as @dlglin ).

Also, I did test that I put the server in a different time zone than my local time and saw the time adjustment.

Alex-Jordan:
I tested, and it mostly fixes #1873. I'm not 100% sure though.

Specifically, I used my older Safari to visit my 2.17 development server (running unadulterated 2.17). I confirmed the issue was still there where I see a timestamp instead of a date when I go to edit a set's dates.

Then I applied this patch, ran npm install, restarted apache, and cleared the cache in Safari. And now I see dates instead of timestamps, good!

However I still see a console error. It's not present on the page where I can actually input dates. It is a console error when I first visit the Hmwk Sets Editor page. The error message is:
```
RangeError: timeZoneName must be "short" or "long"
```
It comes from:
https://webwork-test.pcc.edu/webwork2_files/js/apps/ProblemSetList/problemsetlist.03d2245f.min.js

Is this to be expected? Is the patch not yet applied to some place in a minimized js file in such a way that matters on the Hmwk Sets Editor page, but not the actual page where you edit dates?

drgrice1:
@Alex-Jordan:  Ahh, yes. I forgot to make the change in the other file that uses the same technique.  I will add that.  That

drgrice1:
That is now added.  You can do the same thing for problemsetlist.js.  This is also the only change from the 2.17 version of that file.

Alex-Jordan:
OK, great. I tested again and that console error is gone.

Alex-Jordan (approved):

PR #1878, opened 2023-02-01
Convert LaTeX hardcopy to use the exam document class

drdrew42:
Note: there is a [small change required in PG](https://github.com/openwebwork/pg/pull/771) paired with this PR

## Description

In an attempt to further clean up and enhance the hardcopy generation feature in WeBWorK, it makes sense to use the 'exam' document class in LaTeX. I believe that there is more that can be done in terms of including solutions or incorporating student responses more clearly than is currently done -- but this initial conversion at least allows for the following:
* Customization of header/footer text, both of which are automatically broken into a left-center-right configuration
* Customization of problem label and corresponding point value
* Inclusion of scoring tables, aggregated either by individual question or per-page

The new defaults in `ASimpleCombinedHeaderFile.pg` should be visually consistent with the current default in WeBWorK. I have also tested to make sure that old versions of the file still compile and haven't substantively changed with regards to PDF output. Confirmation is requested.

### Caveats for those familiar with the exam class

This specific implementation of the exam class relies on `\titledquestion` and `\qformat`, which does *not* support any iteration of `\pointsinmargin`. It is necessary to take this approach in order to maintain the existing functionality of sets with problem numbers that are non-sequential (either by deleting without re-sequencing, or in the case of nested problems in J-I-T sets).

### Unresolved Issues

The exam class makes the command `\gradetable` available, which (when used) requires `pdflatex` to be run three times (something to do with building and processing the hardcopy.aux file) in order to render the table. This can be accomplished by wrapping Hardcopy.pm:L749-76 in a for-loop, but there is probably a better approach.

Not an "issue" per se, but with this PR, the only difference between twoColumn and oneColumn formats is the declaration of `\newcommand{\nocolumns}{}`. Do we need to continue having separate folders for one/two column formats if they differ by only one line of code?

And building off of the previous point, the use of the `\gradetable` will likely need to happen in the hardcopySetFooter, which would require *its* own theme directory. I don't know if there's interest in building out additional hardcopy configurations options that are not folder-based.

Finally, I did not address the XeLaTeX formats, though I would expect the necessary changes to be very similar.

### Documentation Needs

When updating to the next release of WeBWorK, the modelCourse **ought** be updated with the new versions of `ASimpleCombinedHeaderFile.pg` and `ASimpleHardcopyHeaderFile.pg`, though this is not mandatory. Old header files _should_ continue to work as is.


drdrew42:
Also, R.I.P. hardcopySetHeader.pg in the theme directories -- you had a good 20-year run, but no one can even use you at this point.

pstaabp:
Out of the box, I'm getting the following error:
```
 LaTeX Error: Option clash for package graphicx.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.11   \usepackage
                  {epstopdf} % allows use of eps files with pdftex
The package graphicx has already been loaded with options:
  []
There has now been an attempt to load it with options
  [pdftex]
Adding the global options:
  ,pdftex
to your \documentclass declaration may fix this.
```

I don't understand this error, because line 10 of `packages.tex` is `\usepackage[pdftex]{graphicx}` which is what the above error recommends.   Wondering if others are seeing this.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -1180,24 +1182,23 @@ async sub write_problem_tex ($c, $FH, $TargetUser, $MergedSet, $problemID = 0, $
...
-			my $id       = $MergedProblem->problem_id;
-			my $prettyID = join('.', jitar_id_to_seq($id));
-			print $FH "{\\bf " . $c->maketext("Problem [_1].", $prettyID) . "}";
-		} elsif ($MergedProblem->problem_id != 0) {
-			print $FH "{\\bf "
-				. $c->maketext("Problem [_1].", $versioned ? $versioned : $MergedProblem->problem_id) . "}";
+			$id = join('.', jitar_id_to_seq($id));
+		} elsif ($id != 0 && $versioned) {
+			$id = $versioned;    # this cannot be right?
 		}
```
drgrice1, line 1191: As wrong as it looks, it is correct.  The value of `$c->{versioned}` is originally set to be the version number, but it is later set equal to the problem number on line 974.  That is done before each call to `write_problem_tex`.  It is done this way because the MergedProblem->problem_id is not correct for randomized gateway tests.  That is the problem_id in the template, but not the order on the test in that case.

Why was the `versioned` variable used?  I guess the original idea was to recycle that variable since it just needs to be non-zero for all other uses in the module.

drgrice1: Note git blame shows that this was how @glarose originally implemented this for gateway quizzes.

conf/snippets/ASimpleCombinedHeaderFile.pg
```diff
@@ -84,7 +75,46 @@ BEGIN_PGML
...
+\pagestyle{headandfoot}
+\firstpageheader{
+	{\large \bf $studentName} \\
+% Uncomment the line below if this course has sections.
+%	{\large \bf Section: \{protect_underbar($sectionName)\}} \\
+	{\large \bf Assignment \{protect_underbar($setNumber)\} due $formattedDueDate}
+}{}{
+	{\large \bf \{protect_underbar($courseName)\} \\ {} }
+}
+\ifdefined\nocolumns\firstpageheadrule\fi
```
drgrice1, line 100: The addition of `\firstpageheadrule` here seems to cause the first page to have two head rules when using the single column theme.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -1180,24 +1182,23 @@ async sub write_problem_tex ($c, $FH, $TargetUser, $MergedSet, $problemID = 0, $
...
 
 		my $problemValue = $MergedProblem->value;
 		if (defined($problemValue)) {
-			my $points = $problemValue == 1 ? $c->maketext('point') : $c->maketext('points');
-			print $FH " {\\bf\\footnotesize($problemValue $points)}";
+			print $FH "\\titledquestion{$id}[$problemValue]\n";
```
drgrice1, line 1195: One problem with this is that previously `point` amd `points` were translated.  Now this will require foreign languages to use a custom hardcopy header file.  Although there were untranslated strings in the previous header files, so that was probably already done.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -1180,24 +1182,23 @@ async sub write_problem_tex ($c, $FH, $TargetUser, $MergedSet, $problemID = 0, $
...
 			# Use the pretty problem number if its a jitar problem
-			my $id       = $MergedProblem->problem_id;
-			my $prettyID = join('.', jitar_id_to_seq($id));
-			print $FH "{\\bf " . $c->maketext("Problem [_1].", $prettyID) . "}";
-		} elsif ($MergedProblem->problem_id != 0) {
-			print $FH "{\\bf "
-				. $c->maketext("Problem [_1].", $versioned ? $versioned : $MergedProblem->problem_id) . "}";
+			$id = join('.', jitar_id_to_seq($id));
+		} elsif ($id != 0 && $versioned) {
+			$id = $versioned;    # this cannot be right?
```
drgrice1, line 1190: This approach has the same translation issue as the points.  Now the only way for the `Problem` text to be translated is via a custom hardcopy header file.

drgrice1 (approved):
This pull request looks good.  Just noting some changes and suggestions.

One change that I notice is that previously with the two column theme, the footer was displayed at the end of the last page.  Now it is displayed in the column at the end of the last problem.

Perhaps the copyright statement should be moved into the runningfooter in the header file, and the `hardcopySetFooter.pg` file deprecated.   This would make it so that instructors can customize that, instead of having it hardcoded.  Is it really correct to have this copyright in the hardcopy as it is?  The content of the exam might not really be webwork's content.  It could be that the problems were created by someone else.  It makes sense in html, because then you can claim ownership of the page (but necessarily all content).  In hardcopy, this is not so true.

drgrice1:
By  the way, I am not getting the option clash for the graphicx package that @pstaabp saw.

drgrice1:
I just noticed one change that I don't like.  You added `\setlength{\columnseprule}{.4pt}` to the one column preamble.  That causes `multicols` used within a problem to have the column separator rule.  Why is that needed for the one column preamble?

drgrice1:
Do you know why the one column theme uses an 11 point font, and the two column theme uses a 10 point font?  It seems that those should be the same.  This is not a change in this pull request, I have noticed this before, and wondered why it is that way.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
drdrew42: > Although there were untranslated strings in the previous header files, so that was probably already done.

My thoughts exactly.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
drdrew42: Yeah, I couldn't make sense of this... but it still worked when creating versioned hardcopies ¯\_(ツ)_/¯ 

I guess I should delete the comment.

drdrew42:
> One change that I notice is that previously with the two column theme, the footer was displayed at the end of the last page. Now it is displayed in the column at the end of the last problem.

Yes, this is one of the minor changes that I could not quite resolve nicely. I ended up throwing in the towel on maintaining consistency. I am also on the same page with your suggestion that perhaps this is could be put in the default footer... there's a config for headers/footers to identify being on the last page. Putting it in the footer has the advantage of allowing for modification by any problem set.

> I just noticed one change that I don't like. You added `\setlength{\columnseprule}{.4pt}` to the one column preamble. That causes `multicols` used within a problem to have the column separator rule. Why is that needed for the one column preamble?

That was an oversight. I was testing how similar I could get the oneColumn and twoColumn preambles... I didn't end up testing with any problems that made use of the multicols envir. This is definitely unnecessary for the oneColumn format.

> Do you know why the one column theme uses an 11 point font, and the two column theme uses a 10 point font? It seems that those should be the same. This is not a change in this pull request, I have noticed this before, and wondered why it is that way.

Nope, but I'm sure there's a historical justification. Perhaps related to fitting problems into narrow columns? IDK, but we could standardize them to use the same font.

pstaabp:
I'm still having those same troubles.  Even saving the hardcopy.tex files (as well as packages.tex, CAPA.tex, PGML.tex), removing the first line `\batchmode` and running my standard way, getting the same errors.  Might be my setup, but a comment https://tex.stackexchange.com/questions/512935/option-clash-for-package-graphicx seems that [pdftex] is no longer needed for graphicx (I think the comment comes from the author of graphicx ).   Removing that did squash that error, but still have other errors to track down.

@drdrew42 you also mentioned about not using different sets of templates for one and two column formats.  Can we leverage the `onecolumn` and `twocolumn` options in the `geometry` package?  I don't think I've used that, but may simplify things.

drgrice1:
The comment regarding the `[pdftex]` option being needed on that stack exchange post was not made by the author of the package.  The authors of the package are David Carlisle and Sebastian Rahtz.  David Carlisle does make another comment on the post, but not that one.  Nonetheless, the comment is correct.  The option is not needed anymore.

As to the onecolumn versus twocolumn issue, changing to using the geometry package ... well @drdrew42 has already done that in this pull request (well technically the paired pg pull request).

@drdrew42:  There is another thing that is needed for this pull request though.  I noticed that you didn't update the other themes (the XeXLaTeX themes).  Those will also need to be updated for this.  Also, I think there are some changes to `bin/check_latex.tex` that are needed.

drgrice1:
Actually, what @drdrew42 does in pg is not using the geometry package twocolumn option, but is what the geometry twocolumn option does internally.  Well essentially.  The geometry twocolumn option sets the \twocolumn switch for the entire document which is actually not what we want.  We don't want it for the header.  So @drdrew42 turns it off for that, and then back on again after the header.

drdrew42:
Precisely, it has the same effect whether or not twocolumn is set in the geometry `\usepackage` or not.

I'll copy over the changes into the XeTeX files and make sure they work, and then tackle the check_latex script.

drgrice1:
The thing is that we need a flag to tell the tex compiler to use two column mode or not.  So the twocolumn option of the geometry package doesn't change anything in regards to having the separate onecolumn and twocolumn themes.  To merge the two themes we would have to inject the option into the tex output via perl code.  That could be done.  However, I am not sure it would go with the webwork hardcopy theming setup.  The current theming setup allows much more than just changing the number of columns.  Injecting the flag via code would have to be determined by a separate option, and could mess with the true theme selection capability.

pstaabp:
I successfully have this running.  I had another `exam.cls` file in the path and it wasn't loading the class file correctly. 

> I'll copy over the changes into the XeTeX files and make sure they work, and then tackle the check_latex script.

I'm assuming we are supporting XeTeX for the unicode/different font support.  It seems like luatex/lualatex is a bit better at this now.  I use lualatex a lot for doing scripting inside of latex documents.   We could look into that--just to add something else to do. :)

taniwallach:
@pstaab From what I can see lualatex supports Polyglossia, https://github.com/reutenauer/polyglossia/ so I think it seems quite reasonable to assume that foreign languages will work well with lualatex instead of xelatex. I would say that supporting either one to get Unicode support suffices.

There is a thread on lualatex vs xelatex at https://www.reddit.com/r/LaTeX/comments/x9acv3/which_would_you_recommend_xelatex_or_lualatex/ and my impression is that the recommendation there is to use lualatex.

drgrice1:
I am seeing a few indentation and line breaking issues with this.  The following images show the issues I am seeing.  These are all generated in single column mode and not showing the source file path.

The first problem is Library/Hope/Multi1/03-01-Vector-spaces/Proof_04.pg.  The first picture is a hardcopy generated the current develop branch.  Make sure to not include answers since those are currently broken for drag and drop problems.
exhibit1
The second is with this branch generated the same way.
exhibit2

The next problem is one of mine.  This is generated with answers. Again first the develop branch result:
exhibit3
Now with this branch:
exhibit4

I realize the exam class indents things, but perhaps that could be overridden?  It certainly doesn't look the best.  Although the issue with the first line of the Hope problem is a bigger problem.  The "Statement" to the far right on the same line as the problem number.

drgrice1:
Note that I am seeing the same issues with two column mode.

One solution is to do as I suggested and remove the exam class indentation.  This can be done by adding the following to the hardcopyPreamble.tex files for the hardcopy themes.
```
\renewcommand{\questionshook}{\leftmargin=0pt\labelwidth=-\labelsep}
\renewcommand{\partshook}{\setlength{\leftmargin}{\leftmargini}%
  \setlength{\labelwidth}{\dimexpr \leftmargin-\labelsep}}
\renewcommand{\subpartshook}{\setlength{\leftmargin}{\leftmargini}%
  \setlength{\labelwidth}{\dimexpr \leftmargin-\labelsep}}
```
I found this at https://tex.stackexchange.com/questions/568225/changing-indentation-in-exam-class.  I haven't tested it extensively, but it seems to work well with what I have tested.

drgrice1:
@drdrew42:  I rebased this onto the develop branch.  So make sure that you update your local clone.  I am going to try to work on updating the other themes, and it helps to have the branch up to date with develop.

drgrice1:
@drdrew42: I put a pull request in to your branch for this pull request that updates the XeLaTeX files, and a few other minor changes.

taniwallach (changes requested):
I tested mainly that the xelatex option works reasonably well after the changes in https://github.com/drdrew42/webwork2/pull/5 and things seem OK in the Hebrew formats with those changes. I made a PR to @drgrice1 branch to make some minor fixes for the Hebrew output. With our without my small changes, I think @drgrice1's changes should be merged and then this PR should be ready to merge.

drgrice1:
@taniwallach:  I merged your pull request into mine.

taniwallach (approved):
Did not test again. About points and translations - it will need to be handled by custom header files. Eventually I'll do that for Hebrew.

Alex-Jordan:
This PR removed `conf/snippets/hardcopyThemes/twoColumn/hardcopySetHeader.pg`, which I assume is OK. However it left in place `hardcopyThemes/twoColumn/hardcopySetHeader.pg`, which is a symlink to the above. So now it's a broken symlink. Can I confirm it would be OK to remove this symlink?

PR #1881, opened 2023-02-03
Fix an incorrect route name on the stats page

drgrice1:
The problem links on a set statistics page were not going to the correct place because the route name was wrong.  The prolem links were going to the general set stats page instead of the problem stats page because the instructor_set_statistics route was accidentally used instead of the instructor_problem_statics route.

To see the problem on the develop branch open the stats page for a set, and click on the problem numbers in either the "Percent of Users" table near the end.  That will just open the page you are currently viewing instead of the stats page for the problem you click on.

There was also a typo in the title of the problem stats page from before that is fixed.


pstaabp (approved):
Fixes the issue.

PR #1882, opened 2023-02-03
Fix a typo in WeBWorK/Utils/Rendering.pm.

drgrice1:
This typo has the effect that if the course settings are to use no entry assist method, then when rendering a problem in the homework sets editor or in the pg problem editor (or anything that uses WeBWorK/Utils/Rendering.pm), then instead of no entry assist, you get wiris.


pstaabp (approved):
Wiris is sneaking in before we have a chance to deprecate it. 😄 

This fixes that.

drgrice1:
Yeah, time for that to go.

Although, part of the problem is the odd way that we have the `entryAssist` option set up to work.  We have separate settings for each instead of a single setting that controls the selection.

PR #1884, opened 2023-02-04
Rework flags handling for essay type answers

drgrice1:
This rework caused a chain reaction that led to other related changes.  Here is the summary of these changes.
    
Instead of saving global problem flags in the form "essay," and user problem flags in the form "graded," or "needs_grading," which gives no real information about what answers the problem has, save the type for all answers.  This is saved as a comma separated list to the user problem.  If an answer does not have a type just save it as the empty string.  If there are essay type answers in the problem, then ":needs_grading" is appended to the end of the comma separated list of answer types.  That is removed after the essay answers in the problem are graded.  Nothing is stored in the global problem flags column.  The reason that it is all saved in the user problem is that different students can have different answer types in certain cases (usually for a gateway test with problems drawing from a grouping set).

In addition, the only use of the global flag previously was to determine if a problem is gradable.  Now, all problems are gradable.  The "Grade Problem" link appears for all problems on the problem set list page for non-gateway assignments.  The grader link also appears for all problems on the problem set detail page (with the render problem, edit problem, and open in new window buttons), and again on the stats page for the set.

The past answers table has been revised so that for courses created with this in effect no problems are rendered.  The flag is used to get answer types instead, and correctly display them for all users.  There is a compatibility layer that basically does the same thing as before if that flag has a value that appears to be in the old format (they all ended with a comma, the new format never does).  The old method is improved however.  As long as students all have the same answer types the problems will still be rendered correctly.  Also, gateway problems that draw from a grouping set will work as long as the problems in the grouping set have essay answers in the same answer order.  Note grouping set problems did not work at all previously.  Answers were displayed but essay answers were not displayed correctly.

The biggest change in this pull request is a complete rework of the old manual problem grader.  That now works for all types of sets including gateway tests.  In that case the problems from all versions are shown.

The page now correctly renders problems (not just showing the problem text).  The problem that renders initially is the problem for the first student in the list.  The problem for any student can be rendered instead by choosing that student from a select field.

You can still click on a students name to open a particular student's problem in a new window.  In addition, for gateway tests the problem is opened in the actual test version for that student and on the correct page of the test.  This technique was also used to make this work on the set detail editor page when editing a set version for a student as well.  This was requested previously (by @Alex-Jordan I believe).  Furthermore, on the manual grader page, when a set is opened in a new window the single problem grader is already active and answers, solutions, and hints shown.

The rpc javascript rendering method is separated into a new file.  That is used by the problem grader, stats problem page, the library browser, and the problem set detail page.  The PG problem editor has its own code for now as it does some additional handling that doesn't fit into that code.  Note that there were some id/name elements that needed to be made uniform across various pages for this to work (which led to minor modifications to the tag widget code).

There are a few changes to the stats page.

For problem rendering the missing language and display mode elements that were looked for in the javascript were added.  Also, instead of having a show/hide button for the problem, just render it when the page loads.  If the display mode is changed with the new select button, then the problem is re-rendered.

The siblings when viewing problem statistics was incorrect.  If viewing statistics for a problem the siblings should be the other problems in the set.  Not all of the sets.  So that is fixed.

Note that this depends on #1880.


drdrew42:
Just running through this quickly, things look good -- though I don't have a good sample set for testing a gateway with grouped problems.

Other than that, the manual grader is working as described, as is the statistics page and both forms of set details.

My only suggestion for possible improvement is to include the other problems in the set as siblings for the sidebar when working in the manual problem grader... as is, there isn't a convenient way to navigate between problems

drgrice1:
I have added other problems as siblings to the manual problem grader.

drdrew42 (approved):
Looks great!

pstaabp:
Starting to dive into this.  On first glance, in the problem grader, I'm seeing:
image

where the rightmost button is a little larger.  Was this intentional?  It looks like the previous two buttons have the `mb-1` class added and this doesn't.

pstaabp:
The above isn't in the problem grader.  It's in the viewing a problem as a student. And it looks like it might be in develop as well.

pstaabp (approved):
I honestly haven't used essay problems much in classes, so am not very familiar with the previous version.  What I see looks good.

drgrice1:
I added the margin to that submit button @pstaabp.

PR #1885, opened 2023-02-04
Remove a FIXME and its related code that were not supposed to still be here

drgrice1:
This was supposed to have been removed before #1857.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -156,19 +156,6 @@ async sub go ($c) {
 	return $c->header;
 }
 
-=item r()
-
-Returns a reference to $c (this object) which is a WeBWorK::Controller object.
-
-FIXME: This method will be removed once all ContentGenerator modules
-are converted.
-
-=cut
-
-sub r ($c) {
-	return $c;
-}
-
 =item do_reply_with_file($fileHash)
 
 Handler for reply_with_file(), used by go(). DO NOT CALL THIS METHOD DIRECTLY.


drgrice1:
@pstaabp:  This is why your code in #1879 on line 459 of WeBWorK/Utils/ProblemProcessing.pm is working.  I thought I had deleted this temporary code used while converting from the old `$r` usage to the new `$c` usage.  Note that all it does is return the `$c` object it is called with and should not be used.

drgrice1:
Why is this failing the perltidy check?

drgrice1:
It seems that something has changed with one of the workflow dependencies.  I am guessing it is `actions/checkout@v3` based on the error.  I think it is now checking out a treeless clone.  So the `git diff` command that checks to see if `perltidy` changed anything is giving the error that `git` is not being run from inside a git repository.

pstaabp (approved):

pstaabp:
We should merge in #1886 before this one to show that the check will now work.

drgrice1:
Yeah, #1886 should be merged before anything else in fact.  The same goes for https://github.com/openwebwork/pg/pull/773 for the pg code.  Until those are in, all of the perltidy checks are null and void.  So merging something else could result in code added that is not perltidy, and that will make later checks fail for later pull requests even though it is not their fault when these are merged.

pstaabp:
Why don't I just merge these two.

drgrice1:
Yes, #1886 and https://github.com/openwebwork/pg/pull/773 should just be merged.  The evidence that they work is in the fact that the workflow succeeded for them.

PR #1887, opened 2023-02-06
Update the tag widget.

drgrice1:
The javascript is updated with javascript that does not use jQuery.

Furthermore, the old widget that was directly displayed in the page and was ugly as hell has been replaced with a button.  When the button is clicked a dialog opens with a dialog which contains the tag editor.

If a problem has invalid tags, instead of those intrusive native alert dialogs a toast is shown.  The user can continue to interact with the page even while the toast is present.  The toast disappears after five seconds on its own.

Note for self:  This has conflicts with #1884.

### DIFF
--- a/htdocs/js/apps/TagWidget/tagwidget.js
+++ b/htdocs/js/apps/TagWidget/tagwidget.js
@@ -293,7 +293,8 @@
 
 			if (currentValue && !select.value) {
 				showMessage(`Provided ${category} "${currentValue}" is not in the taxonomy.`);
-				select.add(new Option(currentValue, currentValue, false, true));
+				select.remove(0);
+				select.add(new Option(currentValue, currentValue, false, true), 0);
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


drgrice1:
I though of adding labels.  I will do that.

The dialog is way better.  There is no need to have the tags in the page.  The row will never work nicely.  That is simply impossible.  This is the same problem your current design of the webwork3 library browser has.  You just cannot fit long text like this into a single line.  If you want to see the tags, simply click on the button.  It opens quickly and there you have it.

drgrice1:
By the way, as to the labels, this is the same as it was before without the labels.

drgrice1:
I added labels to the selects in the dialog.  Also, I eliminated the nonsensical "All ..." option that was used for the default option in each select to mean no subject is selected.  Instead "Select ..." is used.

pstaabp (approved):
Looks good with the labels

drdrew42:
This is great. I am no longer bombarded with alerts when a problem isn't properly tagged...

My only suggestion for improvement is to replace the "Select _____" option with the invalid subject/chapter/section at the top of the dropdown list instead of appending it as the last option.

drdrew42 (approved):

drgrice1:
@drdrew42:  I made your suggested change.  Invalid tags are now added at the beginning and the 'Select ...' option removed.

PR #1888, opened 2023-02-10
Enforce user permissions on Classlist Editor page

drdrew42:
## Description

Perhaps better described as a bugfix than a feature, it is time to prevent the modification/deletion of administrative users by users with non-admin privileges.

In order to accomplish this, several safeguards have been put into place.
1. Accounts with higher permissions than the current user will not have a checkbox, preventing their inclusion in any action using the 'selected' scope.
2. Accounts with higher permissions than the current user will not have an edit icon.
3. Accounts with higher permissions than the current user will not appear when attempting an 'edit' with the 'all' scope.
4. Accounts with higher permissions than the current user display a warning message when attempting a 'password' action with the 'all' scope.
5. An error should be thrown if a valid request is made that circumvents all of the above.

### Bonus

~~Somehow the `text-end` css class was omitted from the login-status column of the math4 theme, it's been added to match the other themes. ***Do we want to add `my-auto` to this column as well?*** This change would vertically center the login-status content.~~

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
@@ -241,7 +241,8 @@ sub pre_header_initialize ($c) {
 	$c->{allUserIDs}    = [ keys %allUsers ];
 	$c->{sortedUserIDs} = [
 		map  { $_->user_id }
-		sort { &$primarySortSub || &$secondarySortSub || &$ternarySortSub } (values %allUsers)
+		sort { &$primarySortSub || &$secondarySortSub || &$ternarySortSub }
+		grep { $c->{visibleUserIDs}{ $_->user_id } } (values %allUsers)
 	];
 
 	return;
@@ -561,7 +562,7 @@ sub save_password_handler ($c) {
 	}
 
 	if (defined $c->param('prev_visible_users')) {
-		$c->{visibleUserIDs} = $c->every_param('prev_visible_users');
+		$c->{visibleUserIDs} = { map { $_ => 1 } @{ $c->every_param('prev_visible_users') } };
 	} elsif (defined $c->param('no_prev_visible_users')) {
 		$c->{visibleUserIDs} = {};
 	}
@@ -611,7 +612,7 @@ sub importUsersFromCSV ($c, $fileName, $createNew, $replaceExisting, @replaceLis
 	my $db   = $c->db;
 	my $dir  = $ce->{courseDirs}->{templates};
 	my $user = $c->param('user');
-	my $perm = $db->getPermissionLevel($user)->permission;
+	my $perm = $c->{userPermission};
 
 	die $c->maketext("illegal character in input: '/'") if $fileName =~ m|/|;
 	die $c->maketext("won't be able to read from file [_1]/[_2]: does it exist? is it readable?", $dir, $fileName)
--- a/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
@@ -140,8 +140,9 @@ sub pre_header_initialize ($c) {
 
 	return unless $authz->hasPermissions($user, 'access_instructor_tools');
 
-	$c->{editMode}     = $c->param('editMode')     || 0;
-	$c->{passwordMode} = $c->param('passwordMode') || 0;
+	$c->{userPermission} = $db->getPermissionLevel($user)->permission;
+	$c->{editMode}       = $c->param('editMode')     || 0;
+	$c->{passwordMode}   = $c->param('passwordMode') || 0;
 
 	return if ($c->{passwordMode} || $c->{editMode}) && !$authz->hasPermissions($user, 'modify_student_data');
 
@@ -375,21 +376,28 @@ sub delete_handler ($c) {
 	my $db    = $c->db;
 	my $user  = $c->param('user');
 	my $scope = $c->param('action.delete.scope');
+	my $perm  = $c->{userPermission};
 
 	my @userIDsToDelete = ();
 	if ($scope eq 'selected') {
 		@userIDsToDelete = @{ $c->{selectedUserIDs} };
 	}
 
-	my %allUserIDs      = map { $_ => 1 } @{ $c->{allUserIDs} };
-	my %visibleUserIDs  = map { $_ => 1 } @{ $c->{visibleUserIDs} };
-	my %selectedUserIDs = map { $_ => 1 } @{ $c->{selectedUserIDs} };
+	my %allUserIDs      = map { $_            => 1 } @{ $c->{allUserIDs} };
+	my %visibleUserIDs  = map { $_            => 1 } @{ $c->{visibleUserIDs} };
+	my %selectedUserIDs = map { $_            => 1 } @{ $c->{selectedUserIDs} };
+	my %permissionbyID  = map { $_->{user_id} => $_->{permission} } @{ $c->{visibleUsers} };
 
-	my $error = '';
-	my $num   = 0;
+	my @resultText;
+	my $num = 0;
 	foreach my $userID (@userIDsToDelete) {
 		if ($user eq $userID) {    # don't delete yourself!!
-			$error = $c->maketext('You cannot delete yourself!');
+			push @resultText, $c->maketext('You cannot delete yourself!');
+			next;
+		}
+
+		if ($permissionbyID{$userID} > $perm) {
+			push @resultText, $c->maketext('You cannot delete someone with permissions higher than your own.');
 			next;
 		}
 		delete $allUserIDs{$userID};
@@ -403,7 +411,8 @@ sub delete_handler ($c) {
 	$c->{visibleUserIDs}  = [ keys %visibleUserIDs ];
 	$c->{selectedUserIDs} = [ keys %selectedUserIDs ];
 
-	return $c->maketext('Deleted [_1] users.', $num) . ($error ? " $error" : '');
+	unshift @resultText, $c->maketext('Deleted [_1] users.', $num);
+	return join(' ', @resultText);
 }
 
 sub add_handler ($c) {
@@ -493,8 +502,7 @@ sub cancel_edit_handler ($c) {
 
 sub save_edit_handler ($c) {
 	my $db                   = $c->db;
-	my $editorUser           = $c->param('user');
-	my $editorUserPermission = $db->getPermissionLevel($editorUser)->permission;
+	my $editorUserPermission = $c->{userPermission};
 
 	my @visibleUserIDs = @{ $c->{visibleUserIDs} };
 	foreach my $userID (@visibleUserIDs) {
@@ -502,6 +510,9 @@ sub save_edit_handler ($c) {
 		die $c->maketext('record for visible user [_1] not found', $userID) unless $User;
 		my $PermissionLevel = $db->getPermissionLevel($userID);
 		die $c->maketext('permissions for [_1] not defined', $userID) unless defined $PermissionLevel;
+		# delete requests for elevated users should never make it this far
+		die $c->maketext('insufficient permission to edit [_1]', $userID)
+			unless ($editorUserPermission >= $PermissionLevel);
 		foreach my $field ($User->NONKEYFIELDS()) {
 			my $param = "user.$userID.$field";
 			if (defined $c->param($param)) {
@@ -547,6 +558,9 @@ sub save_password_handler ($c) {
 	foreach my $userID (@visibleUserIDs) {
 		my $User = $db->getUser($userID);
 		die $c->maketext('record for visible user [_1] not found', $userID) unless $User;
+		# password requests for elevated users should never make it this far
+		die $c->maketext('insufficient permission to edit [_1]', $userID)
+			unless ($c->{userPermission} >= $User->permission);
 		my $param = "user.${userID}.new_password";
 		if ($c->param($param)) {
 			my $newP          = $c->param($param);
@@ -615,6 +629,7 @@ sub importUsersFromCSV ($c, $fileName, $createNew, $replaceExisting, @replaceLis
 	my $db   = $c->db;
 	my $dir  = $ce->{courseDirs}->{templates};
 	my $user = $c->param('user');
+	my $perm = $db->getPermissionLevel($user)->permission;
 
 	die $c->maketext("illegal character in input: '/'") if $fileName =~ m|/|;
 	die $c->maketext("won't be able to read from file [_1]/[_2]: does it exist? is it readable?", $dir, $fileName)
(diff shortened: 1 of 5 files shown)


drgrice1:
A quick note on the `text-end` css class.  It is not actually needed.  The `math4.scss` file adds `text-align: right` to the `#login_status` element (which is what `text-end` does).  `math4.scss` is common to all themes.  If you are using a customized theme with a copy of that file instead of a link, it may not have that though.

Also, what exactly do you mean by the class was added to match other themes?  The system.html.ep file is used by all of the provided themes.  The actual file is in the math4 directory, and the other themes just link to that file.

I think that the login status looks better without the `my-auto` css class.  I like the login status being at the top, rather than centered vertically in the header.  To me things look a little off balance when it is vertically centered in the scope of the rest of the header.  However, if the consensus is that it looks better centered, then that is fine.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
```diff
@@ -389,7 +391,13 @@ sub delete_handler ($c) {
 	my $num   = 0;
 	foreach my $userID (@userIDsToDelete) {
 		if ($user eq $userID) {    # don't delete yourself!!
-			$error = $c->maketext('You cannot delete yourself!');
+			$error .= $c->maketext(' You cannot delete yourself!');
```
pstaabp, line 394: Did you mean to add a space at the beginning of this string?

pstaabp: I see.  You're concatenating $error, so the answer is yes.

drdrew42: Yes, the space was removed from the default message, and space added at the start of these additional messages to support the fact that there may be multiple additional messages, each requiring an initial space.

drdrew42:
> A quick note on the `text-end` css class. It is not actually needed. The `math4.scss` file adds `text-align: right` to the `#login_status` element (which is what `text-end` does). `math4.scss` is common to all themes. If you are using a customized theme with a copy of that file instead of a link, it may not have that though.

I was using the default math4 theme on develop, and my login-status was aligned left before this change. I'm guessing that I probably missed refreshing with npm install... (confirmed)

> Also, what exactly do you mean by the class was added to match other themes? The system.html.ep file is used by all of the provided themes. The actual file is in the math4 directory, and the other themes just link to that file.

I was referring to math4-green/yellow/red, which each contain a system.html.ep file with `text-end` included.

> I think that the login status looks better without the `my-auto` css class. I like the login status being at the top, rather than centered vertically in the header. To me things look a little off balance when it is vertically centered in the scope of the rest of the header. However, if the consensus is that it looks better centered, then that is fine.

The vertical spacing is now resolved after updating my css with npm install. Login is slightly higher than center, and I'm okay with that -- my layout was unpleasant simply for lack of running the correct version of the css files.

drdrew42:
Strange, I see now that all the files are linked as you say. I'm not sure what confusion I'm running on here... I can revert this change, it's unrelated anyhow.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
drdrew42: Good recommendation, though it will create a new string regardless, since the space is now removed from the default message...

drgrice1 (commented):
This generally looks good.  The inefficiency of the `delete_handler` needs a little attention though.

Also, there is another way that an admin user can be modified in the class list editor that this does not prevent.  That is by importing and replacing a user.  If you export users to a file, then edit an admin user in the file (change the password or permission level in particular), and then import the file replacing users, then the admin user is modified.  I was able to demote the admin user to a student, and change the password to something invalid.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
drgrice1: The translation strings in the `po` and `pot` files don't take into account a space that is outside of a `maketext` call.  So it won't create a new string in those files.  Is that what you are referring to by a new string?

However, my point was not about creating a new string.  The point is that a translator may not know why a space is at the beginning of the string, and then not include it in the translation.  Then for that language the strings will not have the separator space.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
```diff
@@ -389,7 +391,13 @@ sub delete_handler ($c) {
...
+			next;
+		}
+
+		my ($thisUser) = List::Util::first { $userID eq $_->{user_id} } @{ $c->{visibleUsers} };
```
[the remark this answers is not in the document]
drdrew42: Just a heads-up that List::Util also appears in Authen.pm, and you're right -- a hash map is a much better solution.

templates/ContentGenerator/Instructor/UserList/user_list.html.ep
```diff
@@ -71,6 +71,7 @@
 		<tbody class="table-group-divider">
 			% my %selectedUserIDs = map { $_ => 1 } @{ $c->{selectedUserIDs} };
 			% for (@{ $c->{visibleUsers} }) {
+				% next if ( $c->{editMode} && $_->{permission} > $c->{userPermission});
```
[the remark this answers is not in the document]
drdrew42: accidental, I don't like them either.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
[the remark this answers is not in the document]
drgrice1: I see that it is.  So it is technically not a new dependency.  I also see that List::Util is a core Perl module.  So it is probably okay to use it without adding it to check_modules.pl.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
drgrice1: I didn't read all of @pstaabp's comments.  I see where the new string stuff came from now.

drdrew42: Your message hadn't appeared when I posted my response -- it was directed at Peter's second comment above. The removal of the trailing whitespace from the default message would still change the text to be translated. Your suggestion of concatenating with a join would preserve both, yes.

drgrice1: Heh!

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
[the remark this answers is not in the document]
drdrew42: Still, the hash map is a much more efficient method of handling this computation.

drdrew42:
Okay, the delete method has been made more efficient, the translation strings should be preserved, and I tidied up the errant space in the conditional. 👍

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
```diff
@@ -375,21 +376,28 @@ sub delete_handler ($c) {
...
+	my %permissionbyID  = map { $_->{user_id} => $_->{permission} } @{ $c->{visibleUsers} };
 
-	my $error = '';
-	my $num   = 0;
+	my @resultText = ();
+	my $num        = 0;
 	foreach my $userID (@userIDsToDelete) {
 		if ($user eq $userID) {    # don't delete yourself!!
-			$error = $c->maketext('You cannot delete yourself!');
+			push $c->maketext('You cannot delete yourself!'), @resultText;
```
pstaabp, line 395: I'm running perl 5.36.0 and am getting an error here
```
Experimental push on scalar is now forbidden at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm line 395, near "@resultText;"
```
I think the order of the push is wrong.   Similar on line 400 and the unshift on 414.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
```diff
@@ -375,21 +376,28 @@ sub delete_handler ($c) {
...
+			next;
+		}
+
+		if ($permissionbyID->{$userID} > $perm) {
```
pstaabp, line 399: Also have an error here.  I think this should be `$permissionbyID{$userID}`

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
```diff
@@ -375,21 +376,28 @@ sub delete_handler ($c) {
...
-	my %visibleUserIDs  = map { $_ => 1 } @{ $c->{visibleUserIDs} };
-	my %selectedUserIDs = map { $_ => 1 } @{ $c->{selectedUserIDs} };
+	my %allUserIDs      = map { $_            => 1 } @{ $c->{allUserIDs} };
+	my %visibleUserIDs  = map { $_            => 1 } @{ $c->{visibleUserIDs} };
+	my %selectedUserIDs = map { $_            => 1 } @{ $c->{selectedUserIDs} };
+	my %permissionbyID  = map { $_->{user_id} => $_->{permission} } @{ $c->{visibleUsers} };
 
-	my $error = '';
-	my $num   = 0;
+	my @resultText = ();
```
drgrice1, line 391: Never write `my @array = ();`.  Perl initializes an array variable to the empty array on creation by default.  So `my @array;` does exactly the same thing.  The same is true of `my %hash = ();`.  That is also unnecessary.  `my %hash;` is equivalent.

Note that this is different from something like `@array = ();` after initialization.  In that case if the array did have elements, those would be removed and the array reset to the empty array.

drdrew42:
This should be all cleaned up now. Users may not use the import action to create or modify accounts with permissions higher than their own.

drdrew42:
Minor issue: UserList.html.ep displays the wrong user count. I may need to rethink the approach to skipping privileged rows or else fixing this is going to be super hack-y.

pstaabp:
On initial look, I don't see an issue with the user count. 

It seems like you can't select an admin in any table in the classlist editor.  Do we want to allow for export, filter or sort.  That doesn't seem like a security problem, although this is quite minor.

drdrew42:
Refactor complete 👍 . Everything should be re-tested, as the refactor changes touched pretty much everything on the Classlist Editor page... I got everything to work as expected, but as always, there are always potential edge cases.

## Changes

Most of the benefit here is in removing duplicated requests to the database. All (non-set proctor) users are fetched and maintained in `allUsers`, they are added or removed from this hash during import/export/delete. Accurate hashes of visible, selected, and editable users are similarly maintained. Using grep, the sorted list of users can be reduced to only those that are visible/selected/editable as needed.

### Data structures
- editable users are stored in a hash keyed by userID (userIsEditable)
- visible users are stored in a hash keyed by userID (visibleUsers)
- previously visible users are stored in a hash keyed by userID (prevVisibleUserIDs)
- all user IDs are stored in an array (allUserIDs)
- sorted user IDs are stored in an array (sortedUserIDs)
- allUsers is a hash of db records keyed by userID (allUsers)

### Cruft

There is no longer the option to filter directly on section or recitation, since we have regex matching on all of the fields. Data structures originally supporting those actions have been removed.

Also `visible_user_string` seems to no longer be in use either, so all references have been removed.

### Parameter Access

UserList requests have several repeated keys in the form data, and these parameters have been migrated to use the Mojolicious `every_param` instead of relying on the 'old' method of WeBWorK request data parsing.

### Bonus

When sorting by a header row, an input field is briefly displayed in the page before the request is completed. I have added the hidden property to the input field so that this no longer occurs. (Requires npm ci to rebuild the javascript)

drdrew42:
Found and fixed a couple more oversights. I think it's possible to ditch the use of allUserIDs... is it worth it?

drgrice1:
@drdrew42:  I put in a pull request to this branch with one suggested change.  Instead of retrieving the permissions from the database one at a time, get them all in one query.  This gives a considerable speed boost in page loading time for large classes.

drgrice1:
Good catch on the visible input on a click on the header column.

drdrew42:
So it should be worth it to sort only the visible users? I can put in that change along with the improved permissions fetch...

drgrice1:
That would certainly be worth it.  Sorting is always a computationally intensive process.  So if it could be avoided if not needed, then that would be worth it.

drgrice1:
Yeah, now timing is back down to what it was when no users are shown for a large class!  Good work.  Only the visible users really need to be sorted anyway.

drdrew42:
Yeah, there's no sense in refactoring just to make things _worse_!

drgrice1:
I removed the usage of the visible_user_string when I switched a click on a header into a post request.  I missed those in pre_header_initialize though.  Another good catch.

pstaabp:
I'm getting the following error on changing a password of a user (I was logged in as a professor, the user is a student):
```
Not a HASH reference at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm line 245.
```

drgrice1:
> I'm getting the following error on changing a password of a user (I was logged in as a professor, the user is a student):
> 
> ```
> Not a HASH reference at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm line 245.
> ```

The problem is line 565.  @drdrew42 forgot to convert the array reference returned by `every_param` into a hash in this case.  He did it everywhere else though!

templates/ContentGenerator/Instructor/UserList/user_list.html.ep
```diff
@@ -69,20 +69,19 @@
 			</tr>
 		</thead>
 		<tbody class="table-group-divider">
-			% my %selectedUserIDs = map { $_ => 1 } @{ $c->{selectedUserIDs} };
-			% for (@{ $c->{visibleUsers} }) {
+			% for (@{ $c->{sortedUserIDs} }) {
 				<%= include 'ContentGenerator/Instructor/UserList/user_row',
-					user => $_, userSelected => exists $selectedUserIDs{ $_->user_id } =%>
+					user => $c->{allUsers}{$_}, userSelected => exists $c->{selectedUserIDs}{$_}, editable => exists $c->{userIsEditable}{$_} =%>
```
drgrice1, line 74: This line is now longer than the 120 character limit.  Could you change it to something like:
```suggestion
					user         => $c->{allUsers}{$_},
					userSelected => exists $c->{selectedUserIDs}{$_},
					editable     => exists $c->{userIsEditable}{$_} =%>
```
I have been trying for a sort of mix of perltidy and html formatting here.  I am not exactly sure what is best.  The 120 character limit is the important thing.

lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
```diff
@@ -615,6 +612,7 @@ sub importUsersFromCSV ($c, $fileName, $createNew, $replaceExisting, @replaceLis
 	my $db   = $c->db;
 	my $dir  = $ce->{courseDirs}->{templates};
 	my $user = $c->param('user');
+	my $perm = $db->getPermissionLevel($user)->permission;
```
drgrice1, line 615: I don't think that this database query is needed.  You should be able to use the value of `$c->{userPermission}` cached in `pre_header_initialize`.

drgrice1 (commented):
Just a few suggestions.

I think after these are dealt with and the mapping added on line 565 of UserList.pm that is giving the error @pstaabp found, this pull request is looking good.

drdrew42:
Okay, updates are in!

drgrice1 (approved):
Looks good.  Thanks for the contribution.

pstaabp (approved):
Looks good now.

PR #1889, opened 2023-02-10
The ContentGenerator helpMacro needs a default value for the args parameter.

drgrice1:
This causes an error when clicking on "Advanced Search" in the library browser (and perhaps other places).

This fix is in #1884, but is unrelated to that pull request.  So I will add this separately.


pstaabp (approved):
This fixes the issue.

PR #1891, opened 2023-02-14
Make the PGProblemEditor show the problem submit buttons for source paths

drgrice1:
Currently if you edit an OPL or Contrib problem from the library browser the submit buttons are not shown in the problem rendered in the page.

To fix this the javascript needs to check for both "problem" and "source_path_for_problem_file" to determine when to show the submit buttons. This actually just checks that the file type contains the string "problem" (so it also catches blank_problem, although I don't actually know how to get the PGProblemEditor to open a true blank problem).


pstaabp (approved):
Fixes the problem

PR #1893, opened 2023-02-15
Fix the stats and progress index page siblings

drgrice1:
There was another issue with the incorrect route being used.  If viewing the index page of either Statistics or Progress then the siblings links for all sets listed just take you back to the index page instead of to the statistics or progress page for that set.


drgrice1:
Nevermind.  This is fixed in #1884.

PR #1894, opened 2023-02-15
Improvements to the layout

drgrice1:
This makes the site navigation element a separate scrollable entity from the rest of the page content.  When a page is open that has a lot of siblings the separate scrolling of the content and the site navigation makes it more convenient to get to site navigation elements from anywhere in the page.
    
The header (which we call the masthead) is a fixed element, and so when the content is scrolled it remains visible at the top of the page, while the site navigation remains fixed at the left.
    
The sidebar toggle button is now in the header with the WeBWorK logo.  So that is always visible on the page even when scrolled to the bottom of a long page.  The icon is changed to a standard menu bars icon instead of the directional chevron.
    
When the screen width is at or below the medium breakpoint only the WeBWorK logo (and sidebar toggle button) is fixed in the masthead.  The rest of the masthead scrolls with the rest of the page.
    
The white space around the entire page has been removed.  That is now on the page content.  Literally, the `container-fluid` class was moved from the entire page to the content div.  In fact the outer div that wrapped the entire page (except the footer) before has been removed.
    
There are some other improvements to the site navigation.  Instead of changing the width of the site navigation to 0 to make it hidden, the site navigation's left position is made negative.  So it is moved off the left of the page.  This gives a much smoother appearance as it transitions off the page, since the text does not appear to be being squished and start wrapping.  Once the transition is completed the bootstrap `invisible` class is added to the site navigation which removes the site navigation from the tab order.  Currently the site navigation remains in the tab order even after it is hidden.  So if focus is on an element in the masthead, and you tab out of it you end up tabbing through all of those now hidden links in the site navigation.

The MAA logo has been updated to an SVG image.  This SVG was downloaded from https://upload.wikimedia.org/wikipedia/en/8/8c/Mathematical_Association_of_America_logo.svg.  The fill color was changed to white and the stroke color added and set to the original bluish fill color.  This was done to make this look good with all of the current webwork2 themes.  This is essentially the same logo that is currently shown on the MAA website, and is a more compact logo.


drdrew42 (approved):
This is terrific -- looks great!

pstaabp (approved):

PR #1896, opened 2023-02-15
PGProblemEditor iframe resize for narrow layout

drgrice1:
When the window width is less than the large breakpoint, resize the iframe to fit its content when it loads.  Note that the iframe can already be resized in a large layout by resizing the CodeMirror editor.  The render iframe is resized to match that.

Also scroll to the top of the render area when a problem loads if the current scroll position is below that.


pstaabp (approved):
Looks good.

PR #1899, opened 2023-02-24
Remove the null byte hack.

drgrice1:
In fact the WeBWorK::Form module is no longer needed at all.  Just use the Mojolicious request params hash directly.

I thought that I was going to need to say that this is paired with https://github.com/openwebwork/pg/pull/778 which removes null byte usage from PG.  However, it turns out this works without that.  This null byte handling just isn't needed.

One thing that I am not certain of is how to encode past answers that come from arrays (for checkbox answers).  At this point this separates entries by a space.  Previously a null byte was used.  The null byte does not show up when viewing past answers.  Is there something better than a space to use?  Note that separate answers for a problem are still separated by a tab character.


taniwallach:
If I understand what you wrote about https://github.com/openwebwork/pg/pull/778 the choice "value" which gets stored in the past answer table is something we will be able to set to text which can include spaces. Thus, it does not seem that using a space character is a good choice here. The null character was not a better choice, but whoever put it into used probably chose it as something could not appear in the answers.

Maybe we can use a Unicode symbol which we can reasonably assume will not be used in the "value" strings and which looks like a "delimiter". There is a long list of "Other symbols" in Unicode at https://www.compart.com/en/unicode/category/So

Some options I noticed:
  * https://www.compart.com/en/unicode/U+236E (a underscored semicolon)
  * https://www.compart.com/en/unicode/U+241E (record separator character)
 
Nothing with "comma" on that page looked relevant.

drgrice1:
We can use whatever we want for this separator.  Remember the only thing that the past_answers are used for is for display in the answer log (past answers table).  Previously the null byte character essentially provide no separation at all.  So the space is an improvement on that.  Also prior to openwebwork/pg#778 these were values were single character values (if coming from PGchoicemacros.pl) or things like "B0", "B1", ...  With that PG pull request the values can be whatever the author sets them to be and could include spaces.  However, they could really include anything.  So whatever is chosen for the separator could still be used in a value.  Even any Unicode symbol that we choose.

drgrice1:
I think it is probably reasonable to assume that either of the Unicode symbols you suggest would not be used. I vote for the understood semicolon. I think the record separator wouldn't look good in the past answers table.

taniwallach:
@drgrice1 I saw you posted while I was typing, and I left what I had written at the bottom. 

I also think the "underscored semicolon" seems visually nicer than "record separator". Go ahead an use it. If someone has a better suggestion later, the choice can be changed.

In potential anyone can parse the data from the database or the answer logs file, and when done by code - the null can be "seen" by the code, as will whatever delimiter we finalize on. As such, I think it would be good to make a decision and stick to it.

---

You asked 
>  Is there something better than a space to use?

Since space characters seem to be something somewhat likely to be potentially included in a "value" - it seems helpful to avoid using that as the delimiter, as the purpose was to give a visible delimiter which can be recognized as such. For the same reason, simple commas and semicolons are probably not a safe choice. With full UTF-8 support in the code, we cannot totally avoid what authors can do, but we can try to pick something which is as **unlikely** to be used frequently. That is why I wrote _reasonably assume_ above, and I think picking something reasonable noticeable and which looks like a separator/delimiter of some sort it is the best we can do.

drdrew42 (approved):
Tested with openwebwork/pg#778 and openwebwork/pg#779

👍

drdrew42:
I don't have a strong preference for delimiter. As has already been said, we can parse it out so long as it is not an expected character in any response label.

drgrice1:
@taniwallach: I switched to using `&#9070;` to separate the array answers in the past answers table.  So in the database you will see `answer 1&#9070;answer 2`, but in html in the past answers table you will see "answer 1&#9070;answer 2".

pstaabp:
I see the `answer 1&#9070;answer 2` in the database.  On the past answer page, it seems to have both ; and &#9070;

Here's an example:

image

pstaabp:
I was testing the openwebwork/pg#778

drgrice1:
That is because the value in the problem has the semicolon. You are using one of the examples I gave that I created before the separator was added to the past answers table.

pstaabp (approved):
In my test above, I must have submitted before checking out this branch.  Looks good now.

Issue #1902, opened 2023-02-28
Close date takes user value in set overrides


pstaabp:
In the user overrides for a set on develop, the "Close Date" column is used epoch 0 for the close date which is probably coming from the user_set table in the `due_date` column.  In this example:

<img width="674" alt="image" src="https://user-images.githubusercontent.com/879338/221892980-b08a016e-f754-40f4-aee6-f6ea4cdf8550.png">

The two non 1969 close dates have overrides in the database, and all of the other values are null.

drgrice1:
This is just a display issue on this page.  Those are not the dates used, just what is displayed here.  The page is not supposed to show close dates unless there is an override.  What is probably happening is that there an incorrect check for when the user set close date is set or not, and it is being displayed in any case.  The NULL database value gets interpreted to zero, and results in the beginning of the epoch.

Should be easy to fix.

drgrice1:
This has been fixed.

PR #1903, opened 2023-02-28
Don't display due dates on the users assigned to set page if not set.

drgrice1:
This is how this was with webwork 2.17 and how it should still be.  A check was forgotten in the conversion to Mojolicious templates.

This fixes issue #1902.


pstaabp (approved):
fixes the issue. I figured it would be an easy fix.

PR #1904, opened 2023-03-02
Add Observer status

somiaj:
Add a new observer status for instructors/teaching assistants/users who have access to the course and the assignments, but are not included in any stats, grades, or emails. Add a style so Observer users are green in the classlist editor.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -548,6 +548,10 @@ $default_status = "Enrolled";
 		abbrevs => [qw/ A a audit /],
 		behaviors => [qw/ allow_course_access include_in_assignment include_in_stats include_in_email /],
 	},
+	Observer => {
+		abbrevs => [qw/ O o observer /],
+		behaviors => [qw/ allow_course_access include_in_assignment /],
+	},
 	Drop => {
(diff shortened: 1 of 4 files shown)


pstaabp:
A few things here:
1. I'm not sure about the green on the Observer status.  Perhaps darken it up.  It doesn't seem to have the same contrast.  Even though it's not crucial to having the Observer status different visually, there's always the issue of accessibility and color blindness.
2. You mention that the observers are not in stats/grades/email, however, in on the Student Progress page, observers are listed on the right side.  Is this the intended behavior?

Overall, the idea is good and doesn't interfere with past courses.

drgrice1:
> I do like how audit is purple and dropped is light grey, just helps them pop out when looking at the course list. So I picked green as it wasn't a color that was being used. I can change this to a darker shade of green, increase the font-weight, or use another color whose contrast will help stand out better. I do think it can be helpful to be able to quickly see the different status in a table list vs looking directly at the status column.

I think the green is fine.  I think it stands out enough, and not too much.  I don't want the font-weight to be increased.  I don't think that will look good.

pstaabp (approved):

PR #1906, opened 2023-03-12
Reset the problem grader inputs on page load.

### LINKED ISSUE #1905: SingleProblemGrader uses previous problem % score in tests when moving through problems.
I noticed in develop when using the single problem grader in a test with multiple problems, each problem on its own page, that when I move between problems, the 'Problem Score (%)' field is not updated to the current problems score, instead the field keeps the same value as the previous problem. Here are the steps I used:

+ Open up a test version with multiple problems.
+ Open up the problem grader.
+ Change the 'Problem Score (%)' to a different number.
+ Link to a new problem, the 'Problem Score (%)' will keep the same value as set in the previous step.

Unsure if this is some issue with my setup or not, as it seems weird it is caching the value from a previous page, even when I don't save the score. I have run 'npm ci', './generate-assets.js', and even done a hard reload, and still seem to have an issue with this.

Testing further, I even have this issue with the problem part sub-scores, and here is a bit stranger, it sets all the sub-parts equal to the same thing (what I entered into the final sub-part score). I also don't have this issue when moving through problems with the problem grader in normal homework sets.

drgrice1:
This fixes #1905.  See the issue for a description of the problem.  Although not mentioned in the issue, the same problem occurs with the comment field.

### DIFF
--- a/templates/HTML/SingleProblemGrader/grader.html.ep
+++ b/templates/HTML/SingleProblemGrader/grader.html.ep
@@ -40,7 +40,8 @@
 						<% end =%>
 					<% end =%>
 					<div class="col-sm">
-						<%= number_field 'answer-part-score' => $scores[$part],
+						% param('answer-part-score', $scores[$part]);
+						<%= number_field 'answer-part-score' => '',
 							class => 'answer-part-score form-control form-control-sm d-inline',
 							id    =>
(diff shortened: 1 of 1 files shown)


somiaj (approved):
I can confirm this fixes the issue I had in #1905

drgrice1:
By the way, the reason this did not occur for homework sets is because changing problems in a homework set is not a form submission.

PR #1907, opened 2023-03-12
Add features to the single problem grader.

somiaj:
Adds the following to the single problem grader:

+ Include an input field for the point value of a problem that can be used to set a problem score. This field modifies the percent score appropriately, and is updated to stay in sync with it if the percent score or sub scores are modified.
+ Add a percent score along side the recorded score in tests.
+ Update the recorded score and percent score in tests when the single problem grader is used to match any changes made.

This is built on top of the bugfix in #1906.


drgrice1 (commented):
I have to say that I find this a rather bad idea.  There is no reason to have two settings for the same thing, one in decimal 0 to 1 form, and one in percent 0 to 100 form.  That just doesn't really make sense.

somiaj:
Not all problems are worth 1 point. I am currently using this, because I find it easier to set the point value then have to compute the percent each time. We could hide this additional input on problems that only have a single point.

somiaj:
Opps, I need to fix the javascript to deal with things when the field is hidden on that last commit.

drgrice1:
All problems are given a status of 0 to 1.  You are talking about the problem value. What do you mean by computing the percent?  Even if the value for the problem is 5, 100% means 5 points.

Even if hidden for 1 point problems, I still think this is a bad idea.  This just makes no real sense.

drgrice1:
Also, you should note that this pull request is going to have rather big conflicts with #1884.

somiaj:
If a problem has a value of 13 points, this field allows you to set the grade to be between 0 and 13 points. Without this if someone wanted to give 9 points on this problem, they have to first compute the percent score, and then set that. I am currently using this because I grade based on the points each problem is worth. I thought others might also find this useful if they grade based on the problem value, and let the computer compute the status to be saved.

I currently find this useful and so do some of the instructors I work with. If overall the feeling is this feature won't be useful, I can just close this pull request and keep a local patch for myself.

drgrice1:
Why do you really need to determine the exact percentage to give 9 points?  Even if you set your new input to 9, the score that is saved to the database is 0 to 1, and that will not be exactly 9.

Pedagogically, you shouldn't be thinking in terms of the value relative to the rest of the set.  It is bad practice.  You should think in terms of the problem only, and score the student for the merit of that problem alone.

somiaj:
I do understand that due to saving the score as a value between 0 and 1 and rounding the score saved in the data base can make point values sightly off, but this doesn't change ones personal preference to grade a problem using points vs a percent.

This isn't a pedagogical issue, grading a problem in terms of the problem only can be done by either giving the problem a % grade, or based off a point value. It is personal preference which one to use, but they are just different methods of awarding partial credit and splitting the value of a problem across problem parts.

If the majority agrees that this feature isn't needed in webwork2, then I will drop the first commit.

The second commit might be useful, this just updates the recorded test grade on the page without having to reload it when the grade is changed via the single problem grader.

drgrice1:
I don't understand why you think you need to grade with precise points out of the total value.  That is an antiquated way of thinking, and leads to a likely misjudgement of student performance.

Also, the additional input is just unnecessary clutter.

htdocs/js/apps/ProblemGrader/problemgrader.js
```diff
@@ -35,25 +35,57 @@
...
+			} else {
+				part.classList.remove('is-invalid');
+
+				let score = 0;
+				answerLabels.forEach((label) => {
+					const partElt = document.getElementById(`score_problem${problemId}_${label}`);
+					score += partElt.value * partElt.dataset.weight;
+				});
+				document.getElementById(`score_problem${problemId}`).value = Math.round(score);
+				document.getElementById(`grader_messages_problem${problemId}`).innerHTML = '';
```
drgrice1, line 49: Line 49 should be moved after the `else` statement.  The messages should be cleared even if the part score does not validate.

templates/HTML/SingleProblemGrader/grader.html.ep
```diff
@@ -60,6 +61,44 @@
...
+						<i class="icon fas fa-question-circle" aria-hidden="true" data-alt="Help Icon"></i>
+						<span class="sr-only-glyphicon">Help Icon</span>
+					<% end =%>
+				<% end =%>
+				<div class="col-sm">
+					% param('grader-problem-points',  wwRound(1, $grader->{recorded_score} * $grader->{problem_value}));
+					<%= number_field 'grader-problem-points' => '',
+						min          => 0,
+						max          => $grader->{problem_value},
+						step         => 0.1,
```
drgrice1, line 93: If this point input is going to persist, the `step` value needs more thought.  A step size of 0.1 is only good for problem values in a very limited range.  Well, basically it is only sufficient is the problem value is basically 10 (perhaps good enough for values close to 10).  For a problem values less than that it is to restrictive.  For problem values more than that it is to granulated.  So adjust this based on the problem value.

drgrice1 (commented):
I still don't like the new point value input (and won't).  However, the rest of this pull request is beneficial.

There are a couple of small issues to take care of.

Alex-Jordan:
Is the number of points equal to the weight of the exercise? If not, having three things (status, weight, and point total) does seem like a source of confusion for some instructors.

Tossing out an idea. There could be a global setting where:

* Default is status quo.
* Alternatively, indicate a preference for points. If point based, the final score input field would not display (%) and would say (__/13) or whatever. That "13" would be the weight of the exercise in the assignment, not something you set independently. Subscores would still be percents and assume whatever distribution is currently assumed. Status would still be recorded as it is now, and something is translating between the status values and the "(__/13)" values.

I know that is not equivalent. You could not have equal weights among exercises in a set and still have 13 points within one exercise but 2 points within another. But if an instructor likes assessing an exercise based on points, is it not likely they would want those to match the weights?

somiaj:
> Is the number of points equal to the weight of the exercise? If not, having three things (status, weight, and point total) does seem like a source of confusion for some instructors.

Yes, the number of points is the problem 'weight'. Inside the problem webwork calls them points, such as something like this, `Problem 3. (12 points)`.

I was also thinking a configuration option that could turn the option on/off. Though if most webwork users don't use points, I do agree that it doesn't need to be merged, I just thought that I wasn't the only one who used the manual grader using a point scale rubric.

drgrice1:
@Alex-Jordan:  Are you suggesting a course configuration option to show either the problem weight input or the usual percent input (or possibly both)?

@somiaj:  If you implement this in that way, I can approve it.  I don't think there is any other way that I can get behind this.

Alex-Jordan:
If it's an config option, I think it's good. I don't apply weights other than 1 to exercises, so I wouldn't use it. But if people do apply weights other than 1 to some exercises (or import someone else's set with that setting), it could be a nice reminder to them that they have such weights in place.

So yes, I'm saying something like a variable that starts in `defaults.config`, maybe `$preferred_scoring = 'percent'`. Can be overrided in all the usual places, including in `simple.conf` from the Course Config page. It would be a course-wide discussion to handle all exercise sets one way or the other.

And `$preferred_scoring = 'percent'` means status quo.

But something like `$preferred_scoring = 'points_out_of_weight'` means:
* leave subscoring as it is now
* replace `Problem Score (%):` with `Problem Score (out of 13):` or whatever "13" is. The field expects decimal numbers in the interval [0,13] (or allow to go higher?)
* whatever is entered/displayed in that field is appropriately translated to/from the 0-to-1 status. (There should be nothing that risks a snafu if an instructor toggles this option.)

If it's better, make the option boolean. Then it will be a countdown until someone asks for a third way of doing things :)

drgrice1:
I do apply weights that are not 1 to my assignments, but I still don't want the points input.  It will just get in the way.

drgrice1:
I think the option should not be boolean.  In my opinion with this pull requests there are already three options.  Show the points input only, show the percent input only, and show both.  The default should be to show the percent input only (the status quo).

@Alex-Jordan:  Your third bullet is how things work with this pull request for the points input already.

Alex-Jordan:
Sure, my three bullets are outlining how I would do this compared with the current tip of develop. Not compared with the tip of this PR branch.

> Show the points input only, show the percent input only, and show both. 

Personally, I would leave out the "show both" as an option.  @somiaj as a user, do you want to see both, or did you just want to see the points option and knew you would need to leave the percent option there too for others?

somiaj:
@Alex-Jordan for my use case, having the percent there doesn't add any additional information I use. I just left it there since that is the field that is actually saved to the database, and treated my points field like the sub-score fields for multi-part problems to allow an additional way to compute the percentage that is saved.

htdocs/js/apps/ProblemGrader/problemgrader.js
somiaj: Thanks, overlooked this being set. I also added this to clear the message when changing the point value.

conf/defaults.config
```diff
@@ -1983,6 +1994,17 @@ $ConfigValues = [
 			),
 			type => 'number'
 		},
+		{
+			var  => 'problemGraderScore',
+			doc  => x('Method to enter problem scores in the signle problem manual grader'),
```
pstaabp, line 1999: typo: 'signle' to 'single'

somiaj: Thanks for catching that.

templates/HTML/SingleProblemGrader/grader.html.ep
pstaabp: I agree with this.  Playing with this in the "Both" mode on a problem with weight 1, if I select a percent like 75%, it becomes 0.8, which doesn't make sense.  You may just want to scale the points to ensure that the percentage is equivalent.

drgrice1:
@somiaj:  I put in a pull request to this branch with some suggestions and a minor css fix.  Please look at it.

drgrice1 (approved):
I think this looks good now.

drgrice1:
I will merge this now that you have fixed the conflict.

PR #1908, opened 2023-03-13
Pass the options isInstructor and forceScaffoldsOpen through with the render_rpc endpoint.

drgrice1:
Since these options are not passed on through the form in the default template, they are not preserved after submitting answers.  As such problems with scaffolds come back after submitting with the latter scaffolds grayed out and unable to be opened when a problem is rendered on the set detail page, the problem sets editor page, or on the library browser page.


pstaabp (approved):
I didn't see it grayed out on develop in all of the places you mentioned, but I did see in the Problem Set Detail and this fixes it.

PR #1909, opened 2023-03-13
Implement LTI 1.3 authentication and LTI Advantage tools.

drgrice1:
Everything is set up to work much like the old LTI 1.1 authentication and the previous grade passback.  There is much more that can be done with the LTI Advantage tools and that could yet be implemented.
    
There is some hackery so that the courseID can be reliably retrieved early in the LTI advantage launch request before the JWT is decrypted.
    
There is a new column added to the set_user (UserSet) table. Other than that the existing lis_source_did columns from other tables were recycled for this.  The new column is name lis_source_did the same as the set_user column so that the merged set method still works without adding a new column to that table.  It is the LMS lineitem that is stored there, and that is common to all users of the set.  Note that there are several new things stored in the course's "setting" table for this, but that doesn't take a change to the database.
    
The authen_LTI configuration files are reworked to relegate the two versions being used at the same time.

LTI 1.3 Setup Instructions:
The attached zip contains screenshots and an authen_LTI_1_3.conf file that demonstrate the settings needed to make LTI 1.3 authentication and grade passback work for Moodle and Canvas.  These were taken from my local test installations that are all on my local machine.  Hence the `localhost:3005` address in the screenshots.  Replace `http://localhost:3005` with an actual domain name.  Usually you will not include the port.  Of course change `http` to `https` if you have ssl on your webwork server.  At some point actual instructions are needed.  Hopefully this gets someone started with testing.
[setup.zip](https://github.com/openwebwork/webwork2/files/11425838/setup.zip)


conf/localOverrides.conf.dist
```diff
@@ -421,34 +421,11 @@ $mail{feedbackRecipients}    = [
...
-# user/set  update of the grade.
-
-# The switches above can be set in course.conf to enable debugging for just one course.
-
-# If you want even more information enable the debug facility for SubmitGrade.pm in
-# the WeBWorK::Constants file.
-# This will print extensive debug messages for all courses.
-
-
+include("conf/authen_LTI.conf");
```
pstaabp, line 428: Do we want this commented out by default?

pstaabp:
I'm trying the 1.3 with the LMS mojolicious app.  Will this work?   I don't see the LTI_debug parameter anymore.  Does this still exist?

conf/localOverrides.conf.dist
```diff
@@ -421,34 +421,11 @@ $mail{feedbackRecipients}    = [
 # Uncomment the following line to enable authentication via IMS LTI.
```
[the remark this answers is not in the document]
drgrice1: IMS is actually correct. The IMS Global Learning Consortium is the non-profit entity that maintains the LTI specification.

drgrice1:
The LTI debugging paramaters have not changed.  There are two of them `$debug_lti_parameters` and `$debug_lti_grade_passback`.  They are still in `authen_LTI.conf.dist`.

drgrice1:
No.  My Mojolicious LMS app will not work with this.

drgrice1:
I think that the important thing to test here for now is the LTI 1.1 functionality.  Test to see that it still works as before.

Testing LTI 1.3 functionality is much harder without an LMS that implements it.  I installed Moodle and Canvas locally to test it (maybe I will try to write an LMS test app for this at some point).  In addition, documentation on how to set this up is lacking.  That still needs to be done.

Note that there are now three authen_LTI configuration files.  `authen_LTI.conf.dist` contains the configuration variables that are common to both LTI 1.1 and LTI 1.3.  `authen_LTI_1_1.conf.dist` contains configuration variables for LTI 1.1, and `authen_LTI_1_3.conf.dist` contains configuration variables for LTI 1.3.  If only using LTI 1.1 then you only need to copy the first to files to the non dist name.

To use LTI 1.1 make sure to make non dist copies of both `authen_LTI.conf.dis` and `authen_LTI_1_1.conf.dist`, and configure them correctly.  Note that you can use my Mojolicious LMS app to test this.
[lms-test-app.zip](https://github.com/openwebwork/webwork2/files/11100381/lms-test-app.zip)

By default the LTI 1.3 module is commented out in `authen_LTI.conf`.  To enable LTI 1.3 you need to uncomment lines 46 and 56 of that file, and then copy `authen_LTI_1_3.conf.dist` to its non dist name.

If you want only 1.3 and not 1.1, then you also should comment out lines 45 and 55 in `authen_LTI.conf`.  Of course, in that case you don't need a non dist copy of `authen_LTI_1_1.conf.dist`.

drgrice1:
A quick note (that just came up for me in changing branches).  If you have an existing copy of `authen_LTI.conf`, it will not work with this.  You will need to relegate changes in this pull request.  The 1.1 settings are moved to `authen_LTI_1_1.conf` and the base settings in `authen_LTI.conf` are not quite the same.

somiaj:
I am attempting to test this out with canvas here, but will have to wait to hear from the canvas admins if they will help setup an LTI 1.3 key since I can't do that myself.

I decided to test this to see if LTI 1.1 and LTIMassUpdate works, and I ran across an issue with LTI 1.1, when webwork is recreating the URL at line 458 in `lib/WeBWorK/Authen/LTIAdvanced.pm`:  `my $path = $ce->{LTI}{v1p1}{BasicToThisSiteURL} || ($c->url_for->to_abs =~ s|/?$|/|r);`. The url that gets recreated is `http://..` instead of `https://...`. And this causes authentication to fail. If I add `$path =~ s/http/https/` to force https, authentication works.

The reason I think the url is only `http://...` is because I'm running this in a virtual machine on port 8080 and access webwork2 through a secure proxy. I have `$server_root_url` in `site.conf` set to `https://...`, which I think is meant to be used to help set the url and know it needs to be https in this case. Configuring the url with `$ce->{LTI}{v1p1}{BasicToThisSiteURL}` would only work for a single login course mode, since the url needs to match what is in the LMS exactly, which varies between assignments.

Outside of this small issue, LTI 1.1 still works as before configuring it with the new files/option names.

drgrice1:
@somiaj:  Have you tested this with the develop branch?  That line hasn't been changed from what is in develop other than changing to the new LIT v1p1 configuration variable.

drgrice1:
It also sounds like you don't have your configuration set up right.  If you are proxying, then you need to forward the protocol.  If the protocol is forwarded, then it will be correctly picked up in the `$c->url_for->to_abs` call.

somiaj:
@drgrice1 I haven't tested develop, though I expect the same result. It could be I have this missconfigured, but unsure where, outside of listen: set to `http://*:8080` in `webwork2.mojolicious.yml`, I'm setting the url to be `https://` of the outside proxy elsewhere.

Outside of this issue, the proxy setup seems to be working fine.

drgrice1:
If you are proxying via apache2, then it should be set in `webwork2.apache2.4.conf`.  That is on line 30 of the dist file, and `https` is the default.  If you are proxying via nginx, then that is set in `webwork2.ngixn.conf`.  For nginx it automatically detects the protocol, so you don't need to change anything.

If you are proxying in some other way, then you will need to figure out how to forward the protocol.  In any case, it is the X-Forwarded-Proto header that needs to be set.

somiaj:
@drgrice1 Okay, that was it, I missed setting `X-Forwarded-Proto https` on the proxy, since everything else worked with it, so didn't realize it was needed.

somiaj:
Anyways, LTI 1.1 and the minion job queue work fine in all my testing, maybe next week I'll see if I can get a setup to test LTI 1.3.

drgrice1:
Yeah, LTI 1.3 is harder to set up than LTI 1.1 in that an administrator must set it up for you.  With LTI 1.1 you could set things up without administrative access on the LMS.

conf/authen_LTI_1_1.conf.dist
```diff
@@ -0,0 +1,212 @@
...
+# should be used carefully!
+$LTI{v1p1}{fallback_source_of_username} = '';
+#$LTI{v1p1}{fallback_source_of_username} = 'lis_person_sourcedid';
+
+# Stripping the domain when creating user_id from an email address:
+# If you set either preferred_source_of_username or fallback_source_of_username to
+# lis_person_contact_email_primary, and the email is being used as the WeBWorK user_id, then if
+# the following setting is enabled, then WeBWorK will strip off the domain portion of the email
+# (after the '@') and just use the username.
+$LTI{v1p1}{strip_address_from_email} = 0;
```
taniwallach, line 83: Once we are making significant chages, maybe replace `strip_address_from_email` with `strip_domain_from_email`?
Also for LTI_1_3.

conf/authen_LTI_1_1.conf.dist
```diff
@@ -0,0 +1,212 @@
...
+# The purpose of the LTI nonces is to prevent man-in-the-middle attacks.  The NonceLifeTime (in
+# seconds) must be short enough to prevent at least casual man-in-the-middle attacks but long
+# enough to accommodate normal server and networking delays (and perhaps non-synchronization of
+# server time clocks).
+$LTI{v1p1}{NonceLifeTime} = 60;    # in seconds
+
+# This allows you to override the URL that Oauth will use to validate the authentication.  This
+# is important if you have some sort of setup (e.g. load distributing) where the path in the LMS
+# does not match the path that ends up in the webwork page.
+$LTI{v1p1}{BasicToThisSiteURL} = '';
```
taniwallach, line 135: If making major changes - maybe we can use a clearer name than `BasicToThisSiteURL`?

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
```diff
@@ -320,18 +297,15 @@ EOS
...
 						if $ce->{debug_lti_grade_passback};
-					$lti_do_send = 1;    # Change back to sending the update
 				} elsif ($oldScore ne '' && abs($score - $oldScore) < 0.001) {
 					# LMS has essentially the same score, no reason to update it
 					debug("LMS grade will NOT be updated - grade unchanges. Old score: $oldScore; New score: $score")
 						if $ce->{debug_lti_grade_passback};
 					$self->warning('LMS grade will NOT be updated - grade unchanged. '
 							. "Old score: $oldScore; New score: $score")
 						if ($ce->{debug_lti_grade_passback});
-					$lti_do_send = 0;    # Do not send
```
taniwallach, line 331: It is not clear to me how the revised code avoids sending updates when the score is essentially unchanged when the use of `$lti_do_send` is removed. The main purpose of check of the current grade in the LMS was to avoid having an update be done for a negligible change, to reduce load on the LMS and the "log/history" spam for updates which are not really important.
Maybe I am missing something.

lib/WeBWorK/DB/Record/Set.pm
```diff
@@ -59,6 +59,7 @@ BEGIN {
 		hide_hint                 => { type => "INT" },
 		restrict_prob_progression => { type => "INT" },
 		email_instructor          => { type => "INT" },
+		lis_source_did            => { type => "BLOB" },
```
taniwallach, line 62: Should the content be `TEXT` or `BLOB`?

taniwallach (commented):
I read over the code, but did not yet test it, and made a few comments.
I hope to be able to do testing, at least of the revised code for LTI 1.1, next week.

conf/authen_LTI.conf.dist
```diff
@@ -1,376 +1,163 @@
...
-# If you want even more information enable the debug facility for SubmitGrade.pm in
-# the WeBWorK::Constants file.
-# This will print extensive debug messages for all courses.
+# If you want even more information enable the debug facility for SubmitGrade.pm in the
+# WeBWorK::Constants file.  This will print extensive debugging messages for all courses.
 
-#############################################################################
-# Authentication settings
-#############################################################################
+# Note that not for LTI 1.3 not all debug message will make it back to the HTML page due to the
```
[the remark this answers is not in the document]
drgrice1: The word "not" should not be there.  It was removed.

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
drgrice1: You are missing something.  You missed it when you implemented it.  After each time that you set the `$lti_do_send` variable to 0 you `return`.  That returns from the subroutine, and does not continue on to the later place in the subroutine where it checks the value of the variable and the sending is done. The only times there is not a return statement after setting `$lti_do_send` is when you set it to 1.  The variable was never needed.  You didn't really think that one out it seems.

conf/authen_LTI_1_1.conf.dist
drgrice1: I am fine with changing this.  Do you have an idea for a better name?  Perhaps `OverrideSiteURL`?

conf/authen_LTI_1_1.conf.dist
drgrice1: Sure.  Why not.

lib/WeBWorK/DB/Record/Set.pm
drgrice1: I set it to `BLOB` to match the other `lis_source_did` columns in the database.  If those are changed to `TEXT`, then this should be as well.

conf/authen_LTI_1_1.conf.dist
drgrice1: This is changed now.

lib/WeBWorK/DB/Record/Set.pm
drgrice1: This is now type TEXT.

Alex-Jordan:
My D2L people cannot figure out what the `PlatformID` config value is supposed to be. I see it associated with `iss` in the code, and then I see that `iss` is the issuer URL in the LTI 1.3 spec. Does that sound right? And if so, should `PlatformID` change to `PlatformURL` or `IssuerURL`?

drgrice1:
This should either be `PlatformID` or `Issuer` (or perhaps `IssuerID`).  It is referred two both ways by different implementations.  It is not a URL in general, although in some cases it is.  For Moodle it is a URL, but for Canvas it is not.  It really just depends on what the LMS platform sets this to in the JWT.  Moodle calls it the `Platform ID` in the information it provides when you set up the tool.  Other platforms don't give that nice summary in the way that Moodle does which is kind of a pain.

PR #1910, opened 2023-03-14
Fix an error when opening a set that is not assigned to any users.

drgrice1:
The user record needs a default value for the fieldTable method of ProblemSetDetail.pm.  Otherwise an error occurs when opening a set that is not assigned to any users, because the method receives only 5 arguments, but expects at least 6 arguments.

Also the 6th optional setType argument needs to be checked if it is defined each time it is used in the method.


pstaabp (approved):
Fixes the error.

PR #1911, opened 2023-03-17
Updates to grades page.

somiaj:
Make test names link to the ProblemSet test page. If it is
  a test version, (version n) links to the actual test version,
  provided the student can see the version's problems. Also
  consistently use table headers for the first column.

  For sets that are not yet opened, don't link to the set and
  don't show the grades for the set.

  Don't show hidden sets on the grades page.

  Refer to tests as tests (instead of set or assignment) in messages
  about no version has been taken and cannot display scores.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Grades.pm
+++ b/lib/WeBWorK/ContentGenerator/Grades.pm
@@ -155,6 +155,9 @@ sub displayStudentStats ($c, $studentID) {
 	my %setVersionsCount;
 	my @allSetIDs;
 	for my $set (@sets) {
+		# Don't show hidden sets unless user has appropriate permissions.
+		next unless ($set->visible || $authz->hasPermissions($c->param('user'), 'view_hidden_sets'));
+
 		my $setID = $set->set_id();
 
 		# FIXME: Here, as in many other locations, we assume that there is a one-to-one matching between versioned sets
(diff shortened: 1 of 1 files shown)


somiaj:
It should be noted that this also changes what is shown on the 'Student Progress' page for individual students, since that is identical to what you would see on the grades page for that student.

lib/WeBWorK/ContentGenerator/Grades.pm
```diff
@@ -202,7 +205,8 @@ sub displayStudentStats ($c, $studentID) {
 			# For other sets we just count the number of problems.
 			$num_of_problems = $db->countGlobalProblems($setID);
 		}
-		$max_problems = $max_problems < $num_of_problems ? $num_of_problems : $max_problems;
+		$max_problems =
+			(($set && after($set->open_date)) && $max_problems < $num_of_problems) ? $num_of_problems : $max_problems;
```
drgrice1, line 209: There are extra parentheses here that aren't needed.  The only necessary parentheses are the ones around the argument of the `after` method.

lib/WeBWorK/ContentGenerator/Grades.pm
```diff
@@ -215,22 +219,56 @@ sub displayStudentStats ($c, $studentID) {
...
-							colspan => $max_problems + 3,
-							$c->tag('em', $c->maketext('No versions of this assignment have been taken.'))
+							'em',
+							after($set->open_date) ? $c->maketext('No versions of this test have been taken.')
+							: $c->maketext(
+								'Will open on [_1].',
+								$c->formatDateTime($set->open_date, undef, $ce->{studentDateDsiplayFormat})
+							)
 						)
+
```
drgrice1, line 268: There is an empty line here that shouldn't be.

drgrice1 (approved):
This looks good.  Just a few quibbles on code style.

PR #1912, opened 2023-03-18
Add remaining time column to student progress table for tests.

somiaj:
This optional column will compute the amount of time each student has left on a timed test, which can vary depending on when the student started the test. I would find this useful when giving a timed test in class.


lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
```diff
@@ -148,10 +150,25 @@ sub displaySets ($c) {
...
 					$testTime = $timeLimit if ($testTime > $timeLimit);
 					$testTime = sprintf('%3.1f min', $testTime);
+					$timeLeft = '0 min';
+					if ($showColumns{timeleft} && time - $set->open_date < $set->version_time_limit) {
+						# Get a problem to determine how many submits have been made.
+						my @ProblemNums = $db->listUserProblems($studentRecord->user_id, $setName);
+						my $Problem =
+							$db->getMergedProblemVersion($studentRecord->user_id, $setName, $vNum, $ProblemNums[0]);
+						my $verSubmits =
+							(defined $Problem && $Problem->num_correct ne '')
```
drgrice1, line 160: There is no need to check that `$Problem->num_correct ne ''`.  I suspect you get this from line 507 of GatewayQuiz.pm.  The check is also not needed there.

somiaj: I probably originally got it from GatewayQuiz.pm when I created the check in ProblemSet.pm. I removed those three locations of that check. I didn't find any more instances grepping for `num_correct ne` in the source.

drgrice1 (approved):
This pull request looks good.

drgrice1:
There is a minor issue in `templates/ContentGenerator/Instructor/StudentProgress/set_progress.html.ep` that I saw when testing this.  I converted the quoting to using `q{...}` when I migrated this from StudentProgress.pm, but missed a backslash on line 80 of that file.  Could you delete that backslash?  If not, I can do that in another pull request.

lib/WeBWorK/ContentGenerator/Instructor/StudentProgress.pm
```diff
@@ -148,10 +150,22 @@ sub displaySets ($c) {
 					my $timeLimit = $set->version_time_limit / 60;
 					$testTime = $timeLimit if ($testTime > $timeLimit);
 					$testTime = sprintf('%3.1f min', $testTime);
+					$timeLeft = '0 min';
```
pstaabp, line 153: Should we wrap the 'min' on this line and L152 and L160, 165 in `maketext`?

somiaj: I considered that, but went with this approach to match what was already being done. Probably something we should fix now.

PR #1915, opened 2023-03-21
Fix the add problem button in the library browser.

### LINKED ISSUE #1913: undefined source value when add problem from Library Browser
on develop, if you add a problem from the library browser to a set and then view the set in ProblemSet Detail page, the problem source is "undefined".

drgrice1:
This was an accidental regression in #1887.

This fixes #1913.


pstaabp (approved):
Fixed and minor enough that I will just merge.

PR #1918, opened 2023-03-23
Pgeditor, option to backup up files when saving.

somiaj:
When saving a file with the PG editor, a backup can be made that could be reverted to if desired. This is done by first setting the number of backups to keep. Then each time a problem is saved there is an option to create a backup of the previous saved file (defaults to yes). Backups are numbered 1, 2, 3, ..., with 1 being the most recent. Backups are saved in the `tmpEdit` directory next to the temp files.

The revert tab then has an additional option to allow users to revert to either the current live version or one of the backups. If a user selects to revert to a backup, only the temporary file is reverted, and the user will still need to save the file to make it available to students.

In addition rename "Update" as "Save" and "New Version" as "Save As" to be more standard.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -995,6 +995,9 @@ $options{useContribDefFiles} = 1;
 # just a textarea
 $options{PGCodeMirror} = 1;
 
+# Number of backups to keep when saving problems in the problem editor.
+$options{editorNumberOfBackups} = 0;
+
 ###########################################################################################
 #### Default settings for the PG translator
 #### This section controls the display of equations, HINTS, answers, SOLUTIONS,
@@ -1566,6 +1569,17 @@ $ConfigValues = [
 			),
 			type    => 'number',
 			hashVar => '{pg}->{answersOpenAfterDueDate}'
+		},
+		{
+			var  => 'options{editorNumberOfBackups}',
+			doc  => x('Number of backups to keep in the PG editor (0 to disable backups)'),
+			doc2 => x(
+					'If this is greater than zero, two additional options become available when saving files with '
+					. 'the problem editor, backup previous file and delete oldest backup. This sets the number of backups '
+					. 'to keep before suggesting to delete the oldest backup. Backup files can be either restored or '
+					. 'deleted under the "Revert" tab. Setting to 0 disables all backup features.'
+			),
+			type => 'number'
 		},
 	],
 	[
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -995,9 +995,6 @@ $options{useContribDefFiles} = 1;
 # just a textarea
 $options{PGCodeMirror} = 1;
 
-# Number of backups to keep when saving problems in the problem editor.
-$options{editorNumberOfBackups} = 0;
-
 ###########################################################################################
 #### Default settings for the PG translator
 #### This section controls the display of equations, HINTS, answers, SOLUTIONS,
@@ -1569,17 +1566,6 @@ $ConfigValues = [
 			),
 			type    => 'number',
 			hashVar => '{pg}->{answersOpenAfterDueDate}'
-		},
-		{
-			var  => 'options{editorNumberOfBackups}',
-			doc  => x('Number of backups to keep in the PG editor (0 to disable backups)'),
-			doc2 => x(
-					'If this is greater than zero, two additional options become available when saving files with '
-					. 'the problem editor, backup previous file and delete oldest backup. This sets the number of backups '
-					. 'to keep before suggesting to delete the oldest backup. Backup files can be either restored or '
-					. 'deleted under the "Revert" tab. Setting to 0 disables all backup features.'
-			),
-			type => 'number'
 		},
 	],
 	[
--- a/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
@@ -192,17 +192,17 @@ sub pre_header_initialize ($c) {
 	}
 
 	# Initialize these values in case of failure in the getFilePaths method.
-	$c->{editFilePath}  = '';
-	$c->{tempFilePath}  = '';
-	$c->{inputFilePath} = '';
-	$c->{backBasePath}  = '';
+	$c->{editFilePath}   = '';
+	$c->{tempFilePath}   = '';
+	$c->{inputFilePath}  = '';
+	$c->{backupBasePath} = '';
 
 	# Determine the paths for the file.
 	# getFilePath defines:
-	#   $c->{editFilePath}:  path to the permanent file to be edited
-	#   $c->{tempFilePath}:  path to the temporary file to be edited with .tmp suffix
-	#   $c->{inputFilePath}: path to the file for input, (this is either the editFilePath or the tempFilePath)
-	#   $c->{backBasePath}:  base path to the backup files
+	#   $c->{editFilePath}:    path to the permanent file to be edited
+	#   $c->{tempFilePath}:    path to the temporary file to be edited with .tmp suffix
+	#   $c->{inputFilePath}:   path to the file for input, (this is either the editFilePath or the tempFilePath)
+	#   $c->{backupBasePath}:  base path to the backup files
 	$c->getFilePaths;
 
 	# Default problem contents
@@ -464,10 +464,10 @@ sub isTempEditFilePath ($c, $path) {
 }
 
 # Determine file paths. This defines the following variables:
-#   $c->{editFilePath}  -- path to permanent file
-#   $c->{tempFilePath}  -- temporary file name to use (may not exist)
-#   $c->{inputFilePath} -- actual file to read and edit (will be one of the above)
-#   $c->{backBasePath}  -- base path to backup files
+#   $c->{editFilePath}    -- path to permanent file
+#   $c->{tempFilePath}    -- temporary file name to use (may not exist)
+#   $c->{inputFilePath}   -- actual file to read and edit (will be one of the above)
+#   $c->{backupBasePath}  -- base path to backup files
 sub getFilePaths ($c) {
 	my $ce   = $c->ce;
 	my $db   = $c->db;
@@ -553,9 +553,9 @@ sub getFilePaths ($c) {
 	}
 
 	# The path to the permanent file is now verified and stored in $editFilePath
-	$c->{editFilePath} = $editFilePath;
-	$c->{tempFilePath} = $c->determineTempEditFilePath($editFilePath);
-	$c->{backBasePath} = $c->{tempFilePath} =~ s/.$user.tmp/.bak/r;
+	$c->{editFilePath}   = $editFilePath;
+	$c->{tempFilePath}   = $c->determineTempEditFilePath($editFilePath);
+	$c->{backupBasePath} = $c->{tempFilePath} =~ s/.$user.tmp/.bak/r;
 
 	# $c->{inputFilePath} is $c->{tempFilePath} if it is exists and is readable.
 	# Otherwise it is the original $c->{editFilePath}.
@@ -565,22 +565,22 @@ sub getFilePaths ($c) {
 }
 
 sub getBackupTimes ($c) {
-	my $backBasePath = $c->{backBasePath};
-	my @files        = glob("$backBasePath*");
+	my $backupBasePath = $c->{backupBasePath};
+	my @files          = glob("$backupBasePath*");
 	return unless @files;
-	return reverse(map { $_ =~ s/$backBasePath//r } @files);
+	return reverse(map { $_ =~ s/$backupBasePath//r } @files);
 }
 
 sub backupFile ($c, $outputFilePath) {
 	my $ce = $c->ce;
 	return unless $ce->{options}{editorNumberOfBackups} && $ce->{options}{editorNumberOfBackups} > 0;
 
-	my $backupTime   = time;
-	my $backFilePath = $c->{backBasePath} . $backupTime;
+	my $backupTime     = time;
+	my $backupFilePath = $c->{backupBasePath} . $backupTime;
 
 	# Make sure any missing directories are created.
-	surePathToFile($ce->{courseDirs}{templates}, $backFilePath);
-	copy($outputFilePath, $backFilePath);
+	surePathToFile($ce->{courseDirs}{templates}, $backupFilePath);
+	copy($outputFilePath, $backupFilePath);
 	$c->addgoodmessage($c->maketext(
 		'Backup created on [_1]',
 		$c->formatDateTime($backupTime, undef, $ce->{studentDateDisplayFormat})
@@ -588,15 +588,15 @@ sub backupFile ($c, $outputFilePath) {
 
 	# Delete oldest backup if option is present.
 	if ($c->param('deleteBackup')) {
-		my @backupTimes    = $c->getBackupTimes;
-		my $backupTime     = @backupTimes[$#backupTimes];
-		my $backFilePath   = $c->{backBasePath} . $backupTime;
-		my $formatBackTime = $c->formatDateTime($backupTime, undef, $ce->{studentDateDisplayFormat});
-		if (-e $backFilePath) {
-			unlink($backFilePath);
-			$c->addgoodmessage($c->maketext('Deleted backup from [_1].', $formatBackTime));
+		my @backupTimes      = $c->getBackupTimes;
+		my $backupTime       = @backupTimes[$#backupTimes];
+		my $backupFilePath   = $c->{backupBasePath} . $backupTime;
+		my $formatBackupTime = $c->formatDateTime($backupTime, undef, $ce->{studentDateDisplayFormat});
+		if (-e $backupFilePath) {
+			unlink($backupFilePath);
+			$c->addgoodmessage($c->maketext('Deleted backup from [_1].', $formatBackupTime));
 		} else {
-			$c->addbadmessage($c->maketext('Unable to delete backup from [_1].', $formatBackTime));
+			$c->addbadmessage($c->maketext('Unable to delete backup from [_1].', $formatBackupTime));
 		}
 	}
 }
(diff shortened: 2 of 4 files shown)


templates/ContentGenerator/Instructor/PGProblemEditor/revert_form.html.ep
```diff
@@ -1,5 +1,65 @@
...
+			% if (scalar(@backupNumbers) == 1) {
+				<%= hidden_field 'action.revert.backup.number' => $backupNumbers[0] =%>
+			% } else {
+			<div class="col-auto">
+				<%= select_field 'action.revert.backup.number' => [
+					map { [ $_ => $_ ] } @backupNumbers
+				],
+				id    => 'action_revert_backup_number_id',
+				class => 'form-select form-select-sm d-inline w-auto' =%>
+			</div>
```
drgrice1, line 55: There is a missing level of indentation here.

templates/ContentGenerator/Instructor/PGProblemEditor/save_form.html.ep
```diff
@@ -10,5 +10,13 @@
 			$c->{file_type} eq 'hardcopy_header' ? (checked => undef) : () =%>
 		<%= label_for newWindowSave => maketext('Open in new window'), class => 'form-check-label' =%>
 	</div>
+	% if ($c->ce->{options}{editorNumberOfBackups} > 0) {
+		<div class="form-check mb-2">
+			<%= check_box backupOldFile => 1, id => 'backupOldFile', class => 'form-check-input',
+				checked => undef =%>
+			<%= label_for backupOldFile => maketext('Create backup of old file'),
+				class => 'form-check-label' =%>
```
drgrice1, line 18: These two lines don't need to be wrapped.  They fit into the 120 character limit without wrapping.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -559,11 +559,27 @@ sub getFilePaths ($c) {
...
-	$problemContents = ${ $c->{r_problemContents} } unless not_blank($problemContents);
+	# Copy backup files to the next file in the chain.
+	for (reverse(1 .. $numberOfBackups)) {
+		my $prevFile   = ($_ == 1) ? $outputFilePath : $tempFilePath . '.bak' . ($_ - 1);
+		my $backupFile = $tempFilePath . ".bak$_";
+		copy($prevFile, $backupFile) if (-e $prevFile);
+	}
+}
+
+sub saveFileChanges ($c, $outputFilePath, $backup) {
```
drgrice1, line 580: Add a default value for `$backup` so that you don't need to pass the argument `0` below in the cases that a backup is not provided.  Something like
```suggestion
sub saveFileChanges ($c, $outputFilePath, $backup = undef) {
```
I prefer `undef` to `0` in this case.  It is a more logical alternative to a string value that is not passed.  It doesn't change the usage in this method if it is undefined rather than zero.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -696,7 +715,7 @@ sub view_handler ($c) {
 	# Grab the problemContents from the form in order to save it to the tmp file.
 	$c->{r_problemContents} = \(fixProblemContents($c->param('problemContents')));
 
-	$c->saveFileChanges($c->{tempFilePath});
+	$c->saveFileChanges($c->{tempFilePath}, 0);
```
drgrice1, line 718: No need to pass `0` if the suggested default is added.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -1030,7 +1050,7 @@ sub save_as_handler ($c) {
 	}
 
 	unless ($do_not_save) {
-		$c->saveFileChanges($outputFilePath);
+		$c->saveFileChanges($outputFilePath, 0);
```
drgrice1, line 1053: And here also.  No need to pass `0` if the suggested default is added.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -559,11 +559,27 @@ sub getFilePaths ($c) {
...
 }
 
-sub saveFileChanges ($c, $outputFilePath, $problemContents = undef) {
-	my $ce = $c->ce;
+sub backupFile ($c, $outputFilePath) {
+	my $ce              = $c->ce;
+	my $numberOfBackups = $ce->{options}{editorNumberOfBackups} || 0;
+	my $user            = $c->param('user');
+	my $tempFilePath    = $c->{tempFilePath} =~ s/.$user.tmp$//r;
+	return unless $numberOfBackups > 0;
```
drgrice1, line 567: This return statement should be immediately after line 564 where `$numberOfBackups` is defined.  The other variables aren't needed to return.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -559,11 +559,27 @@ sub getFilePaths ($c) {
...
 
-	$problemContents = $$problemContents if defined $problemContents && ref $problemContents;
-	$problemContents = ${ $c->{r_problemContents} } unless not_blank($problemContents);
+	# Copy backup files to the next file in the chain.
+	for (reverse(1 .. $numberOfBackups)) {
+		my $prevFile   = ($_ == 1) ? $outputFilePath : $tempFilePath . '.bak' . ($_ - 1);
+		my $backupFile = $tempFilePath . ".bak$_";
+		copy($prevFile, $backupFile) if (-e $prevFile);
+	}
+}
```
drgrice1, line 578: This is a convoluted scheme and involves excessive disk access.  It would be better to use a time stamp in the file name, rather than a simple number.  Something like `20230323-184633.2356` perhaps.  Then that time stamp could also be displayed in the ui select to give the user a more meaningful reference as to which backup to restore than a number that only gives an indication of order.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -906,7 +925,8 @@ sub save_handler ($c) {
 				. 'The problem set may have changed. Please reopen this file from the homework sets editor.'
 		));
 	} else {
-		$c->saveFileChanges($c->{editFilePath});
+		my $backup = $c->param('backupOldFile') || 0;
+		$c->saveFileChanges($c->{editFilePath}, $backup);
```
drgrice1, line 929: There is not need for the intermediary variable that is only used once.  This should be
```suggestion
		$c->saveFileChanges($c->{editFilePath}, $c->param('backupOldFile'));
```
Note there is also no need for the `|| 0` as undefined (or even the empty string) will still be logically false, and so will work the same in the `saveFileChanges` method.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -1193,15 +1212,32 @@ sub revert_handler ($c) {
...
+	my $revertType   = $c->param('action.revert.file') || 'do_not_revert';
+	my $backupNumber = $c->param('action.revert.backup.number');
+	my $backFilePath = $revertType eq 'backup' ? ($c->{tempFilePath} =~ s/.$user.tmp$//r) . '.bak' . $backupNumber : '';
+
+	if ($revertType eq 'tmp') {
+		$c->{inputFilePath} = $c->{editFilePath};
+
+		# Unlink the temp files;
+		unlink($c->{tempFilePath});
+		$c->addgoodmessage($c->maketext('Deleted temporary file "[_1]".',    $c->shortPath($c->{tempFilePath})));
```
drgrice1, line 1225: This reports that a temporary file is deleted even if no temporary file was deleted.  This wasn't an issue previously because the revert form was only shown if a temporary file existed.  That is not longer the case.  So more care is needed to accurately report what has happened.

drgrice1 (changes requested):
This implementation is quite confusing and needs more work.

One of the things that is not helpful is to use a mere count to identify backups.  Adding a timestamp as suggested would help with that as it gives a more meaningful point of reference for the user to pick which backup to choose from.

Another thing that is extremely confusing is that once a backup is made, the revert form is always there.  I saved the file I was working on with a backup, and then reverted to the original file.  After doing so, not only is the revert form still there, but listed in the form is the option to revert to the original file.  Wait, I just did that.

There also needs to be some way to get rid of backups you no longer want.  Once the backups are made you get this list of backups, but no way to clean them up other than going to the file browser (or admin access file deletion of course).

somiaj:
Applied the suggested changes. Backups now use a timestamp to order them. There are now two options on the Revert tab for backups, either restore a backup or delete it. Backups are identified using the date of the timestamp formatted using the `studentDateDisplayFormat`. I did think about making a new Backup Tab for this, though unsure if one would be needed, as it does fit under the Revert tab. Maybe rename it to Revert/Backups.

When updating the message output and adding the logic of when to delete the oldest backup, I first went with changing the 'Create backup' option to 'Create backup and delete oldest backup' when the desired number of backups were reached to inform the user a backup would be deleted. I then morphed this into two options on the Save tab, that is both 'Create backup' and 'Delete oldest backup'. Where 'Delete oldest backup' is checked once the desired number of backups are reached, but the user can uncheck it if wanted.

I need to figure out how to set the param's for checkboxes and radiobuttons so they are rest to defaults when the take action (form is submitted) occurs, otherwise they keep their previous selected values.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
somiaj: I have fixed it so the Revert option will not show up if there isn't a temp file like the way it use to be for the revert tab. Though the only time there isn't a temp file is before the user clicks view. Should the Revert tab and/or option to revert appear when the temp file is created by viewing/reloading the file (otherwise you have to reload the page to see it)?

drgrice1:
I think one tab is all that is needed.  This is better.  If there are backups or a temporary file is in use, then then the revert tab is shown.  Now that you have made what is shown more accurate it is less confusing.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
drgrice1: That is not quite true.  There will also not be a temporary file if the user has just reverted.  That is the case that was confusing before.

I almost implemented making the revert tab appear when viewing/reloading the file when I revising the PG problem editor.  I didn't because it is troublesome to do so.  It isn't as easy to do as you might think.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -195,12 +195,14 @@ sub pre_header_initialize ($c) {
 	$c->{editFilePath}  = '';
 	$c->{tempFilePath}  = '';
 	$c->{inputFilePath} = '';
+	$c->{backBasePath}  = '';
```
drgrice1, line 198: Please change this to `backupBasePath` throughout this file.  That is a really bad shortening of `backup`.

drgrice1:
@somiaj:  I think if you merge the two pull request I have put in to this branch, change `backBasePath` to `backupBasePath` as already mentioned, and remove the course configuration option, then this pull request is good.

I think the course configuration option is really not needed with the check boxes that give the user direct control of when a backup is made, and allows the user to delete backups.

somiaj:
@drgrice1 Accepted pull requests and fixed the variable naming. Two things:

First there is a small issue with disabling the revert radio button, is that the revert option is still selected after reverting from the temporary file. Thus if a user hits take action again it will do the revert action again before a temporary file has been selected. I think this can be fixed by moving the selected option to restore backup in this case.

Second, just to clarify on your suggestion on removing the course configuration. Do you mean just remove the ability to set the `$options{editorNumberOfBackups}` in the UI? So, keep this variable to control the number backups to keep before recommending deleting old backups, but don't give any option to modify it in the UI? Or do you mean remove this variable altogether, and let the user check when they want to delete old backups?

drgrice1:
> > First there is a small issue with disabling the revert radio button, is that the revert option is still selected after reverting from the temporary file. Thus if a user hits take action again it will do the revert action again before a temporary file has been selected. I think this can be fixed by moving the selected option to restore backup in this case.
> 
> Hmm... I will look into that. It should not be switched to restoring a backup either. It should be set to the default view/reload.

Actually, now that I test this I see that this is not actually happening.  After reverting from the temporary file the view/reload action is enabled.  So I am not sure what you mean?

somiaj:
> Actually, now that I test this I see that this is not actually happening. After reverting from the temporary file the view/reload action is enabled. So I am not sure what you mean?

After a revert, go back to the revert tab the radio button for revert from temporary file is disabled, but still selected. So if u click Take Action from this tab, you do another revert even though there is no temporary file. Which is why I suggested to change what is selected on that tab in this case.

drgrice1:
Yeah, I figured out what you meant.  I am looking into it.

drgrice1:
@somiaj: I added a pull request that switches the checked radio to the backup radio after reverting the temporary file.

pstaabp (approved):
I think this looks good. This will save some of those overwrite mistakes.

somiaj:
Does anyone else have an opinion on if there should be a configuration option to control when we suggest deleting old backups like it is now, or should this just removed, and let the user manage backups fully from the editor? My original thought was something that would help keep the backups from piling up. Another option might be delete them after they are X-time old.

pstaabp:
> Does anyone else have an opinion on if there should be a configuration option to control when we suggest deleting old backups like it is now, or should this just removed, and let the user manage backups fully from the editor? My original thought was something that would help keep the backups from piling up. Another option might be delete them after they are X-time old.

Although I didn't test this.  Let's say you select 2 automatic backups.  If you have saved a file 3 times, that there is the current file and two backup files.  If you save another time, the oldest is deleted.  Is this not how the current branch works?

somiaj:
> Although I didn't test this. Let's say you select 2 automatic backups. If you have saved a file 3 times, that there is the current file and two backup files. If you save another time, the oldest is deleted. Is this not how the current branch works?

Correct, currently if you set it to 2 backups, and you currently have 2 (or more) backups, the editor will automatically check the 'delete oldest backup' option when saving, which will make it so you when you save a file you can both create and delete a backup, thus keeping the number of backups from piling up if you save a lot.

@drgrice1 suggested just removing this altogether, since the backups can be manged and deleted from the editor, which will no longer automatically delete any and require the user manually delete old backups.

I do like the idea of not letting backups build up, but the simple 'keep N backups' has its flaws if someone saves and backups the file a lot. So the suggestion is to remove that whole feature (including the 'delete oldest backup' check box) and just let users delete backups when they want (which my guess is most won't and they will pile up).

drgrice1:
I did not say to remove the delete oldest backup checkbox. I only said to remove the limit on the number of backups which is completely misleading. The point is that when you have the delete oldest backup checkbox which gives the user the option to override the limit the option is rather pointless. That option is a soft limit.  If one wants to limit backups they only need to keep the save back up and delete oldest back up checkboxes both checked.

Alex-Jordan:
My two cents. Go with the simple 'keep N backups' model. It's already
infinitely many times more backups than we currently do. And I think we
could replace this with a versioning approach, given more time to develop
it. And then some issues go moot, since versioning can be way more
efficient than backup copies.

drgrice1:
The thing is that the 'editorNumberOfBackups' option is not a 'editorNumberOfBackups' option.  It should be renamed to 'editorNumberOfBackupsBeforeSuggestingALimit' option.  The option does NOT limit the number of backups.  I am saying remove the option, and let the user choose the number of backups to keep.  Once the user reaches the number of backups they want, then they check the 'Delete oldest backup' check box.  That truly does limit the number of backups.  On the other hand, if a user wants no limit, the user simply needs to not check that check box.  If the user doesn't want any backups, then uncheck the 'Save backup' check box, and no backups will be saved.  The check boxes will stay in the state the user sets them to.  Everything can be done with the check boxes as implemented, and the course config option is redundant.

Implementing an actual versioning system would be better in any case.  One way this can be done is by using local a git repository that is created and updated automatically by the server when saving.

conf/defaults.config
```diff
@@ -1567,6 +1570,17 @@ $ConfigValues = [
 			type    => 'number',
 			hashVar => '{pg}->{answersOpenAfterDueDate}'
 		},
+		{
+			var  => 'options{editorNumberOfBackups}',
+			doc  => x('Number of backups to keep in the PG editor (0 to disable backups)'),
+			doc2 => x(
+					'If this is greater than zero, two additional options become available when saving files with '
+					. 'the problem editor, backup previous file and delete oldest backup. This sets the number of backups '
```
drgrice1, line 1578: This line exceeds the 120 character limit.  I added a pull request that fixes this.  See https://github.com/somiaj/webwork2/pull/8.  Either merge that, or just fix it yourself if you like.

somiaj:
I removed the course configuration option, and users can just manage the backups from the editor page.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -559,11 +564,46 @@ sub getFilePaths ($c) {
...
+		'Backup created on [_1]',
+		$c->formatDateTime($backupTime, undef, $ce->{studentDateDisplayFormat})
+	));
 
-	$problemContents = $$problemContents if defined $problemContents && ref $problemContents;
-	$problemContents = ${ $c->{r_problemContents} } unless not_blank($problemContents);
+	# Delete oldest backup if option is present.
+	if ($c->param('deleteBackup')) {
+		my @backupTimes      = $c->getBackupTimes;
+		my $backupTime       = @backupTimes[$#backupTimes];
```
drgrice1, line 592: This should be
```suggestion
		my $backupTime       = $backupTimes[-1];
```
`@backupTimes[$#backupTimes]` gives a warning.
Changing that to `$backupTimes[$#backupTimes]` will work without warning, but perlcritic will complain and say you should use -1 for the index.

drgrice1 (commented):
Something isn't right now.  Now no backups are saved regardless of if 'Create backup' is checked or not.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -559,11 +564,46 @@ sub getFilePaths ($c) {
...
+		my $backupFilePath   = $c->{backupBasePath} . $backupTime;
+		my $formatBackupTime = $c->formatDateTime($backupTime, undef, $ce->{studentDateDisplayFormat});
+		if (-e $backupFilePath) {
+			unlink($backupFilePath);
+			$c->addgoodmessage($c->maketext('Deleted backup from [_1].', $formatBackupTime));
+		} else {
+			$c->addbadmessage($c->maketext('Unable to delete backup from [_1].', $formatBackupTime));
+		}
+	}
+}
```
drgrice1, line 602: This method does not end with return.

somiaj:
Okay, it should work now. I missed adding a file to the previous commit, but it was working for me since I still had the old variable in my simple.conf file. Now that I added removing the check for the option it is working again.

drgrice1 (approved):
Everything seems to be working now.  Looks good.

Alex-Jordan:
Recent work with hardcopy themes has me poring over the editor interface. I noticed that it is a little weird that when you want to delete backups, you choose the radio button for deleting a backup and then you click the `Revert` button. (Maybe at the time this was written, the button still said `Take Action!`.)

I can't think of a quick and easy way to make this more smooth (or I would just do it an open a PR). We could live with it, or maybe someone else has an idea.

PR #1922, opened 2023-03-26
Change syntax highlighting for non PG files in PGCodeMirror.

somiaj:
When using PGCodeMirror to edit course_info pages, use the htmlmixed to highlight html, and when editing achivement files the base perl mode is used.

### DIFF
--- a/htdocs/js/apps/PGCodeMirror/pgeditor.js
+++ b/htdocs/js/apps/PGCodeMirror/pgeditor.js
@@ -69,7 +69,7 @@
 	};
 
 	const cm = webworkConfig.pgCodeMirror = CodeMirror.fromTextArea(document.querySelector('.codeMirrorEditor'), {
-		mode: 'PG',
+		mode: document.querySelector('.codeMirrorEditor')?.dataset.mode ?? 'PG',
 		indentUnit: 4,
 		tabMode: 'spaces',
 		lineNumbers: true,
--- a/lib/WeBWorK/HTML/CodeMirrorEditor.pm
+++ b/lib/WeBWorK/HTML/CodeMirrorEditor.pm
@@ -63,10 +63,10 @@ use constant CODEMIRROR_ADDONS_JS => [
 	'scroll/annotatescrollbar.js', 'edit/matchbrackets.js'
 ];
 
-sub generate_codemirror_html ($c, $name, $contents = '') {
+sub generate_codemirror_html ($c, $name, $contents = '', $mode = 'PG') {
 	# Output the textarea that will be used by CodeMirror.
 	# If CodeMirror is disabled, then this is directly the editing area.
-	return $c->text_area($name => $contents, id => $name, class => 'codeMirrorEditor');
+	return $c->text_area($name => $contents, id => $name, class => 'codeMirrorEditor', data => { mode => $mode });
 }
 
 sub generate_codemirror_controls_html ($c) {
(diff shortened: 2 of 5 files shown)


htdocs/js/apps/PGCodeMirror/pgeditor.js
```diff
@@ -69,7 +69,7 @@
 	};
 
 	const cm = webworkConfig.pgCodeMirror = CodeMirror.fromTextArea(document.querySelector('.codeMirrorEditor'), {
-		mode: 'PG',
+		mode: document.querySelector('.codeMirrorEditor').dataset.mode,
```
drgrice1, line 72: This would be better as
```suggestion
		mode: document.querySelector('.codeMirrorEditor')?.dataset.mode ?? 'PG',
```
This way if the `.codeMirrorEditor` element does not have the `mode` dataset attribute set, then it will still work.

lib/WeBWorK/HTML/CodeMirrorEditor.pm
```diff
@@ -63,10 +63,18 @@ use constant CODEMIRROR_ADDONS_JS => [
...
 ];
 
-sub generate_codemirror_html ($c, $name, $contents = '') {
+# Return an array of javascript dependencies for syntax highlighting modes.
+sub get_codemirror_mode_deps ($mode) {
+	if ($mode eq 'htmlmixed') {
+		return [ 'xml', 'css', 'javascript', 'htmlmixed' ];
+	}
+	return [$mode];
+}
```
drgrice1, line 72: Please delete this method.  There is no reason to add a method that is only called once, and that can be implemented in one line in the one place that it is called.

lib/WeBWorK/HTML/CodeMirrorEditor.pm
```diff
@@ -89,11 +97,12 @@ sub generate_codemirror_controls_html ($c) {
...
 }
 
-sub output_codemirror_static_files ($c) {
+sub output_codemirror_static_files ($c, $mode = 'PG') {
 	return $c->include(
 		'HTML/CodeMirrorEditor/js',
 		codemirrorAddonsCSS => CODEMIRROR_ADDONS_CSS(),
-		codemirrorAddonsJS  => CODEMIRROR_ADDONS_JS()
+		codemirrorAddonsJS  => CODEMIRROR_ADDONS_JS(),
+		codemirrorModesJS   => get_codemirror_mode_deps($mode),
```
drgrice1, line 105: After removing the `get_codemirror_mode_deps`, please change this to
```suggestion
		codemirrorModesJS   => $mode eq 'htmlmixed' ? [ 'xml', 'css', 'javascript', 'htmlmixed' ] : [$mode]
```

lib/WeBWorK/HTML/CodeMirrorEditor.pm
```diff
@@ -63,10 +63,18 @@ use constant CODEMIRROR_ADDONS_JS => [
...
+		return [ 'xml', 'css', 'javascript', 'htmlmixed' ];
+	}
+	return [$mode];
+}
+
+sub generate_codemirror_html ($c, $name, $contents = '', $mode = 'PG') {
 	# Output the textarea that will be used by CodeMirror.
 	# If CodeMirror is disabled, then this is directly the editing area.
-	return $c->text_area($name => $contents, id => $name, class => 'codeMirrorEditor');
+	return $c->text_area($name => $contents, id => $name, class => 'codeMirrorEditor', 'data-mode' => $mode);
```
drgrice1, line 77: This should be
```suggestion
	return $c->text_area($name => $contents, id => $name, class => 'codeMirrorEditor', data => { mode => $mode });
```
Mojolicious tag helpers have special handling for data attributes.  Please use it.

drgrice1 (commented):
There are a few things that I see that could be done better.  See my comments.

Other than those things, this looks good.

drgrice1 (approved):
Thanks for making those changes.  This looks good.

pstaabp (approved):

Issue #1921, opened 2023-03-26
attemptsTable not using displaystyle formatting


somiaj:
I noticed that the attempts table preview and correct answer columns don't seem to use displaymode any longer (at least consistently). Using `Compute('x/3')` as my answer gives the following table:

![image](https://user-images.githubusercontent.com/18082412/227758230-f2528162-9218-4ab8-b09c-a2b76613af77.png)

In looking into this I noticed that the change to mojolicious changed:

```perl
return '<script type="math/tex; mode=display">' . $tex . '</script>';
```

to

```perl
return $self->c->tag('script', type => 'math/tex', mode => 'display', $self->c->b($tex));
```

which took the `mode=display` option and moved it to its own tag. If I move it back, I seem to get displaystyle working on the two columns again.

```perl
return $self->c->tag('script', type => 'math/tex; mode=display', $self->c->b($tex));
```

![image](https://user-images.githubusercontent.com/18082412/227758306-bede70c7-e6de-43a5-8c3b-2b529b405639.png)

drgrice1:
That is an obvious mistake.  Can you put in a pull request?

drgrice1:
Actually, I have a couple of other changes to the attempts table in a branch already.  I will add this, and put it all in together.

pstaabp:
fixed in #1924

PR #1924, opened 2023-03-28
Minor improvements and fixes for the attempts table.

drgrice1:
Change the preview span to a preview div.  We make that a span, but set it as display `block`.  If you want display block, generally you should be using a div.  Furthermore, many answer previews now use divs for layout which results in invalid html inside a span.  (This requires a change to the PG problem css.  See https://github.com/openwebwork/pg/pull/796.)

Also wrap the attempts table in a `table-responsive` div.  This cleans up the page size on narrow screens or when the results table becomes excessively large (as can quite frequently occur with messages and correct answers).

Also fix display mode for MathJax equations in answer previews.  This fixes issue #1921.

### DIFF
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -87,7 +87,7 @@
 	);
 
 	// Set up popovers in the attemptResults table.
-	document.querySelectorAll('table.attemptResults td span.answer-preview').forEach((popover) => {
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
Looks good.

somiaj (approved):
Fixes the displaystyle issue in attempts table.

PR #1925, opened 2023-03-30
initialize two variables

Alex-Jordan:
The last two variables here were not being initialized. It led to a warning when I imported a set.


drgrice1 (approved):
Good catch.  One of the missing counts is my fault.  I should have incremented it when I added the `showHintsAfter` option.  However, this was already incorrect at that time.  In the main branch there are 17 variables but only 16 were initialized.

I will go ahead and merge this.  It is obvious.

PR #1927, opened 2023-03-31
Remove the Wiris editor.

drgrice1:
This was never fully developed, and was never fully functional.  It has been sitting here untouched for years, and so it seems clear that there is little interest in this.  So bye-bye!

Note that this is paired with https://github.com/openwebwork/pg/pull/798.


pstaabp (approved):
Bye Bye 👋

drdrew42 (approved):
👋

PR #1928, opened 2023-03-31
Fix readme and delete duplicate README_js_organization

dlglin:
Update README.md to reduce the amount of clutter, and remove README_js_organization from the root folder.  There is another copy of README_js_organization in htdocs/js which is more up-to-date.


pstaabp:
Just a note, but there is the `/bin/dev_scripts/update-copyright` script that we usually run before the release.   Not totally sure this runs on `.md` files.

pstaabp (approved):

drgrice1:
Can you also delete the bare README file.  We don't need that and the README.md file.

Note that the update-copyright script does not update the README.md file.  That does need to be changed manually.

README.md
```diff
@@ -3,58 +3,33 @@
...
+WeBWorK is an open-source online homework system for math and sciences courses. WeBWorK is supported by the MAA and the NSF and comes with an Open Problem Library (OPL) of over 30,000 homework problems. Problems in the OPL target most lower division undergraduate math courses, some advanced courses and some other STEM subjects. Supported courses include college algebra, discrete mathematics, probability and statistics, single and multivariable calculus, differential equations, linear algebra and complex analysis.  Find out more at the main WeBWorK [webpage](https://openwebwork.org).
 
 ## Information for Users
 
 New users interested in getting started with their own WeBWorK server, or instructors looking to learn more about how to use WeBWorK in their classes, should take a look at one of the following resources:
 *  The [WeBWorK project home page](https://openwebwork.org/) - General information and resources including announcements of events and important project news
```
drgrice1, line 18: Could you also add an empty line before this line.  Lists in makrdown should always be surrounded by blank lines according to markdownlint.

drgrice1 (approved):
This looks good.

PR #1929, opened 2023-04-02
Play the z-index game.

drgrice1:
There are many things that are being shown above the masthead and the sticky nav that should not be.  For example, the scrollbars for the codemirror editor on the PG editor page in Firefox (but not Chrome), and the MathJax elements of the graph for a graphtool problem (those are set to z-index 9). I am sure there are other things as well.  Those are just the two I notice right now.

This increases the z-index from 3 to 20 for the masthead, and from 1 to 20 for the sticky nav.  Note that the sticky nav no longer needs to have higher z-index than the masthead now that it never scrolls up that high.

Note that this is still well below the z-index of the mqeditor toolbar, and any of the bootstrap elements that are meant to be over other things, so those will still show over both the masthead and stick nav.

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -51,7 +51,7 @@ table caption {
 
 .visually-hidden-focusable:active,
 .visually-hidden-focusable:focus {
-	z-index: 4;
+	z-index: 21;
 }
 
 $masthead-height: 70px !default;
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
In Safari, there are no code mirror scrollbars.  (hmmm), but the mathjax in the graphtool is fixed.

Alex-Jordan:
I tested and witnessed the cited changes. Merging.

I too see significant differences in styling of CodeMirror with Safari. But nothing that bothers me.

PR #1931, opened 2023-04-06
Fix issue with generating a single hardcopy test version.

### LINKED ISSUE #1930: hardcopy for tests for students giving error
I was testing #1809 and found an error, but this is in develop, so I don't think it is related to that PR. 

If you go to a test and on the overview page, click the hardcopy download button, you're taken to the Hardcopy page.  Clicking generate Hardcopy results in
```
Can't call method "open_date" on an undefined value at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Hardcopy.pm line 224.
```

somiaj:
The `$selected_set_id` name was updated to be displayed, but was still being used as the set id parameter. This moves the display change to the line it is being used to ensure the set id parameter is unchanged.

Fixes #1930


pstaabp (approved):
Fixes the issue.

drgrice1 (approved):
This fixes the problem.  Thanks.

PR #1934, opened 2023-04-06
Get rid of BLOBs

dlglin:
This changes `user_id` and `set_id` from TINYBLOB to VARCHAR(100).  The 100 character limit is not new, as it already exists in past answers.

`location_id` is set to VARCHAR(100).  This could be a problem for anyone who has existing locations with IDs that are more than 100 characters long.  Any newly created courses wouldn't be able to save those locations.

`lis_source_did` was changed from BLOB to TEXT.  This should probably be tested.

This still needs perltidying.


lib/WeBWorK/DB/Record/LocationAddresses.pm
```diff
@@ -26,7 +26,7 @@ use warnings;
 
 BEGIN {
 	__PACKAGE__->_fields(
-		location_id => { type => "TINYBLOB NOT NULL", key => 1 },    # requires up to 256 bytes
+		location_id => { type => "VARCHAR(100) NOT NULL", key => 1 },    # requires up to 256 bytes
```
taniwallach, line 29: From what I recall, with `utf8mb4` enabled it is necessary to keep the total length of they keys at under 1000 bytes, which only allows about 240 characters (each character takes up to 4 bytes, and from what I recall a few bytes are needed to set the length of the data in the VARCHAR record).

If `ip_mask` remains `VARCHAR(180)` then `location_id` should probably be capped at `VARCHAR(60)`.

However, the use in `lib/WeBWorK/DB/Record/UserSetLocations.pm` with two other `VARCHAR(100)` key fields, seems to imply that a limit of `VARCHAR(40)` may be needed.

See: https://dev.mysql.com/doc/refman/8.0/en/myisam-storage-engine.html

lib/WeBWorK/DB/Record/Locations.pm
```diff
@@ -26,7 +26,7 @@ use warnings;
 
 BEGIN {
 	__PACKAGE__->_fields(
-		location_id => { type => "TINYBLOB NOT NULL", key => 1 },
+		location_id => { type => "VARCHAR(100) NOT NULL", key => 1 },
```
taniwallach, line 29: Keep consistent with the size set in `lib/WeBWorK/DB/Record/LocationAddresses.pm`.

lib/WeBWorK/DB/Record/SetLocations.pm
```diff
@@ -26,8 +26,8 @@ use warnings;
 
 BEGIN {
 	__PACKAGE__->_fields(
-		set_id      => { type => "TINYBLOB NOT NULL", key => 1 },
-		location_id => { type => "TINYBLOB NOT NULL", key => 1 },
+		set_id      => { type => "VARCHAR(100) NOT NULL", key => 1 },
+		location_id => { type => "VARCHAR(100) NOT NULL", key => 1 },
```
taniwallach, line 30: Keep consistent with the size set in `lib/WeBWorK/DB/Record/LocationAddresses.pm`.

lib/WeBWorK/DB/Record/UserSetLocations.pm
```diff
@@ -26,9 +26,9 @@ use warnings;
 
 BEGIN {
 	__PACKAGE__->_fields(
-		user_id     => { type => "TINYBLOB NOT NULL", key => 1 },
-		set_id      => { type => "TINYBLOB NOT NULL", key => 1 },
-		location_id => { type => "TINYBLOB NOT NULL", key => 1 },
+		user_id     => { type => "VARCHAR(100) NOT NULL", key => 1 },
+		set_id      => { type => "VARCHAR(100) NOT NULL", key => 1 },
+		location_id => { type => "VARCHAR(100) NOT NULL", key => 1 },
```
taniwallach, line 31: Keep consistent with the size set in `lib/WeBWorK/DB/Record/LocationAddresses.pm`.
This file seems to be the one which causes the smallest limit on the size of `location_id` within the 1000 byte limit on myISAM key length.

lib/WeBWorK/DB/Record/LocationAddresses.pm
dlglin: I just did some fooling around, and MariaDB 10.5 doesn't complain about key size no matter how large I make the key fields, despite the fact that MariaDB claims to have the same maximum index length (https://mariadb.com/kb/en/myisam-overview/).  Does that mean it's just truncating any index entries that are too long?  I remember earlier versions used to give a "Specified key was too long" error.

`ip_mask` probably doesn't need to be that long.  I'm guessing that that size was chosen based on how big we could get away with subject to the limitations at the time.  Since there is a separate entry for each ip mask, I don't see why it should ever need to be more than 31 characters (something like "123.456.789.012-123.456.789.013" should be the longest format that can be parsed).

dlglin: After some more poking around:
MariaDB 10.4 or newer will allow a unique key that is more than 1000 bytes, but if you specify `KEY` rather than `UNIQUE KEY` it complains.  Both MariaDB 10.3 and mysql 8.0 complain about a `UNIQUE KEY` that is more than 1000 bytes, so the current version of this PR fails when trying to create the set_location_users table.

Since fewer people use locations I'm going to leave `user_id` and `set_id` at 100 characters and change `location_id` to 40 characters.  This could break things for anyone who is upgrading and has existing `location_id`s that are more than 40 characters long, but if such people exist we can help them individually.

taniwallach: Sounds like a reasonable compromise to me.

taniwallach (commented):
I recommend getting this all into once commit built on top of the current `WeBWorK-2.18` branch and force pushing it to your branch.
It seems to me that the current version should be within the 1000 byte key length limit of "older" versions of MariaDB, so this should certainly be OK for new courses on older database versions. An extra round of testing would be good.

lib/WeBWorK/DB/Record/UserSet.pm
```diff
@@ -61,7 +61,7 @@ BEGIN {
 		hide_hint                 => { type => "INT" },
 		restrict_prob_progression => { type => "INT" },
 		email_instructor          => { type => "INT" },
-		lis_source_did            => { type => "BLOB" },
+		lis_source_did            => { type => "TEXT" },
 		external_data             => { type => "TEXT" },
```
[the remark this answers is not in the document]
pstaabp: I think this is a good idea.

taniwallach (approved):
Reviewed code and we discussed these changes.
I have not tested this in practice yet, but it all makes sense.

PR #1935, opened 2023-04-06
Update user information in cached list when saving a user in the Classlist Editor.

### LINKED ISSUE #1933: Editing users with Class List Editor doesn't show changes right after being saved.
When editing users with the class list editor, the table that is generated after saving the changes is showing the previous information. To see this, edit a user in the class list editor, make a change to one of the fields, and save the change. After saving the user is shown in the user table without the changes. Reloading the class editor shows the changes, so they are being saved, just not being displayed right after the save.

drgrice1:
Since the database is not queried to retrieve updated information after actions are performed the data for saved users needs to be updated in the cached lists so that the updated information is displayed when the templates are loaded later.

This fixes #1933.

Also fix sorting that was also broken in #1888.  The fallback methods are needed in order for sorting to work correctly in case the first three do not sort thing properly.  Currently, if you load the page with users that do not have the information that the first three methods sort on, then the users end up in a different order each time you load the page.  For example, if using the default sort and users do not have first or last names.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
@@ -238,12 +238,16 @@ sub pre_header_initialize ($c) {
 	my $secondarySortSub = SORT_SUBS()->{ $c->{secondarySortField} };
 	my $ternarySortSub   = SORT_SUBS()->{ $c->{ternarySortField} };
 
-	$c->{allUserIDs}    = [ keys %allUsers ];
+	$c->{allUserIDs} = [ keys %allUsers ];
+
+	# Always have a definite sort order in case the first three sorts don't determine things.
 	$c->{sortedUserIDs} = [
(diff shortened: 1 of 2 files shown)


drgrice1:
Probably a better test to see the sorting issue than first and last name, and a test that would be more prevalent on an actual system is sort by section, then recitation, then comment.  On many systems you won't have any of these (and you probably wouldn't sort on them but this is to demonstrate the issue).  Of course to test this you need several users that don't have any of those pieces of information.  Select section, recitation, and comment in the sort form, and then click take action several times.  You will see the users that don't have those pieces of information in a different order each time.

On a production server you might have some users that don't have a section (like several instructors), and sort by section.  This would be enough to display this issue.

somiaj (approved):
This fixes both #1933 and the sorting problem in my tests.

pstaabp (approved):
Looks good.

Issue #1938, opened 2023-04-11
PGProblemEditor uses incorrect seed with problem submits


somiaj:
When editing a problem in the problem editor (that is attached to an assignment set), clicking on preview answer, check answer, or show correct answers in the problem frame doesn't use the seed configured in the editor, and instead uses some other seed. Here is an example problem to use in the problem editor.

```
DOCUMENT();
loadMacros(
    'PGstandard.pl',
    'PGML.pl',
);

BEGIN_PGML
Problem Seed: [$problemSeed]
END_PGML
ENDDOCUMENT();
```

View this problem in the problem frame, then click on randomize seed to see a new seed. Click on any of the buttons and watch the seed revert to some predefined (or default) seed. This appears to only happen when editing a problem in a set which you are assigned too. If you edit an unattached problem things works as expected and the seed doesn't get reset. Also the seed you get when editing a problem in a set you are assigned too is not the same as the seed you actually get when viewing the problem in the homework set.

drgrice1:
This is caused by @pstaapb adding the `set_id` to the data that is filled in when answers are submitted on line 209 of `webwork2/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js`.  Since that was added `renderProblem` looks up the set and problem from the database instead of using the data from the passed in parameters.

pstaabp:
Fixed in #1939

PR #1939, opened 2023-04-12
Make the renderProblem method not look up invalid problem numbers.

drgrice1:
When renderProblem is called with a problem number of 0, then don't look the problem up in the database, and instead use the passed parameters for the problem seed.

This fixes issue #1938.

This also adds a missing return statement to formatRenderedProblem when the raw output mode is used.  The lack of this return statement causes the method to continue and attempt to also render the default output template.  That causes an exception because a response has already been rendered.


somiaj (approved):
Confirmed this fixes issues #1938.

PR #1940, opened 2023-04-13
Revised handling of PERSISTENCE_HASH data

taniwallach:
Persistent problem data - directly into a new database field.
Paired with https://github.com/openwebwork/pg/pull/809

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -24,6 +24,7 @@ deal with versioning sets
 =cut
 
 use Mojo::Promise;
+use Mojo::JSON qw(encode_json decode_json);
 
 use WeBWorK::PG::ImageGenerator;
 # Use the ContentGenerator formatDateTime, not the version in Utils.
@@ -315,6 +316,7 @@ async sub pre_header_initialize ($c) {
 	my $setID           = $c->stash('setID');
 	my $userID          = $c->param('user');
 	my $effectiveUserID = $c->param('effectiveUser');
+	my $isFakeSet       = 0;
 
 	# User checks
 	my $user = $db->getUser($userID);
@@ -342,6 +344,7 @@ async sub pre_header_initialize ($c) {
 	# If the set comes in as "Undefined_Set", then we're trying/editing a single problem in a set, and so create a fake
 	# set with which to work if the user has the authorization to do that.
 	if ($setID eq 'Undefined_Set') {
+		$isFakeSet = 1;
 		# Make sure these are defined
 		$requestedVersion = 1;
 		$c->{assignment_type} = 'gateway';
@@ -910,6 +913,7 @@ async sub pre_header_initialize ($c) {
 	if ($c->{submitAnswers} || (($c->{previewAnswers} || $c->param('newPage')) && $can{recordAnswers})) {
 		# If answers are being submitted, then save the problems to the database.  If this is a preview or pages change
 		# and answers can be recorded, then save the last answer for future reference.
+		# Also save the persistent data to the database even when the last answer is not saved.
 
 		# First, deal with answers being submitted for a proctored exam.  Delete the proctor keys that authorized the
 		# grading, so that it isn't possible to log in and take another proctored test without being reauthorized.
@@ -951,6 +955,28 @@ async sub pre_header_initialize ($c) {
 				my ($past_answers_string, $scores);    # Not used here
 				($past_answers_string, $encoded_last_answer_string, $scores, $answer_types_string) =
 					create_ans_str_from_responses($c, $pg_result);
+
+				# Transfer persistent problem data from the PERSISTENCE_HASH:
+				# - Get keys to update first, to avoid extra work when no updated ar
+				#   are needed. When none, we avoid the need to decode/encode JSON,
+				#   to save the pureProblem when it would not otherwise be saved.
+				# - We are assuming that there is no need to DELETE old
+				#   persistent data if the hash is empty, even if in
+				#   potential there may be some data already in the database.
+				my @persistent_data_keys = keys %{ $pg_result->{PERSISTENCE_HASH_UPDATED} };
+				if (@persistent_data_keys) {
+					my $json_data = decode_json($pureProblem->{problem_data} || '{}');
+					for my $key (@persistent_data_keys) {
+						$json_data->{$key} = $pg_result->{PERSISTENCE_HASH}{$key};
+					}
+					$pureProblem->problem_data(encode_json($json_data));
+
+					# If the pureProblem will not be saved below, we should save the
+					# persistent data here before any other changes are made to it.
+					if (($c->{submitAnswers} && !$will{recordAnswers})) {
+						$c->db->putProblemVersion($pureProblem);
+					}
+				}
 			} else {
 				my $prefix         = sprintf('Q%04d_', $problemNumbers[$i]);
 				my @fields         = sort grep {/^(?!previous).*$prefix/} (keys %{ $c->{formFields} });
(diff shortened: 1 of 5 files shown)


taniwallach:
Note: The saving of data to the database is done right after a problem is rendered by a `Problem.pm` and `GatewayQuiz.pm` so that data set by an "initial" render gets saved. If a problem author does not want revised data to be saved on a Preview, the data in the `PERSISTENCE_HASH` should not be modified when processing a Preview.

The functionality provided here is used by the revised code of https://github.com/openwebwork/pg/pull/506 .

Future use by a revised version of https://github.com/openwebwork/pg/pull/482 is likely.

lib/WeBWorK/ContentGenerator/Problem.pm
```diff
@@ -625,6 +627,18 @@ async sub pre_header_initialize ($c) {
...
+
+	my $json_data = decode_json($pureProblem->{problem_data} || '{}');
+	for my $key (keys %{ $pg->{PERSISTENCE_HASH} }) {
+		$json_data->{$key} = $pg->{PERSISTENCE_HASH}{$key};
+	}
+	$pureProblem->problem_data(encode_json($json_data));
+	if (!$db->putUserProblem($pureProblem)) {
+		warn "failed to save problem data";
+	}
+
```
drgrice1, line 641: It doesn't seem like this is the right place to do this.  It seems like this should be done in the `process_and_log_answer` method of `lib/WeBWorK/Utils/ProblemProcessing.pm`.  If done there when the other data from the problem is processed, then you wouldn't need the extra database access that is added here.  Furthermore, I don't think this should be done when answers are not being submitted, and this doesn't take any of that into account.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -1455,6 +1456,18 @@ async sub getProblemHTML ($c, $effectiveUser, $set, $formFields, $mergedProblem)
...
+		$c->db->getProblemVersion($mergedProblem->user_id, $setID, $setVersionNumber, $mergedProblem->problem_id);
+	my $json_data = decode_json($pureProblem->{problem_data} || '{}');
+	for my $key (keys %{ $pg->{PERSISTENCE_HASH} }) {
+		$json_data->{$key} = $pg->{PERSISTENCE_HASH}{$key};
+	}
+	$pureProblem->problem_data(encode_json($json_data));
+	if (!$c->db->putProblemVersion($pureProblem)) {
+		warn "failed to save problem data";
+	}
+
```
drgrice1, line 1470: This also does not seem like it is done in the correct place, and does not take into account if answers are being submitted.  The correct place for this should be from line 968 to 1019 (which is the GatewayQuiz.pm equivalent of the `process_and_log_answers` method.

lib/WeBWorK/Utils/Rendering.pm
```diff
@@ -260,6 +263,10 @@ sub renderPG ($c, $effectiveUser, $set, $problem, $psvn, $formFields, $translati
 				map { $_ => $pg->{pgcore}{PG_alias}{resource_list}{$_}{uri}{content} }
 					keys %{ $pg->{pgcore}{PG_alias}{resource_list} }
 			};
+			$ret->{PERSISTENCE_HASH} = {
+				map { $_ => $pg->{pgcore}{PERSISTENCE_HASH}{$_} }
+					keys %{ $pg->{pgcore}{PERSISTENCE_HASH} }
+			};
```
drgrice1, line 269: There doesn't seem to be any reason to copy the hash in this way, or even to copy the hash at all.  Why not just pass back a reference to the original hash?  So then this would be
```suggestion
			$ret->{PERSISTENCE_HASH} = $pg->{pgcore}{PERSISTENCE_HASH};
```
The special handling for the hashes above is done because the original reference can't be passed back from the sub process verbatim.  They contain blessed objects that need special care.  This hash doesn't have any of that, and even if it did, copying the keys in the way that you have doesn't deal with anything like that.

drgrice1 (commented):
I think this should be rethought a bit.  Is there any real reason that the processing needs to be done where you have done it?

lib/WeBWorK/ContentGenerator/Problem.pm
taniwallach: I did consider saving only in the `process_and_log_answer` method. However, that only gets called when there is a "for credit" submission whose answers are being recorded.

At least for the use to store the state of the scaffold, the intention was that the state should be saved even on the original render and on a "check answers" call and not only for "submit answers". That keeps the Preview behavior with the new options (not opening more sections on preview and not closing open sections on preview) provided in https://github.com/openwebwork/pg/pull/506 consistent when forward progress is made using either "check answers" (after the deadline) or "submit answers".

If the save was done only in the `process_and_log_answer` method, then a student working a scaffold problem (which was set not to open additional sections on preview) after the deadline would have preview behaving either based on the last value saved using a "submit answers" regardless of what answers had been processed "recently" using "check answers" or as if only the first section had been open before the preview (based on a default which would need to be used set if the save was moved to occur only when a score/answer could be saved).

That is why I located the save where I put it. The same rationale applies to the location of the save in  `GatewayQuiz.pm`.

The problem or macro author using `PERSISTENCE_HASH` stored data gets to decide if and when to store data in the hash which will be added to or update the recorded data in the database. It becomes the author's responsibility to avoid changes which are not wanted (ex. on a preview/check answers) if that is what is needed, but the option to save data is always available.

lib/WeBWorK/Utils/Rendering.pm
taniwallach: Point accepted. I'll test the proposed change and update the PR after I test it.

lib/WeBWorK/ContentGenerator/Problem.pm
drgrice1: It is not true that `process_and_log_answer` is only called when there is a "for credit" submission.  In fact, it is always called (at least assuming a valid set and problem).  So your argument is not valid.

The code block that I referenced (circa line 967 of GatewayQuiz.pm) is also called in all cases of interest.

So this could certainly be implemented in those places, and avoid extra database access.  You may need to `put` the problem in a few more cases, but the need to get the "pure" problem another time is at least removed.

taniwallach: Sorry, you are correct. Most of the DB saving in `process_and_log_answer` is done only when `$submitAnswers` is true, which is what I meant. The code was moved to that file for "regular problems".
  - When `$submitAnswers` is false the `$db->putUserProblem($pureProblem)` is now triggered near the top of the method after the persistent data is updated (if necessary).

Also:
  - The `$db->putUserProblem($pureProblem)` call for sticky answers when `$will{recordAnswers}` is false was moved to the relevant `else` block, to avoid making that call twice when it is true.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
taniwallach: Moved, but an additional block with small differences was added to handle saving when processing a case where that location would not be hit (initial problem renders, etc.).

drgrice1 (commented):
This looks good now.  I need to do better testing of it yet though.  I think the best way to do that will be by testing https://github.com/openwebwork/pg/pull/506.

drgrice1 (approved):
I did some testing with https://github.com/openwebwork/pg/pull/506, and I think this is good.

taniwallach:
I just ran into a problem with this... I tried to view a problem (separate window) from the assignment editor for a question in the "main" assignment of a quiz, and got:
```
Can't call method "problem_data" on unblessed reference at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm line 1220.
```
This does not happen when using the link from a specific version of the quiz.
I'm looking into how to prevent this issue.

PR #1942, opened 2023-04-16
Replace bin/dev_scripts/run-perltidy with bin/dev_scripts/run-perltidy.pl

drgrice1:
The primary advantage of this perl script over the previous shell script is that it used File::Find instead of relying on the extglob, globstar, and nullglob shell options.  Those options are not supported by all shells.

If this script is run with no arguments, then it does the same thing that the previous shell script did, and runs perltidy on all .pm, .pl, and .t files in the WEBWORK_ROOT directory.

If file names are passed on the command line, then perltidy is only executed on the passed filenames.  Another advantage of this script over the previous shell script.

The script enforces the use of the .perltidyrc file in the WEBWORK_ROOT directory.  It also makes the default for -bext to be '/', which means that backup files are created with the .bak extension, and deleted if no errors occur.  That can be overridden if backups are desired.  Other that those things, all options accepted by perltidy are also accepted by this script.

One other thing that this script does is enforces that the version of perltidy used is version 20220613 or newer.  Older versions have an issue with the -vxl='q' option.

### DIFF
--- a/bin/dev_scripts/run-perltidy
+++ b/bin/dev_scripts/run-perltidy
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null 2>&1 && pwd)"
-INPUTDIR=$(dirname $(dirname $SCRIPT_DIR))
-
-shopt -s extglob globstar nullglob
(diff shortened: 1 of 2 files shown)


pstaabp:
Also looks good.  I would also include a verbose flag like suggested in openwebwork/pg#812

pstaabp (approved):

drgrice1:
The similar PG pull request was merged.  This should be merged also.

PR #1943, opened 2023-04-17
PGProblemEditor bugfix and add hot key to render problem

somiaj:
Added `Ctrl-Enter` as a hot key to re-render the problem from inside the editor's text area.

Also found and fixed a bug that re-rendering the problem when codeMirror was disabled would cause erase the temp file with a blank file.

### DIFF
--- a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
@@ -149,6 +149,15 @@
 	// CodeMirror is disabled in localOverrides.conf.
 	const editorArea = document.querySelector('.CodeMirror') ?? document.getElementById('problemContents');
 
+	// Add hot key, ctrl-enter, to render the problem
+	editorArea.addEventListener('keydown', async (e) => {
+		if (e.ctrlKey && e.code === 'Enter') {
+			e.preventDefault();
+			await render();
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks okay.  I am not sure that a hotkey is that useful in this situation.  Furthermore, I think that it is something that most users will not even know exists.  It doesn't hurt to add it though.

somiaj:
I do agree that most users won't know it is there. I did think about adding this to the help page or some documentation, but unsure if that would help things out (or were to best add this info). I did set the key to be what I'm use to using in jupyter notebooks, so maybe others may just try that.

pstaabp (approved):

pstaabp:
Where should we add some help for this?  On the PGproblemEditor page? or in the wiki?

There's nothing in the Help about this page in the "Help" on the navigation.

drgrice1:
There is a help page for the PG Problem Editor.  You can access it by clicking on the question mark icon in the site nav.

drgrice1:
@somiaj:  I documented the hot key in the help file in pull request #1954.

drgrice1:
This has two approvals.  So I will merge it.

drgrice1:
Note that this was merged into both develop and the WeBWorK-2.18 release candidate branch.

PR #1945, opened 2023-04-20
remove still_not_right achievement

Alex-Jordan:
This may be unnecessary with other changes happening to 2.18, and in that case let this PR serve as a discussion board.

But I still have problems in 2.17 with the "Still Not Right" achievement when MathQuill is in use. Somewhere something in the database grows larger than is permitted, I presume keeping track of many past answer attempts. Each one extra long because of the extra MathQuill structure.

If it's no longer a problem in 2.18, great. Or maybe it's better just to remove it from the default list, and leave the achievement itself in place for now. But locally I will be trying to eradicate this from people's old courses that they clone from since it causes so much trouble (assuming that it still causes trouble in 2.18).


drgrice1:
This is still an issue with 2.18.  The problem is simply trying to store to much data in the frozen_hash column of the UserAchievement table.  The frozen_hash column is of type VARCHAR(1024) and this achievement tries to store all of the last_answer column from the UserProblem table in that.  The last_answer column is of type TEXT which can be much larger than 1024 bytes.  If MathQuill is enabled and a problem has multiple parts each storing the latex version of the answer as well in the last_answer column, or if a problem uses a Geogebra applet that stores state in the last_answer column, that can easily exceed the 1024 byte limit of the frozen_hash column.  The frozen_hash column should never have been of type VARCHAR and it should have been bigger.  It should have been of type BLOB or even MEDIUMBLOB.  VARCHAR is for textual data, and the frozen hash is binary data (passed through nfreeze and base64 encoded) which is what the BLOB type is for.

Alex-Jordan:
Discussed in a meeting: we'll make this field larger in the database instead of removing this achievement.

Alex-Jordan:
Do we close this now that frozen hashes were addressed?

drgrice1:
I suppose.  Although, as we discussed, what is the point of this achievement?

Alex-Jordan:
> what is the point of this achievement?

Yeah the 9 should change to 3, I think.

drgrice1:
Yeah, making the achievement occur at an earlier point than 3 is probably the way to go.

PR #1946, opened 2023-04-20
WeBWorK 2.18

drgrice1:
This is the release candidate.  Pull requests that are desired to be in the release should be target on this branch.


Alex-Jordan:
If I'd like for somebody else's PR to be targeted to this branch, should I just move the target myself? Or ask that person?

drgrice1:
You can move it.

Alex-Jordan:
At what point do we change the VERSION file to say 2.18?

drgrice1:
We just have to merge #1947 for that.  We can do that anytime now.  I have been holding off as it makes it easier to keep develop and the release candidate branches in sync.  I think it is time now though.

PR #1947, opened 2023-04-20
Update the copyright in file headers and webwork2 VERSION.

drgrice1:
This is the result of executing the bin/dev_scripts/update-copyright script.

In addition the copyright in some of the javascript files was manually updated as the script didn't catch those.

The version in the VERSION file is also updated.


taniwallach (approved):
I would say that it makes sense to go ahead and merge this.
If some new file gets added with an outdated copyright range - it can get fixed in the future.

drgrice1:
The reason I made this a draft was to delay the merge.  The thing is that it is easier to keep the develop and release candidate in sync before they diverge.  This pull request shouldn't go into develop until the after the actual release, and so it diverges the two.  They can still be kept in sync, it just takes a bit more work.  So I guess I will go ahead and make it ready for review.

PR #1948, opened 2023-04-20
Render empty problems via the render_rpc and deprecated html2xml endpoints.

drgrice1:
Instead of failing to render when the body text is the empty string, just render the empty problem.  Only fail if the error_flag is set.

This means that a problem like
```perlcritic
DOCUMENT();
loadMacros('PGstandard.pl');
ENDDOCUMENT();
```
will now render, but will have not content.

Note that this is consistent with how problems are rendered in an actual set.

The message about not being able to "decode" the problem text was removed because that is not valid.  There is no decoding of the problem text anymore.  Also the ugly `format_hash_ref` call and the method itself was removed as that is no longer serving its purpose.

Also remove the "nicer file not found error".  In that case the error_flag is set, so just let that come through.  This happens for files that actually have no content as well as for files that don't exist.  The error message was changed for the general problem renderer used elsewhere to better say what actually causes the message.

### DIFF
--- a/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/apps/PGProblemEditor/pgproblemeditor.js
@@ -300,8 +300,6 @@
 				if (data.error) throw data.error;
 				// This generally shouldn't happen.
 				if (!data.html) throw 'A server error occured.  The response had no content';
-				// Give a nicer file not found error.
-				if (/this problem file was empty/i.test(data.html)) throw 'No Such File or Directory!';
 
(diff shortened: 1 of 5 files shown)


pstaabp:
If using the above problem, you'll need some semicolons:
```
DOCUMENT();
loadMacros('PGstandard.pl');
ENDDOCUMENT();
```

pstaabp (approved):

drgrice1:
Thanks.  I added that to the original post.

taniwallach (approved):

PR #1950, opened 2023-04-21
Rewrite third party asset handling.

drgrice1:
This removes the third-party-assets.json file, and the contents of that file are no longer added to the static-assets.json file.  Instead the lib/WeBWorK/Utils.pm getAssetURL method loads the package.json file and uses that to determine how to serve any file located in node_modules (both from webwork2/htdocs and pg/htdocs).

The option thirdPartyAssetsUseCDN is added in site.conf.  If that is set to 1, then getAssetURL constructs the jsdelivr CDN URL.  If that option is set to 0, then the file is served from the node_modules directories, and furthermore the version is appended as a URL parameter to the node_modules path.

This is paired with the PG pull request https://github.com/openwebwork/pg/pull/817.  Although that pull request is not actually needed to test this.


drgrice1:
Thanks for pointing out the typo.  That was there before.

lib/WeBWorK/Utils.pm
```diff
@@ -1940,50 +1940,102 @@ sub generateURLs {
...
+	}
+
+	# Check to see if this is a third party asset file in node_modules (either in webwork2/htdocs or pg/htdocs).
+	# If so, then either serve it from a CDN if requested, or serve it directly with the library version
+	# appended as a URL parameter.
+	if ($file =~ /^node_modules/) {
+		my $wwFile = getThirdPartyAssetURL(
+			$file, $thirdPartyWWDependencies,
+			$ce->{webworkURLs}{htdocs},
+			$ce->{options}{thirdParyAssetsUseCDN}
```
pstaabp, line 2024: thirdParyAssetsUseCDN -> thirdPartyAssetsUseCDN

lib/WeBWorK/Utils.pm
```diff
@@ -1940,50 +1940,102 @@ sub generateURLs {
...
+		);
+		return $wwFile if $wwFile;
+
+		my $pgFile =
+			getThirdPartyAssetURL($file, $thirdPartyPGDependencies, '/pg_files', $ce->{options}{thirdParyAssetsUseCDN});
```
pstaabp, line 2029: thirdParyAssetsUseCDN -> thirdPartyAssetsUseCDN

pstaabp:
I couldn't get this to work with the CDN.  I tried to update the typos, but still didn't see things pulling from a CDN.  To clarify, if I'm on the PGEditor page, then the codemirror js code should be pulling from a CDN instead of `/webwork_files/node_modules/`, correct?

lib/WeBWorK/Utils.pm
drgrice1: These are fixed.

drgrice1:
Even with the typos it would work (since the typos were consistent!).  Yes, you should not see the node_modules except for MathQuill when the option is set to true.

drgrice1:
You will need to run `npm install` (or just `generate-assets.js`) for both webwork and pg.

pstaabp:
> You will need to run `npm install` (or just `generate-assets.js`) for both webwork and pg.

I had done that, but needed to set the thirdPartyAssetsUseCDN as:
```
$options{thirdPartyAssetsUseCDN} = 1;
```
I had forgotten the `$options` part.   Works now.

pstaabp (approved):

PR #1952, opened 2023-04-22
Losing the z-index game.

drgrice1:
More tweaks to z-indexing.  The achievements toasts, gateway alert toasts and remaining time, and instructor email insertable macros toast all need to be above the masthead.


drgrice1:
The funny thing is that I also specifically tested the things you listed, and also skipped testing the achievement toasts, assuming it would work.  So I decided I better test it.  I did so now, and it looks good!

pstaabp (approved):
👍

PR #1953, opened 2023-04-24
Prevent multiple auto submits of gateway tests.

drgrice1:
When the gateway test option "Number of Graded Submissions per Test" is set something more than one or even worse to 0, then the javascript auto submit submits continuously until all submissions are used up or the full grace period expires.  That can crash a server.

To test this create a gateway quiz and just use the default setting of 0 for that option.  I recommend setting the test time limit to something small (I used 0.05 for 3 seconds), and setting `$gatewayGracePeriod = 11;` in localOverrides.conf.  When the time is up on the develop branch you will see the quiz start submitting, and continue to do so for the next 10 seconds and the full gatewayGracePeriod is expired.

With this pull request, you will see the test only submit once.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8228.

### DIFF
--- a/htdocs/js/apps/GatewayQuiz/gateway.js
+++ b/htdocs/js/apps/GatewayQuiz/gateway.js
@@ -66,6 +66,7 @@
 			const alertStatus = sessionStorage.getItem('gatewayAlertStatus');
 
 			if (remainingTime <= 10 - gracePeriod) {
+				if (!alertStatus) return;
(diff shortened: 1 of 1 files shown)


drgrice1:
I would be okay with that.  Do others agree?  It is generally not something that probably happens much, as I don't think that students typically leave the browser open with a test un-submitted for two minutes after time has run out.  However, it can be rather bad when that does happen.

PR #1956, opened 2023-04-25
Switch to using templates for the webwork2 helpfiles.

drgrice1:
This is an alternate approach to #1954 (or just takes things further perhaps).  It is actually built on top of that pull request.

This converts the html help files into Mojolicious templates.  Furthermore, with a common layout for those templates a dialog with the help content is embedded into the page instead of injecting it with javascript and fetching the html later.  One advantage of this is being able to make the title of the dialog specific to the page instead of using the generic "WeBWorK Help" title for the dialog with the more specific title at the top of the help in the content.

However, the primary advantage of this approach is that the help files are fully translatable via the usual gettext/maketext setup.


pstaabp:
This looks nice and I like it in a bootstrap dialog instead of separate browser window.

It doesn't look like the Student Progress page has a help page though.

drgrice1:
@pstaabp see https://github.com/openwebwork/webwork2/pull/1954#issuecomment-1522431385

pstaabp:
Maybe a revamp of these pages can be done before the final release is made.

pstaabp (approved):
Much improved.

drgrice1:
I now moved the page specific help icon from the site navigation (the main menu) and up by the page title.

pstaabp:
I think it's clearer with the ? on the page header.  As we discussed, we should merge in these new template system and then fix the individual pages.

PR #1959, opened 2023-04-27
Add the equation editor to the manual problem grader instructor comment boxes.

drgrice1:
This just adds the latexentryfield class and a hidden MathQuill input corresponding to the instructor comment box.  Thus if MathQuill is enabled, then the mqeditor activates for the textarea, and if MathView is enabled then the mathviewer activates for the textarea.

For this to work well a few changes to the MathQuill and Mathview equation editors are needed.  Those are in a PG pull request (https://github.com/openwebwork/pg/pull/822).

### DIFF
--- a/htdocs/js/apps/ProblemGrader/problemgrader.js
+++ b/htdocs/js/apps/ProblemGrader/problemgrader.js
@@ -1,26 +1,6 @@
 'use strict';
 
 (() => {
-	// Comment preview popovers.
-	document.querySelectorAll('.preview').forEach((el) => {
-		el.addEventListener('click', () => {
-			el.dataset.bsContent =
-				el.parentNode.querySelector('textarea')?.value.replace(/</g, '< ').replace(/>/g, ' >');
-			if (el.dataset.bsContent) {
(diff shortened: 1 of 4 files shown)


Alex-Jordan:
I tested this and it works. However I notice that when I want to save whatever comment I am leaving for the student, it looks like I have to scroll to the bottom of the page to find the Save button. (The same is true for simply changing a student score on an exercise.) If it's a long list of students that is either inconvenient or worse, the instructor doesn't think to check down there.

It's a separate issue from this PR, but I would like to brainstorm ideas for making this better. Maybe.

drgrice1:
That is the design of that page.  It requires a form submission to save any changes.  It is not like the single problem grader that uses the api to save changes.

Note that this isn't a new page.  Its existence is certainly made more prevalent since it is now always available though.  Its design has been improved considerably, but it could probably still use some work.   Perhaps each row could have a save button that uses the api to save the changes for the row.  Although finding space for that would be a challenge with the already tight layout.

Alex-Jordan (approved):

PR #1961, opened 2023-04-27
Change the authentication failure message when accessing the webservice endpoints.

drgrice1:
Try to make the message a bit clearer as to what is happening when the session times out.  Note the endpoints can not distinguish in any valid way a session timeout from a failed authentication attempt.  So only one message can be sent in both cases.

Is this wording better?  I can't remember the exact wording we discussed.

Note that these messages are now translatable and the "render_rpc: " prefix of the message has been dropped.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/InstructorRPCHandler.pm
+++ b/lib/WeBWorK/ContentGenerator/InstructorRPCHandler.pm
@@ -41,7 +41,7 @@ use WebworkWebservice;
 
 async sub pre_header_initialize ($c) {
 	unless ($c->authen->was_verified) {
-		$c->{output} = 'instructor_rpc: authentication failed.';
+		$c->{output} = $c->maketext('Authentication failed. Log in again to continue.');
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


drdrew42 (approved):
I think this wording is clear enough.

somiaj (approved):
Wording looks good to me. Any reason these strings aren't translated with maketext?

drgrice1:
@somiaj:  These endpoints do not necessarily have what is needed to construct a valid course environment and thus the localization context for which to call `maketext`.

drgrice1:
I guess there actually is enough to get the site wide locale to work with.  Historically, when this was the html2xml endpoint and that actually went through the XML processing that was true.  This authentication message could be translated now.  Perhaps the `render_rpc: ` prefix should be removed as well.  I don't think it is worth translating the other messages in the InstructorRPCHandler.pm file though, because those messages won't occur for the webwork2 ui.  They will only occur if someone is making their own request through other means, and not using the endpoint correctly.  Webwork2 doesn't do that (unless one of us makes a mistake!).

pstaabp (approved):

PR #1962, opened 2023-04-27
Fix issues with tests that have questions with a weight of 0

dlglin:
In gateway tests if a question has weight of 0 then the score is reported incorrectly on both the quiz page and the scoring download.

To test:

- Create a test with two questions, and set the weight of the second question to 0
- Take the test and get both questions correct.  Before this PR the message will say "Your score on this attempt is 2/1".  With the change it will correctly report "Your score on this attempt is 1/1".
- From the scoring tools page download the scores for the above test.  Before the PR your score in the csv will be listed as "2".  With the patch the score will be listed as "1".

Can Scoring.pm be refactored to use `grade_set` from Utils.pm?  It seems like there is a lot of duplicated code here.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1316,7 +1316,7 @@ async sub pre_header_initialize ($c) {
 	if ($will{recordAnswers} || $will{checkAnswers}) {
 		my $i = 0;
 		for my $pg (@pg_results) {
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
Looks good.  Tests out as you say.

The thing about refactoring Scoring.pm to use `grade_set` from Utils.pm is that the scoring tools break the grades up in different ways than anywhere else.  I looked into it a little bit when all other locations were refactored to use `grade_set`.  It maybe could be done, but would take more work than elsewhere, and the `grade_set` method would probably need to return more details.

pstaabp (approved):
👍

Issue #1960, opened 2023-04-27
LibraryBrowser in develop


pstaabp:
Adding problems in the Library Browser on the develop branch results in

<img width="349" alt="image" src="https://user-images.githubusercontent.com/879338/234935596-1b0437ef-13a0-48b0-991e-45219c52836d.png">

This might have occurred with #1846.

Although there is an error, it appears that the problems are added to the set.

drgrice1:
Clearly there is a problem here.  The message is already indicative of something not done right.  It is saying "some problems in the set are already assigned" which should not be a problem if you are adding more problems to the set.  @taniwallach can you take a look at this?

drdrew42:
Resolved by #1964

PR #1964, opened 2023-04-28
Bugfix: Errors when adding problems to set with assigned users

drdrew42:
### Description

An error is thrown, and database corruption occurs, whenever problems are added to a set (with assigned users) from the Library Browser. This is caused by counting too many records from the problem_user table, querying for any assigned problems rather than the specific problem(s) being added to the set.

### Resolution

Restrict the database query to match set, user, *and problem*

Suggestions for updating the error message are welcome, but this is a situation that "should" never happen...

### Testing

Create a set with at least one problem, assign it to at least one user, then visit the library browser and try to add a problem to the set. See #1960 for details. Warning: this causes database issues!

After applying this fix, problems will be added to the set and they will add a UserProblem for each assigned user, as expected.

### DIFF
--- a/lib/WeBWorK/DB.pm
+++ b/lib/WeBWorK/DB.pm
@@ -2148,15 +2148,17 @@ sub addUserMultipleProblems {
 	my @collectedProblemList;
 
 	for my $problemListRef (@problemLists) {
-		my $user_id = $problemListRef->[0]->user_id;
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Nice.  That fixes the problem.  Thanks for taking care of that.

pstaabp (approved):
Great.  Fixed!!

PR #1966, opened 2023-04-29
Fix theme and key map loading in the PG problem editor.

drgrice1:
Adding the version string to the static assets messed up the dynamic javascript and css loading done by the PG problem editor javascript.

As a result, with the current develop/WeBWorK-2.18 code you can't change themes.  This fixes that.


somiaj (approved):
Confirmed the bug and this fixes it, allowing one to be able to change themes again.

pstaabp (approved):

PR #1971, opened 2023-05-05
There is another frozen_hash!

drgrice1:
Change the other frozen_hash column in the UserAchievement table to MEDIUMBLOB.


drgrice1:
In fact this is the one that has been causing the problems.  This is where the `still_not_right` achievement actually stores its data including all of the last_answer column.

pstaabp:
This makes sense that data about user achievements would be in this file.

taniwallach (approved):

pstaabp (approved):

PR #1972, opened 2023-05-06
Fix some issues with Achievements.pm

somiaj:
This includes two fixes.

The first is that if a student has global achievement data but is not assigned to any visible achievement badges, the page errors out due to calling a method on an undefined achievement. Here is the error:

```
Error messages

Can't call method "category" on an undefined value at template
ContentGenerator/Achievements/achievement_badges.html.ep line 3.

Context

    1: <h2 class="my-3"><%= maketext('Badges') %></h2>

    2: %

    3: % my $previousCategory = $achievements->[0]->category;

    4: %

    5: % if (@$achievements) {

    6: 	% for my $achievement (@$achievements) {
```

This fixes the issue by moving the definition of `$previousCategory` inside the if block which first checks that visible achievements exist.

The second is the achievement progress bars height of the inner div can be one pixel too big or too small (probably due to rounding issues) when viewing the achievements on certain resolutions/zoom sizes.

This fixes the issue by making the height of the inner bars all `100%` instead of a set pixel size, which ensures the height always matches the size of the outside div. Though this fix also makes the `.levelinnerbar` and `.cheevoinnerbar` classes the same (should these be combined into a single class or left as is?).

### DIFF
--- a/htdocs/themes/math4/achievements.scss
+++ b/htdocs/themes/math4/achievements.scss
@@ -20,7 +20,7 @@
 }
 
 .levelinnerbar {
-	height: 16px;
+	height: 100%;
 	background-color: var(--ww-achievement-level-color, #88D);
 }
 
@@ -46,7 +46,7 @@
 }
 
 .cheevoinnerbar {
-	height: 6px;
+	height: 100%;
 	background-color: var(--ww-achievement-level-color, #88D);
 }
 
--- a/templates/ContentGenerator/Achievements/achievement_badges.html.ep
+++ b/templates/ContentGenerator/Achievements/achievement_badges.html.ep
@@ -1,8 +1,7 @@
 <h2 class="my-3"><%= maketext('Badges') %></h2>
 %
-% my $previousCategory = $achievements->[0]->category;
-%
 % if (@$achievements) {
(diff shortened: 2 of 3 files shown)


somiaj:
To reproduce the bug, turn achievements on in a course, but don't import or assign any achievements to any students. Then answer a homework problem. Completing problems with achievements enabled will generate global achievement data, but there will be no badges available and the page will error out. (This will also happen if a user is only assigned level and/or hidden badge achievements, so no visible badges are shown).

drgrice1 (approved):
Looks good.  It is always better to use a generic dimension in css than a fixed pixel measure, and accessing the first element of the `$achievements` array would clearly be a problem if that element doesn't exists.

somiaj:
I noticed that the achievement editor was using `gateways` instead of `tests` when select which type of sets to apply an achievement to. This was missed in renaming of gateways in the UI. Put the patch here since this is another small fix to the achievement pages (let me know if you prefer I move this to its own pull request).

drgrice1:
It is fine to add it to this pull request.

pstaabp (approved):

drgrice1:
This can be merged!

PR #1976, opened 2023-05-10
Fetch feedback mail recipients only when needed.

drgrice1:
This moves the call to fetchEmailRecipients out of Problem.pm and into the feedbackMacro method, and only in the case the the feedback form is used.  That is the only case that needs the list, and that expensive call makes changing problems very slow for large classes.

This fixes the issue observed in
https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8214.  Note that this regression was introduced in WeBWorK 2.16.

Also make the Feedback.pm module use the Utils.pm fetchEmailRecipients method, and improve that method to add in the course environemnt $mail{feedbackRecipients} values as the previous Feedback.pm getFeedbackRecipients method did.

There is a larger problem here though.  The method that we use to determine feedback recipients requires retrieving all users from the database.  That is an expensive query, and should not be done for something as simple as determining email feedback recipients.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -994,6 +994,8 @@ sub feedbackMacro ($c, %params) {
 		return $c->link_to(($c->maketext($c->ce->{feedback_button_name}) || $c->maketext('Email instructor')) =>
 				$c->ce->{courseURLs}{feedbackURL});
 	} elsif ($c->ce->{courseURLs}{feedbackFormURL}) {
+		$params{notifyAddresses} =
+			join(';', $c->fetchEmailRecipients('receive_feedback', $c->db->getUser($c->param('user'))));
 		return $c->include(
 			'ContentGenerator/Base/feedback_macro_form',
 			params          => \%params,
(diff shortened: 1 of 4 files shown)


drgrice1:
This now does as we discussed today.  Instead of listing all users and then checking permissions, this gets users at or above the desired permission level from the permissions table, and then only fetches those users from the database.  This is much faster.

I left the move of the database query (in the fetchEmailRecipients call) from Problem.pm to the feedbackMacro method of ContentGenerator.pm because that query is not needed except in the case that a feedbackFormURL is used.

Alex-Jordan:
I'm not sure how best to test. I am still able to send an email. I don't know if I want to try to simulate the original slowdown issue... I suppose I could add a ton of fake users to a course on this dev server.

drgrice1:
Yeah, the only way to really test this is to have a course with say 10,000 or more users.  Of course, testing to see that things work as before with a normal size class is good too.

somiaj (approved):
Tested by sending a feedback email and it correctly found my inbox.

drdrew42 (approved):

PR #1967, opened 2023-04-29
Update instructor help files (team effort?)

drgrice1:
This is hopefully a team effort pull request to update these help files.  If you are a maintainer, then you can push directly to this branch.  Please do not force push, and remember to pull frequently to get changes others may have pushed since you last worked on it.  If you are not a maintainer, then you will need to put in a pull request to this branch to get your changes in.

Of course, feel free to discuss here.

Bare bones help files have been added for pages that do not already have help files.  These should be improved to actually say something useful. Existing help files should be reviewed and improved as well.
    
When updating these help files remember two things.  These are not shown to students, and these are only available when the page that the help is for is already visible.  The second things means that these help files should not just list the content of the page that is already visible for the user.  That is pointless.  Instead the help files should describe more detailed usage of the page, and most importantly, things that are not intuitive from the layout of the page to begin with.

The following is a list of the existing help files.  

If you are working on a particular help file, put your initials by the file to make note of your work and prevent conflict.

- [x] Achievements.html.ep
- [x] Grades.html.ep (JS)
- [x] Hardcopy.html.ep (PS)
- [x] InstructorAchievementEditor.html.ep (JS)
- [x] InstructorAchievementList.html.ep (PS and JS)
- [x] InstructorAchievementUserEditor.html.ep (JS)
- [x] InstructorAddUsers.html.ep (PS)
- [x] InstructorAssigner.html.ep (PS)
- [x] InstructorConfig.html.ep
- [x] InstructorFileManager.html.ep (PS)
- [x] InstructorIndex.html.ep
- [x] instructor_links.html.ep
- [x] InstructorLTIUpdate.html.ep (JS)
- [x] InstructorPGProblemEditor.html.ep
- [x] InstructorProblemGrader.html.ep
- [x] InstructorProblemSetDetail.html.ep
- [x] InstructorProblemSetList.html.ep (PS)
- [x] InstructorScoring.html.ep (PS)
- [x] InstructorSendMail.html.ep (JS)
- [x] InstructorSetMaker.html.ep (PS)
- [x] InstructorShowAnswers.html.ep
- [x] InstructorStats.html.ep (JS)
- [x] InstructorStudentProgress.html.ep (JS)
- [x] InstructorUserDetail.html.ep
- [x] InstructorUserList.html.ep
- [x] InstructorUsersAssignedToSet.html.ep
- [x] Levels.html.ep
- [x] Options.html.ep
- [x] ProblemSets.html.ep

PR #1979, opened 2023-05-11
No more "Take Action!" button.  Stop taking action.  Just sleep.

drgrice1:
On the pages that utilize action tabs, instead of the button reading "Take Action!" it now has the text of the currently active tab.

Note that some of the help files reference the "Take Action!" button. That will be changed in #1967.

Also, there are "Take Action!" buttons on the location management pages in the course administration course.  Those are left for now.  Those pages do not use the action tab javascript.

### DIFF
--- a/htdocs/js/apps/ActionTabs/actiontabs.js
+++ b/htdocs/js/apps/ActionTabs/actiontabs.js
@@ -1,6 +1,12 @@
-document.querySelectorAll('.action-link').forEach((actionLink) => {
-	actionLink.addEventListener('click', () => {
-		actionLink.blur();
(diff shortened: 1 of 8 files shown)


somiaj:
Looks good. One thing that occurred to me is could there be accessibility issues now that the tab and 'formally know as take action button' have the exact same label?

drgrice1:
I don't see why that would be an accessibility issue?

Alex-Jordan (approved):
Tested and it's good. Make it so!

somiaj (approved):

drdrew42 (approved):
Take Action!

Alex-Jordan:
Three approvals, plus Glenn. Merging this now.

PR #1981, opened 2023-05-12
Add a link to the set assigner page in the site nav.

drgrice1:
Previously this page could only be accessed from the "Instructor Tools" page when using the "Assign" button, and when you get to the page the job of assigning the selected sets to the selected users is already done.  Although this page also allows mass unassigning.


drdrew42 (approved):
Wow! A whole new feature, just like that!

Alex-Jordan:
Approved. Including Glenn that's 3. I'm merging this and will update #1982.

PR #1982, opened 2023-05-12
simplify the nav menu and use bootstrap to style instructor tools

Alex-Jordan:
This puts the instructor tools (including "Instructor Tools" all in the main `ul` and puts a bootstrap class on the instructor tool items so they stand out. See what you think, I'm not in love with the look but I just don't like how "Instructor Tools" is currently separate from the other instructor tools. Instructors often think it is just a header, not a link.

If this is accepted, then #1981 would need an update to put the `list-group-item-primary` class on that item.

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -219,6 +219,12 @@ $layout-divider-color: #aaa !default;
 					color: var(--ww-primary-foreground-color, white);
 				}
 			}
+			&.list-group-item {
+				.nav-link:focus {
+					position: relative;
+					z-index: 2;
+				}
+			}
 		}
 	}
 
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -3,13 +3,13 @@
 <h2 class="navbar-brand mb-0"><%= maketext('Main Menu') %></h2>
 <ul class="nav flex-column">
 	% unless ($restricted_navigation) {
-		<li class="nav-item">
+		<li class="list-group-item nav-item">
 			<%= link_to maketext('Courses') => 'root', class => 'nav-link' %>
 		</li>
 	% }
 	% if (defined $courseID && $authen->was_verified) {
 		% # Homework Sets or Course Administration
-		<li class="nav-item">
+		<li class="list-group-item nav-item">
 			% if ($restricted_navigation) {
 				<span class="nav-link disabled"><%= maketext('Homework Sets') %></span>
 			% } else {
@@ -23,7 +23,7 @@
 		</li>
 		%
 		% if (defined $setID) {
-			<li class="nav-item">
+			<li class="list-group-item nav-item">
 				<ul class="nav flex-column">
 					% # Set link. The set record is needed to determine the assignment type.
 					% my $setRecord = $db->getGlobalSet($setID =~ s/,v\d+$//r);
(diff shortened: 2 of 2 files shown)


Alex-Jordan:
The diff looks bigger than it is. All that happens is the `li` and `ul` for the currently indented items are removed, and the class `list-group-item-primary` is added to all top-level instructor tool`li`.

drgrice1 (commented):
Mixing the `list-group` styles with the `nav` styles doesn't seem to be working quite right.  When you use tab to move focus to a link that does not have the `list-group-item-primary` style you see the border around the `nav-link`, but when move focus to a link that has the `list-group-item-primary` style you only see the top border.  This makes it unclear which item is actually focused.  Is it the one above or below the only visible border line?  The text color of the actually focused item is different, but not that different (the contrast between the focused text color and the unfocused text color is not very high).

Bootstrap does not design `nav-links`'s to work well with background colors.  The same thing happened with the tabs on the course configuration page.  To fix this I bumped up the z-index on a focused tab.  That is done on line 914 of math4.scss.  That will probably need to be done here as well.

You might also consider using `list-group-item-secondary` instead of `list-group-item-primary`.

drgrice1:
To add the `z-index` as I suggested, you could add the following after line 221 of math4.scss:
```scss
			&.list-group-item-primary {
				.nav-link:focus {
					position: relative;
					z-index: 2;
				}
			}
```

Alex-Jordan:
Better now. I put the `list-group-item` class on all items, and moved the `nav-item` class to the end of all class lists.

A problem with using `list-group-item-secondary` right now is the secondary background color is the same as the on-hover background color. See what you think now, and if we want to use `list-group-item-secondary` (with a gray background) we could change the hover background color.

drgrice1:
Yeah, I think the primary color looks good.  Stick with that.

drgrice1 (approved):
This looks good.  Maybe users will realize that the "Instructor Tools" item is a link to a page now, instead of a nesting title!

pstaabp:
When a set or problem is selected, I'm not crazy about the background being white.  I think I have the default theme on and this is what I see:

image

To me the "set 2" doesn't look like it is an instructor tool now. It seems like it would look fine with the same background as the other tools/pages--I think the indenting makes it clear that it is part of Hmwk Set Editor or Student Progress.

drgrice1:
@pstaabp:  That is not what I see.  I see the following:
main-menu

pstaabp:
@drgrice1 Yep Safari.

drgrice1:
Can you use the developer tools and see if you can figure out more on this?  Probably compare what you see in Firefox or Chrome to what is shown in Safari.  This probably has something to do with css inheritance.   In Firefox the element inspector shows that the inner `li` for the problem set is inheriting the `list-group-item-primary` styles from the outer `li`.  I suspect on Safari that inheritance is not working.

pstaabp:
It was the `#site-links ul` that is in `math4-overrides.css` which I figured out is an old file I think from 2.17.  I deleted it and looks good now.

pstaabp (approved):

pstaabp:
I approved, but since this is a bit of UI change, maybe others should weigh in before merging.

pstaabp:
I approved, but since this is a bit of UI change, maybe others should weigh in before merging.

somiaj:
I'm okay with the change. Since `Report bugs` is only visible to instructors, think this should be the primary color too? I see the color as indicating only visible to instructors.

Alex-Jordan:
Report bugs is available to anyone with the `report_bugs` permission (`ta` by default). All the things where this applies the `list-group-item-primary` class are tied to the `access_instructor_tools` permission (also `ta` by default).

I have no objection to putting the `list-group-item-primary` class on the bug reporter link. I'll do that in another commit now.

PR #1983, opened 2023-05-12
Include number of incorrect attempts for sets in student progress table.

somiaj:
This makes the problem table consistent with other uses of it on
  grades pages, and correctly displays the number which is mentioned
  in help text at the top of the table.


drgrice1 (approved):
Looks good.

PR #1984, opened 2023-05-12
Add set status message to ProblemSet and StudentProgress pages.

somiaj:
This removes the due dates from the main page title on the ProblemSet
  and StudentProgress for sets pages and instead puts that information
  into a status alert (similar to where the reduced scoring information
  was already) that states when a set will open, is due (enters reduced
  scoring period), or is closed.  If the set uses reduced scoring the
  reduced reduced scoring message is included in the alert.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSet.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSet.pm
@@ -116,14 +116,6 @@ sub nav ($c, $args) {
 	);
 }
 
-sub page_title ($c) {
-	my $ce = $c->ce;
-
-	# Using the url path parameters won't break if the set/problem are invalid.
-	my $setID = $c->stash('setID');
-	return $c->tag('span', dir => 'ltr', format_set_name_display($setID));
-}
-
 sub siblings ($c) {
 	my $db    = $c->db;
 	my $ce    = $c->ce;
--- a/lib/WeBWorK/ContentGenerator/ProblemSet.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSet.pm
@@ -121,32 +121,7 @@ sub page_title ($c) {
 
 	# Using the url path parameters won't break if the set/problem are invalid.
 	my $setID = $c->stash('setID');
-
-	my $title = $c->tag('span', dir => 'ltr', format_set_name_display($setID));
-
-	# Put either due date or reduced scoring date in the title.
-	my $set = $c->db->getMergedSet($c->param('effectiveUser'), $setID);
-	if (defined($set) && between($set->open_date, $set->due_date)) {
-		if ($ce->{pg}{ansEvalDefaults}{enableReducedScoring}
-			&& $set->enable_reduced_scoring
-			&& $set->reduced_scoring_date
-			&& $set->reduced_scoring_date != $set->due_date
-			&& before($set->reduced_scoring_date))
-		{
-			$title .= ' - '
-				. $c->maketext(
-					'Due [_1], after which reduced scoring is available until [_2]',
-					$c->formatDateTime($set->reduced_scoring_date, undef, $ce->{studentDateDisplayFormat}),
-					$c->formatDateTime($set->due_date,             undef, $ce->{studentDateDisplayFormat})
-				);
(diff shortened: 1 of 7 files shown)


templates/ContentGenerator/Instructor/StudentProgress/set_status.html.ep
```diff
@@ -0,0 +1,52 @@
+% use WeBWorK::Utils qw(after before between);
```
drgrice1, line 1: This file does not belong in templates/ContentGenerator/Instructor/StudentProgress.  It is used by both instructor and non-instructor pages.  It should be moved to templates/ContentGenerator/Base.

somiaj: Thanks. I was unsure the best place to put it.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -121,32 +121,7 @@ sub page_title ($c) {
...
-				);
-		} elsif ($set->due_date) {
-			$title .= ' - '
-				. $c->maketext('Closes [_1]',
-					$c->formatDateTime($set->due_date, undef, $ce->{studentDateDisplayFormat}));
-		}
-	}
-
-	return $title;
+	return $c->tag('span', dir => 'ltr', format_set_name_display($setID));
```
drgrice1, line 124: This method should be deleted entirely now.  This is only doing what the default ContenteGenerator page_title method will do for this page anyway.  No need for the override anymore.

drgrice1: Although, there is a mistake in lib/WeBWorK/ContentGenerator.pm that makes that not work right.  Line 694 of that file should be `return route_title($c, $c->current_route, 1);`.  I must have accidentally dropped the optional argument in the conversion to Mojolicious and templates.  It was there in webwork 2.17.

somiaj: I was in the process of deleting the method and testing and it seemed to work for me. Ahh it appears to be missing the LTR span? Also didn't test it with sets that have underscores. (Should I drop it here anyways with anticipation the issue you see will be fixed?)

drgrice1: That optional argument should also be on line 539 of ContentGenerator.pm.  It seems I dropped the optional argument everywhere in that file that it was supposed to be.  It should not be on the other instance of the route_title method in that file.  Can you fix those for me in this pull request?

somiaj: And leave the one at 619 alone I am assuming. Patch incoming.

drgrice1 (approved):
Looks good.  Nice improvement.  The page title was to long when it included the close date.

somiaj:
I noticed that when viewing a closed test with the new set status message, two messages are printed, first the status message, "Set is closed.", and second the test is closed message, "This test is closed.".  I see no reason for both alerts so removed the "This test is closed." message.

templates/ContentGenerator/Base/set_status.html.ep
```diff
@@ -0,0 +1,52 @@
+% use WeBWorK::Utils qw(after before between);
+%
+% my $useReducedScoring = ($ce->{pg}{ansEvalDefaults}{enableReducedScoring}
+	% && after($set->open_date)
+	% && $set->enable_reduced_scoring
+	% && $set->reduced_scoring_date
+	% && $set->reduced_scoring_date != $set->due_date
+% );
+% 
+<div class="alert alert-warning mb-3">
```
drgrice1, line 10: Seeing your last change made me notice this.  `alert-warning` is not right here.  This should be `alert-info`.  This is an information message, not a warning.  Something seemed off before with the color, now I see why.  Please change this.

somiaj: Updated. I was using the same alert that was being used for reduced scoring message prior to this change. But I agree, info looks better.

pstaabp (approved):

PR #1985, opened 2023-05-13
Remove WeBWorK::Constants and improve the webwork2 app configuration.

drgrice1:
The only constants still defined in the WeBWorK::Constants package were the WEBWORK_DIRECTORY and the PG_DIRECTORY.  Furthermore, even those were not used much in the code, and they are even set several times in scripts to "avoid an error message" from not having them defined.  The few places that these constants were used now either use WEBWORK_ROOT and PG_ROOT environment variables directly (from which the previous constants were set) or use values from the course environment.

The CourseEnvironment constructor grabs webwork_dir and pg_dir from the %WeBWorK::SeedCE hash if those parameters are not passed in, and falls back as a last resort to WEBWORK_ROOT and PG_ROOT environment variables.  This makes it so that within the webwork app, you do not need to pass the webwork_dir to construct a CourseEnvironment object.

Scripts will either need to set the %WeBWorK::SeedCE webwork_dir and pg_dir values, or pass webwork_dir and pg_dir to the CourseEnvironment constructor in its now optional argument, or just ensure the WEBWORK_ROOT and PG_ROOT environment variables are set.

Note the old four argument way of calling the CourseEnvironment constructor that has been marked as deprecated since 2004 is now removed entirely.

One upshot of this is that the pg_dir is not needed in site.conf (and so is removed), and this now only needs to be set in one place, namely webwork2.mojolicious.yml.

The other constants defined throughout the code that were previously set in WeBWorK::Constants are now set from configuration variables in webwork2.mojolicious.yml when the app initializes.  The only such constants are the WeBWorK::Debug constants and one WeBWorK::Hardcopy constant.  This means that to enable debug logging, you don't need to modify the code anymore.  Just enable it in the webwork2.mojolicious.yml file.

Note the `clients` and `t` directories were also deleted.  I am tired of seeing matches on things in these directories when grepping the code, and nothing in these directories works or is valid for anything anymore.


pstaabp (approved):
For testing, I just tried to run through all of the pages and looks good.  I would argue that let's merge this soon and if anything was missed we'll see with just general use.

drdrew42:
This will have implications for the standalone renderer... it will be nice to ditch the WeBWorK::Constants!

drgrice1:
Technically since the standalone renderer has its own code for this, it will not need any changes due to this pull request.  However, it would be good to make similar changes to the standalone renderer's code.

drgrice1:
@drdrew42:  I realized that I have already done this for the standalone renderer.  It is part of https://github.com/openwebwork/renderer/pull/7.

somiaj (approved):
Tested pages out for a bit with this, everything seems to work.

PR #1986, opened 2023-05-14
Fix and update Email page.

somiaj:
In looking to write help for the Email page I noticed that due to mojolicious caching of parameters, some of the buttons didn't correctly work on the email page.  I also felt that the page could be organized a bit better.  This restructures the Email page along with address the issues I originally noticed (main one was the open button didn't show the new file it opened). Summary of changes:

+ Removed the option to set the editor rows, since the email body text input can be resized.
+ Removed one of the refresh page buttons. It was mostly used for the editor row setting which was removed.
+ Renamed the student list refresh button to 'Change Display Settings' to match what is said on the instructor tools page.
+ Combined the message and merge file names into a single row.  I don't think we need to state what the last used selection was on the page.
+ Removed an error message about the user not found in the merge file from the editor page.  This message is also on the preview page and just checked the status of the professor which might not be in the CSV. Instead just show this message when they try to preview a message.
+ Don't state the professors name next to the preview button. It will preview the first selected student, and fall back to using the professor if no student was selected. Selecting a student didn't change the name next to the preview button, so it was at best misleading.
+ Move the elements around to better make use of the space.

If this gets approved, I'll update the help file in #1967 to match.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
@@ -88,8 +88,6 @@ sub initialize ($c) {
 		$action = 'saveDefault';
 	} elsif ($c->param('openMessage')) {
 		$action = 'openMessage';
-	} elsif ($c->param('updateSettings')) {
-		$action = 'updateSettings';
 	} elsif ($c->param('previewMessage')) {
 		$action = 'previewMessage';
 	}
@@ -102,8 +100,6 @@ sub initialize ($c) {
 	$c->{defaultReply}   = $ur->rfc822_mailbox;
 	$c->{defaultSubject} = $c->stash('courseID') . ' notice';
 
-	$c->{rows}    = (defined($c->param('rows')))    ? $c->param('rows')    : $ce->{mail}->{editor_window_rows};
-	$c->{columns} = (defined($c->param('columns'))) ? $c->param('columns') : $ce->{mail}->{editor_window_columns};
 	$c->{default_msg_file}     = $default_msg_file;
 	$c->{old_default_msg_file} = $old_default_msg_file;
 	$c->{merge_file}           = $mergefile;
@@ -180,6 +176,7 @@ sub initialize ($c) {
 	my $output_file = 'FIXME no output file specified';
 	if ($action eq 'saveDefault') {
 		$output_file = $default_msg_file;
+		$c->param('openfilename', $output_file);
 	} elsif ($action eq 'saveMessage' or $action eq 'saveAs') {
 		if (defined($savefilename) and $savefilename) {
 			$output_file = $savefilename;
@@ -219,9 +216,11 @@ sub initialize ($c) {
 	# Get inputs
 	my ($from, $replyTo, $r_text, $subject);
 	if ($input_source eq 'file') {
-
 		($from, $replyTo, $subject, $r_text) = $c->read_input_file("$emailDirectory/$input_file");
-
+		$c->param('from',    $from)      if $from;
+		$c->param('replyTo', $replyTo)   if $replyTo;
+		$c->param('subject', $subject)   if $subject;
+		$c->param('body',    ${$r_text}) if $r_text;
 	} elsif ($input_source eq 'form') {
 		# read info from the form
 		# bail if there is no message body
@@ -234,7 +233,6 @@ sub initialize ($c) {
 		$c->addbadmessage($c->maketext('You didn\'t enter any message.'))
 			unless $c->param('body') =~ /\S/;
 		$r_text = \$body;
-
 	}
 
 	my $remote_host = $c->tx->remote_address || "UNKNOWN";
@@ -250,11 +248,6 @@ sub initialize ($c) {
 	#     first time actions
 	#          open new file
 	#          open default file
-	#     choose merge file actions
-	#          chose merge button
-	#     option actions
-	#       'reset rows'
-
 	#     save actions
 	#       "save" button
 	#       "save as" button
@@ -270,13 +263,7 @@ sub initialize ($c) {
 	my $to            = $c->param('To');
 	my $script_action = '';
 
-	if (not $action
-		or $action eq 'openMessage'
-		or $action eq 'updateSettings')
-	{
-
-		return '';
-	}
+	return '' if (not $action or $action eq 'openMessage');
 
 	# If form is submitted deal with filled out forms
 	# and various actions resulting from different buttons
--- a/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
@@ -96,16 +96,17 @@ sub initialize ($c) {
 	my $ur = $db->getUser($user);
 
 	# Store data
-	$c->{defaultFrom}    = $ur->rfc822_mailbox;
-	$c->{defaultReply}   = $ur->rfc822_mailbox;
-	$c->{defaultSubject} = $c->stash('courseID') . ' notice';
+	$c->{defaultPreviewUser} = $ur;
+	$c->{defaultFrom}        = $ur->rfc822_mailbox;
+	$c->{defaultReply}       = $ur->rfc822_mailbox;
+	$c->{defaultSubject}     = $c->stash('courseID') . ' notice';
 
 	$c->{default_msg_file}     = $default_msg_file;
 	$c->{old_default_msg_file} = $old_default_msg_file;
 	$c->{merge_file}           = $mergefile;
 
 	my @classList = (defined($c->param('classList'))) ? $c->param('classList') : ($user);
-	$c->{preview_user} = $classList[0] || $user;
+	$c->{preview_user} = $c->db->getUser($classList[0] || $user);
 
 	# Gather database data
 	# Get all users except set level proctors and practice users.  If the current user has restrictions on viewable
@@ -217,10 +218,10 @@ sub initialize ($c) {
 	my ($from, $replyTo, $r_text, $subject);
 	if ($input_source eq 'file') {
 		($from, $replyTo, $subject, $r_text) = $c->read_input_file("$emailDirectory/$input_file");
-		$c->param('from',    $from)      if $from;
-		$c->param('replyTo', $replyTo)   if $replyTo;
-		$c->param('subject', $subject)   if $subject;
-		$c->param('body',    ${$r_text}) if $r_text;
+		$c->param('from',    $from)    if $from;
+		$c->param('replyTo', $replyTo) if $replyTo;
+		$c->param('subject', $subject) if $subject;
+		$c->param('body',    $$r_text) if $r_text;
 	} elsif ($input_source eq 'form') {
 		# read info from the form
 		# bail if there is no message body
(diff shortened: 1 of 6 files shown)


somiaj:
Note. I didn't test the email functionality (since I'm not setup to send emails), but the preview works and since I didn't touch any of the email code, I am assuming it will work as before any changes here.

drgrice1 (changes requested):
I haven't looked very closely at this yet, but I can already see that you have removed some UI elements from the page that should not be removed.

One immediate thing that I see that you removed is the display of the currently open merge file data on the email page itself. That definitely should still be there.  That is extremely useful to show what data you have to actually work with.

Also, with the removal of the elements that show the currently open message and merge file you have also removed clarity on the page.  It is now not clear which file is actually active and open on the page.

You also removed the element that shows which student a preview will be generated for.  This again makes things less clear as to what the page will do.

You claim that you have fixed something due to parameter caching (parameter caching is not new with Mojolicious and has never caused any problems -- I think you mean auto-filling of form inputs), but I am not sure what you fixed.  I also don't see anything broken to begin with.   What is not working?  Please be specific.

I am not really seeing a lot of improvement in this pull request.  You have removed functionality, actually made the page behavior less clear.  The reorganization is not really a very big improvement either.

somiaj:
The first thing that is broken is the 'open' button, when you try to open a new message, nothing is changed (the from, reply-to, subject, and body are all the previous entered value and not the value from the newly loaded message file). I did state this as an example of something that is broken.

As for the UI elements I removed, they didn't provide any info I found useful and instead found confusing. First the name attached the the preview button only states who the merge file preview was generated for at page load, but doesn't state who the preview will be generated for when you click the button. Instead the preview is generated for the first student that is selected, and falls back to the instructor if no students are selected. So to me it isn't clear what that name is for since the preview may not be generated for the student listed. Instead a preview button with the knowledge it uses the first selected student seems more appropriate to the behavior I saw testing the page out.

As for the merge file preview, to me it initially seemed to show the data of the instructor, or in my test cases because the instructor was not in the merge file no data was shown. Instead when testing I found it more useful to view the merge file data which is also on the preview page for the student I selected. So since the data was available in the previews, I was using that to see the structure of the merge file.

Though in testing again I did miss that if you select a student and refresh the page, the merge file will show the data for the first selected student, similar to the preview button. Maybe we should show the first row of the merge file if the instructor or the first selected student is not found so the structure of the merge file is on the page even if the instructor or selected student is not in the csv. Since I was not in the csv files I created with the scoring tools, the preview on the main page wasn't functional and I ended up using the preview button to preview what was in the csv file. You are correct this should be on the page, but maybe as I suggested show the top row if the student/instructor isn't found.

I don't really see the point of displaying the current open message filename in 3 different places. The only thing this seems to affect is the file that gets saved when you click 'Save' message, which still states that file name. So instead there is now an option to open a previously saved message and save over the last opened message which still gives the info of which message it was.

somiaj:
Update. I put the preview of the merge file back on the main form page. I added a 'View' button to the drop down menu for selecting the merge file, which you can use to view the selected merge file, which will show a single row of data for the first selected student, and fall back to the instructor if no student is selected.

I added a label to the data that states which merge file and which user is being viewed. This is what the open merge file name and preview student name was actually stating on the original page anyways, since it didn't appear to affect any other actions in my testing.

drgrice1:
@somiaj:  I added some pull requests to this branch that for your consideration.

drgrice1:
I think that the "Save As Default" button and its corresponding action should be removed.  That is functionality that really doesn't make sense as it is currently implemented.  The only thing that is "default" about it is that it saves the current message to a file named "default.msg".  However, the file that is opened by default on the email page is not that file, but the message file that was opened last (since that is what is saved to the database).

Furthermore, the file that is attempted to be opened when an instructor visits the page for the first time in a newly created course is the file named "default.msg".  However, that file usually does not exist in a new created course, because it is not in the model course.  The only message file that is in the model course, and hence in a newly created course, is the "welcome.msg" file.  So either the page should open "welcome.msg" initially, or a "default.msg" file should be added to the model course.

In general this thing that is done on this page with the "fixme" messages should be entirely removed.  That is not a good way to set up a UI, and is just bad messaging to begin with.

somiaj:
Removed the "Save as Default" button and better handle the display when no valid message file is found. In the case the message file is not valid, add a selected "None" option to the message file menu, remove the "Save" button, and set the "Save As" file name to `default.msg`.

lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
```diff
@@ -149,7 +137,6 @@ sub initialize ($c) {
 	# Check the validity of the input file name
 	my $input_file = '';
 	# Make sure an input message file was submitted and exists.
-	# Otherwise use the default message.
 	if (defined($openfilename)) {
```
drgrice1, line 140: With the addition of the "None" select option I think that this needs to be changed to
```
	if (defined $openfilename && $openfilename ne 'None') {
```
Otherwise the message "The file /opt/webwork/courses/myTestCourse/templates/email/None cannot be found. Check whether it exists and whether the directory /opt/webwork/courses/myTestCourse/templates/email can be read by the webserver." is shown when you click "Save As" when "None" is currently selected.  No message is needed in this case.

templates/ContentGenerator/Instructor/SendMail/main_form.html.ep
```diff
@@ -169,31 +171,26 @@
 		<%= label_for 'email-body', class => 'form-label', begin =%>
 			<%= maketext("Email Body:") %><span class="required-field">*</span>
 		<% end =%>
-		<%= text_area body =>
-			defined $c->{r_text} ? ${ $c->{r_text} } : 'FIXME no text was produced by initialization!',
-			id => 'email-body', rows => $c->{rows}, class => 'form-control' %>
+		<%= text_area body => $c->{r_text},
+			id => 'email-body', rows => $ce->{mail}->{editor_window_rows}, class => 'form-control' %>
```
drgrice1, line 175: There is an unnecessary dereferencing arrow here.  This should be `rows => $ce->{mail}{editor_window_rows}`.

lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
somiaj: Do you think just setting `$openfilename = undef if $openfilename eq 'None';` when defining that variable will catch this case and others when `None` is selected and the page gets submitted? Vs checking for this each time?

somiaj:
@drgrice1 I notice the following lines of code 165-169 in `templates/ContentGenerator/Instructor/SendMail/main_form.html.ep` that I don't think does anything.

```
	% # Create a textbox with the subject and a textarea with the message.
	% # Print the actual body of message.
	% if (defined $c->{message}) {
		<div class="alert alert-info p-1 my-2"><%= $c->{message} %></div>
	% }
```

The comment seems to suggest this might be left behind from some previous version of this page and use to print the body of the email message, but since `$c->{message}` isn't defined in `SendMail.pm` nothing is actually printed. I'm just double checking there isn't something else outside of `SendMail.pm` that could be setting an information message that should appear in this location.

lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
drgrice1: That should work also.

Although, for both the 'openfilename' and the 'merge_file' what really should be done is that the value should be the empty string instead of 'None'.  The textContent should be 'None' for the option, but the value should be the empty string.  That takes more work to implement though.

In any case, however you fix the openfilename, make sure you do that as well for the merge_file.

somiaj:
I made `None` a more permanent option for the message file, which sets a default From/Reply/Subject and a blank body to be used as a clean template to either save a new message or just send a one time message. Cleaned up all the errors associated with this by converting 'None' to an undefined filename and back.

drgrice1 (approved):
I think this is looking good now.

somiaj:
@drgrice1 thanks for all the help. I notice one tiny issue with the javascript. When you select a user, preview the message, then click the back button (as suggested on the page), the preview message user is reset to the user at page load and not the user the javascript changed the button to, which can make the user name incorrect.  I'm unsure how this can be fixed.

I also thought it would be useful to inform the user which users don't have merge data.  So I added a comment on the list of recipients on the preview page to indicate which users don't have merge data.

drgrice1:
@somiaj:  What browser are you using that you are seeing that on?

somiaj:
@drgrice1 Firefox.

drgrice1:
Odd, I am not seeing this with Firefox, but am with Google Chrome.  With Firefox when I click the back button the preview user shown is the first selected user.

somiaj:
I'm using firefox-esr which is version 102 if that matters.

drgrice1:
I am using Firefox 112.  I am not sure that it matters.  Perhaps...

drgrice1:
@somiaj: I added another pull request to this branch.  It should fix the issue for you.  It does on Google Chrome.  On the version of Firefox I am using this change isn't needed it seems.  I am betting that for your version of Firefox, the timeout is not needed.  Chrome has some odd issue.

somiaj:
Yea, unsure what is going on. I saw it fixed the issue in both firefox and chromium here. Though in firefox I see the change after the timeout is hit, but I don't see it in chromium. And yet another time I was testing this and the issue didn't appear at all, so not quite sure what is going on. Anyways, seems this fix will work just in case things get out of sync.

pstaabp:
This isn't something you've done, but I had a missing default.msg file and got the error:
```
FIXME file /opt/webwork/courses/test_course/templates/email/default.msg doesn't exist
```

This seems like it should be a message  or a warning instead of inside the email body texture.  It's on line 466 of SendMail.pm.

somiaj:
@pstaabp  This is the old behavior, that should have been removed in this change here. I will look at this more closely when I get home to double check.

drgrice1:
@pstaabp:  I am unable to get that to happen with this branch.  Even with no `.msg` files in the courses templates/email directory, the contents of the "Email Body:" textarea come up empty for me.  The message on line 466 of SendMail.pm should be updated, but it also shouldn't occur with this.  Did you run `npm ci` or `generate-assets.js`?

somiaj:
Looking closer it appears I missed the lines around 466 of SendMail.pm, and left them intentionally, but as @drgrice1 pointed out, that message should not be appearing at all.  Currently the subroutine `read_input_file` path is only processed if `$input_file` exists (`-e`) and is readable (`-R`). Then `read_input_file` checks again that the file exists and is readable (`-r`). To me it seems the whole checking again in `read_input_file` is not needed.

Though there is a subtle difference, `-R` readable by real user, while `-r` readable by effective user. I guess if your setup has a difference between the effective and real user it is possible to still get this error message. (I wonder if there is something lingering in the permissions in your courses directory from the switch to mojolicous?)

@drgrice1 I think we should just remove the additional check that `$input_file` exists and is readable from `read_input_file`, since this is already being checked elsewhere. Or should we leave this in for safety and update the else clause to match what is done lines 208-216 if the input file is not found. I'm unsure if the `-R` (real) vs `-r` (effective user) difference is important here, but should probably at least make those match.

lib/WeBWorK/ContentGenerator/Instructor/SendMail.pm
```diff
@@ -441,31 +441,23 @@ sub read_input_file ($c, $filePath) {
...
+		$header .= <$FILE>;
+	}
+	$text = join('', <$FILE>);
+	close $FILE;
 
-		$header =~ /^Reply-To:\s(.*)$/m;
-		$replyTo = $1 or $replyTo = $c->{defaultReply};
+	$text   =~ s/^\s*//;           # remove initial white space if any.
+	$header =~ /^From:\s(.*)$/m;
+	$from = $1 // $c->{defaultFrom};
```
drgrice1, line 454: To be the equivalent of what this was, this should be `$from = $1 || $c->{defaultFrom};`.  The difference is that if the file contains "From:  " with nothing after that (the line ends there), then `$1` will be the empty string, and with `//` that will be taken for the value of `$from`.  With `||` the value of `$from` will be that of `$c->{defaultFrom}`.

The same for `defaultReply` and `defaultSubject` below.

somiaj: Thanks.

pstaabp:
That error I was seeing was the actual default.msg file--I must have saved it earlier.

pstaabp (approved):
I haven't tested the email functionality, but the layout and other functionality looks good.

PR #1988, opened 2023-05-15
Don't allow renaming the current course (namely the admin course).

drgrice1:
I have noticed that when in the admin course the admin course is listed on the "Rename Course" page.  So I tried it out.  I renamed the admin course, changing the courseID from admin to admin_blah.  The page returned from the rename and said this course does not exist.  The rename was of course successful, and the admin course no longer existed. I then had to manually put everything back (rename the tables in the database and rename the course directory).  This probably should not be something that the page allows.  So this pull request removes the current course (which would be the admin course) from the list.  This is the same way that this is done for the "Delete Course" page.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -932,7 +932,7 @@ sub do_delete_course ($c) {
 sub archive_course_form ($c) {
 	my $ce = $c->ce;
 
-	my @courseIDs = listCourses($ce);
+	my @courseIDs = grep { $_ ne $c->stash('courseID') } listCourses($ce);
 	my %courseLabels;
 
 	if (@courseIDs) {
--- a/templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
+++ b/templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
@@ -112,7 +112,7 @@
 		<%= maketext('To copy problem templates from an existing course, select the course below.') =%>
 	</div>
 	<div class="row mb-3">
-		% my @existingCourses = sort { lc($a) cmp lc($b) } listCourses($ce);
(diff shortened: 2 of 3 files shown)


drgrice1:
I also removed the admin course from the "Archive Course" page listing.  There is not point in listing the admin course on this page, and then giving an error saying "You can not archive the course you are using."
    
Although, should we allow archiving of the admin course?  It works perfectly fine to archive the course you are using.  In fact we even have a link to do that in the site nav for a regular course.  The link is visible anytime you are on the "File Manager" page.  I tested it, and you can safely archive the admin course from the admin "Archive Course" page if you comment out the code that currently gives the warning.

somiaj (approved):
It appears that the admin course is listed in the courses you can copy templates from in the add course page. Maybe remove it there too?

Unsure the use of the admin archive, but if comment out said code, might need to put some checks around  the "Delete course after archiving" option. Also with with unarchiving would it make sense to give the unarchived admin course a new name?

drgrice1:
Yes, if archiving the admin course were to be allowed, protection would be needed against deleting after archiving.

Archiving an admin course would be for keeping a backup of the course, or to transfer to another server.

You could not unarchive an admin course.  The current "Unarchive Course" page prevents this because it doesn't allow you to unarchive a course that already exists, and the admin course always exists.  Note that unarchiving an admin course with a new name is useless, but could be done with the "Unarchive Course" page.  A horse by another name may still be a horse, but a webwork2 admin course by another name is no longer an admin course at all.  The webwork2 code only shows the admin home page for a course that is named "admin". If you rename the course and sign in to the course, you see the "Homework Sets" page, and no admin pages.

With a little more though, I don't think there is any point in allowing archiving the admin course.  The system admin can always make backups anyway.

Allowing the admin course to be copied as a template it really not a problem, but I will remove it from that list too.

pstaabp (approved):

drdrew42 (approved):

PR #1990, opened 2023-05-16
Replace more gateways in the ui with test.

somiaj:
I found another instance of `gateway` in the UI and realized I could use the translation files to track down more instances. This cleans up all the instances of `gateway` left I found via the translation files.


drgrice1 (approved):
Looks good.

PR #1993, opened 2023-05-19
Remove the allowedRecipients option.

drgrice1:
This option no longer is used for anything now that PG problems can not send email anymore.
    
Also clean up the generated course.conf file.  The segment regarding $mail{feedbackRecipients} is also no longer added to the course.conf file.  That can be set from the course configuration (written to simple.conf), and is not needed here.  The actual setting is usually not set here anyway.  All the commentary about this course environment setting is certainly not needed here.  The copyright and license certainly do not need to be output to a generated file.  Also remove some excessive newlines and make the output more condensed.

The course.conf file in courses.dist/modelCourse is modified accordingly.  In addition, the commentary about the $achievementExcludeSet variable is removed.  Really this file should be deleted.  The file itself is not used for anything. The only thing that it has of value is the documentation on the $viewable_sections and $viewable_recitations hashes.  Those hashes should be able to be set from the course configuration, and the documenation added in defaults.config.

For now I left the dbLayoutName and dbLayout setting in the course.conf.  Does anyone actually use the sql_moodle layout?  Is it even valid anymore?  Should this be entirely removed from the code, and just have a fixed database layout?

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -42,26 +42,16 @@ include("VERSION");  # get WW version
 # Mail Settings
 ################################################################################
 
-# AllowedRecipients defines addresses that the PG system is allowed to send mail
-# to. this prevents subtle PG exploits. This should be set in course.conf to the
-# addresses of professors of each course. Sending mail from the PG system (i.e.
-# questionaires, essay questions) will fail if this is not set somewhere (either
-# here or in course.conf).
-$mail{allowedRecipients}     = [
-	#'prof1@yourserver.yourdomain.edu',
-	#'prof2@yourserver.yourdomain.edu',
-];
-
 # By default, feedback is sent to all users who have permission to
-# receive_feedback. If this list is non-empty, feedback is also sent to the
-# addresses specified here.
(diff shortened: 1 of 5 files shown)


somiaj:
Do you think that `course.conf` should contain things that can be set via the course configuration in `simple.conf`, such as `$mail{feedbackRecipients}` and `$achievementExcludeSet`? My suggestion is just to remove them since they can be set in the course configuration.

drgrice1:
@somiaj:  Yeah, the `$mail{feedbackRecipients}` really doesn't need to be in the generated course.conf file.  I removed that.  The `$achievementExcludeSet` is not in the generated course.conf file.  You are referring to the course.conf file in courses.dist/modelCourse that I updated.   That file really needs to be deleted.

somiaj:
I thought those two locations were in sync, and didn't realize there was differences. Anyways, I too say remove the unused file since one gets generated, no need to update the same info in two locations.

drdrew42 (approved):

somiaj (approved):

PR #1994, opened 2023-05-19
Make renaming a course or unarchiving a course to a new name work again.

drgrice1:
This was using the old four argument form of the CourseEnvironment constructor.  I missed that in #1985.

To test this rename a course or unarchive a course to a new name.  On the current release candidate branch you will get an error.  With this pull request it will work.

Note that if you test this on the release candidate branch and rename a course, the course directory will be renamed.  So you will need to manually rename that back to what it was.  The database tables will still be under the old course name.  If you attempt to unarchive to a new name, then the course directory will be created with the new name, but the database tables will not be.  So you will need to manually delete that new course directory.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -1233,6 +1233,10 @@ sub unarchive_course_validate ($c) {
 		push @errors, $c->maketext('Course ID cannot exceed [_1] characters.', $ce->{maxCourseIdLength});
 	}
 
+	unless ($courseID =~ /^[\w-]*$/) {    # regex copied from CourseAdministration.pm
+		push @errors, $c->maketext('Course ID may only contain letters, numbers, hyphens, and underscores.');
+	}
+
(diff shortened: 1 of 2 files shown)


drgrice1:
I think that this pull request is not the place for that type of change.  This is a straight forward bugfix pull request.  It fixes something that is broken, and is a critical issue for the release.  Making spaces work in course names is a new feature.

pstaabp:
We'll leave this as is then.  I would say that fixing the space problem would be a bug fix since not checking for them results in a non-functional course. I'll add an issue to flag it.

pstaabp (approved):

drgrice1:
I am not sure that you can classify behavior that has been the norm for 20 years (since the inception of webwork2) as a bug fix.  I am not opposed to making the change though in another pull request.

drgrice1:
So I added a check for course name validity when unarchiving a course to a new name now.

pstaabp:
Thanks for adding the check @drgrice1.  Looks good.

drdrew42 (approved):

PR #1997, opened 2023-05-24
Minor typo on File Manager page.

pstaabp:
Fix a minor typo.


drgrice1 (approved):
If you want a binary file type then you check the second "Text" radio button!

PR #1999, opened 2023-05-24
Remove more of the null byte hack for gateway quizzes.

### LINKED ISSUE #1996: Tests do not save answers for multiple select problems on page change.
When a test has multiple pages and a problem is using a multiple select / checkbox grader, the answers are not saved when switching between pages. Here is the problem I'm testing with:

```
DOCUMENT();      
loadMacros(
    'PGstandard.pl',
    'PGML.pl',
    'PGchoicemacros.pl',
    'PGcourse.pl'
);

$mc = new_checkbox_multiple_choice();
$mc->qa('Select True Statements:', 'True1', 'True2');
$mc -> extra('False1', 'False2', 'False3');

BEGIN_PGML
[@ $mc->print_q() @]*
    [@ ANS(checkbox_cmp($mc->correct_ans())); $mc->print_a() @]*

END_PGML
ENDDOCUMENT();
```

I tested using `parserCheckboxList.pl`, and noticed the same problem too.

drgrice1:
The null byte array concatentation was still being checked for in templates/ContentGenerator/GatewayQuiz.html.ep.  Instead of splitting on the null byte, this should check for an array reference and add a hidden field with the same name for each element of an array.

This fixes #1996.

### DIFF
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -596,8 +596,12 @@
 				% # Print out hidden fields with the current last answers.
 				% my $curr_prefix = 'Q' . sprintf('%04d', $problem_numbers->[ $probOrder->[$i] ]) . '_';
 				% for my $curr_field (grep {/^(?!previous).*$curr_prefix/} keys %{ $c->{formFields} }) {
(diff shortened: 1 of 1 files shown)


somiaj (approved):
Fixes the issue. Thanks.

drdrew42 (approved):

PR #2000, opened 2023-05-25
Make the url for pg htdocs files configurable.

drgrice1:
This is controlled by the variable $pg_htdocs_url in site.conf.

### DIFF
--- a/lib/Mojolicious/WeBWorK.pm
+++ b/lib/Mojolicious/WeBWorK.pm
@@ -59,6 +59,7 @@ sub startup ($app) {
 	# Set important configuration variables
 	my $webwork_url         = $ce->{webwork_url};
 	my $webwork_htdocs_url  = $ce->{webwork_htdocs_url};
+	my $pg_htdocs_url       = $ce->{pg_htdocs_url} // '/pg_files';
 	my $webwork_htdocs_dir  = $ce->{webwork_htdocs_dir};
 	my $webwork_courses_url = $ce->{webwork_courses_url};
 	my $webwork_courses_dir = $ce->{webwork_courses_dir};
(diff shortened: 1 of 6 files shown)


lib/Mojolicious/WeBWorK.pm
```diff
@@ -193,9 +194,9 @@ sub startup ($app) {
 		}
 	);
 
-	# Provide direct access to pg_files.
+	# Provide direct access to the pg htdocs location.
 	$r->any(
-		'pg_files/*static' => sub ($c) {
+		"$pg_htdocs_url/*static" => sub ($c) {
```
Alex-Jordan, line 199: Is it significant that previously, this did not have thee slash before `pg_files`?

Alex-Jordan:
I note that `/pg_files` is written in the apache and nginx config files. Would it be good to have comments in those files to remind people to make it match what they put in `site.conf`?

lib/Mojolicious/WeBWorK.pm
drgrice1: It will work with or without the leading slash.  It probably should have had the leading slash before for consistency.  All of the other root routes in that file have the leading slash.

drgrice1:
It wouldn't hurt to have such comments.  However, note that such comments should be added for all locations in those files.  Every location in those config files, even the base `/webwork2` location is configurable.

drgrice1:
I added comments in `conf/webwork2.apache2.4.dist.conf` and `conf/webwork2.nginx.dist.conf` pointing out which site.conf setting each location is tied to.

Alex-Jordan (approved):

pstaabp:
Thinking about people that might upgrade and miss adding the `$pg_htdocs_url = "/pg_files";` line to `site.conf`.  If you don't have this, you get the response `File Not Found` from the server.  

I wonder if we should add in a more helpful error message--maybe in `lib/Mojolicious/WeBWorK.pm` if it's not defined.

drgrice1:
@pstaabp:  I made it so that if `$pg_htdocs_url` is not set in `site.conf`, then the default is `/pg_files`.  This is set in `lib/Mojolicious/WeBWorK.pm`.

pstaabp (approved):

PR #2002, opened 2023-05-26
Stop using system environment variables in bin scripts to detect the webwork and pg root locations.

drgrice1:
Instead the webwork root directory is determined from the script location.  This uses the Mojo::File curfile method for this.  Note that this approach does work with symlinks to a script file.

The need for the scripts to have the pg root location at all has been removed.  To make this work a few changes were needed.

First, the direct dependence on WWSafe is removed.  The CourseEnvironment.pm and AchievementEvaluator.pm modules don't need the old version of Safe that is WWSafe.  They work perfectly well with the newer versions of Safe (2.43 on Ubuntu 22.04).  So the native perl Safe is used instead.

Next, the WeBWorK::Utils::CourseIntegrityCheck module needed to not use the createDirectory method from WeBWorK::PG::IO.  That method has been copied into WeBWorK::Utils for now.

Finally, the WeBWorK::Utils::CourseManagement module needed to not use the not_null method from PGUtil.  That was a silly dependency in any case.  There is no need to use that method just to see if a variable is defined and a non-empty string.

Note that the addcourse and delcourse scripts are cleaned up quite a bit.  Also the pg scripts that depended on NPL modules that don't even exist anymore were deleted.  Also the integrity_check.pl script was removed.  That wasn't working or well developed to begin with, and all it does is check the directory structure for a hard coded course with the id "tmp_course".

There are some scripts that still use the system environment variables, but they are developer scripts in bin/dev_scripts.

For the bin/dev_scripts/generate-ww-pg-pod.pl script either the PG_ROOT environment variable must be set, or the --pg-root command line option must be set if you want the POD for pg to be generated.  By default only the webwork2 POD will be generated, and the webwork2 root directory is automatically detected.  Note that the method used in this script does not work with symlinks.

The bin/dev_scripts/update-localization-files still needs both of the environment variables to be set.  That is a bash script and it isn't as easy to do this with bash.  It would be possible though.  Since this is a developer script only, I am putting that off for now.


drgrice1:
I found a way to fix the Mojolicious/Safe conflict that seems to be working nicely.

drdrew42:
Does this mean #2003 should be closed in favor of this one?

drgrice1:
Yeah, I prefer this approach.  I will close #2003.

Alex-Jordan:
These are scripts for a server admin, so I'm not sure, but is there a need to add `Mojo::File` to `check_modules`?

drgrice1:
`Mojo::File` is part of the `Mojolicious` stack.  So it already is in `check_modules`.

pstaabp:
@drgrice1 I had trouble running some of the scripts due to the shebang line hard coding perl to `/usr/bin/perl`.  I just put in a PR to this that switches these to `/usr/bin/env perl`.  It doesn't update some of the other scripts that you didn't update here though. I can put this in separate if you prefer.

drgrice1:
@pstaabp:  I merged your pull request.  So this now includes a shebang rework.

drdrew42 (approved):

PR #2003, opened 2023-05-26
ALT Stop using system environment variables in bin scripts to detect the webwork and pg root locations.

drgrice1:
Instead the webwork root directory is determined from the file location. This uses the Mojo::File curfile method for this.  Note that this approach does work with symlinks to a script file.

The difference with this and #2002 is that this reads the pg root directory from the pg_dir setting in the conf/webwork2.mojolicious.yml file (or if that file does not exist the the default in conf/webwork2.mojolicious.yml.dist is used), and does not change the WWSafe usage.  So instead of setting PG_ROOT, you must first set pg_dir in that file if you use a location other than /opt/webwork/pg.   The reason I am submitting this pull request is that I am seeing an issue with #2002 that I am not sure how to resolve.  Some conflict with perl Safe and Mojolicious.

Also the addcourse and delcourse scripts are cleaned up quite a bit.

There are some scripts that still use the system environment variables, but they are developer scripts in bin/dev_scripts.

For the bin/dev_scripts/generate-ww-pg-pod.pl script either the PG_ROOT environment variable must be set, or the --pg-root command line option must be set if you want the POD for pg to be generated.  By default only the webwork2 POD will be generated, and the webwork2 root directory is automatically detected.  Note that the method used in this script does not work with symlinks.

The bin/dev_scripts/update-localization-files still needs both of the environment variables to be set.  That is a bash script and it isn't as easy to do this with bash.  It would be possible though.  Since this is a developer script only, I am putting that off for now.


drgrice1:
You might note, that this was actually written before #2002.

PR #2004, opened 2023-05-26
Fix a possible division by zero issue with achievements.

drgrice1:
I keep seeing this occur.  I am not entirely sure how it happens, but in some cases the previous level points and the next level points end up being the same.  This results in the computed level goal being zero in Achievements.pm, and that causes a division by zero error.  There should always be protection against division by zero in any case.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Achievements.pm
+++ b/lib/WeBWorK/ContentGenerator/Achievements.pm
@@ -76,7 +76,7 @@ sub getAchievementLevelData ($c) {
 			$level_progress   = $c->{globalData}->achievement_points - $prev_level;
 			$level_progress   = 0           if $level_progress < 0;
 			$level_progress   = $level_goal if $level_progress > $level_goal;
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
This protects against division by 0. It would be nice to understand how it occurs that the two things are equal though. Like is it also possible for previous level points to sometimes be more than next level points, and you'd be dividing by a negative?

somiaj:
It is the level achievement files that both test and set the xp values.  I looked through the default achievements we provide in the modelCourse and they all seem to look fine for testing and setting the xp for the next level.  Since users can modify these level achievements it is possible they change things on an active class or miss-configure them to cause a division by zero.  So a test is probably the best we can do if a user modifies the level achievements incorrectly.

I also think if you enable achievements, but don't actually assign the `level_one.at` achievement, the initial setting of making the user level one and assigning the xp required to get level 2 is never run.  Unsure if that could also be causing problems here.

Alex-Jordan (approved):

PR #2006, opened 2023-05-27
Convert the `bin/check_latex` script to perl so Course Environment settings can be used.

drgrice1:
Note that this is built on top of #2002 so that it can use the webwork2 root detection method in that pull request and not depend on the WEBWORK_ROOT and PG_ROOT environment variables.

The only file changed from #2002 is `bin/check_latex`.

### DIFF
--- a/bin/check_latex
+++ b/bin/check_latex
@@ -1,21 +1,69 @@
-#!/bin/bash
+#!/usr/bin/env perl
+################################################################################
+# WeBWorK Online Homework Delivery System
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I checked out this branch and ran `check_latex`, and it works. (It does not work otherwise as the script wants to use my incomplete system latex.)

So I'm registering approval! But #2002 is a bigger thing that I haven't looked into yet. So not approving with GutHub tools yet.

pstaabp:
Yes. `check_latex` works, but we should wait until #2002 is merged.

drdrew42 (approved):

PR #2008, opened 2023-05-29
Make the achievement editor sublinks have the "instructor" color.

drgrice1:
I noticed that if you are on the "Achievement Editor" page, and click on the "Edit Users" column for an achievement, the achievment nav item that is added with the achievement id does not have the "instcutor" coloring, and so stands out oddly.  This adds the `list-group-item list-group-item-primary` classes to make that match the other instructor links.

Also fix some lines that have become overly long in that file.

### DIFF
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -129,7 +129,9 @@
 			% }
 
 			% # Set assigner
-			<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_set_assigner') %></li>
+			<li class="list-group-item list-group-item-primary nav-item">
+				<%= $makelink->('instructor_set_assigner') %>
+			</li>
 			% # Library Browser
 			<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_set_maker') %></li>
 			% # Statistics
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
fixes the miscoloring.

Alex-Jordan:
I was wondering if it would look better to reverse the colors. It would
certainly be simpler that way. Not sure how it would look though for
students with all the menu links in blue.

drgrice1:
I kind of like the current setup.  With this students will see things they way they were before, and only instructors will see this coloring.

I might change my mind if you work it up and it looks better that way though.

Note the coloring is not always blue.  That depends on the theme.  So if you do change it, make sure that you check it at least with the four supported themes.

Alex-Jordan:
Here are screenshots with the `primary...` class on the student-facing menu items. I don't care much, but I think I do like it better this way. Maybe if `MAIN MENU` also had some similar coloring.
Screen Shot 2023-05-29 at 9 21 09 PM
Screen Shot 2023-05-29 at 9 19 58 PM
Screen Shot 2023-05-29 at 9 19 23 PM
Screen Shot 2023-05-29 at 9 16 49 PM
Screen Shot 2023-05-29 at 9 16 09 PM
Screen Shot 2023-05-29 at 9 15 42 PM
Screen Shot 2023-05-29 at 9 15 29 PM
Screen Shot 2023-05-29 at 9 15 09 PM


Compare with current math4 (blue):

Screen Shot 2023-05-29 at 9 24 33 PM
Screen Shot 2023-05-29 at 9 24 22 PM

drgrice1:
I think that it might look alright if "Main Menu" also has the coloring.

Issue #1992, opened 2023-05-18
Change name of Hmwk Sets Editor and possibly Classlist Editor


dlglin:
"Hmwk" needs to be replaced with a full word.  There was discussion about the fact that not all sets are homework, so we should find a different name.  Suggestions include "Assignments" or just "Sets".

Also consensus seemed to be that this should be changed from "Editor" to "Manage".

How about:

- "Classlist Editor" -> "Manage Users"
- "Hmwk Sets Editor" -> "Manage Assignments"

The biggest implication of this would be to existing documentation.  Both internal documentation and custom site-specific documentation refer to the old names, which may make people confused and/or unhappy.

pstaabp:
I would argue to make the new name nouns to roughly fit it with the rest of the name of the tools.  Perhaps:

* "Classlist Editor" -> User Manager
* "Hmwk Sets Editor " Assignment Manager

drgrice1:
It seems that the `Achievement Editor` should also be renamed.  Perhaps to `Achievement Manager`.  The name `Achievement Editor` is confusing when there is also an page that is for editing achievement evaluators that seems more like an `Achievement Editor` than this page.

Alex-Jordan:
For a little while now in develop, in the TOC we've had:
```
User Manager
Set Assigner
Set Manager
```

<img width="252" alt="Screenshot 2023-11-24 at 11 42 49 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/663b4b3d-2005-4e08-840c-d25b0bd7c325">

Each time I go into a course on develop (to test or develop something) I see these and I'm not liking them. At first I thought it was just reflex because they are new and different. But now I'm starting to feel that I dislike these for more than just the bias of familiarity. Does anyone feel similarly? With the repeated use of `Set` and `Manager`, it feels cluttered. Or like it's missing a `User Assigner` to round it out. `Set Assigner` bugs me now, because you can "assign the sets to users" but you could also think of it as "assigning the users to the sets".

It's hard to think of better names. But I'm going to toss out these and see if anything sticks in the opinions of others reading this. In addition to renaming those three, I renamed `Acheivement Editor` as @drgrice1 suggested (except with plural). And also `Homework Sets` at the top of the TOC.

<img width="252" alt="Screenshot 2023-11-24 at 11 48 34 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/5ccf480e-7c69-4c1a-b664-36b97479f324">

[...]

PR #2009, opened 2023-05-30
Add customized menu for the admin course.

somiaj:
Move the links (nav menu) from the top of the admin page over to
  the main menu.  Also remove most of the course links since they
  don't really do much on the admin page.  Label this new menu
  'Admin Menu'.

The only old menu items used are the user settings to be able to change the password, class list editor, and email since the admin users may want to email various instructors or manage them (since they are created automatically with new courses).  There was also some discussion of maybe the file manager being useful, but not added here. I had wondered if the admin file manager could go up one more directory to the courses directory root, so admin users have access to all course files.

### DIFF
--- a/templates/ContentGenerator/Base/admin_links.html.ep
+++ b/templates/ContentGenerator/Base/admin_links.html.ep
@@ -1,0 +1,48 @@
+<h2 class="navbar-brand mb-0"><%= maketext('Admin Menu') %></h2>
+<ul class="nav flex-column">
+	<li class="list-group-item list-group-item-primary nav-item">
+		<%= link_to maketext('Courses') => 'root', class => 'nav-link' %>
+	</li>
+	<li class="list-group-item list-group-item-primary nav-item">
+		<%= $makelink->(
+			'set_list',
+			text   => maketext('Course Administration'),
+			active => !param('subDisplay') && $c->url_for =~ /admin$/ ? 1 : 0
+		) %>
+	</li>
+	% for (
+		% [
+		% 	'add_course',
+		% 	maketext('Add Course'),
+		% 	{
+		% 		add_admin_users      => 1,
+		% 		add_config_file      => 1,
+		% 		add_dbLayout         => 'sql_single',
+		% 		add_templates_course => $ce->{siteDefaults}{default_templates_course} || ''
+		% 	}
+		% ],
+		% [ 'rename_course',        maketext('Rename Course') ],
+		% [ 'delete_course',        maketext('Delete Course') ],
+		% [ 'archive_course',       maketext('Archive Course') ],
+		% [ 'unarchive_course',     maketext('Unarchive Course') ],
+		% [ 'upgrade_course',       maketext('Upgrade Courses') ],
+		% [ 'manage_locations',     maketext('Manage Locations') ],
+		% [ 'hide_inactive_course', maketext('Hide Courses') ],
+	% )
+	% {
+		<li class="list-group-item list-group-item-primary nav-item">
+			<%= $makelink->(
+				'set_list',
+				text              => $_->[1],
+				systemlink_params => { subDisplay => $_->[0], %{ $_->[2] // {} } },
+				active            => (param('subDisplay') // '') eq $_->[0],
+			) %>
+		</li>
+	% }
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('options') %></li>
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_user_list') %></li>
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_mail_merge') %></li>
+	<li class="list-group-item list-group-item-primary nav-item">
+		<%= link_to maketext('Report bugs') => $ce->{webworkURLs}{bugReporter}, class => 'nav-link' =%>
+	</li>
+</ul>
--- a/templates/ContentGenerator/Base/admin_links.html.ep
+++ b/templates/ContentGenerator/Base/admin_links.html.ep
@@ -42,6 +42,7 @@
 	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('options') %></li>
 	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_user_list') %></li>
 	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_mail_merge') %></li>
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_file_manager') %></li>
 	<li class="list-group-item list-group-item-primary nav-item">
 		<%= link_to maketext('Report bugs') => $ce->{webworkURLs}{bugReporter}, class => 'nav-link' =%>
 	</li>
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -1,5 +1,9 @@
 % use WeBWorK::Utils qw(jitar_id_to_seq);
 %
+% if ($ce->{courseName} eq 'admin') {
+	<%= include 'ContentGenerator/Base/admin_links' =%>
+	% next;
+% }
 <h2 class="navbar-brand mb-0"><%= maketext('Main Menu') %></h2>
 <ul class="nav flex-column">
 	% unless ($restricted_navigation) {
@@ -13,12 +17,7 @@
 			% if ($restricted_navigation) {
 				<span class="nav-link disabled"><%= maketext('Homework Sets') %></span>
 			% } else {
-				<%= $makelink->(
-					'set_list',
-					text => $ce->{courseName} eq 'admin'
-						? maketext('Course Administration')
-						: maketext('Homework Sets'),
-				) %>
+				<%= $makelink->('set_list', text => maketext('Homework Sets')) %>
 			% }
 		</li>
 		%
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -286,7 +286,7 @@
 			% && $authz->hasPermissions($userID, 'report_bugs'))
 		% {
 			<li class="list-group-item list-group-item-primary nav-item">
-				%= link_to maketext('Report bugs') => $ce->{webworkURLs}{bugReporter}, class => 'nav-link'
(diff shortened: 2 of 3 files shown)


Alex-Jordan:
I poked around and it looks good. i did not try actually using admin tools but i assume they still all work.

Should there still be a Help menu item? (Specialized to the admin course...) And report bugs?

somiaj:
I think if we want help menu items, we should also give more help to the various admin pages, similar to what has been done for normal courses.  That is beyond this pull request though, maybe another pull request can add admin help pages and a general help menu.

I can add the report bugs link, though for course admins, is that Bugzilla link the appropriate place to send them to report bugs about the admin course?  Though I guess they could also be reporting bugs about a problem or something else as well, just using the admin course for the link.

drgrice1:
I think the "Course Administration link should be above the "User Settings" link in the nav.  That is the home page, and so it should be first.

Probably the coloring should either be dispensed with for the admin course, or made entirely the way the #2010 makes it for students.  In other words, uniform.

somiaj:
Moved user settings more towards the bottom (under the course admin links right before the class list editor). Added report bug link. Made everything the primary color assuming #2010 gets accepted.  The Menu header's color will be updated with #2010 changes.

I think the admin help additions should be put in another pull request, so leaving help link off for now.

templates/ContentGenerator/Base/admin_links.html.ep
```diff
@@ -0,0 +1,50 @@
...
+	% }
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('options') %></li>
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_user_list') %></li>
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_mail_merge') %></li>
+	<li class="list-group-item list-group-item-primary nav-item">
+		%= link_to maketext('Report bugs') => $ce->{webworkURLs}{bugReporter}, class => 'nav-link'
+	</li>
+</ul>
+
+
```
drgrice1, line 50: There are a couple of empty lines here that should be removed.

templates/ContentGenerator/Base/admin_links.html.ep
```diff
@@ -0,0 +1,50 @@
...
+				systemlink_params => { subDisplay => $_->[0], %{ $_->[2] // {} } },
+				active            => (param('subDisplay') // '') eq $_->[0],
+			) %>
+		</li>
+	% }
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('options') %></li>
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_user_list') %></li>
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('instructor_mail_merge') %></li>
+	<li class="list-group-item list-group-item-primary nav-item">
+		%= link_to maketext('Report bugs') => $ce->{webworkURLs}{bugReporter}, class => 'nav-link'
```
drgrice1, line 46: To be consistent this line should be
```suggestion
		<%= link_to maketext('Report bugs') => $ce->{webworkURLs}{bugReporter}, class => 'nav-link' =%>
```
There are also some other places that perl segments should end with `=%>` as well.  I have been trying to add that anytime that a perl segment is has spaces around it (as the `=%>` removes that spacing in the generated html.  I really wish that Mojolicious templates were the opposite in terms of spacing and new lines.  That is I would prefer that spacing and new lines were "opt in" and not "opt out".  Modern html should be minimized, and Mojolicious templates don't support that without a lot of effort.

drgrice1 (approved):
This looks good.  I think this is a nice improvement to the admin course.

Eventually it would be good to remove the `sub_display` url paramenter, and make the admin pages proper routes.  But that is for future work.

templates/ContentGenerator/Base/admin_links.html.ep
somiaj: I grabbed this line from `templates/ContentGenerator/Base/links.html.ep`, should it be updated there too?  I thought it was weird this line was different than other ones, but went with it due to the link actually points to a new site.  I have updated your fix in my PR though.

somiaj:
Applied @drgrice1 code cleanup comments.  I am working on adding some help items to the admin pages (in case someone else was thinking about working on this).  I'll add a PR when its ready, though lots of the help is quite short (most pages already describe what they do).  We can talk about this more once I make the PR.

templates/ContentGenerator/Base/admin_links.html.ep
drgrice1: Yeah, when I converted to Mojolicious templates I started with the admin course, and my techniques changed some as I worked through all of the content generator modules.  I tried to go back and update, but that was a large project and things were missed.

somiaj: Okay, I fixed the occurrence that I copied this from too.

Alex-Jordan:
I'm interested in keeping File Manager available in the admin course.

After Googling today, I find that I can:
```
cd /opt/webwork/courses/admin
sudo mkdir courses
sudo chown apache:apache courses
sudo mount --bind /opt/webwork/courses courses
sudo mount -o bind,remount,ro courses
```

And now in the admin course's file manager, I have a `courses` folder that is like a symlink to `/opt/webwork/courses/` except that everything I access through this `courses` mount is read-only.

This would be useful for me managing the Runestone server, where users request their full archive files from the past, and I could get them through the admin course. And I'm not risking some kind of accident since it's read-only.

Similarly, I might want to mount like this to the `webwork2/logs/` folder.

Thoughts on this?

somiaj:
@Alex-Jordan My thoughts was to just give the admin file manager access to all courses by allowing it to move all the way up to the courses root directory.  This also makes it so the admin file manager can download and upload archives (which cannot currently be done).  It appears this is fairly easy to pull off, I have a patch on this branch, https://github.com/somiaj/webwork2/tree/admin-file-manager. Thoughts?

Alex-Jordan:
Personally I would like to avoid the admin course having so much write access. With what you propose, one of my colleagues (who have access to the admin course) might accidentally do something very bad and lose/corrupt files from the courses folder tree.

somiaj:
Ahh I missed you mounted the link read only.  I still think it would be nice to be able to add archives (I like to copy an archive from my live server to my test server) from the admin course, though since I have server access it isn't needed.  Though agreed it might be too much access.  In either case I can add the file manager back.

pstaabp:
I think this comment got lost above (or lost in translation).  On the User Settings, there are the "Change Display Settings", which isn't relevant for the admin course.  It looks like you are using the existing User Settings page.  Can the "Change Display Settings" block be hidden if the course is the admin course?

Alex-Jordan:
I know this is merged now but this seems like an OK place for discussion. How do you feel about:

1. "Course Administration" -> "Course Listings" ?  It's no longer the page for doing all the administration.
2. Swap the order of "Manage Locations" and "Hide Courses".
3. Make all the singular "Course" into a uniform plural "Courses", or drop the trailing "Course"/"Courses" altogether from the nav links.
4. Remove the "Courses" link. It's not really offering anything that "Course Administration"/"Course Listing" isn't offering already.

(Actually maybe "Courses" can go away for all courses? You can just click the WeBWorK logo to get there if you want to...)

somiaj:
Seems reasonable, maybe add a PR.

> 3\. Make all the singular "Course" into a uniform plural "Courses", or drop the trailing "Course"/"Courses" altogether from the nav links.

The plural is being used to indicate wither the tool can only act on a single course at a time or multiple courses at a time.

drgrice1:
@Alex-Jordan:  All of your suggestions sound good.  Could you work up a PR for those things?  Note that if you remove the "Courses" link from the site nav, then you will also need to find a fix for @somiaj's issue pointed out in https://github.com/openwebwork/webwork2/pull/2010#issuecomment-1572752710.

Alex-Jordan:
Ok I'll work this up. One more I might do is move User Settings to the top
and have it alone use the class that gives it color. To be consistent with
other courses.

If a user below admin level enters the admin course, is it the case that
all they see is user settings? I haven't checked.

drgrice1:
> If a user below admin level enters the admin course, is it the case that all they see is user settings? I haven't checked.

That is a good question. That certainly should be the case. I will check that when I get a chance.

somiaj:
I don't think I tested for this, so most likely all users in the admin course see the same menu.

drgrice1:
I just tested this, and yes, all users see the full site nav.  Users that don't have student permissions see "You are not authorized to create or delete courses" or "You are not authorized to access instructor tools" for all of the pages except the "User Settings" page.  Users that only have "professor" permission can actually create and delete courses (and do everything).  So this is a problem that needs to be fixed.

somiaj:
> > 3. Make all the singular "Course" into a uniform plural "Courses", or drop the trailing "Course"/"Courses" altogether from the nav links.
> 
> The plural is being used to indicate wither the tool can only act on a single course at a time or multiple courses at a time.

Note, it appears that `Archive Course` should then be `Archive Courses`, because you can select multiple courses at once to archive (though then you end up archiving them one at a time with a string of confirmation pages).  I'm not suggesting you keep this naming convention, just pointing out what I thought the plural was being used for.  It will probably look better to be more consistent in the menu.

@Alex-Jordan If you make a PR to update the link names here, might also want to consider some of the suggestions in #1992 and update the names on the main menu as well.

drgrice1:
@somiaj:  Making the renames from #1992 is a little more involved than just changing the links in the site nav.  Then you also need to change all of the references to those things in the help files.

Alex-Jordan:
I am about to open a PR that has my initial suggestions. I have to stop WW stuff for the rest of the day and possibly weekend. So the things from #1992 and the issue about people with low permissions being able to do things they shouldn't will come separately. I don't think I will bee available until late next week, so if anyone else wants to do those, go for it.

somiaj:
I will go fix the issue I overlooked with permissions.

drgrice1:
I just tested on webwork 2.17, and I see that a user with 'student' permissions in the admin course can't see or do anything.  However, a user with 'professor' permissions could create and delete courses (and everything else).  So this change makes the 'student' user issue messy looking, but the 'professor' issue is not new (but should be fixed).

somiaj:
Removing the items from the menus will be easy for 'professor' or anyone with 'access_instructor_tools', but if they actually type in the URL of a given instructor tool they will most likely still have access.  Maybe just increase 'access_instructor_tools' to 'admin' in the admin `course.conf` file?  This may not work for existing admin courses though.

drgrice1:
So all that needs to be done is to hide the links for those that do not have permission to perform those actions.

PR #2010, opened 2023-05-30
move primary coloring in navbar to student-accessible menu items

Alex-Jordan:
This changes the navbar coloring so the student-accessible menu items get the primary coloring instead of the others below.

Also the "MAIN MENU" h2 gets coloring with the primary color. That h2 had `opacity: 0.6` and it looked a bit too light to me, so I changed it to 0.7. (I also tried removing the opacity setting so it was as dark as the main banner and active item, but that looked bad.)

Note that if this is a good way to go, then #2008 should be closed unmerged.

Here are some screenshots. One of them has a set and problem number opened so you can confirm the coloring applies to nested things in the nav tree. "Achievements" is gray in one of them because the mouse was hovering over it when I took the screen shot.

Screen Shot 2023-05-30 at 11 50 25 AM

Screen Shot 2023-05-30 at 11 50 03 AM

Screen Shot 2023-05-30 at 11 49 23 AM

Screen Shot 2023-05-30 at 11 51 44 AM

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -193,7 +193,6 @@ $layout-divider-color: #aaa !default;
 		color: black;
 		font-weight: bold;
 		font-size: 0.9rem;
-		opacity: 0.7;
 		background-color: var(--bs-primary, #038);
 		color: var(--ww-primary-foreground-color, white);
 	}
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -193,7 +193,9 @@ $layout-divider-color: #aaa !default;
 		color: black;
 		font-weight: bold;
 		font-size: 0.9rem;
-		opacity: 0.6;
+		opacity: 0.7;
+		background-color: var(--bs-primary, #038);
+		color: var(--ww-primary-foreground-color, white);
 	}
 
 	.info-box {
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -3,13 +3,13 @@
 <h2 class="navbar-brand mb-0"><%= maketext('Main Menu') %></h2>
 <ul class="nav flex-column">
 	% unless ($restricted_navigation) {
-		<li class="list-group-item nav-item">
+		<li class="list-group-item list-group-item-primary nav-item">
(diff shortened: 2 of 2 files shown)


pstaabp:
Related question: Another thing that I've been playing with is that the links on the breadcrumbs for some of the themes are hard to tell apart from the end of breadcrumb (which is not a link)  I'm using a variation on the green theme that @Alex-Jordan posted above.  Is there a more general way we could handle this? Or tweak on the theme basis?

drgrice1:
I am seeing a contrast issue with the "MAIN MENU" text on all themes except "math4-yellow".

Alex-Jordan:
@pstaabp I put the `link-primary` class on the anchors, and at least in the blue theme, it makes a small difference. Not sure it's enough to get at what you are after though. I think to do that, you might need to wrap the last one in a span and put some class on that span.

Alex-Jordan:
> I am seeing a contrast issue with the "MAIN MENU" text on all themes except "math4-yellow".

You see it locally, or in my screenshots?

drgrice1:
I think that with the background color being the primary color, you should completely remove the opacity.  That looks much better to me, and solves the contrast issue.

Alex-Jordan:
I can do that. But it looked off to me to have the background color be fully opaque and therefore match the "MAA" background, with the WW logo background a lighter version. The effect was that instead of three rectangles, it was one rectangle (the WW logo) and then a jagged octogon. The dividing border line didn't totally kill that illusion for me. Do you see any of that when looking with the opacity removed?

drgrice1:
I did see that.

Alex-Jordan:
I removed the opacity. Now for example in green:

Screen Shot 2023-05-30 at 2 01 11 PM

I( tried moving it to 0.8 first, which fixed the contrast issue for blue, but not green.)

drgrice1 (approved):
This looks good.

It would be nice to figure out a better way to make the header and "Main Menu" more differentiated.  But @Alex-Jordan will probably figure something out at some point.

somiaj:
Now that this is on my live server, I do notice the following looks a bit odd. I remove the 'Courses' link from my menu as once a student is in a course, they don't really need to get out to a list of all courses.  This does make it so the menu header and selected item have the same color.

ww-menu

somiaj:
Note, I can edit my themes CSS to get things to look nicer here, just sharing the result.

PR #2011, opened 2023-05-30
Use column headers on user detail set dates table

Alex-Jordan:
This morning an instructor was confused why they were changing dates for a student on a set, but the "date wouldn't change". They didn't understand that they were seeing the set's global dates alongside the override they were applying to a student.  So I added column headers to the tables that are made for the dates. See pic.

Note that I do not think it is appropriate to split up the table into two columns of the larger table. It is appropriate for a user to be able to keyboard navigate quickly from the override input to the default date and back.

Screen Shot 2023-05-30 at 12 19 55 PM

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -573,8 +573,8 @@ sub fieldTable ($c, $userID, $setID, $problemID, $globalRecord, $userRecord = un
 				'tr',
 				$c->c(
 					$c->tag('td', colspan => '3', ''),
-					$c->tag('th', $c->maketext('User Value')),
-					$c->tag('th', $c->maketext('Class value'))
+					$c->tag('th', $c->maketext('User overrides')),
+					$c->tag('th', $c->maketext('Set values'))
 				)->join('')
 			)
 		);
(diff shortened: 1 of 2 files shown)


pstaabp:
I think this is a good idea, however, on the set detail page if you override a single set for one user, the UI is
image

I would be consistent with the headers with these pages--I'm trying to find if there are any other override pages as well. 

I like "User Overrides", but think "Set Defaults" is a bit confusing (as if there are some general default values for a set), and I'm not crazy about "Class Values". How about "Set Values"?

dlglin:
> I like "User Overrides", but think "Set Defaults" is a bit confusing (as if there are some general default values for a set), and I'm not crazy about "Class Values". How about "Set Values"?

I really want to say "Non-Overridden Value", but that's super clunky.  How about "Default Value" or "Global Value"?

Alex-Jordan:
I changed both to `User overrides` and `Set values`.

Danny's comment just came in. We can keep wordsmithing it.

somiaj:
> This morning an instructor was confused why they were changing dates for a student on a set, but the "date wouldn't change".

A place we have confusion here is wanting to edit a student's test version dates (after it has been started), but instead of overriding the version, they override the student's global or template test, which doesn't work.  Think it is worth adding a help popup on tests giving the user info that these only edit the test's global values, and any current test versions must be modified separately (also, maybe add a link to edit any current test versions).

pstaabp:
> A place we have confusion here is wanting to edit a student's test version dates (after it has been started), but instead of overriding the version, they override the student's global or template test, which doesn't work. Think it is worth adding a help popup on tests giving the user info that these only edit the test's global values, and any current test versions must be modified separately (also, maybe add a link to edit any current test versions).

One possibility is to add underneath each set name: 

image

an box with some information under both the global test and the version tests with a short explanation.  There's plenty of room here if we keep the comment succinct.

PR #2012, opened 2023-05-30
Another tweak to the pg htdocs url configuration.

drgrice1:
If `$pg_htdocs_url` is not set in `site.conf`, then pages load due to a fallback in `lib/Mojolicious/WeBWorK.pm` .  However, if a problem uses pg javascript the problem fails to load and there are lots of errors. This adds a fallback to the course environment as well so that those problems still load.

This stems from https://github.com/openwebwork/webwork2/pull/2000#issuecomment-1569040247.

### DIFF
--- a/lib/WeBWorK/CourseEnvironment.pm
+++ b/lib/WeBWorK/CourseEnvironment.pm
@@ -232,6 +232,9 @@ sub new {
 			keys %{ $self->{statuses} }
 	};
 
+	# Make sure that this is set in case it is not defined in site.conf.
+	$self->{pg_htdocs_url} //= '/pg_files';
+
 	# Fixup for courses that still have an underscore, 'heb', 'zh_hk', or 'en_us' saved in their settings files.
 	$self->{language} =~ s/_/-/g;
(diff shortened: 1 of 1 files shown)


pstaabp:
I couldn't get any errors on develop.  When you say "uses pg javascript", this means that a problem loads javascript on the PG side?  I'm looking at a scaffold problem, which loads scaffold.js and not seeing any errors.

drgrice1:
Are you sure that you have `$pg_htdocs_url` commented out (or removed) from site.conf?  Without this pull request and that not in site.conf I see several lines of "Use of uninitialized value in concatenation (.) or string at /opt/webwork/webwork2/lib/WeBWorK/Utils.pm line 2075." anytime I load any problem.  All problems load files from the pg htdocs location, so that should be the case.

pstaabp:
Hmmm.  I tried again and yes see all of the errors.  This fixes it.

pstaabp (approved):

Alex-Jordan (approved):

drdrew42 (approved):

PR #2014, opened 2023-05-31
Update the docker build with missing modules.

drgrice1:
When I added the LTI 1.3 authentication, I forgot to add the modules for that.

Remove the libcgi-pm-perl package.  That is not needed anymore!

Remove the libcrypt-ssleay-perl package as that is not needed.

Add the dependencies of Mojo::SQLite as that speeds up the cpan installation.

Don't install Mojolicious::Plugin::NotYAMLConfig as that is now part of Mojolicious directly, and so is included in the libmojolicious-perl pacakge which is already installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,11 +75,11 @@ RUN apt-get update \
 	libarray-utils-perl \
 	libc6-dev \
 	libcapture-tiny-perl \
-	libcgi-pm-perl \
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
I haven't set up docker to test, but looks like the needed packages are there now.

PR #2018, opened 2023-06-01
Fix for file manager directory dropdown menu.

somiaj:
When opening a new directory with the file manager, update the directory drop down menu to ensure the current directory is selected.


pstaabp:
I'm confused what the issue is.  I just created a new folder, typed int the name and the file Manager took me to that new folder.  The dropdown at the top of the file listing had the new folder name in the listing.  Is this what you are talking about?

drgrice1:
@pstaabp:  If you double click on a directory in the listing to change directories, the directory is changed, but the directory name in the dropdown at the top stays what it was.  If you use the dropdown directly to change directories, then it does change.

pstaabp:
I see now.  Nothing to do with making a new directory/folder.

drgrice1 (approved):

PR #2019, opened 2023-06-01
change use of 'versions' which is confused with quiz versions

Alex-Jordan:
This link is not about quiz/test versions, so I propose changing this text.

### DIFF
--- a/templates/ContentGenerator/Instructor/ProblemSetDetail.html.ep
+++ b/templates/ContentGenerator/Instructor/ProblemSetDetail.html.ep
@@ -158,8 +158,8 @@
 			</div>
 			<div class="col-md-6 mt-md-0 mt-2">
 				<%== maketext(
-					'Edit [_1] of set [_2].',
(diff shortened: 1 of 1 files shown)


templates/ContentGenerator/Instructor/ProblemSetDetail.html.ep
```diff
@@ -159,7 +159,7 @@
 			<div class="col-md-6 mt-md-0 mt-2">
 				<%== maketext(
 					'Edit [_1] of set [_2].',
-					link_to(maketext('individual versions') =>
+					link_to(maketext('individual user settings') =>
```
drgrice1, line 162: I think the wording needs to be changed a bit in the full sentence with this.  "Edit individual user settings of set ..." isn't quite right.  It should be "Edit individual user settings for set ..." I think.

drgrice1 (approved):
Looks good.

pstaabp (approved):
looks good.

PR #2020, opened 2023-06-01
Update the masthead layout.

drgrice1:
This removes the masthead from bootstrap's grid layout, and instead used display flex.  The div containing the webwork logo and sidebar menu button has a width of 20%, the same as the sidebar.  The institution logo is allowed to take up any space left from the login status to the right.

Since the sidebar and this div have the same width, the sidebar no longer overlaps with the institution logo part of the masthead giving a better appearance with the new coloring of the "MAIN MENU" navbar brand.

Note that on narrow screens there should be no difference from before.

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -58,11 +58,12 @@ $masthead-height: 70px !default;
 $layout-divider-color: #aaa !default;
 
 /* Banner */
-#masthead {
+.masthead {
 	position: fixed;
 	top: 0;
 	left: 0;
 	right: 0;
+	display: flex;
 	height: $masthead-height;
 	background-color: var(--bs-primary, #038);
 	border-bottom: 1px solid $layout-divider-color;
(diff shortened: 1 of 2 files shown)


somiaj (approved):
Looks good.  Would adding a max-width setting be worthwhile? The menu gets a bit too wide on large screens, and I think it would look better being a bit narrower.

drgrice1:
You can't just set max-width.  It would take more than that.  If you set max-width on the site-navigation and webwork-log, then there would just be white space between the site-navigation and the main content in the space where the site-navigation would have previously been.  This is because the site-navigation is set to a width of 20%, and the main content is set to a width of 80%.  If you set max-width on the site-navigation, then it would limit its width, but the main content would still only take 80% of the width the same as it currently does.

The only real option would be to go to a fixed width site-navigation instead, and that may not be as desirable as you think.

In general, go look around at many of the sites out there.  Extremely wide browser windows always end up having a lot of wasted space.  Just look at the Github page you are currently viewing.  Do you see all of the empty space to the left and right?  I just looked through several sites I happen to have open now, and all of them are like this.

somiaj:
Thanks for the explanation.

drgrice1:
However, it would be possible to change the percentages at various breakpoints.

PR #2021, opened 2023-06-02
Change the site-navigation to use a fixed width.

drgrice1:
The width is 250px.  That width could be tweaked if this is what we want.  Note this only is in effect if the browser window width is more than the small breakpoint (768px).

This builds on top of #2020.  There will be another pull request with an alternate approach for comparison shortly.

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -56,6 +56,7 @@ table caption {
 
 $masthead-height: 70px !default;
 $layout-divider-color: #aaa !default;
+$site-nav-width: 250px !default;
 
 /* Banner */
 .masthead {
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Of the three, this one is my preference.

Alex-Jordan (approved):

drgrice1:
Contrary to my comment in https://github.com/openwebwork/webwork2/pull/2020#issuecomment-1573051878, I think I like this one the best also.

pstaabp (approved):
Much better.  I use a wide external monitor and there was a lot of extra space when maxing out the window (which I do with the pg problem editor).

Alex-Jordan:
I think overall, @pstaabp , @drgrice1 , and I have each expressed preference for this one. Is there any objection to me merging it? (And closing the other two.)

drdrew42 (approved):
Count me in!

drgrice1:
Sounds good to me.

PR #2022, opened 2023-06-02
Decrease the width of the site navigation when above the xl breakpoint.

drgrice1:
This switches the width of the site navigation to 15% and the content to 85% when the browser window width is above the extra large breakpoint (1200px).  Below the extra large breakpoint the width is still 20% for the site navigation, and 80% for the content.  Of course below the small breakpoint, things are still the same.

This builds on top of #2020 also.  Compare this to #2021.

@dlglin:  Take a vote.  #2021 or #2022?  Which do we prefer?  Or neither?

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -58,11 +58,12 @@ $masthead-height: 70px !default;
 $layout-divider-color: #aaa !default;
 
 /* Banner */
-#masthead {
+.masthead {
 	position: fixed;
 	top: 0;
 	left: 0;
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I really don't like the effect here when resizing the window. But of course real users aren't resizing windows much.

Still I think I prefer #2021 to this one. Now I will try plain old #2020.

pstaabp:
Reading these PRs out of order.  I prefer #2021.

Alex-Jordan:
Closing in favor of #2021.

PR #2023, opened 2023-06-02
Archived course location

somiaj:
Moves the archived courses into the admin course's `archives` directory so they can be accessed via the file manager.   Adds a direct link to the `archives` directory above the archived courses list in the main admin course page.  Adds help statements stating where the archives are located.  Removes some old text describing the internals of the archive process which isn't needed.  Moves the `Archive This Course` link from the instructor menu to a button inside the file manager.

I was considering also looking in the old location when listing archives and when unarchive archives, use the old location if the course doesn't exist in the new location to make it so old archives are still shown, just not accessible via the file manager.  I have left that off for now, part of the upgrade process should include moving the old archive files. Just mentioning the thought if others think it is worth pursing.

### DIFF
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -132,7 +132,8 @@ Lists the courses which have been archived (end in .tar.gz).
 
 sub listArchivedCourses {
 	my ($ce) = @_;
-	my $coursesDir = "$ce->{webworkDirs}->{courses}/admin/archives";
+	my $coursesDir = "$ce->{webworkDirs}{courses}/admin/archives";
+	surePathToFile($ce->{webworkDirs}{courses}, "$coursesDir/test");    # Ensure archives directory exists.
 	return grep {m/\.tar\.gz$/} readDirectory($coursesDir);
 }
 
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -767,6 +767,7 @@ sub archiveCourse {
 		$archive_path = $options{archive_path};
 	} else {
 		$archive_path = "$ce->{webworkDirs}{courses}/admin/archives/$courseID.tar.gz";
+		surePathToFile($ce->{webworkDirs}{courses}, $archive_path);
 	}
 
 	# fail if the source course does not exist
@@ -823,8 +824,6 @@ sub archiveCourse {
 	##### step 3: cleanup -- remove database dump files from course directory #####
 
 	unless (-e $archive_path) {
-		# Make sure any missing directories are creates.
-		surePathToFile($ce->{webworkDirs}{courses}, $archive_path);
 		unless (move($tmp_archive_path, $archive_path)) {
 			unlink($tmp_archive_path);    #clean up
 			croak "Failed to rename archived file to '$archive_path': $!";
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -37,7 +37,7 @@ use File::Spec;
 use WeBWorK::CourseEnvironment;
 use WeBWorK::DB;
 use WeBWorK::Debug;
-use WeBWorK::Utils qw(runtime_use readDirectory pretty_print_rh);
+use WeBWorK::Utils qw(runtime_use readDirectory pretty_print_rh surePathToFile);
 
 our @EXPORT_OK = qw(
 	listCourses
@@ -132,7 +132,7 @@ Lists the courses which have been archived (end in .tar.gz).
 
 sub listArchivedCourses {
 	my ($ce) = @_;
-	my $coursesDir = $ce->{webworkDirs}->{courses};
+	my $coursesDir = "$ce->{webworkDirs}->{courses}/admin/archives";
 	return grep {m/\.tar\.gz$/} readDirectory($coursesDir);
 }
 
@@ -721,9 +721,10 @@ sub deleteCourse {
  courseID  => $courseID,
  ce        => $ce,
 
-Creates a gzipped tar archive (.tar.gz) of the course $courseID in the WeBWorK
-courses directory. Before archiving, the course database is dumped into a
-subdirectory of the course's DATA directory.
+Creates a gzipped tar archive (.tar.gz) of the course $courseID and places it in the
+archives directory of the admin course or in the location given in the optional archive_path
+option.  Before archiving, the course database is dumped into a subdirectory of the course's
+DATA directory.
 
 Only files and directories stored directly in the course directory are archived.
 The contents of linked files is not archived although the symbolic links
@@ -751,10 +752,10 @@ sub archiveCourse {
 	# grab some values we'll need
 	my $course_dir = $ce->{courseDirs}{root};
 
-	# tmp_archive_path is used as the target of the tar.gz operation
-	# After this is done the final tar.gz file is moved either to the course directory
-	# or the course/myCourse/templates   directory (when saving individual courses)
-	# this prevents us from tarring a directory to which we have just added a file
+	# tmp_archive_path is used as the target of the tar.gz operation.
+	# After this is done the final tar.gz file is moved either to the admin course archives directory
+	# course/admin/archives or the supplied archive_path option if it is present.
+	# This prevents us from tarring a directory to which we have just added a file
 	# see bug #2022 -- for error messages on some operating systems
 	my $uuidStub         = create_uuid_as_string();
 	my $tmp_archive_path = $ce->{webworkDirs}{courses} . "/${uuidStub}_$courseID.tar.gz";
@@ -764,7 +765,7 @@ sub archiveCourse {
 	if (defined $options{archive_path} && $options{archive_path} =~ /S/) {
 		$archive_path = $options{archive_path};
 	} else {
-		$archive_path = $ce->{webworkDirs}{courses} . "/$courseID.tar.gz";
+		$archive_path = "$ce->{webworkDirs}{courses}/admin/archives/$courseID.tar.gz";
 	}
 
 	# fail if the source course does not exist
@@ -821,6 +822,8 @@ sub archiveCourse {
 	##### step 3: cleanup -- remove database dump files from course directory #####
 
 	unless (-e $archive_path) {
+		# Make sure any missing directories are creates.
+		surePathToFile($ce->{webworkDirs}{courses}, $archive_path);
 		unless (move($tmp_archive_path, $archive_path)) {
 			unlink($tmp_archive_path);    #clean up
 			croak "Failed to rename archived file to '$archive_path': $!";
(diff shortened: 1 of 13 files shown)


Alex-Jordan:
Note you have to manually add the new `archives` folder to your existing admin course when you check out this branch. Or else you get errors visiting thee admin course.

The admin course is usually created with the `addcourse` script, which is not affected by this PR. Is the right thing to do, edit that script, and if the course being created is 'admin' then also create this 'archives' folder? Or maybe you can automate its creation when it is not found?

If it is not automatically created, it will need to be in upgrade instructions to add this new folder.

Alex-Jordan:
Hmm, also upgrade instructions may need to include moving all existing archive files into this new folder.

Alex-Jordan:
While in the admin course's file Manager, i clicked the new button to archive this course, and it gives me an error. Maybe this button should just not be an option in the admin course.

somiaj:
I thought I checked for that, but seems I only checked for the directory when creating a new archive, not when listing archives. I'll fix that.

Alex-Jordan:
Aside from the things I mentioned, the functionality of this seems good and it will b a nice thing for admins.

somiaj:
> While in the admin course's file Manager, i clicked the new button to archive this course, and it gives me an error. Maybe this button should just not be an option in the admin course.

What error did you get?  It seems to work just fine for me.

somiaj:
It should also now work just fine if the `archives` directory doesn't exist, and it will create a new directory when saving an archive if it doesn't exist.

Alex-Jordan:
> What error did you get?

> Failed to generate course archive: Failed to archive course directory '/opt/webwork/courses/admin' with command '2>&1 /bin/tar -C /opt/webwork/courses/admin/.. -czf /opt/webwork/courses/aed425e4-010b-11ee-96f3-9329d359ce1b_admin.tar.gz admin' (exit=1 signal=0 core=0): /bin/tar: admin/courses/aed425e4-010b-11ee-96f3-9329d359ce1b_admin.tar.gz: file changed as we read it at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm line 59.

somiaj:
@Alex-Jordan Is this the server you have mount --bind courses into the admin course?

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -132,8 +132,8 @@ Lists the courses which have been archived (end in .tar.gz).
 
 sub listArchivedCourses {
 	my ($ce) = @_;
-	my $coursesDir = $ce->{webworkDirs}->{courses};
-	return grep {m/\.tar\.gz$/} readDirectory($coursesDir);
+	my $coursesDir = "$ce->{webworkDirs}->{courses}/admin/archives";
```
drgrice1: There is an unnecessary dereferencing arrow here.  Please change this to:
```suggestion
	my $coursesDir = "$ce->{webworkDirs}{courses}/admin/archives";
```

somiaj:
I decided to go with creating the `archives` directory if it is missing (when listing the archived courses, this way the directory should almost always exists after accessing the admin course, and that way if an admin clicks the "Download/upload archives" link, the directory exists and they can upload archives without extra work.

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -751,21 +753,19 @@ sub archiveCourse {
...
-	my $archive_path;
-	if (defined $options{archive_path} && $options{archive_path} =~ /S/) {
-		$archive_path = $options{archive_path};
-	} else {
-		$archive_path = $ce->{webworkDirs}{courses} . "/$courseID.tar.gz";
-	}
+	my $archive_path =
+		$options{admin_course}
+		? "$ce->{webworkDirs}{courses}/admin/archives/$courseID.tar.gz"
+		: "$course_dir/templates/$courseID.tar.gz";
```
drgrice1, line 768: This should not be changed in this way.  You have replaced a more versatile method of setting the archive path with a very limited approach.  Furthermore, this made it so that you needed to make other unnecessary changes elsewhere in the code.

One thing that you do not seem to realize is that this module is designed to be used with scripts.  A script may want to call this method and set the archive path to something other than the hard coded ones you are forcing here.  In fact I have scripts that do this.

drgrice1 (changes requested):
I was going to do a more thorough review, but saw this and realized you need to be notified of this immediately.  See my comment in the code.

drgrice1:
You also definitely need to remove the "Archive Course" option from the file manager in the admin course.

lib/WeBWorK/Utils/CourseManagement.pm
somiaj: I didn't mean to remove the ability to set the `archive_path` directly, but missed that I did (for some reason I thought this was tested for later).  All I mean to was to change the default to putting the archive in the courses template directory unless the option to place it in the admin course or archive_path option was provided.  But now I do see I did remove that ability (probably something got lost in a few various attempts).  Anyways, I'll revert this, thanks.

somiaj: Okay reverted and just changed the default path.

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -132,7 +132,8 @@ Lists the courses which have been archived (end in .tar.gz).
 
 sub listArchivedCourses {
 	my ($ce) = @_;
-	my $coursesDir = $ce->{webworkDirs}->{courses};
+	my $coursesDir = "$ce->{webworkDirs}{courses}/admin/archives";
```
drgrice1, line 135: This variable name is not reflective of what it is with this change.  It should be renamed to something like `$archivesDir`.

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -821,6 +823,8 @@ sub archiveCourse {
 	##### step 3: cleanup -- remove database dump files from course directory #####
 
 	unless (-e $archive_path) {
+		# Make sure any missing directories are creates.
+		surePathToFile($ce->{webworkDirs}{courses}, $archive_path);
```
drgrice1, line 827: This is going to have to be rethought.

My script uses a temporary directory in the system tmp directory (/tmp/archiveXXXXXX) for the archive location.  That is not inside of the courses directory, and as a result, when the script is called the directory /opt/webwork/courses/tmp/archiveXXXXXX is created.

Any script that tries to use this method, and uses a location outside of the courses directory will have this happen.  Furthermore, If the user running the script doesn't have permission to write to the courses directory then the script will fail.  My script happens to be run as the same user as the webwork server so the directory is created, but the script otherwise works.

somiaj: One possible fix is to only run this if `$options{archive_path}` is not set, and revert to assuming the path exists if is supplied to the method.

drgrice1: If the directory is going to be created in this method, then that would work.  Do it on line 770 (currently in this pull request).

Another way would be to check that the given path starts with `$ce->{webworkDirs}{course}` and only try to create directories in that case.

You could also not try to create the directories here at all, and only do this in the one time that this method is called in which it might be needed.  That is in the `do_archive_course` method of CourseAdmin.pm.

somiaj:
I'll fix the issues with permissions I overlooked when creating the admin menu initially.

lib/WeBWorK/Utils/CourseManagement.pm
somiaj: I went with your last suggestion, only testing in the CourseAdmin `do_archive_course` method.

somiaj: Though I guess if someone calls `archiveCourse` without `archive_path`, there is no check and it could error out.  This isn't being done, maybe I'll revert back to my first suggestion just to allow that option.

drgrice1: Since the path on line 769 of CourseManagement.pm is relative to the courses directory anyway, it wouldn't hurt to create the directory there, and that would cover any call to the `archiveCourse` method.

somiaj: Updated the last commit to do just that. Thanks.

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -37,7 +37,7 @@ use File::Spec;
 use WeBWorK::CourseEnvironment;
 use WeBWorK::DB;
 use WeBWorK::Debug;
-use WeBWorK::Utils qw(runtime_use readDirectory pretty_print_rh);
+use WeBWorK::Utils qw(runtime_use readDirectory pretty_print_rh surePathToFile);
```
drgrice1, line 40: You could remove `pretty_print_rh` from this list.  That method is not used in this file, and so should not be imported.

drgrice1 (approved):
This is looking good now.

somiaj:
I don't think we need to scan for multiple extensions, just use the one the create archive button uses.  Creating an archive with the file manager will most likely not be a course archive (since it can be any file system), so I don't think we need to test for that.

I do think we should change the file manager default here though, since it will be assumed that archives will be uploaded and we don't want to extract those.

somiaj:
I will update the suggestions with the file manager (default to the root admin directory and uncheck the unpack archive button by default) when I am back home.

somiaj:
I updated the file manager link in the admin course so it defaults to the root "admin" course directory and doesn't check the unpack and autodelete archives by default.  I also removed the help comments about moving up one level to find the archives directory since it isn't needed anymore.

I also updated the unpack and autodelete checkboxes to remember their previous setting when moving through the directories with the file manager (so they stay unchecked once they are unchecked and visa versa).  @drgrice1 want to double check I did this correctly (my tests show it works, but just double checking my solution is the best for this).

drgrice1:
> I also updated the unpack and autodelete checkboxes to remember their previous setting when moving through the directories with the file manager (so they stay unchecked once they are unchecked and visa versa). @drgrice1 want to double check I did this correctly (my tests show it works, but just double checking my solution is the best for this).

The way you did it looks good.

somiaj:
One issue I just ran across with ngnix was its upload size is too small to upload a `2.7M` archive by default, so I had to configure the proxy to allow a bigger upload.  This is a server configuration issue, though wonder if it would be worth mentioning in the install guide that to upload larger archives server configuration might need to be adjusted.

drgrice1:
It would probably be a good idea to mention that.  It seems the setting is `client_max_body_size`.

Note that Mojolicious also has a setting related to this that may need to be changed.  Mojolicious limits request sizes to 16MiB, and responses to 2 GiB.  The response size is probably sufficient, but I have seen issues with the request size limitation in other Mojolicious apps, and had to increase it.  You can do that with the `MOJO_MAX_MESSAGE_SIZE` environment variable (although in my app I did it directly with `$app->max_request_size(0)`).  Note that a setting of `0` for that means no limit.

pstaabp (approved):
This looks good now.

drgrice1:
This has two approvals, and looks good.  So I am merging.

PR #2024, opened 2023-06-02
Tweak logout so that authentication is still attempted

drgrice1:
Previously I made it so that if the logout module was called, it would always skip authentication and just log out.  The problem with that is that then the user information is not filled in, and so logs don't have this information anymore.

So instead of skipping authentication entirely when logging out, still attempt to verify authentication.  However, if authentication fails, then don't switch to the login module and logout anyway.

This still has the benefit that you don't need to login to logout if the session has expired.  Note that in this case the logs won't have user information anyway, and there is nothing that can be done about that. But in the case that the session has not expired, now the logs will have user information.

To test this watch your course's login.log file when you log out. Currently it will alway show something like "[Fri Jun 02 07:10:13 2023] LOGGED OUT user_id= login_type= credential_source= host=127.0.0.1 port=43854 UA=Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0".  There is no user_id, login_type, or credential_source regardless of the session status when you click logout.  With this pull request that will still be the case if the session has expired, but if the session is still valid, then you will see that information filled in.

### DIFF
--- a/lib/WeBWorK.pm
+++ b/lib/WeBWorK.pm
@@ -208,66 +208,66 @@ async sub dispatch ($c) {
 		debug("(here's the DB handle: $db)\n");
 		$c->db($db);
 
-		# Don't check authentication if the user is logging out.
-		if ($displayModule ne 'WeBWorK::ContentGenerator::Logout') {
-			if ($authen->verify) {
-				# If this is the first phase of LTI 1.3 authentication, then return so its special content generator
-				# module will render and submit the login repost form.  This does not contain the neccessary information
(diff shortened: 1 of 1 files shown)


drgrice1:
If you hide whitespace changes, it makes it really easy to see the actual changes in this pull request.

Alex-Jordan (approved):
I tested and I see the before and after as described.

drdrew42 (approved):

Issue #2016, opened 2023-06-01
answers.log contains incomplete or malformed lines


PhilipLoewen:
Here are 3 consecutive lines from the file answers.log in my course:
```
[Thu Mar 09 21:48:44 2023] |ZI7UJNQDCR03|WW06|18|100    1678427324      -ky
x+32)/2
[Thu Mar 09 21:48:44 2023] |7XTJHYSEKZ08|WW06|13|100000 1678427324      tan(arctanx)    [0,pi]
```
Here are 3 others:
```
[Thu Mar 09 21:49:11 2023] |C93SRBEKNW04|WW06|19|1110   1678427351      7       1       3       3.67
8(T-22) 106.3
[Thu Mar 09 21:49:12 2023] |BA05O2XZTR07|WW06|13|111000 1678427352      x       (-inf,inf)      tan(arctanx+arctan(-5))
```
In both cases there is something wrong with the line in the middle. My answers.log file has 1326 lines that don't have the expected structure, out of 3052734 lines in total. For many of these, including the two examples shown above, the lines before and after show timestamps that are very close together (sometimes identical).

Unfortunately I do not have the knowledge or skill to help fix this. All I can do is report it. Thanks for reading.

I harvested the file named "answers.log" from the File Manager in the user-facing web interface. To make sure I got a valid copy  I downloaded the file twice. Both copies (size 301MB) show the same md5 hash.

This is WeBWorK © 1996-2022 | theme: math4 | ww_version: 2.17 | pg_version 2.17

Alex-Jordan:
It may be useful to compare against this course's transaction log from the server. Ask your server admin for the file `/opt/webwork/webwork2/logs/course_name_transaction.log` where of course `course_name` is the name of the course. Just to see if there was anything strange about the student's input in these instances; or if the student inputs here have something in common; or if those lines in the transaction log are also corrupted.

PhilipLoewen:
The IT crew have sent me `*_transactions.log`. I expect each line in this file to start with a timestamp in square brackets. This is true for most of the 1,579,982 lines, but there are around 700 exceptions. The middle line of the following 3 provides a sample; I provide discussion further below:
```
[Thu Jan 12 11:09:23 2023] 1	WW00	Xaaaaaaaaaa7	set101-WW0/WW0explanation.pg	1	-1	1873	1	1	["AnSwEr0001","I have read and understood this explanation."]	1	0
[...]

PR #2025, opened 2023-06-02
Obtain an exclusive lock on log files when writing logs

drgrice1:
More is probably needed with this as there are probably other files that multiple server processes could be attempting to write to at the same time.  Finding these would take careful review of the code.

This is to address issue #2016.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -32,6 +32,7 @@ use Date::Parse;
 use Date::Format;
 use File::Copy;
 use File::Spec::Functions qw(canonpath);
+use Fcntl qw(:flock);
 use Time::Zone;
 use MIME::Base64 qw(encode_base64 decode_base64);
 use Errno;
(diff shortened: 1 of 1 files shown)


drgrice1:
Note that the Fcntl module used by this is part of the Perl core.  So I don't think it needs to be added to check_modules.

pstaabp:
It seems like the best test for this might be to merge it and for those with a fair number of users to check the log files. I think trying to test on a development machine might have probability near zero to reproduce.

somiaj:
> It would also be good to keep an eye on any performance hits resulting from this, particularly in clustered environments where these log entries could be written to a network location.

I was also wondering if an i/o bottle neck here could cause a active server to slow down as each worker has to wait its turn to write its log.

Would pushing log writing off to minion be a possibility (though this would then make minion mandatory).  Though this might put a layer between the process and logs making debugging issues harder.

drgrice1:
@somiaj:  This certainly won't cause a bottleneck.  It is a quick file open, write a few lines, and close.

drgrice1:
Pushing this off to minion is silly.  This is not a time consuming nor resource intensive task.

drgrice1:
@dlglin:  I doubt this will cause any performance hits.  This is exactly the same code that Mojo::Log uses to handling log writing, and the is certainly designed to fit in with Mojolicous non-blocking behavior.

Alex-Jordan:
My school's production server has many users (55 current live courses), with some instructors even giving quizzes with simultaneous start times. I do not see this issue there (I only looked at the transaction logs, searching for lines that do not begin with `[`. That includes searching transaction logs from past terms and years.)

The same with the Runestone hosting server. And that has twice as many live courses, including one school with an emporium model where hundreds of students are simultaneously active at regular times each week.

So I'm not sure what kind of testing I could do to potentially reproduce the issue.

Alex-Jordan (approved):
I'm approving this without testing it or closely examining its code. I am thinking that we should just merge it and use what time we have left before the release to keep an eye out for issues. And it is probably good, since @drgrice1 has experience using file locking.

pstaabp (approved):
Agree with @Alex-Jordan.  Will merge.

PR #2027, opened 2023-06-02
Admin nav tweaks

Alex-Jordan:
The commit messages describe what this does.

For removing 'Courses', I grew the size of "MAIN MENU" to 1.3 rem and brough back the opacity setting, but at 0.85. It gets AAA contrast rating this way, even in green. And for me at least it is visually clear that "MAIN MENU" is separate from the first item. But it may need more tweaking.

Screen Shot 2023-06-02 at 10 47 27 AM

Screen Shot 2023-06-02 at 10 48 36 AM

Of course things will look a little different in combination with #2020, #2021, or #2022

BTW I tried a border line under MAIN MENU and to me it looked bad...

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -192,8 +192,9 @@ $layout-divider-color: #aaa !default;
 		padding: 0.5rem 0 0.25rem 1rem;
 		color: black;
 		font-weight: bold;
-		font-size: 0.9rem;
+		font-size: 1.3rem;
 		background-color: var(--bs-primary, #038);
+		opacity: 0.85;
 		color: var(--ww-primary-foreground-color, white);
 	}
 
--- a/templates/ContentGenerator/Base/admin_links.html.ep
+++ b/templates/ContentGenerator/Base/admin_links.html.ep
@@ -1,8 +1,5 @@
 <h2 class="navbar-brand mb-0"><%= maketext('Admin Menu') %></h2>
 <ul class="nav flex-column">
-	<li class="list-group-item list-group-item-primary nav-item">
-		<%= link_to maketext('Courses') => 'root', class => 'nav-link' %>
-	</li>
 	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('options') %></li>
 	<li class="list-group-item nav-item">
 		<%= $makelink->(
--- a/templates/ContentGenerator/Base/admin_links.html.ep
+++ b/templates/ContentGenerator/Base/admin_links.html.ep
@@ -7,14 +7,14 @@
 	<li class="list-group-item nav-item">
 		<%= $makelink->(
 			'set_list',
-			text   => maketext('Course Administration'),
+			text   => maketext('Course Listings'),
 			active => !param('subDisplay') && $c->url_for =~ /admin$/ ? 1 : 0
 		) %>
 	</li>
 	% for (
 		% [
 		% 	'add_course',
-		% 	maketext('Add Course'),
+		% 	maketext('Add Courses'),
 		% 	{
 		% 		add_admin_users      => 1,
 		% 		add_config_file      => 1,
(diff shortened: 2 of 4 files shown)


somiaj:
Should probably update the main admin help page, `templates/HelpFiles/admin_links.html.ep` to match the naming and order of the menu changes.  I don't think the other help pages will need any updates for this.

somiaj:
Nevermind, I missed you updated those already. I looked over the files and didn't register.

somiaj:
Though with the removal of the `Courses` link, think removing the first `WeBWorK` link in the bredcrumbs menu would also be reasonable (I do this as well in my setup).

drgrice1:
@Alex-Jordan:  I added a pull request to this branch with some tweaks for your consideration, and further discussion.

Alex-Jordan:
I removed "MAIN MENU" but haven't pushed yet. Here's a picture:

Screen Shot 2023-06-03 at 3 09 54 PM

Are there any objections, before I push that commit here?

> removing the first WeBWorK link in the bredcrumb

While I'm slashing things, any reason to keep breadcrumb navigation at all? Does it add anything that the sidebar navigation is not already doing with indentation? (Maybe that indentation could be increased though.)

Screen Shot 2023-06-03 at 3 12 14 PM

somiaj:
> While I'm slashing things, any reason to keep breadcrumb navigation at all? Does it add anything that the sidebar navigation is not already doing with indentation? (Maybe that indentation could be increased though.)

As far as I can tell it is the only way to get out of a test (since there is no side menu's in tests).  I also use it occasionally after I have closed the menu to get around.  It can also help users understand where they are at, though outside of the no other menu during tests, it isn't something I have feelings about being there or not.

So if you were to remove the breadcrumb menu, something should be done about tests.

drgrice1:
I think the breadcrumb navigation should stay.

Alex-Jordan:
That's a good reminder. But is thee test template separate from the regular ("system") template? I haven't looked but I'm guessing nothing would need too be done other than leave the test template alone.

Alex-Jordan:
> I think the breadcrumb navigation should stay.

OK. Do you have an opinion on @somiaj 's suggestion to cut the root listing from breadcrumb navigation?

drgrice1:
Tests do not use a separate template anymore.

I really think the breadcrumb should stay.

drgrice1:
@somiaj have enough reason. Having it there gives an indication of where you are. It is useful, and I don't think it needs to be cut.

Alex-Jordan:
Alright, not touching breadcrumbs. Do the screenshots with "MAIN MENU" gone stir up any reasons I should hit the brakes on that? Of course it would look a bit different with #2020, #2021, or #2022 merged in.

Alex-Jordan:
I went ahead and pushed so everyone can see the code changes.

somiaj:
@drgrice1 You didn't answer the question about your thoughts on my suggestion on if the breadcrumb menu root location should be `WeBWorK` or the `CourseID` of the course.  I remove the `WeBWorK` link in my classes because to me the root location should be the course not the server (Since most students aren't in multiple courses I don't see a reason users need such visible links to the main courses list page).

drgrice1:
@somiaj:  I am fine with that.  I have no strong opinion on that either way though.

pstaabp:
Overall, I like this.  If we completely remove a way to get to the list of courses, then I'm wondering as an instructor with multiple courses flipping back and forth between them.  

About the breadcrumbs, could we have the top level for a student be the CourseID (and the link goes to the homework sets), whereas for instructors, the top level is courses (or WeBWorK) as it is currently?

pstaabp:
> We are not removing the WeBWorK logo which is a link to the list of courses.

Good point. Although a bit of a hidden link.

Alex-Jordan:
The "Course Listings" (formerly "Course Administration") also gives you links to all the live courses. I can't really think of a reason it's not just as good as the main page. Except it is also showing you the hidden courses.

drgrice1:
> The "Course Listings" (formerly "Course Administration") also gives you links to all the live courses. I can't really think of a reason it's not just as good as the main page. Except it is also showing you the hidden courses.

However, this is only available for admin users in the admin course.  It is not generally available for instructors of multiple courses that are not admin users in the admin course.  Also, there are those that don't use WeBWorK solely through an LMS.  In that case the course listing page also needs to be available for students.

Alex-Jordan:
There was a merge conflict here that I tried to resolve with GitHub tools, but it did not do the indentation with tabs. sstand by while I update things locally and.

> However, this is only available for admin users in the admin course. It is not generally available for instructors of multiple courses that are not admin users in the admin course. Also, there are those that don't use WeBWorK solely through an LMS. In that case the course listing page also needs to be available for students.

Are you saying that I should bring back "Courses"? Or is the breadcrumb with "WeBWorK" root enough? Either way I think I will not touch the breadcrumb at all in this PR, not even to remove the "WeBWorK" root.

drgrice1:
No, I am just saying that the "Course Listings" of the admin course is not a valid substitute for the general course listings page.

Alex-Jordan:
OK, this is back ready for review. If there is any change to breadcrumb, let's keep that separate.

Here are some screenshots now that #2021 is merged.
Screen Shot 2023-06-06 at 10 07 54 AM
Screen Shot 2023-06-06 at 10 07 30 AM

pstaabp:
Overall, this looks good.  I think there is still the issue of if (in a regular course) the top of the breadcrumb should be `WeBWorK` or `courseID`? 

I think I'm leaning toward `courseID`, but don't feel too strongly.

Alex-Jordan:
> I think there is still the issue of if (in a regular course) the top of the breadcrumb should be WeBWorK or courseID?

I'd like to leave that out of this, even though it was discussed in this thread. It should be easy for an independent PR though if desired.

taniwallach:
I also think breadcrumbs should stay. Among everything else, they act as a WCAG approved way of helping make location in the site known: https://www.w3.org/WAI/WCAG21/Techniques/general/G65

pstaabp (approved):
Looks good.

PR #2028, opened 2023-06-02
Admin menu perm check

somiaj:
This updates the `create_and_delete_courses` default permission to be `admin` and deletes an unused `fix_courses_database` permission (the upgrade tool just checks `create_and_delete_courses`).  I also don't show any of the admin or instructor tools in the admin menu unless the `create_and_delete_courses` menu is there.

This address the issue mentioned in https://github.com/openwebwork/webwork2/issues/2009#issuecomment-1573859258, though `professor` (and to some lesser extent `ta`) can still access the instructor tools if they type out the url in the admin course.  Probably still best to not have any users except students and admins in this course.  Another option is to setup a new permission hash for the admin course that will then lock professors out of the standard instructor tools too.

I have built this on top of #2027 which put the user settings as the top link instead of doing any modifications of the menu here.

### DIFF
--- a/templates/ContentGenerator/Base/admin_links.html.ep
+++ b/templates/ContentGenerator/Base/admin_links.html.ep
@@ -1,54 +1,55 @@
 <ul class="nav flex-column">
 	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('options') %></li>
-	<li class="list-group-item nav-item">
-		<%= $makelink->(
-			'set_list',
-			text   => maketext('Course Listings'),
-			active => !param('subDisplay') && $c->url_for =~ /admin$/ ? 1 : 0
-		) %>
-	</li>
-	% for (
-		% [
-		% 	'add_course',
-		% 	maketext('Add Courses'),
-		% 	{
-		% 		add_admin_users      => 1,
-		% 		add_config_file      => 1,
-		% 		add_dbLayout         => 'sql_single',
-		% 		add_templates_course => $ce->{siteDefaults}{default_templates_course} || ''
-		% 	}
-		% ],
-		% [ 'rename_course',        maketext('Rename Courses') ],
-		% [ 'delete_course',        maketext('Delete Courses') ],
-		% [ 'archive_course',       maketext('Archive Courses') ],
-		% [ 'unarchive_course',     maketext('Unarchive Courses') ],
-		% [ 'upgrade_course',       maketext('Upgrade Courses') ],
-		% [ 'hide_inactive_course', maketext('Hide Courses') ],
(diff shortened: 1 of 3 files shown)


somiaj:
I added one possible way I could think of to deal with professor being able to access instructor tools in the admin course by typing in a url.  In the permission check, I elevate any permission requirement greater than student to be the same as `create_and_delete_courses`.  This locks a bit more of the course tools (which aren't really used) in the admin course away from anyone except admins.

templates/ContentGenerator/Base/admin_links.html.ep
```diff
@@ -1,36 +1,35 @@
 <h2 class="navbar-brand mb-0"><%= maketext('Admin Menu') %></h2>
 <ul class="nav flex-column">
-	<li class="list-group-item list-group-item-primary nav-item">
-		<%= link_to maketext('Courses') => 'root', class => 'nav-link' %>
-	</li>
-	<li class="list-group-item list-group-item-primary nav-item">
+	<li class="list-group-item list-group-item-primary nav-item"><%= $makelink->('options') %></li>
+	% last unless ($authz->hasPermissions($userID, 'create_and_delete_courses'));
```
drgrice1, line 4: You are not going to be able to use `last` here.  You will need to use an `if .. then` construct here.  Otherwise this will result in an un-ended `ul` tag.

somiaj: Thanks for catching that.

somiaj:
@pstaabp I updated this but don't merge, still need to wait on #2027 to be merged which this is built on top of, and I'll have to rebase and fix conflicts again at that time.  This probably shouldn't go in first because of moving the menu entries around in the other PR.

Alex-Jordan:
I'll review this once #2027 is merged.

pstaabp:
Just merged #2027.  @somiaj as you noted there are conflicts and now can review.

somiaj:
Updated.  Note that some how two links for the `File Manager` ended up in the admin menu in one of the other PRs, so this also removes the second file manager link.

drgrice1 (approved):
This looks good.  At first I thought it was a bit odd to do this sort of manipulation with permissions in the `hasPermissions` method of `Authz.pm`, but I can't think of a better way to code the fact that permissions in the admin course are just different.

It seems that the "File Manager" link got doubled somehow.  Probably some merge conflict resolution error I am guessing.

somiaj:
@drgrice1 The only other way I could think of making the admin permissions different is create a second permissions hash just for the admin course, which I thought was overkill, as I don't think such fine grain control is needed.

PR #2031, opened 2023-06-05
Show student evaluated answers for a fallback in the attempts table

drgrice1:
When the `$pg{options}{showEvaluatedAnswers}` option is false and parsing of the student answer fails, then show the evaluated answer in the "Answer Preview" column instead of showing nothing.  This is what would be shown in the "Entered" column if `$pg{options}{showEvaluatedAnswers}` were true.

### DIFF
--- a/lib/WeBWorK/HTML/AttemptsTable.pm
+++ b/lib/WeBWorK/HTML/AttemptsTable.pm
@@ -254,12 +254,16 @@ sub formatAnswerRow ($self, $rh_answer, $ans_id, $answerNumber) {
 	);
 
 	return $c->c(
-		$self->showAnswerNumbers   ? $c->tag('td', $answerNumber)                                               : '',
-		$self->showAttemptAnswers  ? $c->tag('td', dir => 'auto', $self->nbsp($answerString))                   : '',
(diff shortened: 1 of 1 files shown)


drgrice1:
This is not a bug.  This is a feature, and it has nothing to do with MathQuill.  Student answers fail to parse for many reasons.  The simplest way to make a student answer fail to parse is to enter a variable in an answer that is not in the context.  Typically `z` is not in the context for most problems.  So just enter an answer with `z` in it.  Another reason student answers fail to parse is if they enter a math operation or function that is disabled for the problem.  You should test with the course configuration option "Display the evaluated student answer" under "Problem Display/Answer Checking" set to both true and false to see that this works both ways.  If that option is set to true, you should see nothing different than usual.  If you have it set to false, then with the develop/WeBWorK-2.18 branch or any previous version of webwork, then the preview answer column will show nothing, but with this pull request it will show what would have been in the entered column if that option were true.

Alex-Jordan (approved):
Tested and working. if it is not clear, visit Course config and set "Display the evaluated student answer" to false. Visit a problem and enter something like "3+q" as an answer. Before this commit the results table will be empty in the Answer Preview column. After this commit, it shows `3+q` there.

pstaabp (approved):
See it now.  Didn't realize I need to update the configuration.

PR #2032, opened 2023-06-06
Simplify the combined header and remove the screen- and hardcopy-spec…

Alex-Jordan:
This separates the changes to header files from #2030, with a few additional small changes. (#2030 will be closed soon.)


conf/snippets/ASimpleCombinedHeaderFile.pg
```diff
@@ -1,89 +1,46 @@
...
+#####################################################################
+# Content to print only in PDF output
+# Use PGML markup
 
-% This is a TeX comment
-% Use this space to add header text to pdf hardcopies
-% Text added here will not be displayed in the browser
+$TeXonly = <<'PGML';
+This PDF is available for convenience. Assignments must be submitted within *WeBWorK* for credit.
+PGML
```
pstaabp, line 18: How about using `END_PGML` to be consistent with the standard PGML blocks?  I know it doesn't matter, but it may make more sense for some users.

conf/snippets/ASimpleCombinedHeaderFile.pg
```diff
@@ -1,89 +1,46 @@
...
-$BBOLD WeBWorK  Assignment \{ protect_underbar($setNumber) \}  is due :  $formattedDueDate. $EBOLD
-$PAR
-Here's the list of
-\{ htmlLink(qq!http://webwork.maa.org/wiki/Available_Functions!,"functions and symbols") \}
- which WeBWorK understands.
-$BR
-EOT
+$HTMLonly = <<"PGML";
+This assignment will close on *$formattedDueDate*.
+PGML
```
pstaabp, line 26: And here?

courses.dist/modelCourse/templates/ASimpleCombinedHeaderFile.pg
```diff
@@ -1,89 +1,46 @@
...
-####################################################
+#####################################################################
+# The last line of PGML below prints the items from above.
+# Other content will appear in all output forms.
 
 BEGIN_PGML
+[@ MODES(TeX => $TeXonly, HTML => $HTMLonly) @]**
 
+If all else fails, try answering with [`90^{\circ}`] because that is
+always right.
```
drgrice1, line 36: What does this mean?

Alex-Jordan:
It's a joke. 90 deg => right angle.

I couldn't think of anything good to put in both output forms and it was
late. Got any ideas?

Alex-Jordan:
The joke should go away,  I was not thinking clearly last night. But what to put there? The hardcopy header already lays out the due date, set name, and course name. So we don't want to write those again in print. What piece of sample text can we use that makes sense for both PDF and HTML?

Earlier I thought something like "This assignment has 23 exercises." But that "23" is not an environment variable, so we'd have to add that wherever environment variables are created. I'd rather just work with what we have. But I can't think of anything to say here.

somiaj:
I have a patch where I ass the `$setDescription` as a variable to PG, this way I can display the set description in headers. Would something like that work? (the patch is a simple one liner)

Alex-Jordan:
That is interesting, but sets might not have a description. For this default file, it would be nice if there was some test that just clearly shows up out of the box, in all cases.

`$setDescription` and a few other things might be good to include in a commented section to show header authors what is possible.

somiaj:
Here is my patch needed to add the `$setDescription`, and I do agree, might not be best for default since not all sets have this filled in. Just sharing the patch here for reference if you think it is something worth adding to set headers.

```diff
diff --git a/lib/WeBWorK/Utils/Rendering.pm b/lib/WeBWorK/Utils/Rendering.pm
index 0271995d9d..f4255bc561 100644
--- a/lib/WeBWorK/Utils/Rendering.pm
+++ b/lib/WeBWorK/Utils/Rendering.pm
@@ -133,6 +133,7 @@ sub constructPGOptions ($ce, $user, $set, $problem, $psvn, $formFields, $transla
 	$options{sectionNumber}    = $options{sectionName};
 	$options{recitationName}   = $user->recitation;
 	$options{recitationNumber} = $options{recitationName};
+	$options{setDescription}   = $set->description;
 	$options{setNumber}        = $set->set_id;
 	$options{studentLogin}     = $user->user_id;
 	$options{studentName}      = $user->first_name . ' ' . $user->last_name;
```

Alex-Jordan:
The joke is gone and now there is the set description as Jaimos suggested, along with a comment.

conf/snippets/ASimpleCombinedHeaderFile.pg
Alex-Jordan: I thought to not use `END_PGML`, to avoid any confusion for a new problem writer. This is not the same thing as a `BEGIN_PGML...END_PGML` block that is rendered. It is just storing some PGML markup as a string for uses later.

drgrice1: I agree that using `END_PGML` here could be confusing.  Although, maybe rather than using `PGML` instead use something like `END_HTML` and `END_TEX` for the `$HTMLonly` and `$TeXonly` sections respectively.  I am fine with just leaving it as it is too though.

conf/snippets/ASimpleCombinedHeaderFile.pg
```diff
@@ -1,89 +1,48 @@
...
 
 loadMacros(
-	"PG.pl",
-	"PGbasicmacros.pl",
-	#"source.pl", #uncommenting this will display "show source" button if that is permitted
+	"PGstandard.pl",
+	"PGML.pl",
 	"PGcourse.pl",
-	"PGML.pl"
 );
```
drgrice1, line 11: It seems that this should still have `TEXT($BEGIN_ONE_COLUMN)` here, and `TEXT($END_ONE_COLUMN)` where it was later.  Otherwise in two column mode this header content is placed into the first column in hardcopy.

Alex-Jordan: I noticed that but in the interest of clearing arcane incantations from the file (or as many as possible) I thought it was not so bad to have it there in the first column. Personally I like it better since the reading area is so wide before entering two columns. But I can understand wanting it at the top.

I'd like to look at getting "problem 0" (the header file) to be processed before the two-columns start, automatically. Unless anyone already knows why that won;t be easily doable without big changes.

drgrice1:
I am not sure that a`TEXT` call should be considered an arcane incantation.  There are still times that is useful, even with PGML.

I also don't quite understand the argument that the reading area is so wide before entering two columns.  I understand the argument in the browser on a wide screen monitor that to wide of a text area decreases readability, but I don't think that argument holds water in hardcopy for standard paper sizes.  Personally, I don't even like two column mode and only use one column mode.  Making text too narrow has the same problem as making text to wide, and also decreases readability.

Alex-Jordan:
`$BEGIN_ONE_COLUMN` is the arcane incantation I mean. "What does this mean?" says the author who is looking at screen output.

About textwidth, the thing you may be overlooking is how the margins for hardcopy that we produce are unnaturally small, making the one-column reading area 7.5 inches wide. That's with 10pt font, where out-of-the-box LaTeX `article` class has about 4.8 inches of text area. The `scrartcl` class is a little bit wide, closer to 5 inches. But at 10pt, 7.5 inches is big and bad for reading. These things are population averages though, and not everyone experiences it the same way.

At two columns, though, it's a good width. Smaller than it needs to be, but not small enough to impact reading comprehension speed.

Alex-Jordan:
I've gone down the rabbit hole to build on the improvements that @drdrew42 made to hardcopy. How important is it that the default styles continue to look exactly like they currently look? For example, is it important that I not do things like move something from the first page right header to the first page center header, etc? And how important is it that I not introduce something new to the header like the recitation?

conf/snippets/ASimpleCombinedHeaderFile.pg
pstaabp: I'm fine leaving as is too now.

drgrice1:
I also don't think that `$BEGIN_ONE_COLUMN` should be considered an arcane incantation, and furthermore, it really should be in this set header file.  The `$BEGIN_ONE_COLUMN` and `$END_ONE_COLUMN` variables are really specifically for set headers as you really wouldn't use them in an actual problem.  It would be a good idea to also add a comment in the file explaining it.

Alex-Jordan:
With the changes I made in #2049, I don't think `$BEGIN_ONE_COLUMN` and `$END_ONE_COLUMN` have a role to play. The set header will always be built in the one column area before (potentially) the theme has the problems enter two-columns.

Alex-Jordan:
Closing in favor of #2049.

PR #2036, opened 2023-06-06
move definition of pgmlPercent to within pgmlSetup

Alex-Jordan:
The way `\pgmlPercent` is currently defined, one "pgmlPercent" is defined before the file may have entered two-column mode. So for example something like this in a set's first problem
```
BEGIN_PGML
Text
====
More text
END_PGML
```
will make a horizontal rule that overflows into the second column.

This changes `\pgmlPercent` to be redefined every time `pgmlSetup` is invoked, so it's based on the actual local line width.


pstaabp (approved):
👍

PR #2038, opened 2023-06-07
Add Problem Editor to sidebar nav

Alex-Jordan:
This adds Problem Editor to the sidebar navigation.

Also when you are editing a problem in the context of some set, that will fall under "Problem Editor" instead of under "Hmwk Sets Editor".

And I made a `blankProblem3.pg`, trying to update some things.

There were some outdated references in the help files that this addresses as well.

Screen Shot 2023-06-07 at 12 54 36 PM


pstaabp:
I don't see the `blankProblem3.pg` in this PR.

Alex-Jordan:
Sorry, try now (after I force pushed).  I forgot to stage the new file.

conf/snippets/blankProblem3.pg
```diff
@@ -0,0 +1,53 @@
...
+   "PGcourse.pl",     # Customization file for the course
+);
+
+# Show which answers are correct and which are incorrect
+$showPartialCorrectAnswers = 1;
+
+############################################################
+# Setup
+
+Context("Numeric");
```
pstaabp, line 30: This isn't really needed. I've been removing this in all of the sample problems.

conf/snippets/blankProblem3.pg
```diff
@@ -0,0 +1,53 @@
...
+
+BEGIN_PGML
+Enter a value for [`\pi`].
+
+[_]{$pi}{5}
+
+END_PGML
+
+############################################################
+# Solution
```
pstaabp, line 45: It also seems a bit redundant to say solution here when just below, there is the `BEGIN_PGML_SOLUTION`

Alex-Jordan: The thinking is to have the "blocks" be separated. There are the `Setup` and `Statement` blocks, and it would be inconsistent not to label the last block.

pstaabp:
Although the number of tools has grown now (+2 with the set assigner), I think this makes more sense having it separate from the Problem Sets Editor.  

We should add this and the Set Assigner to the general help that opens with the `Help` on the navigation sidebar.

conf/snippets/blankProblem3.pg
drgrice1: In my opinion, unless you load one of the macros that incorrectly sets the context, then you should not call `Context('Numeric')`.  The only time you should need to set the context initially is if you are using another context.  The point is to get to the point that doing so is not needed, and fix these macros that incorrectly set the context.  See for example line 305 of contextPolynomialFactors.pl which is one of these bad macros.  The comment on that line even points this out.

Alex-Jordan: OK, I was remembering [a time](https://github.com/openwebwork/pg/pull/335) when it mattered.

conf/snippets/blankProblem3.pg
drgrice1: I agree with @pstaabp on this one.  Let syntax highlighting handle the separation.  If you are using the PG problem editor, then you have good syntax highlighting that does this.  I particularly despise the `######################################` lines.

conf/snippets/blankProblem3.pg
```diff
@@ -0,0 +1,53 @@
...
+DOCUMENT();
+
+loadMacros(
+   "PGstandard.pl",   # Standard macros for PG language
+   "PGML.pl",         # PGML markup and Math Objects
+   "PGcourse.pl",     # Customization file for the course
+);
+
+# Show which answers are correct and which are incorrect
+$showPartialCorrectAnswers = 1;
```
drgrice1, line 25: This also is not actually needed.  The only time this is needed is when you set the value of this variable to 0.  Perhaps change this to
```suggestion
# Uncomment the following if you don't want to show which answers are correct and which are incorrect
#$showPartialCorrectAnswers = 0;
```

conf/snippets/blankProblem3.pg
Alex-Jordan: Do we have a style for the stuff at the top of the file? 

```
############################################################
##DESCRIPTION
##  Arithmetic problem: give the value of pi
##ENDDESCRIPTION

##KEYWORDS('template', 'pi')

## DBsubject('Algebra')
## DBchapter('Fundamentals')
## DBsection('Real Numbers')
## Date('')
## Author('')
## Institution('')
############################################################

DOCUMENT();

...
```

Nothing but the lines that open with `##`? Spaces between some lines or no? Always one space after the `##` or sometimes two? Sometimes zero?

Alex-Jordan: How many blank lines surrounding lines like `BEGIN_PGML`, `END_PGML`?

templates/ContentGenerator/Base/links.html.ep
```diff
@@ -109,26 +109,28 @@
...
 			<li class="list-group-item nav-item"><%= $makelink->('instructor_set_assigner') %></li>
+			% # Problem Editor
+			<li class="list-group-item nav-item"><%= $makelink->('instructor_problem_editor') %></li>
+			% if (defined $problemID) {
+				<li class="nav-item">
+					<ul class="nav flex-column">
+						<li class="nav-item">
+							<%= $makelink->(
+								'instructor_problem_editor_withset_withproblem',
+								text     => maketext('[_1] #[_2]', $prettySetID, $prettyProblemID),
```
drgrice1, line 126: The `$prettySetID` needs to be wrapped in a `dir="ltr"` span.  So this will need to be changed to
```suggestion
								text => b(maketext(
									'[_1] #[_2]',
									$c->tag('span', dir => 'ltr', $prettySetID),
								   	$prettyProblemID
								)),
```
Note the `b` call is needed so that this is converted into a bytes object, and thus the result is not escaped in the later `link_to` call on line 538 of ContentGenerator.pm.

I am also not fond of the chosen text.  I would prefer '[_1]: Problem [_2]'.

templates/ContentGenerator/Base/links.html.ep
```diff
@@ -109,26 +109,28 @@
...
-						% }
 					</ul>
 				</li>
 			% }
 
 			% # Set assigner
 			<li class="list-group-item nav-item"><%= $makelink->('instructor_set_assigner') %></li>
+			% # Problem Editor
+			<li class="list-group-item nav-item"><%= $makelink->('instructor_problem_editor') %></li>
+			% if (defined $problemID) {
```
drgrice1, line 120: This needs to be
```suggestion
			% if (defined $prettySetID && defined $problemID) {
```
The `instructor_problem_editor_withset_withproblem` route requires that both the `setID` and `problemID` stash values are defined, and `$prettySetID` is used directly within this block.  Note that `$setID` will be defined if `$prettySetID` is.

templates/ContentGenerator/Instructor/PGProblemEditor/add_problem_form.html.ep
```diff
@@ -9,7 +9,7 @@
 		<div class="col-auto">
 			<%= select_field 'action.add_problem.target_set' => [
 					map { [
-						format_set_name_display($_) => $_, $_ eq $c->{setID} ? (selected => undef) : ()
+						format_set_name_display($_) => $_, $_ eq ($c->{setID} // 'undefined_Set') ? (selected => undef) : ()
```
drgrice1, line 12: This should be
```suggestion
						format_set_name_display($_) => $_, $_ eq ($c->{setID} // '') ? (selected => undef) : ()
```
There is no `undefined_Set` that will be in this list (unless an instructor actually creates a set named 'undefined_Set', but then `$c->{setID}` will be defined).

drgrice1 (commented):
I see several problem that need to be addressed.

I still need to do further testing on this.  There are several possibilities that this opens up for this module that I am not sure could occur before.

Also, this changes the behavior of the site nav when editing a problem that will take some getting used to.

conf/snippets/blankProblem3.pg
drgrice1: There should never be more than one consecutive blank line anywhere in my opinion.

drgrice1: As to the spaces after `##`, I think there should always be one space.  I could agree to two spaces for the lines in between `## DESCRIPTION` and `## ENDDESCRIPTION`.  No spaces looks bad everywhere.  Including `##DESCRIPTION` and `##ENDDESCRIPTION`.

drgrice1:
> We should add this and the Set Assigner to the general help that opens with the `Help` on the navigation sidebar.

@pstaabp:  Good catch.  I should have added the "Set Assigner" before when I added the link.  I didn't think about it then though.

Alex-Jordan:
> Also, this changes the behavior of the site nav when editing a problem that will take some getting used to.

I feel that too. When I started, I just wanted to have a quick way to immediately go to a place where I could start writing a new problem. Once I had that working, I happened to go edit a problem in a set, and it looked wrong that I was editing a problem with the editor, yet the navbar active link placed me under the Hmwk Set Editor instead of under the Problem Editor. So I moved that, and now some things are different for workflow.

I think I may move "Set Assigner" up above "Hmwk Sets Editor". It makes sense there, as a bridge in between "Classlist Editor" and "Hmwk Sets Editor". And then "Hmwk Sets Editor" and "Problem Editor" will be adjacent, which is also natural, and closer to how things currently work.

drgrice1:
That sounds right.

conf/snippets/blankProblem3.pg
Alex-Jordan: I updated the new problem template. Marking this subthread as resolved, but feel free to continue with suggestions to help it become better.

Alex-Jordan:
I've addressed a few small things, and will turn to @drgice1's comments later, maybe tomorrow.

somiaj:
I like this idea.  I have also wanted a faster way to just start editing a new problem.  I did notice that editing and saving of the blank file needs to be looked at.  It starts by stating you are editing the actual location of the file, `Editing blank problem in file "[WW]/conf/snippets/blankProblem3.pg".`  Maybe copy the blank file into the templates directory first similar to what adding a blank file from the set editor does.

Next, when you try to save the file you do get an error about the location needs to be inside the templates directory, but due to a typo the error doesn't work.  Line 611 in `lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm` the typo `$c->maktext(` needs to be fixed.  This should also work similar to using the set editor to add a blank file.

Last the default save as location of the file is the full path of the blank file inside the templates directory.

As an alternative to loading the blank file by default, what about having a landing page that a user can use a some form to select either a blank problem or an existing .pg problem in their templates directory to edit?  Although you can use the library browser to access your local problems, this could provide a faster approach to using a previous problem as a template if you know the filename.

Alex-Jordan:
I did all @drgrice1's suggestions as one commit attributed to Glenn. I just saw some of the things you mentioned @somiaj  and will patch next round :)

somiaj:
Hmm, grepping for `maktext`, I find this located in three different places. I'll make a PR just for fixing the typos @Alex-Jordan

drgrice1:
You seem to have found a GitHub trick that I am not familiar with.  How did you put all of my suggestions together into one commit attributed to me?

Alex-Jordan:
Nope, I manually applied them locally.

drgrice1:
How did you get it to be attributed to me then?

Alex-Jordan:
I used
got commit -m message --author="..."

drgrice1:
I see.  I should have figured that out I guess.  So not GitHub tricks then.  Too bad.  It would be nice if you could add suggestions en masse directly from the UI instead of applying them one at a time.

Alex-Jordan:
Sorry, was on my phone before. But yeah, `git commit -m "Suggestions from PR#2038" --author="Glenn Rice <grice1@missouriwestern.edu>"`.

>  It would be nice if you could add suggestions en masse directly from the UI instead of applying them one at a time.
Feature request for GitHub :)

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -334,7 +332,8 @@ sub page_title ($c) {
 		}
 	}
 
-	return $c->maketext('Problem [_1]', $problemID);
+	return (defined $setID    ? $c->tag('span', dir => 'ltr', format_set_name_display($setID)) . ': ' : '')
```
Alex-Jordan, line 335: Does it make sense that perltidy wanted all the spaces here after `$setID`?

Alex-Jordan:
@somiaj Do you see how to fix the things you cited? I gave it a go, but I went in circles and got nowhere. If you see what needs doing, could you open a PR to this branch?

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
drgrice1: It is trying to line up the `?` on this line with the one on the line below.  It would also try to line up the `:`'s later on those two lines, but doing so pushes the second line past the 120 character line limit, and so it doesn't.

somiaj:
@Alex-Jordan I opened a PR against your branch for one way to fix the issues I mentioned.

Alex-Jordan:
For anyone who wants to probe and test this, it's ready for that now. I don't foresee more changes until someone else sees something to change.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
Alex-Jordan: I see. I was looking at something with different setting for tab width and it didn't align.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -334,7 +332,7 @@ sub page_title ($c) {
 		}
 	}
 
-	return $c->maketext('Problem [_1]', $problemID);
+	return $c->maketext((defined $setID ? "$setID " : '') . 'Problem [_1]', $problemID);
```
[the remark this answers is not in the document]
Alex-Jordan: This is in there now.

templates/ContentGenerator/Base/links.html.ep
Alex-Jordan: This is in there now.

templates/ContentGenerator/Instructor/PGProblemEditor/add_problem_form.html.ep
Alex-Jordan: This is in there now.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -179,6 +179,8 @@ sub pre_header_initialize ($c) {
 				$c->{file_type} = 'problem';
 			}
 		} else {
+			copy($ce->{webworkFiles}{screenSnippets}{blankProblem}, "$ce->{courseDirs}{templates}/local/$BLANKPROBLEM")
```
drgrice1, line 182: I really don't like this, and think this should be removed.  A file should not be copied into the templates directory every time that a link in the site navigation is clicked on.  I also think it is a bad idea to copy the newProblem.pg file into the templates directory, and then tell the user that the file is a template and to use "Save As" to create a new file.

Furthermore, I don't like that this is copied into the "local" subdirectory of templates.  The "local" directory may not be local to the course.  I use a symlink and this is a shared directory between courses that is intended to be read only for these courses.  I know that there are others that do this as well.

somiaj: > Furthermore, I don't like that this is copied into the "local" subdirectory of templates. The "local" directory may not be local to the course. I use a symlink and this is a shared directory between courses that is intended to be read only for these courses. I know that there are others that do this as well.

I have been thinking about this too.  In the use case you mention here, do instructors not use the homework set editor to add a blank problem to homework sets which will also copy a blank problem into a sub directory of the templates directory?  I added this to use the blank problem structure that was already in use for blank problems in homework sets. The other option is to add checks for `blank_problem` to get the same effect without copying the file (I will look into this).

I have been thinking of a way to better deal with blank problems without copying them into the templates directory in both this case and when adding blank problems to homework sets.  This case is straight forward to not have to copy a file, but the homework set case seems a bit more complicated since the file is not only used in the editor, but also used to render the problem, so that change could affect multiple pages.

somiaj: I add a pull request to @Alex-Jordan branch with a way to do this without copying the blank problem.  I would still be interested in seeing if it is also possible to not copy blank problems when adding them to homework sets.

drgrice1: When a blank problem is added to a homework set, the blank problem is not created in the `templates/local` directory.  It is created in the directory `templates/setSetId` where SetId is of course the set name.  However, I already don't like that behavior.  In that case as well you are creating a file that is ultimately not used.  Ideally, when the problem editor link in the site nav is used, there should be a page that opens first asking the user for a file path (name and location in templates) to copy the file to.

Alex-Jordan:
>  I add a pull request to @Alex-Jordan branch with a way to do this without copying the blank problem. I would still be interested in seeing if it is also possible to not copy blank problems when adding them to homework sets.

I don't seem to have any open pull requests on my end. I already merged the earlier one, so if you made edits and pushed to your own branch, there would need to be a new PR for me to see those.

somiaj:
@Alex-Jordan Hmm wonder where the pull request went, maybe I didn't actually finalize it.  Anyways I tracked down the commit and made a pull request hoping it worked this time.

drgrice1 (commented):
This is almost in order now.  There are a few details that need to be fixed though.

conf/snippets/newProblem.pg
```diff
@@ -0,0 +1,38 @@
...
+
+# Uncomment the following if you don't want to show which
+# answers are correct and which are incorrect
+#$showPartialCorrectAnswers = 0;
+
+$pi = Real("pi");
+
+BEGIN_PGML
+Enter a value for [`\pi`].  
+[_]{$pi}{5}
```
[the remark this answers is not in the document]
Alex-Jordan: I changed it to a new paragraph. IIRC, the two spaces is like a `<br>` or a `//`, while a blank line in between is starting a new `<p>` or using `\par`. I often use the two spaces in this construction since logically the answer blank is part of the same "paragraph".  But a "paragraph" doesn't always align with what a `<p>` or a `\par` are for.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -248,16 +248,16 @@ sub initialize ($c) {
 		$c->addbadmessage($c->maketext('The file "[_1]" cannot be found.', $c->shortPath($c->{inputFilePath})));
 	} elsif (!-w $c->{inputFilePath} && $file_type ne 'blank_problem') {
```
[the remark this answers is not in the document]
Alex-Jordan: Change made locally.

templates/ContentGenerator/Instructor/PGProblemEditor/save_as_form.html.ep
```diff
@@ -5,11 +5,12 @@
...
 % last if $c->{file_type} eq 'course_info' && -e $c->{editFilePath};
 %
+% my $isBlank       = $c->{file_type} eq 'blank_problem';
 % my $templatesDir  = $ce->{courseDirs}{templates};
-% my $shortFilePath = $c->{editFilePath} =~ s|^$templatesDir/||r;
+% my $shortFilePath = $isBlank ? 'newProblem.pg' : $c->{editFilePath} =~ s|^$templatesDir/||r;
 %
 % # Suggest that modifications be saved to the "local" subdirectory if its not in a writeable directory
-% $shortFilePath = "local/$shortFilePath" if !-w dirname($c->{editFilePath});
+% $shortFilePath = "local/$shortFilePath" unless $isBlank || -w dirname($c->{editFilePath});
```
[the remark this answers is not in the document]
Alex-Jordan: @somiaj Would you weigh in on what this should be?

templates/ContentGenerator/Instructor/PGProblemEditor.html.ep
```diff
@@ -25,7 +25,7 @@
 % }
 %
 % my %titles = (
-	% blank_problem                => x('Editing <strong>blank problem</strong> in file "[_1]".'),
+	% blank_problem                => x('Editing <strong>blank problem</strong> template.'),
```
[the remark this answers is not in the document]
Alex-Jordan: Made this change.

Alex-Jordan:
This is updated after fixing a merge conflict and making Glenn's changes. Except that one question about `save_as_form.html.ep` remains.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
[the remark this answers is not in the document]
drgrice1: Line 239 of PGProblemEditor.pm can be deleted now.  The `$file_type` variable is no longer used in the `initialize` method.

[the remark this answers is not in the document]
Alex-Jordan: Oops, I missed that you said that before.

templates/ContentGenerator/Instructor/PGProblemEditor/save_as_form.html.ep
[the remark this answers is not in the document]
Alex-Jordan: I used to think of `local/` as a place for problems I wrote for my course. Now  think it's better to think of it as a place to keep parallel copies of problems from various libraries. If that is how others see it too, then I think this is right. If you have a brand new problem that "is blank" (did not come from customizing something like an OPL problem) then it's not right for `local/`.

It's not good to put it in `[TMPL]` either though. If the author has a file folder system in place, they should use it. Otherwise maybe we want something like a `pg/` folder it could land in by default.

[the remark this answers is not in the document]
drgrice1: I have though of `local/` as a local library for the server, like the `Library` link is for the OPL library.  However, it may be that `local/` is intended to be a place for problems for your course, and that my interpretation is incorrect.  I am fine with adapting to that if that is what was intended.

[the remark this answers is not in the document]
Alex-Jordan: FWIW, I currently think the right choice is to not save this into `local/`. I don't love saving it in `templates/` but I think that is better than `local/`. I'm fine with it the way it is though. Authors can type in their own folder names if they want.

The only other thing that occurs to me is to introduce a new folder, like `temeplates/my_problems/` or `temeplates/pg/` or something. But that has to do with my interpretation of `local/`, as a place for parallel customizations of various library problems.

drgrice1 (approved):
This is generally ready to go.  I can approve it as it stands.  We should talk about the local directory, and what is intended with that in our meetings.

pstaabp (approved):
I think this looks good too.  

Should we wait and discuss about the location of `local` versus not?  Or merge?

Alex-Jordan:
If there is a problem file that is not writable, a new file is written into `local`. This is consistent with my current thinking (I did not always think of things this way) that `local` is for parallel, modified copies of some problem from some library outside of the course.

Here we are creating a new file from nothing (or rather from a template). So I lean toward not putting this new file in `local`. If the person is writing a new filie for, say, their linear algebra course in a matrix multiplication set, they should do something like make this new file be saved at `[TMPL]/LinearAlgebra/MatrixMultiplication/myproblem.pg`. Well, not that they "should" do that, but it's what my practice would be at this point.

Related to all this and also a recent forum thread, when we *do* save a copy of an unwritable file into `local`, maybe we should think about inserting a comment at the top of the file with some lineage metadata.  Like `# This file was copied from Library/...` or `## Parent(Library/...)`.

PR #2043, opened 2023-06-09
Fix the raw output format for FormatRenderedProblem.

### LINKED ISSUE #2042: 2.18 html2xml raw
@sean-fitzpatrick  has 2.18 running and tried to wire it into a PreTeXt book, but something is not working. In this [Google Group thread](https://groups.google.com/g/pretext-support/c/uUjq0OkvAv0/m/MIgvam1aBAAJ) (which maybe you have to sign up for) he posts a link to his server that passes through `html2xml` using the `raw` format and it should be returning JSON, but instead you get a server error.

But if you cut off the last two queries (`&answersSubmitted=1&WWsubmit=1`) then it looks like it gives what I think it should give.

Is anyone familiar with what may have changed with that endpoint from 2.17 that could be related to this?

drgrice1:
The AttemptsTable::answerTemplate now returns a ByteStream object instead of a string.  That needs to be converted to a string before being passed to be encode into JSON.  Otherwise JSON's encode method fails because that is a blessed object.

This should fix #2042.

### DIFF
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -211,7 +211,7 @@ sub formatRenderedProblem {
 		$output->{input}      = $ws->{input};
 
 		# The following could be constructed from the above, but this is a convenience
-		$output->{answerTemplate}  = $answerTemplate if ($answerTemplate);
+		$output->{answerTemplate}  = $answerTemplate->to_string if $answerTemplate;
 		$output->{lang}            = $PROBLEM_LANG_AND_DIR{lang};
 		$output->{dir}             = $PROBLEM_LANG_AND_DIR{dir};
 		$output->{extra_css_files} = \@extra_css_files;


sean-fitzpatrick:
I can confirm that this fixes #2042.

pstaabp (approved):
Haven't tested it, but am fine approving with @sean-fitzpatrick recommendation that this fixes it.

sean-fitzpatrick:
Here's a [page in a PreTeXt book]([Exercise Group](https://opentext.uleth.ca/apex-accelerated/sec_chainrule.html#exercisegroup-41) 
https://opentext.uleth.ca/apex-accelerated/sec_chainrule.html#exercisegroup-41) with lots of embedded WeBWorK problems. Prior to making the change here, there was no response when clicking on the Check Responses button, but now it works as expected.

PR #2045, opened 2023-06-09
alternate pathway to user detail

Alex-Jordan:
This is one last (I hope) change I'd like to propose to sidebar navigation. With this, if you are acting as some user, then you have a direct link in the sidebar nav to see their user detail page (which has more details about all the sets assigned to them, including test versions).

The only current way (that I know of) to navigate here is to go to the Classlist Editor and click the link in the "Assigned Sets" column. That still works with this. But also now if you arrive at that page this way, you will see this is where you are in the sidebar nav.

If you are acting as user X and you go to Classlist Editor and click the link in the "Assigned Sets" column for user Y, you will see two links in the sidebar (assuming X ≠ Y). The one for Y is active, and the one for X is just visible because you are acting as X. Here is a screenshot with all of that going on, consistent with Statistics and Student Progress.

Screen Shot 2023-06-09 at 1 59 43 PM

### DIFF
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -89,6 +89,33 @@
 			<li class="list-group-item nav-item"><%= $makelink->('instructor_tools') %></li>
 			% # Class list editor
 			<li class="list-group-item nav-item"><%= $makelink->('instructor_user_list') %></li>
+			% # User Assignments
+			% if (defined $eUserID && $eUserID ne $userID || defined $urlUserID) {
+				<li class="list-group-item nav-item">
+					<ul class="nav flex-column">
+						% if (defined $urlUserID) {
+							<li class="nav-item">
+								<%= $makelink->(
+									'instructor_user_detail',
+									text       => $urlUserID,
+									captures   => { userID => $urlUserID },
+									link_attrs => { dir => 'ltr' }
+								); %>
+							</li>
+						% }
+						% if ($userID ne $eUserID && (!defined $urlUserID || $urlUserID ne $eUserID)) {
+							<li class="nav-item">
+								<%= $makelink->(
+									'instructor_user_detail',
+									text       => $eUserID,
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks good to me.  Although, there is one thing I am not certain of.  We haven't been adding the `dir="ltr"` attribute for user ids in the site nav.  Notice that the student progress and statistics sub links do not have that.  @taniwallach: Should we be doing that?

drgrice1:
Thinking about it more, I suspect that we should have the `dir="ltr"` attribute on all of the student id links in the site nav.  This is because at this point we don't allow utf8 characters like those in the Hebrew alphabet in user ids.  However, it is probably best to let @taniwallach confirm this.

Alex-Jordan:
I added `dir="ltr"` to the other user ID links in the site nav.

There are many other places where a user ID or username is printed to screen. Consider this screenshot after I switched to Hebrew:

Screen Shot 2023-06-12 at 11 20 22 AM

No inline directions are specified here, but the browser is doing the right thing. Reading form the right, there is Hebrew, then a correct switch to ltr for the user names, then back to rtl for the period. The browser is probably seeing that this run of characters are from a code block that typically uses ltr, and makes that change.

Actually if I explicitly set the sidebar nav links to `rtl`, I _still_ see them presented by the browser left-to-right. 

This is only with one browser of course, and I don't know how all this plays out with a screen reader.

pstaabp:
I like this, but wonder about the number of instructor tools now available/visible on the nav menu.  Are there too many?  Perhaps we can visually break them up with horizontal line(s).  I don't have a good location though.

Alex-Jordan:
Note that most of the time, this particular change does not make the nav menu any longer. It only does so when you are acting as another user, or if you do something like click the link under the Assigned Sets column in the Classlist Editor.

pstaabp (approved):
I would say to merge and if @taniwallach runs across this or other places that we need to update the text direction that we have another PR for that.

PR #2048, opened 2023-06-13
Round initial point value to match step size in SingleProblemGrader.

somiaj:
When using the point value in the single problem grader, the initial value needs to also be rounded to the nearest stepsize. Without this it is possible that the initial point value is an invalid number for the point value input.

### DIFF
--- a/templates/HTML/SingleProblemGrader/grader.html.ep
+++ b/templates/HTML/SingleProblemGrader/grader.html.ep
@@ -80,6 +80,11 @@
 			% } else {
 				% $stepSize = int(($grader->{problem_value} - 1) / 100) + 1;
 			% }
+			% # Round point score to the nearest $stepSize.
+			% param(
+				% 'grader-problem-points',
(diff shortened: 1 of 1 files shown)


templates/HTML/SingleProblemGrader/grader.html.ep
```diff
@@ -80,6 +80,14 @@
 			% } else {
 				% $stepSize = int(($grader->{problem_value} - 1) / 100) + 1;
 			% }
+			% # Round point score to the nearest $stepSize.
+			% param(
+				% 'grader-problem-points',
+				% wwRound(
+					% 2,
+					% wwRound(0, $grader->{recorded_score} * $grader->{problem_value} / $stepSize) * $stepSize
+				% )
```
drgrice1, line 89: This could be condensed into one line, and still fit into the 120 character limit.

drgrice1 (approved):
This looks good.

I have just one minor suggestion on code formatting.

Alex-Jordan (approved):

PR #2054, opened 2023-06-14
Do not html escape a link shown on the problem sets page (course home)

drgrice1:
The message that is shown to users that do not have the navigation_allowed permissions is currently html escaped.  That message can have a link to the LMS in it, and in that case the actual html for the link is shown instead of the html rendering of the link.  To fix this that message must not be html escaped of course.


pstaabp (approved):

somiaj (approved):
Looks good. Didn't test this, but it obviously fixes the issue of escaping html tags when it shouldn't.

Issue #2047, opened 2023-06-12
In WeBWorK 2.18 installation instructions, check_latex step is too early


sean-fitzpatrick:
The `check_latex` routine cannot run without the files `conf/site.conf` and `conf/localOverrides.conf`.

But at the stage in the installation instructions where we are advised to run `check_latex`, these files have not yet been created.

To avoid confusion, it seems like it would be better to either leave this test until later, or have us copy the `dist` files earlier in the process.

drgrice1:
Thanks for pointing this out.  I have reorganize things so that section is after the webwork configuration section.

drgrice1:
I believe that all of the things brought up in this issue are now dealt with.

PR #2055, opened 2023-06-14
Update the temporary directory comments and location.

drgrice1:
The comment is updated to state that the directory chosen must be have the correct permissions set.  Note the comment already states that the directory must be created.

The current comment about the system default for apache2's document root is incorrect.  On most systems that is `/var/www/html`, not `/var/www`.  So that comment is corrected, and the example settings adjusted.

This is to address what is left of issue #2047.


somiaj (approved):
Looks good. Just updating commented out options to match standard locations, so nothing to test.

PR #2057, opened 2023-06-15
add css to pod

pstaabp:
To improve the looks of the POD, a css file has been added.

The css just changes the standard font and some padding to the body.  Styling is not my forte, so welcome any additions/changes.


pstaabp:
I have tested this locally, but not on a server.   It looks like it should work, but might need some tweaking for that.

pstaabp:
I think I will close this since #2067 contains this and goes far beyond.

PR #2060, opened 2023-06-15
Fix the LTI config tab.

drgrice1:
The LTIConfigvalues.config inclusion is moved to to the end of defaults.config, instead of the end of authen_LTI.conf.  This means that file is always included.  However, the config variables it defines are only added to the `$ConfigValues` array if there are uncommented entries in the `@LTIConfigVariables` array that is still defined in authen_LTI.conf.  So the behavior is the same as before.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -164,6 +164,4 @@ $LTIMassUpdateInterval = 86400;    #in seconds
 	#'debug_lti_parameters'
 );
 
-include('conf/LTIConfigValues.config');
-
 1;    # final line of the file to reassure perl that it was read properly.
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -2078,4 +2078,6 @@ $ConfigValues = [
 	],
 ];
 
+include('conf/LTIConfigValues.config');
+
 1; #final line of the file to reassure perl that it was read properly.


drgrice1:
Note that if you still have the inclusion of the `LTIConfigValues.config` in your local `authen_LTI.conf` file it won't cause any issues or conflicts with this pull request.  In fact it will not actually do anything.

Alex-Jordan (approved):
Works for me.

pstaabp:
I'm not seeing the LTI tab in the course configuration.  I have `authen_LTI.conf` loaded and am seeing the LTI Grade Update nav item show up. The `authen_LTI.conf` has
```
@LTIConfigVariables = (
	'LTI{v1p1}{LMS_name}',
	# 'LTI{v1p3}{LMS_name}',
	'LTI{v1p1}{LMS_url}',
	# 'LTI{v1p3}{LMS_url}',
	'external_auth',
	'LTIGradeMode',
	'LMSManageUserData',
	'debug_lti_parameters'
);
```
at the bottom, so it seems like should be showing up.  I'm testing with @drgrice1 's LMS tool.

drgrice1:
You don't need to use the LMS tool to test this.  If authen_LTI.conf has the lines uncommented and is included in localOverrides.conf, and that isn't overridden in the course's course.conf or simple.conf file, then you will see the LTI config tab with this pull request.

pstaabp:
Not sure why I don't see it.  

I was also trying to troubleshoot this with print or warn statements, but adding these to any of the config files results in errors.

drgrice1:
In any case, we need to get this in.  See #2068 for problems caused by this.

somiaj (approved):
Fixed the issue for me.

Alex-Jordan:
I would merge this, but we should track down why @pstaabp isn't seeing the LTI tab. Can you check over all config files and how they compare against the `.dist` versions?

pstaabp:
Finally figured this out.  There was a typo in the `authen_LTI_1_1.conf` name.  oops and that file wasn't loading.

PR #2062, opened 2023-06-16
Fix the PG Problem Editor for files that have spaces in the filename.

drgrice1:
The glob in the `getBackupTimes` method needs special quoting to handle spaces in filenames.

To test this try to edit a pg problem with spaces in the filename.  On develop the editor won't open, and you get an exception screen.  With this pull request it will work of course.


Alex-Jordan (approved):

pstaabp (approved):
👍

PR #2064, opened 2023-06-19
Only copy over auxiliary files that are referenced inside a .pg file.

somiaj:
Instead of copying over all non .pg files assuming they are auxiliary
  files, first grep the contents of the .pg file for the file name
  in question, then only copy over the file if it is found inside
  the problem's contents.

This addresses the issue found in the comments here, https://github.com/openwebwork/webwork2/pull/1827#discussion_r1233494179


Alex-Jordan:
This is interesting but I believe I've seen problem files where this would prevent too much from being copied. Liike if there are image files `image1.png`, `image2.png`, neither filename might appear in the problem source. It could have something like `"image${index}.png"`.

somiaj:
Good catch, I wasn't able to think of any common ways of writing problems, but forgot that it might be common to construct the filename from variables.  So the question might be if it is better to copy over too many files, or too few which would then require manually copying some additional files.

somiaj:
I'll close this for now due to not being reliable enough to determine if a file is an auxiliary file or not.

PR #2066, opened 2023-06-20
Sample Problems link in PGeditor now shows the sample problems locally

pstaabp:
This changes the "Problem Techniques" link at the top of the PGeditor to "POD and Sample Problems" and removes the POD link.  In addition, this now points to the new location for PG documentation.


drgrice1:
I think you should remove the change to the POD link.  I have a better solution coming soon.

drgrice1:
Just remove the link to the github POD.  The links to the sample problems can stay for now, but I think a similar solution to what I am doing with the POD will be better for those too.  Basically, I am adding routes directly in webwork to locally served and dynamically generated POD for PG.  The advantage of this is that the links will always be to the POD for the version of PG currently on the server.

pstaabp:
> The advantage of this is that the links will always be to the POD for the version of PG currently on the server.

I've been thinking about how to handle this, but figured we had until 2.19 to determine how to do this.  If we want to still have the POD/Sample problems on an external website (and I think we do), we probably need to figure something out.

drgrice1:
We certainly should still have the POD and sample problems on an external website.  For the POD it is easy to make the POD for various versions of webwork available. Just use a base url that includes the version.  That should work for the sample problems too.

As to serving the sample problems by webwork, a similar technique to #2067 should work for the sample problems with some tweaking to your script.

pstaabp:
> As to serving the sample problems by webwork, a similar technique to https://github.com/openwebwork/webwork2/pull/2067 should work for the sample problems with some tweaking to your script.

@drgrice1 have you started a PR for this?  If not, I can do this.

drgrice1:
I have not started working on that.

pstaabp:
> I have not started working on that.

I'll do this then.

pstaabp:
Just finished updating this.  This now works similarly to #2067 in that the link in the PGeditor now dynamically produces the sample problems. 

A few things
- openwebwork/pg#811 must be paired with this
- The package [Mojolicious::Plugin::RenderFile](https://metacpan.org/pod/Mojolicious::Plugin::RenderFile) needs to be installed--I'll need to update the `check_modules.pl` and docker containers for this. 
- I ended up copying some templates from the pg side to get rendering to work correctly.  If there is a better way to reuse the copy, let me know.
- Mysteriously, the copy to clipboard button on the sample problems has disappeared, although it appears that everything is the same from the PG side.  It appears to be adding the class `position-absolute` to the button.  @drgrice1 if you have any ideas, let me know.

DockerfileStage1
```diff
@@ -144,7 +144,7 @@ RUN apt-get update \
 # ==================================================================
 # Phase 3 - Install additional Perl modules from CPAN that are not packaged for Ubuntu or are outdated in Ubuntu.
 
-RUN cpanm install -n Statistics::R::IO DBD::MariaDB Mojo::SQLite@3.002 \
+RUN cpanm install -n Statistics::R::IO Mojolicious::Plugin::RenderFile DBD::MariaDB Mojo::SQLite@3.002 \
```
drgrice1, line 147: This should instead add the Ubuntu package `libmojolicious-plugin-renderfile-perl` above.

pstaabp: I forgot to look for that in the ubuntu packages.  Will switch that.

templates/layouts/SampleProblemIndexLayout.html.ep
```diff
@@ -0,0 +1,94 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+	<meta charset="utf-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<title>PG Sample Problems</title>
+	<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/css/bootstrap.min.css" rel="stylesheet">
```
drgrice1, line 8: This should not use the hard coded cdn link.  It should use the usual methods for webwork2.  See `htdocs/themes/math4/system.html.ep` for the correct way to add the bootstrap.css file.  Just use our custom css file.

drgrice1: This goes for all of the files you have currently set to use a cdn link.

templates/ContentGenerator/SampleProblemViewer/problem.html.ep
```diff
@@ -0,0 +1,113 @@
...
+	defer => undef =%>
+	<script type="module">
+		for (const pre of document.body.querySelectorAll('pre.CodeMirror')) {
+			CodeMirror.runMode(pre.textContent, 'PG', pre);
+		}
+
+		for (const btn of document.querySelectorAll('.clipboard-btn')) {
+			btn.addEventListener('click', () => navigator.clipboard.writeText(btn.dataset.code));
+		}
+	</script>
```
drgrice1, line 110: This should be moved to a separate js file in htdocs/js.

templates/layouts/SampleProblemIndexLayout.html.ep
pstaabp: I switched the links in the `problem.html.ep` template, but missed this one.

pstaabp: For the codemirror css, should I load it with `getAssetsURL` like in `templates/HTML/CodeMirrorEditor/js.html.ep` ?

pstaabp: Line 5 of that file?

drgrice1: Yes

pstaabp: There's a small javascript at the bottom of this file as well.  Should I put it in its own file or is this small enough?

drgrice1: All JavaScript should be in a separate file, and not inline.

DockerfileStage1
```diff
@@ -79,6 +79,7 @@ RUN apt-get update \
 	libmodule-build-perl \
 	libmodule-pluggable-perl \
 	libmojolicious-perl \
+	libmojolicious-plugin-renderfile-perl \
```
drgrice1, line 82: This also needs to be added to the single stage build Dockerfile.

templates/layouts/SampleProblemIndexLayout.html.ep
```diff
@@ -0,0 +1,93 @@
+<!DOCTYPE html>
```
drgrice1, line 1: Please rename this file to `sample_problem_index.html.ep`.  There is no need to have `layout` in the name as it is in the layouts directory, and the naming scheme that I have been using for files that don't belong to modules like this is snake case.

conf/defaults.config
```diff
@@ -341,9 +341,8 @@ $webworkURLs{WikiMain}      = "https://webwork.maa.org/wiki/";
 $webworkURLs{SiteMap}       = "https://webwork.maa.org/wiki/WeBWorK_Sites";
 
 #URL for help buttons on the PGProblemEditor pages:
-$webworkURLs{problemTechniquesHelpURL}='https://webwork.maa.org/wiki/Category:Problem_Techniques';
+$webworkURLs{problemTechniquesHelpURL}='https://openwebwork.github.io/pg-docs/index.html';
```
drgrice1, line 344: This link should also be deleted with this pull request.  It isn't needed anymore.

templates/ContentGenerator/SampleProblemViewer.html.ep
```diff
@@ -0,0 +1,48 @@
...
+</head>
+
+<body>
+	<div class="container" id="_sample_prob_top_">
+		<h1 class="mt-3"><%= title %></h1>
+		<hr>
+		<div class="list-group">
+			<a href="sampleproblems/categories.html" class="list-group-item list-group-item-action" aria-current="true">
+				<div class="w-100">
+					<span class="h4">Sample Problems by Categories</span>
```
drgrice1, line 22: This should be "Sample Problems by Category".

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
```diff
@@ -0,0 +1,85 @@
...
+			metadata        => $metadata,
+			filename        => $filename,
+			macro_locations => $macro_locations,
+			pod_root        => '/webwork2/pod/macros',
+			pg_doc_home     => '/webwork2/sampleproblems'
+		);
+	} elsif ($c->stash->{filePath} =~ /\.pg$/) {
+		# Render the .pg file as downloadable file.
+		my $probFile = "$pg_root/doc/sample-problems/" . $c->stash->{filePath};
+		$c->app->log->debug($probFile);
```
drgrice1, line 79: This line needs to be deleted.

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
```diff
@@ -0,0 +1,85 @@
...
+	my $pg_root = $c->ce->{pg_dir};
+
+	my $metadata = LoadFile("$pg_root/doc/sample-problems/sample_prob_meta.yaml");
+
+	# Render one of the four indexes.
+	if (grep { $c->stash->{filePath} eq "$_.html" } (qw/categories techniques subjects macros/)) {
+		my $type = $c->stash->{filePath} =~ s/.html$//r;
+		warn 'here';
+		my $params = buildIndex($type, metadata => $metadata);
+		$c->app->log->debug($c->dumper($params));
```
drgrice1, line 54: This line also needs to be deleted.

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
```diff
@@ -0,0 +1,85 @@
...
+		return $c->render(
+			template => 'ContentGenerator/SampleProblemViewer/index_main',
+			layout   => 'SampleProblemIndexLayout',
+			sidebar  => $c->render_to_string('ContentGenerator/SampleProblemViewer/index_sidebar', %$params),
```
drgrice1, line 58: Delete this line.  Instead change line 81 of templates/layouts/SampleProblemIndexLayout.html.ep to `<%= include 'ContentGenerator/SampleProblemViewer/index_sidebar' =%>`.  Note that the stash values (the `%$params` you are passing here) are automatically inherited by included templates.

templates/layouts/SampleProblemIndexLayout.html.ep
drgrice1: Actually, scratch the previous comment about this file.  Instead delete this file and integrate its contents into templates/ContentGenerator/SampleProblemViewer/index_main.html.ep.  You should only add a layout if it is used by multiple templates.  This is only used by one template, and so should not exist.

pstaabp: Since there was no point to splitting the file, I put the layout, sidebar and main pane into a single file--a little large.  And now this renders on a single file.

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
pstaabp: done.

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
pstaabp: done.

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
pstaabp: As I mentioned above, I combined into a single file.

pstaabp:
Made the above fixes.

drgrice1:
@pstaabp:  I added a pull request to this branch that fixes the clipboard icon and many other things, as well as makes rendering more dynamic.  I put in a corresponding pull request to openwebwork/pg#811 with needed changes to make this all work.  See the explanations on those pull requests for more details.

Alex-Jordan:
> This changes the "Problem Techniques" link at the top of the PGeditor to "POD and Sample Problems" and removes the POD link.

I don't see that. Did I miss a change that happened in the interim?

I have both this and the corresponding pg branch, and I double checked commit hashes on each. I ran `npm ci` in each `htdocs` folder, and I restarted `webwork2`. I see:

Screen Shot 2023-06-30 at 3 51 40 PM


What should I be checking here?

drgrice1:
@pstaabp created this pull request before #2067.  With that pull request he changed it so that the "Sample Problems" link stayed with the same name, but it now renders the problem documentation dynamically.

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
```diff
@@ -0,0 +1,122 @@
...
+
+=cut
+
+sub renderSampleProblem ($c) {
+	my $pg_root = $c->ce->{pg_dir};
+
+	(undef, my $macro_files) = Pod::Simple::Search->new->inc(0)->survey("$pg_root/macros");
+	$c->log->info($c->dumper($macro_files));
+	my %macro_locations = map { basename($_) => $_ =~ s!$pg_root/macros/!!r } keys %$macro_files;
+	my $metadata        = generateMetadata("$pg_root/doc/sample-problems", macro_locations => \%macro_locations);
```
drgrice1, line 50: This should be
```suggestion
	my $metadata        = generateMetadata("$pg_root/tutorial/sample-problems", macro_locations => \%macro_locations);
```

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
```diff
@@ -0,0 +1,122 @@
...
+		);
+	} elsif ($c->stash->{filePath} =~ /\.pg$/) {
+		unless ($metadata->{ basename($c->stash->{filePath}) }) {
+			return $c->render(data => $c->maketext('File not found.'));
+		}
+
+		# Render the .pg file as a downloadable file.
+		return $c->render_file(
+			data => parseSampleProblem(
+				"$pg_root/doc/sample-problems/" . $c->stash->{filePath},
```
drgrice1, line 91: "doc" => "tutorial"

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
```diff
@@ -0,0 +1,122 @@
...
+				pg_doc_home => $c->url_for('sample_problem_index')
+			)->{code}
+		);
+	} else {
+		unless ($metadata->{ basename($c->stash->{filePath}) . '.pg' }) {
+			$c->render(data => $c->maketext('Sample problem not found.'));
+		}
+
+		# Render a problem with its documentation.
+		my $problemFile = "$pg_root/doc/sample-problems/" . $c->stash->{filePath} . '.pg';
```
drgrice1, line 103: "doc" => "tutorial"

htdocs/js/SampleProblemViewer/sample-problem.js
```diff
@@ -0,0 +1,7 @@
+for (const pre of document.body.querySelectorAll('pre.CodeMirror')) {
+	CodeMirror.runMode(pre.textContent, 'PG', pre);
+}
+
+for (const btn of document.querySelectorAll('.clipboard-btn')) {
+	btn.addEventListener('click', () => navigator.clipboard.writeText(btn.dataset.code));
```
drgrice1, line 6: This also should have the check that navigator.clipboard is defined.

lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm
```diff
@@ -0,0 +1,122 @@
...
+
+Render the requestedSampleProblem or one of the indexes.
+
+=cut
+
+sub renderSampleProblem ($c) {
+	my $pg_root = $c->ce->{pg_dir};
+
+	(undef, my $macro_files) = Pod::Simple::Search->new->inc(0)->survey("$pg_root/macros");
+	$c->log->info($c->dumper($macro_files));
```
drgrice1, line 48: Please delete this line.

PR #2067, opened 2023-06-20
Update the POD generation scripts and add POD routes.

drgrice1:
The webwork2 routes serve dynamically generated POD directly generated from the PG files on the server.  The advantage of this is that the link to the POD in the PG problem editor will always be to POD for PG that is for the currently installed PG version.

Also add bootstrap and the pod css to the pages generated by the generate-ww-pg-pod.pl script.
    
Remove POD::Find usage and use POD::Simple::Search instead.  POD::Find is deprecated, and recommends using POD::Simple::Search instead.

This builds on top of #2057.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -128,6 +128,8 @@ my @modulesList = qw(
 	PadWalker
 	Path::Class
 	PHP::Serialization
+	Pod::Simple::Search
+	Pod::Simple::XHTML
 	Pod::Usage
 	Pod::WSDL
 	Safe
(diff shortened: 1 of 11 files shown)


drgrice1:
Another advantage of this is that someone working on PG POD, for example while creating a new macro, can see the generated html super easy.  Just open the page in webwork.

pstaabp:
@drgrice1 This looks very nice.  Both the built-in POD viewer linked from the PGeditor and the resulting POD.

Alex-Jordan (approved):

Alex-Jordan:
@pstaabp If you officially approve, please merge and we knock down two more open PRs.

PR #2069, opened 2023-06-20
tweaks to Library Browser

Alex-Jordan:
This clears up some things in the library Browser that can be confusing to the uninitiated.

It changes the `Local Problems` button to `File Manager`. There was some confusion here that `Local Problems` corresponded to the `local/` folder.

Related, the string `My Problems` is currently used to look at problems in the root `templates/` folder, but there has been confusion about what that means. Is there a folder called "My Problems"? No. Has WeBWorK actually tracked "my" problems that I have written? No. This changes that string to `[TMPL]`. (If there is something better, I'm happy to change it.)

This also changes the `From This Course` button to `Course Sets`. There has been confusion about what "From This Course" means. That could mean problem files in this course's File Manager, set definition files in this course's File Manager, or (what it really means) live sets in this course.

This changes the placeholder text `Select a Problem Collection` to `Select a Folder`, making it more clear what you are selecting. A "collection" might suggest some sort of formal data structure. for example, you are not selecting a "collection" like the full tree hierarchy of the OPL. You will just be looking at .pg files that are direct children of the folder you choose.

Lastly, this disables the placeholder text for several HTML select elements.

### DIFF
--- a/templates/ContentGenerator/Instructor/SetMaker/browse_course_sets_panel.html.ep
+++ b/templates/ContentGenerator/Instructor/SetMaker/browse_course_sets_panel.html.ep
@@ -8,7 +8,7 @@
 		<%= select_field library_sets => [
 				@$local_sets == 0 ? [ maketext('No sets in this course yet') => '' ]
 				: (
-					$selected_library eq '' ? [ maketext('Select a Homework Set') => '', selected => undef ] : (),
+					$selected_library eq '' ? [ maketext('Select a Homework Set') => '', selected => undef, disabled => 'disabled' ] : (),
 					map { [ format_set_name_display($_) => $_, $_ eq $selected_library ? (selected => undef) : () ] }
 					@$local_sets
 				)
--- a/templates/ContentGenerator/Instructor/SetMaker/browse_local_panel.html.ep
+++ b/templates/ContentGenerator/Instructor/SetMaker/browse_local_panel.html.ep
@@ -6,17 +6,17 @@
 <div class="InfoPanel">
 	<div class="mb-2">
 		<%= label_for library_sets =>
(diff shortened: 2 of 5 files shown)


pstaabp:
Overall, I like these changes. 

I don't think `[TMPL]` is understood that much by instructors.  Would `Course Home` be better?

Alex-Jordan:
How about `[course templates folder]`?

 I like that the brackets make it clear that this not actually the name of some folder.

pstaabp:
That's pretty clear.

drgrice1:
Perhaps just `[templates folder]` be sufficient?

Alex-Jordan:
`[templates folder]` is good. One of us could make that edit here in GitHub before merging. I'm a little disoriented with my local networks all down, and my own computer is a bit of a mess.

drgrice1:
Perhaps we should hold off a bit on making the `[TMPL]` to `[templates folder]` change.  There is a little more involved than just changing this here.  There are a few other instances where that is used as well, and they all should be changed.  Also, there are places in PG where this is used as well, so those also should be changed.

Alex-Jordan:
Sure, but this PR would have no "TMPL" at all (from start to finish) if we change that here. What I mean is, right now this PR changes `My Problems` to `[TMPL]`. We could change the PR so it is changing `My Problems` to `[templates folder]`.  If we think "TMPL" is bad, we'd just be introducing another instance.

drgrice1:
I see.  I didn't see that this was an added instance.  I thought it was an existing instance that would be changed.  I didn't look that closely.

drgrice1:
I made the change.

pstaabp (approved):

PR #2049, opened 2023-06-13
Header footer

Alex-Jordan:
This is on top of #2032.

It removes `hardcopySetFooter.pg`, so the only PG problem associated with a hardcopy infrastructure is the header file. It's replaced with a tex file for the footer of a set.

It brings back `hardcopySetHeader.tex`, which I think @drdrew42 recently retired. But read on...

Now `Hardcopy.pm` writes LaTeX macros for various environment variables you might want to call in the LaTeX infrastructure. So for example `hardcopySetHeader.tex` can use things like `\webworkFirstName` to get the first name of the user. And we don't need a lot of tex-specific stuff in the PG header file.

This also allows `hardcopySetHeader.tex`  to start two column mode after the PG header file is run, so the header content can still be in the one-column portion of the PDF. Then `hardcopySetFooter.tex` closes two column mode, if needed.

I haven't tested against the XeLaTeX themes, but I tested with oneColumn and twoColumn.

Here is a screenshot to see a sample. Maybe I was too radical rearranging header and footer items, but it's a starting point.
Screen Shot 2023-06-13 at 2 32 50 PM

This also changes the text of the copyright footer and makes it a running footer, not just a footer for the last page/problem.

Screen Shot 2023-06-13 at 2 35 46 PM

PR #2070, opened 2023-06-20
Fix hardcopy generation in the PG problem editor.

drgrice1:
This just writes dummy empty TeX macros that are needed by the inclusion of headandfoot.tex.  Then it overrides the headers set by that file. Then it writes the hardcopySetHeader.tex file so that two column mode is started if that theme is selected.  After writing the problem TeX, it writes the hardcopySetFooter.tex file so that two column mode is appropriately ended.  Since that file includes the copyright.tex file, these files will now have the logo and copyright at the bottom.

All of the common files are also added to the zip file if TeX Source is selected or if PDF file generation fails.

Note that when a set header is being edited, you will not get values for the TeX macros.  So hardcopy generation is not quite complete for those.  This is just a quick fix to get things working again.

### DIFF
--- a/lib/HardcopyRenderedProblem.pm
+++ b/lib/HardcopyRenderedProblem.pm
@@ -128,7 +128,7 @@ sub generate_hardcopy_tex {
 	# Copy the common tex files into the working directory
 	my $ce         = $ws->c->ce;
 	my $common_dir = path($ce->{webworkDirs}{texinputs_common});
-	for (qw{packages.tex CAPA.tex PGML.tex}) {
+	for (qw{packages.tex CAPA.tex PGML.tex headandfoot.tex copyright.tex webwork_logo.png}) {
 		eval { $common_dir->child($_)->copy_to($working_dir) };
 		push(@$errors, qq{Failed to copy "$ce->{webworkDirs}{texinputs_common}/$_" into directory "$working_dir": $@})
 			if $@;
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Tested and it works. But the comments you wrote are specific to the current themes. I'll suggest some edits.

lib/HardcopyRenderedProblem.pm
```diff
@@ -210,7 +210,43 @@ sub write_tex {
 		. ($ws->{inputs_ref}{hardcopy_theme} // $ce->{hardcopyTheme});
 
 	write_tex_file($FH, $ce->{webworkFiles}{hardcopySnippets}{preamble} // "$themeDir/hardcopyPreamble.tex", $errors);
+
+	# Write dummy LaTeX macros to pacify the inclusion of headandfoot.tex in the preamble.
```
Alex-Jordan, line 214: Not just because of headandfoot.tex in the preamble, but these macros might be used (in some future theme) by the setTexHeader and setFooter.

```suggestion
	# Write dummy LaTeX macros possibly called by preamble, setTexHeader, and setFooter
```

lib/HardcopyRenderedProblem.pm
```diff
@@ -210,7 +210,43 @@ sub write_tex {
...
+		ReducedScoringDate
+		DueDate
+		AnswerDate
+		PrettySetId
+	))
+	{
+		print $FH "\\def\\webwork${_}{}\n";
+	}
+
+	# Override the headers and footers set in headandfoot.tex.
```
Alex-Jordan, line 238: The Hebrew themes are an example where headandfoot.tex won't be called. (Maybe headers and footers will be set directly by those theme's preamble file.)

```suggestion
	# Clear any headers and footers.
```

lib/HardcopyRenderedProblem.pm
```diff
@@ -210,7 +210,43 @@ sub write_tex {
...
+	print $FH "\\firstpageheader{}{}{}\\footer{}{}{}";
+
+	# Now write hardcopySetHeader.tex so that two column mode is started if that theme is selected.
```
Alex-Jordan, line 241: A custom theme might do open other environments.

```suggestion
	# Now write hardcopySetHeader.tex, possibly opening environments.
```

lib/HardcopyRenderedProblem.pm
```diff
@@ -210,7 +210,43 @@ sub write_tex {
...
+	write_tex_file($FH, $ce->{webworkFiles}{hardcopySnippets}{setTexHeader} // "$themeDir/hardcopySetHeader.tex",
+		$errors);
+
 	write_problem_tex($ws, $FH);
+
+	# Now write hardcopySetFooter.tex so that two column mode is ended if that theme is selected.
```
Alex-Jordan, line 247:
```suggestion
	# Now write hardcopySetFooter.tex, closing environments from setTexHeader.
```

lib/HardcopyRenderedProblem.pm
drgrice1: I just looked at how this works.  The headandfoot.tex file and its usage is generally a problem, and I should have looked closer when I reviewed #2049.  It should not be in common, and really shouldn't have been done in that way.  It is not translatable, and this approach needs revision.  Generally that pull request crosses boundaries with the hardcopy themes in troublesome ways.  Thus resulting in the need for this pull request among other things.

drgrice1:
I fixed the comments.

lib/HardcopyRenderedProblem.pm
Alex-Jordan: This makes me wonder if you think of "common" as files that _all_ themes use. I am thinking of it as a place where you can hold content that _multiple_ themes (but maybe not all) make use of. For example not being translatable was not a concern to me. A non-english theme would just have the header and footer set directly in the preamble tex file and not use `\input{headandfoot.tex}`.

Would reverting the `headandfoot.tex` directly back into each theme address your concern here? I had it that way, but I was annoyed that each small revision required editing six files.

drgrice1: > This makes me wonder if you think of "common" as files that _all_ themes use.

That was the intent when I created the "common" directory.

> Would reverting the `headandfoot.tex` directly back into each theme address your concern here? I had it that way, but I was annoyed that each small revision required editing six files.

Not really.  The real problem is the crossover of content into the the theme files.

Alex-Jordan:
I have an unrelated branch that I was going to push tonight or tomorrow and
open a PR for. I think it happens to have something in it that will help
straighten things out.

PR #2072, opened 2023-06-21
webwork.sty file for WeBWorK-specific LaTeX macros

Alex-Jordan:
This is an alternative to #2070 in that II think it fixes the hardcopy generation of an individual problem that was broken by #2049.

`headandfoot.tex` is no more. The six styles directly declare headers and footers using
```
\pagestyle{headandfoot}
\firstpageheader{\webworkAssignmentCourseInfo}{}{\webworkUserInfo}
\runningheader{}{}{}
\footer{}{}{\webworkPageNumber}
```

Those tex macros like `\webworkAssignmentCourseInfo` are defined in the new `webwork.sty` file, which also defines empty macros for things like `\webworkCourseName` that will be written over in the case of an actual set hardcopy being built. The macros like `\webworkAssignmentCourseInfo` should be doing the right thing when the macros like `\webworkCourseName` are empty.

Also `webwork.sty` provides a crude way to localize short words and phrases that are used within macros like `\webworkAssignmentCourseInfo`. I successfully localized into Pig Latin:

Screen Shot 2023-06-20 at 11 29 13 PM

I'll need @taniwallach to try this out with Hebrew localization to see if it really works as intended in that rtl+xelatex situation. But I left notes in the Hebrew themes for what to localize. 

FWIW,  on another branch I also have `pg.sty` where I'd like to put PG latex macros. Stuff that makes sense in PG without webwork2, unlike what `webwork.sty` has in it. For starters I'd replace `\\mbox{\\parbox[t]{${tcol}ex}{\\hrulefill}}` with `\\answerRule{$tcol}`, a macro defined in `pg.sty`. This will allow me to redefine that macro as is needed when an answer blank is inside a table and the underline looks terrible alongside the table's horizontal rules.


Alex-Jordan:
If it's verified that this works with an rlt language like Hebrew, then we could automate the localization in the usual way. The two harcopy-making .pm files could write localizations into the tex file at about the same place things like `\webworkCourseName` are written.

Alex-Jordan:
I fixed the typo/nonexistent words. And renamed the sty file to `webwork2.sty`. (I wasn't sure a LaTeX package could have a digit in its name, but it tested out.) This will keep this clear as a webwork2 thing for webwork2 hardcopies.

pstaabp (approved):
Fixes the issue of hardcopy generation in the PGeditor.  

Should we wait to merge for the Hebrew localization?

drgrice1:
I think go ahead and merge this.  @taniwallach can put in a pull request when he gets a chance to work on it.

PR #2073, opened 2023-06-21
remove a tex comment when it has nothing to say

Alex-Jordan:
This is a small piece of clutter I noticed with recent things that have me looking at hardcopy tex files. It was printing

```
%% decoded old answers, saved. (keys = 
```

with nothing more. this closes thee paren, and prints nothing now if that hash is empty.


drgrice1 (approved):
Straightforward enough.  Looks good.  This tracks back to when this code was added in 2006.

drgrice1:
Merging with two approvals.

Issue #2074, opened 2023-06-21
OAuth verification with 2.18 LTI


Alex-Jordan:
My school uses a load balancer in front of the WW server. The load balancer handles all https stuff and communicates with the actual WW server using port 80.

In site.conf, I still declare thee root URL using https. This is important for various links to be created with the "https".

In the mojolicious yml file, I have it listening on port 80 for the load balancer to talk to it. I am serving with hypnotoad.

Regular access to courses works fine, but with an connection from the LMS, I was getting this error:
```
The following path was reconstructed by WeBWorK. It should match the path in the LMS:
http://webwork.pcc.edu/webwork2/redacted/
OAuth verification failed. Check the Consumer Secret and that the URL in the LMS exactly matches the WeBWorK URL as defined in site.conf. E.G. Check that if you have https in the LMS url then you have https in $server_root_url in site.conf at /opt/webwork/webwork2/lib/WeBWorK/Authen/LTIAdvanced.pm line 506.
```
Note that the LMS is using a URL with "https" as well, which it needs to do.

But at some stage, WeBWorK gets the "http" request from the load balancer, and that is responsible for constructing the "http" path that you see above.

I hacked this by editing `lib/WeBWorK/Authen/LTIAdvanced.pm` and adding `$path =~ s/^http:/https:/;` at the appropriate place. And now my LTI connections work. But can we figure out the "right" way to handle this situation?

somiaj:
I think there is a setting for this to allow renaming of the url if needed. Have you seen if that setting works. Form the `conf/authen_LTI_1_1.conf`.


```perl
# This allows you to override the URL that Oauth will use to validate the authentication.  This
# is important if you have some sort of setup (e.g. load distributing) where the path in the LMS
# does not match the path that ends up in the webwork page.
$LTI{v1p1}{BasicToThisSiteURL} = '';
```

Alex-Jordan:
There are a few issues.

* the line you cite is no longer in `conf/authen_LTI_1_1.conf`
* I see now the key was renamed to `OverrideSiteURL`
* nothing in `lib/...` has "OverrideSiteURL"
* `LTIAdvanced.pm` still has code based on `$LTIBasicToThisSiteURL`, so i was searching config files for `BasicToThisSiteURL` and came up empty.
[...]

PR #2075, opened 2023-06-21
Fix the OverrideSiteURL in the code.

drgrice1:
Somehow, when I changed the variable in the config file, I didn't also change the variable in the code.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -455,7 +455,7 @@ sub authenticate {
 
 	# We need to provide the request URL when verifying the OAuth request.
 	# We use the url request by default, but also allow it to be overriden
-	my $path = $ce->{LTI}{v1p1}{BasicToThisSiteURL} || ($c->url_for->to_abs =~ s|/?$|/|r);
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
This is good. But see the discussion at #2074. I'm not sure this is a good way to handle overriding the path in most cases. The config variable might need to be something different for each course, for each ELT/app link from the LMS.

drgrice1:
Each course can set the variable to a different value in the course.conf file.  If the only need is to change http to https, that can be accomplished by the config variable.  There is no need for a code reference.  Generally, we need to move away from using such things in the config at all.  Eventually (webwork3) there will be none of that allowed.

Alex-Jordan:
I'd like to have a way to handle this site-wide. I don't want to manage all the courses' config files for this.

These URLs are more complicated when there is homework grade passback, or an instructor wants the student to go directly to a particular page from the URL. So in that situation there is no one single value you can use for this config variable. Unless the OAuth is truncating those URL paths anyway down to just the course?

drgrice1:
Regardless of any further improvements that you may desire, this pull request should go through for now.  It obviously is needed.  If you want further improvements for your use case, then I suggest that you submit a pull request.  I don't know what you need.

Alex-Jordan:
Agreed, that's why I approved this already.

drgrice1:
Sorry, I didn't see the approval.

pstaabp (approved):

PR #2078, opened 2023-06-23
One column

Alex-Jordan:
There are two commits here. One removes `BEGIN/END_ONE_COLUMN` from all pg files that come with the distribution. The other removes all `beginproblem()` from pg files that come with the distribution. These are all deprecated now and the distribution should not use them.

The four sets that come with model course should be redone (I think into just two sets: an orientation and a demo). But that is a later project.

### DIFF
--- a/conf/snippets/blankProblem.pg
+++ b/conf/snippets/blankProblem.pg
@@ -11,7 +11,6 @@ loadMacros("PG.pl",
 
 ## Do NOT show partial correct answers
 $showPartialCorrectAnswers = 0;
-TEXT(beginproblem());
 
 BEGIN_TEXT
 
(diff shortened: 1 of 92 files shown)


pstaabp (approved):
👍

drgrice1 (approved):
This looks good.

The PG codemirror editor will still highlight BEGIN_ONE_COLUMN and END_ONE_COLUMN.  Those are the only instances to these in the webwork2 code left.

We should consider removing the `$pg{specialEnvironmentVars}{problemPreamble}` and `$pg{specialEnvironmentVars}{problemPostamble}` configuration variables.  Now that we are discouraging `beginproblem` usage the `problemPreamble` will become unreliable since that method is what injects the `problemPreamble` into the output.  They have been noted as "becoming deprecated" since 2013.

Alex-Jordan:
@drgrice1 Should those removals happen here or in a separate PR? This one affects so many files it is nice to keep it as targeted in scope as possible.

drgrice1:
A separate pull request would be good for that. I was just noting it for future to-do.

pstaabp:
I'll go ahead and merge.

PR #2079, opened 2023-06-23
give message if someone tries to visit an archived (and closed) course

Alex-Jordan:
If a course `course_name` has been archived and deleted, so the archives folder has `course_name.tar.gz`, this makes it so that when a user visits `webwork.mysite.org/webwork2/course_name`, they get a message "This course has been archived and closed." instead of the message "This course does not exist.".


drgrice1 (approved):
This looks good.

pstaabp (approved):
👍

PR #2080, opened 2023-06-23
LTI 1.1 override protocol and domain only for OAuth

Alex-Jordan:
This allows you to override the protocol/scheme (http vs https) and the domain for the URL that is used with LTI 1.1 for OAuth without changing the rest of the URL.

I have this working in production. I have
```
$LTI{v1p1}{OverrideSiteProtocolDomain} = 'https://webwork.pcc.edu';
```
and I can get into my WW course from the LMS, even after actively logging out from WW. But if I alter or comment out that config setting, I get "The course XXXXXXXXXXXXX uses an external authentication system ([Desire2Learn](https://online.pcc.edu/)). Please return to that system to access this course." The hack i previously had in place is removed.

### DIFF
--- a/lib/Mojolicious/WeBWorK.pm
+++ b/lib/Mojolicious/WeBWorK.pm
@@ -91,8 +91,8 @@ sub startup ($app) {
 
 	# Helpers
 
-	# This replaces the previous Apache2::RequestUtil method that was overriden in the WeBWorK::Request module to return
-	# the empty string for '/'.
+	# This replaces the previous Apache2::RequestUtil method that was overridden in
+	# the WeBWorK::Request module to return the empty string for '/'.
 	$app->helper(location => sub ($) { return $webwork_url eq '/' ? '' : $webwork_url });
 
 	$app->helper(server_root_url => sub ($) { return $server_root_url; });
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -28,7 +28,7 @@ use warnings;
 
 use Carp;
 use DBI;
-use URI;
+use Mojo::URL;
 use URI::Escape;
 use Net::OAuth;
 
@@ -455,12 +455,12 @@ sub authenticate {
 	my $requestHash = \%request_hash;
 
 	# We need to provide the request URL when verifying the OAuth request.
-	# We use the url request by default, but also allow it to be overriden
-	my $url = URI->new($c->url_for->to_abs);
+	# We use the url request by default, but also allow it to be overridden
+	my $url = $c->url_for->to_abs;
 	if ($ce->{LTI}{v1p1}{OverrideSiteProtocolDomain}) {
-		my $override = URI->new($ce->{LTI}{v1p1}{OverrideSiteProtocolDomain});
-		$url->scheme($override->scheme());
-		$url->host($override->host());
+		my $override = Mojo::URL->new($ce->{LTI}{v1p1}{OverrideSiteProtocolDomain});
+		$url->scheme($override->scheme);
+		$url->host($override->host);
 	}
 	my $path = $ce->{LTI}{v1p1}{OverrideSiteURL} || ($url =~ s|/?$|/|r);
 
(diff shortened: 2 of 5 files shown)


lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -455,7 +456,13 @@ sub authenticate {
...
 	# We need to provide the request URL when verifying the OAuth request.
 	# We use the url request by default, but also allow it to be overriden
-	my $path = $ce->{LTI}{v1p1}{OverrideSiteURL} || ($c->url_for->to_abs =~ s|/?$|/|r);
+	my $url = URI->new($c->url_for->to_abs);
+	if ($ce->{LTI}{v1p1}{OverrideSiteProtocolDomain}) {
+		my $override = URI->new($ce->{LTI}{v1p1}{OverrideSiteProtocolDomain});
+		$url->scheme($override->scheme());
+		$url->host($override->host());
+	}
+	my $path = $ce->{LTI}{v1p1}{OverrideSiteURL} || ($url =~ s|/?$|/|r);
```
drgrice1, line 465: There is no reason to use the `URI` module.  A Mojo::URL object already can do everything that the `URI` module can do, and is what `$c->url_for->to_abs` returns.  So this should be
```suggestion
	my $url = $c->url_for->to_abs;
	if ($ce->{LTI}{v1p1}{OverrideSiteProtocolDomain}) {
		my $override = Mojo::URL->new($ce->{LTI}{v1p1}{OverrideSiteProtocolDomain});
		$url->scheme($override->scheme);
		$url->host($override->host);
	}
	my $path = $ce->{LTI}{v1p1}{OverrideSiteURL} || ($url =~ s|/?$|/|r);
```

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -28,6 +28,7 @@ use warnings;
 
 use Carp;
 use DBI;
+use URI;
 use URI::Escape;
```
drgrice1, line 32: Don't forget to change this to `use Mojo::URL;` after switching to Mojo::URL usage.  Although, technically you don't need to `use Mojo::URL` to use it in the file.

lib/WeBWorK/Authen/LTIAdvanced.pm
Alex-Jordan: Thanks, I made these changes.

drgrice1 (approved):
This looks good.

lib/Mojolicious/WeBWorK.pm
```diff
@@ -91,7 +91,7 @@ sub startup ($app) {
 
 	# Helpers
 
-	# This replaces the previous Apache2::RequestUtil method that was overriden in the WeBWorK::Request module to return
+	# This replaces the previous Apache2::RequestUtil method that was overridden in the WeBWorK::Request module to return
 	# the empty string for '/'.
```
drgrice1, line 95: Change this to
```perl
	# This replaces the previous Apache2::RequestUtil method that was overridden in
	# the WeBWorK::Request module to return the empty string for '/'.
```
to fix the perltidy issue.  Adding the one "d" was enough to push this over the 120 character limit.  When perltidy encounters a comment that exceeds the limit it decreases indentation to try to fit it in.

Alex-Jordan:
I dared to try fixing a common typo in the last commit here, and the linter wants to do something strange. It wants to replace  leading tabs on two lines of comments with four spaces. Does this make sense?

lib/Mojolicious/WeBWorK.pm
Alex-Jordan: Thanks, slowly growing my understanding pf perltidy.

drgrice1: We have many really long comment lines in the code.  Those really mess up formatting and need to be fixed.

This particular comment was within the limit though until you fixed my typo!

Sad that two of the five "overridden" typos were mine.  I can't believe I did that.

pstaabp (approved):

PR #2081, opened 2023-06-23
infrastructure for pg.sty

Alex-Jordan:
This will be paired with a PR to pg.

In this PR, I am unsure that I handled the file copying commands well. I copied what was there for copying from `conf/snippets/common/`, but I altered it as I though appropriate for just copying a single file.

### DIFF
--- a/bin/check_latex
+++ b/bin/check_latex
@@ -49,7 +49,8 @@ die $@ if $@;
 my $pdflatex_cmd =
 	"cd $temp_dir && "
 	. "TEXINPUTS=$ENV{WEBWORK_ROOT}/bin:"
-	. shell_quote($ce->{webworkDirs}{texinputs_common}) . ': '
+	. shell_quote($ce->{webworkDirs}{texinputs_common}) . ':'
+	. shell_quote("$ce->{pg}{directories}{assets}/tex/") . ': '
 	. $ce->{externalPrograms}{pdflatex}
 	. ' -interaction nonstopmode check_latex.tex > check_latex.nfo 2>&1';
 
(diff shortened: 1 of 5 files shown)


drgrice1:
Perhaps it is time to replace the flat line answer rules with a box.  A box would be more like what is actually seen in html, and in many cases would look better than the line.  One particular case where lines look really bad is when answer rules are used in formulas.  For example, answer rules that fit into the numerator and denominator of a fraction, or the limits and integrand of an integral.

Alex-Jordan:
I think by the time this is where I envision it, someone could make that
kind of customization to the latex package at the course or site level.
FWIW I'm not opposed to boxes but I have empirical evidence from PTX
authors that many people are.

drgrice1 (approved):

Alex-Jordan:
A note about an outlined box from the other thread: if the printed hardcopy is something a student will actually write on, then an outlined box that is uniformly one baselineskip tall (or something like that) gets in the way of handwriting something like a fraction, rational expression, or otherwise tall answer.

drgrice1:
Yeah, that is a problem with using boxes. I am accustomed to having a height argument in what I do where I use boxes, and that allows one to use taller boxes in those cases. Webwork answer rules don't have that, so that would make it impossible to make that adjustment.

drgrice1:
What I do works for my purposes.  The boxes provide separation on handwritten tests from the rest of the work, and are big enough for what is needed.  For webwork2, I realize that the boxes are not appropriate.  Even if you had the height setting, it would be unreliable for general purposes.

pstaabp (approved):

PR #2082, opened 2023-06-23
add mhchem package to latex packages

Alex-Jordan:
@limefrogyank has some WW chemistry macros that might be coming to `pg` or the OPL in the coming year. This tex package will support the hardcopy output for those.

Note that if you don't specify a version, this package issues a runtime warning. So we set it to the current version 4. Version 4.09 is the latest, from December 2021.


pstaabp (approved):
Not sure how to test this, but should be fine.

drgrice1 (approved):
This should be okay.  However, it would be nice to have a mechanism that a problem author could add a LaTeX package.  Something for hardcopy like the `HEADER_TEXT` method does for html.

PR #2083, opened 2023-06-25
POD Viewer improvements

drgrice1:
First, improve the efficiency of the WeBWorK::Utils::PODPaser package.

The $podFiles argument must be provided to the new method in order for the module to resolve POD links to local files.  It should be the first return value of the POD::Simple::Search survey method if called in list context, or only return value if called in scalar context.

This is opposed to the previous method of search the root directory for each link that was encountered in the POD.  Now instead, the directory is searched for files containing POD only once, and the results are stored in a hash.  This is done by the caller and passed to the WeBWorK::Utils::PODPaser object on construction, instead of the WeBWorK::Utils::PODPaser object handling it.

The module now expects `L<...>` usage to be correct according to the documentation at https://perldoc.perl.org/perlpod#L-a-hyperlink.  Note that it also now honors links to POD head sections in other local files.

The module also honors the following special cases that fit into that scheme for local files:

* For a link to a local package, use the package name (exactly as the rules for `L<...>` dictate).
* For a link to a `.pod` file, just use the basename without the `.pod` extension.  So for example for a link to doc/MathObjects/UsingMathObjects.pod, use `L<UsingMathObjects>`.
* For a link to a macro file (or script), use the basename of the macro file with the `.pl` extension and no path.

The generate-ww-pg-pod.pl script now uses Mojo::Template to make it easier to make html.

The POD page layouts have been improved with a sidebar for indices. Mojo::DOM is used to parse the generate POD and pull the index out to place in the sidebar.

Also add a layout template for the POD rendered by webwork2.

This needs https://github.com/openwebwork/pg/pull/864 for some of the PG POD links to work right.

### DIFF
--- a/bin/dev_scripts/PODtoHTML.pm
+++ b/bin/dev_scripts/PODtoHTML.pm
@@ -17,20 +17,25 @@ package PODtoHTML;
 
 use strict;
 use warnings;
+use utf8;
 
 use Pod::Simple::Search;
+use Mojo::Template;
+use Mojo::DOM;
+use Mojo::Collection qw(c);
 use File::Path qw(make_path);
+use File::Basename qw(dirname);
 use IO::File;
 use POSIX qw(strftime);
 
 use WeBWorK::Utils::PODParser;
 
 our @sections = (
-	bin    => "Scripts",
-	conf   => "Config Files",
-	doc    => "Documentation",
-	lib    => "Libraries",
-	macros => "Macros"
+	bin    => 'Scripts',
+	conf   => 'Config Files',
+	doc    => 'Documentation',
+	lib    => 'Libraries',
+	macros => 'Macros'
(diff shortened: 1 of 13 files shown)


pstaabp:
Looks good.  There's a warning: `Wide character in print at bin/dev_scripts/PODtoHTML.pm line 123.` Looks like it is probably coming from the unicode characters in `contextCongruence.pl` and `parserLinearRelation.pl`. Seems like everything looks like--not sure why it is throwing that warning.

drgrice1:
It was giving the warning because those macros use unicode characters and I forgot to set the IO encoding to UTF-8 for the file IO in `PODtoHTML.pm`.

pstaabp (approved):
warning fixed.  Looks good.

PR #2084, opened 2023-06-26
after upgrading courses, don't scan all courses again

Alex-Jordan:
I think this is a good change, but not sure if all will agree.

When there are many courses, visiting the upgrade tool in the admin course takes a while for it to scan all courses looking for database tables to upgrade. Later when it's all done doing the upgrades, you click Done, and it reloads the same page, repeating that scan that takes so much time.

This changes it to go to the Course Listings page instead after clicking Done.


drgrice1 (approved):
I agree with this.  I never like that it went back to the upgrade page.  It seemed unintuitive to me.  It seems natural that "Done" should go where this makes it go.

pstaabp (approved):
👍

PR #2085, opened 2023-06-26
Student answers

Alex-Jordan:
This is another small change that I would like but I'm unsure if people like it the way it is now. This makes it so that when you go to print hardcopy, none of the optional components are checked. Presently, the student answers box is checked. This usually leads (for me) to a bunch of empty lists with spaces for answers that haven't been entered yet. If I am printing hardcopy, it is more often before students have had a chance to enter anything. Also this seems more clean in that none of the options are checked, instead of one standing out.


drgrice1 (approved):
This change is fine with me.  Others should weigh in on this though.

somiaj (approved):
I would also prefer this unchecked by default.

PR #2087, opened 2023-06-27
Fix scroll positioning issues.

drgrice1:
Making the masthead fixed has caused some issues with scroll positioning.

First, if you click on a fragment link, the window scrolls the element that is referenced to the top of the page.  Unfortunately, that makes it under the masthead.  This adds a `scroll-margin-top` to all elements with ids in the main content of the page.  One example of this is the "Skip to main content" accessibility link.  If you load a page and hit the "tab" key immediately after the page loads this will appear.  If you then hit "enter", the "page-title" should be focused and scrolled to the top.  Without this pull request the page title will be under the masthead.  With this pull request it will be right under it.  This will also fix any other fragment links like this as well though.

Also on the PG problem editor page to fix scrolling to the top of the problem when the "View/Reload" button is clicked, the masthead height needs to be taken into account.  To test this open a problem in the editor, and scroll as far down as you can.  Make the window shorter if needed to make sure that the top of the problem is above the top of the window.  Then click "View/Reload".  With this pull request the window should scroll so that the top of the problem is visible below the masthead.  Withouth this pull request it will be obscured below the masthead.

### DIFF
--- a/htdocs/js/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/PGProblemEditor/pgproblemeditor.js
@@ -228,7 +228,8 @@
 
 		// Scroll to the top of the render window if the current scroll position is below that.
 		const renderAreaRect = renderArea.getBoundingClientRect();
-		if (renderAreaRect.top < 0) window.scrollBy(0, renderAreaRect.top);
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
Looks good.

Alex-Jordan (approved):

PR #2089, opened 2023-06-27
fontspec in hardcopy preamble

Alex-Jordan:
I use the `xelatex` for making hardcopy, initially just because some problems had things like π directly in their text.

I do not have a need for `xunicode` or `polyglossia`, so the distribution `XeLaTeX-***column` hardcopy themes seem like they are not what I should be using. But if using `xelatex`, then the `one-column` and `two-column` distribution themes fail to show font variants like bold, small caps, etc.

By adding the `fontspec` package, that is fixed. It's harmless for `pdflatex` sites, but lets these themes work when `$externalPrograms{pdflatex}` (a misnomer) is `xelatex --no-shell-escape`.

Try changing `$externalPrograms{pdflatex}` to `xelatex --no-shell-escape` and make a hardcopy before this PR and you should see the absence of bold. But with this commit, it comes back.

### DIFF
--- a/conf/snippets/hardcopyThemes/oneColumn/hardcopyPreamble.tex
+++ b/conf/snippets/hardcopyThemes/oneColumn/hardcopyPreamble.tex
@@ -4,6 +4,10 @@
 
 \usepackage[text={7.5in,9in},centering]{geometry}
 
+% If the "pdflatex" external program is set to xelatex in site.conf,
+% this is needed for bold, small caps, etc variants of Compute Modern
+\usepackage{fontspec}
+
 % This removes the margin for questions from the exam class.  At this point we only use questions and not parts or
 % subparts, but if those are used a similar thing will be needed for those as well.
 \renewcommand{\questionshook}{\leftmargin=0pt\labelwidth=-\labelsep}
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I suppose if this is OK, then maybe `fontspec` belongs in `common/packages.tex`.

Alex-Jordan:
I did some more investigating about this. All while using `xelatex`...

If I put `\usepackage{fontspec}` in `common/packages.tex` anywhere before `\usepackage{pslatex}`, the bold, small caps, etc fails to show up in the PDF. If I put it _after_ `\usepackage{pslatex}` then it all shows up.

So this seems to be some sort of package clash. I don't see any alerts in the tex console log and I'm not sure what is the root cause is. But if this `fontspec` inclusion is OK across the board, and if we put it in `packages.tex`, it needs to be placed after `pslatex`. And I'm not sure what `pslatex` is for and if `fontspec` in turn messes it up.

Alex-Jordan:
See my [question](https://tex.stackexchange.com/questions/689672/boldface-and-other-things-fail-when-xelatex-used-with-pslatex-package?noredirect=1#comment1711214_689672) and David Carlisle's response. I think we can drop `pslatex`.

And then I went looking and I think we can drop `epsf` and `epsfig` as well.

PR #2090, opened 2023-06-27
only show reduced credit date in set header when relevant

Alex-Jordan:
This fixes one essential thing: the new hardcopy header was showing a reduced credit date even when the set did not have reduced credit enabled.

And one other thing: the new hardcopy header was showing a reduced credit date even when it was equal to the close date.

This commit fixes both of these things.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -962,25 +962,25 @@ async sub write_set_tex ($c, $FH, $TargetUser, $setID) {
 
 	# write environment variables as LaTeX macros
 	for (qw(user_id student_id first_name last_name email_address section recitation)) {
-		print $FH '\\def\\webwork' . underscore_to_camel($_) . '{' . handle_underbar($TargetUser->{$_}) . "}\n";
+		print $FH '\\def\\webwork' . underscore_to_camel($_) . '{' . handle_underbar($TargetUser->{$_}) . "}\n"
+			if $TargetUser->{$_};
 	}
 	for (qw(set_id description)) {
-		print $FH '\\def\\webwork' . underscore_to_camel($_) . '{' . handle_underbar($MergedSet->{$_}) . "}\n";
+		print $FH '\\def\\webwork' . underscore_to_camel($_) . '{' . handle_underbar($MergedSet->{$_}) . "}\n"
+			if $MergedSet->{$_};
 	}
 	print $FH '\\def\\webworkPrettySetId{' . handle_underbar($MergedSet->{set_id}, 1) . "}\n";
 	for (qw(open_date due_date answer_date)) {
-		if (!ref($MergedSet->{$_}) && $MergedSet->{$_} =~ /^\d{10}$/) {
+		if ($MergedSet->{$_}) {
 			print $FH '\\def\\webwork'
 				. underscore_to_camel($_) . '{'
 				. WeBWorK::Utils::formatDateTime($MergedSet->{$_}, $ce->{siteDefaults}{timezone}) . "}\n";
-		} else {
-			print $FH '\\def\\webwork' . underscore_to_camel($_) . "{}\n";
 		}
 	}
 	# Leave reduced scoring date blank if it is disabled, or enabled but on (or somehow later) than the close date
-	if ($MergedSet->{enable_reduced_scoring}
-		&& !ref($MergedSet->{reduced_scoring_date})
-		&& $MergedSet->{reduced_scoring_date} =~ /^\d{10}$/
+	if ($MergedSet->{reduced_scoring_date}
+		&& $ce->{pg}{ansEvalDefaults}{enableReducedScoring}
+		&& $MergedSet->{enable_reduced_scoring}
 		&& $MergedSet->{reduced_scoring_date} < $MergedSet->{due_date})
 	{
 		print $FH '\\def\\webworkReducedScoringDate{'
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -974,7 +974,7 @@ async sub write_set_tex ($c, $FH, $TargetUser, $setID) {
 		if ($MergedSet->{$_}) {
 			print $FH '\\def\\webwork'
 				. underscore_to_camel($_) . '{'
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -977,6 +977,15 @@ async sub write_set_tex ($c, $FH, $TargetUser, $setID) {
 			print $FH '\\def\\webwork' . underscore_to_camel($_) . "{}\n";
 		}
 	}
+	# Leave reduced scoring date blank if it is ddisabled, or enabled but on (or somehow later) than the close date
```
pstaabp, line 980: ddisabled -> disabled

pstaabp (approved):
Except for the typo, looks good.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
Alex-Jordan: Sorry, you may (or may not) have noticed more typos from me recently. My keyboard is malfunctioning, with lost of keys double (or triple) registering. And some keys not registering without extra pressure. My school is getting me a new laptop but only after the fiscal year resets.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -977,6 +977,15 @@ async sub write_set_tex ($c, $FH, $TargetUser, $setID) {
...
+	if ($MergedSet->{enable_reduced_scoring}
+		&& !ref($MergedSet->{reduced_scoring_date})
+		&& $MergedSet->{reduced_scoring_date} =~ /^\d{10}$/
```
drgrice1, line 983: These checks are incorrect.  I see that you also added these to Hardcopy.pm in #2049.  I missed that and those should be fixed.  `$MergedSet->{reduced_scoring_date}` comes from the database.  That column is defined to be of type `BIGINT`.  That means that it is impossible for `$MergedSet->{reduced_scoring_date}` to be a `ref`.  So the first check will always be true.  It also means that it will always satisfy the regular expression in the second check.  So both checks will always be true and are not needed.

The same is true for all of the set dates defined in `lib/WeBWorK/DB/Record/UserSet.pm`, and so these checks in Hardcopy.pm should be removed also.

Alex-Jordan: IIRC I hit an issue where `WeBWorK::Utils::formatDateTime` choked on `$MergedSet->{reduced_scoring_date}` in one instance and these checks got me past that.

drgrice1: `WeBWorK::Utils::formatDateTime` would choke on an undefined value.

drgrice1: So both of those checks should be replaced with the single check for `defined $MergedSet->{reduced_scoring_date}`.  That is if a reduced scoring date of `0` is allowed.  If `0` should not be allowed, then drop `defined` from that.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -977,6 +977,15 @@ async sub write_set_tex ($c, $FH, $TargetUser, $setID) {
...
+		&& $MergedSet->{reduced_scoring_date} < $MergedSet->{due_date})
+	{
+		print $FH '\\def\\webworkReducedScoringDate{'
+			. WeBWorK::Utils::formatDateTime($MergedSet->{reduced_scoring_date}, $ce->{siteDefaults}{timezone}) . "}\n";
```
drgrice1, line 987: Here and the other place where this was added in #2049 should be changed to
```suggestion
			. formatDateTime($MergedSet->{reduced_scoring_date}, $ce->{siteDefaults}{timezone}) . "}\n";
```
and then add `formatDateTime` to the list of methods that are imported from the `WeBWorK::Utils` module on line 33.

drgrice1: Actually, what I said about this isn't quite right.  This should be
```suggestion
			. $c->formatDateTime($MergedSet->{reduced_scoring_date}) . "}\n";
```
and don't include `formatDateTime` in the list of methods that are imported from `WeBWorK::Utils`.  The ContentGenerator method just calls `WeBWorK::Utils::formatDateTime` and takes care of filling in the `$ce->{siteDefaults}{timezone}` argument for you as well as the `$ce->{siteDefaults}{locale}` which is missing in your call and should be there.

Alex-Jordan: `formatDateTime` is not a method, it's just a subroutine. I haven't tried but if you call it like `$c->formatDateTime(sometehing)` you trigger the message at line 706 of `Utils.pm`.

Alex-Jordan:
This needs another look with the last commit changing the conditional logic. Note that `formatDateTime` is a general purpose utility and not a method,  and I don't think it can be called like `$c->formatDateTime(...)`.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
drgrice1: Calling `$c->formatDateTimet(...)` will not trigger the warning on line 706 of `Utils.pm`.  This is because calling `$c->formatDateTime` calls the `WeBWorK::ContentGenerator::formatDateTime`  method that is defined on line 1263 of `lib/WeBWorK/ContentGenerator.pm`, and does not directly call the method `WeBWorK::Utils::formatDateTime`.  The first calls the latter, and when the first calls the latter it does not pass its object (`$c`) to the latter.

Both `WeBWorK::ContentGenerator::formatDateTime` (that I am telling you to use here) and `WeBWorK::Utils::formatDateTime` are methods.  The warning on line 706 of that method is using incorrect and confusing terminology.  The word method is a synonym of subroutine.  There are methods (or subroutines) that are intended to be called with a package object as its first argument, and there are methods (or subroutines) that are not intended to be called with a package object as its first argument.  The warning on line 706 is saying that method is not the first kind of method.

Alex-Jordan: OK, I think I am caught up.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
Alex-Jordan: I think I am caught up with everything here.

drgrice1 (approved):
This looks good.  Thanks for making the requested changes.

PR #2091, opened 2023-06-27
Make messages dismissible and PG editor tmp edits more clear

drgrice1:
Any message that is added by the ContentGenerator addgoodmessage or addbadmessage methods is now dismissible.  There is a close button to the right that can be clicked to remove the message from the page.  In additiona, a hidden button is also added to the end of the page.  If there are any dismissible messages, then the button is shown by javascript, and if clicked all of the dismissible messages are removed from the page (and the button hidden again).  This is a convenience for removing all of those messages at once if there are many of them.

In the PG editor, when a temporary file is created by javascript the file name is changed to the temporary file name in the "Editing ..." statement and the `temporaryFile` class is added which gives the text the orange temporary file color.  Furthermore, the "Editing ..." statement is now shown both above and below the codemirror editor in the page.  This makes it more clear that a temporary file is being edited.

Also fix a bug introduced by #1918.  The PG editor revert form needs to be shown for course information files as well.  Otherwise there is no way to revert to the original file once a temporary file is created other than saving the file.  @somiaj: What was your rationale for doing this?

Also add site nav sub links for set headers and the course information file.  Now that the problem editor is directly in the site nav that link is shown as active when editing one of these files.  That is confusing because if you click on that "active" link it does not open the page you are currently viewing that is editing one of these files.  Instead it opens the editor with a new problem.

### DIFF
--- a/htdocs/js/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/PGProblemEditor/pgproblemeditor.js
@@ -54,7 +54,16 @@
 
 		fetch(webserviceURL, { method: 'post', mode: 'same-origin', body: new URLSearchParams(request_object) })
 			.then((response) => response.json())
-			.then((data) => showMessage(data.server_response, data.result_data))
+			.then((data) => {
+				showMessage(data.server_response, data.result_data);
+				if (data.result_data) {
+					// Add the temporary file coloring and change the current file to the saved file.
+					document.querySelectorAll('.set-file-info').forEach((nfo) => nfo.classList.add('temporaryFile'));
+					for (const currentFile of document.querySelectorAll('.current-file')) {
+						currentFile.textContent = currentFile.dataset.tmpFile;
+					}
+				}
+			})
 			.catch((err) => showMessage(`Error saving temporary file: ${err?.message ?? err}`));
 	};
 
@@ -104,6 +113,12 @@
 				if (revertRadio && revertRadio.disabled) {
 					revertRadio.disabled = false;
 					revertRadio.checked = true;
+				}
+
+				// Add the temporary file coloring and change the current file to the saved file.
+				document.querySelectorAll('.set-file-info').forEach((nfo) => nfo.classList.add('temporaryFile'));
+				for (const currentFile of document.querySelectorAll('.current-file')) {
+					currentFile.textContent = currentFile.dataset.tmpFile;
 				}
 
 				if (editorForm) editorForm.target = 'WW_View';
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -124,10 +124,18 @@
 	if (messages.length) {
 		const dismissBtn = document.getElementById('dismiss-messages-btn');
 		dismissBtn?.classList.remove('d-none');
-		dismissBtn?.addEventListener('click', () => {
-			messages.forEach((message) => message.remove());
-			dismissBtn.classList.add('d-none');
-		});
+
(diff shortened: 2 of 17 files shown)


drgrice1:
Based on your image it looks like you need to run `npm ci` or `generate-assets.js`.

pstaabp:
I have the generate-assets.js running with the `-ws` flags so they update automatically.  I just restarted them and still see the same thing.

drgrice1:
`p-1` means a padding of 0.5rem on all four sides and is correct.  `py-1` means a padding of 0.5rem only on the top and bottom, but not on the left and right.  `pl-1` is not a valid bootstrap css class at all and won't doing anything.  `ps-1` sets the padding at the start (the left in left-to-right and right for right-to-left) to 0.5rem.  I changed to that.

drgrice1:
It would help for you to show me a screen shot in the context of the rest of the page.  Does the alert not span the entire page?  It should with this pull request.

pstaabp:
This looks good to be now. 

Here's the screenshot of the top of the PGeditor:

image

Before, the alert div wrapped the text and then the X overlapped the text so only went about 1/3 of the way across the page. 

I think I remember using `pl-1` (p-ell-1) for padding on the left, but bootstrap has changed to s for start.

drgrice1:
That is odd.  I expect the alert to be the full width of the page.  This is what I see.  This is in Firefox, but it it basically the same for me in Chrome.
alert

pstaabp:
I just added the `fade` and `show` classes to the alert and when clicking on them, they don't disappear as abruptly.

drgrice1:
Bootstrap version 4 used `pl-1`.  It had the left and right notation.  Bootstrap 5 switched to start and end, and supports left-to-right languages.

pstaabp:
I took the screen shot from Safari, but it looks the same to me now in Chrome.   I haven't reinstalled FF on this machine.  I could test it there too.

drgrice1:
I added the fade and show classes.  I also added some code to ensure that the dismiss button is removed when the last alert is dismissed.

drgrice1:
I tested on Chrome also, as I said.  I see the alert extending the width of the page there too.  Do you have a custom math4.js file in your theme?  Or is the math4.js file in your theme not a link to that file in the math4 directory?

drgrice1:
I tweaked things a bit more.  Now you should see it NOT extend the full width of the page.  The way it used to be.  I also removed the margin on the bottom of the `.message` divs.  The messages that are injected already have a margin, so when the `.message` divs are display inline-block (which I added back) the padding gets doubled up.

pstaabp (approved):

drgrice1:
The message margin issues are now fixed.  Instead of each message adding its own margin, the message `divs` are `display:inline-flex` with a gap.  The message `divs` itself has a margin below.  This results in a much cleaner and more reliable spacing above, below, and between messages.

Alex-Jordan (approved):
I tested all the things described in the opening post and they all work as described.

Alex-Jordan:
Two approvals here, so I'm going to merge this.

somiaj:
I like the new dismisible alerts, and now that I'm using them, I find I have to close the same alert twice at the top of the page and the bottom of the page (if I don't hit the Dismiss All button).  I was wondering if it would be useful to have the messages close in pairs, so if you dismiss the top alert, the bottom alert is also dismissed?

drgrice1:
I don't think that there is a good mechanism to achieve that.  For now the "Dismiss All" button will have to do.

PR #2093, opened 2023-06-27
Add folding and comment toggle to codemirror

pstaabp:
This adds two feature to the codemirror in the PGeditor

1. toggle comment.  To test, select a line or block of text and enter CTRL-/ (CMD-/ on macOS) to toggle between perl comment on and off
2. code folding. This was requested by @mgage.  Click on a line that start with BEGIN_XXX or Scaffold::Begin or Section::Begin and click CTRL-R.  

image

The block will collapse that looks like: 

image

I'm still working on getting the UI in the gutter to show arrows for the folding.


htdocs/js/PGCodeMirror/PG.js
```diff
@@ -1418,16 +1418,36 @@
...
 	});
-	
+
 	CodeMirror.registerHelper("wordChars", "perl", /[\w$]/);
-	
+
+	CodeMirror.registerHelper('fold','PG', function(cm,start) {
+		var line = start.line, lineText = cm.getLine(line);
+		var lastLineInFold = line;
+		var m1 = /^\s*BEGIN\_(PGML|PGML\_SOLUTION|PGML_HINT)\s*$/.exec(lineText);
```
drgrice1, line 1427: The underscores in the regular expression should not be escaped.

Also this should include BEGIN_TEXT.  We may not want authors to use it, but it still is supported.  Furthermore, if you are converting a problem from using that to using PGML, having support for it in the editor can be useful.

It would be nice to also fold on the `BEGIN_TIKZ` and `BEGIN_LATEX_IMAGE` constructs.  See lines 1343-1359 of `pg/lib/WeBWorK/PG/Translator.pm` for the regular expressions used by the translator.

htdocs/js/PGCodeMirror/pgeditor.js
```diff
@@ -75,11 +75,21 @@
...
+			"Ctrl-R": (cm) => cm.foldCode(cm.getCursor())
+		},
 		highlightSelectionMatches: { annotateScrollbar: true },
 		matchBrackets: true,
 		inputStyle: 'contenteditable',
 		spellcheck: localStorage.getItem('WW_PGEditor_spellcheck') === 'true',
+		gutters: ["CodeMirror-linenumbers", "CodeMirror-foldgutter"],
+		foldGutter: {
+			rangeFinder: new CodeMirror.fold.combine(CodeMirror.fold.PG)
+		},
```
drgrice1, line 91: This nicely condenses onto one line which is consistent with the coding here (and is nicer).  See line 84 above for example.

htdocs/js/PGCodeMirror/pgeditor.js
```diff
@@ -75,11 +75,21 @@
 		tabMode: 'spaces',
 		lineNumbers: true,
 		lineWrapping: true,
-		extraKeys: { Tab: (cm) => cm.execCommand('insertSoftTab') },
+		extraKeys: {
+			Tab: (cm) => cm.execCommand('insertSoftTab'),
+			'Ctrl-/': (cm) => cm.execCommand('toggleComment'),
+			'Cmd-/': (cm) => cm.execCommand('toggleComment'),
+			"Ctrl-R": (cm) => cm.foldCode(cm.getCursor())
```
drgrice1, line 82: This should use single quotes like the rest.  Also on line 92 below.

drgrice1 (commented):
Here are some initial observations.

`CTRL-R` is a bad choice for folding.  If you use the vim key map, `CTRL-R` is the vim key binding for redo.  I use undo and redo a lot when editing.

Please clean up the formatting of the added code in PG.js.  I know the file is a mess, but you can at least add clean new code.  Commas should always have a space after them.  There are almost no exceptions to this.  You can use modern javascript mixed with old es5.  So `var` shouldn't be use in the code you add.

pstaabp:
> CTRL-R is a bad choice for folding. If you use the vim key map, CTRL-R is the vim key binding for redo. I use undo and redo a lot when editing.  

Any thoughts one which key?  I changed to `Shift-Ctrl-F`, but if there is anything better, let me know. 

Just updated the code for cleanup as well as fixed the issue with the gutter showing arrows where the code can be folded.  Additionally, if the line with the cursor does not have a starting block line (BEGIN_PGML for example), then it will scan up to find a possible one.

drgrice1:
`Shift-Crtl-F` works.  I don't have a better suggestion.

It would be nice if that could be dependent on the keybinding in use.  In vim the usual keybinding for opening and closing a fold is `za`.

It would also be nice if you could close or open all folds at once.  In vim the keybinding are `zM` and `zR` to close all folds and open all folds respectively.  Of course vim has more with this.  There is also `zA` to open or close the fold the cursor is on recursively.

drgrice1 (approved):
Although I think there is more that can be done here.  This is still good.

htdocs/js/PGCodeMirror/PG.js
```diff
@@ -1418,16 +1418,34 @@
...
+						from: CodeMirror.Pos(start.line, cm.getLine(start.line).length),
+						to: CodeMirror.Pos(current_line, cm.getLine(current_line).length)
+					};
+				}
+			}
+		}
+		return;
+	});
+
+
```
drgrice1, line 1441: There is an extra empty line here.

pstaabp:
> `Shift-Crtl-F` works. I don't have a better suggestion.
> 
> It would be nice if that could be dependent on the keybinding in use. In vim the usual keybinding for opening and closing a fold is `za`.
> 
> It would also be nice if you could close or open all folds at once. In vim the keybinding are `zM` and `zR` to close all folds and open all folds respectively. Of course vim has more with this. There is also `zA` to open or close the fold the cursor is on recursively.

I can look into adding opening and closing all folds.  I'll see if the key mapping can be dependent on the mode.

htdocs/js/PGCodeMirror/PG.js
pstaabp: I miss the linter for js from webwork3.

drgrice1: Yeah.  Me too.

pstaabp:
Just added fold-all (Shift-Ctrl-A) and unfold-all (Shift-Ctrl-U) to the keybindings.  On first look, not sure how to make it work with the vim keybindings you mentioned above.

drgrice1:
`Ctrl-Shift-U` is a problem in Firefox and Chrome (at least on Linux).  That starts unicode character code entry.

drgrice1:
Also, don't there need to be `Cmd` variants of these for Mac?

drgrice1:
Finding key combinations that don't interfere with other things is always a challenge.  I had the same problem with the graph tool keyboard controls.

Don't worry about making vim specific bindings.  I think it can be done, but would take more effort.  I think that would have to be done when the keymap is set or changed later.

htdocs/js/PGCodeMirror/pgeditor.js
```diff
@@ -75,11 +75,21 @@
...
+			'Cmd-/':        (cm) => cm.execCommand('toggleComment'),
+			'Shift-Ctrl-F': (cm) => cm.foldCode(cm.getCursor(), { scanUp : true }),
+			'Shift-Ctrl-A': (cm) => CodeMirror.commands.foldAll(cm),
+			'Shift-Ctrl-U': (cm) => CodeMirror.commands.unfoldAll(cm),
+		},
 		highlightSelectionMatches: { annotateScrollbar: true },
 		matchBrackets: true,
 		inputStyle: 'contenteditable',
 		spellcheck: localStorage.getItem('WW_PGEditor_spellcheck') === 'true',
+		gutters: ["CodeMirror-linenumbers", "CodeMirror-foldgutter"],
```
drgrice1, line 90: There are still double quotes on this line.

htdocs/js/PGCodeMirror/PG.js
```diff
@@ -1418,16 +1418,33 @@
...
+				if(end_re.test(cm.getLine(current_line))) {
+					return {
+						from: CodeMirror.Pos(start.line, cm.getLine(start.line).length),
+						to: CodeMirror.Pos(current_line, cm.getLine(current_line).length)
+					};
+				}
+			}
+		}
+		return;
+	});
```
drgrice1, line 1439: I didn't look closely at this before.  The regular expressions need to be more precise.  If a line contains `# BEGIN_PGML` it should not start a fold.  Also, there is no need for the m1 and m2 variables.  Here is my suggested way to do this:
```suggestion
	CodeMirror.registerHelper('fold', 'PG', (cm, start) => {
		const match =
			/^\s*BEGIN_(PGML|PGML_SOLUTION|PGML_HINT|TEXT|TIKZ|LATEX_IMAGE)\s*$/.exec(cm.getLine(start.line)) ||
			/^\s*(Section|Scaffold)::Begin/.exec(cm.getLine(start.line)) ||
			/^[$\w]*->\s*BEGIN_(TIKZ|LATEX_IMAGE)\s*$/.exec(cm.getLine(start.line));
		if (match) {
			for (let current_line = start.line + 1; current_line <= cm.lineCount(); ++current_line) {
				if (RegExp(`END_${match[1]}`).test(cm.getLine(current_line))) {
					return {
						from: CodeMirror.Pos(start.line, cm.getLine(start.line).length),
						to: CodeMirror.Pos(current_line, cm.getLine(current_line).length)
					};
				}
			}
		}
		return;
	});
```

drgrice1: Sorry.  You do need the `m1` and `m2`.  I missed the difference between the scaffolds and the others.  So make that:
```suggestion
	CodeMirror.registerHelper('fold', 'PG', (cm, start) => {
		const m1 =
			/^\s*BEGIN_(PGML|PGML_SOLUTION|PGML_HINT|TEXT|TIKZ|LATEX_IMAGE)\s*$/.exec(cm.getLine(start.line)) ||
			/^[$\w]*->\s*BEGIN_(TIKZ|LATEX_IMAGE)\s*$/.exec(cm.getLine(start.line));
		const m2 = /^\s*(Section|Scaffold)::Begin/.exec(cm.getLine(start.line));
		if (m1 || m2) {
			for (let current_line = start.line + 1; current_line <= cm.lineCount(); ++current_line) {
				const end_re = m1 ? RegExp(`END_${m1[1]}`) : RegExp(`${m2[1]}::End`);
				if(end_re.test(cm.getLine(current_line))) {
					return {
						from: CodeMirror.Pos(start.line, cm.getLine(start.line).length),
						to: CodeMirror.Pos(current_line, cm.getLine(current_line).length)
					};
				}
			}
		}
		return;
	});

Comment (drgrice1) [reply htdocs/js/PGCodeMirror/PG.js:1439]: The regular expressions in my suggestion more closely match the perl regular expressions in the translator.

**Comment by Alex-Jordan** [comment]
I'm getting some undesirable behavior with the comment toggling. I accept that these may be things to just live with.

* If a line opens with multiple `#` (possibly with spaces mixed in), then `ctrl /` removes one `#`. Then it removes the next one, and so on. Eventually you are toggling between 1 and 0, but it's not the same as what you started with. Not sure you could do anything about this though unless you put in memory what the original line looked like. But it might be better just to do nothing if the line already starts with more than one `#`.
* If I am inside a BEGIN_PGML block and use `ctrl /`, it adds a `#`. Then it adds another, and so on. It should either do nothing, or wrap the line in `[%   %]`, the PGML comment delimiter. Then if the cursor is within a `[%   %]`, and you use `ctrl /`, it could remove them.
* Same with BEGIN_PGML_HINT, BEGIN_PGML_SOLUTION
* Same with a BEGIN_TEXT block, except I don't know that there is a way to comment in there. So do nothing?
* Same with BEGIN_HINT, BEGIN_SOLUTION
* If it's not hard to do, highlighting some selection (possibly multiline) within a PGML block and using `ctrl /` could wrap that particular selection in `[%   %]`.

**Comment by pstaabp** [comment]
About the multi-character comments.  I assume you're thinking of the ## for meta data, I'll look into that.  I haven't seen a lot of example of custom comment toggling. 

About the ability to perform different comments in different parts of a problem, it appears that the PG mode that we created for codemirror isn't set up in an efficient way.  I've seen other examples where there is a base language (in our case perl) and then inside blocks, there are other modes defined.  For us, we would have a PGML mode for PGML, PGML_SOLUTION and PGML_HINT and then we could use TeX for the LATEX_IMAGE and TIKZ blocks. It seems like to use this, we'd need to rewrite the PG mode that we have.  I think this would be a good idea, but perhaps a bit too much for this PR.  Also, although it appears that the modes work for CodeMirror 6, which we may switch to in the future, I hesitate to put too much work into CM 5 if the modes need to change.

**Comment by drgrice1** [comment]
@pstaabp:  `PG.js` is defined so that inside blocks there are other modes defined.  You aren't looking closely at the code.

**Comment by drgrice1** [comment]
Also, using TeX mode inside LATEX_IMAGE and TIKZ blocks is not as good of an idea as you might think.  The problem is that what is inside those blocks is not valid TeX usually.  Usually that includes perl variables because the whole point of these blocks is for dynamically generated images that depend on random parameters.  TeX balks at that.  I tried that with a syntax highlighter for vim, and it did not work so well.

**Comment by pstaabp** [comment]
What I meant is that I think we encapsulate everything a bit better and create a PG mode for BEGIN_TEXT blocks and a PGML mode for BEGIN_PGML blocks.   The classic example shown is an html mixed mode which uses the defined css and javascript modes when appropriate.  

In terms of this PR, the `toggleComment` function uses either a line or block comment and the mode need to define what these are.  In PGML mode, we would need to define `beginBlockComment` to be `[%` and `endBlockComment` to be `%]` for example.  These aren't done in `PG.js`.

File: htdocs/js/PGCodeMirror/PG.js (hunk 6 above)
Comment (pstaabp) [reply htdocs/js/PGCodeMirror/PG.js]: I removed the `|TIKZ|LATEX_IMAGE` from the regex on line 1443 since it is on 1444.

**Comment by pstaabp** [comment]
> Also, using TeX mode inside LATEX_IMAGE and TIKZ blocks is not as good of an idea as you might think. 

We could do a simplified tex mode that allows perl variables.

File: htdocs/js/PGCodeMirror/PG.js (hunk 6 above)
Comment (drgrice1) [reply htdocs/js/PGCodeMirror/PG.js]: Yes, it should have been removed.  It was in my initial edits, but then it was undone later in successive changes.

**Comment by Alex-Jordan** [comment]
> About the multi-character comments. I assume you're thinking of the ## for meta data,

It's also things like

```
if condition
    line 1 of code
    line2 of code
    line3 of code
```

and you comment out line 2:
```
if condition
    line 1 of code
    #line2 of code
    line3 of code
```

and later you comment out the whole block:
```
#if condition
#    line 1 of code
#    #line2 of code
#    line3 of code
```

If the cursor is on "line 2" and you use `ctrl /`, it will remove thee first `#`. Then do it again and it will remove the other.

I dont' have a great suggestion. Off the top of my head, I would do nothing for a line that matches `^\s*##+` like the metadata lines at the top of a file. But with this other situation I don't know.

Well, I just tried SublimeText comment toggle on a perl file, and it behaves like this is behaving with multiple `#` (and possible whitespace) opening a line. If this behavior is standard, I should be complaining.

**Comment by drgrice1** [comment]
I am not concerned with the double commenting that occurs in the general perl portion of the code as @Alex-Jordan just described.  @pstaabp is just using the built in addon that toggles the comments, and that behavior is standard for text editors that do this sort of thing.  I am more concerned with commenting that occurs inside a PGML block.  The built in addon is clearly not working right there.

**Comment by pstaabp** [comment]
Have updated to customize the `comment.js` code from codemirror.  This detects the mode in a different way and provides either a line or block comment. 

- double `##` is working for the tagging section.
- regular perl comments are working.
- PGML comments have a block structure.  On a single line, it wraps the line in a comment, if a range is selected, it comments the range. 
- comments in BEGIN_TIKZ and BEGIN_LATEX_IMAGE are not toggling.  To comment a line, this works, but uncommenting, provides a second `%`.  We don't provide extra structure for these in `PG.js` and would probably need to create these modes. 

Lastly, to determine the mode of a problem does some scanning of the file and is not very efficient, but I think with the size of typical pg problems, I don't notice any performance problems--I can try to do a single sweep through the file if it seems necessary.

**Comment by pstaabp** [comment]
> `Ctrl-Shift-U` is a problem in Firefox and Chrome (at least on Linux). That starts unicode character code entry.

I still need to update this.  `Ctrl-Shift-G` ? 

I can also add the Mac versions, and I will update the help documentation to mention these new features.

File: htdocs/js/PGCodeMirror/comment.js [hunk 7]
```diff
@@ -0,0 +1,265 @@
...
+				var atCursor = cmp(self.getCursor("to"), to) == 0, empty = !self.somethingSelected()
+				self.replaceRange(endString, to);
+				if (atCursor) self.setSelection(empty ? to : self.getCursor("from"), to)
+				self.replaceRange(startString, from);
+			}
+		});
+	});
+
+	CodeMirror.defineExtension("uncomment", function(from, to, options) {
+		console.log('uncommenting');
```
drgrice1, line 186: Please remove all of the `console.log` calls added in this method (this is the first one, and the last one is on line 208).

drgrice1:
`Ctrl-Shift-G` seems to work fine.

htdocs/js/PGCodeMirror/comment.js
```diff
@@ -0,0 +1,265 @@
...
+		}
+	});
+
+	// Rough heuristic to try and detect lines that are part of multi-line string
+	function probablyInsideString(cm, pos, line) {
+		return /\bstring\b/.test(cm.getTokenTypeAt(Pos(pos.line, 0))) && !/^[\'\"\`]/.test(line)
+	}
+
+	const md_section = RegExp('DESCRIPTION|KEYWORDS|DBsubject|DBchapter|DBsection|Date|Author|Institution ' +
+		'|MO|Static|TitleText|EditionText|AuthorText|Section|Problem|Language|Level';
```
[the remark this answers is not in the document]
pstaabp: Fixed.

drgrice1 (approved):
I think this looks acceptable now.

I see a lot of issues with the syntax highlighting in general.  These are not caused by this pull request, but are problems with PG.js that I have seen for a while.  There are a lot of cases in PGML blocks that syntax highlighting just messes up.  Some of those cause block commenting to not work right in PGML blocks as well.  For example consider the following
```
BEGIN_PGML
Give the solution set.

The solution set is [`\big\{`] [_]{$sol_cmp}{10} [`\big\}`].

Stuff after.
END_PGML
```
In that block everything after the "The solution set is ..." line appears as if it is commented out.  This occurs with or without this pull request.  However, you can't toggle comment blocks anywhere after the line with this pull request.

PR #2094, opened 2023-06-27
remove unused latex packages

Alex-Jordan:
These packages can go (and one of them is causing a problem). These packages were used back with LaTeX2.09

`epsf` only provides `\epsfbox` which I find nowhere in `webwork2`, `pg`, or `OpenProblemLibrary`.

`epsfig` only provides `\psfig` which I find nowhere in `webwork2`, `pg`, or `OpenProblemLibrary`.

`pslatex` is for using PostScript fonts. We are fine with default fonts, or if not, would use something newer.

### DIFF
--- a/conf/snippets/hardcopyThemes/common/packages.tex
+++ b/conf/snippets/hardcopyThemes/common/packages.tex
@@ -15,9 +15,6 @@
   \DeclareUnicodeCharacter{20AC}{\euro} % make it possible to use the UTF-8 character for the euro symbol in problems
 \fi
 
-\usepackage{epsf}
-\usepackage{epsfig}
-\usepackage{pslatex}
 \usepackage{listings}
 \lstset{basicstyle=\ttfamily,breaklines=true,aboveskip=0pt,belowskip=0pt}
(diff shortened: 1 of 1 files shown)


pstaabp (approved):

drgrice1:
I have never been particularly fond of the default LaTeX fonts.  So I would prefer to use something else.  The STIX 2 fonts are nice.  They are available in the LaTeX stix2 package.

Alex-Jordan:
Once some of the current PRs are merged, I'll be working on the hardcopy themes a bit more. I'll put in STIX 2 fonts with that.

drgrice1:
Actually, I was thinking about it more.  Although I do much prefer the STIX 2 fonts, they don't match with the MathJax fonts.  The default LaTeX fonts match better with those.  So in the interest of making hardcopy resemble what is seen in html, it would probably be best not to switch to the STIX 2 fonts.   That is, unless MathJax can also be switched to STIX 2 fonts?

drgrice1 (approved):
I approve this pull request in any case.

PR #2095, opened 2023-06-27
change spacing between problems

Alex-Jordan:
Using `\vfill` between problems in hardcopy causes some bad things. One is just the irregular spacing between problems from one page to the next (or one column to the next). But also this is leading to some unnecessary page (or column) breaks in some hardcopies I was looking at recently. Maybe it was not a problem in 2.17, but is a problem with the new hardcopy themes. I think 1pc of space is adequate, but could also change that to something else.


drgrice1:
I don't think that the `\vfill` messes up things and causes bad spacing.  I think it causes page breaks to occur at better places.  With this pull request I see a lot more of problems being split across pages or columns.  I don't like that.  I would rather have the spacing and problems on a single page or column.

Alex-Jordan:
I will wait until a few hardcopy related prs are merged and see if the
issues that I see are still there, and then show them. I think they are new
with using the exam class with longer exercises. There is something to work
on and at first it may be a game of whack a mole. But will converge to
something good.

Alex-Jordan:
I made a new local branch with everything merged so I could look at this and be sure that things like removing `pslatex` won't affect my issue. Suppose a problem ends, and the next problem is a very long one that would extend into a new page. The way things are now, that problem gets pushed until it can _start_ on a new page. So the PDF can be like this extreme case:
Screen Shot 2023-06-27 at 10 30 58 PM

Very long problems are naturally part of an orientation/demo assignment, which is why I've been seeing this more than on might typically se this happen as I work on a new orientation.

I've tried removing the `\goodbreak` to see if that is really the culprit here, but that did not change things in the screenshot. Of course changing the `\vfill` as I did with this commit does change things.

Do you have any TeX insight as to why it is breaking this way? I have not ruled out that it has to do with the exam class and each problem being inside a `\titledquestion` within one big `questions` environment.

Alex-Jordan:
Instead of having hard coded `exam` class stuff in the hardcopy modules, maybe they should just print things into macros/counters (like the exercise number and the point value) and all the exam class commands move to the theme files. And then it becomes possible to opt for a them that does not use `exam`.

pstaabp:
@Alex-Jordan All the other hardcopy PRs are now merged.  rebase this onto WeBWorK-2.18 and let's see what this looks like.

Alex-Jordan:
I still get thee kind of issue that I see in the screenshot here. Again, I haven't done a careful test, but it seems likely that the `exam` class `\titledquestion` is chunking a problem in such a way to give a strong preference that a question not be split over pages/columns. And that leads to the kind of thing in my screenshot if there is a `\vfill` available to let the second question slide that far down.

I have a branch where `exam` specific stuff has been removed from `Hardcopy.pm` and put into theme files. Maybe it is my penchant for XML a la PTX, but I think the `hardcopyProblemDivider.tex` is a bad idea, and I replaced it with `hardcopyProblemHeader.tex` and `hardcopyProblemFooter.tex`. The thinking is that problems are objects, and the space in between them is not. And then `Hardcopy.pm` writes things like the problem ID and weight where they are needed, like it does with things like `\webworkCourseName`.

Anyway with this branch I could design an `exam` class based theme, a bare bones theme, and maybe even something else like a `tcolorbox` based theme (maybe). I just don't want to go far with it if there is a plan to tie things more closely to the `exam` class. Although if that is the plan, then we need to figure out this terrible page break issue like in my screnshot.

Alex-Jordan:
I couldn't make this week's meeting or I would have brought this up for discussion. Do I have a green light to work on separating the `exam` class from `lib` code? I have a branch that does this, I just need to clean it up. A theme will have `hardocopyProblemHeader.tex` and `hardocopyProblemFooter.tex`and no longer have `hardocopyProblemDivider.tex`. Before `hardocopyProblemHeader.tex` is written to `hardcopy.tex`, the ID and weight of the upcoming problem is written into a tex macro, that `hardocopyProblemHeader.tex` can reference. In this way the `exam` class's `\titledquestion` macro can still be used.

The scheme will have more flexibility for hardcopy themes, at the expense of one more file among hardcopy theme files. And it seems right, because right now we have a theme preamble file that uses `\documentclass{exam}` but then we have `exam`-specific code in `lib`.

And there is still the bad pagebreaking issue to address. (But I might address it by making a theme that does not use `exam`.)

drgrice1:
Sure.  Put in a pull request so we can check it out when you are ready.

PR #2096, opened 2023-06-29
PTX XML

Alex-Jordan:
I think these are typos: `webwork_rpc` instead of `render_rpc` in the mojolicious yml file. And a regex issue that happened to not matter until I started trying to use `render_rpc`. I am not sure of either though. Are these changes good?


drgrice1 (approved):
Yes, the links referenced in `/conf/webwork2.mojolicious.dist.yml` should be `/webwork2/render_rpc`.  Good catch.

The regular expression was not precise in `lib/WeBWorK.pm` either.

Good catch.

Alex-Jordan:
OK, so my next question is how do I actually use `render_rpc`? I swapped
out a working `html2xml` URL with `render_rpc` and I get "Authentication
failed. Log in again to continue."

On Wed, Jun 28, 2023 at 6:20 PM Glenn Rice ***@***.***> wrote:

> ***@***.**** approved this pull request.
>
> Yes, the links referenced in /conf/webwork2.mojolicious.dist.yml should
> be /webwork2/render_rpc. Good catch.
>
> The regular expression was not precise in lib/WeBWorK.pm either.
>
> Good catch.
>
> —
> <https://protect2.fireeye.com/v1/url?k=31323334-501d2dca-3132feb7-454455534531-571f609e8fc73627&q=1&e=3a774be1-f64b-47ad-bd90-1d289f1ac440&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fwebwork2%2Fpull%2F2096%23pullrequestreview-1504429868>,
> or unsubscribe
> <https://protect2.fireeye.com/v1/url?k=31323334-501d2dca-3132feb7-454455534531-a639369aeda15fc7&q=1&e=3a774be1-f64b-47ad-bd90-1d289f1ac440&u=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABEDOAG22YYTYPHSFQSIOGDXNTJ5ZANCNFSM6AAAAAAZX2W334>
> .
> ***@***.***>
>


-- 
Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
In the url parameters that you pass, use `user` instead of `userID`, `courseID` instead of `courseName`, `passwd` instead of `course_password`, and `key` instead of `session_key`.  The new parameters are just the usual parameters that have always been used by the rest of webwork for these things.  The point is that the `/webwork2/render_rpc` route goes through the usual authentication path that all other routes that require authentication go through.  It no longer needs different parameters as it doesn't go through the XML middle man anymore.  Any other parameters that you use should be the same.

Alex-Jordan:
With html2xml, it works with keys:
courseID
userID
course_password
That's what PTX has been using with 2.17. It's not using courseName.

I'm getting it to work now with render_rpc using keys
courseID
user
passwd

Was courseName recognized before? If not, does line 98 of WeBWorK.pm need a
small edit?






On Wed, Jun 28, 2023 at 6:42 PM Glenn Rice ***@***.***> wrote:

> In the url parameters that you pass, use user instead of userID, courseID
> instead of courseName, passwd instead of course_password, and key instead
> of session_key. The new parameters are just the usual parameters that
> have always been used by the rest of webwork for these things. The point is
> that the /webwork2/render_rpc route goes through the usual authentication
> path that all other routes that require authentication go through. It no
> longer needs different parameters as it doesn't go through the XML middle
> man anymore. Any other parameters that you use should be the same.
>
> —
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-be69438b87770d10&q=1&e=96c89264-ef58-4d73-85fb-78c5ddf8e355&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fwebwork2%2Fpull%2F2096%23issuecomment-1612311981>,
> or unsubscribe
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-ee8fa5c4525ecd5b&q=1&e=96c89264-ef58-4d73-85fb-78c5ddf8e355&u=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABEDOAH2TJFQ57ZKC4CAPJTXNTMPPANCNFSM6AAAAAAZX2W334>
> .
> ***@***.***>
>


-- 
Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
There should be no need for an edit.  The mappings that are there are what were in the code before.  I am certain of that.  Note that the xml2html endpoint will also work with the new parameters directly.

drgrice1:
I see that `courseID` was also used by the previous `html2xml` endpoint.  `courseName` was not actually a url parameter, but was passed internally to the XMLRPC call.  So we could drop that from the map.

drgrice1:
Note that leaving it in the mapping effectively adds it as a new parameter that will work with the `html2xml` endpoint as well as `courseID`.

Alex-Jordan:
Thanks for all the explanation. I cleaned up the `ptx` format some, and don't plan to do anything more than what is here now. Unless something comes up.

Alex-Jordan:
OK, I added another thing. I added an HTTP header to make the `render_rpc` requests for static PTX identify as XML, which makes web browsers display it as XML (a collapsible tree and recognition of invalid XML are the features that will be helpful).

At first I thought to make it configurable, but (a) it would be hard to do that since this depends on the query parameters of the URL and (b) I'm not sure it should be configurable anyway. So I hard coded it in. Not sure if that is best or if I did this the right way. I just mimicked what was in there for configurable headers.

lib/Mojolicious/WeBWorK.pm
```diff
@@ -122,6 +122,18 @@ sub startup ($app) {
...
+	$app->hook(
+		before_dispatch => sub ($c) {
+			if ($c->req->url->path =~ /^\/webwork2\/render_rpc/
+				&& $c->req->url->query->param('outputformat') eq 'ptx'
+				&& $c->req->url->query->param('displayMode') eq 'PTX')
+			{
+				$c->res->headers->header('Content-Type' => 'text/xml; charset=utf-8');
+			}
+		}
+	);
```
drgrice1, line 135: This should be done in the RenderViaRPC.pm `content` method, and not here.  There you don't need to check the `$c->req->url->path` because that is guaranteed to be the path (and that is also not the currect way here -- you should have used $c->current_route).  A `before_dispatch` hook should be something that applies more generically to potentially any route, and not just a single route.

drgrice1:
@Alex-Jordan:  I added a pull request to this branch that sets the "text/xml" header at the correct place.  That actually needs to be in the `formatRenderedProblem` method of the `FormatRenderedProblem` package and not in the `RenderViaRPC` package as I suggested earlier, because it needs to be done before the response is rendered.

drgrice1:
@Alex-Jordan:  In my pull request that I merged, I restricted the addition of the "text/xml" header to the render_rpc route as you had done before.  Do you not want that for the html2xml route?  If you just remove the `current_route` check on line 191 of FormatRenderedProblem.pm, then the html2xml route would also have that header set.

Are you also interested in the raw format actually serving JSON?  For this to happen you would change line 234 of FormatRenderedProblem.pm from `return $ws->c->render(data => JSON->new->utf8(1)->encode($output));` to `return $ws->c->render(json => $output);`.  Note that Mojolicious takes care of encoding `$output` to JSON with UTF-8 encoding, and setting the header to `application/json;charset=UTF-8`.  I think that since you use jQuery's `getJSON` method it wouldn't change the way that you deal with the response in the pretext-webwork.js file.

Alex-Jordan:
> Do you not want that for the html2xml route?

I was thinking to just leave that behaving like it has up to now because it's in use, and I don't want to disrupt some workflow people have. Maybe it would make things better, I'm just not sure. Should I change my mind? I'm not of a strong opinion about it.

> Are you also interested in the raw format actually serving JSON?

That could be helpful (although changing it runs contrary to my thinking to not change html2xml). I am catching up with PTX production workflow soon, to verify things will work well with 2.18. I will experiment with this when I get to that point.

drgrice1:
> > Are you also interested in the raw format actually serving JSON?
> 
> That could be helpful (although changing it runs contrary to my thinking to not change html2xml). I am catching up with PTX production workflow soon, to verify things will work well with 2.18. I will experiment with this when I get to that point.

You could only do it for the render_rpc endpoint.

pstaabp (approved):

pstaabp (approved):

PR #2098, opened 2023-06-30
Add the capability in the pg problem editor to pg perltidy code.

drgrice1:
This adds a new tab to the PG problem editor (unless editing a course configuration file).  The tab contains basic information about what it does, but if the "Tidy Code" button is pressed it perltidies the code in the CodeMirror editor window.  The tidied code is also saved to the temprorary file.

The actual tidying is done by a new webwork webservice action (not a form submission).  If errors occur, they are shown in the renderer window.

The webwork webwservice depends on a new PG module defined in a corresponding PG pull request (see https://github.com/openwebwork/pg/pull/868).

### DIFF
--- a/lib/WeBWorK/CourseEnvironment.pm
+++ b/lib/WeBWorK/CourseEnvironment.pm
@@ -101,6 +101,11 @@ sub new {
 	my %ORIG_SIG;
 	$ORIG_SIG{$_} = $SIG{$_} for keys %SIG;
 
+	# The following line is a work around for a bug that occurs on some systems.  See
+	# https://rt.cpan.org/Public/Bug/Display.html?id=77916 and
+	# https://github.com/openwebwork/webwork2/pull/2098#issuecomment-1619812699.
+	%+;
+
 	my $safe = Safe->new;
 	$safe->permit('rand');
 	# seed course environment with initial values
--- a/lib/WebworkWebservice/ProblemActions.pm
+++ b/lib/WebworkWebservice/ProblemActions.pm
@@ -22,6 +22,7 @@ use warnings;
 use Data::Structure::Util qw(unbless);
 
 use WeBWorK::Debug;
+use WeBWorK::PG::Tidy qw(pgtidy);
 
 sub getUserProblem {
 	my ($invocant, $self, $params) = @_;
@@ -130,6 +131,25 @@ sub putPastAnswer {
 			"Updated answer $params->{answer_id} for problem $pastAnswer->{problem_id} of $pastAnswer->{set_id} "
 			. "for user $pastAnswer->{user_id} in course "
 			. $self->ce->{courseName} . '.'
+	};
+}
+
+sub tidyPGCode {
+	my ($invocant, $self, $params) = @_;
+
+	debug('in tidyPGCode');
+
+	local @ARGV = ();
+
+	my $code = $params->{pgCode};
+	my $tidiedPGCode;
+	my $errors;
+
+	my $result = pgtidy(source => \$code, destination => \$tidiedPGCode, errorfile => \$errors);
+
+	return {
+		ra_out => { tidiedPGCode => $tidiedPGCode, status => $result, errors => $errors },
+		text   => 'Tidied code'
 	};
 }
 
--- a/lib/WebworkWebservice/ProblemActions.pm
+++ b/lib/WebworkWebservice/ProblemActions.pm
@@ -21,13 +21,10 @@ use warnings;
 
 use Data::Structure::Util qw(unbless);
 
-use WeBWorK::Debug;
 use WeBWorK::PG::Tidy qw(pgtidy);
 
 sub getUserProblem {
 	my ($invocant, $self, $params) = @_;
-
-	debug('in getUserProblem');
 
 	my $db = $self->db;
 
@@ -43,8 +40,6 @@ sub getUserProblem {
 
 sub putUserProblem {
 	my ($invocant, $self, $params) = @_;
-
-	debug('in putUserProblem');
 
 	my $db = $self->db;
 
@@ -75,8 +70,6 @@ sub putUserProblem {
 
 sub putProblemVersion {
 	my ($invocant, $self, $params) = @_;
-
-	debug('in putProblemVersion');
 
 	my $db = $self->db;
 
@@ -109,8 +102,6 @@ sub putProblemVersion {
 sub putPastAnswer {
 	my ($invocant, $self, $params) = @_;
 
-	debug('in putPastAnswer');
-
 	my $db = $self->db;
 
 	my $pastAnswer = $db->getPastAnswer($params->{answer_id});
@@ -136,8 +127,6 @@ sub putPastAnswer {
 
 sub tidyPGCode {
 	my ($invocant, $self, $params) = @_;
-
-	debug('in tidyPGCode');
 
 	local @ARGV = ();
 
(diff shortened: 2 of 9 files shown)


pstaabp:
Haven't tried this out yet, but was thinking that we should add some documentation to the PGEditor help page for this.

drgrice1:
Yeah. I will add that.

drgrice1:
I added a section about the new tab to the help file.  Although, I didn't have anything else to say other than what the tab already says.

drgrice1:
I decided that that lengthy explanation should only be in the help, and not directly on the tab.  So the tab now simply say "Reformat the code using perltidy."

Alex-Jordan:
I am at commit 651fa85 on `pg`, and 597103b on `webwork2`, so I think I have both branches where they should be. I have run `npm ci` in both places, and restarted `webwork2`. I restored `localOveerrides.conf` following my changes for the hardcopy refactor branch.

But I get this error anywhere on the problem editor page.
Screen Shot 2023-07-01 at 4 45 30 PM

drgrice1:
I added Perl::Tidy to `check_modules.pl` .  It needs to be version 20220613 or newer.  Note that newer versions cause reformatting of webwork2 and pg code from that version, but should be fine for what is needed for the tidying of problems.

drgrice1:
@Alex-Jordan:  I suspect that not having Perl::Tidy installed is your problem.

Alex-Jordan:
That wasn't it. II have `Perl::Tidy` and it is `$VERSION = '20220613';`. Still getting that error.

The webwork2 error long has:

```
[2023-07-01 20:43:35.12892] [164601] [error] [eNZmnALDPO1S] Can't locate loadable object for module main::Tie::Hash::NamedCapture in @INC (@INC contains: /opt/webwork/pg/lib /opt/webwork/webwork2/lib /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5) at /usr/share/perl5/English.pm line 155.
```

Not sure what to make of that. Install `Tie::Hash::NamedCapture`?

Alex-Jordan:
I had `Perl::Tidy` installed already somehow. I ran `sudo cpanm Perl::Tidy` to see what it would do, and it installed version 20230701. But nothing has changed. I still get that error in the webwork2 log file.

Alex-Jordan:
OK, I see quite a bit about this after Googling the error message, for instance here: https://rt.cpan.org/Public/Bug/Display.html?id=77916

drgrice1:
Both `Safe` and `WWSafe` use `Tie::Hash::NamedCapture`.  `Safe` is in `check_modules.pl`, although both `Safe` and `Tie::Hash::NamedCapture` are core perl modules.  If you have perl installed on your system, you should have both of those.

This issue happens for you with this branch, but not with the WeBWorK-2.18 branch?  Also, this doesn't happen when you view a problem in a set in the usual way?

Alex-Jordan:
I have them installed and don't have this issue except with this branch.
Did you see my later post that points to a known perl bug?

drgrice1:
> I have them installed and don't have this issue except with this branch. Did you see my later post that points to a known perl bug?

I saw that, but I don't see how that has anything to do with this pull request.  This pull request doesn't change anything with the `Safe` or `WWSafe` usage, or problem rendering in general.  That is why I asked if the issues you are seeing are restricted to this pull request, or also occur for the WeBWorK-2.18 branch.

drgrice1:
I also can not reproduce the issue on any of my systems.

Alex-Jordan:
It may intersect with this flavor of OS (Oracle 8). I will try some things
based on what people have posted. Seems to be about English.pm that is a
dependency.

Alex-Jordan:
I followed the suggestion from that link, putting `*+;` right before calling `Safe->new;` in `lib/WeBWorK/CourseEnvironment.pm`, and that resolves the bug for me. I confess I have no idea what `*+;` actually does.

Now that I can test the feature, it works!

drgrice1:
I added a small tweak this.  If there are no changes to the code when perltidy runs, then the code in the editor window is not updated, and the toast just says "There were no changes to the code."

@Alex-Jordan: I am not sure what to do about the `Safe` issue.  Should I add `*+;` before calling `Safe->new` in `CourseEnvironment.pm`?

Alex-Jordan:
> Should I add *+; before calling Safe->new in CourseEnvironment.pm?

I would say yes, but I would like to know what that command even does first. If we know what that does, we can at least think about if it could have potential bad side effects. (Knowing how it "fixes" the issue I encountered would be nice, but not as important imho.)

I am not a perl monk and I wouldn't have even guessed that was a legal command. It's been a challenge to Google about it given the nature of that character string.

Alex-Jordan:
OK, I sort of understand now. Maybe 85%.

`Safe` wants to use the special hash `%+` for which you can read its meaning in `perldoc perlvar`. This special hash is implemented by `Tie::Hash::NamedCapture`, which is loaded dynamically when it's needed. Apparently on my system, something about this branch leads to `Tie::Hash::NamedCapture` not being loaded fast enough. So by accessing `%+` before `Safe` is loaded, we make sure `Tie::Hash::NamedCapture` is loaded in time. And `*+` seems like overkill, accessing `%+` but also other things like `$+`. (I do not understand the star sigil...)

Anyway I tested that it is enough on my system to access `%+` right before `Safe`. So I guess my suggestion is that you add `%+;` right before `Safe` is loaded. And include some comment, maybe referencing the [bug report](https://rt.cpan.org/Public/Bug/Display.html?id=77916) or this GitHub thread.

drgrice1:
@Alex-Jordan I added `%+;` before `Safe` is used in `CourseEnvironment.pm`.  Thanks for figuring out what is needed there.

Alex-Jordan (approved):
Just did a fresh test and it's working well.

lib/WebworkWebservice/ProblemActions.pm
```diff
@@ -133,4 +134,23 @@ sub putPastAnswer {
 	};
 }
 
+sub tidyPGCode {
+	my ($invocant, $self, $params) = @_;
+
+	debug('in tidyPGCode');
```
pstaabp, line 140: Remove this line?

pstaabp (approved):
Looks great.

lib/WebworkWebservice/ProblemActions.pm
drgrice1: This has been the standard in the WebworkWebservice action, particularly in this file.  Each method calls `debug('in thismethod');`.  So I just followed suit.  Most of the other WebworkWebservice modules also do this, although not as consistently as this file.  I would certainly be fine with removing all of those.  That was probably more meaningful with the old XML pipeline that this was inside of.  Now it is probably just clutter.

lib/WeBWorK/CourseEnvironment.pm
```diff
@@ -101,6 +101,11 @@ sub new {
 	my %ORIG_SIG;
 	$ORIG_SIG{$_} = $SIG{$_} for keys %SIG;
 
+	# The following line is a work around for a bug that occurs on some systems.  See
+	# https://rt.cpan.org/Public/Bug/Display.html?id=77916 and
+	# https://github.com/openwebwork/webwork2/pull/2098#issuecomment-1619812699.
+	%+;
```
drgrice1, line 107: @Alex-Jordan:  This is not right.  I didn't notice this when I added this before.  This is causing a warning every time the webwork2 app starts about the useless use of a variable in void context. This needs further investigation.

PR #2099, opened 2023-07-01
refactor hardcopy themes

Alex-Jordan:
This pairs with https://github.com/openwebwork/pg/pull/870.

This refactors how hardcopy themes work. Things to note:
* This separates what tex packages and macros are to support PG, to support WeBWorK2 hardcopy production, and are to support a given theme.
* The hardcopy production is now abstracted away from the `exam` class.
* A hardcopy theme is now a single XML file. See examples in `assets/hardcopyThemes`.
* It should be easy now to make a hardcopy theme. Follow the `README` in `assets/hardcopyThemes`, or just copy one of the existing theme files.
* In light of this, IMHO it is not worth keeping the ability to granularly override parts of a theme that could previously be done with files in `snippets`. Those are gone. All of `snippets` is gone. The `pg` files that were there are now in `assets/pg`.
* This uses `XML::LibXML`. I had trouble installing it with cpanm, but no trouble using `yum` (`sudo yum install perl-XML-LibXML`).
* Several things in `localOverrides.conf` need updating with this PR.

I expect there will be things to tweak about the themes. But I left the `\vfill` in place. I think th issue I am seeing there with bad page breaks on long problems is not new to 2.18. But this PR will make it easy to adjust an existing theme or add a new one.


Alex-Jordan:
Here are three themes based on the `tclorbox` package. None of them are universally good for all hardcopy sets with long problems. But these are good for testing what can be done with themes. To try these, put them in `/assets/hardcopyThemes/`. And add the following to `localOverrides.conf`. Then they will be available for hardcopy production.

```
$hardcopyThemes = [
        {file => 'empty.xml',                     pretty => 'Empty'},
        {file => 'basic.xml',                     pretty => 'Basic'},
        {file => 'oneColumn.xml',                 pretty => 'One Column'},
        {file => 'twoColumn.xml',                 pretty => 'Two Columns'},
        {file => 'tcbraster.xml',                 pretty => 'TCBRaster'},
        {file => 'tcolorbox.xml',                 pretty => 'TColorBox'},
        {file => 'tcolorboxmulticol.xml',         pretty => 'TColorBoxMulticol'},
        #{file => 'XeLaTeX-oneColumn.xml',         pretty => "English\x{2014}one Column"},
        #{file => 'XeLaTeX-twoColumn.xml',         pretty => "English\x{2014}two Columns"},
        #{file => 'XeLaTeX-Hebrew-oneColumn.xml',  pretty => "Hebrew/English\x{2014}one Column"},
        #{file => 'XeLaTeX-Hebrew-twoColumn.xml',  pretty => "Hebrew/English\x{2014}two Columns"},
];
```

You will need to change the extension of these files to `xml`.

[tcbraster.txt](https://github.com/openwebwork/webwork2/files/11931186/tcbraster.txt)
[tcolorbox.txt](https://github.com/openwebwork/webwork2/files/11931187/tcolorbox.txt)
[tcolorboxmulticol.txt](https://github.com/openwebwork/webwork2/files/11931188/tcolorboxmulticol.txt)

pstaabp:
Overall this looks great.  However, if I add your custom files above, they are tracked.  How about a `custom` directory inside `assets/hardcopyThemes` that is in `.gitignore`.  

Also, is there a mechanism for instructors to do this for a particular course.   If we update `$hardcopyThemes` in `course.conf`, then that generates the menu of options, but how about searching for the theme files inside the templates directory of a course?

Alex-Jordan:
I'm making some enhancements following today's meeting. So it may be best not to review this quite yet.

Alex-Jordan:
OK, I made some changes that do some of what @pstaabp suggested. Now you can get on this branch (and the corresponding pg branch) and in a course:

- create a `hardcopyThemes/` folder inside `templates/` (this will be created as an empty folder in new courses).
- put in the theme files I posted earlier. Change the extensions to `.xml`.
- extend the list of available themes in `course.conf` with something like:
```
push (@{$hardcopyThemes}, 
   { id => 'myTheme1', file => "$courseDirs{hardcopyThemes}/tcolorbox.xml", pretty => "My Theme 1" },
   { id => 'myTheme2', file => "$courseDirs{hardcopyThemes}/tcolorboxmulticol.xml", pretty => "My Theme 2" },
   { id => 'myTheme3', file => "$courseDirs{hardcopyThemes}/tcbraster.xml", pretty => "My Theme 3" },
);
```

templates/ContentGenerator/Hardcopy/form.html.ep
```diff
@@ -153,9 +153,9 @@
 						<div class="input-group-text">
 							% for (@{ $ce->{hardcopyThemes} }) {
 								<label class="form-check-label me-3">
-									<%= radio_button hardcopy_theme => $_, class => 'form-check-input me-2',
-										$_ eq $ce->{hardcopyTheme} ? (checked => undef) : () =%>
-									<%= $ce->{hardcopyThemeNames}{$_} =%>
+									<%= radio_button hardcopy_theme => ($_->{id}), class => 'form-check-input me-2',
+										($_->{id}) eq $ce->{hardcopyTheme} ? (checked => undef) : () =%>
+									<%= $_->{pretty} =%>
```
drgrice1, line 158: There are unnecessary parenthesis around `$_->{id}`.  Please remove those.  They just make clutter that decrease readability.

conf/defaults.config
```diff
@@ -262,38 +234,43 @@ $perProblemLangAndDirSettingMode = "force::ltr";
 ################################################################################
 
 # The root directory, set by webwork_root variable in Apache configuration.
-$webworkDirs{root}          = "$webwork_dir";
+$webworkDirs{root}           = "$webwork_dir";
```
drgrice1, line 237: Don't try to align these variables.  Generally speaking when you cross comments alignment should restart (perltidy does this).  For now, just put these back the way they were.  Eventually it would be nice to perltidy this file.

conf/defaults.config
```diff
@@ -361,116 +338,115 @@ $webworkURLs{AuthorHelpURL}           ='https://webwork.maa.org/wiki/Category:Au
...
+$hardcopyThemes = [
+	{ id => 'empty',          file => "$webworkDirs{hardcopyThemes}/empty.xml",             pretty => 'Empty' },
+	{ id => 'basic',          file => "$webworkDirs{hardcopyThemes}/basic.xml",             pretty => 'Basic' },
+	{ id => 'oneColumn',      file => "$webworkDirs{hardcopyThemes}/oneColumn.xml",         pretty => 'One Column' },
+	{ id => 'twoColumn',      file => "$webworkDirs{hardcopyThemes}/twoColumn.xml",         pretty => 'Two Columns' },
+	#{id => 'hebrewOneColumn', file => "$webworkDirs{hardcopyThemes}/Hebrew-oneColumn.xml",  pretty => "Hebrew/English\x{2014}one Column"},
+	#{id => 'hebrewTwoColumn', file => "$webworkDirs{hardcopyThemes}/Hebrew-twoColumn.xml",  pretty => "Hebrew/English\x{2014}two Columns"},
+	#{id => 'myTheme',         file => "$courseDirs{hardcopyThemes}/myTheme.xml",            pretty => "My Theme"},
+	#{ id => 'myTheme',        file => "$courseDirs{hardcopyThemes}/tcolorbox.xml",          pretty => "My Theme" },
+];
```
drgrice1, line 425: Please change the "pretty" key to "label".  "pretty" sounds wrong to me here.  This really is a "label".  That is how it is used.

drgrice1 (changes requested):
I haven't delved too deeply into this yet, but here are some things that I see already.

I don't like that when a hardcopy is generated in the PG problem editor via HardcopyRenderedProblem.pm, there is now a horizontal rule at the top of the page.

The `\firstpageheader` from the theme is now written after the set header.  This means that it is impossible for the set header to override that.  This is a show stopper for me.  I don't want the default headers.

I am not entirely fond of having TeX in XML, but I can get over that I suppose.

Alex-Jordan:
I changed the minor things.

> I don't like that when a hardcopy is generated in the PG problem editor via HardcopyRenderedProblem.pm, there is now a horizontal rule at the top of the page.

That rule is part of the theme. It seems wrong to let certain things appear from the theme (like let the twoColumn theme use two columns and have a vertical line between them) but not let other things from the theme appear.

Could you render using the Basic theme or the Empty theme instead? Would it help to have a "Basic Two Columns" theme? Or a new theme that has the things you currently use, including a lack of a vertical rule? Would it help to allow for a different default theme for the PGEditor (eg Basic) than the default for full hardcopy production (eg oneColumn)?

> The \firstpageheader from the theme is now written after the set header. This means that it is impossible for the set header to override that. This is a show stopper for me. I don't want the default headers.

So, the point of much of this is that you could easily have your own hardcopy theme that does not have a `\firstpageheader`, or has different content in the `\firstpageheader`. You are saying that you want to be able to use the PG header to manage alterations to the the theme's header. But you don't have to do that anymore. Your PG headers can now be purely about contextual, pedagogical, instructional information for the students, divorced from the styling of the PDF. 

If you can send me a sample tex bundle for a homework set that is in a style you like, I could add that as a hardcopy theme that comes with the distribution. I don't even care to keep the current oneColumn and twoColumn exactly as they are. if yours is better, I'd change those to what you are using in production. Already this evening I realized that I need to change something about the `\firstpageheader` anyway, since I saw that is easier than I expected to have a long set name and a long user name and they butt into each other.

drgrice1:
> I changed the minor things.

Thanks.

> So, the point of much of this is that you could easily have your own hardcopy theme that does not have a `\firstpageheader`, or has different content in the `\firstpageheader`. You are saying that you want to be able to use the PG header to manage alterations to the the theme's header. But you don't have to do that anymore. Your PG headers can now be purely about contextual, pedagogical, instructional information for the students, divorced from the styling of the PDF.

This must not insist on an instructor writing a new theme to obtain this.  I can not approve this pull request in this way.  As I said, this is a show stopper.

drgrice1:
On the issue of the `\firstpageheader`, the deal is that I see no reason that what could easily be obtained before with only the hardcopy set header should now require writing an entire custom theme.

assets/hardcopyThemes/empty.xml
```diff
@@ -0,0 +1,39 @@
...
+\pagestyle{empty}
+\parindent=0pt
+\begin{document}
+</preamble>
+
+<setheader>
+</setheader>
+
+<problemheader>
+\ifthenelse{\equal{\webworkProblemId}{}\OR\equal{\webworkProblemId}{1}}{}{\makebox[\linewidth]{\rule{\paperwidth}{0.4pt}}}
```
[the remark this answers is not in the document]
Alex-Jordan: I thought of `Empty` as more intended for use with the PGEditor, where separation between exercises is not relevant. It wouldn't be good for an actual hardcopy if it doesn't label problem numbers. I'm going to change this to do a `\newpage` with each new problem.

If there is a better name than `Empty`, let's change the name. I mean for this to be a tool to look at problem output with essentially the absence of a theme.

conf/defaults.config
```diff
@@ -419,58 +399,53 @@ $webworkFiles{environment}                      = "$webworkDirs{conf}/defaults.c
...
+	{ id => 'twoColumn',      file => "$webworkDirs{hardcopyThemes}/twoColumn.xml",         label => 'Two Columns' },
+	#{id => 'hebrewOneColumn', file => "$webworkDirs{hardcopyThemes}/Hebrew-oneColumn.xml",  label => "Hebrew/English\x{2014}one Column"},
+	#{id => 'hebrewTwoColumn', file => "$webworkDirs{hardcopyThemes}/Hebrew-twoColumn.xml",  label => "Hebrew/English\x{2014}two Columns"},
+	#{id => 'myTheme',         file => "$courseDirs{hardcopyThemes}/myTheme.xml",            label => "My Theme"},
+];
 
-# The set divider goes between sets (in multiset output).
-#$webworkFiles{hardcopySnippets}{setDivider}     = "$webworkDirs{conf}/snippets/hardcopySetDivider.tex";
+# Default hardcopy theme (an id)
+$hardcopyTheme = 'oneColumn';
```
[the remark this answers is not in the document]
Alex-Jordan: I think I just lost track of what the default was while changing this to see other themes at some point in the process.

conf/defaults.config
```diff
@@ -1536,8 +1511,8 @@ $ConfigValues = [
 					. 'Columns theme is the traditional hardcopy format.  The One Column theme uses the full page '
 					. 'width for each column'
 			),
-			values  => $hardcopyThemes,
-			labels  => $hardcopyThemeNames,
+			values  => [ map {$_->{file}} @$hardcopyThemes ],
+			labels  => { map {$_->{file} => $_->{label}} @$hardcopyThemes },
```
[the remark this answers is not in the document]
Alex-Jordan: This was right prior to yesterday's changes to support @pstaabp 's suggestions. At that time, the `file` key was just the file name itself, without a path, because we could assume all theme files lived in the same folder. And there was no `id`, since then `file` could serve a dual role as an `id`. Thanks for catching the oversight.

Alex-Jordan:
>  The problem number is now being displayed as "Problem1",

Which theme did you see this with? These should all be written with `\webworkPageNumber` which is defined in `webwork2.sty` using a `~` for the space. I looked through the themes except Hebrew and I only see it like `Problem 1`.

drgrice1:
I see it with both the one and two column theme.

drgrice1:
In both themes this is added as `\webworkLocalizeProblem \thequestiontitle.`  I don't see a tilde between the word 'Problem' that comes from `\webworkLocalizeProblem` and the problem number that comes from `\thequestiontitle`.

Alex-Jordan:
Got it. My eyes/brain turned the "Problem 1" you wrote into "Page 1" and that is what I was looking at. Sorry!

drgrice1:
You also need to update the `check_latex.pl` script for the changes in this pull request.

Alex-Jordan:
> On the issue of the \firstpageheader, the deal is that I see no reason that what could easily be obtained before with only the hardcopy set header should now require writing an entire custom theme.

I accept that I haven't persuaded you yet on this principle, but the PG set header file can now be divorced from styling the PDF. And I feel that it should be divorced, and that will make some things better in the long run. But your method to override things in the theme using the PG header runs contrary to that principle. What happens when you override `\firstpageheader` in a PG header file, but then you (or someone downstream using your problem set) wants to switch to a different theme that is not based on the `exam` class? They probably can't, because your overrides will cause LaTeX errors with `\firstpageheader` not being defined.

Up to now it has been more work to make a new theme than to make a new PG header file. With this PR, they are a lot closer. Something I can still improve on is to make the Editor capable of editing a theme file like how it can edit a few other non-PG files. I'll give that a go. And then writing/editing a theme really would be an equivalent amount of work to writing/editing a PG header file.

A thought: consider a new WW-using instructor with little to no prior PG experience. And this person wants to customize the hardcopy headers/footers. Which will be a steeper obstacle for them?

1. Learning about `MODES` to edit a PG header file appropriately. You don't even get the luxury of disregarding the theme file(s), because you have to be familiar with what exactly you will override (`\firstpageheader` in this discussion).
2. Editing some tex snippets in a theme xml file, where you can see the impact of the `exam` class (or whatever else is in use) all in one place.

I feel that option 2 is easier for people unfamiliar with PG.

Also as I wrote before, I'm more than happy to change the default headers and footers to a format you are already using. I don't have the impression anyone is tied to what is there now. But I don't know what your format is yet.

drgrice1:
Prior to all of your recent hardcopy changes all an instructor needed to do was edit the default set header file and change a few lines.  That was even easier than any of the options you list.  They don't need to learn about `MODES` and such.  It was actually quite straightforward.  There was no need to override headers set in the themes.  The themes provided what was needed -- just layout, and there was no need to mess with themes at all unless you needed translations.  Prior to the exam class change even that was rarely needed.

drgrice1:
In addition, to create a theme you need to know LaTeX.  Most instructors know nothing about LaTeX.  This makes it vastly more difficult than before to do what you are saying.

Alex-Jordan:
You are not understanding me, and/or I am not writing things well. I have
used "create" and "edit" interchangeably and probably imprecisely. Stand by
and give me several days to meet your needs, since it takes me more time to
contribute code.

drgrice1:
I am sorry.  I don't mean to press you on this, and I don't mean to sound so harsh.  I have not focused on the positive aspects of this pull requests.  You have set up a versatile hardcopy theming system that looks very nice.  For my purposes, I can set up my own themes that do what I want.

I just think that there are many instructors that will feel the same way about the default page headers as I do.  Generally, I don't want much information in those headers, and use a more condensed format.  Also, since I don't use sections and recitations the default header becomes rather unbalanced.  There are three lines on the left, and only one ("student name (login id)") on the right.

I also don't like the links so much.  For this it might help if `hypersetup` were used to override the default `hyperref` settings that I think are rather ugly.  I always change those to make links look more like they do in the browser when I use the `hyperref` package.

In any case, an instructor that wants a simple way to change the header information with this system will find it much more challenging than before.  Intuitively speaking the set header pg file is where one would expect to set the header for the set in the pdf.  Until the recent changes was where that was done.  You mention divorcing the PG set header file from PDF styling, but the header in the PDF result is more than just a style and is really a set header (it is shown for each set and user).  It is part of the content.

Alex-Jordan:
It's no problem, your feedback and reviews are very much appreciated. And I
think it is all converging to a better thing all around. I will wait to get
into details because I don't always describe things well in summary, but I
am part way through making it just as easy to edit a theme. One click, then
edit a few lines as you would with a PG header, save, and off you go.

On the last notes, note that setting the content of the PDF's
headers/footers uses different tex macros when it's the exam class, the
article class with fancyhdr package, or something else. So if you override
these things in a PG header file, you won't have flexibility to change to a
different theme (well, maybe if you use a bunch of tex conditional macros)
. But you will still have that option if you really want to when I'm done
with revisions that are underway.



On Sat, Jul 8, 2023 at 4:27 AM Glenn Rice ***@***.***> wrote:

> I am sorry. I don't mean to press you on this, and I don't mean to sound
> so harsh. I have not focused on the positive aspects of this pull requests.
> You have set up a versatile hardcopy theming system that looks very nice.
> For my purposes, I can set up my own themes that do what I want.
>
> I just think that there are many instructors that will feel the same way
> about the default page headers as I do. Generally, I don't want much
> information in those headers, and use a more condensed format. Also, since
> I don't use sections and recitations the default header becomes rather
> unbalanced. There are three lines on the left, and only one ("student name
> (login id)") on the right.
>
> I also don't like the links so much. For this it might help if hypersetup
> were used to override the default hyperref settings that I think are
> rather ugly. I always change those to make links look more like they do in
> the browser when I use the hyperref package.
>
> In any case, an instructor that wants a simple way to change the header
> information with this system will find it much more challenging than
> before. Intuitively speaking the set header pg file is where one would
> expect to set the header for the set in the pdf. Until the recent changes
> was where that was done. You mention divorcing the PG set header file from
> PDF styling, but the header in the PDF result is more than just a style and
> is really a set header (it is shown for each set and user). It is part of
> the content.
>
> —
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-b46a4e6c5840fc5f&q=1&e=37ceecc5-c3e8-4aad-89e8-ad0f74675c03&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fwebwork2%2Fpull%2F2099%23issuecomment-1627145251>,
> or unsubscribe
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-72405446c48c338d&q=1&e=37ceecc5-c3e8-4aad-89e8-ad0f74675c03&u=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABEDOADZNET33S46QOM5AYLXPE725ANCNFSM6AAAAAAZ3BVKCI>
> .
> ***@***.***>
>


-- 
Alex Jordan
Mathematics Instructor
Portland Community College

Alex-Jordan:
I have this close to ready to push and ask for more review, but there's something I want to check in about. (I'll be doing day job stuff through tomorrow and I can't make progress, but I hope to have it pushed by the end of tomorrow.)

There are now two snippets for set header, one before the PG header and one after, which will make it possible to still override things using the PG header. I've simplified things so that hardcopy them files can only be at `webwork2/assets/hardcopyThemes` or `templates/hardcopyThemes`. You can have theme files in both places and config will see them all, but you choose which ones are enabled. If two share a filename, the one from your course will be used (imagine making a slightly altered copy of twoColumn, for instance). You choose which one is the default for hardcopy production, and separately which one is the default for PGEditor hardcopy production.

You can edit an existing theme file with the editor, and it saves to `templates/hardcopyThemes`. Either directly editing what was already there or making a copy of one from `webwork2/assets/hardcopyThemes`. But now i get to my question. I was looking for the simplest mechanism to trigger editing an existing theme. What I came up with was, in the PG editor, go to the hardcopy tab. The button there triggers some javascript call for building the PDF instead of the usual form submit. So i thought to add a button that would do the regular form submit. It submits the chosen hardcopy theme and has text like "Edit selected theme". It directs back to the editor, but with a new `hardcopy_theme` file mode (and the selected theme via the form submission). The theme file is opened from whichever of the two source folders it lives in, and a target file in `templates/hardcopyThemes` (possibly the same as the source file) is what it writes to. The last thing I haven't completed is displaying the XML niceely in the right panel which I think uses `instructor_rpc` but I just haven't looked at that yet.

My question now is if I've overlooked something and made a mistake by adding this form submission button to that form. Maybe it's hard to say without looking at the code in the updated PR. But does anything occur to you as making this a bad idea?

drgrice1:
Something sounds a bit off with that structure.  I can't say for certain though without seeing exactly what you have.

I was thinking about the issue with overriding the pdf page headers from the pg set header.  It seems that if there were TeX commands that were defined that provide the headers, say `\firstPageHeaderLeft`, `\firstPageHeaderCenter`, and `\firstPageHeaderRight`, and the hardcopy generation code would then define those to be the default headers that you are currently using, then the pg file could override those commands and thus change the headers.  Then any theme would just set the headers to the value of those commands.  Then it wouldn't matter if the them is using the exam class and its header mechanism, or using the `fancyhdr` package, and the pg set header could reliably override the default values.  I haven't done any real testing as to the feasibility of this idea, but it seems like it should work.

Alex-Jordan:
I'm trying to make this less of a beast to review. I just force pushed this after regrouping the edits. This only has the new theme structure itself, and not the ability to edit a theme using the Editor. Since there's so much to look at, if it is helpful I recorded a [screen capture](https://spot.pcc.edu/~ajordan/temp/hardcopy.mp4) to walk through the features, the way I think they are supposed to work.

I'll get to the editing of themes on another branch built on top of this one.

Alex-Jordan:
Synopsis:

* themes are XML files. Structure is described in `asssets/hardcopyThemes/README.md` or look at a theme in that folder as an example.
* themes now have a set header tex snippet before and after the PG header. So things in the "presetheader" snippet can be overridden by the PG heeader when that is needed.
* I put a bunch of themes in the distribution. They are all enabled (except hebrew) in `deefaults.config`. We could decide if we only want fewer enabled by default.
* You can have different default hardcopy themes for making hardcopies versus using the PG editor to make a hardcopy for a single problem you are editing.
* Themes can also live in `templates/hardcopyThemes`. This is where edited themes will go when I post the theme editing branch. In the video I posted you can see a course-specific theme put to use.
* If themes from `asssets/hardcopyThemes` and `templates/hardcopyThemes` have the same filename, the course version is used.
* Theme selection for building a regular hardcopy is still radio buttons if only three or fewer themes are enabled, otherwise it is a select.

Alex-Jordan:
I added a commit so that now, all course hardcopy theme files are just available no matter what. You only enable/disable the site hardcopy themes.

If you really don't want to have the option to use the course hardcopy themes, you could move them or rename them to not have a .xml extension. This will make things better once you can also edit themes, and then any new theme (or edited version of a site theme) you make will automatically be available.

Side note: running perltidy on my system has been wanting to make the following changes the past few days. But then GitHub's pertidy check does not like these changes. Do you know what might explain the disconnect?

```
diff --git a/lib/WeBWorK/ContentGenerator/Hardcopy.pm b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
index d1198334b..d76829508 100644
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -1194,10 +1194,10 @@ async sub write_problem_tex ($c, $FH, $TargetUser, $MergedSet, $themeTree, $prob
                                problemID => $MergedProblem->problem_id,
                        ),
                        $MergedProblem->problem_id == 0
-                               # link for a fake problem (like a header file)
+                       # link for a fake problem (like a header file)
                        ? (params =>
                                        { sourceFilePath => $MergedProblem->source_file, problemSeed => $MergedProblem->problem_seed })
-                               # link for a real problem
+                       # link for a real problem
                        : (),
                );
 
diff --git a/lib/WeBWorK/ContentGenerator/Instructor/Config.pm b/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
index 85d371221..636b8706b 100644
--- a/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Config.pm
@@ -120,9 +120,9 @@ sub getConfigValues ($c, $ce) {
        opendir(my $dh2, $localizeDir) || die "can't opendir $localizeDir: $!";
        my %seen      = ();    # find the languages in the localize direction
        my $languages = [
-               grep     { !$seen{$_}++ }                      # remove duplicate items
-                       map  { $_ =~ s/\.[pm]o$//r }               # get rid of suffix
-                       grep {/\.mo$|\.po$/} sort readdir($dh2)    #look at only .mo and .po files
+               grep { !$seen{$_}++ }                      # remove duplicate items
+               map  { $_ =~ s/\.[pm]o$//r }               # get rid of suffix
+               grep {/\.mo$|\.po$/} sort readdir($dh2)    #look at only .mo and .po files
 
        ];
```

lib/WeBWorK/ContentGenerator/Instructor/Config.pm
```diff
@@ -93,6 +93,28 @@ sub getConfigValues ($c, $ce) {
 	opendir(my $dh, $themeDir) || die "can't opendir $themeDir: $!";
 	my $themes = [ grep { !/^\.{1,2}$/ && $_ ne 'layouts' } sort readdir($dh) ];
 
+	# Get the list of all site hardcopy theme files
+	my $hardcopyThemeDirSite = $ce->{webworkDirs}{hardcopyThemes};
+	opendir(my $dhS, $hardcopyThemeDirSite) || die "can't opendir $hardcopyThemeDirSite: $!";
+	my $hardcopyThemesSite     = [ grep {/\.xml$/} (sort readdir($dhS)) ];
+	my $hardcopyThemeDirCourse = $ce->{courseDirs}{hardcopyThemes};
+	opendir(my $dhC, $hardcopyThemeDirCourse) || die "can't opendir $hardcopyThemeDirCourse: $!";
+	my $hardcopyThemesCourse = [ grep {/\.xml$/} (sort readdir($dhC)) ];
```
drgrice1, line 102: This should be

```suggestion
	my @hardcopyThemesCourse;
	if (opendir(my $dhC, $ce->{courseDirs}{hardcopyThemes})) {
		@hardcopyThemesCourse = grep {/\.xml$/} sort readdir($dhC);
	}
```

The same should be done with lines 101-102 of `lib/WeBWorK/ContentGenerator/Config.pm`, and lines 309-310 of `lib/WeBWorK/ContentGenerator/PGProblemEditor.pm`.  This causes an error for a course that does not have such a directory if any of those pages are opened.  Courses that are restored from older versions of webwork will not have that directory, and there is no reason to insist that they must have it.

Also, in all of these files there is no need for the local variables `$hardcopyThemeDirSite` and `$hardcopyThemeDirCourse` that are just copies of the course environment values `$ce->{webworkDirs}{hardcopyThemes}` and `$ce->{courseDirs}{hardcopyThemes}`.  Those hash keys will even work in interpolated strings.

templates/ContentGenerator/Instructor/PGProblemEditor/hardcopy_form.html.ep
```diff
@@ -51,9 +51,9 @@
 		<div class="col-auto">
 			<%= select_field 'action.hardcopy.theme' => [
 					map { [
-						$ce->{hardcopyThemeNames}{$_} => $_,
-						$_ eq $ce->{hardcopyTheme} ? (selected => undef) : ()
-					] } @{ $ce->{hardcopyThemes} }
+						stash('hardcopyLabels')->{$_} => $_,
+						$_ eq $c->ce->{hardcopyThemePGEditor} ? (selected => undef) : ()
+					] } @{ stash('hardcopyThemesAvailable') }
```
drgrice1, line 56: This is perhaps nicer as
```suggestion
						$hardcopyLabels->{$_} => $_,
						$_ eq $c->ce->{hardcopyThemePGEditor} ? (selected => undef) : ()
					] } @$hardcopyThemesAvailable
```
Stash values are made into variables in the templates by Mojolicious.  You do need to make sure that the stash values are defined though, or there are errors when the template is compiled.  So to be sure, you should add `$c->stash->{hardcopyLabels} = [];` to line 232 of the `initialize` method in `PGProblemEditor.pm` if you make the above change.

drgrice1 (commented):
This is getting close.  There is one issue that needs to be addressed though.

lib/WeBWorK/ContentGenerator/Instructor/Config.pm
drgrice1: I guess when an archive course is upgraded the directory structure will be repaired, and the directory will be created.  So this is perhaps not that significant.  It occurred for me in testing with a course that I didn't upgrade.

Alex-Jordan: I made these changes. Even if it wouldn't matter with an upgraded course, as long as I was removing the unnecessary local variables, I went ahead with the other change too.

Alex-Jordan:
Thanks for reviewing this, I made those changes.

drgrice1 (approved):
I think this is looking good now.  Although, I was testing this for a user without sufficient permissions and noticed some warnings.  One of those is due to this pull request, but several due to other changes made by @somiaj and myself.  I added a pull request (https://github.com/Alex-Jordan/webwork2/pull/19) to this branch that fixes these things.

assets/hardcopyThemes/basic.xml
```diff
@@ -0,0 +1,63 @@
...
+\parindent=0pt
+\pagestyle{fancy}
+\begin{document}
+</preamble>
+
+<presetheader>
+\fancyhead{}
+\renewcommand{\headrulewidth}{0pt}
+\ifthenelse{\equal{\webworkPrettySetId}{}}{}{\fancyhead[L]{\ifthenelse{\thepage=1}{\large\scshape\webworkLocalizeSet: \webworkPrettySetId}{}}}
+\ifthenelse{\equal{\webworkUserId}{}}{}{\fancyhead[R]{\ifthenelse{\thepage=1}{\bfseries\webworkLocalizeUsername: \webworkUserId}{}}}
```
drgrice1, line 23: To follow up on https://github.com/openwebwork/webwork2/pull/2099#issuecomment-1629233882, if this were
```suggestion
\fancyhead[L]{\webworkAssignmentCourseInfo}
\fancyhead[R]{\webworkUserInfo}
```
and all of the themes consistently used `\webworkAssignmentCourseInfo` and `webworkUserInfo` like this, the in the hardcopy set header file one could do
```
\renewcommand{\webworkAssignmentCourseInfo}{My custom left header content}
\renewcommand{\webworkUserInfo}{My custom right header content}
```
and reliably modify the headers for all themes without concern of if the exam class or fancyhdr package were used by the theme.

drgrice1: Note the theme itself could also use `\renewcommand` to change the default headers for the theme.  The `basic.xml` theme could do this to get the different headers you currently have set for this.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -309,18 +336,21 @@ sub path ($c, $args) {
...
+			$c->url_for('problem_detail', problemID => $c->stash('problemID') || ''),
+		$c->maketext('Editor') => ''
 	);
 }
 
 sub page_title ($c) {
 	my $setID     = $c->stash('setID');
 	my $problemID = $c->stash('problemID');
 
+	return 'Editor' unless $c->{file_type};
```
drgrice1, line 352: I just realized this should be `return $c->maketext('Editor') unless $c->{file_type};`.  My bad.  Can you make that change?

Alex-Jordan: Sure, I'll update it.

assets/hardcopyThemes/basic.xml
Alex-Jordan: I'll do something like this and update the branch. The purpose of the `\ifthenelse` is, in part, so that if you use the theme from the Editor (where there is no set ID or user ID), it gives no header content at all. But maybe it would be better to still show whatever the skeleton of the header would be. It's also messier with the ones relying on `fanccyhdr` where I don't think there is anything special to single out doing the first page headers differently than running headers.

drgrice1: So the theme could do
```
\renewcommand{\webworkAssignmentCourseInfo}{%
	\ifthenelse{\equal{\webworkPrettySetId}{}}{}%
	{\ifthenelse{\thepage=1}{\large\scshape\webworkLocalizeSet: \webworkPrettySetId}{}}%
}
\renewcommand{\webworkUserInfo}{%
	\ifthenelse{\equal{\webworkUserId}{}}{}%
	{\ifthenelse{\thepage=1}{\bfseries\webworkLocalizeUsername: \webworkUserId}{}}%
}
\fancyhead[L]{\webworkAssignmentCourseInfo}
\fancyhead[R]{\webworkUserInfo}
```
That should do the same thing in any case.

Alex-Jordan: I think I will change `\webworkAssignmentCourseInfo` and `\webworkAssignmentCourseInfo` to have generic names like `\webworkLeftHeader` and `\webworkRightHeader` and add commentary in the right places about customizing them. Probably add `\webworkCenterHeader` too. And as I've noted, I'll change the default content of these to make it less possible for a left and right header overstrike.

Alex-Jordan:
I changed the default `webwork2.sty` headers and footers that apply to several of the themes. See what you think?

There's an issue with using the footer to insert a copyright claim. It can happen that the footer is redefined to be a copyright claim, but then the set extends to one more page because of some unresolved content at that point in the processing. I tried many things to delay the resetting of the footer content, but it was always either not enough of a delay, or delayed it so much that the last page ended up without the copyright claim. So as things stand, occasionally the last _two_ pages get a copyright claim. I can live with this (after wasting too much time battling it). But if anyone would like to try some other approach, you might have more success than I did.

Aside from a footer, there are other ways to claim copyright. It could just be written as the last piece of content like it was before. Or it could be done with some kind of watermark or with thee `eso-pic` package (although these may have the same issues that I've had with footers.)

Alex-Jordan:
> It could just be written as the last piece of content

This comes with the risk of the copyright claim moving to a new page though.

drgrice1:
I think the new LaTeX commands for the headers and footers make this quite versatile.  Thanks for making that change.

I was unable to reproduce the footer issue, but that probably takes just the right (or rather wrong) problem lengths.

Alex-Jordan:
Thanks for the suggestions. The sty and theme files looks a lot cleaner now.

Yeah, I happen to be testing with a set where the last problem is just a bit too long and gets pushed to a new page with all of the two-column themes.

drgrice1:
Oh wait.  I found a set that gave the issue with the copyright.  That is not to serious.

Alex-Jordan:
I can think of one way to handle the copyright issue (that is too much work that I do not plan to do). That would be to change how hardcopies are made, so that each set is its own PDF. Then merge the PDFs at the end. There are several problems with this though, given how we do things now. But it would help address this issue because there are several tools for dealing the actual last page of a PDF, and those tools could make sure it only lands on the last page of each set.

drgrice1:
For now, this should work.  I think that trying to rework hardcopy generation to generate separate PDF's and merge them is a bit ambitious at this point.  I plan to release 2.18 next Thursday.

Alex-Jordan:
It's not only ambitious, I think I would not want the side effect of effectively forcing a newpage between sets.

Alex-Jordan:
> I plan to release 2.18 next Thursday.

OK, I'll clean up my `hardcopy-themes-editor` branch and open a PR then to hurry that up. I still haven't figured out why I can't get the save handler to work the way it should, but more eyes on it will probably see what I haven't been able to.

drgrice1:
There is something that I just noticed that needs to be changed.  Previously the `multicol` LaTeX package was always available even when the two column theme was not used.  That still needs to be the case.  The pg `DragNDrop.pm` package (used by the `draggableProof.pl` and `draggableSubsets.pl` macros) uses this in hardcopy.  So that package will need to be added to `webwork2.sty` (or `pg.sty` if that is more appropriate).

Alex-Jordan:
Yeah, `boxedRows.xml` uses a `tcbraster`. Inside a `tcbraster`, the `tcolorbox` blocks are not breakable and there is a note about this in the description element in the theme. Currently nothing is done with the descriptions, but maybe in the future they can be exposed to some sort of help tool.

Alex-Jordan:
I'll add `multicol` to `pg.sty`, since it supports a PG thing. But then also need to update some comments here on the webwork2 side.

pstaabp (approved):
I think this is ready to go after `multicols` is added in.

Alex-Jordan:
I made the `multicol` move both here and the pg side. Btw I'm suree the meeting is underway, but I will be late/absent. Have a student meeting.

PR #2100, opened 2023-07-04
use scalar return for grade_all_sets

Alex-Jordan:
Grade passback with LTIAdvanced is broken, but this fixes it. `grade_all_sets` returns based on `wantarray` and with some changes to using named subroutine parameters, it was giving the array version when it needs the scalar. The result was either an error if `$LTIGradeOnSubmit` is true, or an error every time a mass update was triggered.

It looks like this was already recognized and applied to the corresponding place in LTIAdvantage.

If we are moving more to named parameters, we may need to be aware of this effect. There are 45 instances of `wantarray` in 13 modules across `webwork2` and `pg`. Plus 20 instances in 10 macro libraries.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
@@ -126,7 +126,7 @@ async sub submit_course_grade ($self, $userID) {
 	$self->warning("lis_source_did is not available for user: $userID")
 		if !$user->lis_source_did && $ce->{debug_lti_grade_passback};
 
-	return await $self->submit_grade($user->lis_source_did, grade_all_sets($db, $userID));
+	return await $self->submit_grade($user->lis_source_did, scalar(grade_all_sets($db, $userID)));
 }
 
 # Computes and submits the set grade for $userID and $setID to the LMS.  For gateways the best score is used.


drgrice1:
Good catch.  I missed that in the clean up of LTIAdvanced/SubmitGrade.pm.

However, I want to clarify the actual reason this is needed.

First, I am not sure what you meant by "named paramaters"?  Do you mean the change to using "signatures" in the SubmitGrade.pm file?  That actually has nothing to do with this issue, and this would be needed even without the change to using "signatures".

The reason this is occurring is because previously the value of `grade_all_sets` was stored in a scalar variable (via `my $score = grade_all_sets($db,$userID);`.  That puts the return value in scalar context.  As such the `scalar` call was not needed as `wantarray` will choose to return the scalar return value.

However, when the return value is placed directly into the `submit_grade` call with `$self->submit_grade($user->lis_source_did, grade_all_sets($db, $userID));` the return value is in list context because it is in the list of parameters passed to the `submit_grade` method.  So `wantarray` returns the list return value.

drgrice1 (approved):

drgrice1:
Note that the change to using "signatures" makes the incorrect call clear.

If "signatures" were not being using for the `submit_grade` method, then the wrong thing would be passed to the method with no error.  That is the list return `($courseTotalRight, $courseTotal)`.  The result is that the `$courseTotalRight` would be used (and the second element `$courseTotal` ignored) instead of the `$courseTotalRight / $courseTotal`.

Using "signatures" turns this into an error because the incorrect number of arguments is passed to the `submit_grade` method.  Making it easier to find the mistake.

Alex-Jordan:
OK, I got the wrong idea then. I wanted to see what happened when I changed it to use a bunch of shift calls, and that made the error message go away. I incorrectly deduced that calling the argument with shift was somehow making `wantarray` false. But I understand you point that it was sending the wrong thing. And it did seem odd to me that `wantarray` would not have been determined at the time of the function call anyway.

pstaabp (approved):

PR #2103, opened 2023-07-06
have LTI 1.1 and 1.3 available by default

Alex-Jordan:


conf/authen_LTI.conf.dist
```diff
@@ -35,24 +35,21 @@ $debug_lti_grade_passback = 0;
...
 
-# This section enables LTI authentication.  If your site only uses LTI 1.1, then leave the
-# LTIAdvantage module commented out.  If your site only uses LTI 1.3, then comment out the
-# LTIAdvanced module and uncomment the LTIAdvantage module.  Both modules can enabled
-# (uncommented) for the site, but only one module can be used per course.  Make sure that the
-# matching configuration files are included below.  Failover to Basic_TheLastOption is necessary
-# to authenticate with cookie keys.
+# This section enables LTI authentication.  If a course uses LTI 1.1 (see $LTIVersion below),
+# then the LTIAdvanced module will be used.  If a course uses LTI 1.3 (see $LTIVersion below),
+# the LTIAdvantage will be used.  If you know a site will not use one or the other, it can be
```
drgrice1, line 40: This should probably read
```suggestion
# then the LTIAdvantage module will be used.  If you know a site will not use one or the other, it can be
```

drgrice1 (approved):
This should work.  Now action must be taken for either LTI version, but only one file additional `.dist` file must be copied if only one is needed.

PR #2104, opened 2023-07-07
Fix a warning that is being emitted due to the hack added in #2098

drgrice1:
This just assigns the result of calling `%+` to a dummy variable to avoid the warning that perl gives stating that this is a useless use of a variable in void context.

If you are using morbo this warning is currently displayed in the terminal as soon as the app starts.

I  also think this hack needs further investigation.


drgrice1:
You will also see this warning if you run the `check_latex` script.

pstaabp (approved):

Alex-Jordan:
I checked that removing the hack entirely still causes the original issue on my dev server.

This change avoids the original issue as well as prevents that perl warning. 

FWIW, I tried `if (%+) {};` as well, and that works too without creating `%dummy`.

Alex-Jordan (approved):

PR #2107, opened 2023-07-11
fix error with blank problem in new window

Alex-Jordan:
This fixes an error that happens when you go to Problem Editor, check the box to open in a new window, and then click to View.


drgrice1 (approved):
Looks good.

pstaabp (approved):

PR #2108, opened 2023-07-11
don't show PG stuff while editing course_info

Alex-Jordan:
When editing `course_info`, this suppress the PG help stuff at the top, as well as the selects for problem seed and display mode.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
@@ -233,6 +233,9 @@ sub initialize ($c) {
 	$c->stash->{formsToShow}      = ACTION_FORMS();
 	$c->stash->{actionFormTitles} = ACTION_FORM_TITLES();
 
+	# Tell the templates if we are working on a PG file
+	$c->{is_pg} = ($c->{file_type} eq 'course_info') ? 0 : 1;
+
 	# Check permissions
 	return
 		unless $authz->hasPermissions($user, 'access_instructor_tools')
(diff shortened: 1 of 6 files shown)


lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -233,6 +233,9 @@ sub initialize ($c) {
 	$c->stash->{formsToShow}      = ACTION_FORMS();
 	$c->stash->{actionFormTitles} = ACTION_FORM_TITLES();
 
+	# Tell the templates if we are working on a PG file
+	$c->{is_pg} = ($c->{file_type} eq 'course_info') ? 0 : 1;
```
drgrice1, line 237: Why do this?  There is nothing wrong with checking the value of `$c->{file_type}` in the templates.  This is unnecessary.

Alex-Jordan: In the hardcopy-theme branch, there is a new file type `hardcopy_theme`. All of the places that check for `'course_info'` will now be checking two things. Making this change now will just make the diff for that branch cleaner. But if you prefer I can leave it all in that branch.

drgrice1: I see.  I suppose it will be okay here then.

Alex-Jordan:
I fixed my typos in the commit message.

drgrice1 (approved):
This is good.  Those things are certainly not needed for a course info file.

pstaabp (approved):

PR #2109, opened 2023-07-11
clean up some comments, logic, and code

Alex-Jordan:
These are things in my hardcopy-themes branch that I cleaned up while working there. They should be unrelated to hardcopy-themes, but I wanted to change them when I noticed them.  If these are all OK and this is merged, there will be fewer distractions in the hardcopy-themes PR diff.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
@@ -89,14 +89,13 @@ The behavior on submit is defined by the value of $file_type and the value of
 the submit button pressed (the action).
 
     Requested actions and aliases
-        Save:                      action = save
-        Save as:                   action = save_as
-        View Problem:              action = view
-        Add this problem to:       action = add_problem
-        Make this set header for:  action = add_problem
-        Revert:                    action = revert
+        View/Reload                action = view
         Generate Hardcopy:         action = hardcopy
         Tidy Code:                 action = pgtidy
+        Save:                      action = save
+        Save as:                   action = save_as
+        Append:                    action = add_problem
+        Revert:                    action = revert
 
 An undefined or invalid action is interpreted as an initial edit of the file.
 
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -365,8 +361,9 @@ sub getRelativeSourceFilePath ($c, $sourceFilePath) {
 sub determineTempEditFilePath ($c, $path) {
 	my $user  = $c->param('user');
 	my $setID = $c->{setID} // 'Undefined_Set';
+	my $ce    = $c->ce;
```
drgrice1, line 364: There is no need to create a local reference to the `$c->ce` object.  That can be efficiently accessed via `$c->ce`.  This is going in the opposite direction of clean up.  I have worked to reduce this sort of local aliasing.  It just adds another step in looking up the object in question.

Alex-Jordan: OK, I get that. to confirm, it is more (or the same) efficiency to repeatedly run `$c` through `ce` than to just do it once and store it? I guess the `ce` method has no overhead now that I write this down.

It's no big deal but this comes from another thing in the hardcopy branch where I want to use `$c->ce->{harcopyTheme}` in a string concatenation like `"$foo$c->ce->{harcopyTheme}"`. Since `ce` is a method, that doesn't work. Changing to `"$foo$ce->{harcopyTheme}"` works. But I could use `$foo . $c->ce->{harcopyTheme}` instead.

drgrice1: Yeah, the accessor has minimal overhead. I suspected you might be using this in string concatenation, and in that case you will need to create a local alias. The accessor won't work for that.

Alex-Jordan: I'll just use `$foo . $c->ce->{harcopyTheme} . $bar` instead. It's only in one place and it's OK that way.

drgrice1 (approved):
This looks good.

pstaabp (approved):

PR #2114, opened 2023-07-15
Hardcopy themes editor

Alex-Jordan:
This is on top of #2099. At present only one commit (85e2f37) is on top of #2099.

Of course, first of all there may be issues with the code that's already here that I need to fix. But beyond that, this is incomplete because the save handler is not working. I've tried various things, and here I tried to leave out all the small messes that I made while trying to figure it out. Towards the end I was making some hail marys that did not land... If you have time and interest to look into it, I appreciate it. I'm running out of time.

With what is here, you can go to the Editor, Hardcopy tab, and click to edit a Hardcopy theme. Either it's already a local theme, and you start editing it, or it is a site theme and you are set up to use "save as" to create a new local theme. But when using the save tab, somewhere along the way, `$c->{editFilePath}` stops being something like `/opt/webwork/courses/mycourse/templates/hardcopyThemes/myTheme.xml` and is reborn as just `/opt/webwork/courses/mycourse/templates/hardcopyThemes/`, leading to issues.

And once that is worked out, there are a few more things.

- I haven't reviewed documentation, but I'm sure there are places I need to add help about this. Waiting to get it actually working first.
- With view and save tabs, I am hiding the "open in a new window" checkbox. In theory it would be nice if those led to building a dummy set in a new tab, using the temporarily edited theme, but I don't know if I'll get to that.
- The codemirror display is just showing the same XML as from the editing side. It would be nice to copy the temporary file somewhere public in `htdocs/tmp`, and then have the display panel load it in an iframe and the browser would use it's XML tree display mode. This would at least let you know when you have invalid XML.
- It might be nice if there were a codemirror  mode for XML with LaTeX coloring within specified elements.  But maybe not, as here we could have unmatched braces, etc, that match up within other XML elements.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
@@ -241,7 +241,7 @@ sub initialize ($c) {
 	$c->stash->{hardcopyLabels}   = [];
 
 	# Tell the templates if we are working on a PG file
-	$c->{is_pg} = !$c->{file_type} || $c->{file_type} ne 'course_info' && $c->{file_type} ne 'hardcopy_theme';
+	$c->{is_pg} = !$c->{file_type} || ($c->{file_type} ne 'course_info' && $c->{file_type} ne 'hardcopy_theme');
 
 	# Check permissions
 	return
--- a/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
@@ -20,11 +20,14 @@ use Mojo::Base 'WeBWorK::ContentGenerator', -signatures;
 
 WeBWorK::ContentGenerator::Instructor::PGProblemEditor - Edit a pg file
 
-This editor will edit problem files, set header files, or files such as
-course_info whose name is defined in the defaults.config file.
+This editor will edit problem files, set header files, hardcopy theme files,
+or files such as course_info whose name is defined in the defaults.config file.
 
 Only files under the template directory (or linked to this location) can be
 edited.
+
+Editable hardcopy themes are in the directory defined by
+$ce->{courseDirs}{hardcopyThemes}
 
 The course information and problems are located in the course templates
 directory.  Course information has the name defined by
@@ -48,7 +51,7 @@ The suffix for a temporary file is "user_name.tmp" by default.
 =item problem
 
 This is the most common type. This editor can be called by an instructor when
-viewing any problem.  the information for retrieving the source file is found
+viewing any problem.  The information for retrieving the source file is found
 using the problemID in order to look look up the source file path.
 
 =item source_path_for_problem_file
@@ -69,6 +72,12 @@ listed as problem 0 in the set's list of problems.  But it is used instead of
 set_header when producing a hardcopy of the problem set in the TeX format,
 instead of producing HTML formatted version for use on the computer screen.
 
+=item hardcopy_theme
+
+This allows editing of a hardcopy theme file, which defines snippets of tex
+code to be inserted before and after problems in a hardcopy.  It can be called
+from this module when using the Hardcopy tab.
+
 =item course_info
 
 This allows editing of the course_info.txt file which gives general information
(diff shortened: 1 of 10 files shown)


Alex-Jordan:
And of course I forgot perltidy. Now the one commit on top of #2099 is 15fbb85.

Alex-Jordan:
I have the save handler working. I wasn't storing the theme as a hidden input in the template.  I'll review this more, and perhaps add other things I mentioned, and push.

Alex-Jordan:
Once #2099 is settled, the diff for this should be more focused. It is ready for review. Some things could be improved upon, including some of the enhancements I noted in this thread. But I don't think there is time before the release.

lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
```diff
@@ -228,9 +238,10 @@ sub initialize ($c) {
 	$c->stash->{problemContents}  = '';
 	$c->stash->{formsToShow}      = ACTION_FORMS();
 	$c->stash->{actionFormTitles} = ACTION_FORM_TITLES();
+	$c->stash->{hardcopyLabels}   = [];
 
 	# Tell the templates if we are working on a PG file
-	$c->{is_pg} = ($c->{file_type} eq 'course_info') ? 0 : 1;
+	$c->{is_pg} = !$c->{file_type} || $c->{file_type} ne 'course_info' && $c->{file_type} ne 'hardcopy_theme';
```
drgrice1, line 244: I recommend changing this to
```suggestion
	$c->{is_pg} = !$c->{file_type} || ($c->{file_type} ne 'course_info' && $c->{file_type} ne 'hardcopy_theme');
```
This will work correctly without the parentheses in this case, but when `||` and `&&` are mixed it is always a good idea to make the intent clear with parentheses.

templates/ContentGenerator/Instructor/PGProblemEditor/hardcopy_form.html.ep
```diff
@@ -51,12 +51,15 @@
...
-					] } @{ $ce->{hardcopyThemes} }
+						$hardcopyLabels->{$_} => $_,
+						$_ eq $c->ce->{hardcopyThemePGEditor} ? (selected => undef) : ()
+					] } @$hardcopyThemesAvailable
 				],
 				id    => 'action_hardcopy_theme_id',
 				class => 'form-select form-select-sm d-inline w-auto' =%>
 		</div>
+		<div class="col-auto">
+			<button id="edit_hardcopy_theme" class="btn btn-secondary p-1" type="submit"><%= maketext('Edit Selected Theme') =%></button>
```
drgrice1, line 62: This line exceeds the 120 character limit.  Please change to
```suggestion
			<button id="edit_hardcopy_theme" class="btn btn-secondary p-1" type="submit">
				<%= maketext('Edit Selected Theme') =%>
			</button>
```

templates/ContentGenerator/Instructor/PGProblemEditor/save_as_form.html.ep
```diff
@@ -45,6 +47,9 @@
 		</div>
 		<%= hidden_field 'action.save_as.source_file' => $c->{editFilePath} =%>
 		<%= hidden_field 'action.save_as.file_type' => $c->{file_type} =%>
+		% if ($c->{file_type} eq 'hardcopy_theme') {
+			<%= hidden_field 'action.save_as.saveMode' => 'new_hardcopy_theme' =%>
+		% }
```
drgrice1, line 52: This is not needed with the handling that is done on lines 1143-1146 of PGProblemEditor.pm.

drgrice1 (commented):
This looks pretty good.  There are a few minor details that should be cleaned up.

Alex-Jordan:
Thanks, I made those changes.

drgrice1 (approved):

pstaabp:
Steps I did:
1. Went to the problem editor, Generate Hardcopy, selected Edit theme.  Opened the theme as expected.
2. Made a small edit, saved as 'mytheme.xml'. Noted that in the File Manager that the theme now appears in the `hardcopyThemes` directory as expected.
3. Returned to the problem editor (default new problem) -> Generate Hardcopy and the new theme is not there.

I figured out my problem. I didn't change the name of the theme--it actually is there.  

I would recommend putting some indication that if the theme name already exists that this will replace it (I know that is intended), but for a user who just didn't change the name, it's not clear that the new theme is there (with the same name as the previous one).

Alex-Jordan:
By "theme name" do you mean the label? Or the file name?  I'm not sure what you are describing because if you made a new file `mytheme.xml` that should show up separately in the list. But if you did not change the label then you should be seeing the same label appearing twice.

The README does describe how if you make a theme file name that matches a filename from the distribution, that the local course file will be the one that is used.

Issue #2115, opened 2023-07-16
With 2.18, using html2xml, sessions are invalidating each other 


Alex-Jordan:
[Here](https://spot.pcc.edu/~ajordan/temp/iframebug.html) is a sample page with two exercises embedded in iframes. Note this is a common situation for PTX books that were most recently built a few years ago (to use iframes directly, not `pretext-webwork.js`). But it is presently uncommon that they rely on a WW server that has upgraded to 2.18. Uncommon for now.

The iframe loading one of these invalidates the session for the other, whichever came first. So you can load this page and try entering answers, and with at least one of them you get a failed authentication method.

Worse, load this page in Firefox and then in Chrome. Return to FF and neither with work since Chrome load wiped the keys. So different people browsing the page can wipe each other's session keys.

This is not how it worked with 2.17 and earlier. It's not great but there just was no session key, I think, for html2xml connections made this way. 

Is it possible to restore the old behavior? I could possibly address the issue with PCC books, given some time. But there are who knows how many books that rely on the AIM WW server (presently 2.17). So I think I'd be stuck for a while unable to upgrade that to 2.18. And the same for Runestone's WW daemon server.

drgrice1:
It is not true that there were no session keys for webwork 2.17 with the html2xml endpoint.  In fact they were essential to the way that the html2xml endpoint was used internally by webwork (in the homework sets editor and such).  Something else is going wrong, and I am looking into it.

drgrice1:
The problem is that something is stripping the `passwd` from the parameters before the webwork webservice is called.  This means that the hidden input that should contain that information is empty.  Thus when you submit an answer, the session key is used instead.  I am not sure where or why this is being removed from the parameters.  I am still looking for that.

Alex-Jordan:
OK, I see. I always looked at this from the perspective of the params we built into the URL, and didn't think about session keys that I guess are put into a hidden input. Thanks for looking into it.

drgrice1:
By the way, the reason this was not a problem prior to 2.18 is that the parameter that was saved in the hidden field and used for authentication was `course_password` instead of `passwd`.  The `passwd` was being removed from the parameters before (although it actually wasn't there to begin with), but the `course_password` was allowed through.

drgrice1:
I believe this is fixed with #2116.

PR #2116, opened 2023-07-16
Don't strip the password from the input parameters for the rpc endpoints.

drgrice1:
Password authentication for the rpc endpoints relies on this, and this is the cause of issue #2115.

There really is no reason that the password should ever be stripped from the parameters.  I suspect someone at some point thought leaving it in the parameters after authentication is complete was some sort of security vulnerability.  It really isn't.  The parameters are secure in the system.  Nothing outside of the application can see them.  In fact no real security is even gained by removing this parameter either.  So the line modified here could simply be safely deleted instead.  For now, I left and only leave the parameter for the rpc endpoints though.


Alex-Jordan (approved):
Thanks, this of course works as you describe, and cleared up #2115 for the PTX books using our server.

pstaabp (approved):

PR #2117, opened 2023-07-16
pick and choose permission levels to receive feedback

Alex-Jordan:
This is an alternative to #1598. It allows you to set `$permissionLevels{receive_feedback} = ['ta', 'professor']` so that only those specific roles can receive feedback. There is no change in behavior if you use a string: `$permissionLevels{receive_feedback} = 'ta'` still means "ta and above".

I copied `checkboxlist.pm` to `permission_checkboxlist.pm` and tweaked it to work specifically for the list of permissions. I am unsure if that left some things present that are not needed.

Perhaps this is a model for any other permissionLevels where it would be better to allow you to pick and choose the roles instead of going with one role and above. Off the top of my head I can't think of another permissionLevel where you'd want this though.

### DIFF
--- a/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
+++ b/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
@@ -15,29 +15,10 @@
 
 package WeBWorK::ConfigObject::permission_checkboxlist;
 use Mojo::Base 'WeBWorK::ConfigObject', -signatures;
-
-my %userRoles = (
-	guest         => -5,
-	student       => 0,
-	login_proctor => 2,
-	grade_proctor => 3,
-	ta            => 5,
-	professor     => 10,
-	admin         => 20,
-	nobody        => 99999999,    # insure that nobody comes at the end
-);
-
-sub role_and_above {
-	my $role       = shift;
-	my $role_array = [$role];
-	for my $userRole (keys %userRoles) {
-		push @$role_array, $userRole if ($userRoles{$userRole} > $userRoles{$role});
-	}
-	return $role_array;
-}
+use WeBWorK::Utils 'role_and_above';
 
 sub display_value ($self, $val) {
-	$val = role_and_above($val) unless ref($val) eq 'ARRAY';
+	$val = role_and_above($self->{c}->ce->{userRoles}, $val) unless ref($val) eq 'ARRAY';
 	return $self->{c}->c(@{ $val // [] })->join($self->{c}->tag('br'));
 }
 
@@ -62,13 +43,13 @@ sub save_string ($self, $oldval, $use_current = 0) {
 }
 
 sub comparison_value ($self, $val) {
-	$val = role_and_above($val) unless ref($val) eq 'ARRAY';
+	$val = role_and_above($self->{c}->ce->{userRoles}, $val) unless ref($val) eq 'ARRAY';
 	return join(',', @{ $val // [] });
 }
 
 sub entry_widget ($self, $default) {
 	my $c = $self->{c};
-	$default = role_and_above($default) unless ref($default) eq 'ARRAY';
+	$default = role_and_above($self->{c}->ce->{userRoles}, $default) unless ref($default) eq 'ARRAY';
 	return $c->c(
 		map {
 			$c->tag(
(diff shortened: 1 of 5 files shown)


Alex-Jordan:
BTW, my thinking with making `permission_checkboxlist.pm` was that maybe this technique would be applied to more permissions in the future, and it would be useful to have that. But if this ends up only being good for `receive_feedback` then it is unnecessary and it could just use the existing `checkboxlist.pm`.

I'm interested in cleaning house with old PRs, and that led to this. #1598 has a reasonable underlying motivation, but I don't think it's best to have competing permissions like happens there. It's too easy to be confused when you change one and don't understand why you don't get the behavior you want, all because you are just unaware of the opposing permission.

lib/WeBWorK/Utils.pm
```diff
@@ -1840,6 +1841,16 @@ ID: foreach my $id (@problemIDs) {
 	return 1;
 }
 
+# Get the array of all permission levels at or above a given level
+sub role_and_above {
+	my ($c, $role) = @_;
```
drgrice1, line 1846: It would probably be better to make the first parameter here the course environment directly, since that is all that is needed for this method.  That means it can not be called as a package method, and it instead will need to be called as `role_and_above($ce, $role)`.  Then `role_and_above` should not be imported into the `WeBWorK::ContentGenerator::Feedback` module.

Or perhaps even better and more direct would be to only pass in a reference to the course environment `userRoles` array since in reality that is all this method needs.

lib/WeBWorK/Utils.pm
```diff
@@ -1848,18 +1859,25 @@ sub fetchEmailRecipients {
...
+		: $c->role_and_above($ce->{permissionLevels}{$permissionType});
+	@$roles = grep { defined $ce->{userRoles}{$_} } @$roles;
+	return unless $roles;
+
+	my $user_ids = [];
+	for my $role (@$roles) {
+		push(@$user_ids,
+			map { $_->user_id }
+				$db->getPermissionLevelsWhere({ permission => { '=' => $ce->{userRoles}{$role} } }));
+	}
```
drgrice1, line 1876: This should be changed to
```suggestion
	my @rolePermissionLevels = map { $ce->{userRoles}{$_} } grep { defined $ce->{userRoles}{$_} } @$roles;
	return unless @rolePermissionLevels;

	my $user_ids = [ map { $_->user_id } $db->getPermissionLevelsWhere({ permission => \@rolePermissionLevels }) ];
```
This is will get the same user ids from the datbase, and is more efficient as it is a single database query.

By the way, your original query could have been `$db->getPermissionLevelsWhere({ permission => $ce->{userRoles}{$role} })`.  Equality is the default, and does not need to be passed.  To go the other direction with my suggested query, it could also be `$db->getPermissionLevelsWhere({ permission => { '=' => \@rolePermissionLevels } })`.

drgrice1 (changes requested):
This has some things that need to be changed.  See the comments in the code.

I certainly think this is a better approach than adding a conflicting permission as in #1598.

Alex-Jordan:
Thanks for your review. I made all those changes and just pushed. I tested and it still works in both config page and actually sending emails to the right users.

drgrice1 (approved):
Looks good.  Thanks for making the changes.

lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
```diff
@@ -0,0 +1,75 @@
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2021 The WeBWorK Project, https://github.com/openwebwork
```
pstaabp, line 3: Update this copyright date.

pstaabp (approved):
Looks good.

somiaj:
I keep running into issues due to this change.  If you have a course which has set `$permissionLevels{receive_feedback}` in the course configuration page via `simple.conf` in the past (so the variable is a string) then this change makes it so the course configuration page errors out when trying to save any configuration tab except the email tab, because `$permissionLevels{receive_feedback}` is a string and not an `ARRAY`.  In essence, the course configuration page is broken until you update this configuration option to be an `ARRAY` in `simple.conf` if it is there.

It might be a bit late for this now, and now that I know what the issue is I can fix it with my courses, but it might be worth thinking about this for other users who may run across this issue.

somiaj:
To see the issue I just mentioned above, find a course and add a single line `$permissionLevels{receive_feedback} = 'nobody';`, then go to any configuration tab except the email tab and hit save changes, and you'll see an error about `'nobody'` not being an array reference.

drgrice1:
@somiaj:  If the course configuration page errors out, then a hot fix is needed.  If you have a fix, then could you please make a pull request?

drgrice1:
@somiaj: Nevermind, it seems that @Alex-Jordan has made a pull request to address this.

PR #2118, opened 2023-07-16
Fix codemirror folding for html and xml.

drgrice1:
Currently when you edit a course info file in the PG problem editor, there is a console log error.  This is because the PG code folding options do not work for htmlmixed mode.  This will also happen with the xml mode that will be added in pull request #2114.

This uses different options for htmlmixed or xml mode than for PG mode so that folding works for html and xml as well.

This also only enables comment toggling for PG mode.  That doesn't work for html and xml.

### DIFF
--- a/htdocs/js/PGCodeMirror/pgeditor.js
+++ b/htdocs/js/PGCodeMirror/pgeditor.js
@@ -69,16 +69,15 @@
 		return configName;
 	};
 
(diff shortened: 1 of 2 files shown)


Alex-Jordan (approved):
Tested. Saw the error before, saw it go away after.

drgrice1:
Note also, that if you merge this into #2114 and test this, you will see that code folding works.  You can collapse the xml sections in a theme file.

pstaabp (approved):
Looks good.

PR #2119, opened 2023-07-17
allow enter key to activate submit/check

Alex-Jordan:
This adds a hidden, unfocusable button preceding the Preview button, so that this new button will be the one that is activated when you have focus in an answer text input and you hit Enter.

This button is a copy of one of the three main buttons: Preview, Check, or Submit. Which one depends on the new config value `$enterKey`.
* `'preview'`: always the preview button (status quo, and the value in defaults.config)
* `'submit'`: always the submit button;  unless if the submit button not present, then the check button;  unless that not present too, and fall back to the preview button.
* `'hybrid'`: like 'submit' except if the problem only has finitely many attempts, then it's the preview button.

This was an issue for me today with a student who did not understand why their answer was "being counted wrong". They were hitting enter, it turns out, and interpreting the pink warning that it was only a preview as indicating they were incorrect. I plan to use `'hybrid'` in production.

I did not apply this to where these buttons come up in RPCRenderFormats. It seems unnecessary, but I could add that.

I know the release is soon, and aside from if I manage to fix the few bugs we know about, I won't make any more new PRs. Or if I do, I will target develop.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -188,6 +188,11 @@ $editor{textEdition} = '';
 $editor{textAuthor} = '';
 
 ################################################################################
+# Enter key behavior
+################################################################################
+$enterKey = 'preview';
+
+################################################################################
 # showMeAnother
 ################################################################################
 # switch to enable showMeAnother button
@@ -1591,6 +1596,21 @@ $ConfigValues = [
 			type    => 'number',
 			hashVar => '{pg}->{answersOpenAfterDueDate}'
 		},
+		{
+			var => 'enterKey',
+			doc => x('Enter Key Behavior'),
+			doc2 => x(
+				'If this is set to "preview", hitting the enter key on a problem page activates the "Preview My Answers" '
+					. 'button. If this is set to "submit", then the enter key activates the "Submit Answers" button '
+					. 'instead. Or if that button is not present, it will activate the "Check Answers" button. Or if '
+					. 'that button is also not present, it will just activate the "Preview My Answers" button. A '
+					. 'third option is "conservative". In this case, the enter key behaves like "preview" when the '
+					. '"Submit" button is available and there are only finitely many attempts allowed. Otherise the '
+					. 'enter key behaves like "submit".'
+			),
+			type => 'popuplist',
+			values => ['preview', 'submit', 'conservative']
+		}
 	],
 	[
 		x('Optional Modules'),
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -190,7 +190,7 @@ $editor{textAuthor} = '';
 ################################################################################
 # Enter key behavior
 ################################################################################
-$enterKey = 'preview';
+$pg{options}{enterKey} = 'preview';
 
 ################################################################################
 # showMeAnother
@@ -1596,21 +1596,6 @@ $ConfigValues = [
 			type    => 'number',
 			hashVar => '{pg}->{answersOpenAfterDueDate}'
 		},
-		{
-			var => 'enterKey',
-			doc => x('Enter Key Behavior'),
-			doc2 => x(
-				'If this is set to "preview", hitting the enter key on a problem page activates the "Preview My Answers" '
-					. 'button. If this is set to "submit", then the enter key activates the "Submit Answers" button '
-					. 'instead. Or if that button is not present, it will activate the "Check Answers" button. Or if '
-					. 'that button is also not present, it will just activate the "Preview My Answers" button. A '
-					. 'third option is "conservative". In this case, the enter key behaves like "preview" when the '
-					. '"Submit" button is available and there are only finitely many attempts allowed. Otherise the '
-					. 'enter key behaves like "submit".'
-			),
-			type => 'popuplist',
-			values => ['preview', 'submit', 'conservative']
-		}
 	],
 	[
 		x('Optional Modules'),
@@ -2037,6 +2022,23 @@ $ConfigValues = [
 			values => [qw(Percent Point Both)],
 			type   => 'popuplist'
 		},
+		{
+			var => 'pg{options}{enterKey}',
+			doc => x('Enter Key Behavior'),
+			doc2 => x(
+				'If this is set to "preview", hitting the enter key on a homework problem page activates the "Preview '
+					. 'My Answers" button.  If this is set to "submit", then the enter key activates the "Submit '
+					. 'Answers" button instead.  Or if that button is not present, it will activate the "Check '
+					. 'Answers" button.  Or if that button is also not present, it will activate the "Preview My '
+					. 'Answers" button.  A third option is "conservative". In this case, the enter key behaves like '
+					. '"preview" when the "Submit" button is available and there are only finitely many attempts '
+					. 'allowed.  Otherise the enter key behaves like "submit".  Note that this is only affects '
+					. 'homework problem pages, not test/quiz pages, and not instructor pages like the PG Editor '
+					. 'and the Library Browser.'
+			),
+			type => 'popuplist',
+			values => ['preview', 'submit', 'conservative']
+		}
 	],
 	[
 		x('E-Mail'),
--- a/conf/localOverrides.conf.dist
+++ b/conf/localOverrides.conf.dist
@@ -312,7 +312,7 @@ $mail{feedbackRecipients}    = [
 # A third option is "conservative". In this case, the enter key behaves like "preview"
 # when the "Submit" button is available and there are only finitely many
 # attempts allowed. Otherise the enter key behaves like "submit".
-#$enterKey = 'conservative';
+#$pg{options}{enterKey} = 'conservative';
 
 ################################################################################
 # Periodic re-randomization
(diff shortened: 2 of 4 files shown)


Alex-Jordan:
I changed `'hybrid'` to `'conservative'`.

templates/ContentGenerator/Problem/submit_buttons.html.ep
```diff
@@ -1,13 +1,28 @@
...
+%		$enter_key_name = 'checkAnswers';
+%		$enter_key_value = maketext('Check Answers');
+%	}
+% }
 %
 % if ($will{requestNewSeed}) {
 	<%= submit_button maketext('Request New Version'), id => 'submitAnswers_id', name => 'requestNewSeed',
 		formtarget => '_self', class => 'btn btn-primary mb-1' =%>
 % } else {
+	<%= submit_button $enter_key_value, id => 'enter_key_submit', formtarget => '_self', name => $enter_key_name, tabindex => -1,
```
drgrice1, line 24: Move `tabindex => -1,` to the next line.  This line exceeds the 120 character limit.

Note that I recommend using `editorconfig` to help with this.  Most text editors support it.  Some out of the box, and some with a plugin.  See https://editorconfig.org/.  You will need to add a `.editorconfig` configuration file to the webwork2 root directory (I intend to add one to the repository soon anyway).  See https://github.com/openwebwork/webwork3/blob/main/.editorconfig for an example of such a file.  That one will work for most of webwork2's file types, although it also covers some not used by webwork2.

conf/defaults.config
```diff
@@ -187,6 +187,11 @@ $editor{textTitle} = '';
 $editor{textEdition} = '';
 $editor{textAuthor} = '';
 
+################################################################################
+# Enter key behavior
+################################################################################
+$enterKey = 'preview';
```
drgrice1, line 193: I think this option should be scoped in one of the existing hashes.  Probably in the `pg{options}` hash.  So then this would be `$pg{options}{enterKey} = 'preview';`.

conf/defaults.config
```diff
@@ -1591,6 +1596,21 @@ $ConfigValues = [
...
+					. 'button. If this is set to "submit", then the enter key activates the "Submit Answers" button '
+					. 'instead. Or if that button is not present, it will activate the "Check Answers" button. Or if '
+					. 'that button is also not present, it will just activate the "Preview My Answers" button. A '
+					. 'third option is "conservative". In this case, the enter key behaves like "preview" when the '
+					. '"Submit" button is available and there are only finitely many attempts allowed. Otherise the '
+					. 'enter key behaves like "submit".'
+			),
+			type => 'popuplist',
+			values => ['preview', 'submit', 'conservative']
+		}
```
drgrice1, line 1613: I think it would be (somewhat) more intuitive if this were in the "Problem Display/Answer Checking" tab instead of the "General" tab.

htdocs/themes/math4/math4.scss
```diff
@@ -743,6 +743,11 @@ input.changed[type=text] { /* orange */
 	row-gap: 0.25rem;
 }
 
+.visually-hidden-unfocusable {
+	position: absolute;
+	left: -9999px;
```
drgrice1, line 748: I think that this should include `visibility: false`.  Otherwise, won't this be announced by a screen reader?  As I understand it screen readers don't care about position.

templates/ContentGenerator/Problem/submit_buttons.html.ep
```diff
@@ -1,13 +1,28 @@
...
+% 	if ($can{getSubmitButton}) {
+%		$enter_key_name = 'submitAnswers';
+%		$enter_key_value = (param('user') ne $effectiveUser)
+%			? maketext('Submit Answers for [_1]', $effectiveUser)
+%			: maketext('Submit Answers');
+%	} elsif ($can{checkAnswers}) {
+%		$enter_key_name = 'checkAnswers';
+%		$enter_key_value = maketext('Check Answers');
+%	}
+% }
```
drgrice1, line 18: Please move this added code into the `else` statement below.  Keep the logic together with the the one place where these new variables are used.

Also, I recommend that you place the indentation before the percent symbol in templates.  That is how it is done elsewhere in this file.  I have done that in most cases in the templates, except some that I did initially, and then didn't go back to and fix.  Putting the percent symbol at the beginning of the block results in mixed tabs and spaces which most linters don't like.  It also results in the indentation after the percent being off by one space due to the first level having the one space indentation.  So make this:
```suggestion
% my $enter_key_name  = 'previewAnswers';
% my $enter_key_value = maketext('Preview My Answers');
% if ($c->ce->{enterKey} eq 'submit' || ($c->ce->{enterKey} eq 'conservative' && $c->{problem}{max_attempts} < 0)) {
	% if ($can{getSubmitButton}) {
		% $enter_key_name  = 'submitAnswers';
		% $enter_key_value = (param('user') ne $effectiveUser)
			% ? maketext('Submit Answers for [_1]', $effectiveUser)
			% : maketext('Submit Answers');
	% } elsif ($can{checkAnswers}) {
		% $enter_key_name  = 'checkAnswers';
		% $enter_key_value = maketext('Check Answers');
	% }
% }
```

drgrice1 (commented):
This will need more thought.  As this is currently implemented, it only works as long as MathQuill is not enabled.  This can be updated to work with MathQuill enabled, but will take a bit more work and PG pull request.

htdocs/themes/math4/math4.scss
drgrice1: Note that with the `visibility: hidden` attribute, the `tabindex => -1` attribute added in `submit_buttons.html.ep` is not needed.

drgrice1:
It is actually quite easy to make this work with MathQuill.  It does take a pg pull request though.  All that is needed is to change line 480 from `document.getElementById('previewAnswers_id')?.click();` to `document.getElementById('enter_key_submit')?.click();`.  (The comment on line 471 should be updated too.)

conf/defaults.config
drgrice1: I think it should be noted somewhere that this only applies to regular sets, and not to gateway quizzes.

Alex-Jordan:
Thanks for reviewing this and your insight about Mathquill. I am out sick, will be slow to respond. Hopefully better tomorrow but have to catch up on day job stuff I missed today.

templates/ContentGenerator/Problem/submit_buttons.html.ep
Alex-Jordan: I'll remove `tabindex` altogether. Didn't know about `visibility: hidden`, thanks. Seems that is even better anyway, because `tabindex=-1` still leaves the button in the accessibility tree and potentially exposed to screen readers.

htdocs/themes/math4/math4.scss
Alex-Jordan: Thanks, I made that change.

conf/defaults.config
Alex-Jordan: OK, I moved it. It seemed `webwork2`-ish, not `pg`-ish, so I didn't put it there before and there did not seem to be a good fit hash. But I now see that `pg{options}` is full of things that are `webwork2`-ish already.

conf/defaults.config
Alex-Jordan: I made it so.

conf/defaults.config
drgrice1: Yeah, there are several that are `webwork2`-ish I think.  Show me another and periodic randomization for instance.  This is similar in nature to those.  That is why I suggested it go there.

templates/ContentGenerator/Problem/submit_buttons.html.ep
Alex-Jordan: Done.

Alex-Jordan:
In the case where the course uses periodic re-randomization, you could be on a page with only the "Request New Version" button. Without MathQuill, hitting enter from an input field would activate this. But with MQ, nothing would happen. So I added another button with the `#enter_key_submit` id so that MQ will do the same as without MQ.

conf/localOverrides.conf.dist
```diff
@@ -301,6 +301,19 @@ $mail{feedbackRecipients}    = [
...
+################################################################################
+# If this is set to "preview", hitting the enter key on a problem page activates
+# the "Preview My Answers" button. If this is set to "submit", then the enter
+# key activates the "Submit Answers" button instead. Or if that button is not
+# present, it will activate the "Check Answers" button. Or if that button is
+# also not present, it will just activate the "Preview My Answers" button.
+# A third option is "conservative". In this case, the enter key behaves like "preview"
+# when the "Submit" button is available and there are only finitely many
+# attempts allowed. Otherise the enter key behaves like "submit".
+#$enterKey = 'conservative';
```
drgrice1, line 315: You forgot to change this to match what is now in `defaults.config`.

drgrice1:
Note to @drdrew42:  The standalone renderer will need an adjustment for this.

conf/localOverrides.conf.dist
Alex-Jordan: Thanks for catching that. I updated and force pushed.

drgrice1 (approved):
I think this looks good.

PR #2120, opened 2023-07-19
stop supporting a typo

Alex-Jordan:
This may be controversial, and I'm fine if this is rejected.

My problem with supporting this typo in old set header files, is that while doing dev work, I saw "formated" appear in a list of hash keys and I thought surely that's a typo. So I wasted time looking into it only to find it is supposed to be there. But is it time to make people correct their typos?

### DIFF
--- a/courses.dist/modelCourse/templates/setDemo/paperHeaderFile1.pg
+++ b/courses.dist/modelCourse/templates/setDemo/paperHeaderFile1.pg
@@ -9,7 +9,7 @@ loadMacros(
 "PGanswermacros.pl"
 );
 
-$dateTime = $formatedDueDate;
+$dateTime = $formattedDueDate;
 
 TEXT(EV2(<<EOT));
 \noindent {\large \bf $studentName}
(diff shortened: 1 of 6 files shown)


courses.dist/modelCourse/templates/setDemo/paperHeaderFile1.pg
```diff
@@ -9,7 +9,7 @@ loadMacros(
 "PGanswermacros.pl"
 );
 
-$dateTime = $formatedDueDate;
+$dateTime = $formattedDueDate;
```
drgrice1, line 12: It would be better to delete this intermediate `$dateTime` variable, and just use `$formattedDueDate` below like all of the other files do.

drgrice1 (approved):
I am fine with this.  I didn't really want to add the `formated...` date parameters in `lib/WeBWorK/Utils/Rendering.pm`, but didn't want to stir up the controversy at that time.

Alex-Jordan:
Thanks, I made those deletions.

pstaabp (approved):

PR #2121, opened 2023-07-19
help tooltips for the set details page

Alex-Jordan:
Please proofread for accuracy, important omissions, and typos.

Perhaps controversial: I moved some things like `(0=infinity)" into the help text, explained with complete sentences and using words like "unlimited" instead of "infty". So now such things are not immediately visible. It just seemed like they were part of the description of the item, not the item's label.


lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -112,22 +112,43 @@ use constant FIELD_PROPERTIES => {
...
-		size     => '25',
-		override => 'any',
+		name      => x('Opens'),
+		type      => 'edit',
+		size      => '25',
+		override  => 'any',
+		help_text => x(
+			'This is generally the date when students can begin visiting the set and submitting answers. '
+				. 'Prior to this date, if the set is assigned to a user and it is flagged "visible", '
+				. 'they can see that it exists and when it will open, but cannot view the problems.'
```
drgrice1, line 122: There is a missing space after the period at the end of this line inside the quote.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -112,22 +112,43 @@ use constant FIELD_PROPERTIES => {
...
+				. 'If using "course" grade passback to an LMS, only those sets that are past their open date '
+				. "are factored in to the overall course grade that is passed back.  Note that certain "
```
drgrice1, line 124: This line should use single quotes like the other lines.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -112,22 +112,43 @@ use constant FIELD_PROPERTIES => {
...
-		size     => '25',
-		override => 'any',
+		name      => x('Closes'),
+		type      => 'edit',
+		size      => '25',
+		override  => 'any',
+		help_text => x(
+			'This is generally the date when students can no longer use the "Submit" button to submit an '
+				. 'answer and have it assessed for credit.  However students can still visit the set, '
+				. 'type or select answers, and use the "Check" button to be assessed without credit. '
```
drgrice1, line 136: The actual button text is "Check Answers".  Although everyone will probably know what is meant, it is still probably best to use the actual button text.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -219,14 +250,32 @@ use constant FIELD_PROPERTIES => {
...
 		},
+		help_text => x(
+			'With "homework", students visit each problem one at a time.  They submit answers for one problem at a '
+				. 'time and immediately receive feedback.  With "test", students will submit all answers for all '
+				. 'problems at once.  They may or may not receive feedback right away depending upon other '
+				. 'settings.  Also a "test" can have a time limit, where the student needs to start between the '
+				. 'open date and the close date, but once started has only so much time.  Also a "test" can '
+				. 'be configured to allow taking new, re-randomized versions.  A "proctored test" is the same as '
+				. 'a "test", but in order to begin, either a classwide password specific to this test is needed, '
+				. 'or a higher level user must enter their username and password on the student \'s screen. '
```
drgrice1, line 261: There should not be a space between "student" and "\'s".

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -270,16 +339,29 @@ use constant FIELD_PROPERTIES => {
...
+		default   => '1',
+		help_text => x(
+			'A test is broken up into pages with this many problems on each page.  Students can move from page to page '
+				. 'without clicking to grade the test, and their temporary answers will be saved.  Use "0" to indicate '
+				. '"all problems on one page".  For tests with many problems, either extreme (1 per page or "all on one '
+				. 'page") has drawbacks.  With 1 per page, the student has many pages and may be frustrated trying to go '
+				. 'back and find a particular problem.  With "all on one page", the student may spend a lot of time on '
+				. 'that one page without clicking anything that lets WeBWorK know they are still active, and their '
+				. 'session might expire for inactivity before they get around to clicking the grade button.'
+		)
```
drgrice1, line 361: I am not sure how detailed we want to get here, but there is more that could be said.

You note that the session may expire if a student is sitting on a page with a large number of problems for a long time, but it might be worth noting that in this situation not only will the session expire, but all answers the student had entered on that page will be lost.

It also may be worth noting that a test with a large number of problems on a single page is more taxing on the server when a student changes pages.  Although this may be more of a system administrative concern.  On a related note, if a page has several problems that use LaTeX/TikZ images, the page can load quite slowly.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -237,29 +286,49 @@ use constant FIELD_PROPERTIES => {
...
+				. 'at the close date or allow them the full time limit.'
+		)
 	},
 	attempts_per_version => {
-		name     => x('Number of Graded Submissions per Test (0=infty)'),
-		type     => 'edit',
-		size     => '3',
-		override => 'any',
-		default  => '0',
+		name      => x("Graded Submissions per Version"),
```
drgrice1, line 295: You removed "Number of" here.  Should that stay?

Also, perhaps for clarity, this should specify "per Test Version" rather than "per Version"?  If done here, then it should probably be done elsewhere also.

Alex-Jordan: > You removed "Number of" here. Should that stay?

It started because after other changes, this was now the widest label and doesn't seem to need to be so wide. I think it's clear from context and from the default value being populated as an integer. But now that you point it out there are other "Number of" labels and maybe it should be consistent on way or the other. I'll go with whichever is your preference.

> Also, perhaps for clarity, this should specify "per Test Version" rather than "per Version"?

This was also motivated by making the labels not be so long. But to me it seems clear from context. If you confirm it's better as "Test Version", I'll them all to be that way.

drgrice1: I think it's clear enough as you have it. Long labels do interfere with a nice display format.

Alex-Jordan:
I kept going down the rabbit hole and added more help content for options I wasn't looking at before. And I applied some consistency with capitalization and other things to the labels. And I made the subdivisions have `th` rows as clear dividers.

drgrice1:
I think this still looks good.  I am not entirely sold on the `th` row dividers for the problems when editing a set for a user.  I think it makes it look a bit bulky.  I can live with it though.

pstaabp (approved):

Alex-Jordan:
I'm looking into adjusting it when editing for a user.

Alex-Jordan:
At least if it's a quiz, editing for a user still has a lot going on and I find the `th` separators helpful. Were you looking at a quiz, or just homework?  Actually for me even in homework it's not bad, but I've enabled everything so I see more fields than I would typically see.

Screen Shot 2023-07-20 at 11 20 22 AM

drgrice1:
It was for all sets.  I wasn't looking at the header in the "Set Parameters" section.  That looks fine.  However, that is also shown for each problem when editing a set for a user.

Alex-Jordan:
Oh I see! I can remove that.

Alex-Jordan:
OK there's a commit on top you can review now about that.

drgrice1:
That looks better I think.

PR #2122, opened 2023-07-19
Correct htmlPath and imagesPath settings in defaults.config

drgrice1:
These paths have been incorrect since 2015 when they were added.  The comments that was there about these are completely incorrect.  They state that these should be complete urls.  They should not be urls at all. Instead they should be directories.  The only place that these are used is in PGalias.pm to search for auxiliary html files or images, and there they are treated as directories.

Also remove the completely unused pdfPath setting.


Alex-Jordan (approved):
Looks good.

pstaabp (approved):

PR #2139, opened 2023-07-20
move help link outside of h1

Alex-Jordan:
This moves the question mark help icons next to each h1 outside of the h1 (but still visually in basically the same place).

Without this, when you look at a heading tree for a page, you see a little question mark at the end of each h1, and that shouldn't be in the actual heading.

### DIFF
--- a/htdocs/themes/math4/math4.scss
+++ b/htdocs/themes/math4/math4.scss
@@ -380,6 +380,11 @@ div.showMeAnotherBox {
 h1.page-title {
 	font-size: 31.5px;
 	line-height: 35px;
+	display: inline-block;
+}
+h1.page-title + a {
+	font-size: 31.5px;
+	margin-left: 0.5rem;
 }
 
 h2.page-title {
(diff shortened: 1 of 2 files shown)


htdocs/themes/math4/math4.scss
```diff
@@ -380,6 +380,11 @@ div.showMeAnotherBox {
 h1.page-title {
 	font-size: 31.5px;
 	line-height: 35px;
+	display: inline-block;
+}
+h1.page-title + a {
+	font-size: 31.5px;
+	margin-left: 0.5rem;
 }
```
drgrice1, line 388: I think the help `h.page-title + a` also needs the `line-height`.  Otherwise the page title gets uneven spacing above and below it.

Also, take advantage of scss here.  So make this:
```suggestion
h1.page-title {
	font-size: 31.5px;
	line-height: 35px;
	display: inline-block;

	& + a {
		font-size: 31.5px;
		line-height: 35px;
		margin-left: 0.5rem;
	}
}
```

drgrice1 (approved):
This looks good.  I suggest a small tweak though.

pstaabp (approved):

pstaabp:
Oops.  The changes hadn't been made yet.

PR #2144, opened 2023-07-22
Possible fix for CAS.pm authentication issue.

drgrice1:
This replaces the mod_perl `$r->unparsed_uri` method with Mojo::Message::Request method `$c->req->url->to_string` which should return the same thing.  This may fix the `CAS.pm` module.  This is the only mod_perl call in that file, and was missed in the conversion to Mojolicious.

I have not way to test this, but clearly that method must be replaced.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8317.

### DIFF
--- a/lib/WeBWorK/Authen/CAS.pm
+++ b/lib/WeBWorK/Authen/CAS.pm
@@ -115,7 +115,7 @@ sub get_credentials {
 		#    CAFile => $cas_certs);
 		my $cas = new AuthCAS(%{ $ce->{authen}{cas_options}{AuthCAS_opts} });
 
-		my $service = $c->unparsed_uri();
+		my $service = $c->req->url->to_string;
 		# Remove the "ticket=..." parameter that the CAS server added
 		# (Not sure if the second test is really needed.)
 		$service =~ s/[?&]ticket=[^&]*$//


taniwallach:
@paulvojta Could you look at WW 2.18 and this. Someone using CAS needs to test this.

drgrice1:
I suspect that the CAS.pm module already had issues starting with webwork 2.16.  I think this fixes the issue with the mod_perl call, but there are other issues.  The external authentication code that shows the message about the course using an external authentication system assumes the authentication method is LTI authentication.  For CAS authentication those variables won't be defined.  That code was added for webwork 2.16.

paulvojta:
Hi,

I'm sorry, but I can't test it right now.  We haven't upgraded to 2.18 here
at Berkeley, and it would be highly unwise to do so right now, since there's
a summer-session course using it (WeBWorK 2.17).  

But, it looks like the fix should work.

Sincerely,


Paul Vojta

paulvojta:
Hi,

We've been using CAS authentication at Berkeley for WW 2.17.
(Strictly speaking, we primarily use LTI authentication for students;
professors and TAs can use either LTI or CAS (sometimes they'd need two
WeBWorK accounts to have both options), and admins use CAS only.

For admins, I don't think there's a reasonable way to let them in via LTI.
(Correct me if I'm wrong.)

We'll always need something else for admin logins, and password
authentication doesn't meet our security standards.

As for the message about external authentication, in WW 2.17 that message
is in lib/WeBWorK/ContentGenerator/Log{in,out}.pm, but I don't see it there
in WW 2.18 (and it's hard to grep things using github.io on the web).
But, as far as 2.17 is concerned, the only variables are $ce->{LMS_url}
and $ce->{LMS_name}, and those can be set if desired.

But also, isn't Shibboleth.pm in the same boat?

Sincerely,


Paul Vojta

Alex-Jordan:
I'm not entirely sure what you mean about admin login and LTI, but maybe
this helps.  If the WW course has an existing user with username that
matches your id in the LMS, you can get in with LTI authentication. For
example with our LMS (D2L) from any LMS course I can create an external
learning tool to any WW course, and let myself in to that WW course. Again
though, this works because I already have an admin account in that WW
course with a username matching my LMS account.

drgrice1:
Those messages are now in `templates/ContentGenerator/Login.html.ep` and `templates/ContentGenerator/Logout.html.ep`.

Shibboleth is in even worse shape that CAS.  It uses environment variables from the apache2 mod_shib module that can not be accessed from the webwork2 Mojolicious app.

sarunasb:
WW 2.18 on Ubuntu 22.04.2. The fix works, i.e. can log in using CAS and use course.

Issue #2146, opened 2023-07-23
error/warning about set-level proctor


Alex-Jordan:
In 2.18, I have a proctored test with a provided password. While I am in the Set Detail page, if I click to save, I get a warning:

```
Error adding set-level proctor: addUser: user exists (perhaps you meant to use putUser?) at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm line 1636. 
```

I guess some pseudo user is created for this type of proctoring, and I'm seeing the natural error if i try to create that user twice.

Here is a screenshot with all the quiz settings.
<img width="1008" alt="Screen Shot 2023-07-22 at 11 46 39 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/e2828715-f345-41e3-80dc-dbd76f6a68f3">

drgrice1:
I believe this is fixed with #2147 and #2150.

PR #2147, opened 2023-07-23
Correctly create or update the restricted login proctor password

drgrice1:
There was logic error that made the `if/else` statement fall through incorrectly.  The conditions of a nested `if` were condensed into the first condition of the `if/else` statement.  This puts it back to the way it was with a nested `if`.

This fixes issue #2146.


Alex-Jordan (approved):
Thanks for figuring this out. This change works for me.

Alex-Jordan:
Maybe since it is so close after the release and this is a bug, we go ahead and merge rather than wait for a third approval? Or I can tag @pstaabp to maybe give it a quick look.

drgrice1:
Yeah, that should be fine.

somiaj:
Do we another PR for develop for this then?

drgrice1:
@somiaj: I have it ready. I will add it later.

Issue #2148, opened 2023-07-24
Error while grading test


Alex-Jordan:
In 2.18, I made a test. May not be relevant, but it uses a proctoring password. The test has 33 exercises, 6 per page. There are 10 total tikz images (just mentioning because they take time to load). There are three essay answer fields (is that allowed in a test?)

I made a fake student account and took the test. After answering everything, I clicked to Grade. I clicked to confirm. Then the page hung for a long time. I figured that must be normal, since it must be rendering all the problems including those tikz images, just for the answer checking. But then an error screen came (contents pasted below at the end). 

Still as the student, I returned to the test. I was prompted to again enter the password to return to the test. That seemed wrong because I had submitted it and it is configured to only allow one graded submission. But the error message might indicate some process wasn't concluded. Anyway I entered the password and it hung for a long time. Too long. I got impatient and clicked to go back to the test's landing page. This time I saw this:

<img width="935" alt="Screen Shot 2023-07-24 at 1 52 42 AM" src="https://github.com/openwebwork/webwork2/assets/4732672/64d3fa67-28f8-416e-b0d6-870777b994e1">

It is "completed" but "not yet submitted". Something is off there.

Lastly, I thought I was answering it all correctly, except of course for the essay questions that would need manual grading. So I was surprised to see my score is 138/200. Looking more closely, starting at number 25, I have all 0s. So it seems that scoring was interrupted starting there. That problem uses a tikz image. I clicked to view it and it hung, eventually giving me the error message `malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 (before "(end of string)") at /usr/local/share/perl5/Mojo/JSON.pm line 31. ` But I am able to click to edit the problem and it loads fine in the editor. Even with the same seed as was used in the test.

Here is the error message:

```
WeBWorK error
An error occured while processing your request.

For help, please send mail to this site's webmaster webwork-group@pcc.edu, including all of the following information as well as what what you were doing when the error occured.

Error record identifier
39f5457c-1988-5a39-a881-60d1dd21acfd::1285033a-29ff-11ee-8630-9afc043bea0e

Error messages
Can't use an undefined value as a HASH reference at /opt/webwork/webwork2/lib/WeBWorK/Utils/ProblemProcessing.pm line 328.

Context
323: 	my @answerTypes;
324: 	my %answers_to_store;
325: 	my @past_answers_order;
326: 	my @last_answer_order;
327: 
[...]

Issue #2041, opened 2023-06-08
Test is completed but not submitted.


somiaj:
Just had a student complete a test, but the grade test submit only partially worked.  I can see their grade as if it were graded, but it isn't considered submitted. Here are some screenshots.

First the test launch screen:

![issue-test1](https://github.com/openwebwork/webwork2/assets/18082412/8354c0da-f18e-4635-8c2e-85f581e01699)

Test page

![issue-test2](https://github.com/openwebwork/webwork2/assets/18082412/8c052668-1b93-47c6-b161-fc20131e87ac)

Student Progress Page

![issue-test3](https://github.com/openwebwork/webwork2/assets/18082412/108470b0-2b39-4bb4-8c28-5dc6bf138817)

As shown, the test was graded and has a score (and the test only allowed a single submit).  The student was not able to get back into to the test (I had them check), since it was graded, but yet the test still isn't considered 'submitted'.

edit: Added labels to the screenshots.

somiaj:
Just to add some more info, this student's laptop cannot connect to the universitie's wifi for some reason, so instead they sit in the corner of the room near the window which they have 1 bar of phone service, and use their phone as a hot spot.  So could it be some lag in the grade test submit making it so the test doesn't fully finish being graded?

drgrice1:
This may be some odd issue with internet connectivity and some database write that was terminated prematurely.  If you can reliably reproduce this, then we have something to work with.  Otherwise, it is unlikely that anything can be done about this.  However, even if this can be reproduced reliably, there may not be anything that can be done.  Most likely we would need to implement transactions that can cancelled and rolled back when errors occur to fix this type of thing.

You might look closely at the database and see if there is anything odd for this student and set.

somiaj:
It appears that the column `version_last_attempt_time` from the versioned `CourseID_set_user` table never got filled in (it is zero).  Since this time never got recorded, the test is considered not completed.  Appears that all of the grade data from the problems was recorded, which is why there was a grade.  Looking at the `CourseID_problem_user` table for this test, I see that all the problems were correctly flagged as `attempted`.  So seems we might need to double check what would cause `version_last_attempt_time` from being recorded.

Alex-Jordan:
Are you able to see if the student has nonzero scores up to a point, but then starting with some problem it's all zero scores? I'm wondering if there are more similarities to #2148 and if something about certain problems can be in play.

drgrice1:
[...]

PR #2153, opened 2023-07-27
Switch the CourseEnvironment and AchievementEvaluator back to WWSafe.

drgrice1:
It turns out that using the default Safe is causing problems still. When a problem fails to render within the (hardcoded default) 60 second `$WeBWorK::PG::TIMEOUT`, then the alarm signal is not caught by the `$SIG{ALRM}` handler that is set in `WeBWorK::PG`.  As a result, instead of getting the information about the problem failing to render within the timeout, you instead get a useless message about decode_json failing in Mojo::JSON.

I suspect this is also the cause of issues #2041 and #2148 (now closed). Although the cause of those issues is really an error in the problems, this was hiding the error in the problems (combined with the issue fixed in pull request https://github.com/openwebwork/pg/pull/883 for issue #2041).

### DIFF
--- a/lib/WeBWorK/AchievementEvaluator.pm
+++ b/lib/WeBWorK/AchievementEvaluator.pm
@@ -26,7 +26,7 @@ use DateTime;
 
 use WeBWorK::Utils qw(sortAchievements nfreeze_base64 thaw_base64);
 use WeBWorK::Utils::Tags;
-use Safe;
+use WeBWorK::WWSafe;
 
 our @EXPORT_OK = qw(checkForAchievements);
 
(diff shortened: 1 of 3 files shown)


drgrice1:
@Alex-Jordan:  If I submit your test from #2148 it now comes back with the error
```
Deep recursion on subroutine "AnswerEvaluator::evaluate" at line 260 of [PG]/macros/contexts/contextFiniteSolutionSets.pl.
Timeout after processing this problem for 60 seconds. Check for infinite loops in problem source.
```
after 60 seconds.

drgrice1:
Unfortunately, this is not this easy.  This breaks all of the scripts.  They now need to know where the pg/lib is.  I think webwork2 will need its own WWSafe.   That is really the only viable solution.

drgrice1:
Okay.  This is ready now.  This now adds webwork2's own version of WWSafe (actually WeBWorK::WWSafe).  This avoids the dependency on PG for scripts, and the namespace doesn't conflict with PG's WWSafe.

drgrice1:
By the way @Alex-Jordan, even though submitting your exam comes back with an error with this pull request, the rest of the problems on the exam are successfully graded and answers saved.  Even those problems after the one that causes the issue.  So this pull request really does fix things.

Alex-Jordan:
Thanks for figuring this out. I will test soon juts to confirm, but calling it a night now. (And I will pursue the contextFiniteSets.pl bug.)

drgrice1:
After investigating this further I see that this is not the best fix for this issue.  The real problem is not using the system Safe module, but the dependency of the `PG::Environment` on the `WeBWorK::CourseEnvironment`.  If that dependency were removed, then webwork2 could use the system Safe module with no problem.  Fixing that will take more work, and so this is probably the way we will need to go for now.

pstaabp (approved):
I think this works now.  I created a problem that takes a few seconds to process.  
* Under develop, I get the error `malformed JSON string,...`
* Under this branch, I see the infinite loop error.

Just want to verify this is correct.

drgrice1:
Yeah, that is what should happen.

drdrew42 (approved):
👍🏻

PR #2154, opened 2023-07-27
Switch the CourseEnvironment and AchievementEvaluator back to WWSafe (for develop).

drgrice1:
It turns out that using the default Safe is causing problems still. When a problem fails to render within the (hardcoded default) 60 second `$WeBWorK::PG::TIMEOUT`, then the alarm signal is not caught by the `$SIG{ALRM}` handler that is set in `WeBWorK::PG`.  As a result, instead of getting the information about the problem failing to render within the timeout, you instead get a useless message about decode_json failing in Mojo::JSON.

I suspect this is also the cause of issues #2041 and #2148 (now closed). Although the cause of those issues is really an error in the problems, this was hiding the error in the problems (combined with the issue fixed in pull request https://github.com/openwebwork/pg/pull/883 for issue #2041).

### DIFF
--- a/lib/WeBWorK/AchievementEvaluator.pm
+++ b/lib/WeBWorK/AchievementEvaluator.pm
@@ -26,7 +26,7 @@ use DateTime;
 
 use WeBWorK::Utils qw(sortAchievements nfreeze_base64 thaw_base64);
 use WeBWorK::Utils::Tags;
-use Safe;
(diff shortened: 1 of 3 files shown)


pstaabp (approved):

drdrew42 (approved):
👍🏻

Issue #2157, opened 2023-08-04
Accessibility observations


Alex-Jordan:
A user has reported some accessibility observations. These were made with 2.17, but I think the same issues are in 2.18.

* when tabbing through the sidebar nav, when focus is on the current route, the outline indicating focus is not a clear indicator that this is where focus is. The outline color is too close to the background color. I am unsure what is a nice way to address this. Possibly change the outline color, or possibly change something about the background color or the text color.

* after submitting an answer, the input field has a glow about it, green or red. This is the case with MathQuill or not. This glow is not a sufficient indicator for users who do not perceive contrast or color well. (The results table gives more information, but that is not helpful for this issue.) They suggested some kind of icon adjacent to the input field. FWIW, PTX inserts a background image (⚠ or ✓) into the input field along with some padding so that does not overlap their answer. Again I'm unsure what wee would like to do to address this best.

drgrice1:
The first point is an issue with changing the background color of some of the size nav.  When focus is on a link in the instructor tools portion the focus is considerably clearer.

One solution is to make the background color lighter.  That background color is not critical, and only differentiates between links that are instructor links and those that are not.

Another solution is to dispense with the background color entirely, and use a separator instead between instructor links and the rest.

drgrice1:
By the way, I don't believe the first point was an issue for 2.17 since it did not have the colored background.

Also, on the second point.  I don't think there will be a solution for that until we eliminate the results table.

Alex-Jordan:
This came from a school testing with the Runestone 2.17 server. When you are, for example, on the Grades page, there is a colored background for Grades. This is less about the coloring from the "student" section of the TOC. It is more about the color being applied to the current route, and if you tab focus to that item.

Alex-Jordan:
Here are 2.17 and 2.18 screenshots. In both cases, we are on the Grades page (giving the dark background). But also I've tabbed focus to "Grades" and you can't really tell. This is their point.

<img width="350" alt="Screen Shot 2023-08-04 at 2 30 03 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/0edb565e-6c64-4fb1-a2df-bc29783ef6b3">

<img width="261" alt="Screen Shot 2023-08-04 at 2 29 20 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/aeb49749-ca84-4c67-9bf9-d4a8ea5340ef">

drgrice1:
[...]

PR #2159, opened 2023-08-05
Tweak the site navigation link colors.

drgrice1:
This makes the background color lighter and the foreground color dark for active links in the the site navigation.  This makes it easier to determine when the active link has focus because the outline color is darker than the new background color.

The math4-green theme needed a slightly darker background than the math4 and math4-red themes, and the the colors are the same as before for the math4-yellow theme for which this is not an issue.

This addresses the first part of issue #2157.

### DIFF
--- a/htdocs/themes/math4-green/_theme-overrides.scss
+++ b/htdocs/themes/math4-green/_theme-overrides.scss
@@ -1,0 +1,3 @@
+:root {
+	--ww-site-nav-link-active-background-color: #{lighten($primary, 50%)};
+}
--- a/htdocs/themes/math4-green/_theme-overrides.scss
+++ b/htdocs/themes/math4-green/_theme-overrides.scss
@@ -1,0 +1,3 @@
+:root {
+	--ww-site-nav-link-active-background-color: #{darken($primary, 2%)};
+}
--- a/htdocs/themes/math4-green/_theme-overrides.scss
+++ b/htdocs/themes/math4-green/_theme-overrides.scss
@@ -1,3 +1,0 @@
-:root {
-	--ww-site-nav-link-active-background-color: #{lighten($primary, 50%)};
-}
--- a/htdocs/themes/math4-red/_theme-overrides.scss
+++ b/htdocs/themes/math4-red/_theme-overrides.scss
@@ -1,0 +1,4 @@
+a:not(.btn):focus {
+        outline-color: #{lighten($link-hover-color, 26%)};
+}
+
--- a/htdocs/themes/math4-yellow/_theme-overrides.scss
+++ b/htdocs/themes/math4-yellow/_theme-overrides.scss
@@ -27,5 +27,4 @@ a:not(.btn):focus {
 
 :root {
 	--ww-site-nav-link-active-background-color: #{$primary};
-	--ww-site-nav-link-active-color: #{color-contrast($primary)};
 }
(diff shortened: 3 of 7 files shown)


Alex-Jordan:
Thanks for doing this. I would have done something but I'm catching up from a week on the road.

Did you do any analysis of the contrast ratios for these? If you did, I trust your results. If not, I can go ahead and do that to see. I haven't checked out the branch yet.

Is an accessibility improvement like this worthy of being a hotfix?

drgrice1:
I checked that the contrast ratios satisfy the accessibility checker in Firefox.

I am fine with this being a hotfix if that is desired.

Alex-Jordan:
We have a problem that Firefox's checker is not picking up on. (I think it is a problem without this change, but then it is a worse problem with this change.) For example in math4 (blue), now the non-active background color for one of the student nav items is #ccd6e7. And the active color is #a2c5ff. Their contrast ratio is 1.20, below the 3 that it is supposed to be for adjacent colors.

Similarly the ratio between the student nav items background color and the other item background colors has too low of a ratio. I should have caught that when I removed the inner list for instructor tools, but I assumed bootstrap's color scheme would just work. Then again, here the lack of contrast is less important. There is not much important meaning for the different background colors for student versus instructor items.

Perhaps what we need is what you suggested. Remove the background color for the student nav items and use some kind of separator. I'm thinking a top border on "Instructor Tools". We still can't use the new active color you have here, because its ratio with the regular background color is only 1.66. The active background needs to be darker to get up to 3:1. And then getting back to what was reported, we need a better indicator for when the active item has focus or hover. 

Given all this, if you agree, I can start a new branch. It seems different enough that it doesn't help to start from this branch.

drgrice1:
Sure, if you want to do it.

drgrice1:
I would not use a border. An <hr> tag would be better, with the appropriate styles it will look much better.

drgrice1:
It seems that this branch is still needed.  Even if you remove the background color, the active link color is still to dark, and you can't distinguish between the focus outline and the background.

Alex-Jordan:
I opened a PR onto your branch for inspection. It makes several changes.

drgrice1:
This now removes the background color on the student sections, and tweaks the outline colors instead of the background colors.  Thanks to @Alex-Jordan.

drgrice1:
I added a couple of additional tweaks.  Instead of using the `#site-links` selector, use the `#site-navigation` selector sot hat siblings also get the new outline.  Also make the focused sibling `.nav-link` have the elevated z-index so that if the outline is not under the active link background if that happens to be the next item.

Alex-Jordan (approved):
Tested with most recent changes and it looks good.

pstaabp (approved):
Looks good.  Didn't do thorough contrast analysis, though.

Alex-Jordan:
Somewhere it may be worth recording the contrast measurements I made.

All themes: nav-links background: `#f8f9fa`

Blue: active link background: `#003388`, focus outline: `#3f80ed`

Green: active link background: `#2e4932`, focus outline: `#749879`

Red: active link background: `#880000`, focus outline: `#ff5252`

Yellow: active link background: `#ffc700`, focus outline: `#be5050`

Blue bg:abg 10.89; bg:fo 3.6; abg:fo 3.01

Green bg:abg 9.41; bg:fo 3.07; abg:fo 3.06

Red bg:abg 9.73; bg:fo 3.02; abg:fo 3.21

Yellow bg:abg 1.48; bg:fo 3.01; abg:fo 4.47

drgrice1:
Yes, it would be good to keep track of those.

PR #2160, opened 2023-08-08
use string if it's a string

Alex-Jordan:
I think this fixes the issue reported by @somiaj at #2117.

Note that with testing, you may want to manually open up `simple.conf` to see what's in there after each test, and clear it for a new test.

Once this is good for `develop` we could do a hotfix.

### DIFF
--- a/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
+++ b/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
@@ -25,7 +25,11 @@ sub display_value ($self, $val) {
 # r->param() returns an array, so a custom version of convert_newval_source is needed.
 sub convert_newval_source ($self, $use_current) {
 	if ($use_current) {
-		return @{ $self->get_value($self->{c}->ce) };
+		if (ref $self->get_value($self->{c}->ce) eq 'ARRAY') {
+			return @{ $self->get_value($self->{c}->ce) };
+		} else {
+			return $self->get_value($self->{c}->ce);
+		}
 	} else {
 		return $self->{c}->param($self->{name});
 	}
--- a/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
+++ b/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
@@ -28,7 +28,7 @@ sub convert_newval_source ($self, $use_current) {
 		if (ref $self->get_value($self->{c}->ce) eq 'ARRAY') {
 			return @{ $self->get_value($self->{c}->ce) };
 		} else {
-			return $self->get_value($self->{c}->ce);
+			return @{ role_and_above($self->{c}->ce->{userRoles}, $self->get_value($self->{c}->ce)) };
 		}
 	} else {
 		return $self->{c}->param($self->{name});
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -2075,20 +2075,19 @@ $ConfigValues = [
 		},
 		{
 			var  => 'permissionLevels{receive_feedback}',
-			doc  => x('E-mail feedback from students automatically sent to this permission level and higher'),
+			doc  => x('Permission levels for receiving feedback email'),
 			doc2 => x(
-				'Users with these permission levels will automatically be sent feedback emails from students when they '
-					. 'use the "Contact instructor" button on any problem page.'
+				'Users with these permission levels will be sent feedback emails from students when they use the '
+					. 'feedback button.'
 			),
 			type => 'permission_checkboxlist',
 		},
 		{
 			var  => 'mail{feedbackRecipients}',
-			doc  => x('Additional addresses for receiving feedback e-mail'),
+			doc  => x('Additional addresses for receiving feedback email'),
 			doc2 => x(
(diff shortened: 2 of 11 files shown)


drgrice1 (approved):
This looks okay.

There is some oddity in the behavior though.  If simple.conf contains `$permissionLevels{receive_feedback} = 'nobody';`, and you save from any config tab except the E-Mail tab, then the string is converted to an array containing that string, so that file contains `$permissionLevels{receive_feedback} = ['nobody'];`.  If you instead save from the E-Mail tab then the current value is essentially dropped, and the emtpy array  is saved, so that the file contains `$permissionLevels{receive_feedback} = [];`.  This should be okay as these values work as expected.  It is just a bit odd.

In looking closely at the `fetchEmailRecipients` method of the `WeBWorK::Utils` module.  This is my fault, and I can put in a separate pull request to fix this.  I moved the addition of the contents of the `$ce->{mail}{feedbackRecipients}` array into the `fetchEmailRecipients` method.  However, that method may return before that is added.  Those recipients should always be added to the list.

somiaj:
This issue fixes things in that the course configuration page no longer errors out if someone has an old course with `$permissionLevels{receive_feedback}` set to a string in `simple.conf`.

I also noticed a difference between saving on the email tab vs other tabs. For instance, if `simple.conf` has `$permissionLevels{receive_feedback} = 'professor'`. If saving on the email tab this is correctly updated to `$permissionLevels{receive_feedback} = ['professor', 'admin']`, but if saved on any other tab this is saved to `$permissionLevels{receive_feedback} = ['professor']`.

I am also a little cautious with this fix, since it seems to be a fix just to ensure that the course configuration page doesn't error out if there is a string vs array issue (which is probably good to have). But it seems that there is already logic to turn `$permissionLevels{receive_feedback}` from a string to an array which is applied correctly on the email tab. I wonder if this logic could also just be used on other tabs (or maybe this change could happen in the course environment so when `get_value` is run the correct array is returned), so that there is a smooth transition from the old use of this setting to the new.

One last thing, since we are fixing things, can we fix the help for this setting on the course configuration page. The help still describes the old meaning of this setting, `E-mail feedback from students automatically sent to this permission level and higher`.  The `and higher` is no longer accurate and should also be updated.

pstaabp:
I agree with @somiaj about this changing the legacy feedback.   Shouldn't be too hard to adapt the existing code to handle the change on other configuration tabs. 

Also, I think it would be a good idea to update the help page too.

lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
```diff
@@ -25,7 +25,11 @@ sub display_value ($self, $val) {
 # r->param() returns an array, so a custom version of convert_newval_source is needed.
 sub convert_newval_source ($self, $use_current) {
 	if ($use_current) {
-		return @{ $self->get_value($self->{c}->ce) };
+		if (ref $self->get_value($self->{c}->ce) eq 'ARRAY') {
+			return @{ $self->get_value($self->{c}->ce) };
+		} else {
+			return $self->get_value($self->{c}->ce);
```
drgrice1, line 31: If you change this line to
```suggestion
			return @{ role_and_above($self->{c}->ce->{userRoles}, $self->get_value($self->{c}->ce)) };
```
it should fix the issue that @somiaj mentioned.  If the simple.conf contains `$permissionLevels{receive_feedback} = 'professor';`, then this will save `$permissionLevels{receive_feedback} = ['professor','nobody','admin'];` if "Save" is clicked on a tab other than the E-Mail tab.  If simple.conf contains `$permissionLevels{receive_feedback} = 'nobody';` then this will save `$permissionLevels{receive_feedback} = ['nobody'];`.  Those should be the equivalent of the previous setting.

somiaj: @drgrice1 Wouldn't this code also be run on the checkbox lists on other configuration tabs which are not permission levels?

somiaj: Oh wait, I guess this check is only for permissions, not general checkboxlist (missed that).

Alex-Jordan:
The `doc2` on this includes `'use the "Contact instructor" button on any problem page.'`

Should I change that to `"Email Instructor"`?  Should I try to change it to use  the config variable `$feedback_button_name`? Not sure how easy it is to use config var values within the `defaults.config` help content. I don't see a precedent. Also it may complicate things that this has to go through maketext.

drgrice1:
That would fail to work with maketext.  I think that the `doc2` should not specifically refer to the button text at all.  That is if that is possible.

Alex-Jordan:
I updated the help language.

As part of that, I changed `e-mail` to `email`. So then I felt compelled to do an additional commit to change all other `e-mail` to `email`.

somiaj:
@Alex-Jordan `e-mail` -> `email` might be a bit too much for a hotfix, since it will require lots of strings to be retranslated. Maybe don't include that commit in the hotfix at the end.

Alex-Jordan:
I thought the same. That is why I made it in a separate commit.

pstaabp (approved):
Looks good now.

PR #2163, opened 2023-08-08
Add zip to archives in File Manager

pstaabp:
This adds the ability to use zip/unzip for archiving/unarchiving in the File Manager.  

Those on both Mac and PCs have zip/unzip ability more than a tar.gz file, so will increase the access to do handle archives. 

This currently has the ability to unarchive zip files in the File Manager, but not yet archive them.  I've been wondering about the UI to do this.  Some of my thoughts are
1. a course configuration to toggle between zip and tar.gz files.  Then "Make Archive" will use the desired method. 
2. Have a bootstrap dropdown button https://getbootstrap.com/docs/5.2/components/dropdowns/ that toggles between the two methods.  
3. Replace "Make Archive" with two buttons, "Make ZIP archive" and "Make tar.gz archive.  (not a fan of this)

### DIFF
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -42,14 +42,6 @@
 		}
 	};
 
-	for (const archiveTypeInput of document.querySelectorAll('input[name="archive_type"]')) {
-		archiveTypeInput.addEventListener('click', () => {
-			document.getElementById('filename_suffix').innerText = `.${
-				document.querySelector('input[name="archive_type"]:checked').value
-			}`;
-		});
-	}
-
 	files?.addEventListener('change', checkFiles);
 	if (files) checkFiles();
 
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -45,6 +45,19 @@
 	files?.addEventListener('change', checkFiles);
 	if (files) checkFiles();
 
+	const archiveFilenameInput = document.getElementById('archive-filename');
+	const archiveTypeSelect = document.getElementById('archive-type');
+	if (archiveFilenameInput && archiveTypeSelect) {
+		archiveTypeSelect.addEventListener('change', () => {
+			if (archiveTypeSelect.value) {
+				archiveFilenameInput.value = archiveFilenameInput.value.replace(
+					/\.(zip|tgz|tar.gz)$/,
+					`.${archiveTypeSelect.value}`
+				);
+			}
+		});
+	}
+
 	const file = document.getElementById('file');
 	const uploadButton = document.getElementById('Upload');
 	const checkFile = () => (uploadButton.disabled = file.value === '');
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -35,7 +35,7 @@
 			if (
 				numSelected === 0 ||
 				numSelected > 1 ||
-				!/\.(tar|tar\.gz|tgz)$/.test(files.children[files.selectedIndex].value)
+				!/\.(tar|tar\.gz|tgz|zip)$/.test(files.children[files.selectedIndex].value)
 			)
 				archiveButton.value = archiveButton.dataset.archiveText;
 			else archiveButton.value = archiveButton.dataset.unarchiveText;
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -43,19 +43,17 @@
 	};
 
 	// Used for the archive subpage to highlight all in the Select
-	const selectAllButton = document.getElementById('select-all-files-button');
-	selectAllButton?.addEventListener('click', () => {
-		const n = document.getElementById('archive-files').options.length;
-		for (const opt of document.getElementById('archive-files').options) {
-			opt.selected = 'selected';
+	document.getElementById('select-all-files-button')?.addEventListener('click', () => {
+		for (const option of document.getElementById('archive-files').options) {
+			option.selected = 'selected';
 		}
 	});
 
-
-	for (const r of document.querySelectorAll('input[name="archive_type"]')) {
-		r.addEventListener('click', () => {
-			const suffix = document.querySelector('input[name="archive_type"]:checked').value;
-			document.getElementById('filename_suffix').innerText = '.' + suffix;
+	for (const archiveTypeInput of document.querySelectorAll('input[name="archive_type"]')) {
+		archiveTypeInput.addEventListener('click', () => {
+			document.getElementById('filename_suffix').innerText = `.${
+				document.querySelector('input[name="archive_type"]:checked').value
+			}`;
 		});
 	}
 
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -42,13 +42,6 @@
 		}
 	};
 
-	// Used for the archive subpage to highlight all in the Select
-	document.getElementById('select-all-files-button')?.addEventListener('click', () => {
-		for (const option of document.getElementById('archive-files').options) {
-			option.selected = 'selected';
-		}
-	});
-
 	for (const archiveTypeInput of document.querySelectorAll('input[name="archive_type"]')) {
 		archiveTypeInput.addEventListener('click', () => {
 			document.getElementById('filename_suffix').innerText = `.${
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -42,6 +42,23 @@
 		}
 	};
 
+	// Used for the archive subpage to highlight all in the Select
+	const selectAllButton = document.getElementById('select-all-files-button');
+	selectAllButton?.addEventListener('click', () => {
+		const n = document.getElementById('archive-files').options.length;
+		for (const opt of document.getElementById('archive-files').options) {
+			opt.selected = 'selected';
+		}
+	});
+
+
+	for (const r of document.querySelectorAll('input[name="archive_type"]')) {
+		r.addEventListener('click', () => {
+			const suffix = document.querySelector('input[name="archive_type"]:checked').value;
+			document.getElementById('filename_suffix').innerText = '.' + suffix;
+		});
+	}
+
 	files?.addEventListener('change', checkFiles);
 	if (files) checkFiles();
 
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -28,7 +28,7 @@ use File::Spec;
 use String::ShellQuote;
 use Archive::Extract;
 use Archive::Tar;
-use Archive::Zip qw(:ERROR_CODES);
+use Archive::Zip::SimpleZip qw($SimpleZipError);
 
 use WeBWorK::Utils qw(readDirectory readFile sortByName listFilesRecursive);
 use WeBWorK::Upload;
@@ -355,7 +355,7 @@ sub MakeArchive ($c) {
 		return $c->Refresh;
 	}
 
-	my $dir = "$c->{courseRoot}/$c->{pwd}";
+	my $dir = $c->{pwd} eq '.' ? $c->{courseRoot} : "$c->{courseRoot}/$c->{pwd}";
 
 	if ($c->param('confirmed')) {
 		my $action = $c->param('action')          || 'Cancel';
@@ -375,23 +375,13 @@ sub MakeArchive ($c) {
 		my ($error, $ok);
 		if ($c->param('archive_type') eq 'zip') {
 			$archive .= '.zip';
-			my $zip = Archive::Zip->new();
-			for (@files) {
-				my $fullFile = "$dir/$_";
-
-				# Skip symbolic links for now. As of yet, I have not found a perl module that can add symbolic links to
-				# zip files correctly.  Archive::Zip should be able to do this, but has permissions issues doing so.
-				next if -l $fullFile;
-
-				if (-d $fullFile) {
-					$zip->addDirectory($fullFile => $_);
-				} else {
-					$zip->addFile($fullFile => $_);
+			if (my $zip = Archive::Zip::SimpleZip->new("$dir/$archive")) {
+				for (@files) {
+					$zip->add("$dir/$_", Name => $_, storelinks => 1);
 				}
+				$ok = $zip->close;
 			}
-			$ok = $zip->writeToFileNamed("$dir/$archive") == AZ_OK;
-			# FIXME: This should check the error code, and give a more specific error message.
-			$error = 'Unable to create zip archive.' unless $ok;
+			$error = $SimpleZipError unless $ok;
 		} else {
 			$archive .= '.tgz';
 			my $tar = Archive::Tar->new;
@@ -430,11 +420,9 @@ sub UnpackArchive ($c) {
 sub unpack_archive ($c, $archive) {
 	my $dir  = "$c->{courseRoot}/$c->{pwd}";
 	my $arch = Archive::Extract->new(archive => "$dir/$archive");
-	my $ok   = $arch->extract(to => $dir);
 
-	if ($ok) {
-		my $n = scalar(@{ $arch->files });
-		$c->addgoodmessage($c->maketext('[quant,_1,file] unpacked successfully', $n));
+	if ($arch->extract(to => $dir)) {
+		$c->addgoodmessage($c->maketext('[quant,_1,file] unpacked successfully', scalar(@{ $arch->files })));
 		return 1;
 	} else {
 		$c->addbadmessage($c->maketext(q{Can't unpack "[_1]": command returned [_2]}, $archive, $arch->error));
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -27,8 +27,8 @@ use File::Copy;
 use File::Spec;
 use String::ShellQuote;
 use Archive::Extract;
-use IO::Compress::Zip qw(zip $ZipError);
 use Archive::Tar;
+use Archive::Zip qw(:ERROR_CODES);
 
 use WeBWorK::Utils qw(readDirectory readFile sortByName listFilesRecursive);
 use WeBWorK::Upload;
@@ -356,14 +356,18 @@ sub MakeArchive ($c) {
 	}
 
 	my $dir = "$c->{courseRoot}/$c->{pwd}";
-	if ($c->param('confirmed')) {
-		chdir($dir);
-		# remove any directories
-		my @files_to_compress = grep { -f $_ } @files;
 
-		unless ($c->param('archive_filename') && scalar(@files_to_compress) > 0) {
-			$c->addbadmessage($c->maketext('The filename cannot be empty.'))      unless $c->param('archive_filename');
-			$c->addbadmessage($c->maketext('At least one file must be selected')) unless scalar(@files_to_compress) > 0;
+	if ($c->param('confirmed')) {
+		my $action = $c->param('action')          || 'Cancel';
+		return $c->Refresh if $action eq 'Cancel' || $action eq $c->maketext('Cancel');
+
+		unless ($c->param('archive_filename')) {
+			$c->addbadmessage($c->maketext('The filename cannot be empty.'));
+			return $c->include('ContentGenerator/Instructor/FileManager/archive', dir => $dir, files => \@files);
+		}
+
+		unless (@files > 0) {
+			$c->addbadmessage($c->maketext('At least one file must be selected'));
 			return $c->include('ContentGenerator/Instructor/FileManager/archive', dir => $dir, files => \@files);
 		}
 
@@ -371,18 +375,39 @@ sub MakeArchive ($c) {
 		my ($error, $ok);
 		if ($c->param('archive_type') eq 'zip') {
 			$archive .= '.zip';
-			$ok    = zip \@files_to_compress => $archive;
-			$error = $ZipError unless $ok;
+			my $zip = Archive::Zip->new();
+			for (@files) {
+				my $fullFile = "$dir/$_";
+
+				# Skip symbolic links for now. As of yet, I have not found a perl module that can add symbolic links to
+				# zip files correctly.  Archive::Zip should be able to do this, but has permissions issues doing so.
+				next if -l $fullFile;
+
+				if (-d $fullFile) {
+					$zip->addDirectory($fullFile => $_);
+				} else {
+					$zip->addFile($fullFile => $_);
+				}
+			}
+			$ok = $zip->writeToFileNamed("$dir/$archive") == AZ_OK;
+			# FIXME: This should check the error code, and give a more specific error message.
+			$error = 'Unable to create zip archive.' unless $ok;
 		} else {
 			$archive .= '.tgz';
-			$ok    = Archive::Tar->create_archive($archive, COMPRESS_GZIP, @files_to_compress);
-			$error = $Archive::Tar::error unless $ok;
+			my $tar = Archive::Tar->new;
+			$tar->add_files(map {"$dir/$_"} @files);
+			# Make file names in the archive relative to the current working directory.
+			for ($tar->get_files) {
+				$tar->rename($_->full_path, $_->full_path =~ s!^$dir/!!r);
+			}
+			$ok    = $tar->write("$dir/$archive", COMPRESS_GZIP);
+			$error = $tar->error unless $ok;
 		}
 		if ($ok) {
-			my $n = scalar(@files);
-			$c->addgoodmessage($c->maketext('Archive "[_1]" created successfully ([quant,_2,file])', $archive, $n));
+			$c->addgoodmessage(
+				$c->maketext('Archive "[_1]" created successfully ([quant,_2,file])', $archive, scalar(@files)));
 		} else {
-			$c->addbadmessage($c->maketext(q{Can't create archive "[_1]": command returned [_2]}, $archive, $error));
+			$c->addbadmessage($c->maketext(q{Can't create archive "[_1]": [_2]}, $archive, $error));
 		}
 		return $c->Refresh;
 	} else {
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -371,7 +371,18 @@ sub MakeArchive ($c) {
 			return $c->include('ContentGenerator/Instructor/FileManager/archive', dir => $dir, files => \@files);
 		}
 
-		my $archive = $c->param('archive_filename') . '.' . $c->param('archive_type');
+		my $archive_type =
+			$c->param('archive_type') || ($c->param('archive_filename') =~ /\.(zip|tgz|tar.gz)$/ ? $1 : 'zip');
+
+		my $archive = $c->param('archive_filename');
+
+		# Add the correct extension to the archive filename unless it already has it. If the extension for
+		# the other archive type is given, then change it to the extension for this archive type.
+		if ($archive_type eq 'zip') {
+			$archive =~ s/(\.(tgz|tar.gz))?$/.zip/ unless $archive =~ /\.zip$/;
+		} else {
+			$archive =~ s/(\.zip)?$/.tgz/ unless $archive =~ /\.(tgz|tar.gz)$/;
+		}
 
 		if (-e "$dir/$archive" && !$c->param('overwrite')) {
 			$c->addbadmessage($c->maketext(
@@ -382,7 +393,7 @@ sub MakeArchive ($c) {
 		}
 
 		my ($error, $ok);
-		if ($c->param('archive_type') eq 'zip') {
+		if ($archive_type eq 'zip') {
 			if (my $zip = Archive::Zip::SimpleZip->new("$dir/$archive")) {
 				for (@files) {
 					$zip->add("$dir/$_", Name => $_, storelinks => 1);
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -26,6 +26,9 @@ use File::Path;
 use File::Copy;
 use File::Spec;
 use String::ShellQuote;
+use Archive::Extract;
+use IO::Compress::Zip qw(zip $ZipError);
+use Archive::Tar;
 
 use WeBWorK::Utils qw(readDirectory readFile sortByName listFilesRecursive);
 use WeBWorK::Upload;
@@ -344,24 +347,37 @@ sub Delete ($c) {
 	}
 }
 
-# Make a gzipped tar archive
+# Make a gzipped tar or zip archive
 sub MakeArchive ($c) {
 	my @files = $c->param('files');
 	if (scalar(@files) == 0) {
 		$c->addbadmessage($c->maketext('You must select at least one file for the archive'));
 		return $c->Refresh;
 	}
+	my $dir = "$c->{courseRoot}/$c->{pwd}";
+	chdir $dir;
+	my @files_to_compress;
 
-	my $dir     = "$c->{courseRoot}/$c->{pwd}";
-	my $archive = uniqueName($dir, (scalar(@files) == 1) ? $files[0] . '.tgz' : "$c->{courseName}.tgz");
-	my $tar = 'cd ' . shell_quote($dir) . " && $c->{ce}{externalPrograms}{tar} -cvzf " . shell_quote($archive, @files);
-	@files = readpipe $tar . ' 2>&1';
-	if ($? == 0) {
-		my $n = scalar(@files);
-		$c->addgoodmessage($c->maketext('Archive "[_1]" created successfully ([quant,_2,file])', $archive, $n));
+	for my $f (@files) {
+		push(@files_to_compress, glob("$f/**")) if -d $f;
+		push(@files_to_compress, $f)            if -f $f;
+	}
+
+	my ($archive, $error, $ok);
+	if ($c->param('archive_type') eq 'zip') {
+		$archive = uniqueName('', scalar(@files) == 1 ? $files[0] . '.zip' : "$c->{courseName}.zip");
+		$ok      = zip [@files_to_compress] => $archive;
+		$error   = $ZipError unless $ok;
 	} else {
-		$c->addbadmessage(
-			$c->maketext(q{Can't create archive "[_1]": command returned [_2]}, $archive, systemError($?)));
+		$archive = uniqueName('', (scalar(@files) == 1) ? $files[0] . '.tgz' : "$c->{courseName}.tgz");
+		$ok      = Archive::Tar->create_archive($archive, COMPRESS_GZIP, @files_to_compress);
+		$error   = $Archive::Tar::error unless $ok;
+	}
+	if ($ok) {
+		my $n = scalar(@files_to_compress);
+		$c->addgoodmessage($c->maketext('Archive "[_1]" created successfully ([quant, _2, file])', $archive, $n));
+	} else {
+		$c->addbadmessage($c->maketext(q{Can't create archive "[_1]": command returned [_2]}, $archive, $error));
 	}
 	return $c->Refresh;
 }
@@ -379,23 +395,16 @@ sub UnpackArchive ($c) {
 }
 
 sub unpack_archive ($c, $archive) {
-	my $dir = "$c->{courseRoot}/$c->{pwd}";
-	my @files;
-	if ($archive =~ m/\.zip$/) {
-		my $unzip = 'cd ' . shell_quote($dir) . " && $c->{ce}{externalPrograms}{unzip} -u " . shell_quote($archive);
-		@files = readpipe "$unzip";
-	} else {
-		my $z   = $archive =~ m/\.tar$/ ? '' : 'z';
-		my $tar = 'cd ' . shell_quote($dir) . " && $c->{ce}{externalPrograms}{tar} -vx${z}f " . shell_quote($archive);
-		@files = readpipe "$tar 2>&1";
-	}
+	my $dir  = "$c->{courseRoot}/$c->{pwd}";
+	my $arch = Archive::Extract->new(archive => "$dir/$archive");
+	my $ok   = $arch->extract(to => $dir);
 
-	if ($? == 0) {
-		my $n = scalar(@files);
+	if ($ok) {
+		my $n = scalar(@{ $arch->files });
 		$c->addgoodmessage($c->maketext('[quant,_1,file] unpacked successfully', $n));
 		return 1;
 	} else {
-		$c->addbadmessage($c->maketext(q{Can't unpack "[_1]": command returned [_2]}, $archive, systemError($?)));
+		$c->addbadmessage($c->maketext(q{Can't unpack "[_1]": command returned [_2]}, $archive, $arch->error));
 		return 0;
 	}
 }
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -354,32 +354,40 @@ sub MakeArchive ($c) {
 		$c->addbadmessage($c->maketext('You must select at least one file for the archive'));
 		return $c->Refresh;
 	}
+
 	my $dir = "$c->{courseRoot}/$c->{pwd}";
-	chdir $dir;
-	my @files_to_compress;
+	if ($c->param('confirmed')) {
+		chdir($dir);
+		# remove any directories
+		my @files_to_compress = grep { -f $_ } @files;
 
-	for my $f (@files) {
-		push(@files_to_compress, glob("$f/**")) if -d $f;
-		push(@files_to_compress, $f)            if -f $f;
-	}
+		unless ($c->param('archive_filename') && scalar(@files_to_compress) > 0) {
+			$c->addbadmessage($c->maketext('The filename cannot be empty.'))      unless $c->param('archive_filename');
+			$c->addbadmessage($c->maketext('At least one file must be selected')) unless scalar(@files_to_compress) > 0;
+			return $c->include('ContentGenerator/Instructor/FileManager/archive', dir => $dir, files => \@files);
+		}
 
-	my ($archive, $error, $ok);
-	if ($c->param('archive_type') eq 'zip') {
-		$archive = uniqueName('', scalar(@files) == 1 ? $files[0] . '.zip' : "$c->{courseName}.zip");
-		$ok      = zip [@files_to_compress] => $archive;
-		$error   = $ZipError unless $ok;
+		my $archive = $c->param('archive_filename');
+		my ($error, $ok);
+		if ($c->param('archive_type') eq 'zip') {
+			$archive .= '.zip';
+			$ok    = zip \@files_to_compress => $archive;
+			$error = $ZipError unless $ok;
+		} else {
+			$archive .= '.tgz';
+			$ok    = Archive::Tar->create_archive($archive, COMPRESS_GZIP, @files_to_compress);
+			$error = $Archive::Tar::error unless $ok;
+		}
+		if ($ok) {
+			my $n = scalar(@files);
+			$c->addgoodmessage($c->maketext('Archive "[_1]" created successfully ([quant,_2,file])', $archive, $n));
+		} else {
+			$c->addbadmessage($c->maketext(q{Can't create archive "[_1]": command returned [_2]}, $archive, $error));
+		}
+		return $c->Refresh;
 	} else {
-		$archive = uniqueName('', (scalar(@files) == 1) ? $files[0] . '.tgz' : "$c->{courseName}.tgz");
-		$ok      = Archive::Tar->create_archive($archive, COMPRESS_GZIP, @files_to_compress);
-		$error   = $Archive::Tar::error unless $ok;
+		return $c->include('ContentGenerator/Instructor/FileManager/archive', dir => $dir, files => \@files);
 	}
-	if ($ok) {
-		my $n = scalar(@files_to_compress);
-		$c->addgoodmessage($c->maketext('Archive "[_1]" created successfully ([quant, _2, file])', $archive, $n));
-	} else {
-		$c->addbadmessage($c->maketext(q{Can't create archive "[_1]": command returned [_2]}, $archive, $error));
-	}
-	return $c->Refresh;
 }
 
 # Unpack a gzipped tar archive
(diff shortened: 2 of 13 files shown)


drgrice1 (commented):
This looks okay for now other than the repeated definition of `$externalPrograms{unzip}`.

Eventually we should eliminate `$externalPrograms{tar}` and `$externalPrograms{unzip}` and use Perl modules for the things they are used for (we already use the zip module and there is a good module for tar as well I believe).  That would be more portable, and not have all of the possible issues with versions of these programs installed on systems that are not compatible.  This should be done with many of the other `$externalPrograms` as well (like `mv`, `cp`, `rm`, and `mkdir` which all can be handled with either built in Perl or a module).

drgrice1:
By the way, I realize this is a draft, but the note on replacing `$externalPrograms` with Perl is one that has been on my mind for a while.

drgrice1:
If you want to do that.  This is okay for now though if you want to put it off.

pstaabp:
Updated this and now has the ability to make either zip or tgz archives.  Also, removed the use of `$externalPrograms` in favor of packages.   It looks like `Archive::Tar` and `IO::Compress::Zip` are already installed on ubuntu, so didn't update the Docker files. 

A few issues:
1) I created a select to choose the archive type, but it stretches the width of the column. Changing the column width improved this, but that changes the overall layout.  Thoughts on improving this would be appreciated. 
2) I can't figure out why when making an archive, it always reports 0 files added.  On line 403 of `FileManager.pm`, including a `warn $n;` reports the right number, but the `makeText` isn't reading this right.

drgrice1:
The message that is reported when an archive is created is not on line 403 (nor 404 now).  That message would be on line 378, and that maketext call has invalid maketext syntax for the quant.  It is incorrect in develop and main (and has been since 2016).  The syntax does not allow for spaces after commas.  So `'Archive "[_1]" created successfully ([quant, _2, file])'` should be `'Archive "[_1]" created successfully ([quant,_2,file])'`.

drgrice1:
>     1. I created a select to choose the archive type, but it stretches the width of the column. Changing the column width improved this, but that changes the overall layout.  Thoughts on improving this would be appreciated.

Yeah, that isn't going to work.  That looks particularly bad.  I am not even sure that css styling can fix it.  It is probably going to have to go somewhere else.  Or most likely this is going to need to be reworked entirely to have the "Make Archive" button open a new page on which the user selects the type of archive.  In my opinion this is needed anyway, rather than just blindly creating an archive file.  The user should also have the option to select the archive file name, and a report should tell the user what files are going to be added to the archive.

drgrice1:
Also, you added the usage of the `Archive::Extract` perl module.  That is not installed by default.  It needs to be added to `check_modules.pl` and to the docker build.  The Ubuntu package is `libarchive-extract-perl`.

pstaabp:
> Also, you added the usage of the `Archive::Extract` perl module. That is not installed by default. It needs to be added to `check_modules.pl` and to the docker build. The Ubuntu package is `libarchive-extract-perl`.

Oops. I was looking at `Archive::Tar` for a while, but `Archive::Extract` handled both zip and tar.  Will update the check_modules and docker files.

pstaabp:
> > ```
>  Or most likely this is going to need to be reworked entirely to have the "Make Archive" button open a new page on which the user selects the type of archive.  In my opinion this is needed anyway, rather than just blindly creating an archive file.  The user should also have the option to select the archive file name, and a report should tell the user what files are going to be added to the archive.

We might be able to do this in a bootstrap dialog--otherwise I can create a new page.

drgrice1:
The FileManager.pm module has a mechanism for separate pages.  It uses it for confirmations, file deletion, file view and edit, and the default view is the "refresh".  So it is pretty easy to do this without creating a new module.

pstaabp:
This adds Archive::Extract to check_modules.pl and libarchive-extract-perl to docker files.

In addition, the UI now has a secondary page for creating an archive.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,62 @@
...
+					<button type="button" class="btn btn-sm btn-secondary" id="select-all-files-button">
+						<%= maketext('Select All Files') =%>
+					</button>
+				</div>
+			</div>
+		</div>
+
+
+		% my @files_to_compress;
+		% chdir($dir);
```
drgrice1, line 40: Never call `chdir` without calling `chdir` back to the original working directory.  Calling `chdir` changes the working directory for the ENTIRE process from this point on.  It does not have scope.  It is global for the process.  That can have drastic consequences.  Even better than calling `chdir` and then calling `chdir` again later to change back to the original working directory, find a way to not call `chdir` at all.  This can be done.

To see this add the line `$c->log->info(Mojo::File->new);` to line 63 of `lib/WeBWorK.pm`.  Note that `Mojo::File->new` called with no argument gives the current working directory.  Then go to the file manager page.  When you go to that page you should see `/opt/webwork/webwork2` in the terminal log (assuming that is where you started morbo from).  Now select some files and click "Make Archive".  You will again see `/opt/webwork/webwork2` in the log to the terminal.  Now click "Cancel" on the new page.  Now you will see "/opt/webwork/courses/your_course_id/templates" in the logs.  Now go to another page in webwork2 entirely.  You will still see "/opt/webwork/courses/your_course_id/templates" in the logs.  Now go log in as another user in another browser.  The logs will still show ""/opt/webwork/courses/your_course_id/templates"".  The working directory for that entire process has now been changed.

lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
```diff
@@ -344,52 +347,72 @@ sub Delete ($c) {
...
-	my $dir     = "$c->{courseRoot}/$c->{pwd}";
-	my $archive = uniqueName($dir, (scalar(@files) == 1) ? $files[0] . '.tgz' : "$c->{courseName}.tgz");
-	my $tar = 'cd ' . shell_quote($dir) . " && $c->{ce}{externalPrograms}{tar} -cvzf " . shell_quote($archive, @files);
-	@files = readpipe $tar . ' 2>&1';
-	if ($? == 0) {
-		my $n = scalar(@files);
-		$c->addgoodmessage($c->maketext('Archive "[_1]" created successfully ([quant, _2, file])', $archive, $n));
+	my $dir = "$c->{courseRoot}/$c->{pwd}";
+	if ($c->param('confirmed')) {
+		chdir($dir);
```
drgrice1, line 360: Very bad.  Very very bad.

lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
```diff
@@ -344,52 +347,72 @@ sub Delete ($c) {
...
 
-	my $dir     = "$c->{courseRoot}/$c->{pwd}";
-	my $archive = uniqueName($dir, (scalar(@files) == 1) ? $files[0] . '.tgz' : "$c->{courseName}.tgz");
-	my $tar = 'cd ' . shell_quote($dir) . " && $c->{ce}{externalPrograms}{tar} -cvzf " . shell_quote($archive, @files);
-	@files = readpipe $tar . ' 2>&1';
-	if ($? == 0) {
-		my $n = scalar(@files);
-		$c->addgoodmessage($c->maketext('Archive "[_1]" created successfully ([quant, _2, file])', $archive, $n));
+	my $dir = "$c->{courseRoot}/$c->{pwd}";
+	if ($c->param('confirmed')) {
```
drgrice1, line 359: This is always false as there is no input named 'comfirmed' on the archive page.  So archives are never made.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,62 @@
...
+		% for my $f (@$files) {
+		% 	push(@files_to_compress, glob("$f/**")) if -d $f;
+		% 	push(@files_to_compress, $f)            if -f $f;
+		% }
```
drgrice1, line 44: Make the indentation before the percent symbol.  Otherwise you end up with mixed tabs and spaces or indentation is just off.  That is how I did it in most of the template files, but there are a few that I created early in the conversion to using templates that I indented as you did here.  So don't follow those examples.

So make this
```suggestion
		% for my $f (@$files) {
			% push(@files_to_compress, glob("$f/**")) if -d $f;
			% push(@files_to_compress, $f)            if -f $f;
		% }
```

drgrice1 (changes requested):
There are some serious problems with this pull request.  The `chdir` issue is the biggest.  The fact that archives can't actually be made at all with the new page is pretty big too though.

drgrice1:
There are some other problems with this as well.  This does not deal with symlinks or directories correctly.  Your code follows symlinks treating them as directories instead.  The glob that you use for directories also does not recurse into subdirectories.

pstaabp:
Working on this now.  I think I can use `Mojo::File` to handle a number of the file related issues.  There's a method `list_tree` to get all files recursively in a directory.

drgrice1:
I have a pull request to your branch coming soon that fixes all of the issues I listed.  So you might want to wait for that instead of working on it yourself.

drgrice1:
I added a pull request to this branch with what I have for now.  There is one issue left unresolved though.  That is symbolic links in zip files.  That should work, but there is a permissions error when attempting to add them.  I don't know why yet.

somiaj:
> Should we include symlinked files in the archives?

Yes, not only is the `Library` and `Contrib` links used by default for the OPL, others may be using symlinks as well, and the archive won't be complete without them.

drgrice1:
Symlinks and directories should definitely be included.  That is just obvious.  With my pull request to this branch that is working for tar archives.  I have a way to make it work for zip as well.  The only way I have found to do this is to use the Zip::Archive::SimpleZip module.  I will add that to my pull request to this branch soon.

drgrice1:
It is now added to my pull request to this branch.

Now the other cases where Archive::Zip is used should be switched to using the same module (those are in HardcopyRenderedProblem.pm and Hardcopy.pm).  We don't need to be using two different zip modules.  I though this initially when you were using the IO::Compress::Zip module here as well.

drgrice1:
@pstaabp:  Note that I used Mojo::File and the list_tree method.  It does not follow symlinks (and should not), but does list them.

drgrice1:
Why didn't you just merge my branch?  Now this is a mess.

drgrice1:
I added another pull request to this branch that fixes everything.  Just merge it.

pstaabp:
I had worked on this yesterday before your PR and the merge didn't go well.  Thanks.

drgrice1 (approved):
This looks good now.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,64 @@
...
+		% my @files_to_compress;
+		% for my $file (@$files) {
+			% push(@files_to_compress, $file);
+			% my $path = path("$dir/$file");
+			% push(@files_to_compress, @{ $path->list_tree({ dir => 1, hidden => 1 })->map('to_rel', $dir) })
+				% if (-d $path && !-l $path);
+		% }
+		%
+		% # Select all files initially. Even those that are in previously selected directories or subdirectories.
+		% param('files', \@files_to_compress);
```
drgrice1, line 52: I didn't quite get this right.  This should be
```suggestion
		% param('files', \@files_to_compress) unless param('confirmed');
```
This way all files are selected when the user initially comes to the make archive page, but if the user is coming from the make archive page the user's selection is maintained.

drgrice1:
@pstaabp: I added another pull request to this branch with a couple of tweaks.  First, it adds an "Overwrite existing archive" checkbox to the make archive page.  If that is not checked and the user attempts to create an archive that already exists, then it won't overwrite, and will show a message telling the user to check the box if an overwrite is desired.  Second it adds a default filename if a single file is selected that is the filename without extension.

pstaabp:
> Now the other cases where Archive::Zip is used should be switched to using the same module (those are in HardcopyRenderedProblem.pm and Hardcopy.pm). We don't need to be using two different zip modules. I though this initially when you were using the IO::Compress::Zip module here as well.

I've got a draft PR to remove `externalPrograms{zip}` from hardcopy.  I'll update this PR to include this zip module instead.

drgrice1:
> I've got a draft PR to remove `externalPrograms{zip}` from hardcopy. I'll update this PR to include this zip module instead.

I would prefer that you put that into a separate pull request.  Also this is something I have also already implemented.

pstaabp:
This is a separate PR.  #2172

pstaabp:
> I would prefer that you put that into a separate pull request. Also this is something I have also already implemented.

Or maybe you mean the PG side?  #2172 has removed other external programs from webwork2.

drgrice1:
I see.  I thought you meant you were going to update this pull request to change the Hardcopy.pm change.  Not update that pull request to use the module from this one.

pstaabp:
@drgrice1 I was looking at the `Archive::Zip::SimpleZip` documentation.  I think the `storelinks` option you added on line 388 of `FileManager.pm` should be `StoreLink`, unless this was something different.   I'll update if so.

drgrice1:
The documentation is incorrect.  I originally tried `StoreLink` , and it clearly wasn't working.  When I looked at the code I saw that it is `storelinks`.  Note this is not actually an option of `Archive::Zip::SimpleZip` but it is passed to the underlying libraries it uses so it took a bit of digging to find.  Also note that the options are case insensitive, so `StoreLinks` would also work.  The `s` is needed though.

drgrice1:
Note also that `Archive::Zip::SimpleZip` uses `IO::Compress::Zip` (and other modules distributed with it) under the hood.  I also did figure out how to do this using `IO::Compress::Zip` directly, but the `Archive::Zip::SimpleZip` module just makes it nice and easy, and without it you would just have to do much of what it does.

pstaabp:
Thanks for the explanation and digging.  I'll keep it `storelinks`.

drgrice1:
The option is actually in the code for the `Archive::Zip::SimpleZip` module directly.  Although the option for the `IO::Compress::Zip` module is the same.

drgrice1:
If someone else could take a look at this, that would be good.  It has my approval.

Alex-Jordan:
Trying this out. Here are some things I see. About half is just aesthetic opinion you could take or leave :)

* The File Manager help still says `Selecting a directory or set of files and clicking "Make Archive" creates a compressed tar file with the name COURSE_NAME.tgz`. So that could be updated.
* Do we want to use a `<b>` tag instead of CSS? But for that matter, should this entire (two-sentence) paragraph be bold? I think it looks fine without any styling.
* The `mx-auto` class gives the card a big left margin on my screen, and maybe it's just my taste, but it seems ill-positioned that way. Now that I look I see that that the Rename and Copy cards are the same way though. The main File Manager, the view, and edit pages are all more flush left, and maybe it's the change from one screen to the next that is hitting me. If you like it centered like this, that's fine. Just want to check it is intentional.
* Currently if you select multiple files for archiving, WW automatically uses a name for the file derived from the course name. With this, the input for the archive file name is empty. Could it populate that input with something reasonable? Maybe the current thing based on the course name is not ideal, but I might take anything over being obligated to come up with a file name very time I make an archive.
* We have the radio buttons for zip and tgz, and then also the automatic extension to the right of the filename input field. Would it be better as a select element to the right of the input field? It could have "Archive Type" as a placeholder, even with one of the types automatically selected.
* The "Select All Files" button is maybe unnecessary clutter? When you get to this page, all the files are already selected. And you probably want to archive them all since you already selected these files on the previous screen. And if you accidentally click a file and want to select them all again, you could do that in the usual ways with shift and ctrl/command, with spacebar or mouse click. Having a button for that here makes me wonder, why here and not other multiple select interfaces?

drgrice1:
>     * The File Manager help still says `Selecting a directory or set of files and clicking "Make Archive" creates a compressed tar file with the name COURSE_NAME.tgz`. So that could be updated.

I agree the help should be updated.  I missed that.

>     * Do we want to use a `<b>` tag instead of CSS? But for that matter, should this entire (two-sentence) paragraph be bold? I think it looks fine without any styling.

I think the sentence does not need to be bold.

>     * The `mx-auto` class gives the card a big left margin on my screen, and maybe it's just my taste, but it seems ill-positioned that way. Now that I look I see that that the Rename and Copy cards are the same way though. The main File Manager, the view, and edit pages are all more flush left, and maybe it's the change from one screen to the next that is hitting me. If you like it centered like this, that's fine. Just want to check it is intentional.

I think the `mx-auto` in this case does not look the best.  I think it looks okay for the "Rename" and "Copy" cards.  Those are a little different.  They are allowed to become wider (75% of the container width), and they don't have the file list content.  They are sort of a poor man's dialog (I think those pages should be made actual dialogs not new pages) which would typically be centered.   This card is restricted to a maximum of 700 px.  I added that to this pull request because it doesn't look very good for this card to become to wide.  The other cards are fine being wider.

>     * Currently if you select multiple files for archiving, WW automatically uses a name for the file derived from the course name. With this, the input for the archive file name is empty. Could it populate that input with something reasonable? Maybe the current thing based on the course name is not ideal, but I might take anything over being obligated to come up with a file name very time I make an archive.

I added a default when a single file is selected.  That is the selected filename without extension (the develop branch uses the full file name in this case). That works well when a single directory is selected.  I also wanted a general default when multiple files are selected, but don't like using the course name.

>     * We have the radio buttons for zip and tgz, and then also the automatic extension to the right of the filename input field. Would it be better as a select element to the right of the input field? It could have "Archive Type" as a placeholder, even with one of the types automatically selected.

I am not sure that I like a dropdown for this.  I kind of like @pstaabp's choice on this.  I am not sure that a dropdown would convey that the ".zip" or ".tgz" extension is appended to the filename, and should not be added into the filename input as well. I also think the placeholder will make the dropdown wider, and not look as good. Perhaps if I saw it in action I would change my mind, but my initial instinct is that this is better.

>     * The "Select All Files" button is maybe unnecessary clutter? When you get to this page, all the files are already selected. And you probably want to archive them all since you already selected these files on the previous screen. And if you accidentally click a file and want to select them all again, you could do that in the usual ways with shift and ctrl/command, with spacebar or mouse click. Having a button for that here makes me wonder, why here and not other multiple select interfaces?

I agree that the button is not needed anymore.  With @pstaabp's initial implementation not all files were selected when you came to this page.  If you selected a directory, then the files contained in it were not selected (actually with @pstaabp's initial implementation they weren't even listed on this page).  I added the code that makes it so that all files are initially selected.

pstaabp:
Thanks for the feedback @Alex-Jordan.  I should be able to update today.

pstaabp:
Just updated to include comments from above.  The second to last commit is other changes and the last commit is a switch from radio buttons to a select dropdown.  I think I prefer the radio buttons, but take a look.  It's easy enough to roll back.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,54 @@
...
+				. 'of archive and any subset of the requested files.') =%>
+		</div>
+		<div class="row">
+			<div class="col-12">
+				<div class="input-group input-group-sm mb-2">
+					<span class="input-group-text"><%= maketext('Archive Filename') %>:</span>
+					<%= text_field archive_filename => @$files == 1 ? $files->[0] =~ s/\..*$//r : 'webwork_files',
+						'aria-labelledby' => 'archive_filename', placeholder => maketext('Archive Filename'),
+						class => 'form-control', size => 30 =%>
+					<%= select_field archive_type => ['zip', 'tgz'], class => 'form-select' =%>
```
drgrice1, line 16: This does not have the placeholder as @Alex-Jordan suggested.  Although, technically it is not a placeholder.  `select`'s don't have placeholders.  I think he meant a disabled option.

Also, this really causes the issue that I mentioned that it does not convey to the user that the extension should not be added to the file name.  Particularly so because you omitted the `.` on the extensions.

To fix those issues this should be
```suggestion
					<%= select_field archive_type => [
							[ 'File Type' => '', disabled => undef ],
						   	[ '.zip' => 'zip', selected => undef ],
						   	[ '.tgz' => 'tgz' ]
						],
```

Also, this `select` is rather wide for its content.  I suggest adding `style => 'max-width: 7em'` to the `select` to make it smaller.

Although this has another problem.  A disabled option is not a valid substitute for a visible form label.  I have no suggested fix for this.  This is another reason that the radio button approach is better for this.

htdocs/js/FileManager/filemanager.js
```diff
@@ -35,13 +35,21 @@
...
 	};
 
+	for (const archiveTypeInput of document.querySelectorAll('input[name="archive_type"]')) {
+		archiveTypeInput.addEventListener('click', () => {
+			document.getElementById('filename_suffix').innerText = `.${
+				document.querySelector('input[name="archive_type"]:checked').value
+			}`;
+		});
+	}
+
```
drgrice1, line 52: This is no longer needed and should be deleted if we do switch to the `select` for the archive type.

drgrice1 (changes requested):
I see a few other things that need to be fixed.

Also, I am not sold on using a `select` instead of `radio` buttons.

Alex-Jordan:
I'm good with going back to radio buttons. Thanks for trying it out. Can't
test it all now but I think I will tomorrow.

drgrice1:
> I'm good with going back to radio buttons. Thanks for trying it out. Can't test it all now but I think I will tomorrow.

You generally know more about accessibility than I do.  Is it valid to use a default option on a select to replace a visible label on the page?  Most accessibility checkers don't recognize a default option in such a way.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
pstaabp: > Also, this select is rather wide for its content. I suggest adding style => 'max-width: 7em' to the select to make it smaller.

I was trying to figure out how to make it narrower.

pstaabp:
@drgrice1  I'll make the code suggested above after we decide on radio button v. select.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,54 @@
...
+			% push(@files_to_compress, @{ $path->list_tree({ dir => 1, hidden => 1 })->map('to_rel', $dir) })
+				% if (-d $path && !-l $path);
+		% }
+		%
+		% # Select all files initially. Even those that are in previously selected directories or subdirectories.
+		% param('files', \@files_to_compress) unless param('confirmed');
+		<%= select_field files => \@files_to_compress, id => 'archive-files', class => 'form-select mb-2',
+			size => 20, multiple => undef =%>
+		%
+		<p><%= maketext('Create the archive of the select files?') %></p>
```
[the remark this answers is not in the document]
Alex-Jordan: Or even just "Create archive of selected files?"

Alex-Jordan:
>  Is it valid to use a default option on a select to replace a visible label on the page?

I don't know, and it might violate some automated accessibility checker. But a `title` on the `select` could be all you need, not a label. Either way I feel like this is a special case with only two options (if they had dots), with the context provided by the previous element on the page making it clear. 

However I'm OK with it going back to radio buttons. To reduce clutter, could those buttons and the "Overwrite existing archive" checkbox be on the same line?

Thanks for all thee other changes.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
Alex-Jordan: > A disabled option is not a valid substitute for a visible form label. 

It's not, but a `title` attribute might be. The disabled option would just bee for the benefit of sighted users. But I'm OK with the radio buttons approach.

Alex-Jordan:
>   Is it valid to use a default option on a select to replace a visible label on the page?

I wonder though if the disabled option could be the target of an `aria-labelledby` attribute, and that could meet an automated checker's needs (if the `title` does not).

drgrice1:
A title is still not visible on the page.  I guess we can be lax with accessibility here, and a title will serve well enough.  It isn't like this is the only place in the instructor tools that we don't quite follow the rules.

To answer your question about the radio buttons being on the same line as the "Overwrite existing archive" if we go back to the radio buttons, absolutely not.  That is a layout disaster.  On narrow screens it does not work well.  In my opinion, it looks bad and adds clutter on wide screens also.  To be frank, I really don't think there is "clutter" as you are saying.

drgrice1:
> I wonder though if the disabled option could be the target of an `aria-labelledby` attribute, and that could meet an automated checker's needs (if the `title` does not).

That does work.  So that is the way to go then.  Add an id to the disabled option, and make the select `aria-labelledby` that.

drgrice1:
@pstaabp:  So make all of the changes I have suggested, and add `id="archive-type-label"` to the disabled option of the `archive_type` select and add `aria-labelledby="archive-type-label"` to the select.  By the way, also change that disabled option to be "Archive Type", not "File Type".  It is the type of archive that is being made, not just a file type.  In addition, put it in a `maketext` call.

@Alex-Jordan: Should he still add a title to the select also?

Alex-Jordan:
I guess I'd have to see it to see what you mean. And of course, with flex
box or similar setting so on a small screen they would separate. My first
impression before was the visual of so many horizontal lines for these
things was too much. Maybe clutter is not the word for that though. The
"select all" button was still there at that time and maybe that makes a
difference.

pstaabp:
I think this looks good now.  Double check that the accessibility labels are correct though.

Alex-Jordan:
I think it looks good and that `select` meets accessibility checks. This is perhaps unorthodox, but I changed the `text-align` on that input field to `right`, and it looks like:

Screen Shot 2023-08-27 at 10 43 42 PM

So it puts the filename and extension next to each other. It's not needed, but curious if you think either one is better.

However there is an issue I will comment about in the code.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,54 @@
...
+		% for my $file (@$files) {
+			% push(@files_to_compress, $file);
+			% my $path = path("$dir/$file");
+			% push(@files_to_compress, @{ $path->list_tree({ dir => 1, hidden => 1 })->map('to_rel', $dir) })
+				% if (-d $path && !-l $path);
+		% }
+		%
+		% # Select all files initially. Even those that are in previously selected directories or subdirectories.
+		% param('files', \@files_to_compress) unless param('confirmed');
+		<%= select_field files => \@files_to_compress, id => 'archive-files', class => 'form-select mb-2',
```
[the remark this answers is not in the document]
Alex-Jordan: This should be `aria-labelledby`, not `arialabelled-by`.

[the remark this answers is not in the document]
drgrice1: Thanks @Alex-Jordan.  I frequently type that one wrong. @pstaabp, please make that change.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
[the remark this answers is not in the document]
drgrice1: After looking at this closer, I realized there is still an issue here.  The answers to this question that are presented are not valid answers to this question.  Currently it reads "Create archive of the selected files?", and the buttons that are presented to answer this question are "Cancel" or "Make Archive", neither of which are valid responses to that yes/no question.  So there are two options I see to fix this.  Either change the submit button values to 'No' and 'Yes', or simply drop the question from the page.  I much prefer the latter.  Since the instructions at the top of the page make it clear as to how the page works, namely the sentence "The following files have been selected for archiving. Select the type of archive and any subset of the requested files.", there really is no need for the question.

I see that the file manager deletion confirmation sub page has the same issue.  Although on that page, I think the question is warranted, and the submit button values should be changed.  This is an issue for another pull request I think though.

drgrice1:
> I think it looks good and that `select` meets accessibility checks. This is perhaps unorthodox, but I changed the `text-align` on that input field to `right`, and it looks like:

I am okay with this idea.  It is a bit unorthodox though.

Note that this can be accomplished by adding the bootstrap css class `text-end` to the `archive_filename` text input.  Then this will be reversed for right-to-left languages (although I am not sure it should be).

Perhaps @taniwallach should weight in on this in general.  This whole setup is a bit off for a right-to-left language.  Currently what happens is it reads "zip. filename" or "tgz. filename" (yes the period is moved to the right) if the Hebrew language is selected for the course.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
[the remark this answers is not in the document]
drgrice1: @pstaabp: Also add the `dir => 'ltr'` attribute to this `select_field`.  This is done for the files select on the main file manager page due to issues with file names in left-to-right languages (since they are really not in that language).  This doesn't fix the issue for the archive file name input and archive type select though.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,59 @@
...
+		</div>
+		<div class="row">
+			<div class="col-12">
+				<div class="input-group input-group-sm mb-2">
+					<label class="input-group-text" for="archive-filename"><%= maketext('Archive Filename') %>:</label>
+					<%= text_field archive_filename => @$files == 1 ? $files->[0] =~ s/\..*$//r : 'webwork_files',
+						id => 'archive-filename', placeholder => maketext('Archive Filename'),
+						class => 'form-control', size => 30 =%>
+					<%= select_field archive_type => [
+							[ 'File Type' => '', disabled => undef, id => 'archive-type-label' ],
```
drgrice1, line 17: Please change this to be 'Archive Type'.  This is the type of the archive, not just a file type.  This is more specific.

Also add a `maketext` call to 'Archive Type'.

So please change this to
```suggestion
							[ maketext('Archive Type') => '', disabled => undef, id => 'archive-type-label' ],
```

Alex-Jordan:
Given the RTL considerations, should we just drop the part that has you
select an archive type? And then detect if user input ends in .zip, .tgz,
or if neither, it gives an error? The instructions could make it clear what
is needed.

drgrice1:
To add to that, it should also be the case that if the default option to determine the archive type from extension is selected, and the user doesn't give a recognizable extension, then a default archive type would be used (probably zip) and the extension for that added.

drgrice1:
@pstaabp: I added a pull request to this branch that does what I described in my last two comments.  I realized that this is basically the way you did it originally.  The problem with what you did originally is that you tried to put the extension into a separate `span` instead of adding the extension to the filename input value.

drgrice1:
> I guess I'd have to see it to see what you mean. And of course, with flex box or similar setting so on a small screen they would separate. My first impression before was the visual of so many horizontal lines for these things was too much. Maybe clutter is not the word for that though. The "select all" button was still there at that time and maybe that makes a difference.

I realized that putting the check with the select on the same line isn't that hard. What I have been doing that made it seem so hard is that I was using the wrong break point.  I was trying to use the medium break point, and the problem is that in the range from the medium break point up to the large break point, things often become to narrow to accommodate having them on the same line.  It is fine above the large break point though.

Note that using the `flex` display doesn't work out well because you can't control when things wrap very well, and that results in ugliness at many widths.

pstaabp:
Just merged that.  Looks good.  I think we're getting close.

templates/ContentGenerator/Instructor/FileManager/archive.html.ep
```diff
@@ -0,0 +1,64 @@
...
+			% push(@files_to_compress, $file);
+			% my $path = path("$dir/$file");
+			% push(@files_to_compress, @{ $path->list_tree({ dir => 1, hidden => 1 })->map('to_rel', $dir) })
+				% if (-d $path && !-l $path);
+		% }
+		%
+		% # Select all files initially. Even those that are in previously selected directories or subdirectories.
+		% param('files', \@files_to_compress) unless param('confirmed');
+		<%= select_field files => \@files_to_compress, id => 'archive-files', class => 'form-select mb-2',
+			'arialabelled-by' => 'files-label', size => 20, multiple => undef, dir => 'ltr' =%>
```
drgrice1, line 55: This `arialabelled-by` also needs to be `aria-labelledby`.  I had this locally in the branch you just merged, but must have forgotten to push it.

drgrice1:
There is another serious problem with this that I just found.  If you are in the course's templates directory (the default location for the file manager), select some files and click "Make Archive", and then choose a file name like "../../archive.zip", then the file "archive.zip" is saved in the webwork courses root directory.  It is outside of this course's directory.  You can also enter an archive filename of "../../admin/archive.zip" and save the file to `/opt/webwork/courses/admin`.  Or save to another course's directory.  So the `MakeArchive` method needs to check that a legal file name is specified.

drgrice1:
@pstaabp: I added another pull request to this branch that fixes the issue with being able to save to a location outside of the directory that belongs to the course.  It simply doesn't allow path components in the file name.  I also added a check that the filename contains valid characters for a filename.

pstaabp:
Looks good.  Just merged.

drgrice1 (approved):
I think this looks good now.

@Alex-Jordan: Could you take another look when you get a chance?

Alex-Jordan:
I'm afraid I found more problems with this. I made a zip file with these contents:
```
Archive:  testing.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  09-01-2023 17:06   ../../TACO.txt
        0  09-01-2023 17:06   ../PIZZA.txt
        0  09-01-2023 17:06   SUSHI.txt
---------                     -------
        0                     3 files
```
I put the file in a course's `templates/` folder. When I use the Unpack Archive button, it indeed put `PIZZA.txt` in the course root folder, and worse, put `TACO.txt` in `/opt/webwork/courses/`. I didn't try but I think I could attack other courses' folder trees this way.

I tried doing this with a `.tgz` file too, but I couldn't work out how to archive files with relative paths. Not saying it can't be done, but I couldn't work out how. If it can't be done with `.tgz`, then this issue is new to the addition of the `.zip` format.

drgrice1:
Did you try the same thing with the develop branch?

drgrice1:
Oh, wait.  You just said you tested that for a zip archive.

drgrice1:
I did test what you are saying with a gzipped tar file, and for that it does not work.  Either with the develop branch or this pull request.  `tar` refuses to unzip outside of the working directory.

drgrice1:
To create a tar file with `..` in it you use the `-P` flag for `tar`.

Alex-Jordan:
Is that because the `tar` to extract would also need `-P`?

Maybe there is an option for `unzip` that will disallow extraction to above the current folder. Or something like that that would help.

drgrice1:
Yes, in order for `tar` to allow extracting outside of the working directory you also need the `-P` flag when extracting.

I have not found something like that for `unzip`.  I also have not found a way to pass any options to the `Archive::Extract` perl module for any archive type.  So most likely we will not be able to use that module.  Instead we will need to use `Archive::Tar` to handle tar file extraction, and another module for zip archives.  Unfortunately `Archive::Zip::SimpleZip` does not handle extracting zip archives.  Only creation of zip archives.  So we may need to switch back to `Archive::Zip` which does both.  Although I couldn't get symbolic links to be correctly added to zip archives with `Archive::Zip`.  So we may need both.  Also, I can't promise that `Archive::Zip` won't have the same problem.  It may be that zip archives are a security vulnerability, and we can't allow them.  Still researching...

somiaj:
If you could figure out how to pass options, it might be possible with `[-x xfile(s)]`, which from the manual page:

> An optional list of archive members to be excluded from processing.
> Since  wildcard characters normally match (`/') directory
> separators (for exceptions see the option -W), this  option  may
> be  used  to  exclude any files that are in subdirectories.  For
> example, ``unzip foo *.[ch] -x */*'' would extract all C  source
> files  in  the  main  directory, but none in any subdirectories.
> Without the -x option, all C source  files  in  all  directories
> within the zipfile would be extracted.

I haven't tested, but would the option with `-x ../*` then exclude extracting things outside of the main directory? Though maybe you would need `-x */../*` to keep stuff from `foo/../../../location.foo` (if such a thing was possible in a zip archive).

drgrice1:
@somiaj:  Although that means you need to know what the offending files in the archive are to exclude them.

Alex-Jordan:
Another approach is we could copy the zip file to a temp folder,where the server user has write access, but has no write access to higher level folders. And then copy the results back to where the unpacking was executed.

drgrice1:
That would still take some care.  `Archive::Extract` throws an exception if it contains a file that would be extracted to a location that it doesn't have write permissions for, and then doesn't continue the extraction.  I guess we can just refuse to extract such archives though.

drgrice1:
I think it would be better to parse the archive before extraction, and only extract the files that are valid.  `Archive::Zip` supports that.  This would also allow us to prevent the `course.conf` file being overwritten by a file in an archive.

somiaj:
What version of zip are you testing? Or maybe I can't make a .zip to test correctly.

```
$ unzip -l test.zip
Archive:  test.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2023-09-01 19:40   test5
        0  2023-09-01 19:40   test6
        0  2023-09-01 19:39   ../test2
        0  2023-09-01 19:40   ../dir2/test3
        0  2023-09-01 19:40   ../dir2/test4
        0  2023-09-01 19:39   ../../test1
---------                     -------
        0                     6 files
```

But when I try to extract it, it won't extract things into `../` for me, and just gives me a warning.

```
$ unzip test.zip
Archive:  test.zip
 extracting: test5                   
 extracting: test6                   
warning:  skipped "../" path component(s) in ../test2
 extracting: test2                   
warning:  skipped "../" path component(s) in ../dir2/test3
 extracting: dir2/test3              
warning:  skipped "../" path component(s) in ../dir2/test4
 extracting: dir2/test4              
warning:  skipped "../" path component(s) in ../../test1
 extracting: test1                   
```

so the version of unzip I'm using in bookworm (6.0-28) seems to already remove `../` from any paths, but then extracts it in the cwd.

I think @drgrice1 suggestion of parsing the list and only extracting the valid files is best in the long run.

drgrice1:
We were testing this with the `Archive::Extract` module and this pull request.  We were not using `unzip` directly.

somiaj:
Ahh, I incorrectly assumed `Archive::Extract` was using `unzip` in the backend, and looking at the depends now I see that is the case. Thanks.

Alex-Jordan:
Parsing ahead of time would be nice because you could identify which files you might be overwriting and report that to the user.

drgrice1:
@pstaabp: I added yet another pull request to this branch that fixes the issue with extracting archives with `..` in the path attempting to extract outside the course directory.  It switches to using `Archive::Zip` for zip file extraction, and `Archive::Tar` for tar file extraction.  So `Archive::Extract` is not used.  It furthermore does not blindly overwrite existing course files with files from the archive.  You now need to check `Overwrite existing files silently` to extract those files from an archive.  At this point this change does not prevent overwriting `course.conf` and `simple.conf`, but it would be easy to do that with this approach.  I put that off for another pull request for now, since there are other things that need to be done with that.

pstaabp:
Merged @drgrice1 PR and also updated the error handling.  Instead of list every existing file, it lists the number and the first filename or two.  

I had an example with 30+ existing files and the previous version listed every one.

drgrice1:
> Merged @drgrice1 PR and also updated the error handling. Instead of listing every existing file, it lists the number and the first filename or two.
> 
> I had an example with 30+ existing files and the previous version listed every one.

I don't really think you should have done that.  This is an edge case that really should never happen.  Archive files should never be created that do this sort of thing.  If you are creating an archive that you are going to share with someone else in particular, this is considered extremely bad practice. If you have an archive that does this the message should show it all.  Furthermore, you did this for zip archives, but you didn't do it for tar archives.  Very inconsistent behavior.  If you extract an improperly made tar archive that has files that would be extracted outside of the working directory, you still will get each file listed.  Note that for tar archives this is handled by the `Archive::Tar` module.

lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
```diff
@@ -344,54 +348,208 @@ sub Delete ($c) {
...
+					$outside_files[0]
+				));
+			} else {
+				$c->addbadmessage($c->maketext(
+					'There are [_1] files that already exist including [_2] and [_3]. '
+						. 'Check "Overwrite existing files silently" to unpack these files',
+					scalar(@outside_files),
+					$outside_files[0],
+					$outside_files[1]
+				));
```
drgrice1, line 497: Very bad idea.  The user needs to know ALL files that would be overwritten to make the determination to check the "Overwrite ..." checkbox.  This is not thought out.

drgrice1 (changes requested):
This now needs more work.  All files need to be shown.  Not a summary potentially showing at most 2.  Those can be combined into a single alert, but none of them can be omitted.  Perhaps in the extreme of a very large number this can be done, but showing only 1 or 2 does not give the user enough information.

drgrice1:
@pstaabp: And another pull request to this one.  This pull request makes it so that the first thirty files are listed for both files outside the working directory or already existing, and if there are more than that a note is made to that effect.

drgrice1:
Furthermore, each of those lists are in a single alert.

pstaabp:
Looks good.  Merged.

drgrice1 (approved):
Okay, I think this is good again, and ready for another round of testing by someone else.

somiaj:
Finally getting around to test this. I have only been partially following the discussion here, so I may have missed somethings already discussed. My initial observations (I didn't test it in a way to try to break things with malformed archives). I only think the first one needs to be addressed. The others are just my thoughts when using the interface.

When you select a single directory the file name is just directory name (Without `.zip` appended) while if you select multiple files/directories the filename becomes `webwork_files.zip` with the `.zip` extension. In both cases `by extension` is the default, and if no extension is added `.zip` is used. I would suggest making this consistent, and would slightly prefer that the name `webwork_files` is used as it is easier to add an extension than delete one (though adding `.zip` to the directory name would make this consistent).

When creating archive files in sub-directories, the archive is placed in the current working directory. This is reasonable, but it also starts to scatter archives around if people archive sub-directories. Would placing all archives in a single location like an `archives` directory of the `templates` directory be reasonable? Though this may just add other issues that are less desirable. The main ones I see is it will be easier to duplicate filenames and possible issues with extracting archives.

When unpacking archives, they are unpacked into the current directory. Would adding an option to unpack them to some other directory and a confirm page be useful? I also didn't realize (until I got an error) that `Overwrite existing files silently` also affects unpacking archives from then unpack archive button. I also didn't see mention of this in the help for the page. Maybe add something to the help about this?

You can make archives of archives, and if someone starts to put archives in various sub directories, they may accidentally create archives of archives. Would it be reasonable to exclude archives from archives? Or have an option to exclude them checked by default? The archive page could also just not select archives by default, but list them, and maybe give the user a message that archives were found and are not selected by default. Similar issue would happen with the `Archive Course` functionality as archives including archives.

When no files are selected the `Make Archive` button is greyed out. Since you have to go to another page to select what is finally archived, would having this default to archiving the current directory be reasonable? Or maybe open up the subpage with all files in the current directory listed, but not selected? Current functionality is fine by me, but I might have missed the reasoning of why it is greyed out, because to me it could just bring one to the next page even if nothing is selected.

drgrice1:
> When you select a single directory the file name is just directory name (Without `.zip` appended) while if you select multiple files/directories the filename becomes `webwork_files.zip` with the `.zip` extension. In both cases `by extension` is the default, and if no extension is added `.zip` is used. I would suggest making this consistent, and would slightly prefer that the name `webwork_files` is used as it is easier to add an extension than delete one (though adding `.zip` to the directory name would make this consistent).

The `.zip` extension should be there in all cases by default.  I will look into why it is not added for a directory.  Although it is easier to add an extension than delete one, it is not as clear to the user what is happening.

> When creating archive files in sub-directories, the archive is placed in the current working directory. This is reasonable, but it also starts to scatter archives around if people archive sub-directories. Would placing all archives in a single location like an `archives` directory of the `templates` directory be reasonable? Though this may just add other issues that are less desirable. The main ones I see is it will be easier to duplicate filenames and possible issues with extracting archives.

This is not feasible or even really realistic.  The current working directory approach is the most reasonable way to do this.  Placing the file in an `archives` directory doesn't prevent the instructor from moving it elsewhere.  An archive file can be uploaded into any location as well. We don't need another directory that is needed in the course directory structure in any case.

> When unpacking archives, they are unpacked into the current directory. Would adding an option to unpack them to some other directory and a confirm page be useful? I also didn't realize (until I got an error) that `Overwrite existing files silently` also affects unpacking archives from then unpack archive button. I also didn't see mention of this in the help for the page. Maybe add something to the help about this?

Unpacking to the current working directory is good enough for now.  It is the way it currently works, and is all that is really needed at this point.  You can add this as a feature request for future development if you like.  A note on where archives are extracted could be added to the help.  Using `Overwrite existing files silently` when unpacking archives is new with this pull request.  Previously files were blindly overwritten. A comment on this could be added to the help, but the behavior is straightforward, and instructions are given when a file in an archive would be overwritten.

> You can make archives of archives, and if someone starts to put archives in various sub directories, they may accidentally create archives of archives. Would it be reasonable to exclude archives from archives? Or have an option to exclude them checked by default? The archive page could also just not select archives by default, but list them, and maybe give the user a message that archives were found and are not selected by default. Similar issue would happen with the `Archive Course` functionality as archives including archives.

Archives can be excluded by un-selecting them on the make archive page.  Note that the previous behavior was to blindly add all files in sub-directories.

> When no files are selected the `Make Archive` button is greyed out. Since you have to go to another page to select what is finally archived, would having this default to archiving the current directory be reasonable? Or maybe open up the subpage with all files in the current directory listed, but not selected? Current functionality is fine by me, but I might have missed the reasoning of why it is greyed out, because to me it could just bring one to the next page even if nothing is selected.

The current design makes more sense and is consistent with how the other buttons work on the page.  You can quite easily select all files in the current working directory if you want all files in the archive.  The point of the second page is to allow you to see what you have selected (including all files in subdirectories) and allow you to remove files (particularly from subdirectories) you don't want in the archive.  I don't see your confusion on the greying out to be frank.

drgrice1:
@pstaabp: I added another pull request that makes sure the `.zip` extension is always there by default, even if a file or directory is selected that does not have an extension to begin with.

drgrice1:
@somiaj: I want to clarify that I think some of the things you suggest are reasonable, but I think they are more aggressive than is needed for this pull request.  This pull request started merely to add the feature of zip archive extraction and creation.  I asked @pstaabp to add the new archive sub page to what he had initially, and since then several other things were requested.  I think that this pull request has gone far enough, and needs to be merged.  We can then add other features and such.  I already have some other changes, and am tired of making pull requests to this one.

Alex-Jordan:
> Would placing all archives in a single location like an archives directory of the templates directory be reasonable?

I think an issue with this would be that if I'm in one working directory and make a generic `webwork_files.zip` archive, later in the proposed central archives directory I don't really see where the archive originated. If some time passes I'm not going to know. And then this would be compounded if I were in some other working directory and made another generic `webwork_files.zip` archive that I suppose would be named `webwork_files2.zip` or something.

I will try to test this later today or tomorrow.

Alex-Jordan:
I wanted to test this, but I'm just swamped with day job stuff now. If @pstaabp, @drgrice1, and @somiaj all approve it as it is, please don't wait on me.

Issue #2164, opened 2023-08-08
contrast on progress bar


Alex-Jordan:
A school has reported an accessibility contrast issue with the progress bar:

<img width="248" alt="Screen Shot 2023-08-08 at 3 57 05 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/ca904c0f-89e4-4fe5-b6f2-a731aa3c280e">

There are five background colors here to deal with, and it's possible for each of them to be adjacent to each of the others. It's not possible for them all to have contrast ratio above 3:1 in every combination. So I think more should be done with borders/outlines to separate the blocks.

Additionally, while hovering reveals a tooltip that explains each block, I wonder if more should be done. It's not clear to me how a non-mouse-user gets the explanation of what each block is saying. There are aria labels, but it's not clear how a user would naturally get to those and read them. It seems that we presently rely on the mouse-driven tooltip and the implication of the color to convey what these blocks mean.

drgrice1:
Yes, the progress bar is not properly accessible. Reach block should probably be focusable, and the tooltips activate by focus.

drgrice1:
I want to correct you on the color count.  There are actually only four colors.  There is the background color, the correct color, the incorrect color, and the in progress color.  I think you are counting the unattempted progress as a fifth color, but that is transparent and is intended to be the same as the background color.  So it is not in reality another color.  I also think that the background/unattempted color does not necessarily need to have a 3:1 contrast ratio with the surrounding region.  That is assuming that the border on the progress bar is sufficient.

This means that theoretically it is possible to find four colors that satisfy the 3:1 ratio with each other.  I am not certain that this can be done with a green, yellow, red, and grayish color that work well for this though.

Alex-Jordan:
The screenshot I posted has five background colors. The unattempted is `#dddddd` iirc, and it's not the same as the `#f8f9fa` that is the ambient background surrounding the progress bar.

>  I also think that the background/unattempted color does not necessarily need to have a 3:1 contrast ratio with the surrounding region.

Maybe. But there is something there that carries information. Even if it is 100% unattempted. So the situation to think about is a very colorblind user who doesn't see that there is a bar of any sort there when it's all unattempted. But I agree that using borders better can remedy this. We don't need 3:1 ratios with background colors at all if strong borders work to separate the regions. Note that with the nav links, we were dealing with an outline that dynamically came and went with focus, so there the background color ratios were more important. 
 


[...]

PR #2165, opened 2023-08-09
Progress bar accessibility improvements.

drgrice1:
The progress bar elements are now focusable, and tooltips active on both hover and focus.  Note that this makes all `.set-id-tooltips` active on hover and focus.  That means that this also applies to the actual set ids as well on the problem sets page.

The box shadows on the progress bar and contained elements are now black and thicker.

When a progress bar element is focused the border is thicker and box shadow a bit lighter.  This may not be good enough by itself, but should be good enough considering that the tooltip is really what is important here.  In fact I think that visual aspects of the progress bar in general are not as important with the tooltip being active on focus.

Also make the style more specific.  The `.progress` and `.progress-bar` classes are bootstrap classes that could conceivably be used for another progress bar.  So a `.set-progress` class is added to distinquish this from other progress bars.

This is to address #2164.

### DIFF
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -97,9 +97,9 @@
 	// Homework sets editor config
 	// FIXME: These are really general purpose tooltips and not just in the homework sets editor.  So the class name
 	// should be chosen to better reflect this.
-	document.querySelectorAll('.set-id-tooltip').forEach(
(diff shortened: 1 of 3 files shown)


Alex-Jordan (approved):
Tested and I think it's good. I like the set descriptions showing up on focus now too. If it goes into production, I will ask the original reporter of the accessibility issue to reassess.

pstaabp (approved):
Looks good.

PR #2166, opened 2023-08-10
Progress bar accessibility improvements (hotfix).

drgrice1:
The progress bar elements are now focusable, and tooltips active on both hover and focus.  Note that this makes all `.set-id-tooltips` active on hover and focus.  That means that this also applies to the actual set ids as well on the problem sets page.

The box shadows on the progress bar and contained elements are now black and thicker.

When a progress bar element is focused the border is thicker and box shadow a bit lighter.  This may not be good enough by itself, but should be good enough considering that the tooltip is really what is important here.  In fact I think that visual aspects of the progress bar in general are not as important with the tooltip being active on focus.

Also make the style more specific.  The `.progress` and `.progress-bar` classes are bootstrap classes that could conceivably be used for another progress bar.  So a `.set-progress` class is added to distinquish this from other progress bars.

This is #2165 for consideration as a hotfix.

### DIFF
--- a/htdocs/themes/math4/math4.js
+++ b/htdocs/themes/math4/math4.js
@@ -97,9 +97,9 @@
 	// Homework sets editor config
 	// FIXME: These are really general purpose tooltips and not just in the homework sets editor.  So the class name
 	// should be chosen to better reflect this.
-	document.querySelectorAll('.set-id-tooltip').forEach(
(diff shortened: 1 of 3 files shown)


drgrice1:
Okay, the branch name for this pull request is misleading.  So ignore that.

PR #2170, opened 2023-08-11
Quick fix for the new hardcopyThemes dir (hotfix).

drgrice1:
This simply adds the hardcopyThemes dir to the list of "updateable" directories.  That means that the directory will be created if it does not exist when a course is upgraded.

This is a quick fix that should be sufficient for a hotfix to main to fix the issue of this directory that all courses from old releases won't have.

The plan is to do much better for the next release.  Any missing directory (other than root) should either be copied from the model course if that directory exists (as is currently done for the achievements directory) or the directory should simple be created.  Care needs to be taken on the order that this is done for the directories due to nesting.


pstaabp (approved):

Alex-Jordan:
Tested in production and it works as expected.

PR #2172, opened 2023-08-11
remove references to external programs in favor of packages

pstaabp:
This removes the use of the programs mv, cp, rm, mkdir, tar, gzip (some were no longer used).  Instead, use functions in perl modules.  

This should be tested on a few different platforms.  Especially
1. Hardcopy generation
2. Course Archiving
3. Course Unarchiving
4. Add new course

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -92,8 +92,9 @@ $admin_course_id = 'admin';
 ####################################################
 # system utilities
 ####################################################
+$externalPrograms{tar} = "/bin/tar";
 $externalPrograms{git} = "/usr/bin/git";
-$externalPrograms{tar}   = "/bin/tar";
+
 
 ####################################################
 # equation rendering/hardcopy utiltiies
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -92,7 +92,6 @@ $admin_course_id = 'admin';
 ####################################################
 # system utilities
 ####################################################
-$externalPrograms{tar}   = "/bin/tar";
 $externalPrograms{git} = "/usr/bin/git";
 
 ####################################################
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -93,6 +93,7 @@ $admin_course_id = 'admin';
 # system utilities
 ####################################################
 $externalPrograms{git} = "/usr/bin/git";
+$externalPrograms{tar}   = "/bin/tar";
 
 ####################################################
 # equation rendering/hardcopy utiltiies
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -92,13 +92,8 @@ $admin_course_id = 'admin';
 ####################################################
 # system utilities
 ####################################################
-$externalPrograms{mv}    = "/bin/mv";
-$externalPrograms{cp}    = "/bin/cp";
-$externalPrograms{rm}    = "/bin/rm";
-$externalPrograms{mkdir} = "/bin/mkdir";
 $externalPrograms{tar}   = "/bin/tar";
-$externalPrograms{gzip}  = "/bin/gzip";
-$externalPrograms{git}   = "/usr/bin/git";
+$externalPrograms{git} = "/usr/bin/git";
 
 ####################################################
 # equation rendering/hardcopy utiltiies
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -23,8 +23,6 @@ problem sets.
 
 =cut
 
-use File::Path;
-use File::Copy qw(move copy);
 use File::Temp qw/tempdir/;
 use Mojo::File;
 use String::ShellQuote;
@@ -500,8 +498,8 @@ async sub generate_hardcopy ($c, $format, $userIDsRef, $setIDsRef) {
 	my $userID   = $c->param('user');
 
 	# Create the temporary directory.  Use mkpath to ensure it exists (mkpath is pretty much `mkdir -p`).
-	my $temp_dir_parent_path = "$ce->{webworkDirs}{tmp}/$courseID/hardcopy/$userID";
-	eval { mkpath($temp_dir_parent_path) };
+	my $temp_dir_parent_path = Mojo::File->new("$ce->{webworkDirs}{tmp}/$courseID/hardcopy/$userID");
+	eval { $temp_dir_parent_path->make_path };
 	if ($@) {
 		$c->add_error("Couldn't create hardcopy directory $temp_dir_parent_path: ", $c->tag('code', $@));
 		return;
@@ -598,14 +596,14 @@ async sub generate_hardcopy ($c, $format, $userIDsRef, $setIDsRef) {
 
 	# Try to move the hardcopy file out of the temp directory.
 	my $final_file_final_path = "$temp_dir_parent_path/$final_file_name";
-	my $mv_ok                 = move($final_file_path, $final_file_final_path);
-	unless ($mv_ok) {
+	eval { Mojo::File->new($final_file_path)->move_to($final_file_final_path) };
+	if ($@) {
 		$c->add_error(
 			'Failed to move hardcopy file "',
 			$c->tag('code', $final_file_name),
 			'" from "', $c->tag('code', $temp_dir_path),
 			'" to "',   $c->tag('code', $temp_dir_parent_path),
-			'":',       $c->tag('br'), $c->tag('pre', $!)
+			'":',       $c->tag('br'), $c->tag('pre', $@)
 		);
 		$final_file_final_path = "$temp_dir_rel_path/$final_file_name";
 	}
@@ -630,11 +628,10 @@ async sub generate_hardcopy ($c, $format, $userIDsRef, $setIDsRef) {
 
 # helper function to remove temp dirs
 sub delete_temp_dir ($c, $temp_dir_path) {
-	Mojo::File->new($temp_dir_path)->remove_tree;
-
-	if ($!) {
+	eval { Mojo::File->new($temp_dir_path)->remove_tree };
+	if ($@) {
 		$c->add_error('Failed to remove temporary directory "',
-			$c->tag('code', $file, '":', $c->tag('br'), $c->tag('pre', $!)));
+			$c->tag('code', $temp_dir_path, '":', $c->tag('br'), $c->tag('pre', $@)));
 	}
 	return;
 }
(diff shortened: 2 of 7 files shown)


conf/site.conf.dist
```diff
@@ -82,12 +82,7 @@ $webwork_server_admin_email = '';
 ####################################################
 # system utilities
 ####################################################
-$externalPrograms{mv}    = "/bin/mv";
-$externalPrograms{cp}    = "/bin/cp";
-$externalPrograms{rm}    = "/bin/rm";
```
drgrice1, line 87: Note that `cp`, `mv`, and `rm` are used by pg.  So changes will need to be made to pg as well.  `cp` is actually easy.  It isn't technically  actually used.  The `copyCommand` returns the value of `$externalPrograms{cp}`, but that method isn't actually used, and can be deleted.  `mv` and `rm` are used by `LaTeXImage.pm`.

pstaabp: I was planning to get to pg as well.

drgrice1: I actually have a branch that removes the usage of `mv`, `cp`, and `rm` on the pg side.  Removing the `rm` usage is actually the trickiest.  You can't use the `remove_path` (or the legacy `rmtree`) method from the `File::Path` module.  To use that you would have to share the `File::Spec` module to the safe compartment.  That should not be done.  So I implemented a `remove_path` method in `WeBWorK::PG::IO` that only uses `unlink` and `rmdir`.

It most likely will not be possible to remove the usage of `tar` on the pg side either.  The `Archive::Tar` module also depends on the `File::Spec` module, as well as a host of others that should not be shared to the safe compartment.

drgrice1: That should have been `remove_tree` not `remove_path`.  There is no `remove_path` method.  `rmtree` is legacy and should not be used.  You should take note of that as you used that in this pull request.

pstaabp: I can probably use your code that you just committed to do the `remove_tree` to remove a directory.

drgrice1: You can just use the `File::Path` `remove_tree` method for webwork2.  It doesn't have the safe compartment issues.

drgrice1: Actually, for webwork2, it would be better to use `Mojo::File` methods.  That covers almost all of the file operations.  The only thing it doesn't cover is a recursive directory copy.  Although it isn't hard to implement a simple recursive directory copy using `Mojo::File`.  In fact I did so in #2173.

pstaabp:
This is ready for review.  Switched to `Archive::Zip::SimpleZip` like #2163 and improved the creation of tar balls to not use `chdir`. 

Definitely needs some testing in the area listed above.

Alex-Jordan:
Is it relevant that I see a direct use of `mkdir` in `lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm?`

pstaabp:
> Is it relevant that I see a direct use of `mkdir` in `lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm?`

That's the perl `mkdir`.  https://perldoc.perl.org/functions/mkdir

drgrice1:
This pull request has merge conflicts that need to be resolved.

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -335,18 +339,8 @@ sub addCourse {
...
-			my $cp_out = readpipe $cp_cmd;
-			if ($?) {
-				my $exit   = $? >> 8;
-				my $signal = $? & 127;
-				my $core   = $? & 128;
-				warn
-					"Failed to copy templates from course '$sourceCourse' with command '$cp_cmd' (exit=$exit signal=$signal core=$core): $cp_out\n";
-			}
+			my $copy_ok = dircopy("$sourceDir", $destDir);
+			warn "Failed to copy templates from course '$sourceCourse': $! " if $!;
```
drgrice1, line 343: This should be
```suggestion
			unless (dircopy("$sourceDir", $destDir))
				warn "Failed to copy templates from course '$sourceCourse': $! ";
```
There is no need for the unused `$copy_ok` variable.  This is also similar to how the `dircopy` usage is documented by the `File::Copy::Recursive` module.

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -838,15 +791,15 @@ sub archiveCourse {
...
 	my ($ce, $dump_dir) = @_;
-	my $rm_cmd = "2>&1 " . $ce->{externalPrograms}{rm} . " -rf " . shell_quote($dump_dir);
-	my $rm_out = readpipe $rm_cmd;
-	if ($?) {
-		my $exit   = $? >> 8;
-		my $signal = $? & 127;
-		my $core   = $? & 128;
-		carp
-			"Failed to remove course database dump directory '$dump_dir' with command '$rm_cmd' (exit=$exit signal=$signal core=$core): $rm_out\n";
+	rmtree($dump_dir, { error => \my $err });
```
drgrice1, line 794: Use `remove_tree` from `Mojo::File` (which uses `File::Path::remove_tree` instead of the legacy `File::Path::rmtree` method).  It is called in the same way, and you can use the `error` option with it too.  Call it with
```suggestion
	Mojo::File->new($dump_dir)->remove_tree({ error => \my $err });
```

drgrice1 (commented):
I haven't testing things yet, but I see some things that need to be fixed in the code.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -751,12 +737,16 @@ sub generate_hardcopy_tex ($c, $temp_dir_path, $final_file_basename) {
 	}
 
 	# Create a zip archive of the bundle directory
-	my $zip = Archive::Zip->new();
-	$zip->addTree($temp_dir_path);
+	my $zip_file = "$temp_dir_path/$final_file_basename.zip";
+	my $zip      = Archive::Zip::SimpleZip->new($zip_file);
```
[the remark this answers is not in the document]
pstaabp: @drgrice1  Should I switch this back to Archive::Zip or alternatively I can wrap in an eval and check for errors.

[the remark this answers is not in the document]
drgrice1: You pick.  It doesn't matter much.  The easier approach is to just leave it how it was since it was already set up and working.

pstaabp:
@drgrice1 Just fixed a bunch of these.

drgrice1 (commented):
There are still some things that aren't right.

Also, I don't really think it is worth introducing the `Archive::Extract` module when the `Archive::Tar` module covers this and is a core module.

pstaabp:
@drgrice1 Fixed these.   I've noticed in HardCopy and HardcopyRenderedProblem.pm there are a bunch of file IO that could benefit from using `Mojo::File`.  I've left these as is for now.

drgrice1:
@pstaabp:  Have you tested your latest changes?

pstaabp:
I had tested the course admin features and didn't test the hardcopy creation.  Fixed the error.

pstaabp:
@drgrice1 and others, when you get a chance, this is ready to be looked at and tested gain.

conf/site.conf.dist
```diff
@@ -82,13 +82,8 @@ $webwork_server_admin_email = '';
...
-$externalPrograms{mkdir} = "/bin/mkdir";
+$externalPrograms{git} = "/usr/bin/git";
 $externalPrograms{tar}   = "/bin/tar";
```
drgrice1, line 86: Please change this to
```suggestion
$externalPrograms{tar} = "/bin/tar";
$externalPrograms{git} = "/usr/bin/git";
```
This just puts the order back to how it was, and fixes alignment of the equality symbols.

drgrice1 (commented):
Please, just revert to using `Archive::Zip` in `HardcopyRenderedProblem.pm` and `Hardcopy.pm`.  Here are the justifications for this request:

* `Archive::Zip` is already NOT using a reference to an external program, and so this change does not fit into this pull request as titled.
* `Archive::Zip` is still needed because `Archive::Zip::SimpleZip` can not do archive extraction.
* `Archive::Zip` has a nicer API than `Archive::Zip::SimpleZip`.  Look at how much more code it took to implement the same thing in `HardcopyRenderedProblem.pm` and `Hardcopy.pm` with `Archive::Zip::SimpleZip` than with `Archive::Zip`.
* Quite frankly we wouldn't be using `Archive::Zip::SimpleZip` at all if it weren't for the bug in `Archive::Zip` with adding symbolic links to zip archives.   The zip files created in these two modules don't have symbolic links, so that is irrelevant here.

In fact, please revert all changes to `HardcopyRenderedProblem.pm` in this pull request.  Other than switching from `Archive::Zip` to `Archive::Zip::SimpleZip`, you change several cases where the `Mojo::File::path` method was called into `Mojo::File->new` calls.  You do realize that the `path` call is just a shorthand for `Mojo::File->new` don't you?  The definition of the `path` method in `Mojo::File` is `sub path { __PACKAGE__->new(@_) }`.  The `path` method could also be imported and used in the `CourseManagement.pm` file.  Do NOT do this in `Hardcopy.pm` (or any module that derives from `ContentGenerator.pm`), because that module has another `path` method that will conflict if the `path` method is imported from `Mojo::File`.

conf/site.conf.dist
pstaabp: done.

pstaabp:
reverted to `Archive::Zip` in `Hardcopy.pm` and `HardcopyRenderedProblem.pm`.

drgrice1:
> In fact, please revert all changes to `HardcopyRenderedProblem.pm` in this pull request. Other than switching from `Archive::Zip` to `Archive::Zip::SimpleZip`, you change several cases where the `Mojo::File::path` method was called into `Mojo::File->new` calls. You do realize that the `path` call is just a shorthand for `Mojo::File->new` don't you? The definition of the `path` method in `Mojo::File` is `sub path { __PACKAGE__->new(@_) }`. The `path` method could also be imported and used in the `CourseManagement.pm` file. Do NOT do this in `Hardcopy.pm` (or any module that derives from `ContentGenerator.pm`), because that module has another `path` method that will conflict if the `path` method is imported from `Mojo::File`.

This has still not been addressed.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -760,7 +739,6 @@ sub generate_hardcopy_tex ($c, $temp_dir_path, $final_file_basename) {
 		$c->add_error('Failed to create zip archive of directory "', $c->tag('code', $bundle_path), '"');
 		return "$bundle_path/$src_name";
 	}
-
```
drgrice1, line 763: Please add this empty line back as well.  Generally removing empty lines that provide good code separation and improve readability is not a good idea.

drgrice1 (approved):
Just a few minor issues to be fixed, and then this is ready to go.

drgrice1 (commented):
I just noticed that you added the usage of the module File::Copy::Recursive, but did not add it to the `check_modules.pl` script.  Make sure to add that.

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -274,15 +275,16 @@ sub addCourse {
...
+			warn "Can't create $courseDirName directory '$courseDir', since the parent directory is not writeable. "
+				. "You will have to create this directory manually.\n";
 			next;
 		}
 
 		# try to create it
-		mkdir $courseDir
-			or warn
-			"Failed to create $courseDirName directory '$courseDir': $!. You will have to create this directory manually.\n";
+		eval { Mojo::File($courseDir)->make_path };
```
Alex-Jordan, line 284: Is this an error? Should be `Mojo::File->new`?

Alex-Jordan: If so and if someone confirms how I should fix it, I can fix it in #2290.

drgrice1: Yes, that should be `eval { Mojo::File->new($courseDir)->make_path };`.  Good catch.

Alex-Jordan: OK, less of a catch though than just debugging when testing #2290.

PR #2169, opened 2023-08-10
Hotfix/receive feedback

Alex-Jordan:
Hotfix parallel to #2160, but without the "e-mail" to "email" edits.

PR #2173, opened 2023-08-11
Improvements to the course directory update method.

drgrice1:
This always creates a course directory if it is missing.  It also attempts to fix permissions if those are not correct.  It does not attempt to change ownership or the group as that will almost always fail, and will need to be done manually.

The directories that can be copied from the model course are copied if a directory is created.  This is only done if the path did not exist to begin with.

This is a much improved approach to #2169 or #2170.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -339,10 +339,6 @@ $webworkURLs{AuthorHelpURL}           ='https://webwork.maa.org/wiki/Category:Au
 # Defaults for course-specific locations (directories and URLs)
 ################################################################################
 
-# Developer note: Make sure to keep the list of directories in the updateCourseDirectories method
-# of lib/WeBWorK/Utils/CourseIntegrityCheck.pm up to date with the keys of the $courseDirs variable
-# defined below.
-
 # The root directory of the current course. (The ID of the current course is
 # available in $courseName.)
 $courseDirs{root}        = "$webworkDirs{courses}/$courseName";
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -339,6 +339,10 @@ $webworkURLs{AuthorHelpURL}           ='https://webwork.maa.org/wiki/Category:Au
 # Defaults for course-specific locations (directories and URLs)
 ################################################################################
 
+# Developer note: Make sure to keep the list of directories in the updateCourseDirectories method
+# of lib/WeBWorK/Utils/CourseIntegrityCheck.pm up to date with the keys of the $courseDirs variable
+# defined below.
+
 # The root directory of the current course. (The ID of the current course is
 # available in $courseName.)
 $courseDirs{root}        = "$webworkDirs{courses}/$courseName";
--- a/lib/WeBWorK/Utils/CourseIntegrityCheck.pm
+++ b/lib/WeBWorK/Utils/CourseIntegrityCheck.pm
@@ -342,15 +342,10 @@ sub updateCourseDirectories {
 
 	my @messages;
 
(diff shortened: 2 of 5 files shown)


pstaabp (approved):
This was bugging me this week on our production server.  Good fix.

Alex-Jordan:
This now has conflicts.

drgrice1:
The conflicts are fixed.  Thanks.

lib/WeBWorK/Utils/CourseIntegrityCheck.pm
```diff
@@ -323,85 +324,105 @@ sub checkCourseDirectories {
...
 sub updateCourseDirectories {
 	my $self = shift;
 	my $ce   = $self->{ce};
 
-	my @courseDirectories = keys %{ $ce->{courseDirs} };
+	my @messages;
 
-	#FIXME this is hardwired for the time being.
-	my %updateable_directories = (html_temp => 1, mailmerge => 1, tmpEditFileDir => 1, hardcopyThemes => 1);
+	# All $courseDirs keys should be listed here except for root.  The keys of the $courseDirs hash can not be used
```
Alex-Jordan, line 345: Did you consider sorting the keys? This is not tested, but I mean using sort like:

```
@course_dirs  = sort { $courseDirs{$b} =~ /^$courseDirs{$a}/ } (keys %courseDirs);
```

so it sorts based on if a path is a super path to another path, by assuming the super path's string is a starting substring of the sub path's string.

It would correctly sort things so that for example `aa/bb` would come before `aa/bb/cc`. It would also harmlessly sort things like `aa/bb` to come before `aa/bbb`. It would not sort things like `aa//bb` to come before `aa/bb/cc` but that seems like someone shouldn't configure that way.

drgrice1: That is a good idea.  Although I think you meant
```perl
    @course_dirs = sort { $ce->{courseDirs}{$a} =~ /^$ce->{courseDirs}{$b}/ } keys %courseDirs;
```
With your code it would be a reverse lexical sort on the directories, and so `aa/bb` would come after `aa/bb/cc`.

I have tested this with the forward sort, and I will add it to this pull request (also removing the `root` directory).

We probably don't need to worry about things like `aa//bb`.  Although that sort of thing is not unlikely if a path is built from another path and the first path already has a forward slash at the end and the built path adds another (this is very common with paths in webwork).  However, I don't think that has happened with these paths in the course environment (with default settings at least).

conf/defaults.config
```diff
@@ -339,6 +339,10 @@ $webworkURLs{AuthorHelpURL}           ='https://webwork.maa.org/wiki/Category:Au
 # Defaults for course-specific locations (directories and URLs)
 ################################################################################
 
+# Developer note: Make sure to keep the list of directories in the updateCourseDirectories method
```
Alex-Jordan, line 342: Is this comment needed now with the recent change?

drgrice1: No its not.  I removed it.

drgrice1: I don't know of a utility that checks if a path is a sub directory of another directly.  There are utilities that split a path into parts though.  So you could split the path into parts, and then sort on the parts much like sorting by last name, then first name.  If a path has more parts then another, then it would be sorted later.

somiaj: There is `path_is_subdir` in `WeBWorK/Utils.pm`, though unsure how efficient that would be.

drgrice1: @somiaj: Yeah, we just discussed that in the meeting.

PR #2183, opened 2023-08-17
Correct the version number for the modify_user method called in LTIAdvantage.pm.

drgrice1:
Clearly the LTI 1.3 code should be using the v1p3 method from authen_LIT_1_3.conf, and not the v1p1 method from authen_LTI_1_1.conf.


somiaj (approved):
Obvious error and simple fix. I say next person to approve just merges it.

Issue #2185, opened 2023-08-18
downloading templates folder and re-uploading breaks problem editor 2.18


ofloveandhate:
## Problem and how to manifest

My university's server is at 2.18, and this evening I followed a practice I used all summer with 2.17, to edit many problems at once on my computer -- much faster than one-by-one on webwork:

1. make archive of templates on webwork
2. download templates.tgz to my computer
3. unpack and edit a bunch of files
4. re-tar the folder
5. upload, and unpack-and-delete on webwork

I've successfully used this for a while now, to do things like merge content from courses, change filenames, move problems with filters, etc.  Dozens of times this summer I've walked this loop to accomplish goals practically impossible through the Webwork online interface, using Python and other tools.  So I had reason to believe it would still work this evening.  (*Please* don't tell me "you shouldn't have been doing that" -- that won't be helpful in solving this problem, and it would belittle my experience and goals)

This was my first time doing this on 2.18, and 🔥.  Now when I try to use the problem editor, I get this:

<img width="994" alt="Screenshot 2023-08-17 at 22 46 40" src="https://github.com/openwebwork/webwork2/assets/1388063/04051150-87d5-49b6-9a52-ccafc614f139">

*I think this is a bug having to do with the hardcopy template system in 2.18*

## Evidence

In the Course Configuration page, I can use the Hardcopy theme and Hardcopy theme for problem editor dropdowns, and I can see some `._` files available:

<img width="502" alt="Screenshot 2023-08-17 at 22 34 40" src="https://github.com/openwebwork/webwork2/assets/1388063/6d8cd7d6-4af2-412e-bdcb-c2bdfe9cb3b8">

## 🧐 my guess at an explanation

downloading/uploading cannot create files that start with ._    I tried to make a file named that just now, and it refused giving me a message "Your file name may not begin with a dot".  so somehow a config is pointing to this ._ file, and the download-upload loop loses it?

Something about the hardcopy selecting is picking up files it shouldn't. before i replaced the `templates` folder, the files were there? (were they? I can't see dotfiles)  now they're not, and for some reason that breaks the problem editor.

That's my hunch.

---

## My immediate ask

I'm about to launch my course for the Fall 2023 semester, and I have more work to do before it can happen.  Is there a quick fix my sysadmin could use?  Copy in those files, or just make them symlinks to the actual files?

Alex-Jordan:
[...]

PR #2186, opened 2023-08-18
only offer valid XML files for hardcopy theme

Alex-Jordan:
This is a hotfix to address #2185.

Before this PR, if there is some ".xml" file in the course's `hardcopyThemes` folder that can't be parsed into an XML tree, there is an error in the problem editor page as well as the hardcopy page.

This culls files in that folder from consideration if they do not parse successfully as XML.

It is still the case that you can go to Course Config and activate the dropdown to choose the default hardcopyTheme (either for general us or for  the editor) and you will see unwanted files from the `hardcopyThemes` folder in that list. It doesn't seem worth trying to cull this list here. And if someone chooses a file that won't actually work, thin in the problem editor page as well as the hardcopy page, no harm done I think. The dropdown list there will just default to the top option. It won't actually try to use the bad file.

There may be better ways to do this. Once this is reviewed I can make the parallel `develop` PR.

To test this, put some bad file in a course's `hardcopyThemes` folder. "Bad" just means not valid as XML, but with a `.xml` extension. Then visit the problem editor and hardcopy pages.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -445,19 +445,17 @@ sub display_form ($c) {
 		my $themeTree = XML::LibXML->load_xml(location => "$ce->{webworkDirs}{hardcopyThemes}/$hardcopyTheme");
 		$hardcopyLabels{$hardcopyTheme} = $themeTree->findvalue('/theme/@label');
 	}
-	my @hardcopyThemesCourseAll;
+	my @files;
 	if (opendir(my $dhC, $ce->{courseDirs}{hardcopyThemes})) {
-		@hardcopyThemesCourseAll = grep {/\.xml$/} sort readdir($dhC);
+		@files = grep { /\.xml$/ && !/^\./ } sort readdir($dhC);
 	}
-	my @hardcopyThemesCourse = @hardcopyThemesCourseAll;
-	for my $hardcopyTheme (@hardcopyThemesCourseAll) {
+	my @hardcopyThemesCourse;
+	for my $hardcopyTheme (@files) {
 		eval {
 			my $themeTree = XML::LibXML->load_xml(location => "$ce->{courseDirs}{hardcopyThemes}/$hardcopyTheme");
 			$hardcopyLabels{$hardcopyTheme} = $themeTree->findvalue('/theme/@label') || $hardcopyTheme;
+			push(@hardcopyThemesCourse, $hardcopyTheme);
 		};
-		if ($@) {
-			@hardcopyThemesCourse = grep { $_ ne $hardcopyTheme } @hardcopyThemesCourse;
-		}
 	}
 	my $hardcopyThemesAvailable = [
 		sort(do {
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -445,13 +445,19 @@ sub display_form ($c) {
 		my $themeTree = XML::LibXML->load_xml(location => "$ce->{webworkDirs}{hardcopyThemes}/$hardcopyTheme");
 		$hardcopyLabels{$hardcopyTheme} = $themeTree->findvalue('/theme/@label');
 	}
-	my @hardcopyThemesCourse;
+	my @hardcopyThemesCourseAll;
 	if (opendir(my $dhC, $ce->{courseDirs}{hardcopyThemes})) {
-		@hardcopyThemesCourse = grep {/\.xml$/} sort readdir($dhC);
+		@hardcopyThemesCourseAll = grep {/\.xml$/} sort readdir($dhC);
 	}
-	for my $hardcopyTheme (@hardcopyThemesCourse) {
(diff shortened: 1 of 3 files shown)


lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -445,13 +445,19 @@ sub display_form ($c) {
...
-		$hardcopyLabels{$hardcopyTheme} = $themeTree->findvalue('/theme/@label') || $hardcopyTheme;
+	my @hardcopyThemesCourse = @hardcopyThemesCourseAll;
+	for my $hardcopyTheme (@hardcopyThemesCourseAll) {
+		eval {
+			my $themeTree = XML::LibXML->load_xml(location => "$ce->{courseDirs}{hardcopyThemes}/$hardcopyTheme");
+			$hardcopyLabels{$hardcopyTheme} = $themeTree->findvalue('/theme/@label') || $hardcopyTheme;
+		};
+		if ($@) {
+			@hardcopyThemesCourse = grep { $_ ne $hardcopyTheme } @hardcopyThemesCourse;
+		}
```
drgrice1, line 460: It is highly more efficient if this is
```suggestion
	my @hardcopyThemesCourse;
	for my $hardcopyTheme (@hardcopyThemesCourseAll) {
		eval {
			my $themeTree = XML::LibXML->load_xml(location => "$ce->{courseDirs}{hardcopyThemes}/$hardcopyTheme");
			$hardcopyLabels{$hardcopyTheme} = $themeTree->findvalue('/theme/@label') || $hardcopyTheme;
			push(@hardcopyThemesCourse, $hardcopyTheme);
		};
	}
```
Instead of initially copying the `@hardcopyThemesCourseAll` array, add the valid elements to the array as they are found.  If an exception is thrown by the `load_xml` method, then none of the code after it in the `eval` will be executed, so the invalid file will not be added to the `@hardcopyThemesCourse` array.

Your code iterates through the `@hardcopyThemeseCourseAll` array, and inside that loop it iterates through the `@hardcopyThemesCourse` array each time an invalid file is found.  That latter iteration is not needed.

Also, this really should be done on the site wide hardcopy themes as well.  Really, this should be done any time `load_xml` is called.  This is the same as any time you try to open a file in general.  Even more generally, always check for failure for something that can fail, and handle failure appropriately.  One reason that this might fail for site wide themes is if file permissions aren't correct.  For example, if the `assets/hardcopyThemes` folder is not readable by the server user, then this fails.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -445,13 +445,19 @@ sub display_form ($c) {
 		my $themeTree = XML::LibXML->load_xml(location => "$ce->{webworkDirs}{hardcopyThemes}/$hardcopyTheme");
 		$hardcopyLabels{$hardcopyTheme} = $themeTree->findvalue('/theme/@label');
 	}
-	my @hardcopyThemesCourse;
+	my @hardcopyThemesCourseAll;
 	if (opendir(my $dhC, $ce->{courseDirs}{hardcopyThemes})) {
-		@hardcopyThemesCourse = grep {/\.xml$/} sort readdir($dhC);
+		@hardcopyThemesCourseAll = grep {/\.xml$/} sort readdir($dhC);
```
drgrice1, line 450: I think it would be good to change this to
```suggestion
		@hardcopyThemesCourseAll = grep { /\.xml$/ && !/^\./ } sort readdir($dhC);
```
I see no reason to include hidden files in this at all.  This alone would resolve the original issue that this pull request was created for.  Although, the `eval` around the `load_xml` call definitely is more important than this.  I really should have seen that when this was initially implemented.

This should also be done in all of the cases that these hardcopy theme xml file lists are created.  Then the hidden files won't be shown in the course config dropdowns either.  Eventually I want to replace the `readdir` calls with `Mojo::File` usage, and then we can use the `list` method of that module.  The `list` method skips hidden files by default.  It is common practice for GUI's to skip hidden files unless they are specifically needed.

drgrice1 (commented):
This generally looks good, but I suggested some improvements.

lib/WeBWorK/ContentGenerator/Hardcopy.pm
Alex-Jordan: Committed here, but I realized I need to apply it to both files. Will do locally and force push.

Alex-Jordan:
OK, I force pushed with basically the changes you suggested. And I extended thee same logic to the dropdown lists in Config. To do that, I had to add `XML::LibXML` to `Config.pm` to check for valid XML.

drgrice1 (approved):
This looks good for now.

Using `Mojo::File` won't help with my comment about protecting all `load_xml` calls, but the site hardcopyThemes folder is less likely to have issues at least.  Using `Mojo::File` just simplifies the `opendir` usage.

pstaabp (approved):
I'm on MacOS and I couldn't reproduce Finder creating ._ files, but this works in ignoring bad xml files, so hopefully this will work for #2185

pstaabp:
I approved this, however, reproducing the same code in 3 places makes me cringe a bit, although the block of code is quite small.  Maybe we can put this in a Utils file.

drgrice1:
I agree that a utility method would be good here.  But for now, this is fine.

pstaabp:
I'll go ahead and merge.

PR #2188, opened 2023-08-18
Fix some minor issues I have been seeing since release.

drgrice1:
Use `format_set_name_display` for the set id in the title when editing set headers in the PGProblemEditor.

Add a margin (class `mb-2`) to the bottom of the row containing the body in the system.html.ep template.  If there are messages on these pages, the messages are currently flush up against the end content.  This fixes that in a consistent way.  The margins that were added on some pages to handle this issue have been removed.  The problem with adding a margin to some element in the body aside from the inconsistency of handling this issue is that there can be content between the body and the bottom messages.  Page warnings can appear there.  The warnings margin has been adjusted to fit into this scheme.  Ideally everything would add a margin to bottom, and only the to most element in the page (the page title) add a margin to the top.  (Sounds like a niceTables.pl concept.)  In a few cases this may add a bit too much space between the last body element and the messages (probably because there are missed instances where special handling for this was added before), but that is better than no margin.

Fix some strings that were incorrectly interpolated for maketext in the view.html.ep template for the FileManager and in FileManager.pm itself. The first can be seen by selection a file that is neither a text or image file (like a tgz file), and clicking "View".  This will show the message `The file $file does not appear to be a text or image file.` `$file` is literally what is shown in that message.  The second can be seen by selecting some files in the file manager, and clicking "Make Archive".  The message that comes back always tells you there were 0 files in the archive, even though the correct number of files were added to the archive.

I would like this to be considered for a hotfix even though these are minor.  These issues are quite noticable (at least the margin issue).

### DIFF
--- a/htdocs/themes/math4/system.html.ep
+++ b/htdocs/themes/math4/system.html.ep
@@ -165,7 +165,7 @@
 		% if ($c->can('message')) {
 			<div id="message_bottom" class="message"><%= $c->message %></div>
 			<div>
-				<button type="button" id="dismiss-messages-btn" class="btn btn-primary d-none">
+				<button type="button" id="dismiss-messages-btn" class="btn btn-secondary d-none">
 					<%= maketext('Dismiss All Messages') %>
 				</button>
 			</div>
--- a/htdocs/themes/math4/system.html.ep
+++ b/htdocs/themes/math4/system.html.ep
@@ -146,7 +146,7 @@
 		% }
 		%
 		% # Display the page body.
-		<div class="row">
+		<div class="row mb-2">
 			<div class="body <%= $c->can('info') ? 'col-md-8' : 'col-12' %>">
 				<%= content =%>
 			</div>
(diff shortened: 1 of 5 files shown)


pstaabp (approved):
Looks good and I think a hot fix would be good.

Alex-Jordan (approved):
This is all good and yes, let's apply to `main` as well.

drgrice1:
Would anyone be opposed to changing the "Dismiss All Messages" button from `btn-primary` to `btn-secondary` (or perhaps `btn-info`)?  I often find that I get confused with other more important buttons on the page, and click this expecting the primary action to occur.  Then I move on, and the actual thing I wanted to occur doesn't get done.  This seems to happen to me often when the this button is close to the main button that does the important action for the page.  Of course, some of this is due to habit, and this button not being there before.

pstaabp:
> Would anyone be opposed to changing the "Dismiss All Messages" button from `btn-primary` to `btn-secondary` (or perhaps `btn-info`)? 

I'm good with the change.

Alex-Jordan:
Not opposed.

drgrice1:
I added that to this and #2191.

PR #2191, opened 2023-08-18
Fix some minor issues (hotfix).

drgrice1:
Use `format_set_name_display` for the set id in the title when editing set headers in the PGProblemEditor.

Add a margin (class `mb-2`) to the bottom of the row containing the body in the system.html.ep template.  If there are messages on these pages, the messages are currently flush up against the end content.  This fixes that in a consistent way.  The margins that were added on some pages to handle this issue have been removed.  The problem with adding a margin to some element in the body aside from the inconsistency of handling this issue is that there can be content between the body and the bottom messages.  Page warnings can appear there.  The warnings margin has been adjusted to fit into this scheme.  Ideally everything would add a margin to bottom, and only the to most element in the page (the page title) add a margin to the top.  (Sounds like a niceTables.pl concept.)  In a few cases this may add a bit too much space between the last body element and the messages (probably because there are missed instances where special handling for this was added before), but that is better than no margin.

Fix some strings that were incorrectly interpolated for maketext in the view.html.ep template for the FileManager and in FileManager.pm itself. The first can be seen by selection a file that is neither a text or image file (like a tgz file), and clicking "View".  This will show the message `The file $file does not appear to be a text or image file.` `$file` is literally what is shown in that message.  The second can be seen by selecting some files in the file manager, and clicking "Make Archive".  The message that comes back always tells you there were 0 files in the archive, even though the correct number of files were added to the archive.

### DIFF
--- a/htdocs/themes/math4/system.html.ep
+++ b/htdocs/themes/math4/system.html.ep
@@ -165,7 +165,7 @@
 		% if ($c->can('message')) {
 			<div id="message_bottom" class="message"><%= $c->message %></div>
 			<div>
-				<button type="button" id="dismiss-messages-btn" class="btn btn-primary d-none">
+				<button type="button" id="dismiss-messages-btn" class="btn btn-secondary d-none">
 					<%= maketext('Dismiss All Messages') %>
 				</button>
 			</div>
--- a/htdocs/themes/math4/system.html.ep
+++ b/htdocs/themes/math4/system.html.ep
@@ -146,7 +146,7 @@
 		% }
 		%
 		% # Display the page body.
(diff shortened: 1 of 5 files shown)


pstaabp (approved):

pstaabp:
Shouldn't we put in a PR to develop as well or are these in another PR?

drgrice1:
This is the hotfix version of #2188.  I used the default pull request title from the commit, instead of editing as I did for #2188.

pstaabp:
I saw that a few minutes later and tried to delete the comment.  😵‍💫

PR #2161, opened 2023-08-08
Always send feedback emails to those in the `$mail{feedbackRecipients}` list.

drgrice1:
This is how this was in previous versions of webwork2.  Restructuring of things to clean up how the email recipient list is generated resulted in this being dependent on the `$permissionLevels{receive_feedback}` setting.

To test this go to the E-Mail tab in the course configuration and uncheck all of the check boxes for the "E-mail feedback from students automatically sent to this permission level and higher" setting, and add an email address to the "Additional addresses for receiving feedback e-mail" setting.  With develop or main when you click on the feedback buttons on a homework set or problem page it will show "No e-mail recipients are listed for this course."  With this pull request the email address you add will be shown.

PR #2192, opened 2023-08-18
Always send feedback emails to those in the `$mail{feedbackRecipients}` list (hotfix).

drgrice1:
This is how this was in previous versions of webwork2.  Restructuring of things to clean up how the email recipient list is generated resulted in this being dependent on the `$permissionLevels{receive_feedback}` setting.

This is a hotfix version of #2161 in case we want that.


pstaabp (approved):
I'm good with a hot fix.

Issue #2167, opened 2023-08-10
Click Upgrade Courses twice causes an error


Alex-Jordan:
In the admin course in 2.18, if I click Upgrade Courses and then click it again while it is still processing, it comes back with a huge error message (maybe the same message repeated for each course). Of course the explanation is that the database has a lock on it while performing one check, and the second click is initiating another check.

Alex-Jordan:
This is fixed now.

PR #2193, opened 2023-08-19
Fix the database locking code in `WeBWorK::Utils::CourseIntegrityCheck`.

drgrice1:
A process should not attempt to unlock the database if it has not first locked the database.

What is currently happening is that the database integrity checker locks the database, then unlocks the database, and then when the integrity checker object is destroyed in clean up it attempts to unlock the database again (in the DESTROY method).  If there is no lock by any process when that last unlock attempt is fine.  However, if another request is sent to the server that also attempts to lock the database while the first request is still processing, then as soon as this request unlocks the database, then the second request locks the database.  Then when the first request attempts to unlock the database in clean up, it finds the lock now held by the second process and dies because it can't release the lock now held by the second process.

This now sets a flag when the database integrity checker locks the database successfully, clears that flag when the lock is successfully unlocked, and the database is only attempted to be unlocked in clean up if that flag is still set.  Furthermore, this never dies when attempting to release a lock if the release attempt fails.  There is no reason for that.  It just warns in the two cases that a release attempt can fail, and with the new flag the case of a lock owned by another process should not occur (unless someone erroneously calls the `unlock_database` method without first calling the `lock_database` method).

This fixes issue #2167.

More work needs to be done here.  The course integrity checker is really not well thought out.

The course database checker and the course directory checker need to be separated into different modules.  Directory checking and upgrading does not need a database handle.  If one wanted to check or upgrade directory structure without checking or upgrading the database, then this would acquire an unnecessary database handle.  In fact directory checking and upgrading don't need an object at all.  Those should be exported methods in another non-object package.

In addition, there should be checks for exceptions that may be thrown when the database is checked or upgraded.  There are lots of ways that can happen in addition to a failed lock attempt (which also really shouldn't die).

### DIFF
--- a/lib/WeBWorK/Utils/CourseIntegrityCheck.pm
+++ b/lib/WeBWorK/Utils/CourseIntegrityCheck.pm
@@ -79,8 +79,7 @@ sub confirm { my ($self, @args) = @_; my $sub = $self->{confirm_sub}; return &$s
 
 sub DESTROY {
 	my ($self) = @_;
-	$self->unlock_database;
-	$self->SUPER::DESTROY if $self->can("SUPER::DESTROY");
+	$self->unlock_database if $self->{db_locked};
 	return;
 }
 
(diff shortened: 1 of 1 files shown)


drgrice1:
I forgot to point out some things here.  This can not be tested with morbo.  In order for the error to occur you need multiple processes.  So hypnotoad with more than one worker is required to test this.  You also need enough courses to give you enough time to click a second time before the response is received.

On the develop (or main) branch, if you do as stated in #2167, (in the admin course click on "Upgrade Courses" in the site nav, and before that request response is received click on it again), then you will see the error `(in cleanup) Couldn't release lock because the lock is not held by this thread.` followed by the backtrace from that, and typically that will be repeated several times (one for each course integrity checker object for each course being destroyed).

Of course with this pull request you will not see any of the errors.

Also note that this is not a serious error, and won't result in any damage to the database.  What is happening is that the upgrade check is successfully completing and the database lock is successfully released, and then it dies for really no real valid reason in clean up.

pstaabp (approved):
I haven't tested this, as @drgrice1 comment that needs to be run on hypnotoad.  The code looks good to me.

drdrew42 (approved):

PR #2196, opened 2023-08-23
Remove the course_id from the past_answer table.

drgrice1:
It is time for this to go. It is an entirely unnecessary column. This is taken care of by the usual course upgrade procedure in the admin course.

Don't test this on a course that you want to go back to other branches with! It isn't so easy to add the course_id column back. It can be done though if needed.


pstaabp (approved):
Looks good.  Both tested for a given course as well as an unarchived course (needed to upgrade first).

drdrew42 (approved):

PR #2201, opened 2023-09-01
Change 'File Manager' to 'Course Files' in Library Browser

dlglin:
In the Library Browser I think the name 'File Manager' for the button to access PG files in the course templates folder is misleading.  We already have another page called 'File Manager'.  The button doesn't take you there, nor to a page where you can actually manage files.

I'm suggesting 'Course Files' as an alternative, but I'm not married to that name.  Perhaps "Files in This Course", "Problems From This Course", or "Course Folders"?  I'd even be okay with going back to "Local Problems", which is a lot less confusing now that the next button has been renamed to "Course Sets".

### DIFF
--- a/templates/ContentGenerator/Instructor/SetMaker/browse_local_panel.html.ep
+++ b/templates/ContentGenerator/Instructor/SetMaker/browse_local_panel.html.ep
@@ -6,7 +6,7 @@
 <div class="InfoPanel">
 	<div class="mb-2">
 		<%= label_for library_sets =>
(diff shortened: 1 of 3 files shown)


pstaabp:
I think "Course Files" sounds good.

pstaabp:
@dlglin the help page `InstructorSetMaker.html.ep` needs to be updated as well.

pstaabp (approved):

drdrew42 (approved):

drgrice1 (approved):

PR #2202, opened 2023-09-03
Add a PGML converted to the Problem Editor

pstaabp:
This introduces a second option to cleaning code which is convertToPGML, which changes the code in text blocks in the code to use PGML features. 

To test, make sure that openwebwork/pg#882 is checked out, since this is where the conversion scripts are located. And see that PR for more details about what conversions are done.


templates/HelpFiles/InstructorPGProblemEditor.html.ep
```diff
@@ -129,12 +129,14 @@
...
 	</dd>
 
-	<dt><%= maketext('Tidy Code') %></dt>
+	<dt><%= maketext('Clean Code') %></dt>
 	<dd>
-		<%= maketext('Reformat the code using perltidy. This will change the code in the editor window, and save '
-			. 'changes to the temporary file. In some cases (if the code contains backslashes or double tildes) this '
-			. 'can result in odd spacing in the code. So make sure to inspect the formatted code, and edit further or '
-			. 'revert if needed.') =%>
+		<%= maketext('Reformat the code using perltidy or a conversion to PGM>. Using perltidy will change the code '
```
drdrew42, line 134:
```suggestion
		<%= maketext('Reformat the code using perltidy or a conversion to PGML>. Using perltidy will change the code '
```

drdrew42 (commented):
Not tested yet, but I caught a typo.

drgrice1:
I haven't looked closely at this, but all problems fail to render in the problem editor with this pull request.  So something is broken.

drgrice1 (changes requested):
"Clean Code" is a poor choice of wording.  I think it should be changed to "Format Code".  What perltidy does could be interpreted as cleaning the code, although formatting the code is still better wording even in this case.  What your new script does is not cleaning the code at all. In fact, at this point it is leaving the code often in an even messier state with the current script result.  It is reformatting the code for PGML in any case, so "format" is a word that makes sense for this.

I am not sure that adding this is such a good idea at this point.  I will look at the script in openwebwork/pg#882, but this certainly depends on the readiness of that script.

templates/ContentGenerator/Instructor/PGProblemEditor/clean_code_form.html.ep
```diff
@@ -0,0 +1,32 @@
+<div>
+	<div class="form-check">
+			<%= radio_button 'action.clean_code' => 'tidyPGCode',
```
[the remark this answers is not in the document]
pstaabp: fixed.

lib/WebworkWebservice/ProblemActions.pm
```diff
@@ -22,6 +22,7 @@ use warnings;
 use Data::Structure::Util qw(unbless);
 
 use WeBWorK::PG::Tidy qw(pgtidy);
+use Webwork::PG::ConvertToPGML qw(convertToPGML);
```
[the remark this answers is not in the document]
pstaabp: Yep.  This is a case where MacOS is case-insensitive, which is why it's working on my computer.

lib/WebworkWebservice/ProblemActions.pm
```diff
@@ -142,4 +143,18 @@ sub tidyPGCode {
 	};
 }
 
+sub convertCodeToPGML {
+	my ($invocant, $self, $params) = @_;
+	my $code = $params->{pgCode};
+
+	# calling convertToPGML($code) without the module path returns an error.  Not sure why.
```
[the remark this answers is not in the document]
pstaabp: fixed.

lib/WebworkWebservice/ProblemActions.pm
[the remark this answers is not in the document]
pstaabp: fixed.

templates/HelpFiles/InstructorPGProblemEditor.html.ep
```diff
@@ -129,12 +129,14 @@
...
-	<dt><%= maketext('Tidy Code') %></dt>
+	<dt><%= maketext('Clean Code') %></dt>
 	<dd>
-		<%= maketext('Reformat the code using perltidy. This will change the code in the editor window, and save '
-			. 'changes to the temporary file. In some cases (if the code contains backslashes or double tildes) this '
-			. 'can result in odd spacing in the code. So make sure to inspect the formatted code, and edit further or '
-			. 'revert if needed.') =%>
+		<%= maketext('Reformat the code using perltidy or a conversion to PGML>. Using perltidy will change the code '
+			. ' in the editor window, and save changes to the temporary file. In some cases (if the code contains '
+			. 'backslashes or double tildes) this can result in odd spacing in the code.   The convert to PGML '
```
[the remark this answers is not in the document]
pstaabp: fixed.

pstaabp:
Changed the tab name to "Format Code" and fixed the errors.

htdocs/js/PGProblemEditor/pgproblemeditor.js
```diff
@@ -132,16 +132,17 @@
 					if (webworkConfig?.pgCodeMirror) webworkConfig.pgCodeMirror.setValue(data.result_data.tidiedPGCode);
 					else document.getElementById('problemContents').value = data.result_data.tidiedPGCode;
 					saveTempFile();
-					showMessage('Successfuly perltidied code.', true);
+					showMessage('Successfully '
+						+ (request_object.rpc_command == 'tidyPGCode' ? 'perltidied code.' : 'converted code to PGML'), true);
```
drgrice1, line 136: This line exceeds the 120 character limit.  Move the `true);` part to the next line to fix this.

lib/WebworkWebservice/ProblemActions.pm
```diff
@@ -142,4 +143,17 @@ sub tidyPGCode {
...
 }
 
+sub convertCodeToPGML {
+	my ($invocant, $self, $params) = @_;
+	my $code = $params->{pgCode};
+
+	my $converted_code = convertToPGML($code);
+
+	return {
+		ra_out => { tidiedPGCode => $converted_code },
```
drgrice1, line 153: It doesn't make sense to have this return an object with the `tidiedPGCode` key.  This is not tidied code.  It is PGML code.  This should be renamed.  The javascript should be updated to handle the correct return key.

htdocs/js/PGProblemEditor/pgproblemeditor.js
```diff
@@ -98,15 +98,15 @@
 			?.addEventListener('change', () => deleteBackupCheck.checked = true);
 	}
 
-	// Send a request to the server to perltidy the current PG code in the CodeMirror editor.
-	const tidyPGCode = () => {
+	// Send a request to the server to either perltidy or convert_to_PGML the current PG code in the CodeMirror editor.
+	const formatPGCode = () => {
```
drgrice1, line 102: I don't like the doubling up of the code in this method for the PG perltidy formatting and the PGML conversion.  The two requests have completely different return objects with different information, and it makes this a kludge.  You are cutting corners with this.

templates/ContentGenerator/Instructor/PGProblemEditor/format_code_form.html.ep
```diff
@@ -0,0 +1,32 @@
+<div>
```
drgrice1, line 1: This file should begin with `% last unless $c->{is_pg};` the same as it was in `pgtidy_from.html.ep`.  This form should not be shown for the course information file or hardcopy themes.

templates/HelpFiles/InstructorPGProblemEditor.html.ep
```diff
@@ -129,12 +129,14 @@
...
 	</dd>
 
-	<dt><%= maketext('Tidy Code') %></dt>
+	<dt><%= maketext('Format Code') %></dt>
 	<dd>
-		<%= maketext('Reformat the code using perltidy. This will change the code in the editor window, and save '
-			. 'changes to the temporary file. In some cases (if the code contains backslashes or double tildes) this '
-			. 'can result in odd spacing in the code. So make sure to inspect the formatted code, and edit further or '
-			. 'revert if needed.') =%>
+		<%= maketext('Reformat the code using perltidy or a conversion to PGML>. Using perltidy will change the code '
```
drgrice1, line 134: This should be
```suggestion
		<%= maketext('Reformat the code using perltidy or a conversion to PGML. Using perltidy will change the code '
```

drgrice1 (commented):
This is getting closer on this side of things.  There are still a few things that need to be fixed though.

htdocs/js/PGProblemEditor/pgproblemeditor.js
pstaabp: I was hoping not to duplicate code, but can write two separate functions to handle this.

drgrice1: I don't think two separate functions are needed, just separate fetch calls that handle the return data properly for the two different requests.

drgrice1 (approved):
I think that this is in decent shape now.  This is still pending on https://github.com/openwebwork/pg/pull/882 though.

Alex-Jordan:
Where is this at now (on both sides)? Is it a good time for me to test?

drgrice1:
On this side it has my approval.  On the PG side, it is probably okay also.  I think there is more to think about with that side of things though.

Alex-Jordan:
I tested this on a random OPL problem (`Library/NewHampshire/NECAP/grade11/gr11-2009/n11-2009-2s.pg`) and the problem has several errors after the conversion. Is this supposed to be working already, or is it just a first run and then we start refining the conversion details?

The before/after diff includes:
```
< BEGIN_TEXT
< You may use this blank grid to help you answer this question. $PAR
< If you don't choose to print it out, a piece of graph paper would$BR
---
> BEGIN_PGML
>
> You may use this blank grid to help you answer this question.
> If you don't choose to print it out, a piece of graph paper would
```

So it removed that `$PAR` and did not do anything in the PGML version that corresponds to a paragraph break. I see other places where a `$PAR` that would make an impact has nothing as its counterpart in the PGML version. (Not even two spaces at the end of a line for a line break.)

----

Also in the diff is this:
```
51,54c45,48
< $BR
< \{ image( insertGraph($graph), tex_size=>$TEXSIZE,
<     height=>$HEIGHT, width=>$WIDTH
<     ) \}
---
>
> [@  image( insertGraph([$graph]), tex_size=>[$TEXSIZE],
>     height=>[$HEIGHT], width=>[[$WIDTH]]
>     )  @]*
```

I think this is wrong for the variables within `[@ ... @]` to be there using PGML substitution markup. This is not the best example, but pretend `$TEXSIZE` was a non-numeric string like `'taco'`. Then `tex_size=>[$TEXSIZE]` will become `tex_size=>taco` after PGML substitution, and you have a bareword value. 

Additionally, why is `[[$WIDTH]]` coming out with double brackets?

In any case, the problem no longer displays the image that it did before the conversion, and among the many error messages is probably one or more that have to do with this.

----

Also in the diff, I see a line:
```
ANS(radio_cmp($radio->correct_ans()));
```
that becomes
```
# ANS(radio_cmp($radio->correct_ans()));
```
and it's not like back up in the PGML block the answer is declared in some PGMLish way.

pstaabp:
@Alex-Jordan Thanks for testing.
- A recent change to the code reintroduced the `$PAR` bug that you saw.  That's an easy fix. 
- I need to leave the code inside of `\{ \}` alone and not process it.  I'll work on that.
- The replacement of ANS( ... ) with [_]{...} isn't that straightforward.  This was an initial pass and in the documentation I note that an author will need to fix these.  It provides the [_]{}, but doesn't fill them in.  I may try to see if I can come up with a way to handle this, but a very general way of handling it is difficult.

Alex-Jordan:
Maybe with `[_]{}` it could come out as `[_]{FIXME}` or something with a clear indicator that the editor should do something now. "FIXME" is probably not the best though.

somiaj:
> * The replacement of ANS( ... ) with [_]{...} isn't that straightforward.  This was an initial pass and in the documentation I
> 
> * note that an author will need to fix these.  It provides the [_]{}, but doesn't fill them in.  I may try to see if I can come up with a way to handle this, but a very general way of handling it is difficult.

Since `ANS()` works with `PGML` just fine, maybe don't replace these, just leave the `ANS` alone, but ensure you put `[____]` in the correct position. The the documentation could then point users at updating these to use `[___]{....}`.

htdocs/js/PGProblemEditor/pgproblemeditor.js
```diff
@@ -138,10 +138,41 @@
 			.catch((err) => showMessage(`Error: ${err?.message ?? err}`));
 	};
 
+	// Send a request to the server to convert_to_PGML the current PG code in the CodeMirror editor.
```
drgrice1, line 141: I just noticed this comment.  Please change this to something like
```suggestion
	// Send a request to the server to convert the current PG code in the CodeMirror editor to PGML.
```
Comments should not refer to some non-existent method like `convert_to_PGML`.

Alex-Jordan:
I checked out a branch from the tip of `develop`, then merged this into that branch. When I go to the problem editor, I get:
Screenshot 2024-03-27 at 11 10 21 PM

I also tried checking out this branch and rebasing from `develop`, but that led to merge conflicts.

Alex-Jordan:
OK now I have both this and the corresponding PG branch going.

I loaded an OPL problem (the same one I tried to test before) and when I click to Format Code (either using perltidy or the new conversion to PGML) I get this error:


Screenshot 2024-03-27 at 11 34 16 PM


I went to just try this on `develop` (where the tab is still called "Tidy Code" and you only have the one option) and there I get lots of perltidy errors:

```


PG perltidy errors:

Perltidy version is 20230701

: Begin Error Output Stream
80: Unexpected leading ',' after a '('
81: Unexpected leading ',' after a '('
82: Unexpected leading ',' after a '('
83: Unexpected leading ',' after a '('
84: Unexpected leading ',' after a '('
86: Unexpected leading ',' after a '('
87: Unexpected leading ',' after a '('
88: Unexpected leading ',' after a '('
89: Unexpected leading ',' after a '('
92: Unexpected leading ',' after a '('
95: Unexpected leading ',' after a '('
98: Unexpected leading ',' after a '('
101: Unexpected leading ',' after a '('
104: Unexpected leading ',' after a '('
107: Unexpected leading ',' after a '('
110: Unexpected leading ',' after a '('
113: Unexpected leading ',' after a '('
116: Unexpected leading ',' after a '('
119: Unexpected leading ',' after a '('
122: Unexpected leading ',' after a '('
125: Unexpected leading ',' after a '('
128: Unexpected leading ',' after a '('
131: Unexpected leading ',' after a '('
134: Unexpected leading ',' after a '('
137: Unexpected leading ',' after a '('
140: Unexpected leading ',' after a '('
143: Unexpected leading ',' after a '('
146: Unexpected leading ',' after a '('

150: --More--(57%)
           --^
found ( where operator expected (previous token underlined)

153: $func = Compute("2+sqrt{1-(x-3)**2}");
     ^
found Scalar where operator expected

Missing ';' or ',' above?
162: Unexpected leading ',' after a '('

198: IN THIS PROBLEM ON THE NECAP EXAM YOU WERE EXPECTED $BR
     ^
found bareword where operator expected

Missing ';' or ',' above?
200: To see 28 non-critical warnings rerun with -w

```

So something may be wrong with this branch, but also it seems something is wrong now with `develop`. I also get this latter error when I try "Tidy Code" on my production server, which is on `develop` plus a bit more.

drgrice1:
I can't reproduce the issue on the develop branch that you are stating.  Are you sure you don't actually have a syntax error in the problem that you are trying to run perltidy on?  That would cause exactly the errors you are showing, and is what it should be doing.

drgrice1:
I do get the `Error: o.result_data is undefined` on this branch when I select "Convert the code to PGML" and click "Format Code".

drgrice1:
@pstaabp:  You need to either merge `origin/develop` in to this branch or rebase onto `origin/develop`.  I recommend that you merge.  Rebasing will be more challenging since there are conflicts.  That won't fix the error that @Alex-Jordan and I saw, but currently this branch is way behind develop, and does not include the `prettier` code formatting check.  So if this is merged, then all subsequent workflow runs will fail for that action because the `pgproblemedit.js` file is not correctly formatted in this branch.  So after merging `origin/develop` make sure to also run `npm run prettier-format` and commit the changes.

To fix the error that is occurring you need to update the authentication parameters for the `request_object` in the `convertCodeToPGML` method.  Make that the same as the code for that in the `tidyPGCode` method.

Alex-Jordan:
Off topic from this PR, but:
> I can't reproduce the issue on the develop branch that you are stating. 

Did you specifically use this OPL problem? `Library/NewHampshire/NECAP/grade11/gr11-2009/n11-2009-2s.pg`

I just tried with a different OPL problem and did not get the error. But when I returned to this  one and did "Tidy Code", I still get the error. In that problem, around line 80, it has this:

```
$graph->moveTo(,);
$graph->lineTo( , ,'black');
...
```

It seems questionable to have null on either side of a comma. So that would explain the error I get, but then why no error from just regular rendering?

drgrice1:
No, I did not test specifically with that problem (you didn't state the problem before).  With that problem I see the errors.  Unfortunately, there is not much that can be done about that.  It is a syntax error in Perl to have `$graph->moveTo(,);`.  In a regular Perl script that would fail even without `use strict` or `use warnings`.  Somehow it works when compiled in the safe compartment even though it shouldn't.

Most problems will work fine with the perltidy action, but those that do things wrong might have problems.  That is why the errors are shown.  I see that there should be an `overflow: auto` on the rendering `div` for this though for long errors like this.

Alex-Jordan:
I did state the problem...back on Nov 6 :)

drgrice1:
I see.  I didn't realize that you were using the same problem to test from before!

Alex-Jordan:
Well it turns out I randomly picked a problematic problem. When I tested with it in November, it had the issues that I reported and they've either been addressed or postponed. Does it make sense that at that time, I did not get the perltidy errors?

drgrice1:
I tested that problem on WeBWorK 2.18 on my production server, and it has the same issue.  So that is unlikely.

pstaabp:
There were some conflicts due to some changes.  Still working this out and should have some fixes soon.

pstaabp:
Just fixed all of the merge conflicts and pushed.  Ready for testing again.

htdocs/js/PGProblemEditor/pgproblemeditor.js
```diff
@@ -110,7 +110,11 @@
 
 	// Send a request to the server to perltidy the current PG code in the CodeMirror editor.
 	const tidyPGCode = () => {
-		const request_object = { courseID: document.getElementsByName('courseID')[0]?.value };
+		const request_object = {
+			user: document.getElementById('hidden_user')?.value,
+			courseID: document.getElementsByName('courseID')[0]?.value,
+			key: document.getElementById('hidden_key')?.value
+		};
```
drgrice1, line 117: This won't work.  Don't change this from what it is in develop.

htdocs/js/PGProblemEditor/pgproblemeditor.js
```diff
@@ -150,10 +154,43 @@
 			.catch((err) => showMessage(`Error: ${err?.message ?? err}`));
 	};
 
+	// Send a request to the server to convert the current PG code in the CodeMirror editor.
+	const convertCodeToPGML = () => {
+		const request_object = {
+			user: document.getElementById('hidden_user')?.value,
+			courseID: document.getElementsByName('courseID')[0]?.value,
+			key: document.getElementById('hidden_key')?.value
+		};
```
drgrice1, line 163: This also won't work.  Make this the same as it is for the `tidyPGCode` method in develop.

drgrice1 (changes requested):

htdocs/js/PGProblemEditor/pgproblemeditor.js
pstaabp: I thought something had changed between develop and this branch.  Will update this.

drgrice1:
With your current code in `pgproblemeditor.js`, each time you use either of the formatters, it results in you being signed out when using `session_management_via = 'session_cookie'`.  So you get `Error: o.result_data is undefined`, and if you try to do anything else, then you have to log in again.

pstaabp:
> With your current code in `pgproblemeditor.js`, each time you use either of the formatters, it results in you being signed out when using `session_management_via = 'session_cookie'`. So you get `Error: o.result_data is undefined`, and if you try to do anything else, then you have to log in again.

Right.  reverted this code.

drgrice1 (approved):
That's better.

There is one thing that I have noticed with this that I think could use improvement.  That is if a file is using PGML already, and you use the `Convert the code to PGML` action on it, it can mess things up.  For example, consider the following problem code.

```
DOCUMENT();
loadMacros('PGstandard.pl', 'PGML.pl', 'PGchoicemacros.pl', 'PGcourse.pl');

$cmc = new_checkbox_multiple_choice();
$cmc->qa('Which of the following is a letter?', 'a', 'b', 'c', 'd');
$cmc->extra('1', '2', '3', '4');

BEGIN_PGML
[@ $cmc->print_q @]*  
[@ $cmc->print_a @]*
END_PGML

ANS(checkbox_cmp($cmc->correct_ans));

ENDDOCUMENT();
```

This problem of course uses the deprecated `new_checkbox_multiple_choice` method from `PGcoursemacros.pl`, but the code as shown is the best you can do with it in PGML.  However, when this is "converted to PGML" with the new action, the `ANS` line is commented out.  It would be nice if the convert script would not modify this file when it is executed on a problem like this.

There are cases with PGML, even valid cases when  you are doing some advanced things, where you still need an `ANS` call.  In those cases, the `ANS` call should not be commented out.

pstaabp:
Good idea.  One option would be to just skip running the code if it seems there are already PGML blocks or (as you suggested) run the code, but don't comment out `ANS` lines.

PR #2205, opened 2023-09-08
Make the library browser database connections use a cached connection

drgrice1:
Currently the database connections used by the library browser are obtained with a `DBI->connect` call, and those connections are never closed.  As a result each time the library browser is accessed, a new database connection is obtained, and the previous connections are left open until the server process that created the connection is recycled. This results in the database connection count being unnecessarily incremented.

This changes those calls to `DBI->connect_cached`.  This means that if the process already has a database connection, it will be reused instead of a new one being made.  Thus keeping the database connection count down.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8355 for the discussion that led to this pull request.

### DIFF
--- a/lib/WeBWorK/Utils/LibraryStats.pm
+++ b/lib/WeBWorK/Utils/LibraryStats.pm
@@ -13,61 +13,43 @@
 # Artistic License for more details.
 ################################################################################
 
-###########################
-# Utils::LibraryLocalStats
-#
-# This is an interface for getting local statistics about library problems
-# for display
-###########################
-
+# This is an interface for getting global and local statistics about library problems for display.
 package WeBWorK::Utils::LibraryStats;
 
-use base qw(Exporter);
 use strict;
 use warnings;
+
 use DBI;
 
-our @EXPORT    = ();
-our @EXPORT_OK = qw();
-
 sub new {
-	my $class = shift;
-	my $ce    = shift;
+	my ($class, $ce) = @_;
 
-	my $dbh = DBI->connect(
-		$ce->{problemLibrary_db}->{dbsource},
-		$ce->{problemLibrary_db}->{user},
-		$ce->{problemLibrary_db}->{passwd},
+	my $dbh = DBI->connect_cached(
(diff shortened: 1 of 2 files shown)


drgrice1:
By the way, you can monitor the connection count in mysql/mariadb by entering `show status where variable_name='threads_connected';`.  If you test this with morbo, you should see that before starting morbo there is only one thread connected (the client you are using in the terminal) assuming you don't have anything else running that has a database connection.  After starting morbo and loading a non library browser page in the browser, you should see the connection count increase by one.  Loading any other non library browser page will not increase the connection count, no matter how many of these pages you open.  That is with this pull request or with the develop or main branches.  However, if you load the library browser page, you will see another connection appear.  With this pull request if you load the library browser again, the connection count will not increase further.  With develop or main, each time that you load the library browser page the connection count will continue to increase by one.

drdrew42 (approved):
Thanks for helping to track this down! I'd also argue the case for a hotfix here...

drgrice1:
I will make a hotfix pull request.

pstaabp:
There are a couple of other places that `DBI->connect` is used.  Many are in scripts--I'm guess these are irrelevant, but in `DBUpgrade`, `CourseManagement` and `CourseIntegrityCheck`, do these need to be change?

drgrice1:
The scripts should not use `connect_cached`.  That wouldn't make any sense.  There will not be a cached connection to work with ever for them.

The `CourseManagement` case can not be a `connect_cached` call.  It is a different connection than the default one, and it is also disconnected, so it won't permanently increase the connection count.

I am not entirely certain about the case in `CourseIntegrityCheck`.  In testing it did not permanently increase the the connection count.  Although I am not entirely certain why, since there is not disconnect call.  I think that it needs a separate connection than the default one as well.

The case in `DBUpgrade` doesn't matter. That module isn't actually used except for by the scripts in `bin/old_scripts`.  Those scripts and that file need to be deleted.

pstaabp (approved):

PR #2206, opened 2023-09-08
Use the `followSymlinks` instead of the `usePolling` chokidar option in generate-assets.js.

drgrice1:
With node version 20 the `usePolling` option is resulting in high cpu usage.  The `followSymlinks` serves the purpose more correctly.  The `usePolling` option was initially used because either the `followSymlinks` option did not exist or there was a bug in chokidar and that option wasn't working.  I can't remember the exact details.

The actual high cpu usage occurs with the PG generate-assets.js script, and not this one with the `usePolling` option.  I don't really know why though.  In any case the `followSymlinks` option is more correct for what we want here.

Note this only affects usage of the script with the `-w` flag.  So it only affects developers.


pstaabp:
According to https://github.com/paulmillr/chokidar, the `followSymLinks` defaults already to `true`.  Not bad to explicitly have it there.

pstaabp (approved):

PR #2207, opened 2023-09-10
Make the library browser database connections use a cached connection (hotfix)

drgrice1:
Currently the database connections used by the library browser are obtained with a `DBI->connect` call, and those connections are never closed.  As a result each time the library browser is accessed, a new database connection is obtained, and the previous connections are left open until the server process that created the connection is recycled. This results in the database connection count being unnecessarily incremented.

This changes those calls to `DBI->connect_cached`.  This means that if the process already has a database connection, it will be reused instead of a new one being made.  Thus keeping the database connection count down.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8355 for the discussion that led to this pull request.

### DIFF
--- a/lib/WeBWorK/Utils/LibraryStats.pm
+++ b/lib/WeBWorK/Utils/LibraryStats.pm
@@ -13,61 +13,43 @@
 # Artistic License for more details.
 ################################################################################
 
-###########################
-# Utils::LibraryLocalStats
-#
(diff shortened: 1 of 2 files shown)


drdrew42 (approved):
👍

pstaabp (approved):

PR #2208, opened 2023-09-12
Fix incorrect row label in stats page.

somiaj:
Accidentally used 'Point Value' twice, when one of the rows was meant to be 'Average Percent'.  This is a simple fix for that (probably not worth a hot fix, but simple enough I can make one if others think we should clean up this visual issue).


drgrice1 (approved):
It looks like this is actually my fault.  When it was moved from Stats.pm into the template it was not transferred correctly.

I think a hotfix is justified.

pstaabp (approved):

pstaabp:
I think a hotfix would be good.

drgrice1:
I am going to merge this with two approvals.

PR #2213, opened 2023-09-16
Fix `maketext` interpolation in PG.

drgrice1:
The key to making this work is to override the `_compile` method of `Locale::Maketext` and remove the `use strict` usage inside of its code eval call. That is the only essential difference between the override method added here and the orignal method in `Locale::Maketext`.

With this you can now use maketext interpolation in PG. For example, you can call `maketext('[quant,_1,dog] ate my homework', $num_dogs)`.

This is an important part of removing the attempts table, and moving the code the provides feedback to PG.

The WeBWorK::Localize module is also cleaned up considerably.

### DIFF
--- a/lib/WeBWorK/Localize.pm
+++ b/lib/WeBWorK/Localize.pm
@@ -1,70 +1,49 @@
 package WeBWorK::Localize;
+use parent 'Locale::Maketext';
+
+use strict;
+use warnings;
 
 use File::Spec;
-use Locale::Maketext;
 use Locale::Maketext::Lexicon;
 
 use WeBWorK::Utils qw(x);
 
-my $path     = "$ENV{WEBWORK_ROOT}/lib/WeBWorK/Localize";
-my $pattern  = File::Spec->catfile($path, '*.[pm]o');
-my $decode   = 1;
-my $encoding = undef;
-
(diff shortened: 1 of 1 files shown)


pstaabp:
This seems to perform the translation fine, but not sure how to test the interpolation, since we don't have any of that in PG.  Perhaps this just will pave the way for removing the attempts table, or @drgrice1, do you have a good way to test?   

Would we now have the ability to do maketext in a macro perhaps?

drgrice1:
There is an easy way to test.  Add `maketext('[quant,_1,dog] ate my homework', $num_dogs)` to a problem, and then run `/opt/webwork/webwork2/bin/dev_scripts/update-localization-files -p`.  Then modify that string in your chosen language `.po` file.  Then switch the course to that language, and see that the translation comes through.  Test it with various values of `$num_dogs`.

Note you can also test this for English without running the script testing for various values of `$num_dogs`.  It should say "1 dog ate my homework" if `$num_dogs = 1`, and "2 dogs ate my homework" if `$num_dogs = 2`.  If you then test the problem with the develop branch, you will get an error.

drgrice1:
I am closing this in favor of #2225 and https://github.com/openwebwork/pg/pull/936.

PR #2214, opened 2023-09-18
Send Email Notification for Achievements

drdrew42:
## New Achievement Feature

This PR enables email notifications on a per-achievement basis.

Each achievement can be paired with an email template that will be filled out and sent to the student when they earn the achievement.

The email template itself is a hybrid of `Mojo::Template` and our own email template system (as implemented on the Email page). A default template (default.html.ep) must be present in the `[TMPL]/achievements/` folder. This file is provided in courses.dist/achievements/

I have left the template as an HTML template, in case there is desire to support MIME::HTML email messages.

To enable an email notification:
* the site.conf (or course.conf, at least) must have nonempty `$mail{smtpSender}` or `$mail{set_return_path}`
* the default.html.ep must be copied from `courses.dist/achievements/` into the `[TMPL]/achievements` folder (I'm hoping that #2173 will help ensure this happens)
* achievements must be enabled in course config
* import and assign default_achievements.axp (or whatever else you might have)
* head to the Achievement List page and click "Enable Email Notification" (this should default to default.html.ep)
* make modifications to the template or "use existing template" default.html.ep

Once enabled, the email notification can be disabled by clicking into "Edit Email Template" and selecting the "Disable Notifications" tab.

### Caveat

This PR affects the webwork2-job-queue service, and as such you must restart both webwork2 and webwork2-job-queue when testing.


lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
```diff
@@ -431,7 +431,7 @@ sub import_handler ($c) {
 		$achievement->achievement_id($achievement_id);
 
 		# Fall back for importing an old list without the number or assignment_type fields
-		if (scalar(@$data) == 9) {
+		if (scalar(@$data) == 7) {
 			# Old lists tend to have an extraneous space at the front.
 			for (my $i = 1; $i <= 7; $i++) {
 				$$data[$i] =~ s/^\s+//;
```
pstaabp, line 437: If you are checking if the length is 7, then the indices will be 0..6, so isn't `$$data[7]` undefined?

drdrew42: Something was/is weird here. 

The old check used up to `$$data[7]`  if length was 9, and used up to `$$data[9]` otherwise.

I honestly think we should deprecate the use of 'old' achievement files for import and focus on using the current format. Shall we call for a vote? 😛

drgrice1: Clearly the old code was messed up here.  A current achievement import file should have 10 fields in each row (prior to this pull request).  So if two fields were added (one for assignment_type and one for number), then previously there would have been 8 fields, not 9 or 7.  The code is consistent with this in terms of which indices of the `$data` array are accessed.  So clearly the check should be `if (@$data == 8) {`.

I am fine with deprecating the usage of the "old" achievement files.  The error in the conditional means that those files have not been working anyway.  With the previous condition of `@$data == 9` the code for dealing with the "old" files would never have been executed.  With the new condition of `@$data == 7` it will still not be executed.  This means that if one of these "old" files were imported all of the fields would have been put into the wrong achievement columns, and most likely there would be errors.  Since this has been this way since 2015, most likely no one has any of these "old" files anymore.  Just drop that code.

lib/WeBWorK/ContentGenerator/Instructor/AchievementNotificationEditor.pm
```diff
@@ -0,0 +1,394 @@
...
+				'Write permissions have not been enabled for "[_1]".  '
+					. 'Changes must be saved to another file for viewing.',
+				$c->shortPath($outputFilePath)
+			);
+		} else {
+			$errorMessage =
+				$c->maketext('Unable to write to "[_1]": [_2]', $c->shortPath($outputFilePath), $writeFileErrors);
+		}
+
+		$c->{failure} = 1;
```
drgrice1, line 203: Delete `$c->{failure}` everywhere.  It is only used locally, and it will be true if and only if `$writeFileErrors` is true.  So it is not needed.

lib/WeBWorK/ContentGenerator/Instructor/AchievementNotificationEditor.pm
```diff
@@ -0,0 +1,394 @@
...
+	{
+		$c->addgoodmessage($c->maketext('Saved to file "[_1]"', $c->shortPath($outputFilePath)));
+	}
+
+	return;
+}
+
+sub save_handler ($c) {
+	my $courseName      = $c->{courseID};
+	my $achievementName = $c->{achievementID};
```
drgrice1, line 220: Delete these variables.  They are not used anywhere in this method.

lib/WeBWorK/ContentGenerator/Instructor/AchievementNotificationEditor.pm
```diff
@@ -0,0 +1,394 @@
...
+
+	if (!$achievement) {
+		$c->addbadmessage("Achievement $c->{achievementID} not found!");
+		return;
+	}
+
+	$c->{achievement} = $achievement;
+	my $template_filename = $achievement->email_template || "default.html.ep";
+	$c->{sourceFilePath} =
+		$ce->{courseDirs}{achievements} . "/$template_filename";
```
drgrice1, line 66: This should be one line.

lib/WeBWorK/ContentGenerator/Instructor/AchievementNotificationEditor.pm
```diff
@@ -0,0 +1,394 @@
...
+use Mojo::Base 'WeBWorK::ContentGenerator', -signatures;
+
+=head1 NAME
+
+WeBWorK::ContentGenerator::Instructor::AchievementNotificationEditor - edit an achevement evaluator file
+
+=cut
+
+use HTML::Entities;
+use File::Copy;
```
drgrice1, line 26: Unless I am missing something, these modules are not used in this package.  So these lines should be removed.

lib/WeBWorK/ContentGenerator/Instructor/AchievementNotificationEditor.pm
```diff
@@ -0,0 +1,394 @@
...
+
+use constant ACTION_FORMS => [qw(save save_as existing disable)];
+use constant ACTION_FORM_TITLES => {
+	save     => x('Save'),
+	save_as  => x('Save As'),
+	existing => x('Use Existing Template'),
+	disable  => x('Disable Notifications'),
+};
+
+use constant DEFAULT_ICON => 'defaulticon.png';
```
drgrice1, line 38: This constant is not used by this module.  So this line should be deleted.

templates/ContentGenerator/Instructor/AchievementList/default_table.html.ep
```diff
@@ -55,6 +56,11 @@
 						) =%>
 					</td>
 					<td class="text-nowrap">
+            % my $templateText = $_->email_template ? maketext('Edit Email Template') : maketext('Enable Email Notification');
+						<%= link_to $templateText => $c->systemLink(
+							url_for('instructor_achievement_notification', achievementID => $achievement_id),
+						) =%>
```
drgrice1, line 62: Rather than using an intermediate variable once, do
```suggestion
						<%= link_to $c->systemLink(
							url_for('instructor_achievement_notification', achievementID => $achievement_id)
						), begin =%>
							<%= $_->email_template
								? maketext('Edit Email Template')
								: maketext('Enable Email Notification') =%>
						<% end =%>
```
The existing code also had an indentation issue, and a line that exceeded the 120 character limit.

templates/ContentGenerator/Instructor/AchievementNotificationEditor/save_as_form.html.ep
```diff
@@ -0,0 +1,36 @@
+% # There are three things you can do with a new achievement editor.
+% # You can replace the current achievement, use it in a new achievement, or not use it at all.
```
drgrice1, line 2: Either update these comments for the notification editor, or delete them.

lib/WeBWorK/ContentGenerator/Instructor/AchievementNotificationEditor.pm
```diff
@@ -0,0 +1,394 @@
...
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+package WeBWorK::ContentGenerator::Instructor::AchievementNotificationEditor;
+use Mojo::Base 'WeBWorK::ContentGenerator', -signatures;
+
+=head1 NAME
+
+WeBWorK::ContentGenerator::Instructor::AchievementNotificationEditor - edit an achevement evaluator file
```
drgrice1, line 21: This line should be updated for the notification editor instead of the evaluator editor.

lib/WeBWorK/ContentGenerator/Instructor/AchievementNotificationEditor.pm
```diff
@@ -0,0 +1,394 @@
...
+	$c->{sourceFilePath} =
+		$ce->{courseDirs}{achievements} . "/$template_filename";
+
+	my $actionID = $c->param('action');
+
+	# Perform a save or save_as action
+	if ($actionID) {
+		unless (grep { $_ eq $actionID } @{ ACTION_FORMS() }) {
+			die "Action $actionID not found";
+		}
```
drgrice1, line 74: This is a good candidate for using the `die ... unless ...` syntax instead.

drgrice1 (commented):
There is quite a bit of clean up needed.  Most of this is the result of copying the `AchievementEditor.pm` file and its related files which also need a lot of clean up.  Since these are new files, it would be good to add them already cleaned up.

lib/WeBWorK/ContentGenerator/Instructor/AchievementEditor.pm
```diff
@@ -25,7 +25,7 @@ WeBWorK::ContentGenerator::Instructor::AchievementEditor - edit an achevement ev
 use HTML::Entities;
 use File::Copy;
 
-use WeBWorK::Utils qw(not_blank path_is_subdir x);
+use WeBWorK::Utils qw(fix_newlines not_blank path_is_subdir x);
```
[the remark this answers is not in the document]
drdrew42: This is my mistake with the new `fix_newlines`, I thought I'd included "?" in the regex: `/\r\n?/\n/`

I'll update the util method.

templates/ContentGenerator/Instructor/AchievementList/default_table.html.ep
drdrew42: I had a tough time making this ternary work embedded in the link_to... thanks for finding the begin/end approach.

drdrew42:
Okay, I think that all of the feedback has been resolved. In multiple places:
* extra dereferencing arrows are removed
* stash values are not duplicated on the controller
* ui messages are passed through maketext

Also, cleanup of conditionals (e.g. not checking conditions that already returned early); removal of unused variables and modules; and related template cleanup.

lib/WeBWorK/AchievementEvaluator.pm
```diff
@@ -245,12 +245,25 @@ sub checkForAchievements ($problem_in, $pg, $c, %options) {
...
+
+			# if email_template is defined, send an email to the user
+			$c->minion->enqueue(
+				send_achievement_email => [ {
+					recipient       => $user_id,
+					subject         => 'Congratulations on earning a new achievement!',
+					courseName      => $ce->{courseName},
+					achievementID   => $achievement_id,
+					setID           => $set_id,
+					nextLevelPoints => $nextLevelPoints,
```
drgrice1, line 263: I am getting the warning `Argument "" isn't numeric in subtraction (-) at /opt/webwork/courses/myTestCourse/templates/achievements/default.html.ep line 7.` when an achievement is earned that is not a level achievement.  It may also because in testing I have just reassigned all achievements, and the user does not have any achievement points yet.  In any case, I think this needs to be
```suggestion
					nextLevelPoints => $nextLevelPoints || 0,
```

templates/ContentGenerator/Instructor/AchievementNotificationEditor/existing_form.html.ep
```diff
@@ -0,0 +1,7 @@
+<div class="row align-items-center mb-2">
+	<%= label_for 'action.existing.target_file_id' => maketext('Use existing template:'), class => 'col-form-label col-auto' =%>
```
drgrice1, line 2: This line exceeds 120 characters.  Move the `class => ...` to the next line to fix this.

drgrice1 (commented):
There are a few more things I observed in the code.   I also got to testing functionality, and I am seeing some warnings.

First, I am getting `Use of uninitialized value $val in pattern match (m//) at /usr/share/perl5/Email/MIME/Encode.pm line 36.` warnings anytime an email is sent by the `send_achievement_email` task.  I am not sure what is causing that exactly.  This does not happen with the `send_instructor_email` task.

lib/Mojolicious/WeBWorK/Tasks/AchievementNotification.pm
```diff
@@ -0,0 +1,112 @@
...
+	# parse email template similar to how it is done in SendMail.pm
+	my $msg = processEmailMessage(
+		$mail_data->{body}, $user_record,
+		$ce->status_abbrev_to_name($user_record->status),
+		$mail_data->{merge_data}
+	);
+
+	my $email =
+		Email::Stuffer->to($user_record->email_address)->from($mail_data->{from})->subject($mail_data->{subject})
+		->text_body($msg)->header('X-Remote-Host' => $mail_data->{remote_host});
```
drgrice1, line 97: The `$mail_data->{remote_host}` is guaranteed to be undefined.  This is never set in `AchievementEvaluator.pm` when this task is enqueued.  This is where the messages `Use of uninitialized value $val in pattern match (m//)` in the `Email/MIME/Encode.pm` module are coming from.  `SendMail.pm` sets this from the current request transaction.  See line 237 of that file.

drgrice1 (commented):
The achievement notification editor page needs a help file.  It should be `templates/HelpFiles/InstructorAchievementNotificationEditor.html.ep`.  One thing that is not immediately clear is how you actually enable email notifications, and what the "Use Existing Template" form is for.  I see that the "Use Existing Template" form activates sending emails for the achievement, but a help file to explain that would help.  Also, maybe it would be better if that were a `select` that shows the available templates to choose from.

lib/Mojolicious/WeBWorK/Tasks/AchievementNotification.pm
```diff
@@ -0,0 +1,112 @@
...
+		return $job->fail($job->maketext("An error occurred while trying to send email: [_1]", $@));    # fail silently
+	}
+	$job->app->log->info("Message sent to $mail_data->{recipient}");
+	return $job->finish($job->maketext("Message sent to [_1]", $mail_data->{recipient}));    # succeed silently
+}
+
+sub send_achievement_notification ($job, $ce, $db, $mail_data) {
+	if ($ce->{mail}{smtpSender} || $ce->{mail}{set_return_path}) {
+		$mail_data->{from} =
+			$ce->{mail}{smtpSender} || $ce->{mail}{set_return_path};
```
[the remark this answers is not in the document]
dlglin: Would it make sense to have a configuration variable for the from address on these messages?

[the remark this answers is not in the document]
drgrice1: I think that is needed.  This could use a course setting for this.

lib/WeBWorK/AchievementEvaluator.pm
```diff
@@ -245,12 +245,26 @@ sub checkForAchievements ($problem_in, $pg, $c, %options) {
...
+			$c->minion->enqueue(
+				send_achievement_email => [ {
+					recipient       => $user_id,
+					subject         => 'Congratulations on earning a new achievement!',
+					courseName      => $ce->{courseName},
+					achievementID   => $achievement_id,
+					setID           => $set_id,
+					nextLevelPoints => $nextLevelPoints || 0,
+					pointsEarned    => $achievementPoints,
+					remote_host     => $c->tx->remote_host || "UNKNOWN",
```
drgrice1, line 265: This should be
```suggestion
					remote_host     => $c->tx->remote_address || "UNKNOWN",
```

drgrice1 (approved):
This is looking pretty good now.  One minor issue needs to be fixed.

somiaj:
Overall seems like a nice feature. Some things I noticed playing around with it.

First, the Achievement Notification editor still needs a help page (previously mentioned by @drgrice1).

Do you think instead of the link just saying, "Edit Email Template", it could state which template is being used and edited, since you can use the same template for multiple achievements? Maybe "Edit Email Template default" (the name of the template minus the .html.ep suffix). This way it is clear from the editor which template is being used.

Also when selecting a template in the "Use Existing Templates" form, do you think using a drop down menu that lists all .html.ep files in the templates/achievements directory would be nice (so a user doesn't have to know them by name)?

The Save As form doesn't select an option by default, and one needs to be selected to work (though I guess this is the same behavior as the Evaluator Editor, so if you decide to provide a default, that should be adjusted as well). I was just comparing this to the problem editor which selects an option by default under Save As.

The "Save As" form selecting "use in new achievement" both seems weird and doesn't work. If you enter in an achievement name that exist you get an error, "Achievement ID exists! No new achievement created. File not saved.", but if you enter in a new name you get the error "Don't recognize saveMode: |use_in_new|. Unknown error.", so this doesn't appear to do much. I don't think creating a new achievement here is the correct thing to do (since the best you can do is attach a email template to it). Maybe instead this should be add to existing achievement, and a drop down menu with all the current achievements you can add the email template to.

When using the achievement notification editor, a sub menu item for the achievement appears, but isn't fully clear what it is for, maybe call it "name evaluator" to make it clearer that links is for the evaluator editor. Would adding one for the email template editor be helpful too (just a thought, this will probably be too clunky)?

Also, for those who don't know mojolicious the suffix html.ep might be a bit odd for them to understand. Would it be helpful to hide this from the user and append it as needed? I guess they can see this in the file manager though, so this might just add useful info for some.

drgrice1 (changes requested):
Unfortunately, I must revoke my approval of this pull request.  I started looking into things a little bit deeper after @somiaj's review.  I realized I didn't test some things very well when I say his comments.  Those things are not to bad.  I did quite a bit of work to fix those, and was going to put in a pull request to this branch to fix those things and some other issues I encountered along the way.  Then I found another problem, that can not be easily resolved.  That is the usage of Mojo::Template to render templates that are created by instructors.  That is a major security vulnerability, and unfortunately can't be allowed.  Any perl code can be executed in these templates.  Modules can be `use`d, any files that the server has permissions to modify or delete can be deleted, etc.  If the templates could be compiled in the safe zone, then perhaps this could be fixed, but I haven't found a way to do that yet.  It most likely can not be securely done.  So another approach for email templates is going to need to be found.

drgrice1:
I went ahead and put the work that I had done to address @somiaj's comments and some other things that I saw into a pull request to this branch.  I will see what alternatives I can devise to using the `Mojo::Template` module.

somiaj:
Emails don't need to include any html, you could use a template method similar to what is being done on the email page.

drgrice1:
Yes, that is obvious.  But that is a far more limited method.

drgrice1:
@drdrew42:  The pull request now has been switched to using the basic interpolation method of `WeBWorK::Utils::processEmailMessage` again, and then interpolates a few other variables specific to achievements.  It no longer uses Mojolicious templates ... sigh.  There may be other variables that you might want interpolated in as well though.  If you merge that pull request, then this can pull request can proceed again.  I am still probably going to see if I can find a way to get Mojolicious templates to work in the safe zone, but it doesn't look likely.

drgrice1:
@drdrew42:  I got Mojolicious template compilation to work in the safe zone!  I removed interpolation approach that I had added to the pull request to this branch, and switched it to using templates again with this approach.  So if you merge that pull request into this one, then this is good to go again.

drgrice1 (approved):
Looks good now!

Issue #2215, opened 2023-09-25
Single Problem Grader gives an error in GatewayQuizes if problems don't start at 1


somiaj:
I ran across this with a quiz whose problems didn't start at one. Each time I used the grader to update the score or save a comment I get an error, `document.gwquiz.elements[("probstatus" + t.problemId)] is undefined`. In my case I have a single problem in the quiz whose problem number (Id) is 2.

This is due to the fact that the hidden elements `probstatusX` assume that the problem numbers start at one and are sequential.  I looked into this, but didn't see a good way to fix it because X needs to be sent to an array to find the correct problem if random order is being used.

Due to when the error is returned, any change in the grade is saved in the database, but comments are not saved.

somiaj:
Actually looks like my initial observations was incorrect, in the field `probstatusX`, the `X` needs to be the problem ID, and was incorrect being set to the array value, which works unless the problems are not in sequential order. As such I think I have a fix, PR incoming.

somiaj:
Issue fixed in two different PRs. Closing.

PR #2216, opened 2023-09-25
Fix issue with hidden input probstatusX in gateway quizzes.

somiaj:
The number `X` in the hidden input `probstatusX` needs to be the
  problem ID, but was being set to the index + 1 of the array used
  to randomized the problems.  This causes the single problem grader
  javascript to hit an error if a test's problems do not start at
  1 and are not in sequential order, as mentioned in #2215.

  Also, since GatewayQuiz.pm assumes the X is the problem ID, and uses
  this hidden input to get the status of its problems when moving though
  different pages, a quiz whose problems don't start with 1 or are not
  in sequential error could retrieve an incorrect status as a result.

  This update fixes the issue by setting X equal to the problemID.

### DIFF
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -589,7 +589,8 @@
 					% }
 				</div>
 				% # Store the problem status for continued attempts recording.
-				<%= hidden_field 'probstatus' . ($probOrder->[$i] + 1) => $c->{probStatus}[ $probOrder->[$i] ] =%>
+				<%= hidden_field 'probstatus' . $problems->[ $probOrder->[$i] ]{problem_id}
+					=> $c->{probStatus}[ $probOrder->[$i] ] =%>
 				%
 				<div class="gwDivider"></div>
 			% } else {
@@ -605,7 +606,8 @@
 					% }
 				% }
 				% # Store the problem status for continued attempts recording.
-				<%= hidden_field 'probstatus' . ($probOrder->[$i] + 1) => $c->{probStatus}[ $probOrder->[$i] ] %>
+				<%= hidden_field 'probstatus' . $problems->[ $probOrder->[$i] ]{problem_id}
+					=> $c->{probStatus}[ $probOrder->[$i] ] %>
 			% }
 		% }
 		%


somiaj:
Note, assuming this fix is correct, I would suggest a hotfix.

drgrice1 (approved):
Looks good and is correct logically.  Furthermore it checks out in testing.  I tested with a test that had problems numbered 1, 2, 3, 4, 6 and set to have the problems ordered randomly.  When grading the problem in position 5, with the develop branch I see the error, and with this branch the comment is saved to the correct past answer in the past answers table.

Hopefully this is the last of the issues for gateway quizzes with non-consecutive problems.  Crossing fingers!

drgrice1:
@somiaj:  It is exactly where you are looking in `GatewayQuiz.html.ep` to set the hidden field.  The index of the name is the problem order plus one, and not the problem id.  That is inconsistent with any of the places that are using that name with an index by problem id, and that is a problem.

drgrice1:
Actually, the values are also accessed on lines 293-303 of `GatewayQuiz.html.ep` by problem order and not by problem id.

drgrice1 (approved):
Okay.  After looking quite closely at this, I see that it is properly fixing the issue.  The structure for this in the GatewayQuiz.pm module is very confusing though.

drgrice1:
I think a hotfix is justified.

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -589,7 +589,8 @@
 					% }
 				</div>
 				% # Store the problem status for continued attempts recording.
-				<%= hidden_field 'probstatus' . ($probOrder->[$i] + 1) => $c->{probStatus}[ $probOrder->[$i] ] =%>
+				<%= hidden_field 'probstatus' . ($problems->[ $probOrder->[$i] ]{problem_id})
```
drgrice1, line 592: Note that the parentheses are no longer needed here (and also on the other line changed).

pstaabp (approved):
Seems to fix this.  🤞

somiaj:
Okay, I removed the unneeded parenthesis, it should be ready to go now.

drgrice1:
Okay. I will merge it.

Issue #2218, opened 2023-09-29
CodeMirror editor issues when resizing (hiding/showing main menu)


somiaj:
When hiding/showing the main menu in the problem editor, this resizes the codemirror editor which won't act correctly when using the mouse to move the cursor, select text, etc until you first type into the editor.  After typing in the editor, codemirror editor updates itself for its new size, and then functions normally.  Would it be possible to tell codemirror to update after the resize?

drgrice1:
Could you please be more explicit in your description of the problem?  I don't see the issue, and can't glean what it is from what you have said.  If I hide or show the main menu, the code mirror window resizes, but everything seems to be updated and functioning as it should.  The mouse can move the cursort, select text, etc normally.

somiaj:
I am still not exactly sure how to trigger it, and the behavior only shows up some times. A few more tests seems to show it is an issue with long lines whose text is wrapped to the next line, but after the resize, are no longer wrapped to the next line. so the number of lines in the codemirror editor change. Also, I should have been more clear, there are only certain areas you are not able to move the mouse.  Here are some screenshots to hopefully help out.  I start with a narrow browser that still shows the main menu.

![image](https://github.com/openwebwork/webwork2/assets/18082412/56a2bb07-de7c-4266-9389-9e820e6164a8)

After the resize the few long lines you see wrapped are no longer wrapped

![image](https://github.com/openwebwork/webwork2/assets/18082412/560fddb8-6854-4889-9408-9ce0ec8562d4)

When I click on the `PGcourse.pl` line, the cursor isn't placed where I click, instead it is placed on the `PGML.pl` line. As you click on lines below that one, such as the `$pi = Real("pi");` line, the cursor is placed on the line which starts `# Uncomment`. But as soon as you type any character into the editor, it seems to update itself and you can then move the cursor to the location you click.

Also I guess my original comment about selecting text was incorrect. When I try to select text, it initially selects the wrong text due to not putting the cursor in the correct location, the selecting of text then seems to update codemirror so things function normally again too.

somiaj:
Issue fixed. Closing.

PR #2219, opened 2023-09-29
Refresh the code mirror window when its containing div resizes.

drgrice1:
This fixes issue #2218.


somiaj (approved):
This fixes the issue. Thanks.

pstaabp (approved):

PR #2220, opened 2023-09-29
allow set list to be sorted by reduced scoring date

Alex-Jordan:
A local instructor requested this, and this way of doing it was quick. I'm not 100% sure on the code formatting since this is a template file.

Sorting sets requires going to the Sort tab in the Hmwk Sets Editor. Is there a reason we don't enable simply clicking on column headers if you want to sort the table of sets, like we do with the Classlist Editor to sort users?  Is it too hard to sort that way when the cells have dates?


drgrice1 (approved):
Looks good.

I think sorting by column headers is just something that has not been implemented for the homework sets editor.

pstaabp (approved):

PR #2225, opened 2023-10-05
Remove PG translations and don't add those to the po and pot files.

drgrice1:
PG will no longer use the language handle from webwork2.  It will now provide its own language handle.  So the PG code is no longer scanned for translations.

Also, the new WeBWorK::PG::Localize module is shared to PG's safe compartment.  That is PG's language handle module.

This is the alternate to #2213 that we discussed.

This is paired with https://github.com/openwebwork/pg/pull/936.

Note that the po and pot files have not actually been updated.  So at this point all of PG's translations are still in those files.  Those are saved for now, but can be removed later.

### DIFF
--- a/bin/dev_scripts/update-localization-files
+++ b/bin/dev_scripts/update-localization-files
@@ -39,8 +39,8 @@ do
 	esac
 done
 
-if [ -z "$WEBWORK_ROOT" ] || [ -z "$PG_ROOT" ]; then
-    echo >&2 "You need to set both the WEBWORK_ROOT and PG_ROOT environment variables.  Aborting."
+if [ -z "$WEBWORK_ROOT" ]; then
+    echo >&2 "You need to set the WEBWORK_ROOT environment variable.  Aborting."
     exit 1
 fi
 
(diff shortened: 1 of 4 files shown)


bin/dev_scripts/update-localization-files
```diff
@@ -53,9 +53,9 @@ LOCDIR=$WEBWORK_ROOT/lib/WeBWorK/Localize
 
 cd $LOCDIR
 
-echo "Updating $WEBWORK_ROOT/webwork2.pot"
+echo "Updating $LOCDIR/webwork2.pot"
 
-xgettext.pl -o webwork2.pot -D $WEBWORK_ROOT/lib -D $PG_ROOT/lib -D $PG_ROOT/macros -D $WEBWORK_ROOT/templates \
+xgettext.pl -o webwork2.pot -D $WEBWORK_ROOT/lib -D -D $WEBWORK_ROOT/templates \
```
pstaabp, line 58: I think there's an extra `-D` on this line.

pstaabp (approved):
This looks good.  Tested inside a problem both the ability to do plurality as well as translation using an already existing translation.

bin/dev_scripts/update-localization-files
drgrice1: Thanks for pointing that out. I did that in this and an the paired PG pull request.  I fixed both.  I am a little surprised the script was still working with that.

PR #2228, opened 2023-10-07
Remove the attempts table.

drgrice1:
PG will now provide feedback and generate the results summary.  So the AttemptsTable module will not be needed anymore.  It also conflicts with the PG feedback in some cases.

Without the attempts table the problem page top matter looked too spaced out, so the styles were tweaked to improve spacing.

The ResultsWithError and ResultsWithoutError classes were dispensed with, and instead bootstrap alerts are used with the alert-danger, alert-success, and alert-warning classes.  This is done in the last commit, and we can revert it if we aren't happy with it.

Also remove the unused `$pg{options}{correct_answer}` and `$pg{options}{incorrect_answer}` variables from defaults.config

Note that this includes #2225.

This is paired with the PG pull request https://github.com/openwebwork/pg/pull/939.


pstaabp:
Just started looking at this and overall, this is great.  

One first thought is that on most problem types, I don't think it is necessary to show an answer preview.  This especially includes graph tool problems and checkboxes/radio buttons and matrices.  I know you just moved what was in the attempts table to the popovers.   (Although with math quill, many formulas don't need to be previewed anymore either).

Alex-Jordan:
Here's some feedback. This is an excellent advancement overall.

* There's something of a candy cane look to the cards, and I think part of that is the `.card-header`s are the same height as the `.card-body`s (usually). What do you think about reducing the top/bottom padding on the card headers?
* Could you make it so that if you click on the `.popover-header`, that would be an additional way to dismiss it?
* Suppose a problem has several answer blanks and MathQuill is in use. A keyboard navigating user submits answers and learns that all was correct except the last answer blank. They want to get focus to that last answer input. With the old answer table, they can use a link from a certain cell in that table that jumps focus to the incorrect answer blank. With this change, they probably have to tab their way to that bad answer blank. Now that's not so bad, except that as they reach each of the earlier inputs, they have to also tab their way through all of MathQuill tools. I am not sure what to do about this, but I think it will be annoying to a keyboard navigator. It's already bothering me as I try to test things with keyboard-only navigation :)
* Similarly you have to tab through all the MQ tools just to get from an answer input to the adjacent info button. Is there a hotkey for disabling the MQ toolbar? I know I can right click to disable/enable it but is there a keyboard way to activate that?
* The div with the popover is in the DOM following the whole problem statement. I'm guessing the aria attributes you have will be good for the screen reader to read the popover content. However I might want to move focus into the popover content. (Maybe to activate the MathJax menu on some math content; maybe to get to and explore an image answer like from the graph tool; etc.) After activating the popover, a user would probably think they can tab right on into the popover  content, but actually they will have to tab all the way to the end of the rendered problem. Is it possible to place the div with the popover right after the button that activates it?
* If a set has a problem with an essay answer, it seems I always get a blue info button and the card is "Ungraded" with "This answer will be graded at a later time." Even if the problem has actually been graded, and I return as the student and click to check answers.  I'm not sure what to do though, because even if an instructor grades that problem and even if they leave a feedback message, it's not clear that the *essay* textarea input has been graded, as opposed to other elements of a problem. What do you think?

drgrice1:
First, I think your comments should be moved to https://github.com/openwebwork/pg/pull/939.  That is where everything you are talking about is really implemented.  This just removes the attempts table from webwork2.

* candy cane look: That is easy.  Adding `p-1` to the card-headers will accomplish a slightly smaller header.
* popover header click dismiss:  Making the popover header dismissible by a click is easy.  However, if that is done then it also needs to be made focusable and needs to be also disimissible by a keyboard space or enter key down.  In that case it also needs interactive semantics.  This can all be done, but is quite a bit of work for something that is not necessary.  I am not sure it adds that much either.  Also, not all of the feedback popovers even have a header.  This is the case for a "Preview".
* MathQuill editor toolbar and keyboard navigation: Unfortunately there is not much that can be done about this.  Note that once you have tabbed into the toolbar you can hit "Escape" to close the toolbar, and focus will be left after the MathQuill input.  So pressing "Tab" once more after that will move to the feedback button.  So in short, if focus is in the MathQuill input typing Tab, Escape, Tab will put focus on the feedback button.

  The popover that enables/disables the MathQuill editor toolbar is activated by the contextmenu event.  That means that the standard browser ways of opening the context menu will work.  The mouse can do this with a right click.  The keyboard can do this with Shift-F10 (in Firefox or Chrome -- not sure on other browsers).  Some keyboards have a context menu button (mine does) that will open this as well.
* Popover added to DOM in wrong place:  I will see what I can do about this.  The `container` option needs to be set, but finding a reliable container to use will be the challenge.
* For essay answers, what you describe is the way the current attempts table handles essay answers.  That is not changed.  I also don't think that anything more can actually be done with this.

drgrice1:
@Alex-Jordan: I added a few tweaks based on your comments.

The padding on the card headers in the popover is reduced some.

The popover should now be in the DOM immediately after the feedback button in most cases (all cases I know of, but if someone places the feedback button in a different place it may not work).

I made clicking on the popover header close it (assuming the popover has a header).  I realized that focus and interactive semantics are probably not needed or even desirable in this case.  Keyboard users can already open and close the popover, and if the header is focusable, it would probably just be an annoying tab stop.

Alex-Jordan (approved):
I did not review the code changes, but I did test the new way of giving feedback (at https://github.com/openwebwork/pg/pull/939) and it all works very nicely as far as I can tell.

PR #2230, opened 2023-10-15
Improved file manager button states and other file manager improvements.

drgrice1:
The buttons that can only act on one file are now disabled when multiple files are selected. Only the "Delete" and "Make Archive" buttons can act on multiples files.
    
The type of file (regular, directory, or link) is stored in a data attribute for each option in the `files` select.  The javascript also takes this into consideration and is disabled for a file type a button can not act on.  The "View", "Edit", "Download", and "Copy" buttons can not act on symbolic links.  The "Edit", "Download", and "Copy" buttons can not act on directories.

Editing a pg file from the file manager opens in the pg problem editor.

There is also an edit button added to the file manager file viewer when viewing a text file.

Also make the download link on that page a button.

A toast message is shown instead of submitting when a non-viewable file is double clicked.

### DIFF
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -12,7 +12,47 @@
 
 		document.getElementsByName('directory')[0]?.addEventListener('change', () => doAction('Go'));
 		document.getElementsByName('dates')[0]?.addEventListener('click', () => doAction('Refresh'));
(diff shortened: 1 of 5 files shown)


Alex-Jordan (approved):
This is great. I tested all of the things mentioned and they work as described.

drdrew42 (approved):
👍

PR #2232, opened 2023-10-19
Add a page to manage jobs in the job queue.

drgrice1:
All jobs for a course are listed on this page. The table displays the job id (this is used to reference specific jobs in action messages, otherwise it would not be shown), task name, created time, started time, finished time, and state.  Also a button that opens a popover containing the job result is in the state column if the job has completed.

Note that the Minion job queue automatically removes jobs from the job queue after two days (that is the default at least which we don't change).  So the real importance of this page is to allow the instructor to see the status of recently completed or in progress jobs.

At this point the actions available on the page are filter, sort, and delete.  Jobs can be filtered by id, task name, or state. Jobs can be sorted by clicking on the headers, or by using the sort form.  Jobs that are not active can be deleted.

Minion does not allow deletion of active jobs.  Note that an active job means a job that is currently running. As such they can not be selected on this page. Perhaps an option to stop running jobs could be added at if there is a problem with jobs hanging, but active jobs can not be directly stopped. The Minion worker is in a different process so the Mojolicious app needs to broadcast a signal to the Minion worker to do so.

An inactive job (i.e., a job that has been queued but has not started running yet) can be selected and deleted.  However, it is possible that the inactive job could start before the form is submitted.  In that case the job can not be deleted, and so an alert will show that.

In order to reliably associate a course with a job there is a new rule for tasks.  The job must pass the course id via the "notes" option of the Minion enqueue method.  The existing tasks have been updated to do this. There is also a backwards compatibility check to find jobs that passed it one of the ways the two jobs did it before in the job arguments.

Since the job fail/finish messages are now displayed in the UI, those messages are now translated. That is all except the first few messages in each task before the course environment is established, since a course environment is required to obtain the language of the course.

The send_instructor_email task no longer sends an email to the instructor after sending the emails to the students.  Instead the job result contains all of the information that would have been in that email. This is a far more reliable way of getting that information to the instructor sending the email.  The instructor just needs to go to the "Job Manager" page to see the result.  The message on the "Email" page tells the instructor this.

This page is also available for the admin course.  In the admin course all jobs for all courses are shown.  There is an additional column in the jobs table that shows the course id for the course the job was enqueued by.

Also, the errors that are reported when sending emails are made less verbose by calling the `message` method of the Mojo::Exception which does not include the traceback.

### DIFF
--- a/htdocs/js/JobManager/jobmanager.js
+++ b/htdocs/js/JobManager/jobmanager.js
@@ -1,0 +1,47 @@
+(() => {
+	// Show/hide the filter elements depending on if the field matching option is selected.
+	const filter_select = document.getElementById('filter_select');
+	const filter_elements = document.getElementById('filter_elements');
+	if (filter_select && filter_elements) {
+		const toggle_filter_elements = () => {
+			if (filter_select.value === 'match_regex') filter_elements.style.display = 'block';
+			else filter_elements.style.display = 'none';
+		};
+		filter_select.addEventListener('change', toggle_filter_elements);
+		toggle_filter_elements();
+	}
+
+	// Submit the job list form when a sort header is clicked or enter or space is pressed when it has focus.
+	const currentAction = document.getElementById('current_action');
+	if (currentAction) {
+		for (const header of document.querySelectorAll('.sort-header')) {
+			const submitSortMethod = (e) => {
+				e.preventDefault();
+
+				currentAction.value = 'sort';
+
+				const sortInput = document.createElement('input');
+				sortInput.name = 'labelSortMethod';
+				sortInput.value = header.dataset.sortField;
+				sortInput.type = 'hidden';
+				currentAction.form.append(sortInput);
+
+				currentAction.form.submit();
+			};
+
+			header.addEventListener('click', submitSortMethod);
+			header.addEventListener('keydown', (e) => {
+				if (e.key === ' ' || e.key === 'Enter') submitSortMethod(e);
+			});
+		}
+	}
+
+	// Activate the results popovers.
+	document.querySelectorAll('.result-popover-btn').forEach((popoverBtn) => {
+		new bootstrap.Popover(popoverBtn, {
+			trigger: 'hover focus',
+			customClass: 'job-queue-result-popover',
+			html: true
+		});
+	});
+})();
--- a/htdocs/js/JobManager/jobmanager.js
+++ b/htdocs/js/JobManager/jobmanager.js
@@ -11,31 +11,6 @@
 		toggle_filter_elements();
 	}
 
-	// Submit the job list form when a sort header is clicked or enter or space is pressed when it has focus.
-	const currentAction = document.getElementById('current_action');
-	if (currentAction) {
-		for (const header of document.querySelectorAll('.sort-header')) {
-			const submitSortMethod = (e) => {
-				e.preventDefault();
-
-				currentAction.value = 'sort';
-
-				const sortInput = document.createElement('input');
-				sortInput.name = 'labelSortMethod';
-				sortInput.value = header.dataset.sortField;
-				sortInput.type = 'hidden';
-				currentAction.form.append(sortInput);
-
-				currentAction.form.submit();
-			};
-
-			header.addEventListener('click', submitSortMethod);
-			header.addEventListener('keydown', (e) => {
-				if (e.key === ' ' || e.key === 'Enter') submitSortMethod(e);
-			});
-		}
-	}
-
 	// Activate the results popovers.
 	document.querySelectorAll('.result-popover-btn').forEach((popoverBtn) => {
 		new bootstrap.Popover(popoverBtn, {
--- a/htdocs/js/JobManager/jobmanager.js
+++ b/htdocs/js/JobManager/jobmanager.js
@@ -10,13 +10,4 @@
 		filter_select.addEventListener('change', toggle_filter_elements);
 		toggle_filter_elements();
 	}
-
-	// Activate the results popovers.
-	document.querySelectorAll('.result-popover-btn').forEach((popoverBtn) => {
-		new bootstrap.Popover(popoverBtn, {
-			trigger: 'hover focus',
-			customClass: 'job-queue-result-popover',
-			html: true
-		});
-	});
 })();
--- a/htdocs/js/JobManager/jobmanager.scss
+++ b/htdocs/js/JobManager/jobmanager.scss
@@ -1,8 +1,0 @@
-.job-queue-result-popover {
-	--bs-popover-max-width: 500px;
-
-	.popover-body {
-		overflow-y: auto;
-		max-height: 25vh;
-	}
-}
--- a/htdocs/js/JobManager/jobmanager.scss
+++ b/htdocs/js/JobManager/jobmanager.scss
@@ -1,0 +1,8 @@
+.job-queue-result-popover {
+	--bs-popover-max-width: 500px;
+
+	.popover-body {
+		overflow-y: auto;
+		max-height: 25vh;
+	}
+}
--- a/lib/Mojolicious/WeBWorK.pm
+++ b/lib/Mojolicious/WeBWorK.pm
@@ -26,7 +26,7 @@ use Env qw(WEBWORK_SERVER_ADMIN);
 
 use WeBWorK;
 use WeBWorK::CourseEnvironment;
-use WeBWorK::Utils qw(x writeTimingLogEntry);
+use WeBWorK::Utils qw(writeTimingLogEntry);
 use WeBWorK::Utils::Routes qw(setup_content_generator_routes);
 
 sub startup ($app) {
@@ -84,7 +84,8 @@ sub startup ($app) {
 	# Add the themes directory to the template search paths.
 	push(@{ $app->renderer->paths }, $ce->{webworkDirs}{themes});
 
-	# Setup the Minion job queue.
+	# Setup the Minion job queue. Make sure that any task added here is represented in the TASK_NAMES hash in
+	# WeBWorK::ContentGenerator::Instructor::JobManager.
 	$app->plugin(Minion => { $ce->{job_queue}{backend} => $ce->{job_queue}{database_dsn} });
 	$app->minion->add_task(lti_mass_update        => 'Mojolicious::WeBWorK::Tasks::LTIMassUpdate');
 	$app->minion->add_task(send_instructor_email  => 'Mojolicious::WeBWorK::Tasks::SendInstructorEmail');
--- a/lib/Mojolicious/WeBWorK/Tasks/AchievementNotification.pm
+++ b/lib/Mojolicious/WeBWorK/Tasks/AchievementNotification.pm
@@ -31,30 +31,31 @@ use WeBWorK::SafeTemplate;
 
 # send student notification that they have earned an achievement
 sub run ($job, $mail_data) {
-	my $ce = eval { WeBWorK::CourseEnvironment->new({ courseName => $mail_data->{courseName} }); };
-	return $job->fail("Could not construct course environment for $mail_data->{courseName}.")
+	my $courseID = $job->info->{notes}{courseID};
+
+	my $ce = eval { WeBWorK::CourseEnvironment->new({ courseName => $courseID }); };
+	return $job->fail("Could not construct course environment for $courseID.")
 		unless $ce;
 
-	$job->{language_handle} =
-		WeBWorK::Localize::getLoc($ce->{language} || 'en');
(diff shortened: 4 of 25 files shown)


somiaj:
I like this idea, trying to think of a good way to test this, as the only place I use the queue is on my live server. Is there a way I can copy the queue from a live server just to check out the page?

drgrice1:
You could copy the `DATA/webwork2_job_queue.db` file from your server to your test server.  Archive a course that uses the job queue, and also copy the archived course file to your test server.  Then open the Job Manager page in that course on your test server.

You can also just run some jobs in a course on your test server by configuring LTI grade passback (but with invalid credentials or even an invalid LMS to begin with).  The jobs will still run.  Furthermore the failure will show up in the Job Manager.   You can do the same with sending emails.  Even if you don't have smtp configured, the job will run, and the errors will be shown in the Job Manager.

drgrice1:
@somiaj: By the way, if you copy the sqlite db file and a course archive, make sure you either disable grade pass back for the course, or don't have the job queue running when you test this.  You don't want to pass back grades from your test server. Note that you don't need the job queue running to test this pull request. It doesn't use the job queue worker process, just the database.

somiaj:
@drgrice1 thanks for the warning/remind, I always just make the secret invalid and use a fake url on my dev system, this should give me lots of errors to see in progress. What are the webwork2_job_queue.db-{shm,wal} for?

drgrice1:
I am not entirely sure. They are used internally by sqlite.  I suspect sqlite uses them for caching and locking and such. You only need the `.db` file though.

somiaj:
Here are the results of my tests. I like the new page, though even though all of my jobs should have all failed, they didn't (though maybe the errors of trying to send to localhost with no response wasn't caught). But instead some of my lti mass updates finished and others failed due to what appears to be a bug.

This is the message I get on the updates that failed. Note I could not find a way to actually copy/paste any of the info in the info popup, so I had to take a screenshot.

image

Unsure if this bug is new with this PR, or something that has been hiding.

somiaj:
The updates that are failing are for a single user and all sets.

drgrice1:
There was a mistake on line 79 of LTIMassUpdate.pm.  It has been fixed.

Note that the LTI update job will often succeed even if it can not actually send grades.  The actual update method in WeBWorK::Authen::LTIAdvanced::SubmitGrade (or WeBWorK::Authen::LTIAdvantage::SubmitGrade) can fail and the job still succeed.  For those failures you will need to check the logs/webwork2.log file.  Job success only means that the submit_course_grade method was called and returned.

somiaj:
Now I get this error when restarting webwork2.service

```
Oct 22 20:10:51 webwork hypnotoad[1998]: Can't load application from file "/opt/webwork/webwork2-develop/bin/webwork2": Version control conflict marker at /opt/webwork/webwork2-develop/lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm line 79, near "<<<<<<<"
Oct 22 20:10:51 webwork hypnotoad[1998]: Version control conflict marker at /opt/webwork/webwork2-develop/lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm line 81, near "======="
Oct 22 20:10:51 webwork hypnotoad[1998]: Version control conflict marker at /opt/webwork/webwork2-develop/lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm line 83, near ">>>>>>>"
```

somiaj:
grr, nevermind...my error --- I did a pull instead of reset.

somiaj:
Thanks, this is working fine now.  Would it be possible to get errors from `submit_course_grade`? I could see this confusing a user who sees that the update job has finished sucessfully, but yet when they check their LMS, the grades haven't been updated. In general it should be fine, but it could help users who don't have access to the logs if something goes wrong.

somiaj (approved):
Nice addition.

drgrice1:
I plan to do more work on the LTI mass update task to make failures come through to this page.  The challenge is that the same code is used for when students submit answers. So care is needed to get messages out of the task without interfering with that. I have some ideas though.

I think that is for another pull request though.

As to sending emails, you must have localhost set up to accept emails in some way. Otherwise, the message would show up on this page.

somiaj:
Actually, I don't see any records of my updates failing in `logs/webwork2.log`, and enabling `$debug_lti_grade_passback` doesn't give any info about the jobs failing either. So I can't seem to find any evidence of LTI update failures from the webwork size, as it seems to think they will always be successful.

somiaj:
@drgrice1 thanks for the info, yea that can be in another pull request, as this isn't changing the fact that info about failed LTI updates doesn't seem to currently be logged anywhere.

drgrice1:
Are you sure the users have LTI pass back info? If not there is no failure.  There is nowhere to pass grades back to, so the users will be quietly skipped.  To have pass back info the user must have accessed an assignment from the LMS.

somiaj:
They should, this is an archive of a live course that grade pass back is currently working, that I made right before doing all these tests.  The only thing I changed was to make the secret equal to `invalid` and the url `https://localhost` so any attempt to pass the grades back would fail.

drgrice1:
I suspect there is some issue with your test setup.  Check permissions on the webwork2.log file.  Make sure that the job queue and the webwork2 app are both running as the same user, and that both have permission to write to the file.

I am certain that the log is written to by the job.

drgrice1:
Also, on your test server are you running the minion worker in production mode?  If not, then it will not write to the webwork2.log file.  Instead it will show the messages in the terminal that you are running the worker in.

somiaj:
Yea, I figured it out, I looked at the logs for the service `webwork2-job-queue.service` and discovered I didn't have the LTI setup correctly, so an undefined variable `$ce->{LTIVersion}` was either failing or assuming I was using LTI 1.3, and things weren't working as a result. The job is now running, but taking a long time to finish, thanks for working though this with me. In any case, this PR seems ready to go, and issues are on my end now.

drgrice1:
I added a tweak to the LTIMassUpdate.pm task.  Now all logs are captured by the job and reported in the result.  The jobs will still almost always succeed, but the messages will hold any logs that the grade submission methods send.  Note that if `debug_lti_grade_passback` is true then this will be a rather extensive report.  If `debug_lti_parameters` is also true it will be even more extensive.

somiaj:
Thanks,

Now I can see the errors, and when I set the debugging variables, I get full details on the errors in both the jobs page and the webwork log file.

I do notice that when the request is correctly sent to the LTI and returned, that the returned JSON is `{"errors":[{"message":"text"}]}`. I don't know if this is standard in LTI or just something with canvas. But it makes me wonder if you can capture any returned errors (even when debugging is turned off), to help users if there are any errors in the request.

drgrice1:
I don't know what you mean by the return JSON.  The requests to and from the LMS are not JSON requests, so that doesn't really make sense.

Unfortunately, you can't capture errors if debugging is turned off.

somiaj:
It just looked like the body of the response was JSON, and the header was `Content-Type: application/json;`. Here is the more of the response from the log (I removed most of the headers  as I don't think it is needed and might have sensitive info in it):

```
[2023-10-23 08:56:39.40426] [4250] [warn] The response is:
 HTTP/1.1 401 Unauthorized
Cache-Control: no-cache
Connection: keep-alive
Content-Length: 55
Content-Type: application/json; charset=utf-8
Date: Mon, 23 Oct 2023 14:56:39 GMT

{"errors":[{"message":"Invalid authorization header"}]}
```

somiaj:
This is the response I see in `logs/webwork2.log` file. The two lines before it were about the submit (I removed the exact strings being used).

```
[2023-10-23 08:56:39.10911] [4250] [warn] Submitting grade for user USERNAME and set SETNAME.
[2023-10-23 08:56:39.11224] [4250] [warn] Submitting grade using sourcedid: SOURCEDID and score: SCORE
```

drgrice1:
I think part of the problem here is that you are not seeing valid data from the real requests that would occur on your test server setup.  That is giving you misleading information on what actually happens.  If you were using an actual server you would only get those messages while setting LTI grade passback initially and using debugging.  Once you have things set up, you won't get those anymore.

somiaj:
This should be a valid response, as the request was being sent to my canvas server. I forgot to change the url in the database, so grade_passback requests were being sent to my canvas server, but since the secret was invalid, the request was being rejected by canvas due to an invalid authorization, and this was the response I got from canvas.

Once I updated the url in the database to not send requests to my Canvas server, the messages changed, and I no longer got any response. I just saw this response from Canvas and though it could be useful to parse out the error from the JSON if it was there. But if this isn't something that will happen on a system that is setup correctly, then not needed.

drgrice1:
But that is not a valid response at this point.  You have invalidated it by changing the secret.  That is something that you would be doing when setting up the server with debugging active.

drgrice1:
I added another commit that makes the report for LTI grade passback more thorough.  It is perhaps a tad experimental because it may be too thorough.  Basically, now you will see a report for each user and set that is updated in the job queue result.  This commit can be reverted (or tweaked) if this is too much.  It is perhaps not that different from the instructor email report though.  For that job you also see a report for each email sent.

drgrice1:
Note that a downside (or upside depending on your perspective) is that when mass grade passback occurs (either automated or initiated from the "LTI Grade Update" page) all of these messages will also always be sent to the app log file.

somiaj:
Looks good. Would it be reasonable to have a configuration option in site.conf to let admins decide how much detail is being logged and thus shown on this page?

In regards to the info popup with the logged data, I am not able to select/copy any data from it, and when there is lots of output, I wish could view it in a bigger window. Could javascript be used to open up the data from the div in its own window/tab via a link/button inside the info popup?

drgrice1:
In general, a pop-up should not have interactive elements like a link or button in it, and thus certainly a bad case for something that opens in a new window.

There are two options.

The trigger could be switched from the current trigger of "hover focus" to "click".  With the "hover" trigger the pop-up is open while the mouse hovers over the button, and with the "focus" trigger it is open when the button has focus. So if you click in the pop-up itself the button loses focus and the mouse is not hovering over the button, and the pop-up closes. With the "click" trigger the pop-up opens and closes exactly when you click on the button. Meaning you can click in the pop-up, and it will stay open. This option will not make the contents bigger though.

The other option is to switch to a modal dialog. That can be bigger, and you can select it's contents.

I like the "hover focus" trigger with a pop-up for this because it is less intrusive. When you click somewhere else the pop-up goes away.  I don't really understand the desire to select/copy the contents, and the pop-up is scrollable so you can see all contents. My goal with this was to have something instructors can see that shows what jobs have recently run, with a brief synopsis.  The important thing to me is just to show that jobs have run.

I don't like how verbose the messages now are. This was not intended to be an LTI debugging page.

somiaj:
I like to select things when I read them to help direct my focus when there is lots of scrolling, and if I needed to share the info, it would be nice to copy it. Though this might just be something I see in testing the feature and not something that would actually be needed, so I'm okay with the way it is, unless it can nicely be implemented.

Under normal circumstances the verbosity won't be needed, so maybe having a debugging option to turn it on/off would be best, that way the page mostly functions as just a log of the recent/active/future jobs.  But I could see the extra logging being useful in some case to help debug what the job-queue is actually doing in cases that things aren't functioning properly.

Now that I understand what 'finished' and 'failed' actually mean, I'm okay with how it was before of not having to much detail. The info is about the state of the job-queue, not if there were any issues with the grade passback of the job itself.

drgrice1:
@somiaj: I switched to a modal dialog for showing the results.  This should do what you want.

somiaj:
I do like the modal a bit better if you decide to keep all the details logged, thanks. Though if you do decide to revert it so it only gives a few lines of info, the modal won't be needed.  I could see that info being helpful in rare situations to debug the jobs, but in most cases probably not needed.

Probably best to see if anyone else has any opinions on how much info should be shared via this page, but overall it gives a nice record of grade updates and emails sent via the job queue.

One thing I noticed is 'failed' jobs seem to stick around a lot longer than the other jobs. All my 'finished' jobs from last testing this on Oct 22nd are no longer available, but the 'failed' ones are still there.  Looking at the admin course, I see some failed jobs from other courses back in may still present (from when I was having trouble with the lti update then). So it seems that failed jobs aren't removed from the database automatically like finished jobs are.

drgrice1:
Yes, "failed" jobs are not removed from the database automatically.  Only jobs that have transitioned to the "finished" state are automatically removed.

drgrice1:
The achievement email notification task is updated with the changes needed to work with this.  So this is ready for further review.

Alex-Jordan (approved):
I just did basic testing with this and I don't have anything to report. I did not do more thorough testing involving copying stuff from production over to my test server. 

This seems like a good new feature, where perhaps, maybe if something was overlooked and not working as intended, that would come out if we merge this now and try using it more before 2.19 is released.

PR #2235, opened 2023-10-22
When "tags written" is displayed, make it a good (green) message.

pstaabp:
Currently, when saving a tag on a problem, the toast message "tags written" shows up in red (bootstrap alert). This gives the toast message in green.


drgrice1 (approved):
Good catch ... or is writing tags a bad thing!

PR #2236, opened 2023-10-22
Rename Hwmk Sets Editor and Classlist Editor

drgrice1:
This changes the names in the site nav as discussed in https://github.com/openwebwork/webwork2/issues/1992.


Alex-Jordan:
Do we want to have "Assignment Manager" but right above it we have "Set Assigner"? What I mean is, referring to a set as an "Assignment" in one place, and as a "Set" right next to it.

somiaj:
We also have 'Homework sets', and 'sets' gets used to refer to. Maybe just call it 'Set Manager' to be consistent with what they are referred to elsewhere. Advantage of this is a shorter name.

drgrice1:
@Alex-Jordan: Yeah, that is a good question.  I have had this branch sitting here for a while (since shortly after @dlglin made that issue), and decided to put it out there to further the discussion.

I like "Set Manager".  I will wait for a bit before making the change to see what others have to say.

drgrice1:
I went ahead and made the change from "Assignment Manager" to "Set Manager".

somiaj (approved):
Lke the Chnge

pstaabp:
Also, just searching "Hmwk" there are a number of translations (in webwork2.pot) that include this.  Do we need to update this, or is it sufficient to wait until release time which will then update. 

Lastly, there are a couple of comments in defaults.config with "hmwk" that should probably be updated as well.

drgrice1:
I fixed the things you noticed.  Thanks for pointing those out.

The `po` files should not be modified directly.  At some point we will run the `bin/dev_scripts/update-localization-files` script, and commit the changes it makes.  It will take care of those things.

pstaabp (approved):

drdrew42 (approved):
👍 -- although the branch name had me banging my head 💀

Alex-Jordan:
Three approvals, plus I presume Glenn's.

PR #2237, opened 2023-10-30
check_modules.pl allow newer versions of node.

drgrice1:
Allow any version of node that is 16 or newer.  Don't insist on version 16 exactly.  I am currently using node 20 and was using 18 before that. I haven't seen any issues with these newer versions.


drdrew42 (approved):
Same. Although we will have to be careful if the dev team migrates past v16 and doesn't catch the need for v18 or v20 in the future.

drgrice1:
That is true. So we will need to monitor past versions of node and increment the minimum required version as needed. At this point version 16 still works.

PR #2241, opened 2023-11-05
some redesign of PGML lab

Alex-Jordan:
This is a start to redoing the PGML lab examples. Of course it should get the usual review with possible edits, but for some things that are not yet perfect, I think that subsequent community PRs could improve it. (The whole thing could be done differently, but for now this is using most of the code already here.)

To test, you need to get the new PGML-lab.pg file into a course, as well as the image file here that is used in one example's image demo. Then go to the problem editor and click the PGML button. I don't think you need to check out this branch; you can just put these things in the right place in an existing course to try.

* I changed the layout. I don't like the rendered content inserting itself up high, pushing the input fields down off screen. So now the rendered content is below.
* Also the select menus are now adjacent to the input textareas, as long as the screen is wide enough. And I moved them first so that in a narrow screen, the input textareas are still as close as they can be to the rendered output.
* The "Examples" are mostly revised, except I didn't touch the "Problems" category. I added "Images" and "Tables" examples. I condensed and reorganized some existing examples.
* I use the perl indented heredoc construct (from 5.24 onward) to write the sample code into the file. It's more human readable this way. The cost is that some lines in this file are nothing but a bunch of tabs.
* When you select an "Example", the other selects all reset. I found it less confusing this way.
* I haven't (yet) touched the content of the "Reference" selects, but I did change the options to be disabled. I found it confusing that you would select one of those and nothing would happen. Eventually I figured out they are just for looking at, and making them disabled helps make that more clear. But maybe it's odd we use a select for that purpose in the first place.
* I made the input textareas use monspace font. If we could have syntax highlighting here, that would be nice.


Alex-Jordan:
The diff here in GitHub at line 142 (and many other places) shows a completely blank line. But actually lines like that have 9 tab characters. This is needed for the heredoc construction.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -221,496 +1033,43 @@ window.addEventListener('DOMContentLoaded', () => {
 				el.value = unescapeHTML(selectedExample.dataset[id]);
 				el.rows = Math.max(dataType[id], selectedExample.dataset[id].split(/\n/).length);
 			}
+			for (const otherselect of document.querySelectorAll('.example-selector')) {
+				if (otherselect != select) {
+					otherselect.value = otherselect.firstChild.value;
+				}
+			}
 			document.getElementById('resultsBox').style.display = 'none';
 			window.scrollTo(0, 0);
```
drgrice1, line 1042: I think that with the layout change and the textareas and rendering below, this line should be deleted.  It is a bit annoying when you select something in one of the example dropdowns, and it scrolls to the top of the page.  This is most noticeable on a narrow screen.

drgrice1 (commented):
This is just some quick initial observations.

Please run the perltidy-pg.pl script on the file.  If you want to, you can use a custom perltidy-pg.rc file with `-l=120` (allowing 120 character long lines).

Alex-Jordan:
Are we able to use it on PG files? Or is this one special because it does
not have BEGIN_TEXT or BEGIN_PGML blocks?

On Sun, Nov 5, 2023 at 3:48 PM Glenn Rice ***@***.***> wrote:

> ***@***.**** commented on this pull request.
>
> This is just some quick initial observations.
>
> Please run the perltidy-pg.pl script on the file. If you want to, you can
> use a custom perltidy-pg.rc file with -l=120 (allowing 120 character long
> lines).
> ------------------------------
>
> In courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
> <https://protect2.fireeye.com/v1/url?k=31323334-501d2dca-3132feb7-454455534531-6e7a9f87f19a30c9&q=1&e=273e36d3-1b9f-46ac-b8ea-304b2c0b882c&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fwebwork2%2Fpull%2F2241%23discussion_r1382673617>
> :
>
> > @@ -221,496 +1033,43 @@ window.addEventListener('DOMContentLoaded', () => {
>  				el.value = unescapeHTML(selectedExample.dataset[id]);
>  				el.rows = Math.max(dataType[id], selectedExample.dataset[id].split(/\n/).length);
>  			}
> +			for (const otherselect of document.querySelectorAll('.example-selector')) {
> +				if (otherselect != select) {
> +					otherselect.value = otherselect.firstChild.value;
> +				}
> +			}
>  			document.getElementById('resultsBox').style.display = 'none';
>  			window.scrollTo(0, 0);
>
> I think that with the layout change and the textareas and rendering below,
> this line should be deleted. It is a bit annoying when you select something
> in one of the example dropdowns, and it scrolls to the top of the page.
> This is most noticeable on a narrow screen.
>
> —
> <https://protect2.fireeye.com/v1/url?k=31323334-501d2dca-3132feb7-454455534531-8d63eeef80b9b347&q=1&e=273e36d3-1b9f-46ac-b8ea-304b2c0b882c&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fwebwork2%2Fpull%2F2241%23pullrequestreview-1714092792>,
> or unsubscribe
> <https://protect2.fireeye.com/v1/url?k=31323334-501d2dca-3132feb7-454455534531-efa3ee4ac08991c0&q=1&e=273e36d3-1b9f-46ac-b8ea-304b2c0b882c&u=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABEDOAHTZYJDHNLMZ56GSNLYDAJTHAVCNFSM6AAAAAA66U7NNKVHI2DSMVQWIX3LMV43YUDVNRWFEZLROVSXG5CSMV3GSZLXHMYTOMJUGA4TENZZGI>
> .
> ***@***.***>
>


-- 
Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
The perltidy-pg.pl script is for PG files, not for the PG code (run-perltidy.pl is for that).  It is basically what the tidy code option in the problem editor uses.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
Alex-Jordan: OK, I removed that.

Alex-Jordan:
Oh I didn't catch the distinction. I ran `run-perltidy.pl` on this file and the changes looked OK. I supposed I should go see if it still works...

drgrice1:
Since this file doesn't have any BEGIN_TEXT/END_TEXT or BEGIN_PGML/END_PGML blocks, it will actually be pretty much the same.

Alex-Jordan:
Still working.

I just increased the heights of the two textareas, which is a good use of space given that now there are two more categories over in the Examples selects.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -221,496 +1048,42 @@ window.addEventListener('DOMContentLoaded', () => {
 				el.value = unescapeHTML(selectedExample.dataset[id]);
 				el.rows = Math.max(dataType[id], selectedExample.dataset[id].split(/\n/).length);
```
drgrice1, line 1049: This is the problem that led me to ask for taller textareas.  This line should be
```suggestion
				el.rows = Math.max(dataType[id], selectedExample.dataset[id].split(/~~n/).length);
```
The translator is turning the single backslash into two as usual.  So that code wasn't working.  The textareas should be resized to fit the content, but this is making that not work.

drgrice1:
Sorry, I should have noticed this when I made my last comment.  The values on line 1045 should be updated to match the new textarea defaults for the rows (3 and 10 now).

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -22,8 +22,11 @@ sub Menu {
 	return tag(
 		'select',
 		aria_labelledby => 'reference-label',
-		style           => 'width:15em',
-		join('', map { tag('option', $_) } @_)
+		style           => 'width:11em',
+		value           => (@_)[0],
```
drgrice1, line 26: I am not sure what you were going for in adding this line?  The `value` attribute is not valid for a `select` tag. So this line will need to be deleted.

Alex-Jordan:
Fixed that, thanks.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -36,8 +39,8 @@ sub Examples {
 		class           => 'example-selector',
 		id              => $title,
```
drgrice1, line 40: Since some of the new titles that you added have spaces in them, this will need to be changed to
```suggestion
		id              => $title =~ s/ /_/gr,
```
Spaces are not valid in html `id`s.

drgrice1 (commented):
I see a couple of html validation issues.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
drgrice1: Or you could change spaces to dashes (`-`).  Take your pick.

drgrice1 (commented):
There are a couple of issues I noticed when I used the "Process this text" button with some examples.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -22,8 +22,11 @@ sub Menu {
 	return tag(
 		'select',
 		aria_labelledby => 'reference-label',
-		style           => 'width:15em',
-		join('', map { tag('option', $_) } @_)
+		style           => 'width:11em',
+		value           => (@_)[0],
+		join('',
+			tag('option', selected => 'selected', shift),
+			map { tag('option', disabled => 'disabled', value => $_, $_) } @_)
```
drgrice1, line 29: The `selected` and `disabled` attributes for an `option` are boolean attributes.  This means that they do not have a value.  So in html this would look like `<option disabled value="option value">Option label</option>`.  To do this with the `tag` method you use `disabled => undef` or `selected => undef`.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
Alex-Jordan: Hmm, I think it is a remnant from an early attempt at part of this. Removing it.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
Alex-Jordan: Done.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -104,98 +106,923 @@ if ($pgml ne '') {
...
+									The formula is [_]{'1+x'}{16}.
+									
+									The argument should be a string that is understood by Compute() in the current context.
+									ENDPGML
+							]
+						],
+						[
+							'PG variable' => [
+								<<~'ENDPG',
+									Formula('1+x^2');
```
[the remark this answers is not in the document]
Alex-Jordan: Thanks. I think this and the next one were casualties of the way I converted to use the heredoc.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
Alex-Jordan: Thanks. I knew that was the better way to write them in HTML, but I didn't know or look up how the `tag()` tool would work for that.

drgrice1: I didn't document how to use boolean attributes for the `tag` method.  This is how the Mojolicious::Plugin::TagHelpers `tag` methods works (or really Mojo::DOM `new_tag` method which it calls), and the PGbasicmarcos.pl `tag` method is designed to work the same.

drgrice1:
I have to admit that I did not know about indented here-docs.  That is a very useful perl feature.

Alex-Jordan:
Yeah I learned about that this morning wrestling with these examples. I Googled for a better way and the Google giveth.

Alex-Jordan:
I forgot to include one thing in the summary here. I removed some code that handled the hardcopy output for this PG file. I couldn't see why this file would be processed for hardcopy output. Maybe there is a historical reason. But if there is a reason to keep that, let me know and I can work to bring it back.

pstaabp:
Overall, this is a great improvement over the previous version.  Looking over everything, Some of the answer blanks have quite a few underscores.  For example, line 905 has 36 underscores.  Without mathquill, I'm guessing this would be a huge text input.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -104,98 +105,986 @@ if ($pgml ne '') {
...
+							Bulleted => [
+								'',
+								<<~'ENDPGML'
+									Bullet items can be indicated with a star:
+									* Apple
+									* Banana
+									
+									Or with a plus:
+									+ Apple
+									+ Banana
```
pstaabp, line 351: Also 
```
or with dashes
- item 1
- item 2
```

pstaabp: Never mind.  I see it below.

pstaabp:
I want to add to make sure that if you already have a PGML-lab.pg file in your templates directory of your course, you'll need to replace that with the one in the courses.dist in this PR.  I was reviewing the old version. ☹️

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -104,98 +105,986 @@ if ($pgml ne '') {
...
+									
+									The answer may be declared outside of the PGML markup, by passing ANS() a cmp() routine.
+									ENDPGML
+							]
+						],
+						[
+							'Old style' => [
+								'',
+								<<~'ENDPGML'
+									Twelve is [_]{num_cmp(12)}{16}.
```
pstaabp, line 306: Even though this still works, I think we're trying to discourage this.  I would remove this example.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
```diff
@@ -104,98 +105,986 @@ if ($pgml ne '') {
...
+									[#
+										[. A .]
+										[. B .]*{headerrow => 1}
+										[. C .]
+										[. D .]*
+										[. E .]
+										[. F .]
+									#]{padding => [0,1]}
+									
+									These are all the options. For details about how they work, see the niceTables.pl documentation.
```
pstaabp, line 784: It would be nice to have this link to the POD for niceTables.pl.  

Is there a was to do links in PGML?  I know of the `htmlLink` command.

pstaabp:
Overall, this looks great.

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
Alex-Jordan: No objection from me. Does anyone want to keep this?

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
Alex-Jordan: In PGML code the construct `[< ... >]` is reserved for links. But it was never implemented.

Alex-Jordan:
> For example, line 905 has 36 underscores.

So, that's part of the examples that I did not change. I did not want to get into best overall practices that might be controversial. Also I didn't want to think about what kinds of complete problems would be best for these "exemplar" exercises. So I just left that block of examples as it was before. Other changes are making it look like a new addition in the diff.

Personally I always use the `[_]{answer}{width}` construction and I only left multiple underscores in the first example of answer blanks (among examples I edited).

Alex-Jordan:
I merged this into production to test it. I created a quiz with unlimited graded submissions, requiring proctor to start and grade. But when I used my credentials to start the quiz, I got an error screen with 
```
Too few arguments for subroutine 'WeBWorK::ContentGenerator::GatewayQuiz::can_recordAnswers' at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/LoginProctor.pm line 68. 
```

courses.dist/modelCourse/templates/PGMLLab/PGML-lab.pg
drgrice1: I am fine with removing it.  I agree that the usage of `num_cmp` should be discouraged.

Alex-Jordan:
I just rebased off develop and removed the example with `num_cmp`.

drgrice1:
> I forgot to include one thing in the summary here. I removed some code that handled the hardcopy output for this PG file. I couldn't see why this file would be processed for hardcopy output. Maybe there is a historical reason. But if there is a reason to keep that, let me know and I can work to bring it back.

I am fine with dropping support for hardcopy output of this file.  It doesn't output anything useful anyway.  However, if that is the case, then I think the pretense of this working for hardcopy output should also be removed.  In other words take out the remaining `MODES` calls and just directly insert the `HTML` mode text.

drgrice1:
The `if ($displayMode ne 'TeX') {` on line 1080 (and the ending brace for that block) can also be removed.  Of course there is special handling for outputting the `LaTeX` code on lines 89-106 that should not be removed.

Alex-Jordan:
There is an issue with the PGML lab problem that exists even with the develop version. I would like to try and fix it here with this PR that is otherwise revising the PGML lab problem file. But I need to reach out for help.

To see the issue, whether with the develop PGML-lab.pg file or the one from this branch, select an example for math markup. Check the box to show TeX code, and process the problem. The TeX code that you will see is not correct. In places, it used `HTML_MathJax` as the displayMode instead of using `TeX` like it was supposed to. For example, in the develop branch version with the first math markup example, I get:

```
{\pgmlSetup
In-line math: <script type="math/tex">\frac{x+1}{x-1}</script>, display math: <script type="math/tex">\displaystyle{\frac{x+1}{x-1}}</script>
\vskip\baselineskip
{\pgmlIndent
<script type="math/tex">\displaystyle{\frac{x+1}{x-1}}</script> (indented)
\par}%
\vskip\baselineskip
{\pgmlCenter{}<script type="math/tex">\displaystyle{\frac{x+1}{x-1}}</script> (centered)
\par}%
\par}%
```

With script tags instead of `\(...\)`, etc. This is not limited to math markup. Certain other things like the image markup in the new version of the file leave you seeing img tags instead of working LaTeX code.

The cause of this seems to be that `$displayMode` is set once in a global fashion near the beginning of `PGbasicmacros.pl`, and in this case it is set to `HTML_MathJax` because that is the overall mode for the problem being rendered. But then the lines within PGML-lab.pg that temporarily set `$displayMode` to `TeX` only sort of get the job done. At places within `PGbasicmacros.pl` (for example within the `general_math_ev3` subroutine) the values of `$displayMode` and `$main::displayMode` end up being different. One is still `HTML_MathJax` and the other is `TeX`. I've seen this by printing warn statements to show their values while running the PGML lab problem. So some parts of the problem get translated into tex, and other parts into HTML.

I do not know how to address this issue, as it's getting into Perl global variables in a way that I don't have experience with.

Alex-Jordan:
I had a mess of edits that I couldn't cleanly separate into commits, so I just combined everything into one and force pushed this.

It does the things @drgrice1 commented on most recently: remove use of `MODES` and remove that one conditional about the displaymode not being `TeX`.

Also I added the option to see `PTX` code along with the existing `HTML` and `TeX`.

Note that both `TeX` and `PTX` are currently suffering from the display mode issue I posted about in a comment before this one.

Alex-Jordan:
Note that the "Rules" example here might not behave without https://github.com/openwebwork/pg/pull/956.

drgrice1:
I noticed the same thing with the `TeX` output when I made the first revision of the PGML lab problem for the last release.  I also do not have a solution.  The same thing happens with the original PGML lab problem.  That problem 
[lab-PGML.zip](https://github.com/openwebwork/webwork2/files/13342820/lab-PGML.zip)
and the lba2.pl macro it used are attached.

Alex-Jordan:
If I go inside `general_math_ev3` in `PGbasicmacros,pl` and set `$displayMode = $main::displayMode`, that fixes the issue for math markup. Or at least it appears to. And I cannot say if it has negative side effects. But anyway that is just one corner of this issue and not a path to a solution.

Do you understand why `PGbasicmacros,pl` needs to globally define `$displayMode` in the init subroutine? I guess this is important or why would it be there. But that seems to be a root cause here.

drgrice1:
The only fix for the `$displayMode` issue that will work will involve changes to the `PGbasicmacros.pl` file.

One solution that works is to delete lines 36, 60-64, and 69, and then change all other instances of `$displayMode` in the file to `$main::displayMode`.  That works for this and does not cause issues with other problems.

In general `PG_restricted_eval` is used way to often in PG macros, and rarely needs to be used.  Macros are already in the safe zone, and so are already in a restricted evaluation environment to begin with.  There are some uses that are needed, but the majority are not.  One example of a place where it should not be used is in the `_..._init` method for a macro when it is used purely to make a method or variable global.  For example if you change  lines 18-20 of `parserMultiAnswer.pl` to
```perl
sub _parserMultiAnswer_init { }

sub MultiAnswer { parser::MultiAnswer->new(@_) }
```
it works equally well (actually more efficiently since it avoids the unnecessary things done by `PG_restricted_eval`).

Alex-Jordan:
OK, I propose leaving this issue unaddressed in this PR. But reviewers should be aware that the TeX and PTX code comes out wrong because of this issue. We could separately change `PGbasicmacros.pl`. Unless you are really sure that the changes you suggested would not have unwanted side effects.

drgrice1:
I believe that at some point more of the `PG_restricted_eval` calls were needed.  As I understand it, it used to be the case that more macros were loaded like `PG.pl`, and then that would have been needed.  It certainly is not for `PGbasicmacros.pl` now.  In fact I suspect that most of the `PG_restricted_eval` calls in that file are not needed.  There are some that are though.  For example the call in `EV3P`.

I think that making the `$displayMode` change in another pull request is a good idea, and I wouldn't go any further than that for now either.

drgrice1 (approved):

pstaabp (approved):

PR #2242, opened 2023-11-07
Fix grading proctored test after entering incorrect proctor credentials.

drgrice1:
When grading a proctored test that is set for a finite number of graded submittsions, if you enter an incorrect username or password, then multiple hidden past_proctor_user and past_proctor_key fields are in the form (one from the `hidden_fields` method and one added in LoginProctor.html.ep).  Then if you submit the correct proctor username and password an error occurs because more than three arguments are passed to the WeBWorK::Controller `param` method on lines 900 and 901 of WeBWorK::ContentGenerator::GatewayQuiz.  So this forces the
    return value of `param` into scalar context so only the first of the
input values is returned.

To test this create a gateway test with a finite number of graded submissions per version, start the test, and click "Grade Test".  Then enter a bogus username or password and click "Continute".  Then enter the correct username and password.  On the develop branch this will give the error `Too many arguments for subroutine
'WeBWorK::Controller::param' (got 4; expected at most 3) at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm line 900`, and will work with this pull request.

This may be worthy of a hotfix.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -897,8 +897,8 @@ async sub pre_header_initialize ($c) {
 				# In this case there may be a past login proctor key that can be kept so that another login to continue
 				# working the test is not needed.
 				if ($c->param('past_proctor_user') && $c->param('past_proctor_key')) {
-					$c->param('proctor_user', $c->param('past_proctor_user'));
-					$c->param('proctor_key',  $c->param('past_proctor_key'));
+					$c->param('proctor_user', scalar $c->param('past_proctor_user'));
+					$c->param('proctor_key',  scalar $c->param('past_proctor_key'));
 				}
 			}
 			# This is unsubtle, but we'd rather not have bogus keys sitting around.


drgrice1:
The fact that this does not happen when an unlimited number of submissions per version is set is probably related to the other issue that @Alex-Jordan pointed out in the meeting today.  Also note that I have tested that issue with WeBWorK 2.17, and that issue already existed then, and is not a new issue (just not observed until now).  I am still not sure why that is happening.

Just to record that issue, what is happening is that if you grade a gateway test with an unlimited number of submissions per version, when you return to the test after entering the proctor username and password and then click to change to another page of the test it takes you back to the proctor login page instead of to the page of the test you clicked to switch to.

Alex-Jordan:
I'm having trouble reproducing the error on develop. I made a test with:
* Assignment Type: Proctored Test
* Test Time Limit: 60
* Graded Submissions per Version: 1
* Proctor Authorization Type: Both Start and Grade

I start the quiz, and enter my correct username/password to get started. Then I filled out one answer. I click Grade, and enter a correct username but bad password. It says "Invalid user ID or password." and I do it again with the correct username and correct password. And I get in with no error.

Is there a part of that I should be doing differently?

drgrice1:
What you did sounds right. I will check if there is some other setting that affects this.

drgrice1:
Oh wait.  I see that is must be more than one submission allowed per test version.  It won't happen with only one allowed.

Alex-Jordan (approved):
OK, I was able to reproduce the issue with it set to 3 graded attempts. And this cleared up the issue.

somiaj (approved):
Confirmed this fixed the issue. Fix looks simple I say a hotfix would be nice.

PR #2243, opened 2023-11-07
Don't delete proctor keys for tests with unlimited number of attempts

drgrice1:
Currently if a proctored test is set to have an unlimited number of attempts per version, the user clicks to grade the test, and the proctor credentials are validly entered, then if the user tries to move to another page of the test the user is taken to the proctor login page again.  This is because of an error in the logic that causes the proctor keys to be deleted instead of the proctor grading key.

Perhaps this is worthy of  a hotfix also?

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -879,7 +879,7 @@ async sub pre_header_initialize ($c) {
 
 	# Save results to database as appropriate
 	if ($c->{submitAnswers} || (($c->{previewAnswers} || $c->param('newPage')) && $can{recordAnswers})) {
-		# If answers are being submitted, then save the problems to the database.  If this is a preview or pages change
+		# If answers are being submitted, then save the problems to the database.  If this is a preview or page change
 		# and answers can be recorded, then save the last answer for future reference.
 		# Also save the persistent data to the database even when the last answer is not saved.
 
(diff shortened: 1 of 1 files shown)


drgrice1:
With this pull request you can now make the error fixed in #2242 occur for tests with unlimited attempts!

somiaj (approved):
Confirmed this fixed the stated issue. I think a hotfix would be good, fix seems is straight forward and will fix a potential issue.

Alex-Jordan:
Consider the case where there is a test with unlimited graded submissions and it allows multiple versions to be taken. That only makes sense if there is a time limit on the test, so assume there is a time limit.

Then does this change open the door for a student to be signed in by a proctor, let a test time out, and then start a new version without the proctor entering their credentials?

drgrice1:
No.  The proctor credentials would still be needed to start a new test.  This actually doesn't change the behavior at all for the case that you are describing.  This code is only executed if `submitAnswers` is true (i.e. if "Grade Test" is clicked).  So for what you are describing this never happens with the current code in develop anyway, and the proctor keys are left in the database.  That being in the database does not allow the student to start a new version (easily).  The `WeBWorK::Authen::Proctor` module still needs to receive the proctor_user and either the proctor_passwd or proctor_key in the parameters to authenticate.  So unless the student knows the proctor_passwd or proctor_key the student can not start a new version.

That being said, the student can get the proctor key.  While the student is taking the test, the student can open the developer tools and find the hidden proctor_key input and copy the key value from there (the proctor_user can also be obtained this way).  Then the student can leave the proctored environment, click on "Start New Test", add `&proctor_user=proctor_username&proctor_key=copied_key` to the end of the url in the address bar, and a new test version will start.  I have tested this and this works.  I tested this on the develop branch without this pull request.  This pull request has nothing to do with that, and you have found a security vulnerability that already exists.

drgrice1:
Note that the student can not use the same technique to grade the test.  Of course if you have the test set to ask for a proctor password only to start the test, the student can grade the test.

Alex-Jordan:
OK. There is a comment in the code rat lines 886,887 that I don't fully understand. Or if I understand what it says, then what you just explained means that comment is misleading.

drgrice1:
I think you understand what the comment says, but the comment clearly is not comprehensively true.  The problem is that the proctor keys are only deleted when a proctored test is submitted.  So if a test is never submitted, the keys are never deleted.  If the test is submitted to be graded, then the keys are deleted, and the student can't use the keys to start a new test.  That is all the comment is true with regards to.

Alex-Jordan:
Tested, and it works as described. I support hotfixes for this and #2242. Would you mind opening those? I'll file the other issue discovered here.

Alex-Jordan:
I thought this was already merged. Merging now.

Issue #746, opened 2016-10-06
Enable who have "modify_problem_sets" permission to view a changed seed problem and use check answer button


xcompass:
I would like to have a discussion before making a PR just to make sure my proposal is good.

So currently, Edit permission for problem sets are defined in "modify_problem_sets" setting and default is professor. We have changed it to TA to allow TA to update problems. However, then they are testing the question using changed seed, they can't use "Check Answer" button, because the changed seed is not submitted when "Check Answer" is clicked. It is because this line: https://github.com/openwebwork/webwork2/blob/master/lib/WeBWorK/ContentGenerator/Problem.pm#L1735 
Webwork will not generate the form field for changed seed unless the user is a professor. There is no issue with default permission settings but not for TAs.

I'm proposing to change the permission checking for above line to be consistent with the line which controls the display of the "Edit" button:https://github.com/openwebwork/webwork2/blob/master/lib/WeBWorK/ContentGenerator/Problem.pm#L1242.

Alex-Jordan:
There is still an issue here, 7 years later.

1. First, apply #2251 assuming it is not already merged.
2. Then set the permission for `modify_problem_sets` to `ta`.
3. Then while logged in as a `ta` user, visit the PG editor and have a problem that should be randomly different for different seeds.
4. Change the seed to something new like "314", and click to render the problem in a new tab.
5. Click "Check Answers" in the new tab. Now the problem will re-render with some default seed, not the "314" that you set.

The code surrounding this seems to be kind of a mess. Because maybe once upon a time, you could assume that having permission to use the Editor and having permission to modify a set would be the same thing. But in 2023, you might enter the editor to create a new PG file not attached to any set, to edit a hardcopy theme, or to edit the course_info file. For example maybe you would like to let a ta do some of these things even though you do not want to give a ta the ability to modify an actual problem set.

So anyway right now the permission `modify_problem_sets` is doing too much. But also, not enough, because there are places where some sort of permission level (either this one or some new permission level) should be used, but instead the code strictly requires the user to be professor level.

PR #2251, opened 2023-11-14
fix a permission mismatch

Alex-Jordan:
It seems to me that the wrong permission is referenced here. The problem editor page has nothing to do with student data.

What I'm changing it to seems like the best option. In some cases, the problem editor page allows you to edit a problem set. Although, if you enter it directly then it would seem that ability to edit a problem set is irrelevant, and why not let someone code a new problem? But this change is at least a better option than what is in there now, I think.

I came upon this reviewing old issues, in particular #746. For that issue to make sense, my guess is that `modify_problem_sets` is what should be here too. Assuming we don't make something better now that the problem editor is more general purpose than it once was.

### DIFF
--- a/templates/ContentGenerator/Instructor/PGProblemEditor.html.ep
+++ b/templates/ContentGenerator/Instructor/PGProblemEditor.html.ep
@@ -18,7 +18,7 @@
 	% last;
 % }
 %
-% unless ($authz->hasPermissions(param('user'), 'modify_student_data')) {
+% unless ($authz->hasPermissions(param('user'), 'modify_problem_sets')) {
 	<div class="alert alert-danger p-1 mb-0"><%= maketext('You are not authorized to modify problems.') %></div>
 	% last;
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Seems right to me.

drgrice1:
Another possibility would be to use the `modify_problem_template_files` permission.  That permission is currently not even used for anything.

Alex-Jordan:
It seemed like a few other places there is logic based on `modify_problem_sets` giving access to the editor, although I did not do a thorough investigation. Unless you think `modify_problem_template_files` would be better here, I propose leaving this PR the way it is. But also I was going to separately do a comprehensive refactoring of what grants access to the editor. Both to address #746 and also to make it make more sense now that the editor has more uses than it once did. I haven't thought it through yet but I think it would make this current PR moot.

pstaabp (approved):

PR #2252, opened 2023-11-14
give better information to certin users who are denied entry

Alex-Jordan:
Currently, if you are a user with a role of `Dropped` or `Proctor`, when you try to log in, you get the message `Invalid user ID or password.` This is particularly confusing/misleading for a student who was dropped who tries to get back in to WeBWorK using the username/password that worked for them in the past.

You also get the same message if the `login` permission level has been set high. I sometimes do this if I want to temporarily deny all students access to a course. Seeing `Invalid user ID or password.` is again misleading.

This PR changes things so that if one of the above situations happens, if that person gets past the point of entering a valid username/password combination, they are told more specifically `This user is not allowed to log in to this course`. It lets them know they have the right credentials, but for some reason are denied access right now.

I don't believe this leaks information, since they have to have the correct username/password to get this message.

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -63,17 +63,7 @@ use WeBWorK::Localize;
 use Caliper::Sensor;
 use Caliper::Entity;
 
-#####################
-## WeBWorK-tr modification
-## If GENERIC_ERROR_MESSAGE is constant, we can't translate it
-
-#use vars qw($GENERIC_ERROR_MESSAGE);
-our $GENERIC_ERROR_MESSAGE = "";    # define in new
-
-## WeBWorK-tr end modification
-#####################
-
-#use constant GENERIC_ERROR_MESSAGE => "Invalid user ID or password.";
+our $GENERIC_ERROR_MESSAGE = "";                # define in new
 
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
I think this is okay.  However, note that this does not apply to any authentication module that overrides the `get_credentials` or `check_user` methods.  Both of the LTI authentication modules do this.  Furthermore, both of those modules have similar code for this.  They should probably be modified in the same way.

Although, it seems that there is a missing step in the authentication modules in general.  Authentication should not be checking permissions before verifying credentials to begin with (so the change in order here is correct).  It would be better if the permission checks would be in an additional method that comes after the verification.  Then the other modules could just call that even if they override the verification method.  This is easy to implement.  Just add another step in the `do_verify` method, and the authentication procedure becomes
* get_credentials
* check_user
* verify_user
* check_user_permissions

drgrice1:
By the way, there is something odd here.  I checked out this branch from GitHub and got `access_denied`.  :stuck_out_tongue_winking_eye:

pstaabp (approved):

pstaabp:
I changed the login permission level to professor and got the new login message.

PR #2253, opened 2023-11-14
Show template problem ID in tests.

somiaj:
For users who can see the single problem grader, also show the template problem ID if the problems are in random order  or not enumerated from 1 to N. This way instructors can see what the real problem ID is in the case it is different from the problem number.

I'm unsure on the best label for this, I went with `Template ID` as `Template Problem ID` was a bit too long and I thought it was more descriptive than `Problem ID`.

This was mentioned in #2248.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -847,6 +847,11 @@ async sub pre_header_initialize ($c) {
 		push(@pg_results, $pg);
 	}
 
+	# Show the template problem ID if the problems are in random order
+	# or the template problem IDs are not in order starting at 1.
+	$c->{can}{showTemplateIds} = $c->{can}{showProblemGrader}
+		&& ($set->problem_randorder || $problems[-1]->problem_id != scalar(@problems));
+
 	# Wait for all problems to be rendered and replace the undefined entries
 	# in the pg_results array with the rendered result.
 	my @renderedPG = await Mojo::Promise->all(@renderPromises);
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -522,6 +522,10 @@
 					% # Output the problem header.
 					<h2><%= maketext('Problem [_1].', $i + 1) %></h2>
 					<span class="problem-sub-header">
+						% if ($c->{can}{showTemplateIds}) {
+							<%= '('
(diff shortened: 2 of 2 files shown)


drgrice1 (approved):
This looks good.  I am not entirely certain that tying this to the permission to use the problem grader is correct.  I am fine with that though.

Alex-Jordan:
I don't have an opinion, but alternative permission levels to use that occur to me are `can_show_source_file` and `modify_problem_sets`. The first one feels closer to semantically correct, but not perfectly so. The latter one is a bit of a misnomer because it grants access to the Set Details page, regardless of whether you would save any changes or just view what is there. But I guess if someone needs to act on the problem numbers they would see from this, they would be someone who has to go to the Set Details page.

somiaj:
My logic was pigging back off of being able to see the problem grader, since I assume it would be most useful for someone grading a test to know what the real problem ID is when they are randomized. I have no strong opinions on which permission is best, and didn't see one that would clearly be preferred.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -847,6 +847,11 @@ async sub pre_header_initialize ($c) {
 		push(@pg_results, $pg);
 	}
 
+	# Show the template problem ID if the problems are in random order
+	# or the template problem IDs are not in order starting at 1.
+	$c->{can}{showTemplateIds} = $c->{can}{showProblemGrader}
+		&& ($set->problem_randorder || $problems[-1]->problem_id != scalar(@problems));
```
Alex-Jordan, line 853: I might be overthinking something but are we sure that `$problems[-1]` will be the highest numbered problem? Like if a set def file defined problems in reverse order, could they then be entered into the database in reverse order, and then could that lead to `@problems` being something like `(3,2,1)`?  I know from experience that if a set def file defines problems out of order, that they will still be presented to the user in order. I'm just unsure at what step the sorting happens. If it happens right at the end when problems are displayed, then this check would not work as intended. I sort of suspect sorting happens at the end because of how we can reorder problems in a set. I would guess that is just changing the problem ID in the database, not sorting anything in the database.

somiaj: Looking a bit closer problems are retrieved via `getAllMergedProblemVersions` in `DB.pm`, and it seems to be ordered by `problem_id` via `my $order = ["problem_id"];`, so the problems should be in order by `problem_id`, making the last item of the array the biggest `problem_id`.

drgrice1:
Yes, @somiaj is correct.  The problems are ordered by problem_id when they are obtained from the database.  So the last element in the `@problems` array will have the largest `problem_id`.

Issue #2259, opened 2023-11-19
Nested maketext translations don't appear to be recoginzed.


somiaj:
When testing out some translations I noticed that one string wasn't translated in the ProblemSetDetail.

The code in the template nests `maketext` from lines 163-168 of `ContentGenerator/Instructor/ProblemSetDetail.html.ep`:

```
				<%== maketext(
					'Edit [_1] for set [_2].',
					link_to(maketext('individual user settings') =>
						$c->systemLink(url_for 'instructor_users_assigned_to_set')),
					tag('span', dir => 'ltr', format_set_name_display($setID))
				) =%>
```

I don't think the nested `maketext('individual user settings')` gets noticed, when looking at the .po files I don't find any `msgid` for `individual user settings`. The only match for `individual user settings` I find in the `.po` files is:

```
#. (link_to(maketext('individual user settings')
#: /opt/webwork/webwork2/templates/ContentGenerator/Instructor/ProblemSetDetail.html.ep:162
msgid "Edit %1 for set %2."
msgstr ""
```

So it appears the nested `maketext` doesn't get recognized and a translation string setup.

I think a similar case happens for a nested `maketext('class list data')` on line 47 of `ContentGenerator/Instructor/UserDetail.html.ep`

somiaj:
Fixed in #2261.

PR #2261, opened 2023-11-19
Remove nested maketext calls.

drgrice1:
The `xgettext.pl` script does not detect these.  So the nested calls need to be moved out and stored in a variable.

This fixes issue #2259.


pstaabp (approved):
Tested in a couple of places.  Looks good now.

somiaj (approved):
Double checked the translations screens get added the webwork2.pot file.

PR #2262, opened 2023-11-21
Trim date strings for the date picker and such (hotfix of #2258)

drgrice1:
It has been suggested to make this a hotfix.  So here it is.

This includes everything that was done in #2258, so see that for details.


somiaj (approved):
Thanks.

PR #2264, opened 2023-11-26
icons indicated exercise set type

Alex-Jordan:
We have a clock icon for sets that are timed tests which are still open to take. This appears on the main "Homework Sets" page.

This adds another icon for all other tests, so that there is an indicator that the set is a test (because that is not always clear from the name an instructor gives to the test.)

Screenshot 2023-11-26 at 11 49 09 AM

Additionally, these icons now appear in the Set Manager table so that the instructor can easily see where the tests are.

Lastly, the instructor might also want to see which sets are JITAR sets while in the Set Manager, so there is an icon for this too. I left that out of the "Homework Sets" page because I am not sure how useful that indicator would be for students.

Screenshot 2023-11-26 at 11 49 22 AM

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSets.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSets.pm
@@ -139,18 +139,20 @@ sub setListRow ($c, $set) {
 
 	my $display_name = format_set_name_display($set->set_id);
 
-	# Add clock icon if timed gateway
+	# Add icons for sets that are not "Homework"
+	my $iconClass;
+	my $iconTitle;
 	if ($gwtype && $set->{version_time_limit} > 0 && time < $set->due_date) {
-		$display_name = $c->c(
-			$c->tag(
-				'i',
-				class => 'icon far fa-clock',
-				title => $c->maketext('Test/quiz with time limit.'),
-				data  => { alt => $c->maketext('Test/quiz with time limit.') }
-			),
-			' ',
-			$c->tag('span', $display_name)
-		)->join('');
+		$iconClass = 'icon far fa-clock';
+		$iconTitle = $c->maketext('Test/Quiz with time limit.');
+	} elsif ($gwtype) {
+		$iconClass = 'icon fa-solid fa-list-check';
+		$iconTitle = $c->maketext('Test/Quiz.');
+	}
+
+	if ($iconClass) {
+		$display_name = $c->c($c->tag('i', class => $iconClass, title => $iconTitle, data => { alt => $iconTitle }),
+			' ', $c->tag('span', $display_name))->join('');
 	}
 
 	# This is the link to the set, it has tooltip with the set description.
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
Looks good to me.

Alex-Jordan:
I forget if we had conclusions about this from today to change it. Some questions might be:
* Do we keep the clock icon as it is, or simply make all tests (both proctored and unproctored) use the test icon? (You learn things like whether or not the test is timed when you visit the test landing page.)
* If all tests get the basic test icon, do timed tests also get the clock icon?

I lean toward simplicity and just having the one test icon and no more clock. So the "Homework Sets" page ("Assignments" once #2265 is merged) would only have one possible icon showing.

And then in "Set Manager" ("Sets Manager") there would also be the icon indicating JITAR.

drgrice1:
That sounds good to me.

Alex-Jordan:
For the JITAR icon, we could go with [this icon](https://fontawesome.com/search?q=guitar&o=r&m=free) instead :)

lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -139,18 +139,17 @@ sub setListRow ($c, $set) {
...
+	my $iconTitle;
+	if ($gwtype) {
+		$iconClass = 'icon fa-solid fa-list-check';
+		$iconTitle = $c->maketext('Test/Quiz');
+	}
+
+	if ($iconClass) {
+		$display_name = $c->c($c->tag('i', class => $iconClass, title => $iconTitle, data => { alt => $iconTitle }),
+			' ', $c->tag('span', $display_name))->join('');
 	}
```
drgrice1, line 153: Now that this is added if and only if the set is a gateway, there is no need for the successive conditionals. This could be
```suggestion
	# Add icons for sets that are not "Homework"
	if ($gwtype) {
		my $iconTitle = $c->maketext('Test/Quiz');
		$display_name = $c->c(
			$c->tag(
				'i',
				class => 'icon fa-solid fa-list-check',
				title => $iconTitle,
				data  => { alt => $iconTitle }
			),
			' ',
			$c->tag('span', $display_name)
		)->join('');
	}

```

templates/ContentGenerator/Instructor/ProblemSetList/set_list_row.html.ep
```diff
@@ -3,11 +3,24 @@
 % my $visibleClass   = $set->visible ? 'font-visible' : 'font-hidden';
 % my $set_id         = $set->set_id;
 % my $prettySetID    = format_set_name_display($set_id);
+% my $gwtype         = $set->assignment_type() =~ /gateway/;
+% my $iconClass;
+% my $iconTitle;
+% if ($gwtype) {
```
drgrice1, line 9: Since `$gwtype` is only used once, it could be eliminated by changing this to
```suggestion
% my $iconClass;
% my $iconTitle;
% if ($set->assignment_type() =~ /gateway/) {
```

drgrice1 (approved):
With the latest changes there are a couple of code clean up issues, but other than that this still looks good.

Alex-Jordan:
Thanks, I committed both suggestions.

somiaj (approved):
Looks good.

PR #2265, opened 2023-11-27
Rename several pages

Alex-Jordan:
This makes more name changes discussed at #1992.
```
'Homework Sets'      => 'Assignments',
'User Settings'      => 'Account Settings',
'User Manager'       => 'Accounts Manager'
'Set Assigner'       => 'Assigner Tool',
'Set Manager'        => 'Sets Manager',
'Achievement Editor' => 'Achievements Manager'
```

It also changes the order of nav items a little, moving "Assigner Tool" (formerly "Set Assigner") higher up.

Lastly, while reviewing all the places to make name changes, I found a few vestiges of previous names that had never been changed.

### DIFF
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -6,12 +6,12 @@
 % }
 <ul class="nav flex-column">
 	% if (defined $courseID && $authen->was_verified) {
-		% # Homework Sets or Course Administration
+		% # Assignments or Course Administration
 		<li class="list-group-item nav-item">
 			% if ($restricted_navigation) {
-				<span class="nav-link disabled"><%= maketext('Homework Sets') %></span>
+				<span class="nav-link disabled"><%= maketext('Assignments') %></span>
 			% } else {
-				<%= $makelink->('set_list', text => maketext('Homework Sets')) %>
+				<%= $makelink->('set_list', text => maketext('Assignments')) %>
 			% }
 		</li>
 		%
@@ -88,7 +88,9 @@
 		% if ($authz->hasPermissions($userID, 'access_instructor_tools')) {
 			<li><hr class="site-nav-separator"></li>
 			<li class="list-group-item nav-item"><%= $makelink->('instructor_tools') %></li>
-			% # Class list editor
+			% # Assigner Tool
+			<li class="list-group-item nav-item"><%= $makelink->('instructor_set_assigner') %></li>
+			% # Accounts Manager
 			<li class="list-group-item nav-item"><%= $makelink->('instructor_user_list') %></li>
 			% # User Assignments
 			% if (defined $eUserID && $eUserID ne $userID || defined $urlUserID) {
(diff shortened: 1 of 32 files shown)


drgrice1 (approved):
I think this looks good.

However, I don't like having the "Assigner Tool" first after "Instructor Tools".  In my opinion the two most important and most accessed instructor tools are the "Accounts Manager" and the "Sets Manager", and I think those should be listed first.  The "Assigner Tool" connects accounts with sets, and so it makes more sense that it be listed third.

I don't understand how moving the "Assigner Tool" first after "Instructor Tools" helps in any real way with getting used to the new menu (as @somiaj put it).

Alex-Jordan:
I'll move it to fourth after Instructor Tools, Accounts Manager, and Sets
Manager.

Any preference on the order of Accounts Manager and Sets Manager?

Should Sets Manager be Assignments Manager? I avoided that for being too
long, but we do have Achievements Manager already, which is longer

Alex-Jordan:
Actually I am starting to think of a Set as what we have always thought of
it. And an Assignment as a Set in context of being assigned to a user. So
Sets Manager seems right.

drdrew42 (approved):
👍

templates/ContentGenerator/Base/links.html.ep
```diff
@@ -6,12 +6,12 @@
 % }
 <ul class="nav flex-column">
 	% if (defined $courseID && $authen->was_verified) {
-		% # Homework Sets or Course Administration
+		% # Assignments or Course Administration
```
drgrice1, line 9: This comment should be updated.  This file no longer has anything to do with "Course Administration", and there isn't even such a link in the Admin course anymore.

drgrice1 (approved):
This generally looks good.

There are several instances of "Set Manager" that need to be changed to "Sets Manager" in `templates/HelpFiles/InstructorUserList.html.ep` on lines 153-168.

There are comments on lines 238 and 516 of `conf/localOverrides.conf.dist` that refer to the "homework sets editor". 

There are references to "homework sets" in `templates/ContentGenerator/Hardcopy/form.html.ep` lines 11 and 68, `templates/ContentGenerator/Instructor/Assigner.html.ep` line 9, `templates/ContentGenerator/Instructor/ProblemGrader.html.ep` line 37, `templates/ContentGenerator/Instructor/ProblemSetList.html.ep` line 35, `templates/ContentGenerator/Instructor/UsersAssignedToSet.html.ep` lines 9, 25, and 92, `templates/HelpFiles/InstructorScoring.html.ep` line 24, `templates/HelpFiles/InstructorUserList.html.ep` at numerous places, and `templates/HelpFiles/instructor_links.html.ep` on lines 40 and 46-48.  Should all of those be changed?  Should we go all of the way and eliminate the word "homework" displayed in the UI?  That might be a lot of work.

Alex-Jordan:
> There are comments on lines 238 and 516 of conf/localOverrides.conf.dist that refer to the "homework sets editor".

Do I understand right that both of these should be referring to the Library Browser?

Alex-Jordan:
I am replacing lots of these words/phrases with the approach that it is an "assignment" if it's in the context of something that has been assigned to a user. Otherwise, it is a "set". Does that sound OK?

Alex-Jordan:
> Both of those lines actually apply to both the Library Browser and the Set Detail page.

I see how line 238 applies to both. But line 516 is about set definition search depth. How does that apply to Set Detail?

drgrice1:
> I am replacing lots of these words/phrases with the approach that it is an "assignment" if it's in the context of something that has been assigned to a user. Otherwise, it is a "set". Does that sound OK?

Sounds good to me.

drgrice1:
> I see how line 238 applies to both. But line 516 is about set definition search depth. How does that apply to Set Detail?

The search depth is also used for the import sets drop-down on the sets manager page. Your right, not the set detail page.  However, that was the homework sets editor.

Issue #1151, opened 2020-10-28
Feature proposal: rerandomize seed for selected user on a selected set


taniwallach:
Consider adding to the Instructor tools a new feature: rerandomize seed for selected user on a selected set.

Suggested by @Alex-Jordan in https://webwork.maa.org/moodle/mod/forum/post.php?reply=15043#mformforum

It would be nice to for the code to also change the `psvn` of the assignment also, as that can be used instead of the `seed` by some problems.

Maybe the `set_user` table could also include a notes field, which could record when this function was applied (and in the future other things.)

---

Somewhat related: Consider whether it is possible for past answers database table to record the `seed` and `psvn` at the time of a submission, and to provide a manner to view the relevant version when viewing the past answers data.

Alex-Jordan:
I'd like to make this happen (or something like this). I propose the following, welcoming requests to do it differently.

1. On the Set Detail for one User page, for each problem, have a randomize button for the seed:
<img width="806" alt="Screenshot 2023-11-28 at 11 16 48 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/56db8cc5-8e29-4581-9598-0ea231c7c561">
You would still need to click to save changes like with any other edits here. The button would just be a handy unbiased tool to make a new seed.

2. On the Set Detail for one User page, up at the top next to the display mode select, have a button to randomize all seeds. Again it would just put random seeds in the input fields on the page, and you would still need to click the Save button.

3. On the Set Detail for multiple Users page, have a similar button to randomize all seeds for all users. It would randomize differently for each user. This time, the button actually writes new seeds to the database since you are not seeing all the seeds for all the users. That is, clicking Save is not part of the process.

4. In Instructor Tools, be able to select one or more users, with one or more sets, and there is a button to randomize all problems in all selected sets for all selected users. When you click the button, it should do the randomization, take you back to the Instructor Tools page, and give a message that the seeds were successfully reandomized.

Alex-Jordan:
This was (partly) accomplished with #2271. Not the part about psvn. But I'm going to close this issue and a new one could be opened.

PR #2266, opened 2023-11-28
improvements to Instructor Tools

Alex-Jordan:
This makes changes to Instructor Tools. Also it makes changes to `scrollingRecordList` which is used by Instructor Tools, and also by the Assigner Tool, Email, Hardcopy, and Past Answer pages. So testing should look at all five pages.

There are cosmetic changes that could be better. I won't describe them here; feel free to make suggestions for improving anything cosmetic. The functional improvements are:

* The sort options are only ones that make sense for the available display formats.
* There are now three format options for displaying sets. (There was only one before.) The two I added might not be best, but it's clear how to change them or add more. *Note:* the new format methods for sets include displaying the set's due date. This calls on `formatDateTime` and presently generates an error message. But #2263 will make that go away.
* Previously there were no filter options for sets. Now you can filter by type (e.g. "Proctored Test") and you can filter by Visible/NotVisible. Note that you have to have at least two types to see those filtering options. And you have to have both visible and not visible sets present to see those filtering options.
* I tried to make the functionality of all buttons more clear with changes to the wording.
* "Statistics for one user" is a silly thing, that is really just "Progress for one user". So I removed "Statistics for one user".
* If you are using this to add new users, you can choose ahead of time how many you want to add. Previously it was fixedat 5.
* Previously there was a button to take you to the Set Detail page for a set, and a separate button to take you to the Set Detail page for a set but with some subset of users selected. I combined these into one button. De-select all users if you want the regular Set Detail page.

A screenshot of what to expect:

Screenshot 2023-11-27 at 10 00 55 PM

### DIFF
--- a/lib/WeBWorK/Utils/FilterRecords.pm
+++ b/lib/WeBWorK/Utils/FilterRecords.pm
@@ -62,12 +62,14 @@ our @EXPORT_OK = qw(
 
 =over
 
-=item getFiltersForClass($class)
+=item getFiltersForClass($c, @records)
 
 Given a list of database records, returns the filters available for those
 records.  For all database records from the WeBWorK::DB::Record::User class
-the filters are by section or recitation or no filter at all.  For all other
-classes the only filter is no filter at all.
+the filters are by section or recitation or by that user's permission level
+in the permissionLevel table. For all database records from the
+WeBWorK::DB::Record::Set class the filters are by assignment type and by
+visibility. For other classes the only filter is no filter at all.
 
 The return value is a reference to a list of two element lists. The first
 element in each list is a a string description of the filter and the second
@@ -77,28 +79,46 @@ second value argument to the Mojolicious select_field tag helper method.
 =cut
 
 sub getFiltersForClass {
-	my (@records) = @_;
+	my ($c, @records) = @_;
 
 	my @filters;
 	push @filters, [ "\x{27E8}Display all possible records\x{27E9}" => 'all', selected => undef ];
 
 	if (ref $records[0] eq 'WeBWorK::DB::Record::User') {
-		my (%sections, %recitations);
+		my (%sections, %recitations, %permissions, %roles);
 
+		my %permissionName = reverse %{ $c->ce->{userRoles} };
 		for my $user (@records) {
 			++$sections{ $user->section };
 			++$recitations{ $user->recitation };
+			++$roles{ $user->status };
+			++$permissions{ $permissionName{ $c->db->getPermissionLevel($user->user_id)->permission } };
 		}
 
 		if (keys %sections > 1) {
 			for my $sec (sortByName(undef, keys %sections)) {
-				push @filters, [ ($sec ne '' ? "Section: $sec" : 'No Section') => "section:$sec" ];
+				push @filters, [ 'Section: ' . ($sec ne '' ? $sec : "\x{27E8}blank\x{27E9}") => "section:$sec" ];
 			}
 		}
 
 		if (keys %recitations > 1) {
 			for my $rec (sortByName(undef, keys %recitations)) {
-				push @filters, [ ($rec ne '' ? "Recitation: $rec" : 'No Recitation') => "recitation:$rec" ];
+				push @filters, [ 'Recitation: ' . ($rec ne '' ? $rec : "\x{27E8}blank\x{27E9}") => "recitation:$rec" ];
+			}
+		}
+
+		if (keys %roles > 1) {
+			for my $role (sortByName(undef, keys %roles)) {
+				my @statuses = keys %{ $c->ce->{statuses} };
+				for (@statuses) {
+					push @filters, [ "Role: $_" => "status:$role" ] if ($c->ce->{statuses}{$_}{abbrevs}[0] eq $role);
+				}
+			}
+		}
+
+		if (keys %permissions > 1) {
+			for my $perm (sortByName(undef, keys %permissions)) {
+				push @filters, [ "Permission Level: $perm" => "permission:$perm" ];
 			}
 		}
 	} elsif (ref $records[0] eq 'WeBWorK::DB::Record::Set') {
@@ -125,7 +145,7 @@ sub getFiltersForClass {
 	return \@filters;
 }
 
-=item filterRecords($filters, @records)
+=item filterRecords($c, $filters, @records)
 
 Given a list of filters and a list of records, returns a list of the records
 after the selected filters are applied.
@@ -137,7 +157,7 @@ C<$filters> should be a reference to an array of filters or be undefined.
 =cut
 
 sub filterRecords {
-	my ($filters, @records) = @_;
+	my ($c, $filters, @records) = @_;
 
 	return unless @records;
 
@@ -151,9 +171,18 @@ sub filterRecords {
 	for my $record (@records) {
 		for my $filter (@filtersToUse) {
 			my ($name, $value) = split(/:/, $filter);
-			if ($record->$name eq $value) {
-				push @filteredRecords, $record;
-				last;    # Only add a record once.
+			# permission level is handled differently
+			if ($name eq 'permission') {
+				my %permissionName = reverse %{ $c->ce->{userRoles} };
+				if ($permissionName{ $c->db->getPermissionLevel($record->user_id)->permission } eq $value) {
+					push @filteredRecords, $record;
+					last;    # Only add a record once.
+				}
+			} else {
+				if ($record->$name eq $value) {
+					push @filteredRecords, $record;
+					last;    # Only add a record once.
+				}
 			}
 		}
 	}
--- a/lib/WeBWorK/Utils/FilterRecords.pm
+++ b/lib/WeBWorK/Utils/FilterRecords.pm
@@ -148,11 +148,12 @@ sub getFiltersForClass {
 	return \@filters;
 }
 
-=item filterRecords($c, $filter_combine, $filters, @records)
+=item filterRecords($c, $intersect, $filters, @records)
 
 Given a list of filters and a list of records, returns a list of the records
-after the selected filters are applied. C<$filter_combine> should be 'union'
-or 'intersect'.
+after the selected filters are applied. If C<$intersect> is true then the
+intersection of the records that match the filters is returned.  Otherwise the
+union of the records that match the filters is returned.
 
 C<$filters> should be a reference to an array of filters or be undefined.
 
@@ -161,7 +162,7 @@ C<$filters> should be a reference to an array of filters or be undefined.
 =cut
 
 sub filterRecords {
-	my ($c, $filter_combine, $filters, @records) = @_;
+	my ($c, $intersect, $filters, @records) = @_;
 
 	return unless @records;
 
@@ -180,7 +181,6 @@ sub filterRecords {
 			$c->db->getPermissionLevelsWhere({ user_id => { not_like => 'set_id:%' } }))
 		: ();
 
-	my $intersect       = ($filter_combine eq 'intersect');
 	my @filteredRecords = $intersect ? @records : ();
 	if ($intersect) {
 		for my $filter (@filtersToUse) {
--- a/lib/WeBWorK/Utils/FilterRecords.pm
+++ b/lib/WeBWorK/Utils/FilterRecords.pm
@@ -113,7 +113,7 @@ sub getFiltersForClass {
 			for my $role (sortByName(undef, keys %roles)) {
 				my @statuses = keys %{ $c->ce->{statuses} };
 				for (@statuses) {
-					push @filters, [ "Role: $_" => "status:$role" ] if ($c->ce->{statuses}{$_}{abbrevs}[0] eq $role);
+					push @filters, [ "Enrollment Status: $_" => "status:$role" ] if ($c->ce->{statuses}{$_}{abbrevs}[0] eq $role);
 				}
 			}
 		}
--- a/lib/WeBWorK/Utils/FilterRecords.pm
+++ b/lib/WeBWorK/Utils/FilterRecords.pm
@@ -87,13 +87,15 @@ sub getFiltersForClass {
 	if (ref $records[0] eq 'WeBWorK::DB::Record::User') {
 		my (%sections, %recitations, %permissions, %roles);
 
-		my %permissionName = reverse %{ $c->ce->{userRoles} };
 		for my $user (@records) {
 			++$sections{ $user->section };
 			++$recitations{ $user->recitation };
 			++$roles{ $user->status };
-			++$permissions{ $permissionName{ $c->db->getPermissionLevel($user->user_id)->permission } };
 		}
+
+		my %permissionName = reverse %{ $c->ce->{userRoles} };
+		++$permissions{ $permissionName{$_} }
+			for map { $_->permission } $c->db->getPermissionLevelsWhere({ user_id => { not_like => 'set_id:%' } });
 
 		if (keys %sections > 1) {
 			for my $sec (sortByName(undef, keys %sections)) {
@@ -167,14 +169,22 @@ sub filterRecords {
 		return @records;
 	}
 
+	my %permissionName = reverse %{ $c->ce->{userRoles} };
+
+	# Only query the database for permission levels if a permission level filter is in use.
+	my %permissionLevels =
+		(grep {/^permission:/} @filtersToUse)
+		? (map { $_->user_id => $_->permission }
+			$c->db->getPermissionLevelsWhere({ user_id => { not_like => 'set_id:%' } }))
+		: ();
+
 	my @filteredRecords;
 	for my $record (@records) {
 		for my $filter (@filtersToUse) {
 			my ($name, $value) = split(/:/, $filter);
 			# permission level is handled differently
 			if ($name eq 'permission') {
-				my %permissionName = reverse %{ $c->ce->{userRoles} };
-				if ($permissionName{ $c->db->getPermissionLevel($record->user_id)->permission } eq $value) {
+				if ($permissionName{ $permissionLevels{ $record->user_id } } eq $value) {
 					push @filteredRecords, $record;
 					last;    # Only add a record once.
 				}
--- a/lib/WeBWorK/Utils/FilterRecords.pm
+++ b/lib/WeBWorK/Utils/FilterRecords.pm
@@ -51,7 +51,7 @@ use warnings;
 use Carp;
 
 use WeBWorK::Utils qw(sortByName);
-use WeBWorK::ContentGenerator::Instructor::ProblemSetDetail qw/FIELD_PROPERTIES/;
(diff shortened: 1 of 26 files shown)


Alex-Jordan:
I forgot something. Before these changes, the Instructor Tools page has a button to take you to the Email page, and another button to take you to the File Manager. I removed these. They seem unnecessary with those things being in the nav menu, and also having nothing to do with selecting users or sets.

somiaj:
@Alex-Jordan would having a button 'Email selected users' be useful? They can do this directly from the email page, so not really needed but could be an alternative to just removing it.  Note, I think removing it is fine, this is just a though that passed my mind.

drgrice1 (commented):
Line 26 of `templates/HelpFiles/Hardcopy.html.ep` and line 24 of `templates/HelpFiles/InstructorAssigner.html.ep` need to be changed.  The "Change Display Settings" buttons on those pages are now the "Refresh List" buttons.

I think that the button on the instructor "Email" page should also be changed to be a "Refresh List" button for consistency.  That is set on line 134 of `templates/ContentGenerator/Instructor/SendMail/main_form.html.ep`.  Probably just delete that line and use the default so that it will be in sync with the other places that the scrolling record list is used.  Also make sure to change line 22 of `templates/HelpFiles/InstructorSendMail.html.ep`.

I think that the restructuring of the input groups and the three categories looks pretty bad.  Trying to fit all three onto one row just really doesn't work.  At almost any window width, the input groups wrap.  Input groups do not wrap well at all.    The border radii are incorrect when wrapping occurs.  Things also look unbalanced above the `xl` breakpoint because of the different sizing of the categories and buttons.

I also don't like the way that the input texts are split into parts.  That problematic for translations (it already was some but now it is worse).  I think this was done to mitigate wrapping of input groups since the last input text grows, but that still doesn't really help because a wrapped input group just looks bad due to the border radius not working out when wrapping occurs.

The upper/lower case mixing also doesn't look very nice.  I think you are trying to associate the action with the menu link names, but I don't think the capitalization is needed for that.

I have added pull request https://github.com/Alex-Jordan/webwork2/pull/20 to this branch with my suggested way to deal with the issues noted in this review.

drgrice1:
I agree that the "Email" and "File Manager" buttons are not needed by the way.

Alex-Jordan:
Thanks @drgrice1  for the review and improvements.

More tools could be added. I'm polling to see what we would like.
* @somiaj asked about "Email all selected users"
* "View past answers for one user for one set"
* "Generate hardcopy for all selected sets for all selected users"

And as mentioned at #1151, I'd like to add "Randomize seeds for all selected assignments for all selected users". But that can come later when that feature is a thing in the first place.

Alex-Jordan:
I added the three tools mentioned in my previous comment.

The text description for those tools was long and I needed to grow the 350px size to 370px. This approach has fragility in at least two ways. A user might rescale font size and then these lines are still not long enough. Or a localization might be longer.

Since the `prob_id_sort` sorting routine is now used by both Instructor Tools as well as the Answer Log, I moved it to Utils and both pages import it from there.

There was an adjacent `byData` subroutine that appears to have not been used since 2012, and I removed that.

I had to change the name of the user select element in the Email page to make it match the name of the user select element from the Instructor Tools page. There was one place in `htdocs` where I then followed up with the same name change.

drgrice1:
Switch to using the `em` instead of `px`, and that should resolve the fragility of the width.

drgrice1:
That is with the font scaling. Not the translation issue.

drgrice1:
Another option that will generally resolve most issues is to just use a large enough width to leave excess. That won't entirely resolve the issue, but probably nothing will.

drgrice1:
It seems that `28em` is not enough.  The "View answers for selected users, for selected sets" is wrapping for me.  I tested this in both Firefox and Chrome.  If I increase it to `29em` it is enough, but not by much.

somiaj:
The size seems fine for me, I don't see the wrapping for "View answers for selected users, for selected sets" here. Could this one be shortened? Maybe "View selected users' answers for selected sets".

Also "Score selected sets for ALL users", ALL is not fully valid here, maybe "Score selected sets for active users" or "score selected sets for active students"?

drgrice1:
> Also "Score selected sets for ALL users", ALL is not fully valid here, maybe "Score selected sets for active users" or "score selected sets for active students"?

I think that "all" is fine here, and I think is better than "active".  "active" can mean many things.  Actively logged into the course.  Actively working on the assignment in question.  Furthermore, when a set is scored, all users that the set is assigned to are scored, even if they have not "actively" worked the set.

drgrice1:
Actually, testing further, it really is ALL users.  Even those that are not assigned the set are scored.

drgrice1:
Okay, so dropped students are not scored, but there is no reason to really be that detailed here.

drgrice1:
There is another sizing aspect to fiddle with.  The buttons could possibly be made a little smaller.  Currently we are using a width of 25%.  That could be reduced to 20%.  Unfortunately, bootstrap doesn't have a class for that, so it would have to be done with an inline style attribute.  Of course, if a translation of one of the button labels is longer there could be problems with that.

drgrice1:
I suppose you could also go back to the different sizing of the three categories that you had in this pull request to start with.  Make the second category with all of the longer phrases a bit bigger, and the other two smaller.

Alex-Jordan:
Since "Score" is not dependent on any user selections, I could move it to the left column. Is that a good move?

I should add at least one user format option that makes use of permission level and role. Then you would be able to sort by those too. And I could add filtering by those things as well.

The 28 em (and I believe 27em) is working for me with FF, Chrome, Safari. I wonder if it's something about me (and I assume @somiaj ) using a Mac. Anyway I will make it bigger the next round of edits. BTW I was thinking to make that a setting in some css file, but if we really are adjusting it to accommodate certain string lengths, it seems appropriate to set the width inline.

We could use `width:20%;` on the buttons instead of the bootstrap`w-25` class, and that recovers some space.

We could move to a tabbed menu for the three categories to make the spacing issues moot. At the cost of not seeing all the tools right away.

drgrice1:
> Since "Score" is not dependent on any user selections, I could move it to the left column. Is that a good move?

Although all of the left column's actions involve only users.  Also, then the left column would have more than all of the others, making it unbalanced.

> I should add at least one user format option that makes use of permission level and role. Then you would be able to sort by those too. And I could add filtering by those things as well.

That would be nice to have.  That would take some rewriting of the scrolling record list though since the permission level is not in the same database table.

> The 28 em (and I believe 27em) is working for me with FF, Chrome, Safari. I wonder if it's something about me (and I assume @somiaj ) using a Mac. Anyway I will make it bigger the next round of edits. BTW I was thinking to make that a setting in some css file, but if we really are adjusting it to accommodate certain string lengths, it seems appropriate to set the width inline.

I also thought that a css file might be needed.  Although, we don't have that many inline styles here (yet).

> We could move to a tabbed menu for the three categories to make the spacing issues moot. At the cost of not seeing all the tools right away.

I think a different approach like that is what is going to be needed in the end.  This approaches of trying to fit the three categories in at either a fixed width or a variable width using the bootstrap grid layout are just not entirely working.

Alex-Jordan:
Oh I meant whichever side has the set selection, so the right side. I got
mixed up. With that move plus eventually adding "randomize" to the middle,
I think the three columns end up with equal size.

drgrice1:
That makes more sense.  That is a good move then.

Alex-Jordan:
If I go to 29em, then even at full screen width on a MacBookPro, if the nav toc is there, the three columns do not fit on one row. Anyway I will explore a tab setup for this now.

drgrice1:
> I could change it to "Score selected sets for scorable users"?

How about just "Score selected set" and don't say anything about users just like it does on the scoring page itself?

somiaj:
> > I could change it to "Score selected sets for scorable users"?
> 
> How about just "Score selected set" and don't say anything about users just like it does on the scoring page itself?

I second this suggestion.

Alex-Jordan:
It still has the clunky look on a narrow screen when the button and text break on two lines. Maybe there is no need for `.input-group` now and we can fix this by dropping that?

somiaj:
What is your reasoning for not keeping all three buttons left aligned? I don't know if having them center/right aligned helps. Also why buttons and not tabs?

Alex-Jordan:
> What is your reasoning for not keeping all three buttons left aligned? 

So that, for example, the menu of buttons associated with users is over on the right, which is where the user selection is.

> why buttons and not tabs?

I thought the pills look better. It's ultimately just a small change to bootstrap classes.

Alex-Jordan:
> So that, for example, the menu of buttons associated with users is over on the right, which is where the user selection is.

Er, once again I forgot which thing in on which side. *Set* actions are over on the right.

Alex-Jordan:
Are you seeing it like this? Did you run `npm ci` and restart `webwork2`?

Screenshot 2023-11-30 at 3 01 35 PM

somiaj:
Yea, I ran `npm ci` and am seeing the updates. I was just curious, I would have just used tabs and kept things left aligned.

Alex-Jordan:
I may be overthinking it, but if you are up on the right part of the screen selecting some sets to Score for example, it seems helpful to have the relevant button be closer. It's also a visual cue that the action buttons available to you at that time are relevant to the right select menu only.

Alex-Jordan:
bootstrap is putting tabindex -1 on the non-active pills...not sure why yet. But it makes it so you can't keyboard navigate to the other button menus. Will need to fix that.

Alex-Jordan:
The bootstrap tabs/pills get `tabindex="-1"` when bootstrap js does its thing. It even overrides if I put `tabindex="0"` in the source. I don't get why bootstrap would do this. Why make the tabs untabbable?

Alex-Jordan:
Here is the default styling as "tabs" instead of "pills", if you are interested.

Screenshot 2023-11-30 at 4 41 43 PM

drgrice1:
> The bootstrap tabs/pills get `tabindex="-1"` when bootstrap js does its thing. It even overrides if I put `tabindex="0"` in the source. I don't get why bootstrap would do this. Why make the tabs untabbable?

That is okay here.  You use the arrow keys to switch tabs.

drgrice1:
@Alex-Jordan: I added another pull request to this pull request branch that tweaks the styles some.  I am seeing a lot of wrapping of the input groups at various window widths with the current grid layout (in fact really bad below the `sm` breakpoint).  The pull request tweaks the `col` sizes for those narrower window widths to fix that, and a few other minor tweaks.

drgrice1:
By the way, there is more about why the tabs do not individually receive tab focus at https://getbootstrap.com/docs/5.3/components/navs-tabs/#accessibility.

Alex-Jordan:
Thanks for the assist. I'm becoming more familiar with bootstrap, but I'm not there yet.

Alex-Jordan:
Now you can filter users by permission level or role.

Question: If you selected to filter by `Section: Foo` and `Role: Enrolled`, would you expect to get:
* the students from section Foo who are Enrolled
* all students from section Foo, union with all students who are Enrolled

You currently get the second outcome. To me, a "filter" is more of an intersection thing than a union thing, and I would expect the first outcome.

Which outcome is actually useful for a WeBWorK instructor?

Alex-Jordan:
Another question. Are the "Section: ⟨blank⟩" and "Recitation: ⟨blank⟩" filters useful, or just clutter?

drgrice1:
> Question: If you selected to filter by `Section: Foo` and `Role: Enrolled`, would you expect to get:
>     * the students from section Foo who are Enrolled
>     * all students from section Foo, union with all students who are Enrolled
> You currently get the second outcome. To me, a "filter" is more of an intersection thing than a union thing, and I would expect the first outcome.
> Which outcome is actually useful for a WeBWorK instructor?

I think that the most useful would be the intersection in the example given.  However, for some other combinations of the filters a union is more useful.  For example, if you want to view all users in two different sections, or in two different recitations.

> Another question. Are the "Section: ⟨blank⟩" and "Recitation: ⟨blank⟩" filters useful, or just clutter?

I think that if you are a course coordinator working on assigning sections to all students, then it might be useful to filter by students with sections not yet set so you can clearly see which students you have left to assign sections to.  So there are probably use cases that justify this filter.

drgrice1 (commented):
I see a few more things.

I added another pull request to this branch that fixes the efficiency issues mentioned.

drgrice1:
By the way, to test the inefficiency, you can create a course and import the attached class list file which contains 5000 users.  With this pull request as is, I am seeing that it takes more than one second for a request that uses a permission filter to complete, while with my changes in my current pull request to this branch it decreases to less than a fifth of a second.

[longDemoCourse.lst.txt](https://github.com/openwebwork/webwork2/files/13527782/longDemoCourse.lst.txt)

Alex-Jordan:
> I also noticed that none of the scrolling record list texts are translated.

I noticed that too. At that level, `maketext()` is not available and I opted not to investigate what is needed to get it for now. Sounds like you know what must be done.

drgrice1 (approved):
I think this is looking good now.  The translations can wait, as previously stated, and since that will take some work to implement.

I am approving this pending the fix of the minor perltidy issue!

Alex-Jordan:
perltidy fixed.

I have one loose thread with this that I'm working on. If I finish it this morning I will push. Otherwise I'll sit on it for a small future PR.

Alex-Jordan:
I added radio buttons to choose if you would like to intersect or union the filter results. In my testing, it is functional. Perhaps improvements can be made with the layout.

somiaj:
The layout of the buttons needs improvement here.

image

Each button should be on its own row, and some padding is needed imo.

Alex-Jordan:
Thanks @somiaj , that is in there now.

The word "Intersect" is long, and even on a mildly narrowed screen, will break onto the next line. It also happens with "Union". I'm OK with it but we could also use union and intersect symbols here as icons with a title that has the full word. Open to that or any better ideas.

somiaj:
I think the words are better than symbols. Though you could use the logic equivalent of "And" / "Or", though I don't think that would be as clear (just shorter).

Personally I would adjust the css so the column that has the labels is large enough it won't wrap (though unsure the best way to do this). In almost all cases on the narrow screens in my testing there is still plenty of space in the other column that could be used to avoid wrapping.

somiaj:
I also think adding a small description of the options in the help for the page would be useful.

Alex-Jordan:
The width of those regions is discretely sized with bootstrap grid columns. Here is what it looks like (on the left) if I bump it up to the next possibility:

Screenshot 2023-12-01 at 11 15 47 AM

I like the idea to change the wording to "AND" and "OR". I'll wait to see if more people want to give input.

somiaj:
Adding `text-nowrap` to the div that contains the button / label, keeps wrapping from happening, but then it does get to small in some cases. I also tried switching from `col-2` / `col-10`, to `col-3` / `col-9`, and this mostly works (there is a small point right before the smallest break point that this causes a small issue, and on larger widths there is to much white space, but we could also configure the column size based on break points, so smaller break points swap to `col-3` / `col-9`, but use the current setup for the larger breakpoints.

somiaj:
Note there is already an issue right before the smallest break point before this change where the end of `Sort By:` and `Format:` are getting cutoff, but I don't know css/bootstrap well enough to have any suggestions other than just playing around.

lib/WeBWorK/Utils/FilterRecords.pm
```diff
@@ -117,7 +161,7 @@ C<$filters> should be a reference to an array of filters or be undefined.
 =cut
 
 sub filterRecords {
-	my ($filters, @records) = @_;
+	my ($c, $filter_combine, $filters, @records) = @_;
```
drgrice1, line 164: The `$filter_combine` parameter really shouldn't be a string value.  It should be a boolean since it is either union or intersect (an on/off state).  Basically, just use `$intersect` for the parameter, and delete line 183.

drgrice1:
Instead of trying to squeeze the intersection and union radio buttons to the left, where it is really never going to fit, just put it in a horizontal format below the filter select (sort of as @somiaj suggested, but leave the filter label to the left as it is).

drgrice1:
Also, you have the problem that came up before with the current structure of the grid layout also.  There are elements with `col` classes that are not direct descendants of `row` divs.

drgrice1:
I added another pull request that implements the suggestions I made in the last few comments.

Alex-Jordan:
There is now one little cosmetic thing that @somiaj found. If you make the screen narrow so that the main content is a little over 500px wide, you don't have space for some words: "Sort By:", "Format:", and "Filter(s):".
Screenshot 2023-12-01 at 1 12 18 PM

drgrice1:
> There is now one little cosmetic thing that @somiaj found. If you make the screen narrow so that the main content is a little over 500px wide, you don't have space for some words: "Sort By:", "Format:", and "Filter(s):".

Yeah, that was that way before though.  Although before it wasn't quite as bad, and only really affected "Format".  This only happens right at the cut of from the `md` to the `sm` break point (assuming you have the site nav open), and if you make the window really narrow below the `sm` breakpoint.  The latter case is not really concerning.  Even on a mobile device with really the narrowest layout that I think can validly be supported it fits fine.  It is not to bad right at the cut of from `md` to `sm` either.  So I don't think that this is to big of a deal. 

Also, I think that @somiaj is right that comments in the help files might be a good idea for the union/intersect radios.  Unfortunately, that will need to be added to all of the pages that use the scrolling record list.  Or perhaps you could create a separate template help file for scrolling record lists, and include that in the help files that use it.

drgrice1:
One option would be to drop the horizontal layout when you are below the `lg` breakpoint.  That would fix that issue entirely, but would make the scrolling record list longer.

drgrice1:
It is to bad that there is not a css media query that applies to container width instead of window width, because things are quite different in the range from the `md` to the `lg` breakpoints when the site nav is open and when it is not.

drgrice1:
There is something else that I missed with the union/intersect radio button labels.  The font sizes for those are a bit larger than the other labels.  This is because the others use the col-form-label-sm class.  There is not `sm` variant for `form-check-labels`.  However, there is a `math4.scss` class that I created to accommodate this that gives the same font size.  It is `font-sm`.  So you might want to add that class to those labels.

drgrice1:
I added the `font-sm` to the pull request.  Here is a screen shot of what it looks like:
labels-above

Alex-Jordan:
OK, unless I forgot about something, I think this is ready for last looks.

drgrice1 (commented):
There is an indentation issue in three of the help files.  Probably a cut and paste issue.

Alex-Jordan:
> There is an indentation issue in three of the help files. Probably a cut and paste issue.

Fixed now.  Searching for `^ `, I found a few unrelated instances among the template files as well.

drgrice1:
Yeah, I try to do that on occasion.  I want to keep the template code clean, because there is nothing like perltidy for those files to help with formatting the template code.

pstaabp (approved):
This is a nice improvement.

Issue #1093, opened 2020-03-06
Classlist editor - minor feature request - UTF-8 support


taniwallach:
When using a foreign language - the filter by "what text" for "Permission level" searches properly if the English name of the permission level is used, while the table lists the translation. It would be nice if both options worked.

somiaj:
Fixed in #2268, so closing.

PR #2268, opened 2023-11-30
Fix filtering the user list by permission level or status for other languages.

drgrice1:
Currently when filtering by permission level or status the text entered in the "Filter by what text?" field is matched against the untranslated permission level name or status name.  As a result other languages must enter the English names for these to match what is displayed translated which is probably very confusing.  This just matches against the translated names.

This fixes issue #1093.


drgrice1:
Yeah, I was thinking the same thing, but I just wanted to close that relatively old issue!

Alex-Jordan (approved):

taniwallach (approved):
Thanks for letting the old methods keep working.

PR #2269, opened 2023-12-02
reposition footer

Alex-Jordan:
This puts the footer in a bootstrap row, and makes it a col-10 above the large breakpoint.

On large screens, the footer was looking too far to the right, imo. This generally nudges it a bit to the left.

### DIFF
--- a/htdocs/themes/math4/system.html.ep
+++ b/htdocs/themes/math4/system.html.ep
@@ -172,7 +172,7 @@
 		% }
 		%
 		% # Footer
-		<div id="footer" role="contentinfo"><%= include 'ContentGenerator/Base/footer' %></div>
+		<div id="footer" role="contentinfo" class="row"><%= include 'ContentGenerator/Base/footer' %></div>
 	</div>
 </div>
 %
--- a/htdocs/themes/math4/system.html.ep
+++ b/htdocs/themes/math4/system.html.ep
@@ -172,7 +172,7 @@
 		% }
 		%
 		% # Footer
-		<div id="footer" role="contentinfo" class="row"><%= include 'ContentGenerator/Base/footer' %></div>
+		<%= include 'ContentGenerator/Base/footer' =%>
 	</div>
 </div>
 %
--- a/templates/ContentGenerator/Instructor/Scoring.html.ep
+++ b/templates/ContentGenerator/Instructor/Scoring.html.ep
@@ -13,56 +13,58 @@
 % }
 %
 % my $scoringDir = $ce->{courseDirs}{scoring};
-% $c->stash->{footerWidthClass} = 'col-md-10 col-lg-8';
+% $c->stash->{footerWidthClass} = 'col-md-12 col-lg-10 col-xl-8';
 %
 <div class="row">
-	<div class="col-md-10 col-lg-8 border border-dark p-2">
-		<%= form_for current_route, name => 'scoring-form', id => 'scoring-form', method => 'POST', begin =%>
-			<%= $c->hidden_authen_fields =%>
-			<%= hidden_field returning => 1 =%>
-			<div class="row">
-				<div class="mb-2 col-6">
-					<%= label_for selectedSet => maketext('Selected sets:'), class => 'form-label' =%>
-					<%= select_field selectedSet =>
-							[ map { [ format_set_name_display($_) => $_ ] } @{ $c->{ra_set_ids} } ],
-						id => 'selectedSet', class => 'form-select', size => 10, multiple => undef, dir => 'ltr' =%>
+	<div class="col-md-12 col-lg-10 col-xl-8">
+		<div class="border border-dark p-2">
+			<%= form_for current_route, name => 'scoring-form', id => 'scoring-form', method => 'POST', begin =%>
+				<%= $c->hidden_authen_fields =%>
+				<%= hidden_field returning => 1 =%>
+				<div class="row">
+					<div class="col-sm-6 mb-2">
+						<%= label_for selectedSet => maketext('Selected sets:'), class => 'form-label' =%>
+						<%= select_field selectedSet =>
+								[ map { [ format_set_name_display($_) => $_ ] } @{ $c->{ra_set_ids} } ],
+							id => 'selectedSet', class => 'form-select', size => 10, multiple => undef, dir => 'ltr' =%>
+					</div>
+					<div class="col-sm-6 mt-sm-5">
+						<div class="form-check">
+							<label class="form-check-label">
+								% param('padFields', 1) unless param('returning');
+								<%= check_box padFields => 1, class => 'form-check-input' =%>
+								<%= maketext('Pad fields') =%>
+							</label>
+						</div>
+						<div class="form-check">
+							<label class="form-check-label">
+								% param('includePercentEachSet', 1) unless param('returning');
+								<%= check_box includePercentEachSet => 1, class => 'form-check-input' =%>
+								<%= maketext('Include percentage columns') =%>
+							</label>
+						</div>
+						<div class="form-check">
+							<label class="form-check-label">
+								<%= check_box includeIndex => 1, class => 'form-check-input' =%>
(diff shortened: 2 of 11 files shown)


htdocs/themes/math4/system.html.ep
```diff
@@ -172,7 +172,7 @@
 		% }
 		%
 		% # Footer
-		<div id="footer" role="contentinfo"><%= include 'ContentGenerator/Base/footer' %></div>
+		<div id="footer" role="contentinfo" class="row"><%= include 'ContentGenerator/Base/footer' %></div>
```
drgrice1, line 175: When I converted the footer into an included Mojolicious template from the old template system in which this was a template escape, I maintained the structure which had the containing `div` here.  I really should have moved that containing div into the included footer template though.  This is made worse conceptually with this change though, because this `div` has the `row` class which is important for the `col` class on the `div` in the included template.  So I suggest that this be changed to
```suggestion
		<%= include 'ContentGenerator/Base/footer' =%>
```
and the `<div id="footer" role="contentinfo" class="row">...</div>` be moved into the footer.html.ep file.

drgrice1 (approved):
I am okay with this.  However, I do notice that on some pages where the content is not also restricted to the `col-10` width (like the "User Manager", the "Set Manager", etc.), things look a little unbalanced.  The footer is to the left of center relative to the full width content directly above it.

Alex-Jordan:
> The footer is to the left of center relative to the full width content directly above it.

Yeah, I'm noticing that now too on the login screen, which I really don't like since that's a screen everyone sees a lot. I'll rethink this.

drgrice1:
> Is it possible for a template (like the main template for a page) to store a value in some way so the footer template can access it? Then the main templates could store something about their width (e.g. `col-10`) and the footer template could access that.

Yes.  I added a pull request to this branch that implements the mechanism for this.  It is not complete though.  I left the work of going through page by page and determining what width is needed to you.  I think most pages will not need anything as the default of `col-12` will work.  You just need to figure out what pages need something different.

drgrice1:
I think that in most cases the `footerWidthClass` should be set in the templates in the same place where the primary layout is determined, and not in the .pm files. I just pointed out that it could be done there in case a situation is encountered where that makes more sense.

drgrice1 (commented):
This generally looks good except the scoring page.  I am not entirely sold on centering the footer relative to the left part of the page content for the pages that have the info box, but I can live with it.

Alex-Jordan:
The last commit here gets a bit off topic, and makes some styling changes to the Scoring template. For now it still has the footer center aligned with the control panel at the top.

Screenshot 2023-12-04 at 4 10 06 PM

I'm happy to take feedback about the styling changes and change them. And if it's the majority opinion to center the footer with the page instead of with the control panel, I'll change that too. I was trying to avoid the status quo that is like:

Screenshot 2023-12-04 at 4 07 55 PM

where the footer feels mis-positioned. But I know it might just be trading one bad thing for another.

drgrice1 (commented):
I am not liking the new layout for the form on the scoring page.

I do like the lists for the set scoring file downloads though.

templates/ContentGenerator/Instructor/Scoring.html.ep
```diff
@@ -13,83 +13,94 @@
...
-						<%= check_box padFields => 1, class => 'form-check-input' =%>
-						<%= maketext('Pad Fields') =%>
-					</label>
+<div class="row">
+	<div class="col-md-10 col-lg-8 border border-dark p-2">
+		<%= form_for current_route, name => 'scoring-form', id => 'scoring-form', method => 'POST', begin =%>
+			<%= $c->hidden_authen_fields =%>
+			<%= hidden_field returning => 1 =%>
+			<div class="row">
+				<div class="mb-2 col-6">
```
drgrice1, line 24: The changes here and the wrapping `row` and `col` divs above make wrapping of the radio button labels even worse.  Even with the shortened labels.  I also don't like that the border is now flush up against the site nav.  Generally, borders don't work very well on the `div`'s that are part of a bootstrap grid layout.  You are probably going to need to add another `div` inside grid layout `div`'s that contains the form, and put the border on that.  The following screenshots are taken at the same window widths, the first with this pull request, and the second with the develop branch.

scoring-1
scoring-1-develop

Also, I really don't like that you removed the `sm` breakpoint from this.  Below the `sm` breakpoint (the `xs` layout) is generally for mobile devices.  Here is what things look like with this pull request compared to develop on an android device in Google Chrome:

scoring-2
scoring-2-develop

Notice that selects behave differently on mobile devices, and it is better if this is not in the `col` layout.

drgrice1:
I added a pull request to this branch that fixes the things I mentioned in one way.

Alex-Jordan:
Thanks for that, it's merged now and it looks better.

Still accepting opinions if the footer should go back to being `col-12` or stay aligned with the controls.

drgrice1:
> Still accepting opinions if the footer should go back to being `col-12` or stay aligned with the controls.

I am okay either way on that.  The issue is that no matter what you choose, the footer will not always be aligned with whatever is directly above it (with the current layout -- see below).  If there are no scoring files, then the scoring section below is not shown.  If there are scoring files, then the data shown may be wide or narrow depending on the contents of the scoring file, and generally aligning the footer with that is completely impossible since `fit-content` is used.

One option would be to remove `width:fit-content; max-width:100%;` style that was added to the scoring file content display, and change to `$c->stash->{footerWidthClass} = -f "$scoringDir/$c->{scoringFileName}" ? 'col-12' : 'col-md-12 col-lg-10 col-xl-8';` for the `footerWidthClass`.  Then when the score file contents are not shown, the footer will be centered below the form, and when the scoring file contents are shown the footer will be centered below the full width `pre` containing the contents.

Alex-Jordan:
This had conflicts that I just resolved with a rebase and some editing. It was just about the presence of the word "at" in the footer for "Page generated ...".

Alex-Jordan:
I never saw or used the `s...scr.csv` or `s...ful.csv` files before working on this page for this PR. Such files are not created until you check the box "Record scores for each set" and then score the selected sets. I get it that the purpose is to score lots of sets all together, but also created the individual `csv` files for each set. Those could be used in a mailmerge, for example.

I am not really sure why have two of these. Did only the `scr` variant exist at first, then `ful` was added but to not break people's mail templates, they kept both?

Anyway suppose you select sets A, B, C, and check the box "Record scores for each set", and click to score. Next week you come back and you select only set C, and you do not check that box, and you score. The two files `sC....csv` have not been updated. You are shown the main scoring csv file you just created for set C, and also shown links for the two old files for set C, and it is suggested you can use these in a mailmerge. You are misled to think those have been updated.

I'm not going to mess with any of this, but it seems like it needs attention. Possibly revisiting how many of the features here are useful and/or actually used.

Alex-Jordan:
> One option would be to remove width:fit-content; max-width:100%; style that was added to the scoring file content display, and change to $c->stash->{footerWidthClass} = -f "$scoringDir/$c->{scoringFileName}" ? 'col-12' : 'col-md-12 col-lg-10 col-xl-8'; for the footerWidthClass. Then when the score file contents are not shown, the footer will be centered below the form, and when the scoring file contents are shown the footer will be centered below the full width pre containing the contents.

I tried this. It rarely applies the `col-md-12 col-lg-10 col-xl-8` class. It only seems to do so if I clear the `scoring/` folder. So it does align the footer well the first time an instructor visits scoring tools, but in practice after that it will always be `col-12`.

Another thing about this I was trying to do, is when the csv file being shown is not wide, having it artificially extend all the way to the right of the page seems wrong. The illusion of that happening is there now because I gave the `pre` a background color. The `width:fit-content; max-width:100%;` constrains the `pre` so it won't get that wide, and the gray background is more of a tight hull surrounding the csv file. All that said, it's not important to do this. I could do what you suggested and live with the extra gray "padding" of the csv presentation. I'll take anyone's expressed preference about it (and sorry of that's already been given but I missed it.)

pstaabp (approved):

PR #2270, opened 2023-12-05
Improvements to the essay "needs_grading" flag.

drgrice1:
First, the single problem grader now removes this flag when an essay answer is graded.  A new `mark_graded` parameter for the WebworkWebservice::ProblemActions `putUserProblem` and `putProblemVersion` methods is added to accomodate this.

When a problem is rendered this flag now needs to be passed to PG to indicated whether a problem still needs to be graded so that essay answers can not always show "Ungraded".  This is needed for https://github.com/openwebwork/pg/pull/979.

The `create_ans_str_from_responses` method now handles adding this flag instead of that code being duplicated in ProblemProcessing.pm and GatewayQuiz.pm.

### DIFF
--- a/htdocs/js/ProblemGrader/problemgrader.js
+++ b/htdocs/js/ProblemGrader/problemgrader.js
@@ -104,7 +104,8 @@
 						set_id: saveData.setId,
 						version_id: saveData.versionId,
 						problem_id: saveData.problemId,
-						status: parseInt(scoreInput.value) / 100
+						status: parseInt(scoreInput.value) / 100,
+						mark_graded: true
 					}),
 					signal: controller.signal
 				});
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -926,7 +926,8 @@ async sub pre_header_initialize ($c) {
 			if (ref $pg_result) {
 				my ($past_answers_string, $scores);    # Not used here
 				($past_answers_string, $encoded_last_answer_string, $scores, $answer_types_string) =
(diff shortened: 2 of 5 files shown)


Alex-Jordan:
Suppose a student answered an essay, and the instructor graded it. The student returns to the problem and clicks submit again, without changing anything. They see the feedback button that says "Graded", and it does not have an additional message about being graded at a later time. So far, so good.

Then, without changing anything, the student hits submit again. Now the feedback is "Ungraded" with "This answer will be graded at a later time." So one or the other of these last two submissions reset something to mark the problem as needing to be graded.

Would it be possible to compare the essay answer text currently being submitted with the essay answer text that is stored from the previous submission, and don't change the status for needing to be graded if there is no change in the essay text content?

drgrice1:
That can be done.  However, there is a caveat.  If a problem has multiple essay answers, and the problem was graded, then if the student changes one of the essay answers but not the other, both will still show up as needing grading.  There is only one flag and it is per problem, not per answer.

drgrice1:
I set it up so that if a student submits a problem that has been graded without changing any of the essay answers in the problem, then the answer remains marked as having been graded (and is reported as such in the feedback).  Of course, the caveat in my previous comment applies to problems with multiple essay answers.

drgrice1:
Note that if a student answered an essay question previously so that the answer has been marked as needing grading (and the instructor has not graded it), and then the student deletes the answer entirely and submits that, then the flag marking the answer as needing grading will be removed.

That is actually an unintended consequence of how I set up the logic.  I realized this, and was going to fix it.  However, I think that is actually good.  So I will leave it.  An empty answer should not be marked as needing grading.

pstaabp (approved):

drdrew42 (approved):

Issue #2272, opened 2023-12-05
error editing details for a jitar set with multiple users


Alex-Jordan:
If there is a JITAR set assigned to two users...

And you go to Instructor Tools, select that set and those two users...

And click "Edit one set's details for some or all users"...

then there is an error message: `Can't call method "counts_parent_grade" on unblessed reference at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm line 984. `

I haven't explored what causes this. But I'm filing the issue, discovered while testing things with #2271.

PR #2273, opened 2023-12-06
Fix issue #2272.

drgrice1:
Make sure that the `$userRecord` that is passed to the `fieldHTML` method in `ProblemSetDetail.pm` exists before trying to access it.

Note the actual error occurs for the `counts_parent_grade` column of the problem record which is the only problem record field that is of the "choose" type (which is why this does not occur for non JITAR sets). The earlier checks that the $userRecord exists are needed as the hacked method of accessing the data directly via the hash key rather than using the accessor turns the undefined value into a hash reference if that is done.  Then the check if the $userRecord exists in the "choose" case later will fail because it will exist.

I don't know how long this bug has existed, but I know it goes back to at least WeBWorK 2.17 (I suspect it goes back considerably further though).  The fieldHTML method in general needs a rewrite.  There is to much hackery in use there.  The comment on lines 855-857 is about part of what is causing this problem.  That comment dates back to 2007, and the problem it refers to goes back further.


drgrice1:
I upgraded this to a more proper fix and removed the direct access to the hash keys of the database record.

Alex-Jordan (approved):

pstaabp (approved):
Fixes the issue.

PR #2274, opened 2023-12-06
Add problem seed to past answers.

drgrice1:
Add a `problem_seed` column to the `past_answer` database table.  This is displayed in the first column of the table shown on the "Answer Log" page.  Since that table is not a fully qualified data table and is missing headers, it is displayed in the table as "Seed: 123".  If a problem does not have a seed (because it is from a past answer before this pull request like an archived course), then "Seed: unknown" is shown.

### DIFF
--- a/templates/ContentGenerator/Instructor/ShowAnswers/past-answers-table.html.ep
+++ b/templates/ContentGenerator/Instructor/ShowAnswers/past-answers-table.html.ep
@@ -8,11 +8,11 @@
 		% for my $problemNumber (sort { $a <=> $b } keys %{ $records->{$studentUser}{$setName} }) {
 			% my @pastAnswerIDs = sort { $a <=> $b } keys %{ $records->{$studentUser}{$setName}{$problemNumber} };
 			% my $prettyProblemNumber = stash->{prettyProblemNumbers}{$setName}{$problemNumber};
-			<h3>
+			<h2 class="fs-3">
 				<%== maketext('Past Answers for [_1], set [_2], problem [_3]',
 					$studentUser, tag('span', dir => 'ltr', format_set_name_display($setName)), $prettyProblemNumber
 				) =%>
-			</h3>
+			</h2>
 			<div class="table-responsive">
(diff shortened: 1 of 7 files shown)


taniwallach (commented):
If recording this, maybe also save the PSVN which is an alternate value which PG can use to seed the RNG?

drgrice1:
I don't think that the problem set version number is needed in the past answer table.  The problem set version number is already saved in the user set.  That does not change for the user set, and so it does not also need to be saved in the past answer table, as it can always be retrieved as needed from the user set table.

templates/ContentGenerator/Instructor/ShowAnswers/past-answers-table.html.ep
```diff
@@ -27,8 +27,20 @@
...
+											'Seed: [_1]',
+											defined $record{seed} && $record{seed} ne ''
+												?  $record{seed} : maketext('unknown')
+										) %>
+									</small>
+								</td>
+							% }
 							<td class="px-3"><small><%= $c->formatDateTime($record{time}) %></small></td>
-							% for (my $i = 0; $i <= $upper_limit; $i++) {
+							% for (my $i = 0; $i <= $upper_limit; ++$i) {
```
Alex-Jordan, line 43: Does moving this increment operator to be a prefix make a difference here? Is it a preferable coding standard to use prefix when it doesn't make a difference?

Alex-Jordan (approved):

templates/ContentGenerator/Instructor/ShowAnswers/past-answers-table.html.ep
drgrice1: Pre-incrementation should always be used if the value prior to incrementing is not needed.  So in a `for` loop like this it should definitely always be pre-incrementation.  No matter what programing language you are working with pre-incrementation is more efficient.  Post-incrementation requires saving the original value, incrementing, and returning the original value.

pstaabp (approved):

PR #2276, opened 2023-12-09
Make SQL::Abstract work.

drgrice1:
This removes the optional dependence on SQL::Abstract::Classic or a version of SQL::Abstract prior to 2.000000, and only uses version 2.000000 or newer of SQL::Abstract.

For this to work the capability of a field override needs to be removed. There doesn't seem to be a reliable way to do that with the newer versions of SQL::Abstract, although I didn't try very hard.  I also don't see the need for having a field override.  The only field that has an override is the `key_not_a_keyfield` or `key` field.  Why was that ever done?  Did someone think that MySQL would make a field a database key field if it had the name "key"?  Did someone think it would be confusing to have a field named "key" that was not a database key field? It was probably the latter, but in my opinion the field override was even more confusing.

The `key` field in the `key` table does not contain any permanent data, so upgrading courses does not require any special handling.  The data from the old `key_not_a_keyfield` field can be safely dropped, and the new `key` field used in its place.  You can leave the `key_not_a_keyfield` field in the database if you want to switch back and forth between other pull requests that use the `key_not_a_keyfield` field.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -141,6 +141,7 @@ my @modulesList = qw(
 	Scalar::Util
 	SOAP::Lite
 	Socket
+	SQL::Abstract
 	Statistics::R::IO
 	String::ShellQuote
 	SVG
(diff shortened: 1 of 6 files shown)


pstaabp (approved):
Just for documentation, see #1236.  This appears to fix the script that wasn't working in that issue.

drdrew42 (approved):

PR #2277, opened 2023-12-11
Make the PG problem editor always regenerate dynamic images.

drgrice1:
Currently if a problem that uses dynamically generated images is edited and the code for generated an image changes the image is not regenerated unless you force it by adding `$refreshCachedImages = 1` to the problem itself.  This is problematic as an author may forget to remove that later which can cause unneeded image regeneration in production use.

The problem editor now forces image generation by setting `refreshCachedImages` in the PG environment.  For this to work a minor change is needed to PG.   See https://github.com/openwebwork/pg/pull/983.


drgrice1:
Closing in favor of #2283.

PR #2283, opened 2023-12-12
Include the problem source hash in the problemUUID passed to PG.

drgrice1:
If a problem is rendered using any of the source hash parameters (`problemSource`, `rawProblemSource`, or `uriEncodedProblemSource`), then an MD5 hex of the utf8 encoded source is added tot he problemUUID that is passed to PG.

This means that if a problem is rendered by source and the source changes, then dynamically generated images will be regenerated.  The PG problem editor uses the rawProblemSource, so this means that if the code for a problem changes in the code mirror window, then dynamic images will be regenerated.

This replaces #2277 and https://github.com/openwebwork/pg/pull/983. This is better as it reliably detects code changes, and it does not need a change to PG.  Thanks @drdrew42 for pointing this out.

### DIFF
--- a/lib/WeBWorK/Utils/Rendering.pm
+++ b/lib/WeBWorK/Utils/Rendering.pm
@@ -25,6 +25,8 @@ WeBWorK::Utils::Rendering - utilities for rendering problems.
 use Mojo::IOLoop;
 use Mojo::JSON qw(decode_json);
 use Data::Structure::Util qw(unbless);
+use Digest::MD5 qw(md5_hex);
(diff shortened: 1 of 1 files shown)


drgrice1:
By the way, another advantage of this is that is avoids a potential browser caching issue with the approach in #2277.  Since the problemUUID changes, the alias also changes (i.e., the link is to a completely different file).

Alex-Jordan (approved):
This looks good

pstaabp (approved):

PR #2288, opened 2023-12-17
offer to upgrade course after unarchiving it

Alex-Jordan:
This makes it so when you unarchive a course, you have an option to immediately go and upgrade it. Without this, after unarchiving a course, if it needs to be upgraded you must go to the Upgrade Courses page and wait while all courses are assessed for upgrade needs. But this lets you bypass that page and go straight to the confirmation page for only the one newly unarchived course.

Since now there are three things you can do (upgrade the new course, log in to the new course, and move on to unarchive something else) I changed the layout to look like three buttons in a row.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -1291,26 +1291,42 @@ sub do_unarchive_course ($c) {
 			),
 			$c->tag(
 				'div',
-				class => 'text-center',
-				$c->link_to(
-					$c->maketext('Log into [_1]', $new_courseID) => 'set_list' => { courseID => $new_courseID }
-				),
-			),
-			$c->form_for(
-				$c->current_route,
-				method => 'POST',
+				class => 'row',
 				$c->c(
-					$c->hidden_authen_fields,
-					$c->hidden_fields('subDisplay'),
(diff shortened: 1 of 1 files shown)


pstaabp (approved):

drgrice1 (commented):
This generally looks good, but I see a couple of minor problems.

First, the usage of the Bootstrap grid layout for the buttons doesn't look the best to me.  Mostly the issue is that the "Log into Course" button is too wide.  To fix this I suggest using a direct flex layout instead.  Change the class on the `div` containing the buttons (on line 1294) to `d-flex justify-content-between`, delete the `class` attribute on the forms contained therein (on lines 1299 and 1319), and remove the `col-4` class from the "Log into Course" button (on line 1314).

Second, there are duplicate id's created on the page by calling the ContentGenerator `hidden_authen_fields` method twice in the two forms for the "Upgrade Course" and "Unarchive More" buttons.  There is an optional id prefix argument to the `hidden_authen_fields` method that should be utilized to fix this.  So change line 1301 to something like `$c->hidden_authen_fields('upgrade_course_')` and line 1321 to something like `$c->hidden_authen_fields('unarchive_more_')`.  The actual prefixes used here don't matter as these id's are not used for anything in this case.  What matters is that they are different.

Alex-Jordan:
OK, I made the changes you suggested here.

drgrice1 (approved):
This looks good now.  Thanks for making the suggested changes.

PR #2291, opened 2023-12-19
option to 'clean' a course just after unarchiving it

Alex-Jordan:
This adds a checkbox to the unarchive course page, and if checked, then when the course is unarchived it will be "cleaned". Cleaning means:
* remove all student users and associated data from the database
* delete the log files
* delete any scoring files

Another small thing is that this removes the checkbox for giving a new course name. If you've entered something into the field for the new course name, that is enough, and the checkbox is redundant.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -27,7 +27,6 @@ use File::Path 'remove_tree';
 use Mojo::File;
 use File::stat;
 use Time::localtime;
-use String::ShellQuote;
 
 use WeBWorK::CourseEnvironment;
 use WeBWorK::Debug;
@@ -1290,7 +1289,7 @@ sub do_unarchive_course ($c) {
 			my $ce_new = WeBWorK::CourseEnvironment->new({ courseName => $new_courseID });
 			my $db_new = WeBWorK::DB->new($ce_new->{dbLayout});
 
-			for my $student_id ($db_new->listPermissionLevelsWhere({ permission => 0 })) {
+			for my $student_id ($db_new->listPermissionLevelsWhere({ permission => $ce->{userRoles}{student} })) {
 				$db_new->deleteUser($student_id->[0]);
 			}
 
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -26,6 +26,7 @@ use Net::IP;    # needed for location management
 use File::Path 'remove_tree';
 use File::stat;
 use Time::localtime;
+use String::ShellQuote;
 
 use WeBWorK::CourseEnvironment;
 use WeBWorK::Debug;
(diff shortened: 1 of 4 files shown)


lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -1275,14 +1276,55 @@ sub do_unarchive_course ($c) {
...
+							$c->c($c->tag('p', $c->maketext('Failed to remove file  [_1]:', $file)),
+								$c->tag('div', class => 'font-monospace', $rm_out))->join('')
+						);
+					}
+				}
+			}
+
+			if (-d "$ce_new->{courseDirs}{scoring}") {
+				my $rm_cmd =
+					"2>&1 $ce_new->{externalPrograms}{rm} -f " . shell_quote($ce_new->{courseDirs}{scoring}) . "/*";
```
pstaabp, line 1313: We're trying to remove externalPrograms.  This can be done with `Mojo::File`.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -1275,14 +1276,55 @@ sub do_unarchive_course ($c) {
...
+					return $c->tag(
+						'div',
+						class => 'alert alert-danger p-1 mb-2',
+						$c->c(
+							$c->tag('p',   $c->maketext('Failed to remove scoring files:')),
+							$c->tag('div', class => 'font-monospace', $rm_out)
+						)->join('')
+					);
+				}
+			}
```
drgrice1, line 1325: This can be switched to using `Mojo::File` quite succinctly with

```suggestion
				eval { Mojo::File->new($ce_new->{courseDirs}{scoring})->remove_tree({ keep_root => 1 }) };
				$c->addbadmessage($c->maketext('Failed to remove scoring files: [_1]', $@)) if $@;
```

Of course this presumes that `use Mojo::File` has been added to the beginning of the file.

Note this also should not `return` for the same reason as mentioned before.

drgrice1 (changes requested):
There are some things that I see that need to be fixed.  I have implemented all of these in a pull request that I will submit to this branch.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -1275,14 +1277,33 @@ sub do_unarchive_course ($c) {
...
 			)->join('')
 		);
 	} else {
 		writeLog($ce, 'hosted_courses', join("\t", "\tunarchived", '', '', "$unarchive_courseID to $new_courseID",));
 
+		if ($c->param('clean_up_course')) {
+			my $ce_new = WeBWorK::CourseEnvironment->new({ courseName => $new_courseID });
+			my $db_new = WeBWorK::DB->new($ce_new->{dbLayout});
+
+			for my $student_id ($db_new->listPermissionLevelsWhere({ permission => 0 })) {
```
drgrice1, line 1292: Sorry, one other thing that I missed before.  This shouldn't use a hardcoded permission value of 0.   It should use `$ce->{userRoles}{student}`.  So this should be
```suggestion
			for my $student_id ($db_new->listPermissionLevelsWhere({ permission => $ce->{userRoles}{student} })) {
```

Alex-Jordan:
I made the two edits that you caught, thanks.

pstaabp (approved):

PR #2292, opened 2023-12-21
do not allow LTI access to admin course

Alex-Jordan:
If the authentication module is an LTI authentication module and the course is the admin course, this moves on to the next authentication module.

This is to kill any potential for an LMS user to spoof their LTI data and enter the admin course as one of its admin users.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -45,6 +45,26 @@ $authen{user_module} = [
 	{ '*' => 'WeBWorK::Authen::Basic_TheLastOption' }    # fallback authorization method
 ];
 
+# List of authentication modules that may be used to enter the admin course.
+# This should be a non-empty sublist of whatever is in $authen{user_module}.
+# Since the admin course provides overall power to add/delete courses, access
+# to this course should be protected by the best possible authentication you
+# have available to you.  The current default is
+# WeBWorK::Authen::Basic_TheLastOption which is simple password based
+# authentication for a password locally stored in your WeBWorK server's
+# database.  On one hand, this is necessary as the initial setting, as it is the
+# only option available when a new server is being installed.  However, since
+# this option does not make use of multi-factor authentication or provide any
+# capabilities to prevent dictionary attacks, etc.  At the very least you should
+# use a very strong password.  If you have the option to use a more secure
+# authentication approach to the admin course (one which you are confident
+# cannot be spoofed) that is preferable.
+$authen{admin_module} = [
+	'WeBWorK::Authen::LTIAdvantage',
+	'WeBWorK::Authen::LTIAdvanced',
+	'WeBWorK::Authen::Basic_TheLastOption'
+];
+
 # Include configurations.  You must uncomment at least one of the following. You may uncomment
 # both if the site may be using both LTI 1.1 and 1.3 in different courses. After uncommenting
 # the LTI_1_x line, you must copy the file authen_LTI_1_x.conf.dist to authen_LTI_1_x.conf and
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -60,8 +60,8 @@ $authen{user_module} = [
 # authentication approach to the admin course (one which you are confident
 # cannot be spoofed) that is preferable.
 $authen{admin_module} = [
-	'WeBWorK::Authen::LTIAdvantage',
-	'WeBWorK::Authen::LTIAdvanced',
+	#'WeBWorK::Authen::LTIAdvantage',
+	#'WeBWorK::Authen::LTIAdvanced',
 	'WeBWorK::Authen::Basic_TheLastOption'
 ];
 
--- a/conf/authen_ldap.conf.dist
+++ b/conf/authen_ldap.conf.dist
@@ -12,6 +12,24 @@ $authen{user_module} = {
 	"*" => "WeBWorK::Authen::LDAP",
 };
 
+# List of authentication modules that may be used to enter the admin course.
+# This should be a non-empty sublist of whatever is in $authen{user_module}.
+# Since the admin course provides overall power to add/delete courses, access
+# to this course should be protected by the best possible authentication you
+# have available to you.  The current default is
+# WeBWorK::Authen::Basic_TheLastOption which is simple password based
+# authentication for a password locally stored in your WeBWorK server's
+# database.  On one hand, this is necessary as the initial setting, as it is the
+# only option available when a new server is being installed.  However, since
+# this option does not make use of multi-factor authentication or provide any
+# capabilities to prevent dictionary attacks, etc.  At the very least you should
+# use a very strong password.  If you have the option to use a more secure
+# authentication approach to the admin course (one which you are confident
+# cannot be spoofed) that is preferable.
+$authen{admin_module} = [
+	'WeBWorK::Authen::LDAP'
+];
+
 $authen{ldap_options} = {
 	# hosts to attempt to connect to, in order. For example:
 	#   auth.myschool.edu             -- uses LDAP scheme and port 389
--- a/conf/authen_ldap.conf.dist
+++ b/conf/authen_ldap.conf.dist
@@ -16,16 +16,7 @@ $authen{user_module} = {
 # This should be a non-empty sublist of whatever is in $authen{user_module}.
 # Since the admin course provides overall power to add/delete courses, access
 # to this course should be protected by the best possible authentication you
-# have available to you.  The current default is
-# WeBWorK::Authen::Basic_TheLastOption which is simple password based
-# authentication for a password locally stored in your WeBWorK server's
-# database.  On one hand, this is necessary as the initial setting, as it is the
-# only option available when a new server is being installed.  However, since
-# this option does not make use of multi-factor authentication or provide any
-# capabilities to prevent dictionary attacks, etc.  At the very least you should
-# use a very strong password.  If you have the option to use a more secure
-# authentication approach to the admin course (one which you are confident
-# cannot be spoofed) that is preferable.
+# have available to you.
 $authen{admin_module} = [
 	'WeBWorK::Authen::LDAP'
 ];
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -716,27 +716,28 @@ $modelCoursesForCopy = [ "modelCourse" ];
 # Select the authentication module to use for normal logins.
 #
 # If this value is a string, the given authentication module will be used
-# regardless of the database layout. If it is a hash, the database layout name
-# will be looked up in the hash and the resulting value will be used as the
-# authentication module. The special hash key "*" is used if no entry for the
-# current database layout is found.
-# If this value is a sequence of strings or hashes, then each
-# string or hash in the sequence will be successively tested to see if it
-# provides a module that can handle
-# the authentication request (by calling the module's
-#   sub  request_has_data_for_this_verification_module ).
-# The first module that responds affirmatively will be used.
+# regardless of the database layout.
 #
+# If it is a hash, the database layout name will be looked up in the hash and
+# the resulting value will be used as the authentication module.  The special
+# hash key "*" is used if no entry for the current database layout is found.
 #
-$authen{user_module} = {
-	# sql_moodle => "WeBWorK::Authen::Moodle",
-	# sql_ldap   => "WeBWorK::Authen::LDAP",
-	"*" => "WeBWorK::Authen::Basic_TheLastOption",
-};
+# If this value is a sequence of strings or hashes, then each string or hash in
+# the sequence will be successively tested to see if it provides a module that
+# can handle the authentication request by calling the module's sub
+# request_has_data_for_this_verification_module().  The first module that
+# responds affirmatively will be used.
+
+$authen{user_module} = {"*" => "WeBWorK::Authen::Basic_TheLastOption"};
 
 # Select the authentication module to use for proctor logins.
 # A string or a hash is accepted, as above.
 $authen{proctor_module} = "WeBWorK::Authen::Proctor";
+
+# List of authentication modules that may be used to enter the admin course.
+# This should always be an array reference with a subset of the modules named
+# in $authen{user_module}.
+$authen{admin_module} = ['WeBWorK::Authen::Basic_TheLastOption'];
 
 ################################################################################
 # Authorization system (Make local overrides in localOverrides.conf )
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -152,6 +152,18 @@ sub verify {
 	debug('BEGIN VERIFY');
 
 	return $self->call_next_authen_method if !$self->request_has_data_for_this_verification_module;
+	my $authen_ref = ref($self);
+	if ($c->ce->{courseName} eq $c->ce->{admin_course_id}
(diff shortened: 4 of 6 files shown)


taniwallach (changes requested):
I think this should be configurable. In general, I do agree that the worry is valid, but some institutions may trust LTI if they use MFA and other methods, while password based access to WeBWorK does not have MFA or any protection against dictionary attacks.

Alex-Jordan:
@taniwallach I made it configurable. Does the comment I left starting at line 722 sound right to you?

conf/defaults.config
```diff
@@ -710,15 +710,25 @@ $modelCoursesForCopy = [ "modelCourse" ];
 #
 #
 $authen{user_module} = {
-	# sql_moodle => "WeBWorK::Authen::Moodle",
-	# sql_ldap   => "WeBWorK::Authen::LDAP",
+	#sql_moodle => "WeBWorK::Authen::Moodle",
+	#sql_moodle => "WeBWorK::Authen::Moodle",
```
taniwallach, line 714: You have `WeBWorK::Authen::Moodle` twice, and dropped `WeBWorK::Authen::LDAP` accidentally.

conf/defaults.config
```diff
@@ -710,15 +710,25 @@ $modelCoursesForCopy = [ "modelCourse" ];
...
 	"*" => "WeBWorK::Authen::Basic_TheLastOption",
 };
 
 # Select the authentication module to use for proctor logins.
 # A string or a hash is accepted, as above.
 $authen{proctor_module} = "WeBWorK::Authen::Proctor";
 
+# Authentication modules that may be used to enter the admin course.
```
taniwallach, line 722: I would add
> Since the admin course provides overall power to add/delete courses, access to this course should be protected by the best possible authentication you have available to you. The current default is to use the `WeBWorK::Authen::Basic_TheLastOption` option which is simple password based authentication for a password locally stored in your WeBWorK server's database. On one hand, this is necessary as the initial setting, as it is the only option available when a new server is being installed. However, since this option does not make use of multi-factor authentication or provide any capabilities to prevent dictionary attacks, etc. at the very least you should use a very strong password. If you have the option to use a more secure authentication approach to the admin course (one which you are confident cannot be spoofed) - that is most likely preferable.

taniwallach (changes requested):
Suggestions made about the comment + 1 small error noted.
I have not tested the code yet.

Alex-Jordan:
@taniwallach Thanks for seeing that mistake and the suggestion. I fixed the mistake and lightly edited the suggestion.

somiaj:
@Alex-Jordan did you see my review? Anyways, I don't think `defaults.conf` is the correct place for that comment, and you should add that to files that are actually edited by users. You should probably also remove the commented out lines for Moodle and LDAP as well, since I don't think users should be editing defaults.conf.

somiaj:
@Alex-Jordan that is odd, I wonder why you cannot see my reviews. Anyways Tani caught the same thing I did, and you summarized my other one. I have attached some screenshots of what I see in this PR, so I'm confused as to why you (or maybe others) don't see them.

Okay, an error on my end, for some reason the review was just 'pending' and I had to click a finalize button, new to me, since I saw the review in the PR, I assumed others did too...my mistake.

Anyways, for new users I don't think looking in defaults.conf is a common thing to do. The instructions state making copies of the various .dist files and editing those, so options that only appear in defaults.config might get missed.

My original suggestion was just to follow what is currently being done with `$authen{user_module}`, and add this new option to the various auth files. For LDAP I suggest enabling it by default, but not for LTI. I could see users being confused if they wanted to use LDAP and for some reason it won't work for the admin course since the only info about a different setting is hidden in defaults.conf file.

Although it may not be new to have options hiding in defaults.conf with descriptions on them, to me I think the descriptions in defaults.conf should be more to developers, and the description for users and and options to edit should be in the .dist files that system admins are meant to edit when setting up their server.

Alex-Jordan:
I saw your review and comments came through this afternoon.

I updated it to put commented out options for all this in `localOverrides`. And I put uncommented options in each of the authentication module config files that match what those files set for `user_module`.

More feedback is welcome.

conf/authen_LTI.conf.dist
```diff
@@ -45,6 +45,26 @@ $authen{user_module} = [
...
+# this option does not make use of multi-factor authentication or provide any
+# capabilities to prevent dictionary attacks, etc.  At the very least you should
+# use a very strong password.  If you have the option to use a more secure
+# authentication approach to the admin course (one which you are confident
+# cannot be spoofed) that is preferable.
+$authen{admin_module} = [
+	'WeBWorK::Authen::LTIAdvantage',
+	'WeBWorK::Authen::LTIAdvanced',
+	'WeBWorK::Authen::Basic_TheLastOption'
+];
```
somiaj, line 66: Do you want to comment out the LTI methods here, to make default behavior LTI cannot be used to access the admin course (like your original intent was). I think this as reasonable since LTI can also create student users (they can't do much in the admin course), but that could be another reason it should be disabled by default?

somiaj:
@Alex-Jordan I figured out what I did wrong, and submitted my old review just so you could see the comments and I wouldn't have to copy them again. I figured out what I was doing incorrectly, there is both a 'enter single comment' and 'start a review', the second needs to be finalized. I have only left single comments in the past so didn't realize that 'start a review' as an option until now.

somiaj:
In localOverrides.conf you added the commented out authentication options before the includes that include the auth_LTI.conf and auth_LDAP.conf, which have the same options not commented out. So if a user removes the comments to configure the auth methods there, they will just be overridden by what is in the last auth_Foo.conf file included. I think you should move these to after the includes, so they can be used to override anything the configuration files may be doing when a user removes the comments.

conf/authen_LTI.conf.dist
Alex-Jordan: Now that it's configurable, it feels OK to let the default behavior go unchanged from what it has been up to now. I don't really have an opinion though. In my own servers, I'll make sure that LTI can't be used to access the admin course either way.

Alex-Jordan:
> if a user removes the comments to configure the auth methods there, they will just be overridden by what is in the last auth_Foo.conf file included.

My perspective is that is exactly what should happen. The authentication module config files are "loaded last" conceptually, even if they come in the middle of `localOverrides.conf`. In some installations, the authentication module config files won't even be loaded until somewhere in the `course.conf` file.

In the case of LDAP, it is odd to have a commented out line that suggests using it (in `localOverrides.conf` at line 464) before loading the LDAP config file (at line 512). But what is at line 464 is just what was previously in `defaults.config`, which also came before the loading of the LDAP config file.

drgrice1 (approved):
This looks good to me at this point.

Alex-Jordan:
This is rebased now, I think just awaiting @taniwallach's review.  I did lose track if I got to everything @somiaj mentioned, and please alert me if there is more to adjust.

somiaj:
I recall you addressed everything I mentioned, but I am going to mention it again just in case. Note, I'm okay with things the way they are, I'm just sharing my observations on hypotheticals.

I would still remove LTI access by default (maybe comment it out) in the admin course. This seems like a reasonable default to me, vs just allowing admins to remove it. Sure it changes current behavior, but I think for the better.

I still think the order in which the config options are read should be addressed a bit more. Or maybe add comments about this, to avoid possible confusing. Currently if someone wanted to override the settings in `localOverrides.conf` and uncomments/edits the options there, they will be overridden in `authen_LTI.conf`, `authen_LDAP.conf`, etc, without first modifying those files. For my setup, I would have to copy/paste the settings in `localOverrides.conf` and move them (so I can use both LTI and LDAP) to below the include lines.

I did understand your argument that in some cases that the config files may not be read until `course.conf`, so probably not an easy answer for this.

Alex-Jordan:
I think you are right that it would be good to change the default. I wonder if there is anyone it would even effect. Who is currently intentionally accessing their admin course using LTI? So I will make that change unless someone else stops me.

As we saw with a recent forum thread, the situation with the config files for authentication is a mess. Some solution may be needed that is beyond the scope here.

lib/WeBWorK/Authen.pm
```diff
@@ -189,6 +189,12 @@ sub verify {
 		return ($self->call_next_authen_method());
 	}
 
+	my $authen_ref = ref($c->authen);
+	if ($c->ce->{courseName} eq 'admin' && !(grep(/^$authen_ref$/, @{ $c->ce->{authen}{admin_module} }))) {
+		$c->stash(authen_error => maketext("Cannot authenticate into admin course using $authen_ref."));
```
drgrice1, line 194: This should be 
```suggestion
		$c->stash(authen_error => $c->maketext("Cannot authenticate into admin course using $authen_ref."));
```
since the `WeBWorK::Authen` module does not have a `maketext` method.

conf/authen_ldap.conf.dist
```diff
@@ -12,6 +12,24 @@ $authen{user_module} = {
...
+# have available to you.  The current default is
+# WeBWorK::Authen::Basic_TheLastOption which is simple password based
+# authentication for a password locally stored in your WeBWorK server's
+# database.  On one hand, this is necessary as the initial setting, as it is the
+# only option available when a new server is being installed.  However, since
+# this option does not make use of multi-factor authentication or provide any
+# capabilities to prevent dictionary attacks, etc.  At the very least you should
+# use a very strong password.  If you have the option to use a more secure
+# authentication approach to the admin course (one which you are confident
+# cannot be spoofed) that is preferable.
```
drgrice1, line 28: This comment is not accurate for this authentication module.  LDAP does not use a default of `WeBWorK::Authen::Basic_TheLastOption` for the `$authen{admin_module}`.  Furthermore, LDAP does not use a password.

The same is true for the comment added in `conf/authen_CAS.conf.dist`.

lib/WeBWorK/Authen.pm
drgrice1, line 194: Actually, this should be
```suggestion
		$self->write_log_entry("Cannot authenticate into admin course using $authen_ref.");
		$c->stash(
			authen_error => $c->maketext(
				'There was an error during the login process. '
					. 'Please speak to your instructor or system administrator.'
			)
		);
```
The current `authen_error` gives away to much information about the inner workings of the system authentication methods.  Never display the actual authentication module in the UI.  The error that is displayed should be generic.  I know that you may want to give some hint to the admin user as to the reason that authentication failed, but you just can't do that.

The log message could be tweaked a bit further probably though.  In the log files the first message is probably not very well worded.  Perhaps something like "An authentication attempt using the $authen_ref module was denied."  Note that log message goes to the admin courses `logs/login.log` file (and to the debug log if debugging is enabled).

Alex-Jordan:
@drgrice1 I believe I addressed the recent things you mentioned here.

drgrice1:
It looks like you have addressed the issues that have been put forth.

@taniwallach, could you look at this, and if you are satisfied give it an approval to remove the block on merging?

Alex-Jordan:
@drgrice1 This has a conflict now following the merge of #2333. Since I don't follow everything that you are changing in #2333, would you mind looking this over and suggesting how to address the conflict?

And then @taniwallach, following addressing the conflict, can you look at this and lift your block on merging?

drgrice1:
Yeah, I will look at it.

lib/WeBWorK/Authen.pm
```diff
@@ -189,6 +189,17 @@ sub verify {
...
+	my $authen_ref = ref($c->authen);
+	if ($c->ce->{courseName} eq 'admin' && !(grep(/^$authen_ref$/, @{ $c->ce->{authen}{admin_module} }))) {
+		$self->write_log_entry("Cannot authenticate into admin course using $authen_ref.");
+		$c->stash(
+			authen_error => $c->maketext(
+				'There was an error during the login process.  Please speak to your instructor or system administrator.'
+			)
+		);
+		return ($self->call_next_authen_method());
+	}
```
drgrice1, line 201: The easiest way to fix the conflict here is to merge `origin/develop` into this branch.  When you do that the following merge conflict will occur:
```
<<<<<<< HEAD
	my $authen_ref = ref($c->authen);
	if ($c->ce->{courseName} eq 'admin' && !(grep(/^$authen_ref$/, @{ $c->ce->{authen}{admin_module} }))) {
		$self->write_log_entry("Cannot authenticate into admin course using $authen_ref.");
		$c->stash(
			authen_error => $c->maketext(
				'There was an error during the login process.  Please speak to your instructor or system administrator.'
			)
		);
		return ($self->call_next_authen_method());
	}

	my $result    = $self->do_verify;
	my $error     = $self->{error};
	my $log_error = $self->{log_error};
=======
	return $self->call_next_authen_method if !$self->request_has_data_for_this_verification_module;
>>>>>>> origin/develop
```
Just delete the version control conflict markers and move `return $self->call_next_authen_method if !$self->request_has_data_for_this_verification_module;` before the changes you added here.

Although, I also see some code issues that need to be fixed.  I will comment on those separately.

lib/WeBWorK/Authen.pm
```diff
@@ -189,6 +189,17 @@ sub verify {
 		return ($self->call_next_authen_method());
 	}
 
+	my $authen_ref = ref($c->authen);
```
drgrice1, line 192: This should be
```suggestion
	my $authen_ref = ref($self);
```
`$self` is the same thing as `$c->authen` here.

lib/WeBWorK/Authen.pm
```diff
@@ -189,6 +189,17 @@ sub verify {
...
+	if ($c->ce->{courseName} eq 'admin' && !(grep(/^$authen_ref$/, @{ $c->ce->{authen}{admin_module} }))) {
```
drgrice1, line 193: This needs to be updated to use the `admin_course_id` from the course environment.  Also, this should not use the expression form of grep.  So this should be
```suggestion
	if ($c->ce->{courseName} eq $c->ce->{admin_course_id}
		&& !(grep {/^$authen_ref$/} @{ $c->ce->{authen}{admin_module} }))
	{
```

lib/WeBWorK/Authen.pm
```diff
@@ -189,6 +189,17 @@ sub verify {
...
+		$self->write_log_entry("Cannot authenticate into admin course using $authen_ref.");
+		$c->stash(
+			authen_error => $c->maketext(
+				'There was an error during the login process.  Please speak to your instructor or system administrator.'
+			)
+		);
+		return ($self->call_next_authen_method());
```
drgrice1, line 200: There is no need for the additional parentheses around the return value.  So this should be
```suggestion
		return $self->call_next_authen_method();
```

drgrice1 (commented):
Here are the things that I saw before.

Alex-Jordan:
OK, I made all those updates. @taniwallach this is ready for you to review.

somiaj (approved):
Looked over things again and looks good. I tested that by default I get the error `There was an error during the login process.  Please speak to your instructor or system administrator.` when trying to login via LTI by default (so it is correctly not letting me log in via LTI), and if I add `WeBWorK::Authen::LTIAdvanced` to `$authen{admin_module}` then it tries to login via LTI (though I get an error since I didn't have LTI setup correctly to login, but see via the debug parameters it attempted to).

drgrice1:
Since this now has two approvals, I will override @taniwallach disapproval, and merge it.  I believe that his issues have been addressed, and that he just has not had the time to look at this.

Issue #1991, opened 2023-05-18
Remove "which users" from Classlist Editor and "which sets" from Homework Sets Editor?


dlglin:
I guess this is a follow-up to #1480 

Now that there is a javascript "select all" on both the Classlist Editor and Hmwk Sets Editor, do we really need a drop-down for "which users" or "which sets" on the respective pages?  On the tabs that have such an option can we get rid of the drop-down and make the only option to be that the action applies to selected sets/users?  If the instructor wants to apply the action to all users or all shown users they can filter, then click the select all button.

I think that this would be a cleaner user experience.  Right now it's a little unclear that if you have a filtered list of users/sets, and you select "all users/sets" the action is being done to sets that you don't see.

pstaabp:
This was addressed in #2294

PR #2294, opened 2023-12-31
Remove action scope and add action form validation.

somiaj:
On the UserList, ProblemSetList, and AchievementList managers, remove the scope option that helps determine which items to act on, instead users will always select which items to act on and can use filters to change the list of items to select from. This address #1991.

In addition add javascript form validation that will create a modal popup if the form is missing information, such as no items are selected, a text string is not provided, a valid file is not selected, and so on.

Last, make the import tabs not create a form if no valid files are found to import from.

### DIFF
--- a/htdocs/js/AchievementList/achievementlist.js
+++ b/htdocs/js/AchievementList/achievementlist.js
@@ -1,0 +1,76 @@
+(() => {
+	// Action form validation.
+	const is_achievement_selected = () => {
+		for (const achievement of document.getElementsByName('selected_achievements')) {
+			if (achievement.checked) return true;
+		}
+		document.getElementById('select_achievement_err_msg')?.classList.remove('d-none');
+		document.getElementById('achievement-table')?.addEventListener('change', e => {
+			document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
+		}, { once : true });
+		return false;
+	};
+
+	document.getElementById('achievement-list')?.addEventListener('submit', e => {
+		const action = document.getElementById('current_action')?.value || '';
+		if (['edit', 'assign', 'export', 'score'].includes(action)) {
+			if (!is_achievement_selected()) {
+				e.preventDefault();
+				e.stopPropagation();
+			}
+		} else if (action === 'import') {
+			const import_file = document.getElementById('import_file_select');
+			if (!import_file.value.endsWith('.axp')) {
+				e.preventDefault();
+				e.stopPropagation();
+				document.getElementById('import_file_err_msg')?.classList.remove('d-none');
+				import_file.classList.add('is-invalid');
+				import_file.addEventListener(
+					'change',
+					() => {
+						document.getElementById('import_file_err_msg')?.classList.add('d-none');
+						document.getElementById('import_file_select')?.classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
+			}
+		} else if (action === 'create') {
+			const create_text = document.getElementById('create_text');
+			if (create_text.value === '') {
+				e.preventDefault();
+				e.stopPropagation();
+				document.getElementById('create_file_err_msg')?.classList.remove('d-none');
+				create_text.classList.add('is-invalid');
+				create_text.addEventListener('change', e => {
+					document.getElementById('create_file_err_msg')?.classList.add('d-none');
+					document.getElementById('create_text')?.classList.remove('is-invalid');
+				}, { once : true });
+			} else if (document.getElementById('create_select')?.selectedIndex == 1 && !is_achievement_selected()) {
+				e.preventDefault();
+				e.stopPropagation();
+			}
+		} else if (action === 'delete') {
+			const delete_confirm = document.getElementById('delete_select');
+			if (!is_achievement_selected()) {
+				e.preventDefault();
+				e.stopPropagation();
+			} else if (delete_confirm.value != 'yes') {
+				e.preventDefault();
+				e.stopPropagation();
+				document.getElementById('delete_confirm_err_msg')?.classList.remove('d-none');
+				delete_confirm.classList.add('is-invalid');
+				delete_confirm.addEventListener('change', e => {
+					document.getElementById('delete_select')?.classList.remove('is-invalid');
+					document.getElementById('delete_confirm_err_msg')?.classList.add('d-none');
+				}, { once : true });
+			}
+		}
+	});
+
+	// Remove select error message when changing tabs.
+	for (const tab of document.querySelectorAll('a[data-bs-toggle="tab"]')) {
+		tab.addEventListener('shown.bs.tab', e => {
+			document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
+		}, { once : true });
+	}
+})();
--- a/htdocs/js/AchievementList/achievementlist.js
+++ b/htdocs/js/AchievementList/achievementlist.js
@@ -5,13 +5,17 @@
 			if (achievement.checked) return true;
 		}
 		document.getElementById('select_achievement_err_msg')?.classList.remove('d-none');
-		document.getElementById('achievement-table')?.addEventListener('change', e => {
-			document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
-		}, { once : true });
+		document.getElementById('achievement-table')?.addEventListener(
+			'change',
+			() => {
+				document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
+			},
+			{ once: true }
+		);
 		return false;
 	};
 
-	document.getElementById('achievement-list')?.addEventListener('submit', e => {
+	document.getElementById('achievement-list')?.addEventListener('submit', (e) => {
 		const action = document.getElementById('current_action')?.value || '';
 		if (['edit', 'assign', 'export', 'score'].includes(action)) {
 			if (!is_achievement_selected()) {
@@ -41,10 +45,14 @@
 				e.stopPropagation();
 				document.getElementById('create_file_err_msg')?.classList.remove('d-none');
 				create_text.classList.add('is-invalid');
-				create_text.addEventListener('change', e => {
-					document.getElementById('create_file_err_msg')?.classList.add('d-none');
-					document.getElementById('create_text')?.classList.remove('is-invalid');
-				}, { once : true });
+				create_text.addEventListener(
+					'change',
+					() => {
+						document.getElementById('create_file_err_msg')?.classList.add('d-none');
+						document.getElementById('create_text')?.classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
 			} else if (document.getElementById('create_select')?.selectedIndex == 1 && !is_achievement_selected()) {
 				e.preventDefault();
 				e.stopPropagation();
@@ -59,18 +67,26 @@
 				e.stopPropagation();
 				document.getElementById('delete_confirm_err_msg')?.classList.remove('d-none');
 				delete_confirm.classList.add('is-invalid');
-				delete_confirm.addEventListener('change', e => {
-					document.getElementById('delete_select')?.classList.remove('is-invalid');
-					document.getElementById('delete_confirm_err_msg')?.classList.add('d-none');
-				}, { once : true });
+				delete_confirm.addEventListener(
+					'change',
+					() => {
+						document.getElementById('delete_select')?.classList.remove('is-invalid');
+						document.getElementById('delete_confirm_err_msg')?.classList.add('d-none');
+					},
+					{ once: true }
+				);
 			}
 		}
 	});
 
 	// Remove select error message when changing tabs.
 	for (const tab of document.querySelectorAll('a[data-bs-toggle="tab"]')) {
-		tab.addEventListener('shown.bs.tab', e => {
-			document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
-		}, { once : true });
+		tab.addEventListener(
+			'shown.bs.tab',
+			() => {
+				document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
+			},
+			{ once: true }
+		);
 	}
 })();
--- a/htdocs/js/AchievementList/achievementlist.js
+++ b/htdocs/js/AchievementList/achievementlist.js
@@ -79,14 +79,16 @@
 		}
 	});
 
-	// Remove select error message when changing tabs.
+	// Remove all error messages when changing tabs.
 	for (const tab of document.querySelectorAll('a[data-bs-toggle="tab"]')) {
-		tab.addEventListener(
-			'shown.bs.tab',
-			() => {
-				document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
-			},
-			{ once: true }
-		);
+		tab.addEventListener('shown.bs.tab', () => {
+			const actionForm = document.getElementById('achievement-list');
+			for (const err_msg of actionForm.querySelectorAll('div[id$=_err_msg]')) {
+				err_msg.classList.add('d-none');
+			}
+			for (const invalid of actionForm.querySelectorAll('.is-invalid')) {
+				invalid.classList.remove('is-invalid');
+			}
+		});
 	}
 })();
--- a/htdocs/js/AchievementList/achievementlist.js
+++ b/htdocs/js/AchievementList/achievementlist.js
@@ -9,6 +9,9 @@
 			'change',
 			() => {
 				document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
+				for (const id of ['edit_select', 'assign_select', 'export_select', 'score_select']) {
+					document.getElementById(id)?.classList.remove('is-invalid');
+				}
 			},
 			{ once: true }
 		);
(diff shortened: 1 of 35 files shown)


htdocs/js/ActionTabs/actiontabs.js
```diff
@@ -45,4 +45,57 @@
 			});
 		}
 	}
+
+	// Error modal popup added to window so it can be accessed from other modules.
```
drgrice1, line 49: Instead of making this a global variable and polluting the global namespace (which is what adding it to the `window` object does), use the `webworkConfig` object.  The `webworkConfig` variable was defined to limit webwork2's pollution of the global namespace to just one variable.  So change this from `window.takeActionErrorModal` to `webworkConfig.takeActionErrorModal`, and call it that way as well.  The `webworkConfig` is defined in `htdocs/themes/math4/system.html.ep` and is available for all of webwork2's javascript.  It is already used in this way for the `renderProblem` method defined in `htdocs/js/RenderProblem/renderproblem.js`.

Note that this is not accessed by other modules (as your comment states), but by other javascript.  If we actually had javascript modules this could be done in a much better way by exporting this method here, and importing it where needed.  I have implemented javascript modules for webwork2, but am holding off on it for a bit for certain reasons.  It will be needed sooner or later though.  Note that one major advantage of using javascript modules is that with that we will be able to have direct javascript translations using the i18next javascript module instead of using the messy dataset hack to pass translations from perl to javascript.

drgrice1 (changes requested):
Here are a few things that I noticed with an initial look at the javascript files.

There is an issue with this pull request in general though.  Basically, this is a reason that issue #1991 is not entirely valid.  At least for the Accounts Manager page in part.  I have been meaning to add a comment to that issue about this.

If there are more than 200 users in a course, then no users are shown on that page by default.  So when you select the "Edit" tab, there is initially no way to select users to edit (for example) with this pull request.  You now have to go to the filter tab and get some users to be shown on the page before you can do anything.  That is not ideal.  Previously when none of the users were selected, you could still edit "all users".

htdocs/js/UserList/userlist.js
```diff
@@ -24,4 +24,62 @@
 		export_select_target.addEventListener('change', classlist_add_export_elements);
 		classlist_add_export_elements();
 	}
+
+	// Action form validation.
+	const is_user_selected = () => {
+		const users = document.getElementsByName('selected_users');
+		for (let i = 0; i < users.length; i++) {
+			if (users[i].checked) { return true; }
+		}
```
drgrice1, line 33: C-style for loops should not be used in modern javascript unless explicitly needed.  Iterating over the elements in an array is not a case where a C-style for loop is explicitly needed.  Change this to:

```suggestion
		for (const user of document.getElementsByName('selected_users')) {
			if (user.checked) return true;
		}
```

Note that the intermediate `users` variable is not needed with this approach.

Also, note the removal of the braces around `return true;`.  Those aren't needed for a single statement inside an `if`.

Please make these changes in the `problemlist.js` and `achievementlist.js` files as well.

drgrice1 (commented):
This was supposed to be in the last review, but I must not have clicked the button to add it.

drgrice1:
Also note that using a modal dialog for form validation messages is a rather intrusive and ugly way of doing so.  You should take a look at https://getbootstrap.com/docs/5.3/forms/validation/ for a better way of doing this sort of thing.

somiaj:
Updated the javascript and moved the validation from a modal to in-form messaging and toggling the `is-invalid` style.

@drgrice1 I noticed the extra step when I was testing a class with 500+ users of having to first filter before acting on all users. I wasn't sure how big of an issue it would be, with a large class is it common to act on all users or first filter? Another option could be to leave the scope menus in the few places they are beneficial, such as the example stated.

somiaj:
I also moved to disabling the import tab if no valid files are found to import from vs show a different form.

drgrice1:
> Updated the javascript and moved the validation from a modal to in-form messaging and toggling the `is-invalid` style.

I like this better.  It is less intrusive, and it is good that you clear the validation message when the tabs change.  Perhaps this could be taken further though and this could be used for the other validation messages also.  For example, if the "Filter" tab is used with field matching and no text is entered that message could also be shown in this hidden div (and cleared when the tabs change).

> @drgrice1 I noticed the extra step when I was testing a class with 500+ users of having to first filter before acting on all users. I wasn't sure how big of an issue it would be, with a large class is it common to act on all users or first filter? Another option could be to leave the scope menus in the few places they are beneficial, such as the example stated.

You generally don't want to edit all users if the course is large enough (say 5000 or more) as the page takes to long to load, and who would seriously want to manually edit that many users at one time anyway.  So filtering is almost a necessity.  I think that pagination should be used for dealing with large numbers of users instead of just not showing any.  With proper pagination, all users (to which the current filter applies) could be available in pages and you could select users from multiple pages.

htdocs/js/AchievementList/achievementlist.js
```diff
@@ -0,0 +1,60 @@
+(() => {
+	// Action form validation.
+	const is_achievement_selected = () => {
+		for (const achievement of document.getElementsByName('selected_achievements')) {
+			if (achievement.checked) return true;
+		}
+		document.getElementById('select_achievement_err_msg')?.classList.remove('d-none');
+		document.getElementById('achievement-table')?.addEventListener('change', e => {
+			document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
+		}, { once : true });
```
drgrice1, line 10: The variable `e` is not used here.  So this should be
```suggestion
		document.getElementById('achievement-table')?.addEventListener('change', () => {
			document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
		}, { once : true });
```
This is done several times in the javascript changes that you have made.  So make this change in those places as well.

htdocs/js/AchievementList/achievementlist.js
```diff
@@ -0,0 +1,60 @@
...
+		return false;
+	};
+
+	document.getElementById('achievement-list')?.addEventListener('submit', e => {
```
drgrice1, line 14: Since `e` is used here it is of course needed, but it should still have parentheses around it.  It is just clearer that way that it is a parameter for the arrow function.  This is a common requirement of many style guides.

drgrice1 (commented):
If you know how to do this, could you apply `prettier` to the javascript files that you have edited.  I have started doing this when I change javascript files, and eventually I plan to add a workflow to check this (much like we use perltidy for perl files).

The contents of the `.prettierrc` file that I use for this (and plan to use for the workflow) are:
```
{
	"arrowParens": "always",
	"bracketSpacing": true,
	"printWidth": 120,
	"semi": true,
	"singleQuote": true,
	"tabWidth": 4,
	"trailingComma": "none",
	"useTabs": true
}
```

htdocs/js/AchievementList/achievementlist.js
somiaj: Thanks, I got confused which was preferred and edited the files the wrong direction. Running prettier seems to have found all the instances.

somiaj:
I updated the javascript to remove all the error messages when switching tabs.

htdocs/js/AchievementList/achievementlist.js
drgrice1: Yeah, I ran prettier and that is what led to this comment, and the general comment regarding prettier usage.

somiaj:
Last commit removes `none` from the list of options to filter from. Since you need some users/sets to act on, I don't think this option is useful. I can revert if others think we should keep it.

drgrice1:
There is another case on the UserList page (the accounts manager) that the removal of the scope selector is particularly bad with for large courses.  That is the "Export" tab.  Now if you have a large class and you want to export all users, the only way to do this is to first show all users and then select them all.  For really large classes that can take a really long time for the browser to load all of that.  Previously you could just choose the export "all users" option.

I think in general removing these scope selectors for the accounts manager is not a good idea, and will need to be added back.  For the sets manager and the achievements manager it is fine, since those lists are usually not that large.

somiaj:
I added the action scope back to the Edit, Password, and Export actions on the UserList. In this case it only has two options, either all or selected, since there is a select all checkbox for visible users. Though in practice it will probably only be the export option this is most likely used, since manually editing or changing the password of all users in a class of >200 isn't something that is probably done that often.

templates/ContentGenerator/Instructor/UserList/import_form.html.ep
```diff
@@ -1,11 +1,9 @@
-% use WeBWorK::Utils::Instructor qw(getCSVList);
-%
 <div>
 	<div class="row mb-2">
 		<%= label_for import_select_source => maketext('Import users from what file?'),
 			class => 'col-form-label col-form-label-sm col-sm-auto' =%>
 		<div class="col-auto">
-			<%= select_field 'action.import.source' => [ getCSVList($ce) ],
+			<%= select_field 'action.import.source' => [ @$CSVList ],
```
drgrice1, line 6: This would be better as
```suggestion
			<%= select_field 'action.import.source' => $CSVList,
```
Currently it is actually copying the array that is referenced by `$CSVList` and creating a new reference to the copy which is unnecessary.

somiaj: Thanks.

drgrice1 (approved):
I think this is okay now.  I agree that the "scope" selector is most important for the export form of the user list page.  Although I think it should be there for the edit and password forms as well due to the fact that there is the case where all users are hidden by default for large courses.

somiaj:
I did consider only showing the scope drop downs to Edit, Password, and Export users when less than the full list of students is shown, though unsure if that is useful.

drgrice1:
At least for the export page I think that the option to select all users for export via the dropdown should always be there.  Even if the class is not large. this makes sense as it allows not needing to deal with the checkboxes at all.

somiaj:
That logic could be extended to any place that scope use to be used, though export is the place it is most likely used.

As an alternative to removing scope completely, would be just reduce it to `all` and `selected`. Having `none` there isn't useful, and the checkboxes take care of the `visible` option. This would also be useful if you have done a filter, but want to quickly act on the full class without having to reset the filter. I will work on adding back this limited scope approach.

@dlglin do you have any input from your intent behind #1991?

pstaabp (approved):
I'm good with this.  @drgrice1 I know you approved, but okay with merging now or any other changes?

drgrice1:
Lets wait a bit on this.  There are still open questions.

somiaj:
I have updated this so that both the Accounts and Sets manager still have action scope, but it is limited to two options, either 'all' or 'selected' and 'selected' is always default. In essence only 'visible' and 'none' was removed (since this can be done via the checkboxes). My thoughts here are that on the Accounts and Sets manager can filter the options, so it could be useful to have a quick way to act on all users without having to update the filter (along with large classes not showing any users without filtering).

I have a patch to also add this back to the achievements manager too, but didn't add it here. Logic being there is no filter on the achievements page, so the checkboxes can be used to act on all achievements at any time. If others prefer to have the scope added back here too, let me know and I'll add that patch.

It would be nice if @dlglin weighs in on the original intent behind #1991

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
```diff
@@ -198,6 +189,7 @@ sub initialize ($c) {
 	$c->stash->{fieldTypes}     = FIELD_TYPES();
 	$c->stash->{sortableFields} = SORTABLE_FIELDS();
 	$c->stash->{sets}           = [];
+	$c->stash->{setDefList}     = [ getDefList($ce) ];
```
drgrice1, line 192: I don't exactly like that this is done here.  The issue is that the `getDefList` method is an expensive operation (not nearly as expensive as it used to be when it searched the Library and Contrib directories, but still expensive in that it still involves a directory search), and with it here, it is called even if the import form is not going to be displayed (for instance if the user does not have the permissions checked below.  It would be better to set this to the empty array here, and then update it later when it is known that this is going to be used.

The same is true for the csv list on line 258 of `lib/WeBWorK/ContentGenerator/Instructor/UserList.pm` and the axp list on line 85 of `lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm`.  Those are perhaps not as expensive as the set def list, but they still involve file IO that should be avoided in cases where it is not needed.

drgrice1 (commented):
I think that this is better now.  Allowing acting on all sets or users regardless of selection for the accounts and sets manager pages is good.

There is one detail that needs attention, and then this can be merged.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm
somiaj: I updated it so the list is only generated if the import form could be shown.

drgrice1: This is good now.  I have no further reservations with this pull request.

dlglin:
When I posted #1991 I wasn't thinking about large classes where all students are not automatically displayed, so I see @drgrice1's point on it being extra work to have to show the whole list before exporting.  My thought was to simplify the interface, since right now the user has to mess with two things: the dropdown and the checkboxes to get the subset of students/sets that they want.

There is one thing that I find counterintuitive with this setup: if you filter the list and then perform an action (e.g. export), you are given the option of "all users", which leads to exporting users that are not currently being shown.

I'm not sure if the following suggestion is too complicated, but would it make sense to have the options be "all" and "selected" when no filter is applied, but replace "all" with "displayed" if the list is filtered (with appropriate logic to only apply the action to the shown items)?  Maybe that's beyond the scope of this PR.

drgrice1:
I personally find that counterintuitive.  To me it seems clear that "all" users means all users from the course, and not all visible users.  Furthermore, I think the "all" option should always be there.  If you do as you suggest, and make that switch to "displayed" when a filter is applied, then you are back to the how this initially was and you add steps needed to export all users again if you have previously applied a filter.  That also seems counter to your initial issue that led to this pull request.  What you can do with the "displayed" option can always be done with the check boxes.

dlglin:
I'm having trouble envisioning a scenario where you would apply a filter, and then subsequently perform an action on all users.  My intuition is that once I've filtered I'm only working on that subset.  If I want to do something to all users I would expect that I would remove the filter.

drgrice1:
I can.  You might want to edit some users and so apply a filter to find those users in your large class.  After doing so you might want to export all users.

There is also the issue that there really isn't a way to remove a filter other than to start from scratch by reloading the page.

Alex-Jordan:
Can we do this with more clear labeling? Instead of "all" or "all users", something like "all course users" would (to me) clearly imply everyone, not just all the ones being displayed.

drgrice1:
I am fine with that.

Alex-Jordan (approved):
I just tried this and it looks good. I suggest changing "all users" to "all course users" as mentioned in the thread, but mark for approval either way.

somiaj:
@Alex-Jordan Should `All sets` be kept as is, or do you suggest updating that as well?

somiaj:
I just noticed an issue with the validation javascript. It requires a user to be selected even when 'all course users' is selected. Fixing that now, so don't merge.

Alex-Jordan:
> @Alex-Jordan Should All sets be kept as is, or do you suggest updating that as well?

Oh, I realize now I was only looking in the Accounts Manager, not several other places this impacts. But sure, "all course sets" seems fine. Just looking for clarity between literally all of them versus all that are presently displayed.

Does "all course achievements" come up too?

somiaj:
Because achievements don't have a filter action, there is no option to select between `all course achievements` and `selected achievements`, since the checkboxes already deal with this case. I have a patch to add that option to the achievements page if desired for consistency, but currently is not there.

Alex-Jordan:
> Because achievements don't have a filter action, there is no option to select between all course achievements and selected achievements, since the checkboxes already deal with this case. I have a patch to add that option to the achievements page if desired for consistency, but currently is not there.

I'd like to have that. Of course, it can be separate from this PR. The reason I'd like it is that I often load the default collection of achievements, but then go and delete the Challenge achievements, plus a few more. Filtering would just make that a little more smooth. (I should just export the sets that I actually use into an achievement collection file, but then I still have colleagues to mentor, etc.)

drgrice1:
Since a filter on the achievements list page is new, that is something that I think should be in a separate pull request.

somiaj:
I can not remove the action scope from achievements here in anticipation of such a change, but agree it should be in a different PR.

dlglin:
Another observation: when I first looked at this I was confused as to why the Import tab was greyed out.  Is there a good way to let the user know that in order to import Sets/Users that they need to first upload a file via the file manager?

It would be nice if the instructor could upload a .def/.lst file here directly, but that's beyond the scope of this PR, so it's captured in #2325.

drgrice1:
I looked this over another time to be sure, but I think this pull request is ready.

As to @dlglin's comment about the import tab being grayed out, perhaps a comment could be added to the help file about that.  I don't think that is a particularly serious issue though, as the typical course will have `.def` files in the templates directory, since some are copied from the model course in the default course creation process.

I personally think that import/export should ONLY be done the way that @dlglin suggested.  Files are imported from a local client side file, and files are exported for download.  There is no need for the clutter in the courses templates directory of set definition and classlist files (or even achievement .axp files).  For exporting multiple set definition files, a zip file could be offered.

drgrice1:
We are going to merge this.  If someone wants to add a comment in the help regarding the grayed out import tabs, that might be nice.

PR #2296, opened 2024-01-02
tailor AsciiMath better for MathObjects

Alex-Jordan:
This makes changes to the AsciiMath portion of the MathJax config so that it behaves more consistently with MathObjects. AsciiMath input is used by MathView and can be used when composing text that can have math that will be rendered (a student's essay answer or instructor feedback). Changes are:

- disable `in`, so you don't see a `∈` on your way to making `∞`.
- disable `ne` so you can enter `dne` and `none` without seeing `d≠` and `no≠`.
- as has been the case for a long time, let `**` be interpreted as `^`.
- replace AsciiMath's `infty` with `inf`.
- add `Re`, `Im`, `log10`, `U`, and `><`.

This is well on its way to addressing #638. Perhaps more can be done to get greater consistency with MathObjects and AsciiMath input, but the process is demonstrated now for removing, modifying, or adding triggers.

### DIFF
--- a/htdocs/js/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/MathJaxConfig/mathjax-config.js
@@ -6,9 +6,9 @@ if (!window.MathJax) {
 		loader: { load: ['input/asciimath', '[tex]/noerrors'] },
 		startup: {
 			ready: function() {
-				var AM = MathJax.InputJax.AsciiMath.AM;
+				const AM = MathJax.InputJax.AsciiMath.AM;
 				// modify existing AsciiMath triggers
-				var i = AM.names.indexOf('**');
+				let i = AM.names.indexOf('**');
 				AM.symbols[i] = { input: "**", tag: "msup", output: "^", tex: null, ttype: AM.TOKEN.INFIX };
 
 				i = AM.names.indexOf('infty');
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I made some changes to this.

- more concise code (and I think more efficient, but that depends on efficiency of javascript looping and indexOf method)
- In addition to adding what I added before, also added `infinity`, `Infinity`, `Inf`, `INFINITY`, `INF`, `none`, `None`, `NONE`, `dne`, `Dne`, `DNE`.
- No longer disabling `in`. If you type `in`, you see `∈`. But as soon as you type an `f` the underlying text is `inf` and now you see `∞`. I think this is fine, and allows `in` to still be there in case (for example) an instructor wants that in some feedback message.
- No longer disabling `ne`, since `none` and `dne` are now officially recognized.

drgrice1 (approved):
This looks okay to me for the most part.

This file needs some work though.

It needs to be updated to use modern javascript.  The `var` usage needs to be updated to `const` or `let` appropriately.  I have done this, and was going to make a pull request at some point.  Note that you have two `var i` definitions (on lines 11 and 38).  Those are actually the same variable as the second gets hoisted.  That is why `var` should never be used anymore.  The `for` loop does not limit scope of a `var`.

Also, the formatting will need to be cleaned up.  Soon I plan to add a workflow to use `prettier` on javascript files for formatting.  The `newTriggers` definition will be expanded quite a bit since it exceeds the 120 character limit.

Alex-Jordan:
I updated the var declarations I added (but not the ones that were already in the file and not part of this change).

I'm not sure how best to change the formatting on `newTriggers`. Is the `prettier` change coming soon?

drgrice1:
Don't worry about the prettier formatting. That will be taken care of when it is ready.  I just wanted to point out that the wide formatting wont work with that.  Like perltidy, prettier will auto format the code.

pstaabp (approved):
👍

pstaabp:
Should we close #638 now?

PR #2298, opened 2024-01-11
option to have courses hidden by default

Alex-Jordan:
This adds a new variable to site.conf, `$hide_new_courses`. With this PR, unarchiving a course and keeping its old name, you get a course with its hidden status unchanged. But for unarchiving a course with a new name or making a completely new course, this variable controls whether or not it will be born hidden. More specifically:

If it is false (the distribution default):
* new courses from "Add Courses" do not have the `hide_directory` file (no change)
* new courses that are unarchived with their original name have or don't have `hide_directory` as they were when archived (no change)
* new courses that are unarchived with a _new_ name get `hide_directory` removed, if it is present (this is a change)

If it is true:
* new courses from "Add Courses" get `hide_directory` file and this are born hidden (this is a change)
* new courses that are unarchived with their original name have or don't have `hide_directory` as they were when archived (no change)
* new courses that are unarchived with a _new_ name get `hide_directory` added, if not already present (this is a change)

This will be useful for Runestone, where we always hide all courses (except the demonstration courses). Up to now, I have to manually hide each course after its creation.

This is also helpful for my regular PCC server. Our practice is to leave the prior term's courses open, but hidden. Then they get archived and closed two terms after the course ended. So later, if an instructor wants a new course that is a clone of the old course, we unarchive their old course with a new name (that includes the current term). But without this PR, the new course is born hidden when we don't want it that way.

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -79,11 +79,19 @@ $webwork_server_admin_email = '';
 $admin_course_id = 'admin';
 
 # When new courses are created using the admin course, this setting controls
-# whether or not they will be hidden.  This applies to courses created using
-# "Add Courses" as well as to courses created using "Unarchive Courses" when
-# a new ID is given to the unarchived course. (If it is not given a new ID,
-# its hidden status is preserved.)
-$hide_new_courses = 0;
+# whether or not they will be hidden.  Setting this to anything other than
+# "hidden" or "visible" (or leaving it unset) means that courses created using
+# "Add Courses" are not hidden, and all unarchived courses have whatever hidden
+# status they had when archived.
+# Setting this to "hidden" means that courses created using "Add Courses" and
+# courses created using "Unarchive Courses" with a new course ID will be hidden.
+# Unarchived courses that keep their original name will keep their hidden
+# status.
+# Setting this to "visible" means that courses created using "Add Courses" and
+# courses created using "Unarchive Courses" with a new course ID will be visible.
+# Unarchived courses that keep their original name will keep their hidden
+# status.
+#$new_courses_hidden_status = 'hidden';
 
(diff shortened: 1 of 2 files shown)


drgrice1:
I don't agree with the change to behavior when `$hide_new_courses` is false, and a new course is unarchived with a new name.  I frequently unarchive courses to a new name that were hidden, and want the unarchived course to still be hidden.  (In fact, I don't want any courses visible on my server ever.)

Instead of creating a new course environment option, why not add a checkbox to the add course and unarchive course pages to control this?

Alex-Jordan:
Wouldn't you want it set to true then, if you never want courses visible?

drgrice1:
Yeah, I suppose that works.

Alex-Jordan:
I was thinking about the trinary option too this morning after the earlier messages. I'll go ahead and do that sometime soon. I think I am avoiding the checkboxes because I foresee myself having to check the box a lot. Runestone would have (almost) all courses hidden, and my PCC server would have new courses not be hidden. So one way or the other, I'd be checking that box over and over again.

Alex-Jordan:
Now the variable is named `$new_courses_hidden_status`. It can be `'hidden'`, `'visible'`, or anything else (including unset) for the three behaviors. The comments in `site.conf.dist` hopefully explain it well.

BTW, with this and the custom admin course name PR (#2295), there are new types of things in the config files: config options that affect how the admin course behaves. I put these things in `site.conf`. Does that seem correct? As opposed to `defaults.config` with an option to override using `localOverrides.conf`?

drgrice1 (approved):
This looks good.

I think that setting this (and the custom admin course name in #2295) in site.conf is the correct place for this.

pstaabp (approved):
Looks good.

Issue #2279, opened 2023-12-12
Achievement item that cannot be used


Alex-Jordan:
If you have an achievement item, but it cannot presently be used (e.g. it only applies to open sets and there are none right now) then either the button to use that item should be disabled, or if you click it you should get a special message. Right now you click it and get empty select menus.

pstaabp:
Fixed in #2299

Issue #2281, opened 2023-12-12
Achievement items and problem numbers


Alex-Jordan:
Suppose an achievement item applies to a particular problem from a set. Now suppose you have a set with irregular numbering, like the set problems are numbered 2, 5, 6. The user will be presented a select menu with all the numbers 1, 2, 3, 4, 5, 6. If they choose a number like 1, they get an "error accessing that problem" message. Ideally they would just be presented with 2, 5, 6.

pstaabp:
Fixed in #2299

PR #2299, opened 2024-01-16
Fix achievement issues #2281 and #2279

drgrice1:
When the achievement item modal dialogs for the  'Use Reward' buttons on the achievements page are generated, instead of just storing a max problem id for each set option, store the json encoded list of problem ids.  Initially only the options for the first set are provided.  The javascript then rebuilds the options list when a new set is selected that only includes the problem ids for that set.

Make the "Use Reward" button disabled if there are no sets that the achievement item can act one.  With a little more effort a message could be shown when the button is clicked instead.

This also restructures the code that obtains the achievement item data from the database to be a bit more efficient.  When this data is obtained gateway tests are no longer filtered out.  The current code for each achievement item that does not act on gateway tests is already set up to filter these out.  The achievement items that do act on gateway tests no longer need to load data from the database on their own.  They can use the given list and filter out non-gateway sets.

There seems to have been a minor bug in the `ResurrectGW` achievement in that it showed all gateway sets to be possibly "resurrected" when it should only show gateway sets that are past due.  That has also been fixed, so now it only shows past due gateway sets for "resurrection".

### DIFF
--- a/htdocs/js/AchievementItems/achievementitems.js
+++ b/htdocs/js/AchievementItems/achievementitems.js
@@ -1,17 +1,33 @@
 (() => {
 	for (const setSelect of document.querySelectorAll('select[data-problems]')) {
 		setSelect.addEventListener('change', () => {
-			const max = parseInt(Array.from(setSelect.querySelectorAll('option'))
-				.find((option) => option.value === setSelect.value)?.dataset.max ?? '0');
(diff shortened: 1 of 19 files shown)


pstaabp:
Just so things link, this fixes #2279 and #2281

pstaabp (approved):
Fixes the issues.

Issue #2285, opened 2023-12-12
Achievement points not awarded during reduced credit period


Alex-Jordan:
It seems that if you are in the reduced credit window and correctly answer a question, you get no points towards leveling up. I suspect that somewhere, a test is done based on the score being 1, but during reduced credit the score is 0.75 or something less than 1.

Not sure if this should be changed. Users could either get the full amount of achievement points for correctly answering a question during the reduced credit window, or could get that amount scaled by the reduced credit factor. I would support just awarding all the achievement points.

drgrice1:
We need to discuss what we want to do with this.  I have implemented it to make achievements work on the unreduced score.  We just need to decide if this is what is desired.  I have implemented it to give all achievement points and completely ignore the reduced score using the full unreduced score instead.  This can be tweaked to give a reduced score if desired.

somiaj:
In my limited use of achievements, I thought it might be nice to reduce the number of achievement points in the reduced scoring period. So instead of earning 5 points per correct problem, they only earn 3. I was personally thinking instead of using a percentage reduction, maybe have a configuration field that you can set the number of achievement points correct answers earn during the reducing scoring period. Note I would only reduce the points for the 'per problem earnings', the additional achievements should still earn full points.

pstaabp:
Fixed in #2301

PR #2301, opened 2024-01-18
Reduced scoring improvements for achievements and the progress bar

drgrice1:
Achievements now compute and use the unreduced score for problems for determining if a problem is complete/correct.  That is set as the `status` that is passed into the achievement evaluators.  The reduced score is cached in a hash key of the `$problem` object named reduced_score in case an achievement evaluator wants that information also (although none of them use this at this point).
    
This fixes issue #2285.

Make the problem siblings work with unreduced scores for determining correctness/completion of a problem.  The siblings problem list shows a checkmark by problems for which the maximum possible score has been obtained, even if that is not 100% for the problem due to being in the reduced scoring period.  Also when all problems are complete in the set and there are problems with reduced scores, then the progress bar shows "Complete" without stars instead of the usual "Perfect" with stars.
    
Mark "Perfect" for translation.  For some reason it was not before.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -143,6 +143,7 @@ $institutionName = 'MAA (Mathematical Association of America)';
 $achievementsEnabled = 0;
 $achievementItemsEnabled = 0;
 $achievementPointsPerProblem = 5;
+$achievementPointsPerProblemReduced = 3;
 $achievementPreambleFile = "preamble.at";
 $achievementExcludeSet = [];
 $mail{achievementEmailFrom} = '';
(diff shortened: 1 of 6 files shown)


drgrice1:
I added a new course configuration option `$achievementPointsPerProblemReduced` which is the number of achievement points awarded for completing a problem in the reduced scoring period for a set.  Note that if you want students to earn the same number of points in the reduced scoring period, then set this to be the same thing that you have `$achievementPointsPerProblem` set to.  By default `$achievementPointsPerProblem = 5` and `$achievementPointsPerProblemReduced = 3`.

This was @somiaj's request in issue #2285.

drgrice1:
I should also point out that this does not change the number of points that are awarded for earning an achievement even in the reduced scoring period.

pstaabp (approved):
👍

drdrew42 (approved):

PR #2302, opened 2024-01-20
Rewrite the dump_past_answers script (renamed to dump-past-answers.pl).

drgrice1:
This script currently does not work well (if at all).  This gets the script working again with better help and options.

Note the following changes:

The script no longer saves ids (course id, set id, user id, etc)  as useless hashes.  I don't know why it was thought that doing so was a good idea.

The script does not gzip the resulting csv file.  You can do that on your own if so desired.

The script does not have an upload option (that uploads to wwdata@52.88.32.79).  You can upload it to where you want on your own if so desired.

The unix timestamp for the time at which the script is executed is no longer saved in each row of the CSV file.  It is part of the default file name if you want it.

Past answers from gateway quiz versions are also dumped.  I think it was intended that they would be dumped before, but it wasn't working.

Past answers from problems with more than one answer in them are now dumped.  It certainly was intended that this be done before, but wasn't because of a bug in the script.


drgrice1:
Note that this is already setup to work with #2295, and will also work without that pull request.

pstaabp:
Looks good.  I notice, this doesn't dump the reduced_scoring_date.  I know that doesn't always make sense if the course doesn't have it, but maybe should be included for courses that do.

drgrice1:
Yeah, the script did not include the reduced scoring date previously, and I didn't add it.  An issue with adding that in the region of the csv is that it will change the column numbers of the data after it making the format more variable.  That will also change the format if someone has been using the data from this script previously with reduced scoring enabled.

drgrice1:
I added the reduced scoring date to the dumped data.  It is always in the data even if it is not defined.  In that case the column for it in that row will be empty in the csv file.

PR #2308, opened 2024-01-25
Share strict to the PG safe zone.

drgrice1:
Versions of Perl prior to 5.34 need this.  Otherwise warnings are issue by the `BEGIN { strict->import }` approach now used in place of the old `BEGIN { be_strict() }` approach.

This fixes the issue reported in https://github.com/openwebwork/pg/pull/968#issuecomment-1910760107.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1276,6 +1276,7 @@ ${pg}{modules} = [
 	[qw(Rserve Class::Tiny IO::Handle)],
 	[qw(DragNDrop)],
 	[qw(Types::Serialiser)],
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
This cleared up the warnings I was getting.

Others already tested [the](https://github.com/openwebwork/pg/pull/968) and did not have the issue with the warnings. So they are most likely already on perl 5.36. I guess they could use perlbrew to test this, but what do you think about me just merging this without a second approval?

drgrice1:
That should be okay.

Issue #2312, opened 2024-02-02
Error message when email address is empty


Alex-Jordan:
If a student has no email address and emailing the instructor is allowed, when they use the Email Instructor button but do not fill out the email address at that time, and they send a message, there is an error `Argument contains empty host portion of address at .../Encode.pm line ...`.

Perhaps they should not be allowed to click send until they have entered a validated email address in that field.

somiaj:
I didn't even realize you could enter in an email address in the from box. Checked that if a user doesn't have an email address you can enter one in, but if they do, they cannot edit or change the email address.

I wonder if we should allow entering in of an email address in the 'from' category in this case, or maybe make it a course option if students without email address can enter in their own when sending emails or not. I guess I'm not sure what sort of issues could arise from students entering in bad or maybe malicious emails in the from category.

So another option could be just disable the email button for students who don't have email address, making it something an instructor would have to fix for them.

somiaj:
@Alex-Jordan I just tried to reproduce your bug in 2.18, and I cannot send emails without entering in a valid email address. When I click "Send E-Mail", the javascript keeps me from submitting the form, and a popup on the from field appears telling me to "Please fill out this field".  What version did you experience this on (assuming you have updated all the javascript files as well with `npm`).

Alex-Jordan:
I haven't tried this myself. It comes from a Runestone subscriber report,
where the version is 2.18. Perhaps the real issue is a student entered an
invalid email address, and it passed that JavaScript bar, but was not a
valid email address string for Encode.pm. That would be more consistent
with the error message.

somiaj:
Okay, I tried with entering in `bademail` without any `@` and am able to reproduce the error, so yea javascript is only checking if from is empty.

drgrice1:
Note that javascript is not doing any form validation on this.  This is native browser form validation.

pstaabp:
Fixed in #2313

PR #2313, opened 2024-02-02
Add pattern validation to the email address field on the feedback page.

drgrice1:
Also add some other attributes to the field that are useful for such a field (placeholder, autocorrect: off, and autocapitalization: off), and make the field of type "email" instead of "text" to begin with.

It also wouldn't hurt to add server side validation of this fields contents in case an email saved for the user in the database is not valid, but that is not done yet here.

This addresses issue #2312.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -99,27 +99,29 @@ sub initialize ($c) {
 
 		# Determine the sender of the email.
 		my $sender;
-		if ($user) {
-			if ($user->email_address) {
-				$sender = $user->rfc822_mailbox;
-			} else {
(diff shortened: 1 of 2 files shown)


somiaj:
I was looking at this when I saw your PR, and added a PR to your branch with server side validation if you want to add that as well.

One thing I learned from this is email address can have way more characters in them that I would have initially thought.

drgrice1:
This now includes server side validation as well.  Thanks @somiaj.

somiaj:
Note, to test the server side validation, you can either enter in a bad email address for your user in the accounts manager, or remove the `pattern="..."` from the from input box via a web browser's development tools (inspect).

pstaabp (approved):
Both client and server-side validation works.

PR #2314, opened 2024-02-06
Move settings from site.conf into webwork2.mojolicious.yml

pstaabp:
This replaces the `conf/site.conf` file by adding many configuration variables to `conf/webwork2.mojolicious.yml`. 

The idea is that many of the server configurations shouldn't be overridden by course configuration in `site.conf` or `simple.conf`.  How this happens is that the new order of configurations reads is:

1. `webwork2.mojolicious.yml` is loaded and some basic configuration is done. 
2. `defaults.config`
3. `localOverrides.conf`
4. `simple.conf`
5. `course.conf`
6. Configuration variables that shouldn't be overriden are then copied from `webwork2.mojolicious.yml` to CourseEnvironment.

Note that the URL and directory structure is now included in `webwork2.mojolicious.yml` so that has been removed from `defaults.config`. 

Since yaml files cannot contain code, the loading of `webwork2.mojolicious.yml` is not done in the safe compartment. 

Also, the code in `conf/database.conf.dist` has been moved to `lib/WeBWorK/DB/Utils.pm` -- This can be moved elsewhere if there is a better place for it.  This was removed because it doesn't appear that there are the options that we used to allow in earlier versions. 

Another strange thing that I noticed that when the session times out, I couldn't log back in. The addition of including the password on line 362 of `lib/WeBWorK/Authen.pm` fixed this, but I can't figure out how this work has changed that, but also, why the password was missing from that block. 

Some other ideas I've thought about:

* If this bloats `webwork2.mojolicious.yml`, we can move much of this to a different yaml file.
* Move other settings from `defaults.config` to yaml files that wouldn't need to be loaded in the Safe
* Move some of the items in the CourseEnvironment into the Mojolicious Controller `config` object or other helpers.  Note: @drgrice1 suggested this, but will require rewriting a bunch of the codebase for a new location.


conf/webwork2.mojolicious.dist.yml
```diff
@@ -205,3 +205,516 @@ debug:
...
   # If 1, don't delete temporary files created when a hardcopy is generated.
   preserve_temp_files: 0
+
+
+# Seed/Server variables
+environment:
+
+  # Set these variables to correspond to your configuration.  It is not
+  # recommended to change any of the settings in this file once your
+  # web server has been initially configured.
```
drgrice1, line 215: Remove the recommendation.  That is a bad recommendation.  If server needs change, then the configuration will need to be changed.  I know you copied it from `site.conf`, but don't just copy the comments.  Go through them and clean them up.  There was a lot of crap said in `site.conf`.

conf/webwork2.mojolicious.dist.yml
```diff
@@ -205,3 +205,516 @@ debug:
...
+
+  # URL of WeBWorK handler. If WeBWorK is to be on the web server root, use "". Note
+  # that using "" may not work so we suggest sticking with "/webwork2".
```
drgrice1, line 218: Clearly the comments about quoting do not apply anymore.  Another reason to go through the comments.

drgrice1 (commented):
This is just a quick perusal of a few things, but I already see that there is a lot of work that needs to be done here.  So much in fact, that I recommend not pursuing it further at this point.  At least not in its entirety all at once.

How about you start with the removal of `conf/database.conf.dist`.  Separate that into another pull request, and we will consider that.  Then we will see what else here we want to pursue further after that.

One huge drawback of doing this is that those that have been using webwork will need to do a lot of work to transfer all of there site.conf settings to the webwork2.mojolicious.conf file.  This is perhaps to much work when there may be only a few more releases of webwork2 before we start trying to switch to webwork3.

pstaabp:
Let's chat about this for our next meeting.

Alex-Jordan:

pstaabp:
I'm going to close this in favor of #2338

PR #2317, opened 2024-02-10
Add prettier formatting of JavaScript, style, and HTML files, and check formatting in a workflow.

drgrice1:
Prettier is used to format JavaScript, style, and HTML files in htdocs.  The configuration chosen is in .prettierrc and .editorconfig.  Prettier doesn't really have many options as it is very opinionated.  I am not entirely happy with everything prettier does, but it does give a more uniform code format to files.  Furthermore, it is nice to be able to be able to just run prettier and not need to worry about formatting code.
    
A workflow now checks formatting of JavaScript, style, and HTML files with prettier in addition checking formatting of perl files with perltidy.  Note the "linter" workflow has been renamed to "check-formats", because that is what it does, and this really does not have anything to do with linting.
    
Note that developers can (and should) run `npm run prettier-format` in the htdocs directory to format files.  You can also run `npm run prettier-check` to check that formatting of these files is correct, and not actually change the files.  This is what the workflow does.

Note that this also changes to `actions/checkout@v4` as is done in #2316.  I plan to close that pull request in favor of this one.


drgrice1:
I forgot to note that this is split into two commits.  The first sets things up by adding prettier and changing the workflow.  The second commit is just the result of running `npm run prettier-format` on javascript, style, and html files.

drgrice1:
Actually, there is one exception in the second commit.  The htdocs/js/MathJax/mathjax-config.js file is also updated to modern JavaScript.

pstaabp:
@drgrice1 Looks like there is a conflict with this now. I think #2316 was merged and that's causing the conflict.

pstaabp (approved):
Looks good. Agreed that more formatting will be better.

Alex-Jordan (approved):

PR #2319, opened 2024-02-10
Fix an issue that can cause LTI 1.3 grade passback to fail when called from the job queue.

drgrice1:
The `get_access_token` method calls `$c->url_for` (on line 141 of `WeBWorK::Authen::LTIAdvantage::SubmitGrade`.  This is only called if the access token is not saved in the database or if the access token that is saved is expired.  When this is called from the minion job queue, `$c` is not a `Mojolicious::Controller`.  Instead it is an unblessed hash reference that has the keys `ce`, `db`, and `app` defined.  Clearly that unblessed hash reference will not have the `url_for` method.  As such, the job will fail if a new access token is needed.  So as is done on line 249 in the `submit_grade` method, the `$c` object needs to be switched to the passed in `$c->{app}` object, which does have the `url_for` method.

Note that with the current code this method will still succeed if the database has an unexpired access token stored.  Also, this method will succeed if called with a `Mojolicious::Controller` object (for instance when called in the case that a student initially logs in via LTI 1.3 authentication or in the case that  a student submits an answer to a problem or grades a test).

This could be considered for a hotfix.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvantage/SubmitGrade.pm
+++ b/lib/WeBWorK/Authen/LTIAdvantage/SubmitGrade.pm
@@ -117,16 +117,17 @@ async sub get_access_token ($self) {
 	my $c  = $self->{c};
 	my $ce = $c->{ce};
 	my $db = $c->{db};
+	$c = $c->{app} if $self->{post_processing_mode};
 
 	my $current_token = decode_json($db->getSettingValue('LTIAdvantageAccessToken') // '{}');
 
-	# If the token is still valid (and not about to expire) then it can still be used.
+	# If the token has not expired and is not about to expire, then it can still be used.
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
It sounds like this is unlikely to explain why I have not been able to get grade passback working with D2L, because one should expect that most of the time, that access token should be present and not expired. Is that right? Or is it possible this plays into the issues I had with grade passback and LTI 1.3?

drgrice1:
It is probably not your problem.  If you were able to log in via LTI 1.3 authentication, then it should obtain an access token at that time. If a job to pass grades back follows, then it should work.

Although, in my recent work on this, I thought of something that might be your issue, or at least something to try. I will give details next week.

Alex-Jordan (approved):
Did not test, but the explanation sounds good and code and commentary is good.

pstaabp (approved):
I can't test this, but looks reasonable based on @drgrice1's explanation.

PR #2321, opened 2024-02-11
Add optional certbot renewal routes.

drgrice1:
To enable these routes set `enable_certbot_renewal_routes: 1` in `conf/webwork2.mojolicious.yml`.

Then execute
```
sudo certbot certonly --webroot -w /opt/webwork/webwork2/tmp \
  -d your.domain.edu \
  --post-hook "chown -R www-data:www-data /etc/letsencrypt && systemctl reload webwork2"
```
to renew certificates without needing to stop the webwork2 app. That command will renew the certificate for the first time, and also set up autorenewal in the future. Obviously your.domain.edu needs to be changed to your actual domain name. Note that `/opt/webwork/webwork2/tmp` is the default value of `$webworkDirs{tmp}`. If you customize `$webworkDirs{tmp}` in `localOverrides.conf`, then you will need to use what you have that variable set to instead. Be careful since the default value of `$webworkDirs{tmp}` depends on the value of `$webworkDirs{root}` (which is `/opt/webwork/webwork2` by default). So if you customize `$webworkDirs{root}`, then you will need to adjust the path accordingly. Also, change `www-data:www-data` in the command to be `server_user:server_group` where `server_user` and `server_group` are the values of those settings above. The post hook in the command will run every time that certificates are automatically renewed, and will fix permissions on the new certificates so that the webwork2 app can read them, and will hot reload the webwork2 app to load the new certificates (with zero downtime).

Edit: Note that for this to work `http://*:80` needs to be included in the hypnotoad `listen` list in addition to `https://*:443?...`.  I added a comment about this in `webwork2.mojolicious.dist.yml`.

### DIFF
--- a/conf/webwork2.mojolicious.dist.yml
+++ b/conf/webwork2.mojolicious.dist.yml
@@ -78,6 +78,32 @@ server_group: www-data
 # used when serving the webwork2 app directly.
 redirect_http_to_https: 0
 
+# Change enable_certbot_webroot_routes to 1 to enable routes in the webwork2 app
+# used by certbot for certificate renewal with the webroot option. Note that
+# this should only be used when serving the webwork2 app directly. You will also
+# need to add "- http://*:80" as well as "- http://*:443" to the hypnotoad
+# listen values below for this to work.
+# Then execute
+# sudo certbot certonly --webroot -w /opt/webwork/webwork2/tmp \
+#   -d your.domain.edu \
+#   --post-hook "chown -R www-data:www-data /etc/letsencrypt && systemctl reload webwork2"
+# to renew certificates without needing to stop the webwork2 app. That command
+# will renew the certificate for the first time, and also set up autorenewal in
+# the future. Obviously your.domain.edu needs to be changed to your actual
+# domain name. Note that /opt/webwork/webwork2/tmp is the default value of
+# $webworkDirs{tmp}. If you customize $webworkDirs{tmp} in localOverrides.conf,
+# then you will need to use what you have that variable set to instead. Be
+# careful since the default value of $webworkDirs{tmp} depends on the value of
+# $webworkDirs{root} (which is /opt/webwork/webwork2 by default). So if you
+# customize $webworkDirs{root}, then you will need to adjust the path
+# accordingly. Also, change www-data:www-data in the command to be
+# server_user:server_group where server_user and server_group are the values of
+# those settings above. The post hook in the command will run every time that
+# certificates are automatically renewed, and will fix permissions on the new
+# certificates so that the webwork2 app can read them, and will hot reload the
+# webwork2 app to load the new certificates (with zero downtime).
+enable_certbot_webroot_routes: 0
+
 # hypnotoad server configuration
 # See https://docs.mojolicious.org/Mojo/Server/Daemon
 # Any of the attributes listed there can be set in this section.
--- a/lib/Mojolicious/WeBWorK.pm
+++ b/lib/Mojolicious/WeBWorK.pm
@@ -243,6 +243,17 @@ sub startup ($app) {
 		}
 	}
 
+	# Letsencrypt renewal route.
+	if ($config->{enable_certbot_webroot_routes}) {
+		$r->any(
+			"/.well-known/*static" => sub ($c) {
+				my $file = "$ce->{webworkDirs}{tmp}/.well-known/" . $c->stash('static');
+				return $c->reply->file($file) if -r $file;
+				return $c->render(data => 'File not found', status => 404);
+			}
+		);
+	}
+
 	# Note that these routes must come last to support the case that $webwork_url is '/'.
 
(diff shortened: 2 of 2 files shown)


sean-fitzpatrick:
I might be able to test this next week -- we have a break from classes.

It looks like there's a typo in your initial description (but not in the changed file): you have `routs` instead of `routes`.

I was about to suggest that documentation changes be included but then remembered that that's the wiki, which is completely separate.

drgrice1:
The wiki will need something, but this will go into the release manual on the wiki for the 2.19 release.

Alex-Jordan (approved):
This looks good, but I am not set up to actually test it. The development server I work with is not set up for https. If testing is really needed, I'll see what I can do.

pstaabp (approved):
As with @Alex-Jordan, not sure I can test, but looks good.

drgrice1:
A dry run is sufficient to test.

LetsEncrypt does limit the number of certificate requests, but the limit is 50 per week.  That is usually sufficient to do some renewal testing without going over.  I did several in the process of testing this pull request, but I am not concerned.

With this pull request certificate renewal is automatic.  Once you run the given command once, then automatic renewal is set up, and you don't need to do anything after that.  When you run that command, certbot will update the file `/etc/letsencrypt/renewal/your.domain.edu.conf` with the parameters for the command, and will use them to perform the autorenewal.  So there is no need for a reminder.

sean-fitzpatrick:
Ok, that's fantastic.

We've only recently switched to using CertBot. IT policy used to require enterprise certificates from Entrust that they would issue once per year.

But now that some browsers want renewal every 6 months (or even 3!) they've realized that manually issuing and installing certificates for every server on campus is too big of a nuisance...

sean-fitzpatrick:
I made the two edits in this PR and then ran CertBot with the webroot option, and renewed the certificate successfully.

The dry run failed. But this was due to an oversight on my part: if you were previously using the `standalone` option, it is now necessary to go back into `webwork2.mojolicious.yml`, and uncomment the line to have it listen on port 80. (It is then also a good idea to change the `redirect_http-to_https` value from 0 to 1.

With those changes and a restart of the webwork2 service, the dry run succeeds.
As far as I can test, this works as expected.

(The Active Calculus people who are using my server as the backend for their book on Runestone will be pleased to learn that their students will no longer temporarily lose access every 3 months.)

drgrice1:
Ahh, I knew that, but I forgot to mention it.  I will add a comment in the `webwork2.mojolicious.dist.yml` file.  The `redirect_http_to_https` is not required for the certificate renewal to work, but probably is a good idea in that case.

Alex-Jordan:
I tried this tonight without success. One question I have is that even before trying this, I have
```
hypnotoad:
  listen:
    - https://*:443?cert=...&key=...
    - http://*:80
```

The instructions in the comments make it sound as though I should add `- http://*:443`. Is that right? Or am I good as this is with the 443 line I already have?

When I tried, this, I added the `- http://*:443`.  Not sure if that caused the problem, but I got:

```
Saving debug log to /var/log/letsencrypt/letsencrypt.log

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
An RSA certificate named webwork-hosting.runestone.academy already exists. Do
you want to update its key type to ECDSA?
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
(U)pdate key type/(K)eep existing key type: U
Renewing an existing certificate for webwork-hosting.runestone.academy

Certbot failed to authenticate some domains (authenticator: webroot). The Certificate Authority reported these problems:
  Domain: webwork-hosting.runestone.academy
  Type:   connection
  Detail: 67.205.179.165: Fetching https://webwork-hosting.runestone.academy/.well-known/acme-challenge/0qq0emCf5BoyGIqFDPtwQxkj6fQbXVz23axoFhoX4ro: Error getting validation data

Hint: The Certificate Authority failed to download the temporary challenge files created by Certbot. Ensure that the listed domains serve their content from the provided --webroot-path/-w and that files created there can be downloaded from the internet.

Some challenges have failed.
Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /var/log/letsencrypt/letsencrypt.log or re-run Certbot with -v for more details.
```

I ran the command with `--webroot -w /opt/webwork/webwork2/tmp`. I see that it was looking for something written within `https://webwork-hosting.runestone.academy/.well-known/...` and it couldn't find it.

Is it really correct that we would want `-w /opt/webwork/webwork2/tmp`? That folder is for temporary non-web-accessible files.

drgrice1:
That should be `https://*:443` with the certificates and such.  The `http` is a mistake.  So the line you already have together with `http://*:80` is all that you should have.

The `--webroot -w /opt/webwork/webwork2/tmp` is correct.  That is a special route that is added when the `enable_certbot_webroot_routes` is true.  Actually, only files in the `.well-known` subdirectory are served.  Note that that subdirectory is created by `certbot`.

PR #2324, opened 2024-02-13
make field in add user form in accounts manager have type number

Alex-Jordan:
This is just a small thing I saw while looking at #2294.

### DIFF
--- a/templates/ContentGenerator/Instructor/UserList/add_form.html.ep
+++ b/templates/ContentGenerator/Instructor/UserList/add_form.html.ep
@@ -3,7 +3,7 @@
 		<%= maketext('Add how many users?') %><span class="required-field">*</span>
 	<% end =%>
 	<div class="col-auto">
-		<%= text_field number_of_students => 1, id => 'add_entry', size => 3, 'aria-required' => 'true',
+		<%= number_field number_of_students => 1, id => 'add_entry', size => 5, min => 1, 'aria-required' => 'true',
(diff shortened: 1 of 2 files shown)


templates/ContentGenerator/Instructor/UserList/add_form.html.ep
```diff
@@ -3,7 +3,7 @@
 		<%= maketext('Add how many users?') %><span class="required-field">*</span>
 	<% end =%>
 	<div class="col-auto">
-		<%= text_field number_of_students => 1, id => 'add_entry', size => 3, 'aria-required' => 'true',
+		<%= number_field number_of_students => 1, id => 'add_entry', size => 5, 'aria-required' => 'true',
```
drgrice1, line 6: It would probably be a good idea to add `min => 1` to the attributes of this number field.  Currently this allows negative numbers and 0.  Neither of which make any sense here, and any of those values open the add users page with no inputs for adding a user.

drgrice1 (commented):
It is a good idea to convert this to a number field.  It would probably also be good to change the corresponding field in `templates/ContentGenerator/Instructor/AddUsers.html.ep` to a number field as well.

Alex-Jordan:
@drgrice1 I updated this in the ways that you suggested.

drgrice1 (approved):
Looks good.  Thanks.

pstaabp (approved):

PR #2328, opened 2024-02-17
Add filter action to achievements manager.

somiaj:
This adds a action to filter achievements as suggested by @Alex-Jordan.

### DIFF
--- a/htdocs/js/AchievementList/achievementlist.js
+++ b/htdocs/js/AchievementList/achievementlist.js
@@ -22,10 +22,10 @@
 		const action = document.getElementById('current_action')?.value || '';
 		if (action === 'filter') {
 			const filter_select = document.getElementById('filter_select');
-			const filter = filter_select?.selectedIndex || 0;
+			const filter = filter_select?.value || '';
 			const filter_text = document.getElementById('filter_text');
 			const filter_category = document.getElementById('filter_category');
-			if (filter === 1 && !is_achievement_selected()) {
+			if (filter === 'selected' && !is_achievement_selected()) {
 				e.preventDefault();
 				e.stopPropagation();
 				filter_select?.classList.add('is-invalid');
@@ -37,7 +37,7 @@
 					},
 					{ once: true }
 				);
-			} else if (filter === 2 && filter_text?.value === '') {
+			} else if (filter === 'match_ids' && filter_text?.value === '') {
 				e.preventDefault();
 				e.stopPropagation();
 				document.getElementById('filter_text_err_msg')?.classList.remove('d-none');
@@ -61,7 +61,7 @@
 					},
 					{ once: true }
 				);
-			} else if (filter === 3 && filter_category?.value === '') {
+			} else if (filter === 'match_category' && filter_category?.value === '') {
 				e.preventDefault();
 				e.stopPropagation();
 				document.getElementById('filter_category_err_msg')?.classList.remove('d-none');
@@ -177,7 +177,7 @@
 					},
 					{ once: true }
 				);
-			} else if (document.getElementById('create_select')?.selectedIndex == 1 && !is_achievement_selected()) {
+			} else if (document.getElementById('create_select')?.value == 'copy' && !is_achievement_selected()) {
 				e.preventDefault();
 				e.stopPropagation();
 			}
@@ -221,9 +221,9 @@
 	const filter_text_elements = document.getElementById('filter_text_elements');
 	const filter_category_elements = document.getElementById('filter_category_elements');
 	const filterElementToggle = () => {
-		if (filter_select?.selectedIndex == 2) filter_text_elements.style.display = 'flex';
+		if (filter_select?.value === 'match_ids') filter_text_elements.style.display = 'flex';
 		else filter_text_elements.style.display = 'none';
-		if (filter_select?.selectedIndex == 3) filter_category_elements.style.display = 'flex';
+		if (filter_select?.value === 'match_category' ) filter_category_elements.style.display = 'flex';
 		else filter_category_elements.style.display = 'none';
 	};
 
--- a/htdocs/js/AchievementList/achievementlist.js
+++ b/htdocs/js/AchievementList/achievementlist.js
@@ -9,7 +9,7 @@
 			'change',
 			() => {
 				document.getElementById('select_achievement_err_msg')?.classList.add('d-none');
-				for (const id of ['edit_select', 'assign_select', 'export_select', 'score_select']) {
+				for (const id of ['filter_select', 'edit_select', 'assign_select', 'export_select', 'score_select']) {
 					document.getElementById(id)?.classList.remove('is-invalid');
 				}
 			},
@@ -20,7 +20,73 @@
 
 	document.getElementById('achievement-list')?.addEventListener('submit', (e) => {
 		const action = document.getElementById('current_action')?.value || '';
-		if (action === 'edit') {
+		if (action === 'filter') {
+			const filter_select = document.getElementById('filter_select');
+			const filter = filter_select?.selectedIndex || 0;
+			const filter_text = document.getElementById('filter_text');
+			const filter_category = document.getElementById('filter_category');
+			if (filter === 1 && !is_achievement_selected()) {
+				e.preventDefault();
+				e.stopPropagation();
+				filter_select?.classList.add('is-invalid');
+				filter_select?.addEventListener(
+					'change',
+					() => {
+						document.getElementById('select_achievement_err_msg').classList.add('d-none');
+						document.getElementById('filter_select').classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
+			} else if (filter === 2 && filter_text?.value === '') {
+				e.preventDefault();
+				e.stopPropagation();
+				document.getElementById('filter_text_err_msg')?.classList.remove('d-none');
+				filter_select?.classList.add('is-invalid');
+				filter_text?.classList.add('is-invalid');
+				filter_text?.addEventListener(
+					'change',
+					() => {
+						document.getElementById('filter_text_err_msg')?.classList.add('d-none');
+						document.getElementById('filter_text')?.classList.remove('is-invalid');
+						document.getElementById('filter_select')?.classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
+				filter_select?.addEventListener(
+					'change',
+					() => {
+						document.getElementById('filter_text_err_msg')?.classList.add('d-none');
+						document.getElementById('filter_text')?.classList.remove('is-invalid');
+						document.getElementById('filter_select')?.classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
+			} else if (filter === 3 && filter_category?.value === '') {
+				e.preventDefault();
+				e.stopPropagation();
+				document.getElementById('filter_category_err_msg')?.classList.remove('d-none');
+				filter_select?.classList.add('is-invalid');
+				filter_category?.classList.add('is-invalid');
+				filter_category?.addEventListener(
+					'change',
+					() => {
+						document.getElementById('filter_category_err_msg')?.classList.add('d-none');
+						document.getElementById('filter_category')?.classList.remove('is-invalid');
+						document.getElementById('filter_select')?.classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
+				filter_select?.addEventListener(
+					'change',
+					() => {
+						document.getElementById('filter_category_err_msg')?.classList.add('d-none');
+						document.getElementById('filter_category')?.classList.remove('is-invalid');
+						document.getElementById('filter_select')?.classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
+			}
+		} else if (action === 'edit') {
 			const edit_select = document.getElementById('edit_select');
 			if (edit_select.value === 'selected' && !is_achievement_selected()) {
 				e.preventDefault();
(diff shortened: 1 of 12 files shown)


lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
```diff
@@ -135,17 +144,51 @@ sub initialize ($c) {
...
+		if (@{ $c->{visibleAchievementIDs} }) {
+			$result = $c->maketext('Showing achievements in category [_1].', $category);
+		} else {
+			$result = $c->maketext('No achievements in category [_1].', $category);
+		}
+	} elsif ($scope eq 'enabled') {
+		$result = $c->maketext('Showing enabled achievements.');
+		$c->{visibleAchievementIDs} = [ map { $_->[0] } $db->listAchievementsWhere({ enabled => 1 }) ];
+	} elsif ($scope eq 'disabled') {
+		$result = $c->maketext('Showing enabled achievements.');
```
drgrice1, line 185: I think this should be
```suggestion
		$result = $c->maketext('Showing disabled achievements.');
```
I was a little confused when I selected "disabled achievements" and clicked "Filter", and the message said "Showing enabled achievements".

htdocs/js/AchievementList/achievementlist.js
```diff
@@ -20,7 +20,73 @@
 
 	document.getElementById('achievement-list')?.addEventListener('submit', (e) => {
 		const action = document.getElementById('current_action')?.value || '';
-		if (action === 'edit') {
+		if (action === 'filter') {
+			const filter_select = document.getElementById('filter_select');
+			const filter = filter_select?.selectedIndex || 0;
```
drgrice1, line 25: I know that we have used the `selectedIndex` in this way elsewhere, but it would be better to use `const filter = filter_select?.value || ''` instead.  Then you can use `if (filter === 'selected' && !is_achievement_selected()) {` below.  Of course replace the other places with their values instead of indices also.  There are two advantages to this approach.  First, instead of a meaningless index, you have a human readable string that makes it more clear which filter you are working with.  Second, if another filter is added at a later time and it is not the last in the select, then you don't have to change all of the later indices in the code to fix it.

I was going to comment on this for your earlier pull request that changed the action scope selects, but decided not to because that was the way it was done previously.

htdocs/js/AchievementList/achievementlist.js
```diff
@@ -20,7 +20,73 @@
...
+				);
+				filter_select?.addEventListener(
+					'change',
+					() => {
+						document.getElementById('filter_category_err_msg')?.classList.add('d-none');
+						document.getElementById('filter_category')?.classList.remove('is-invalid');
+						document.getElementById('filter_select')?.classList.remove('is-invalid');
+					},
+					{ once: true }
+				);
```
drgrice1, line 87: There is a lot of redundancy in the above code.  I was also going to comment on this in the other pull request.  I think there are better ways to achieve this without so much repetition.

You can leave this for now though since I let it go for the other pull request.

drgrice1 (commented):
There are a few details that need attention.

If a filter happens to filter out all achievements, then the message shown is "No achievements shown. Create an achievement!"  For example, in my test course all achievements are enabled.  So when I selected the "disabled achievements" filter, that gave me none.  That message is not appropriate anymore if there are achievements, but none are shown because of the filter.

lib/WeBWorK/ContentGenerator/Instructor/AchievementList.pm
somiaj: Thanks, overlooked this. Also fixed another bug with edit/export all achievements not working correctly if a filter was already applied.

drgrice1 (approved):
This looks good now.  Everything seems to be working in my tests.

pstaabp:
Overall, looks good.  Seems like the filter by category should be a dropdown/select list of categories, instead of requiring it to be typed in exactly.  The default list only has 6 categories, and this should be easy to pull out of the list.

somiaj:
@pstaabp I thought about that, but the only way I could think to get the list of all categories is to first pull all achievements from the database and cycle through them getting all categories. But once a filter is applied not all achievements are pulled from the database, only the filtered ones.

templates/ContentGenerator/Instructor/AchievementList/filter_form.html.ep
```diff
@@ -0,0 +1,43 @@
...
+			<span class="required-field">*</span>
+		<% end =%>
+		<div class="col-auto">
+			<%= text_field 'action.filter.category' => '', id => 'filter_category', 'aria-required' => 'true',
+				class => 'form-control form-control-sm', dir => 'ltr' =%>
+		</div>
+	</div>
+	<div id="filter_category_err_msg" class="alert alert-danger p-1 mb-2 d-inline-flex d-none">
+		<%= maketext('Please enter a category to match.') %>
+	</div>
```
drgrice1, line 42: There needs to be a div that contains lines 30 through 42, and that div should have the `filter_category_elements` id (which should be removed from the `row mb-2` div).

The same thing needs to be done with the `filter_text_elements` above.

Similar changes need to be made for all of the content generator modules that have a similar filter form.

Currently if you click the submit button without entering anything into the text input, and then you change the select to another filter, then the form validation message is still shown, even though it no longer applies.

somiaj: I am not seeing the behavior you describe. I click `Filter` when `Enter matching category below` is selected, and get an error message that I must enter a category to match. I then use the select to select any other filter option, and the validation message goes away along with the `is-invalid` matches on the form.

I can still add the extra divs, but currently I do not see any validation messages being left behind when they no longer apply (I'm assuming you updated the javascript?).

drgrice1: Your right, the validation message does go away on the achievements page, but not on the accounts manager page and the sets manager page.

However, the structure I am asking for still should be done.  The `filter_category_err_msg` should be grouped structurally with the elements it applies to for accessibility reasons.

drgrice1: Actually, never mind that last comment.  Divs don't really provide that kind of structural semantics for accessibility anyway.  So you can just leave it.

So the issue is that on the accounts manager page and sets manager page, the messages don't go away when they should.

somiaj: Thanks for the clarification, I'll go update those messages to correctly be hidden.

drgrice1:
There are some real problems with the javascript that you are using.  This goes back to what you did in #2294.

You are repeatedly adding event handlers to various elements each time the form is submitted.  Those event handlers pile up.  Note that the `once: true` option makes the event handler go away if it is called, but each time the form is submitted while that has not been called another event handler is added.  Eventually all of those fire at the same time if the event does occur.

Also you `getElementById` in many cases where you have already obtained a reference to that element.  That is an expensive query.  Each time it is called, the entire DOM has to be searched until that element is found.  For example on line 24 of achievementlist.js you find the `filter_select` element (and even save a reference in a variable), and then on line 36 you search the DOM again, instead of just using the already found reference.  Several places you just blindly search the DOM each time without saving a reference.  For example, on line 43 you find the `filter_text_err_msg`, and then on line 49 you search for it again.

I mentioned the redundancy in the code before, but this is just a total kludge really.  The achievementlist.js file is 233 lines long, merely to validate a few form fields.  That is really excessive.

somiaj:
I have been thinking about ways to clean up the code and remove the redundancy, though unsure on the best approach here. My thought was to have helper functions do the work.

As for the change event methods, the reason I was calling `getElementById` again is I didn't realize that inside the event function I had access to the previously defined variables. Now that I know that I do have access, I'll update things to avoid calling `getElementById` multiple times when not needed.

I'm not quite sure the what to do about multiple events, it sounds like I should also be removing these events to avoid them piling up when one of the companion events fires?

somiaj:
I will have time to look at this and see what I can come up with this weekend.

drgrice1:
I did a little with the multiple event thing for userlist.js in #2335.  Look at what I did there for a start.

somiaj:
@drgrice1 There is my approach at reducing the code duplication and adding/removing event listeners, so they don't pile up. If you prefer this, and any code cleanup you may see, I can apply this to the other two pages as well (though maybe I should leave userlist.js alone since you have gotten it in your PR).

drgrice1:
Yeah, that looks considerably better.  Go ahead and do that for the other two pages as well.  I will work my changes to `userlist.js` into your scheme.

somiaj:
I have updated the other two action form validation location to match what was done here.

somiaj:
There was some achievement transition code that was suggested to remove around 2017....I can open another PR for that if preferred.

drgrice1:
Deleting that code is fine.

The updated javascript is considerably better.  Perhaps in the future it would be nice to do this more uniformly for all pages that use action tabs (and the `actiontabs.js` file).  But this is good for now.

htdocs/js/AchievementList/achievementlist.js
```diff
@@ -1,119 +1,96 @@
 (() => {
 	// Action form validation.
+	const show_errors = (ids, elements) => {
+		for (const id of ids) elements.push(document.getElementById(id));
+		for (const element of elements) {
+			if (element?.id.endsWith('_err_msg')) {
+				element?.classList.remove('d-none');
+			} else {
+				element?.classList.add('is-invalid');
+				element?.addEventListener('change', hide_errors([], elements));
```
drgrice1, line 10: This won't work.  This `removeEventListener` call won't remove the event listener added on lines 64-69 or on line 36.  The method returned by `hide_errors` here will not be literally the same method returned when `hide_errors` is called there.  This would fail even if `hide_errors` were called in both places with the same arguments, because the `removeEventListener` method needs to be called with a reference to the same function that `addEventListener` was called with to work.

drgrice1 (changes requested):
I missed a few things before.

I see that you removed the `{ once: true }` option to the error hiding event listeners.  That is needed.  Otherwise the handler will continue to be called even after the form validation messages have been cleared any time a checkbox is checked or unchecked.

htdocs/js/AchievementList/achievementlist.js
drgrice1: The similar code in `achievementlist.js` and `problemsetlist.js` also needs to be fixed.

drgrice1: To see what I am talking about, add a `console.log` statement to line 17 (the beginning of the function returned by the `hide_errors` method) in `achievementlist.js`.  Then on the achievements list page, select the "selected achievements" option in the filter tab.  Then click on "Filter" several times with no achievements selected.  Now select an achievement.  You will see all of the console logs, and you will see all of those logs each time you select or unselect an achievement.

drgrice1: Really, what you need to do is call `removeEventListener` each time before you call `addEventListener`.  It is certainly not going to work to have the handler added by `addEventListener` try to remove itself.  The handler has to be called for that to happen, and it is not being called each time the event listener is added.  Look again at what I did in `userlist.js` in my pull request.

somiaj: Thanks for catching that. I incorrectly thought the solution in the comments here would work. I should have read (tested it) deeper.

https://stackoverflow.com/questions/256754/how-to-pass-arguments-to-addeventlistener-listener-function#answer-45696430

It does appear that currying can work here (at least under my tests) if you save the curried functions to access later so you can remove them.  So I have updated the code to store the curried functions so the event listeners can be removed.

The reason I went with this approach vs `{ once: true }`, was I was adding event listeners to multiple objects, and wanted to remove them all, not just the one the fired. I couldn't see in your code how that was achieved (or missed it).

somiaj: This also allowed me to not run any error hiding code when moving tabs if there are no event listeners saved when switching tabs, as any error will populate `event_listeners` and the events to fix the action should clear all error messages and listeners.

drgrice1 (approved):
This is better, and works to prevent handlers for events from piling up.

That stack exchange post you linked to is certainly not accurate.  You have to read the comments responding to it to really see what is needed.

In the code I was using in `userlist.js` I was only dealing with one handler being added, and so didn't need something more complicated.  Although, as mentioned in that #2335, there was more work needed.  I saw that there were other event listeners that would need similar handling.

Alex-Jordan:
Thanks for this. I was testing it, and I thought I would make the category selection be a dropdown. I opened a PR to your branch (@somiaj ):

https://github.com/somiaj/webwork2/pull/13

somiaj:
@Alex-Jordan updated this so the category filter is now a drop down menu instead of a text field.

Alex-Jordan (approved):

drgrice1:
I will merge this with 2+1 approvals.

lib/WeBWorK/DB.pm
```diff
@@ -1290,6 +1290,11 @@ sub getAchievements {
 	return $self->{achievement}->gets(map { [$_] } @achievementIDs);
 }
 
+sub getAchievementCategories {
+	my ($self) = shift->checkArgs(\@_);
+	return map {@$_} $self->{achievement}->get_fields_where("DISTINCT category", undef, "category");
+}
+
```
drgrice1, line 1297: @Alex-Jordan:  I realized I didn't look closely at your commit at the end of this.  I just saw this.  This is a nice.  A good new database method.

PR #2330, opened 2024-02-22
Add a patch for the dvisvgm issue observed in https://github.com/openwebwork/pg/issues/1023.

drgrice1:
This patch is applied in the docker build.

It can be applied on your server with the command
`sudo patch -p1 -d / < /opt/webwork/webwork2/docker-config/pgfsys-dvisvmg-bbox-fix.patch`

Note this patch is specific to Ubuntu 22.04, so don't try it on other versions of Ubuntu or other linux distributions (or other operating systems).

This was the change made in https://github.com/pgf-tikz/pgf/pull/1276 that was merged into the master branch of the pgf-tikz latex package on October 23, 2023 that fixes the issue reported in https://github.com/pgf-tikz/pgf/issues/1275.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -213,6 +213,7 @@ COPY --from=base /opt/base/pg $APP_ROOT/pg
 
 # Patch files that are applied below
 COPY docker-config/imagemagick-allow-pdf-read.patch /tmp
+COPY docker-config/pgfsys-dvisvmg-bbox-fix.patch /tmp
 
 RUN echo "PATH=$PATH:$APP_ROOT/webwork2/bin" >> /root/.bashrc \
 	&& mkdir /run/webwork2 /etc/ssl/local \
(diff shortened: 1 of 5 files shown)


drgrice1:
In testing this I realized the docker build is broken, and has some other issues.  So this pull request now fixes those things.  These changes are in the first commit.  The second commit is what this originally did.

The changes are:

The dependence on File::Copy::Recursive module was added in #2172, but not added to the docker build.
    
Start using node 20 in the docker build.  The node 16 install script from nodesource is deprecated, and they have imposed a 60 second install delay when you use it.
    
Fix the ownership on the courses directory and admin course directory when the docker entrypoint runs.  Fixing the ownership on the courses directory and admin course directory will not cause any slow down to the execution of that script.  It is not a recursive ownership change, just two directories for which ownership is changed which is super fast.

somiaj (approved):
I tested that the patch also correctly applies in Debian bookworm, and fixes the issue.

Note, you have `.patch.patch` in the above command, so copy/pasting it will fail.

drgrice1:
Thanks.  I fixed the patch command.

PR #2331, opened 2024-02-22
skip over hidden folders and files in the library browser

Alex-Jordan:
Chemistry faculty here have a library they manage with git. It has a `.git` folder at the top level that shows up in the library browser's list of folders containing PG files (if you click the button to view this library). It could be managed like the OPL, with an intermediate folder, but this seems like a good change anyway. Namely, filter out any files or folders that start with a dot.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
@@ -85,7 +85,7 @@ sub get_library_sets ($c, $top, $dir) {
 	}
 	return (0) if grep {/^=library-ignore$/} @lis;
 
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks fine.

This is another place where `Mojo::File` usage could simplify things quite a bit.

Also, I think that the `=library-no-combine`, `=library-combine-up`, and `=library-ignore` crap should be removed.  I don't know why it was thought that it would be a good idea to use directories by those names to implement this odd behavior that is not documented anywhere.

Alex-Jordan:
> I think that the =library-no-combine, =library-combine-up, and =library-ignore crap should be removed

I agree. Unfortunately I just brought this to someone's attention last week, so I know of one person using it. No big deal if they have to find another way. They are using it so that a folder with testing .pg files within their personal library tree won't appear.

pstaabp (approved):
Looks good.

drgrice1:
This can be merged.

PR #2333, opened 2024-02-23
Switch to using a Mojolicious session for cookies.

drgrice1:
The primary advantage of this switch is that now cookies are signed. That means that unlike webwork2's current cookies, the data stored in the session cookies can be trusted to not have been tampered with. Note that setting the secrets in the `conf/webwork2.mojolicious.yml` file is now important (and not just something that we tell system administrators to do).  That secret is the private key that is used to sign cookies.

The documentation on session management in defaults.config and localOverrides.conf.dist has been updated.  Most of that was obsolete nonsense.

Note that the $debugPracticeUser has been removed.  As stated in the comments about it in defaults.config, "Why de we even have this?"

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -886,12 +886,6 @@ $sessionKeyLength = 32;
 # (you can comment this out to remove practice user support)
 $practiceUserPrefix = "practice";
 
-# There is a practice user who can be logged in multiple times.  He's
-# commented out by default, though, so you don't hurt yourself.  It is
-# kindof a backdoor to the practice user system, since he doesn't have a
-# password.  Come to think of it, why do we even have this?!
-#$debugPracticeUser = "practice666";
-
 # Option for gateway tests; $gatewayGracePeriod is the time in seconds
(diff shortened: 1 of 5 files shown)


drgrice1:
Note that this is part 1 of 3 in a rather exhaustive revamp of webwork2's authentication system.

pstaabp:
Is the switch to mojolicious cookies the code in the session subroutine in the `Controller` module?  Overall, except for that, it seems to be mostly some code cleanup and documentation improvement.

drgrice1:
Most of it is just code clean up.  I tried to put all of that into this pull request so that once this is merged, the other pull requests are more transparent as to what they are doing.  The main part of the code for switching to Mojolicious cookie is the `session` subroutine in the `WeBWorK::Controller` module (as you observed), and the calls to that method in the `fetchCookie`, `sendCookie`, and `killCookie` methods of `WeBWorK::Authen`.

pstaabp (approved):
Looks good.

Alex-Jordan:
Tested some, with no problems.

Issue #2336, opened 2024-02-23
Cross site scripting vulnerability


drgrice1:
This was reported in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4295 (4 years ago), and I don't know why an issue was not made at that time.  The issue is with how we use the `status_message` URL parameter to pass html when redirects occur.  An example of this in action was given in that forum post.  Here is a simpler example that work now:

`https://your.server.domain/webwork2/courseID?status_message=<script>console.log('hello')</script>`

The `console.log` does get executed.

The approach of passing HTML in a URL parameter, and then injecting that into the DOM without begin properly sanitized is a clear vulnerability.  There is a note in the forum post about using HTML::Scrubber to sanitize input, but I do not believe this URL parameter has ever been put through any such scrutiny.  Furthermore, I don't think that even HTML::Scrubber would catch the instance noted in that post.  That is (modified a bit):

`https://your.server.com/webwork2/courseID?status_message=<audio%20src=x%20onerror=alert('hello')>`

The alert in that is executed also.  HTML::Scrubber might catch the first example with a `script` tag, but wouldn't catch the `onerror` attribute.

The good news is that with #2332 there is a safe mechanism that could be implemented to pass such messages securely via the session without using a URL parameter.

I am noting this issue mainly for myself to work on.

drgrice1:
I should note that this does not affect the internal `status_message` usage that is not passed through a URL parameter.  So `addmessage` and methods that go through that (`addbadmessage` and `addgoodmessage`) are not a problem when they only use internal content.  It is passing the `status_message` URL parameter into those methods that is the problem.

Alex-Jordan:
I believe this was fully addressed by #2337.

drgrice1:
Yes it was.  Thanks for closing this.

PR #2337, opened 2024-02-23
Add a session flash and use it to pass status messages for redirects.

drgrice1:
The session `flash` method is similar to the `session` method added previously, and is a method of the WeBWorK::Authen object attached to the controller.  It uses the `flash` method of `Mojolicious::Plugin::DefaultHelpers` if `session_management_via` is "session_cookie" and imitates that with the database session otherwise.  This method saves data to the session that will persist only for the next request.
    
This is then used to save `status_messages` when redirects occur.  This fixes issue #2336, since the `status_message` URL parameter is no longer used. We need to make sure that we never again use a URL parameter to pass HTML.

This builds on #2333, #2334, and #2335 and is part 4 of 3 of the authentication system revamp.  So long, and thanks for all the fish!

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -799,6 +799,39 @@ sub session {
 	return $c->session(@params);
 }
 
+=head2 flash
+
+This sets data in the session that only persists for the next request.
+
+=cut
+
+sub flash {
+	my ($self, @params) = @_;
+	my $c = $self->{c};
+
+	# If session_management_via is not "session_cookie" (so should be "key"), then use the database session.
+	if ($c->ce->{session_management_via} ne 'session_cookie' || $c->stash('disable_cookies')) {
+		# Note that the return values are the same as those returned by the
+		# Mojolicious::Plugin::DefaultHelpers flash method in the following cases.
+
+		# Note that this will be the database session in this case since
+		# the conditions are the same as in the session method above.
+		my $session = $self->session;
+
+		# Get old values.
+		return $session->{flash} ? $session->{flash}{ $params[0] } : undef if @params == 1 && !ref $params[0];
+
+		# Initialize new flash and merge values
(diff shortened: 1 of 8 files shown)


Alex-Jordan:
Can you suggest anything specific to test here? Or just switch to the branch and see that logging in with cookies works fine?

drgrice1:
That is of course something that should still work correctly.  So that is something to test.

The main thing that this deals with is the `status_message` URL parameter.  An easy way to test that is to use the PG problem editor and save a file using the `Save As` tab.  With the develop branch when you do that, you will see the `status_message` in the URL with a long URL escaped html string.  Those are all of those dismiss-able messages that are displayed after the file is saved.  With this pull request, you won't see the URL parameter, but the messages will still be there.  Instead they are in the session flash.  Although, you won't really be able to see that.  They are only in the session cookie (or database session when using `$manage_session_via = 'key'`)  for one request.  So by the time the redirect finishes it has already been removed from the session.

Note that there is one message you won't see if you are saving a new problem from the PG problem editor opened from the site navigation menu.  That is the message "No changes have been made to set".  You will of course still see that if you are saving when editing a problem from a set (and not modifying the set), and then it will say "No changes have been made to set setID".

An important thing to test is what is stated in issue #2336.  With the develop branch if you enter
```https://your.server.domain/webwork2/courseID?status_message=<script>console.log('hello')</script>```
into the address bar of the browser, then you will see 'hello' in the developer console.  With this pull request you will not.

Alex-Jordan (approved):
This all tested out. I have no issues with maintaining a cookie based session. The messages are no longer in the URL. The one message mentioned is no longer there when it shouldn't be. And the scripting test no longer printed anything in the console log.

pstaabp (approved):
Tested the status_messages are either no longer added to the URL as a query string and if they exist, they are not parsed.

PR #2338, opened 2024-02-24
Move database and external program settings

pstaabp:
This PR
* moves database and external program settings in both `site.conf` and `defaults.config` to `webwork2.mojolicious.yml`
* moves code from `database.conf.dist` to `WeBWorK::DB::Utils` 
* loads these settings into the `CourseEnvironment`.

These settings will be loaded after `simple.conf` and `course.conf` so instructors will not be able to override these settings.


conf/defaults.config
```diff
@@ -1093,10 +1056,8 @@ $pg{displayModeOptions}{images} = {
 
 	# If dbsource is set to a nonempty value, then this database connection information will be used to store dvipng
 	# depths.  It is assumed that the 'depths' table exists in the database.
-	dvipng_depth_db => {
-		dbsource => $database_dsn,
-		user     => $database_username,
-		passwd   => $database_password,
+	# These are set in the CourseEnvironment.
+	dvipng_depth_db => {}
```
drgrice1, line 1060: This comment is not good enough.  These can be set here.  A site administrator may want to use a different database for this.  The options should be added back commented out with example defaults.  The comment needs to give a good explanation of how this works.

drgrice1 (changes requested):
This needs a lot of work to be acceptable.

conf/defaults.config
```diff
@@ -1093,10 +1056,8 @@ $pg{displayModeOptions}{images} = {
...
 	},
```
drgrice1, line 1061: I forgot to mention that this line needs to be deleted.

drgrice1: That is line 1061.  It causes this file to fail to load.

lib/WeBWorK/CourseEnvironment.pm
```diff
@@ -365,6 +378,48 @@ sub status_abbrev_has_behavior {
...
+		$ce->{database_dsn} =
+			$config->{database}{use_socket_if_localhost}
+			? "DBI:$config->{database}{driver}:database=$config->{database}{name}"
+			: "DBI:$config->{database}{driver}:database=$config->{database}{name};"
+			. "host=127.0.0.1;port=$config->{database}{port}";
+	} else {
+		$ce->{database_dsn} = "DBI:$config->{database}{driver}:database=$config->{database}{name};"
+			. "host=$config->{database}{host};port=$config->{database}{port}";
+	}
+	$ce->{dbLayoutName}                = 'sql_single';
```
drgrice1, line 399: Hmm.  Somehow my comment here did not make it into my review.  This is the comment that was most important, and that spurs my remark about this pull request needing a lot of work.

`dbLayoutName` needs to be deleted here, and all references to it removed from the code.  You have broken many things by not taking care of this.  The `adduser` script and several things in `CourseManagement.pm` (and probably other places too) access the previous `dbLayouts` hash using this name.  Since there is no `dbLayouts` hash anymore those things fail.

conf/webwork2.mojolicious.dist.yml
```diff
@@ -231,3 +231,90 @@ debug:
...
+  port: 3306
+
+  # name of the database
+  name: webwork
+
+  # Standard permissions command used to initialize the webwork database
+  # GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, ALTER, DROP, INDEX, LOCK TABLES ON webwork.* TO webworkWrite@localhost IDENTIFIED BY 'passwordRW';
+  # where webworkWrite and passwordRW must match the corresponding variables in the next section.
+  username: ''
+  password: ''
```
drgrice1, line 292: Another thing here.  Change this to
```suggestion
  password: passwordRW
```
There are two reasons this should be done.  First, this is the current default in `site.conf.dist`.  Second, this shows that the quotes are not needed.

Although, documentation in the comments is needed here similar but different than what was in site.conf.dist.  In many cases quoting will not be needed here.  However, there are some cases where it will be needed.  The comments here should make this clear.  See https://symfony.com/doc/current/reference/formats/yaml.html for when quoting is needed in a YAML file.

conf/webwork2.mojolicious.dist.yml
```diff
@@ -231,3 +231,90 @@ debug:
...
+# You can use "which tar"  for example to find out where the "tar" program is located
+
+externalPrograms:
+  mv: /bin/mv
+  cp: /bin/cp
+  rm: /bin/rm
+  mkdir: /bin/mkdir
+  tar: /bin/tar
+  gzip: /bin/gzip
+  git: /bin/git
```
drgrice1, line 247: This should be `/usr/bin/git` as it was in `site.conf.dist`.

conf/webwork2.mojolicious.dist.yml
```diff
@@ -231,3 +231,90 @@ debug:
...
+  dvipng: /usr/bin/dvipng
+  convert: /usr/bin/convert
+  dvisvgm: /usr/bin/dvisvgm
+  pdf2svg: /usr/bin/pdf2svg
+
+  giftopnm: /usr/bin/giftopnm
+  ppmtopgm: /usr/bin/ppmtopgm
+  pnmtops: /usr/bin/pnmtops
+  pnmtopng: /usr/bin/pnmtopng
+  pnmtopnm: /usr/bin/pnmtopnm
```
drgrice1, line 270: You have dropped the `gif2eps`, `png2eps`, and `gif2png` external program definitions.  I don't think the first two are ever used anymore, but the last one is.

conf/webwork2.mojolicious.dist.yml
somiaj: With `usrmerge` now standard in the main distros, would it be worth updating all the paths to use `/usr/bin`. Not strictly necessary since `/bin -> /usr/bin`, but might be worth updating to the new standard location.

conf/defaults.config
pstaabp: The database variables aren't available in defaults.config now, which is why I moved them to the CourseEnvironment.  I can provide some documentation in localOverrides to provide this.

drgrice1: That is my point.  Either these DO need to be available in defaults.config (or really in localOverrides.conf), or a separate section in webwork2.mojolicious.dist.yml needs to be created for them.  The same for `$problemLibrary_db`.

conf/webwork2.mojolicious.dist.yml
pstaabp: I meant to keep the code on old lines 156-158 of `site.conf.dist`.   Should I just include the `gif2png`?

drgrice1: You can just leave lines 156-158 of `site.conf.dist`.

conf/defaults.config
pstaabp: I'll add to localOverrides

pstaabp:
This removes all references to dbLayoutName and cases where the dbLayouts hash is used. With this, there is only the standard dbLayout, which is now defined in `WeBWorK::DB::Layout`.

Also, removed the older style unarchive database archive file.  It was unclear when this was last in the codebase, probably > 10 years.

Alex-Jordan:

pstaabp:
Fixed now.

Issue #2326, opened 2024-02-14
File Manager creates file even when not saved


Alex-Jordan:
If you use the File Manager to create a file `foo`, it takes you to the editor. Even if you abort (don't save anything) the file `foo` has already been created and saved as an empty file. I think it should not actually save a file to disk until you click Save or Save As.

pstaabp:
Fixed in #2339

PR #2339, opened 2024-02-26
Delay file creation when a new file is created in the file manager.

drgrice1:
Instead of creating the file when the new file confirmation page is submitted, create the file when "Save" (or "Save As") is clicked from the edit page that opens next.

This was requested in issue #2326.

Also, don't focus the `name` input on the edit page.  That is a bit annoying.  Only focus that input on the confirmation page.  That is the only page that has a `data` input.

### DIFF
--- a/htdocs/js/FileManager/filemanager.js
+++ b/htdocs/js/FileManager/filemanager.js
@@ -54,8 +54,8 @@
 			}
 		});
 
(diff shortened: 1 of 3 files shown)


pstaabp (approved):
👍

Alex-Jordan:
Works as described, thanks!

PR #2347, opened 2024-03-02
Fix a warning that occurs for proctor authentication.

drgrice1:
This is a stopgap measure since this is already fixed in #2334, but this was missed in #2333.  #2334 completely rewrites this module.

The `WeBWorK::Authen::create_session` method no longer accepts a prior database key object for its second argument, so the optional third argument (which is now the optional second argument) is in the wrong place for the `WeBWorK::Authen::Proctor` modules override of this method.


Alex-Jordan:
I'll merge this so it won't somehow get in the way of someone testing before and after with #2334.

PR #2348, opened 2024-03-03
Remove user override check boxes when editing set details for a user.

drgrice1:
This builds on #2345, and makes the same sort of changes for this page.

The dates check on the problem set detail was synchronized with the check for the user detail page.  Most importantly, a test version's open, close, and answer dates can not be nullified. In fact, no set's open, close, and answer dates can be entirely nullified.  In other words, a global set must have the open, close, and answer dates set.
    
Note that the set values column was left at the end on this page (instead of before the user override values as it now is on the user detail page).  There are issues with moving that first on this page.  There are cases where there are no set values, but there are user values at all (for example, for the problem seed, problem status, and problem attempts). So moving set values first would leave an ugly gap between the label and the user value.
    
The 10 year limit on future set dates was removed as was done for the user detail page.
    
There were some check boxes (not override checks) that had the check box input inside the label instead of the input and label being siblings.  That is not correct for Bootstrap's expectations for the `form-check` class.  In particular, the check box and the label "if status is less than 1" (for the "Randomize Seeds" button) were not aligned correctly due to this.  So those issues were fixed.

### DIFF
--- a/htdocs/js/ProblemSetDetail/problemsetdetail.js
+++ b/htdocs/js/ProblemSetDetail/problemsetdetail.js
@@ -421,37 +421,6 @@
 	// Render all problems on page load if requested.
 	if (document.getElementById('auto_render')?.checked) renderAll();
 
-	// Make the override checkboxes for text type inputs checked or unchecked appropriately
-	// as determined by the value of the input when that value changes.
-	document
-		.querySelectorAll('input[type="text"][data-override],input[type="hidden"][data-override]')
-		.forEach((input) => {
-			const overrideCheck = document.getElementById(input.dataset.override);
-			if (!overrideCheck) return;
-			const changeHandler = () => (overrideCheck.checked = input.value != '');
-			input.addEventListener('change', changeHandler);
-			if (input.parentElement.classList.contains('flatpickr')) {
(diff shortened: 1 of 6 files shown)


Alex-Jordan (approved):
Tested, and I can change dates as expected.

There is one thing though that your js skills may be able to improve on. (And if so, maybe separate from this PR.) Suppose a user does not have any date overrides and you go to this page. And then you input an override for, let's say, the close date. And the new override close date is later than the global answer date. This does not cause the answer date field to automatically populate with something.

Note that if you come to this page and there are already overrides for the close date and answer date, then editing the close date to a later time than the answer override date *will* cause the answer override date to update. What I'm describing in the last paragraph is only when you reach this page and there are no override dates. But that is a common situation, and it is common for an instructor to set a close date override, and forget to also update the answer date.

Of course this is not just about close dates and answer dates, but also open dates and reduced scoring dates. This is a behavior that is in 2.18 and probably earlier, not anything about the changes made in this PR.

drgrice1:
I know of the behavior that you speak of, and I have wanted to do something about it.  I will try and get to it sometime in the near future.

pstaabp (approved):

PR #2354, opened 2024-03-11
let instructor's change course title

Alex-Jordan:
This makes some changes to the Course Config module so that in addition to course environment variables, settings from the course's `_setting` table in the database can be set. Then it applies this possibility to the course title.

The upshot is an instructor can visit Course Config and change the course title on their own now. Previously the only way to update the course title is to have the WW admin use the admin course to do that (or otherwise do it from the command line).

### DIFF
--- a/templates/ContentGenerator/Instructor/Config.html.ep
+++ b/templates/ContentGenerator/Instructor/Config.html.ep
@@ -39,6 +39,10 @@
 				</tr>
 				% for my $con (@configSectionArray) {
 					% my $conobject = $c->objectify($con);
+					% next unless (
+						% defined $conobject->{var}
+						% || $authz->hasPermissions(param('user'), "setting_table_$conobject->{setting}")
+					% );
 					% my $name      = defined $conobject->{var}
 							% ? ($conobject->{var} =~ s/[{]/-/gr) =~ s/[}]//gr
 							% : $conobject->{setting};
--- a/templates/ContentGenerator/Instructor/Config.html.ep
+++ b/templates/ContentGenerator/Instructor/Config.html.ep
@@ -39,11 +39,15 @@
 				</tr>
 				% for my $con (@configSectionArray) {
 					% my $conobject = $c->objectify($con);
-					% my $name      = ($conobject->{var} =~ s/[{]/-/gr) =~ s/[}]//gr;
+					% my $name      = defined $conobject->{var}
+							% ? ($conobject->{var} =~ s/[{]/-/gr) =~ s/[}]//gr
+							% : $conobject->{setting};
 					<tr>
 						<td><%= $conobject->what_string %></td>
 						<td class="text-center">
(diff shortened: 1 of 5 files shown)


dlglin:
I'm wondering if there needs to be a way to disable this.  Will there be scenarios where an instructor should not be allowed to change the course name?

Alex-Jordan:
You wrote course "name", but I want to be clear this is the course title.
It's not the course ID that appears in urls, course files folder paths,
etc. I believe it only appears at the top of the assignments page, and
possibly as the html title for pages.

dlglin:
Yes, I meant course title.  I'm envisioning a scenario where an institution wants to insist that the course title match the official name of the course in their calendar, or perhaps a WW administrator includes section information in the course title and doesn't want an instructor to remove this.

I don't know if these are real use cases, so I'm asking whether it's worth the extra work to have a configuration option to disable this feature or have a permission associated with it.

drgrice1 (commented):
I think this is generally fine.

I think that @dlglin might be right that there are those that won't want the instructor to be able to change the title.  If it is not to hard to do, perhaps an option would be merited.

There is one thing that needs attention though.  The help for the course title (that opens when you click the question mark by it) doesn't look the best.  Here is what I see:
title-help
This is in `templates/ContentGenerator/Instructor/Config/config_help.html.ep`.   Line 25 of that file needs to be wrapped in an `if (defined $configObject->{var})` statement.  Probably the modal title should be different too.  Perhaps it should say "Setting Documentation" instead of "Variable Documentation"?

Alex-Jordan:
OK, I fixed the help issue @drgrice1 found, and I added permission level for being able to do this.

templates/ContentGenerator/Instructor/Config.html.ep
```diff
@@ -39,11 +39,19 @@
 				</tr>
 				% for my $con (@configSectionArray) {
 					% my $conobject = $c->objectify($con);
-					% my $name      = ($conobject->{var} =~ s/[{]/-/gr) =~ s/[}]//gr;
+					% next unless (
+						% defined $conobject->{var}
+						% || $authz->hasPermissions(param('user'), "setting_table_$conobject->{setting}")
+					% );
+					% my $name      = defined $conobject->{var}
```
drgrice1, line 46: Now that there is not an `=` to align with on the line above, the extra spacing before `=` should be removed.

drgrice1 (approved):
Looks good now.  Thanks.

pstaabp:
Just to test the permissions, I added 
```
$pemissionLevels{setting_table_courseTitle}      = 'admin';
```
to localOverrides.conf

and still was able to change the courseTitle.  Am I missing something here?

drgrice1:
The permission setting seems to be working for me.  Are you sure that the user you are testing with is not an admin user?

pstaabp:
Silly typo.

pstaabp (approved):
This looks good.

PR #2358, opened 2024-03-12
new Student Orientation

Alex-Jordan:
This removes "set0", the "MAATutorial" set, and the "Orientation" set from the distribution model course. I find most of the content there out of step with modern webwork2. If people want to keep those, I can change that part of this PR.

The main thing is this also adds a "Student Orientation" set, keeping all the files centrally in `webwork2/assets/pg/` so this can be maintained with the distribution and not propagate as content in courses that get copied from term to term.

The new student orientation has a problem (number 3) that displays content contingent upon whether MathQuill, MathView, or neither is in use. Suggestions welcome for improving this orientation, although those could always come in a later PR to make changes easier to assess.

To try this, it is not enough to check out the branch. You also need to update your modelCourse's templates folder to match the `courses.dist/modelCourse/templates/` folder. Then create a new course, and import the Student Orientation set.


drgrice1 (commented):
I think this is generally good.  I don't mind getting rid of the old sets that are largely obsolete.  I commented on specific things that I observed.

drgrice1:
Also, could you wrap the long lines in the added problems.  Many of those problems have excessively long lines, and that makes it harder to read the source code.  It doesn't matter for how PGML displays the rendered problem.

drgrice1:
One other thing.  Also apply the PG perltidy to the added problems.

assets/pg/Student_Orientation/enteringMath.pg
```diff
@@ -0,0 +1,83 @@
...
+}
+########
+elsif ($envir{entryAssist} eq 'MathView') {$body = <<END_BODY;
+When you want to enter an answer, you might want some assistance understanding how to type it and what your answer looks like after it is formatted as a math expression. To the right of each answer blank there is a button[@
+    MODES(
+        HTML => ' <button class="btn btn-sm btn-secondary codeshard-btn" disabled><i class="fa-solid fa-th"/></button>',
+        TeX => ''
+    )
+@]* you can use to reveal a palette with tools for constructing math expressions as well as a preview of what the expression will look like. Try using the palette to help enter the expression [::pi/sqrt(x+1)::]. For this particular expression:
+1. You should start with the fraction building button. It will place [|()/()|]* into the answer blank.
```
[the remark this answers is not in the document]
Alex-Jordan: I think you clicked the horizontal layout fraction button `a/b`, and that only makes `/`. But to the right of that button is a vertical layout fraction button that makes `()/()`. I will put `\frac{a}{b}` in there to make it clear which button.

assets/pg/Student_Orientation/decimals.pg
```diff
@@ -0,0 +1,40 @@
...
+
+    [`\sqrt{110} =10.4880885\ldots=`] [_]{sqrt(110)} (Try [|10|]*, [|10.4|]*, [|10.5|]*, [|10.48|]*, [|10.49|]*, etc.)
+
+    [`0.002038 =`] [_]{0.002038}{4} (Try [|0.0020|]*, [|0.00203|]*, [|0.00204|]*, [|0.002038|]*, etc.)
+
+    [`20380.2 =`] [_]{20380.2}{4} (Try [|20000|]*, [|20300|]*, [|20400|]*, [|20380|]*, etc.)
+
+With [`0.002038`], students sometimes feel they should be able to get away with [`0.002`], since that is only off by a tiny amount of [`0.000038`]. However, percentage-wise this would be off by over [`1.8%`], which is considered too much.
+
+With [`20380.2`], students mistakenly believe they must enter the entire number. While that's fine, it's also OK to round to four significant digits. So you may simply enter [`20380`].
```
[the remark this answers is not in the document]
Alex-Jordan: I wanted to make this work for the default context settings. Would it work to follow up this line here with a note that these settings can change on specific problems, for example a monetary problem requiring answers be exact to the cent?

assets/pg/Student_Orientation/emailInstructor.pg
```diff
@@ -0,0 +1,29 @@
...
+    PGML.pl
+    PGcourse.pl
+));
+
+$a = random(5,10);
+
+BEGIN_PGML
+## Email Instructor
+
+If you feel stuck on a problem,do not understand what the problem asks for, suspect there is a bug with the problem, do not understand what syntax to use to enter the answer, or all of the above, there is an "Email Instructor" button below; use it.
```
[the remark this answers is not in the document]
Alex-Jordan: I changed it to grab the value of `feedback_button_name`. If the course is not using Email Instructor for students, the instructor should just remove this exercise entirely from the Orientation.

assets/pg/Student_Orientation/hardcopy.pg
```diff
@@ -0,0 +1,34 @@
...
+
+$a = random(100,900);
+
+BEGIN_PGML
+## PDF Hardcopy ##
+
+From the *Homework Sets* page, you may push a download button[@
+    MODES(
+        HTML => ' <i class="hardcopy-tooltip icon far fa-arrow-alt-circle-down fa-lg"></i>',
+        TeX => ''
```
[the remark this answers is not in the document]
Alex-Jordan: I added a line mentioning the button on the set's top level page. And I changed the text to "Download Hardcopy for Current Set". And corrected "Homework Sets"  to "Assignments".

I changed it to acknowledge "Assignments" page may be unavailable.

assets/pg/Student_Orientation/enteringMath.pg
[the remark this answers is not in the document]
drgrice1: Ahh, that is it.  I though I remembered there being parentheses.  This is more clear now.

assets/pg/Student_Orientation/decimals.pg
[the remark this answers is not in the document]
drgrice1: Yeah, I think saying something like that would be good.  My problems that require such accuracy also have directions in the problem that state the accuracy that is required.  Perhaps saying something along the lines of "follow the directions in the problem" would be good.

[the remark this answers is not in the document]
drgrice1: In general, I think that even if a problem is using the default tolerance, problems should always state what is expected for the accuracy in an answer.  This is something that should be improved in OPL/Contrib problems.

assets/pg/Student_Orientation/emailInstructor.pg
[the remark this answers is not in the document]
drgrice1: Good.  I didn't realize we passed that variable in to PG!

drgrice1 (approved):
Looks good.  Thanks for making those changes.

assets/pg/Student_Orientation/emailInstructor.pg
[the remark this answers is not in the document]
drgrice1: Oh, I see that you added that to the data passed to PG in this pull request.

assets/pg/Student_Orientation/welcome.pg
```diff
@@ -0,0 +1,34 @@
...
+    PGcourse.pl
+));
+
+$isOpen = time() > $envir{openDate} && time() < $envir{dueDate};
+$instruction =
+    ($isOpen)
+    ? 'To get started, click in the answer blank, type the correct answer, and click the "Submit Answers" button.'
+    : 'This set is not currently open. You may want to ask your instructor to check if the open and close dates for '
+    . 'this set are correctly set. In the meantime, you can click in the answer blank, type the correct answer, '
+    . 'and click the "Check Answers" button to see if your answer is correct.';
```
pstaabp, line 19: Trying to be a complete WW noob.  I would put the  `click the "check Answers" button` instruction after the answer blank.

assets/pg/Student_Orientation/welcome.pg
```diff
@@ -0,0 +1,34 @@
...
+## Welcome
+
+Welcome to the WeBWorK online homework platform. This orientation will familiarize you with basic features and usage of
+WeBWorK.
+
+[$instruction]**
+
+[`2+2={}`][_]{Context("LimitedNumeric"), 4}{4}
+
+[@ MODES(HTML => 'When you are ready, click "Next Problem".', TeX => '') @]
```
pstaabp, line 31: Maybe include that the "Next Button" is at the top of the page.

assets/pg/Student_Orientation/feedback.pg
```diff
@@ -0,0 +1,68 @@
...
+If you type an answer and click to "Preview my Answers", you will see an info button[@
+    MODES(
+        HTML => ': <button class="ww-feedback-btn btn btn-sm btn-info" disabled><i/></button>',
+        TeX => ''
+    )
+@]* instead of the correct/incorrect buttons.
+
+In addition to automated feedback, your instructor can leave messages for you in WeBWorK. To see these messages, you
+must visit the problem for which a message has been left. The message will appear above the exercise. If your course
+uses essay questions, there is no automated feedback, and these messages from the instructor will be your only feedback.
```
pstaabp, line 63: I think an essay example should be a separate problem.  I think the student really won't register this comment (nor know what an essay problem is here)  Also, if an instructor doesn't use essay questions, they can skip this problem.

assets/pg/Student_Orientation/navigating.pg
```diff
@@ -0,0 +1,74 @@
...
+    PGML.pl
+    PGcourse.pl
+));
+
+$a = random(5, 10);
+
+BEGIN_PGML
+## Navigating WeBWorK ##
+By now you've had experience with the "Next Problem" button. The "Previous Problem" button takes you to the previous
+problem, and the "Problem List" button takes you back to the beginning of this problem set.
```
pstaabp, line 18: I would say the "Problem List" takes you to a listing of all problems in the set.

assets/pg/Student_Orientation/hardcopy.pg
```diff
@@ -0,0 +1,37 @@
...
+    )
+@]* to download a PDF version. Alternatively, when you are at the page listing all of one set's exercises, there is a
+"Download Hardcopy for Current Set" button.
+
+This file is something that you can read onscreen while you are offline. You can even print it off and take it to a
+tutoring center or somewhere comfortable to work on.
+
+To check that you understand how this works, download the PDF version of this Orientation assignment. At the end of
+this problem in the PDF, you will find the answer that is expected here: [_]{Compute("$a")}{4}
```
pstaabp, line 30: I did this and got the error that it doesn't support Unicode.  Perhaps we should default the latex engine to xelatex.

assets/pg/Student_Orientation/miscellaneous.pg
```diff
@@ -0,0 +1,60 @@
...
+character if you have a way to do that.
+* All of your submissions are permanently recorded. That means you can log off at any time and all of your work will be
+saved. So you can do assignments in little bits at a time and work out of order if you like.
+* WeBWorK can be used on tablets and smartphones.
+* Most problems are coded with randomization. This means that while you and your classmate may have the same skeleton of
+a problem, the numbers invloved are different. This encourages you to teach each other the _process_ of answering a
+problem without being able to simply share the answers.
+
+The collection of all of the numbers between [`0`] and [`\pi`] including the endpoints is represented as [`[0,\pi]`] in
+interval notation. Practice entering [`[0,\pi]`] in the answer blank: [_]{Context("Interval"), '[0,pi]'}{4}.
```
pstaabp, line 49: [0,pi] didn't work for me here.  I haven't used this format for switching Contexts.  Does this work?  I did switch the Context, and defined the answer above the PGML block which worked.

assets/pg/Student_Orientation/hardcopy.pg
drgrice1: I do not think that we should switch the default to xelatex. Xelatex does not always render the best results. I also don't think that a problem should assume Unicode will work in hardcopy. That is something that needs to be reconsidered in the problems here that use Unicode characters.

assets/pg/Student_Orientation/miscellaneous.pg
```diff
@@ -0,0 +1,60 @@
...
+
+DOCUMENT();
+
+loadMacros(qw(
+    PGstandard.pl
+    PGML.pl
+    PGcourse.pl
+));
+
+$a = random(100, 900);
```
drgrice1, line 13: I don't think this variable is used in this problem.

assets/pg/Student_Orientation/miscellaneous.pg
drgrice1: I am not sure exactly what is going on here.  If I enter `[0,pi]` it is also counted incorrect for me.  If you delete the other answers from the problem, then it works.  If I change the answer on that line to `[_]{Context("Interval"), Compute('[0,pi]')}{4}` it also works.

assets/pg/Student_Orientation/hardcopy.pg
pstaabp: I just switched back from xelatex and it's compiling now--not sure what was going on. I withdraw my comment.

assets/pg/Student_Orientation/welcome.pg
Alex-Jordan: Since this is conditional upon whether the set is currently open, it's a bit messy to make that move. Unless you can pass along specific code to make it work well, I'll leave it this way for now.

assets/pg/Student_Orientation/welcome.pg
Alex-Jordan: Is it always at the top of the page? Or will it always be? It feels like a layout consideration that might change someday (more likely on small screen). It's also not strictly speaking at the "top" of the page.

assets/pg/Student_Orientation/feedback.pg
Alex-Jordan: One thing I want to avoid is an actual need for the instructor to give feedback on a problem here, as part of demonstrating how manual feedback works. So I avoided having an actual essay question. Is there any way to fake some feedback? That is, code something in the problem that makes PG think there is feedback in the database and it presents that feedback? In another exercise I have it faking the number of attempts used, but that seems easier.

assets/pg/Student_Orientation/enteringMath.pg
```diff
@@ -0,0 +1,104 @@
...
+    [_]{'3^(x+1)/(2(x+2)^2)'}{16}
+
+END_BODY
+
+BEGIN_PGML
+## Entering Math
+
+[@ if ($envir{entryAssist} eq 'MathQuill') {$MathQuill} elsif ($envir{entryAssist} eq 'MathView') {$MathView} else {$None} @]**
+
+[@ MODES(HTML => 'When you are ready, click "Next Problem".', TeX => '') @]
```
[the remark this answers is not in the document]
Alex-Jordan: the first problem has conditional code if the set is open, but I'd prefer to avoid that on all the problems. So if I reference "Submit Answers" in other problems, that is less than ideal if there is no "Submit Answers" button present (i.e. the "Check Answers" button is there instead).

assets/pg/Student_Orientation/navigating.pg
Alex-Jordan: I changed to this in my local edits to commit and push soon.

assets/pg/Student_Orientation/hardcopy.pg
Alex-Jordan: It must have been some other problem that caused that error because there is no special character in this problem. And no reason you would have submitted a special character for the answer to this one. So my guess is you had entered a raw π or something as an answer to a different question, and the tex was trying to print your submitted answer.

assets/pg/Student_Orientation/miscellaneous.pg
Alex-Jordan: Switching to `Compute('[0,pi]')` for now.

assets/pg/Student_Orientation/miscellaneous.pg
Alex-Jordan: Thanks, it was a remnant from building each problem on top of the previous one.

assets/pg/Student_Orientation/decimals.pg
```diff
@@ -0,0 +1,54 @@
...
+[|2/6|]*, etc.)
+
+Sometimes, you will be allowed to use decimal _approximations_ to the real answer. So for instance, in the next answer
+blank we can get away with typing [|0.3333|]* even though that is slightly different from [:1/3:].
+[`\frac13={}`][_]{Context("Numeric"),'1/3'}{4}  (Try [|1/3|]*, [|0.33|]*, [|0.3333|]*, [|2/6|]*, etc.)
+
+Why is [|0.3333|]* accepted and [|0.33|]* is not? _If_ a decimal approximation is acceptable at all, then you need to
+use enough significant digits so your answer is "close enough" to the actual correct answer (which was [:1/3:] in this
+case). In general, using _four_ significant digits in your decimals will be enough. You will often be able to get away
+with fewer, but using four is recommended.
```
somiaj, line 28: When I talk to my students about this I mention the default is a relative error of 0.1%. That might be worth adding here (note you do talk about relative error below, and this could make it clearer why the 1.8% error is not close enough). Also note with 0.1% error,  you could provide an example which has 3 sig figs, and another which has 4 sig figs, to try to better explain why 4 is recommended.

assets/pg/Student_Orientation/emailInstructor.pg
```diff
@@ -0,0 +1,40 @@
...
+When you use this button, you should write a message to your instructor (or TA) explaining what you have tried. It will
+help a lot if you write down the steps of the math that you have worked out too. You can even attach a file, for example
+a picture of your hand-written work. Then the instructor gets your message along with a link to the problem from _your
+perspective_, and they can also see all of your previously attempted answers. This way your instructor can write you
+back with good help/hints.
+
+This answer blank is expecting a certain mystery answer: [_]{Real(112358)->cmp(tolType=>'absolute',tolerance=>0.1)}{4}.
+The only way to get the answer is to use the "[$button_name]" button and ask for the answer to this Orientation Problem
+[$envir{probNum}]. The instructor will reply to your email as soon as they are able to. (Instructors: you can
+Check/Submit an answer, then view the feedback to reveal the expected correct answer.)
```
somiaj, line 32: Isn't there a variable that states if the user is an instructor or not? I have used permission level to show stuff for instructors but not students (thinking the instructor comment could be hidden from students).

assets/pg/Student_Orientation/miscellaneous.pg
```diff
@@ -0,0 +1,58 @@
...
+certain number of times.
+* Once the due date has passed, you can still enter answers in the answer blanks and use the "Check" button, and WeBWorK
+will still give you the same feedback; just not for credit.
+* Captial letters are different than lowercase letters when it comes to variables. For exmaple, [`\pi r^2`] might be
+correct while [`\pi R^2`] is not.
+* When you click on a problem set, there is set-specific information in a panel to the right. Always read this before
+you begin the problems set. It may give you important hints.
+*  Repeated guessing is not helpful.  If you have gotten a problem incorrect several times, you should ask for help.
+That's what your instructor, TAs, and tutors are here for.
+* Make sure you read the feedback messages that WeBWorK gives you.
```
somiaj, line 33: These feedback messages can be a bit too mathy in my experience for some students to understand what it is telling them. You could add a comment about that here, and maybe remind them if they don't understand what the feedback message is saying to email their instructor/ta/etc.

somiaj (approved):
looks good. I've added a few comments you can address or not, just things that I saw when reading through the set.

assets/pg/Student_Orientation/emailInstructor.pg
drgrice1: There is.  It is `$envir{isInstructor}`.  I thought I made a comment about this in my review.  Maybe I forgot.  Or what there another problem with a similar instructor comment?  In any case, I agree that this should only be shown for instructors.

drgrice1: Actually, you can just use `$isInstructor` in a problem.

somiaj: Another thing I noticed, would randomizing the mystery answer be worthwhile? Otherwise students might just start telling other students, and avoid emailing the instructor.

assets/pg/Student_Orientation/decimals.pg
Alex-Jordan: I'm biased by the situation at my own school, where I've changed the default tolerance type to `digits` instead of relative, and I am hoping to avoid getting too much into describing relative tolerance. But I'll see what I can do here.

somiaj: If I use these problems, I'd probably edit to particular things I do, so don't change just on my account. Just something I observed.

Alex-Jordan: Even when we used the distribution default of relative error, that phrase went over our students' heads if it wasn't fully explained in an extended discussion. Of course, "significant digits" can do the same, with students here usually assuming it means "decimal places".

assets/pg/Student_Orientation/emailInstructor.pg
Alex-Jordan: > would randomizing the mystery answer be worthwhile?

It would be a bit tedious for the instructor to have to look up the answer each time. This answer is the start of the Fibonacci sequence and easy to remember. A few faculty here even respond with a canned message linking to the Fibonacci sequence Wikipedia article as a way to start building interest in math. But you have to ask whether students are really going to share orientation question answers with each other, and if they do, if we really care.

Alex-Jordan:
Thanks @somiaj, I made some changes based on your review.

drgrice1:
I will merge this now with two approvals.

PR #2359, opened 2024-03-13
move PGML lab files to assets folder

Alex-Jordan:
This moves the PGML lab files to the assets folder. Otherwise as a course is copied from term to term, it will copy potentially outdated PGML lab files. This way we can maintain them centrally in the assets folder and make sure they are relevant to the current version of WeBWorK/PG.


drgrice1 (commented):
This is generally okay.

However, there are some problems with this approach.

One problem will also occur for #2358 and this symbolic link approach.  This assumes that the `$webworkDirs{courses}` directory is set to the default value (or at least is positioned relative to the `webwork2` directory so that the link will point to the correct place.

Another problem is that this only fixes the issue of making sure the PGML lab problem is current for problems created for WeBWorK 2.19 and beyond.  That of course can't be retroactively fixed without manual changes made by the system administrator.

drgrice1:
Perhaps we are in a place that the `modelCourse` could not be in the courses directory at all anymore, and could be used directly from the `webwork2/courses.dist` directory?  I mean, at this point we can now copy from any course we want to.  So if an administrator wants to have their own "model course", then the administrator can create a course that is set up the way they want (and could even call it "modelCourse" -- of course we would remove all code that skips that in course listings).  Yeah, it would actually have the database tables, but is that a problem?

Of course that still doesn't solve the issue of courses having the up to date PGML lab problem, but it is worth considering.

drgrice1:
Note the same problem can occur with the `capaLibrary`, `Contrib`, and `Library` links.  I think that these links should be created programmatically when a course is created using the values in the directory settings from the course environment.  This would ensure that they point to the correct location regardless of the directory structure.

We could even go further and detect broken links in an existing course, and then update them to point to the correct locations.

Alex-Jordan:
I had in mind the situation with `Library` as a precedent, and if someone changes those locations, there would be consequences. The model course would need its symlinks rewired.

It would not be a problem for a model course to have database tables. In fact, that is a feature! Our model courses at PCC have exercise sets already loaded, and we are using the new ability to copy those when creating a new course. This saves instructors who just want to load the default sets for our courses the step of importing sets.

I don't think we can automate solving the issue of a course that was created in 2.18 and has the `PGMLLab/` folder directly in its `templates/` folder. They are out of luck until someone recognizes there is this issue and updates that course. The same is true by the way for the `achievements/` folder, where if we want to update how achievements behave, we are out of luck with courses that are propagating old `achievements/` folders.

drgrice1:
I think these are things to think about for webwork3.  That is if there even is a course directory (probably better to avoid that entirely).

drgrice1:
This can be merged.  The link issue can be resolved later since it is not new with this pull request.

PR #2364, opened 2024-03-19
accessibility updates to student orientation

Alex-Jordan:
Changes here make the disabled buttons in the student orientation sound more like the actual buttons an AT user will run into while using WeBWorK. This is perhaps pending https://github.com/openwebwork/pg/pull/1039, in case the text "Preview Information" changes.

This also adds an alert to users that a Braille hardcopy might be available upon request from institutional staff. They can usually work with .tex source to make that. (In the future, we could make a PTX hardcopy file and use PTX Braille production to automate Braille hardcopy production. I don't think PTX Braille production is ready yet in the PTX command line tool, so it's too early to try that yet.)

### DIFF
--- a/assets/pg/Student_Orientation/feedback.pg
+++ b/assets/pg/Student_Orientation/feedback.pg
@@ -27,24 +27,33 @@ When you Submit an answer, a feedback button appears near the answer blank.
 
 * If you answer correctly, you see a green checkmark[@
     MODES(
-        HTML => ': <button class="ww-feedback-btn btn btn-sm btn-success" disabled><i class="correct"></i></button>',
+        HTML => ': <button class="ww-feedback-btn btn btn-sm btn-success" disabled aria-label="Correct">'
+		. '<i class="correct"></i>'
+		. '</button>',
         TeX => ''
     )@]*.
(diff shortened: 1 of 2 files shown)


assets/pg/Student_Orientation/feedback.pg
```diff
@@ -53,7 +62,8 @@ feedback message will tell you the correct answer. What number am I thinking of?
 
 If you type an answer and click to "Preview my Answers", you will see an info button[@
     MODES(
-        HTML => ': <button class="ww-feedback-btn btn btn-sm btn-info" disabled><i/></button>',
+        HTML => ': <button class="ww-feedback-btn btn btn-sm btn-info" disabled aria-label="Preview information"><i/>'
```
drgrice1, line 65: This will need to be updated to "Answer Preview" to match https://github.com/openwebwork/pg/pull/1039.

Alex-Jordan: Thanks, I got this changed now.

drgrice1 (approved):
Looks good.

pstaabp (approved):

PR #2367, opened 2024-03-21
Fix the OPL-update script (or really the download-OPL-metadata-release.pl script)

drgrice1:
The `download-OPL-metadata-release.pl` is extracting the downloaded metadata release to the wrong directory, and so the rest of the script fails.  It fails quietly, and doesn't actually restore any tables.  So a die statement was added to cover this case.


drgrice1:
One way to test this is to drop one of the OPL tables in the database (like OPL_DBsubject), and then run the script.  With the script currently in the develop branch that table will not be recreated.  With this pull request it will be.

pstaabp (approved):
Tested. Fixed error.

PR #2369, opened 2024-03-21
skeleton of student activity table/graph

Alex-Jordan:
This could grow to replace #2365. I want to ask if the module structure is right. Also I'm not sure about the purpose of `use Mojo::Base 'Exporter', -signatures;` and if I would need that here. Marking this as draft until further notice.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -81,6 +81,7 @@ my @modulesList = qw(
 	Date::Format
 	Date::Parse
 	DateTime
+	DateTime::Format::Strptime
 	DBD::mysql
 	DBI
 	Digest::MD5
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -207,6 +207,7 @@ sub verify {
 	{
 		$self->{was_verified} = 0;
 		$self->session(two_factor_verification_needed => 1);
+		$self->write_log_entry("LOGIN OK") if $self->{initial_login};
 		$self->maybe_send_cookie;
 		$self->set_params;
 	} elsif ($self->{was_verified}) {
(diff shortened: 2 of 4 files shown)


drgrice1:
I would say yes, but you also need more.  Add `our @EXPORT_OK = qw(studentActivityTable studentActivityGraph);`, and then in `templates/ContentGenerator/Instructor/Stats/student_stats.html.ep` (or any other file using an exported method) add `use WeBWorK::HTML::Activity qw(studentActivityTable);` to the beginning of the file, and call it with just `studentActivityTable` instead of `WeBWorK::HTML::Activity::studentActivityTable`.

Alex-Jordan:
Thanks, I made those changes and started work on a table. No styling yet, and not locked into any choices about what to show and what not to show.

Alex-Jordan:
This is what the activity table looks like now. Each login row is a `thead`. In this shot, there is a single test submission, then a single submission from a regular homework set. The links in the "Assignment" column take you to the set details page for that user (jumping down to the particular exercise for that row). The links in the "Problem Number" column take you to the past answers page for that user, set, problem number.

Screenshot 2024-03-24 at 1 07 16 PM

This is all still a draft, so people can check it out only if they are interested (especially to look for inefficiencies in the parsing/processing). I'm fairly satisfied for now with the table, and I may start the graph soon. One thing about this is it's not using Mojo templates, just because I'm not sure how to get started with those. Maybe for an HTML helper, that is OK.

Also, it seems that every successful login (regardless of the authentication method) adds a `LOGIN OK` line to the log, so I use that to identify logins. But the new 2FA logins were not adding that line, so I added it.

Alex-Jordan:
If this table is absurdly long (like towards the end of a semester when a user has had lots of activity) we could only print the login rows, and clicking on one of those could expand it to show the answer submissions from that session.

Also this is parsing the answer log file, and I guess it could be querying the `past_answer` table in the database instead. Details, details...

Alex-Jordan:
Looks like multiple `thead` is not valid HTML, so I will get around to updating that aspect.

PR #2378, opened 2024-03-25
tweaks affecting problems rendered in an iframe with iFrameResizer

Alex-Jordan:
This will pair with a PR to `pg`. This affects how problems behave when they are in an iframe that uses iFrameResizer. With this PR and the one in `pg`, the feedback popover should trigger the iframe resizing, but the MathQuill editor should not. And the MQ editor should never have a horizontal scrollbar. And when there is a vertical scrollbar, it should not obscure the buttons too much.

### DIFF
--- a/htdocs/js/RenderProblem/renderproblem.js
+++ b/htdocs/js/RenderProblem/renderproblem.js
@@ -28,7 +28,7 @@
 				send_pg_flags: 1,
 				extra_header_text:
 					'<style>' +
-					'html{overflow-y:auto;}body{padding:1px;background:#f5f5f5;}.container-fluid{padding:0px;}' +
+					'html{overflow-y:hidden;}body{padding:1px;background:#f5f5f5;}.container-fluid{padding:0px;}' +
 					'</style>',
 				...renderOptions
 			};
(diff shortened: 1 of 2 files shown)


drgrice1 (commented):
You should also probably add `data-iframe-height="1"` to the `footer` div on line 183.  I know that you don't show the footer for PreTeXt, but if someone were to show the footer, then it would be hidden outside of the iframe without the data attribute.

drgrice1 (commented):
You also need to add the `heightCalculationMethod: 'taggedElement'` option to the `iFrameResize` call on line 79 of `htdocs/js/RenderProblem/renderproblem.js`.

drgrice1 (approved):
This looks good in combination with https://github.com/openwebwork/pg/pull/1045.

pstaabp:
In the library browser with this branch, if a feedback button is at the bottom of a problem, this is what I see:

image

and the feedback is cutoff now without being able to scroll.

pstaabp:
Never mind.  Didn't pull the PG side as well.

pstaabp (approved):
Looks good.

PR #2379, opened 2024-03-26
Adjustment to the PGresource restructuring for PG.

drgrice1:
This is needed for the changes in https://github.com/openwebwork/pg/pull/1046.


drgrice1:
This also removes all of the netpbm `externalPrograms` since those are all no longer used.

pstaabp (approved):

Issue #2381, opened 2024-03-26
getting logged out if I visit problem editor or library browser


Alex-Jordan:
Since advancing my production server to `develop` (plus a few uncommitted changes), when I am in a course and visit the library browser, and I refine the subject select (say to Calculus - single variable) then I get an error like:

<img width="415" alt="Screenshot 2024-03-26 at 1 57 15 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/31a8edad-2113-4a77-959b-64fdd43f4d51">

Also if I go the problem editor, it fails to render the problem and I get:

<img width="1103" alt="Screenshot 2024-03-26 at 2 00 56 PM" src="https://github.com/openwebwork/webwork2/assets/4732672/b83fe1f9-48fe-48a3-82ef-e3e703100f44">

After either of these things, I am logged out (if I try to navigate somewhere, I'm taken to the login screen).

I have cleared all cookies for the site, and I reverted all session/cookie config settings to the distribution defaults, except I have only the admin course using 2FA. But I get the same behavior. I have not been able to reproduce this on my pure development server though, so I'm not sure what is going on.

Alex-Jordan:
And the same thing if I try to render problems in a set detail page.

Alex-Jordan:
Changing session management to key gets me past this for the problem editor and set details pages, but not for the library browser page issue.

drgrice1:
So I am seeing this issue with `session_management_via = 'key'` on the library browser page and the develop branch.  I will look into that.

I am also seeing the more general issue with cookies when using the lti-content-item-selection branch for pull request #2322 which I believe you are using.  So I will look into that also.

Alex-Jordan:
OK, thanks for your insight with this. If you do make changes to that branch, for now would you mind keeping any changes on new commits for now? It will help me follow along until this is ironed out.

I can always revert to not having that branch, but it would be nice to have that for spring term, and it's also helping to go through this process now with our D2L team.

drgrice1:
#2382 fixes the develop branch.  Most likely there was a merge issue for #2322, but I will investigate that now.

drgrice1:
Okay, the `lti-content-item-selection` branch is also fixed.  I pushed it in the last commit.  Note that that branch is not on top of develop anymore although it does merge (or rebase) cleanly.

Alex-Jordan:
Thanks, I can confirm things are working for me too. I'll close this.

drgrice1:
[...]

PR #2382, opened 2024-03-26
Fix the instructor_rpc endpoint usage for the library browser.

drgrice1:
The value attribute of the hidden key input does not have a value attribute.  Need to stop at the element in the first step.

This fixes the first issue observed by @Alex-Jordan in issue #2381.


drgrice1:
Note that this is the only part of issue #2381 that applies to the develop branch.  The rest only applies to the as yet unmerged #2322.  The issue has also been fixed on that branch though.

Alex-Jordan (approved):

pstaabp (approved):
This fixed it.

Issue #2384, opened 2024-03-29
Courses with names that end with "modelCourse" not appearing on develop.


Alex-Jordan:
Recently I moved our production server from 2.18 to develop. Prior to the move, we had model courses for classes with `0-` at the start of the course name. For example `0-mth060-modelCourse`. This was just a hack to make them appear at the top of the list.

Since the upgrade, these courses do not appear in the admin course. They don't appear in Course Listings, and they don't appear in the dropdown list of courses to copy from. But their folders are still there in the `courses/` folder, and their tables are still there in the database. So something is preventing the admin course from "seeing" them, and I guess it has to do with the "0-" at the start of the course name.

Alex-Jordan:
I changed the title to describe what is actually going on. Line 117 of `lib/WeBWorK/Utils/CourseManagement.pm` is excluding courses that end with `modelCourse` from showing up in a course list. Is there any reason not to make that check be an exact match on `'modelCourse'`?

Alex-Jordan:
Now that I can list those courses after changing that one line, I renamed them to end in `model-course` instead of `modelCourse`. I will revert the edited line in case there is a reason to leave it the way it is.

drgrice1:
Yeah, that is a mistake.  That should be `$_->basename ne 'modelCourse'`.  Note that `$_` will be the full path, and not just the directory name.  So for the model course it will be `/opt/webwork/courses/modelCourse`.

somiaj:
Are we going to want the model course to always have the exact name of `modelCourse`? What if an institution wants multiple model courses they can copy from when creating new courses but not want them to appear in course lists? Would it be over kill to add a configuration option (in `localOverrides.conf` or another appropriate place) that can list what the names of the model (or other courses) to skip are? Maybe even allow it to support regex/globbing? With the default being what is done here, skip only the course named `modelCourse`.

drgrice1:
There is no need for such a thing.  Those courses can be hidden if you do not want them to appear on the WeBWorK home page, and they absolutely should appear in other lists in the admin course.  So I say no to such an option.

Alex-Jordan:
I see, so that's why it was using `modelCourse$`.

Perhaps we should evolve the model course to become an actual course? There could be a course archive file that comes with the distribution, and it could even have the student orientation pre-loaded.

drgrice1:
[...]

PR #2385, opened 2024-03-29
Fix `listCourses` for the model course.

drgrice1:
Make `listCourses` skip only the `modelCourse`, and not all courses ending in `modelCourse`.

This fixes issue #2384.


Alex-Jordan (approved):

somiaj (approved):
I haven't tested this, but the code looks good and is a simple fix.

PR #2390, opened 2024-04-06
Restructure the problem sets page (the course home page).

drgrice1:
The table is replaced with list groups.  There are two or three list groups, depending on which view you select.  There are two options, "Show By Date" and "Show By Type".

If "Show By Date" is chosen (this is the default), then sets are grouped into the three categories "Open Assignments", "Unopen Assigments", and "Past Due Assignments".  Within the categories the sets are sorted by urgency (the previous status sort).

If "Show By Type" is chosen, then sets are grouped into the two categories "Regular Assignments" and "Tests/Quizzes".  Within each group the sets are ordered by name (the previous name sort).  Note that JITAR sets are grouped with "Regular Assignments".

Now all sets have an icon before them.  For tests that is the same icon that was used before.  For other assignments an open book icon is used. After the icon, the set name is shown with the status below it, and then other messages (reduced scoring and restricted release messages) below that.  The download set icon is to the right as before.

Instructors and students now see the same thing regarding the download set options.  There are no check boxes for an instructor to select and print multiple sets.  There are still numerous ways that an instructor can do that.  It is not needed here.  That can be done from the "Instructor Tools" page or even technically from here by clicking on any download icon, and then selecting the desired sets on the "Hardcopy Generator" page.

Several settings are saved in local storage that make it so that when you return to this page it is the same as it was the last time you visited it in the same browser.  The last chosen sort method will be what you see the next time you open the page, and collapsed sections will still be collapsed.

This is the result of a discussion in our meeting on Wednesday, April 3. As discussed, some wording may need to be ironed out.

For example, a past due set no longer says "Closed, answers available". Instead it says "Available for review. Answers available."  There are too many "availables" in that statement.

Also, is "Regular Assignments" good?  I was going to use "Homework Assigments", but we are trying to get away from the "homework" name.

### DIFF
--- a/htdocs/js/ProblemSets/problemsets.js
+++ b/htdocs/js/ProblemSets/problemsets.js
@@ -65,7 +65,8 @@
 
 		for (const [type, fallbackTitle] of [
 			['open', 'Open Assignments'],
-			['not-open', 'Unopen Assignments'],
+			['reduced', 'Reduced Scoring Assignments'],
+			['not-open', 'Future Assignments'],
 			['past-due', 'Past Due Assignments']
 		]) {
 			const section = addSection(
--- a/htdocs/js/ProblemSets/problemsets.js
+++ b/htdocs/js/ProblemSets/problemsets.js
@@ -1,0 +1,112 @@
+(() => {
+	const setListContainer = document.getElementById('set-list-container');
+
+	if (!setListContainer) return;
+
+	const addSection = (type, title, contents) => {
+		const isCollapsed = localStorage.getItem(`${settingStoreID}.collapsed.${type}`) === 'true';
+
+		const accordion = document.createElement('div');
+		accordion.classList.add('accordion', 'mb-3');
+
+		const item = document.createElement('div');
+		item.classList.add('accordion-item');
+
+		const header = document.createElement('h2');
+		header.classList.add('accordion-header');
+
+		const button = document.createElement('button');
+		button.classList.add('accordion-button', 'fs-5', 'fw-bold');
+		if (isCollapsed) button.classList.add('collapsed');
+		button.type = 'button';
+		button.dataset.bsToggle = 'collapse';
+		button.dataset.bsTarget = `#${type}-collapse`;
+		button.setAttribute('aria-expanded', isCollapsed ? 'false' : 'true');
+		button.setAttribute('aria-controls', `${type}-collapse`);
+		button.textContent = title;
+		header.append(button);
+
+		const collapse = document.createElement('div');
+		collapse.classList.add('accordion-collapse', 'collapse');
+		if (!isCollapsed) collapse.classList.add('show');
+		collapse.id = `${type}-collapse`;
+
+		const body = document.createElement('div');
+		body.classList.add('accordion-body', 'p-0');
+		collapse.append(body);
+
+		const list = document.createElement('ol');
+		list.classList.add('list-group', 'list-group-flush', 'rounded-bottom');
+		list.append(...contents);
+		body.append(list);
+
+		item.append(header, collapse);
+		accordion.append(item);
+
+		setListContainer.append(accordion);
+
+		collapse.addEventListener('shown.bs.collapse', () =>
+			localStorage.setItem(`${settingStoreID}.collapsed.${type}`, 'false')
+		);
+		collapse.addEventListener('hidden.bs.collapse', () =>
+			localStorage.setItem(`${settingStoreID}.collapsed.${type}`, 'true')
+		);
+	};
+
+	const setList = Array.from(setListContainer.querySelectorAll('.list-group-item'));
+	const showByDateBtn = document.getElementById('show-by-date-btn');
+	const showByTypeBtn = document.getElementById('show-by-type-btn');
+	const settingStoreID = `WW.${document.getElementsByName('courseID')[0]?.value ?? 'unknownCourse'}.${
+		document.getElementsByName('userName')[0]?.value ?? 'unknownUser'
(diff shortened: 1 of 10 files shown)


pstaabp:
Overall, this looks good.  "Regular Assignments" sounds fine.  "Unopen Assignments" doesn't sound right to me.  How about  "Unopened assignments"? or "Assignments Not Yet Open"?  

How about instead of "Answers available" to "Answers and Solutions may be viewed"?

Alex-Jordan:
We have it so that "open" or not is more than just a boolean state. There's a date in time when the transition happens, so "Future Assignments" might work.

Maybe "Available for review. Answers available." can just be "Answers available for review". If the answers are available, I think that implies that the set is closed since we enforce and answer date to be ≥ the close date.

I can't think of anything better than "Regular Assignments".

I haven't tried this, but looking at the code, am I right that it does not distinguish between sets that are before/after the reduced scoring date? Students tend to see the "past due" date as the reduced scoring date, not the close date. There could be some complaints if a set is in the "Open Assignments" category but is actually in the reduced scoring period.

drgrice1:
If a set is in the reduced scoring period, then the set is still open.  We could separated the open and reduced scoring sets though.  That would be easy with this structure.

"Future Assignments" instead of "Unopen Assignments" sounds good to me.  I don't like "Assignments Not Yet Open".  In fact, I initially had it as "Not Yet Open Assignments".  Neither of which sounds good.

So to summarize, when the "Show By Date" option is active, then we have "Open Assignments", "Reduced Scoring Assignments" (is that good wording), "Future Assignments", and "Past Due Assignments".  Does that sound good?

pstaabp:
> So to summarize, when the "Show By Date" option is active, then we have "Open Assignments", "Reduced Scoring Assignments" (is that good wording), "Future Assignments", and "Past Due Assignments". Does that sound good?

I think this sounds good.  I can't think of better wording than "Reduced Scoring Assignments".

drgrice1:
I made the changes as we discussed them.

I made one other change in the last commit that I think is better than how it was.  Rather than sorting the past due assignments from the last two weeks to the top (but still in chronological order), and then having the other past due assignments in chronological order after that, the past due assignments are simply in reverse chronological order.  Here the chronological order means by due date.  So the most recently past due assignment will be at the top, and the furthest back past due assignment at the bottom.  The message "Answers recently available" has been dropped, since the most recently past due are just at the top of the past due section anyway.

drgrice1:
I forgot to mention one other change made here of note.  Instead of using the `font-visible` and `font-hidden` styles for sets that are visible for students and sets that are hidden for students, if a set is hidden from students then the message "_(This set is hidden from students.)_" is shown.  The following picture shows what this looks like.
image

drgrice1:
The point of that last change is that the previous method of using only styles does not clearly convey what is meant.  An instructor may notice that one set's status is italicized and the other set's statuses are bold, but what does that mean?

Alex-Jordan:
These are all good changes, thanks for this move. I'll checkout the branch and just confirm things work soon.

pstaabp:
I also think this looks good.  Only thing I notice is that the difference in contrast between the "Show by Type" and "Show By Date" buttons are small. 
image

Is this just the default bootstrap colors?  I don't see anywhere else where we use this setup (with active versus inactive buttons next to each other) elsewhere in WW. 

I was also looking for some contrast tools on the web and everything seems to only do an object (or text) on a background, not buttons next to each other. 

This isn't a big deal to me, but just noticed it and wondered if it might be an accessibility issue.

drgrice1:
It shouldn't be a serious accessibility issue.  That is the contrast of the bootstrap active versus inactive colors.  Here is what it looks like on Bootstraps site with their default colors:
active

drgrice1:
@pstaabp: I switched to outline buttons.  That should resolve any contrast issues.  The math4-yellow theme needed a tweak to make that work though.

Alex-Jordan:
When a category (like "Future Assignments") is empty, would it be better for that block not to appear? I am seeing like:

Screenshot 2024-04-07 at 9 55 54 AM

and the two items in the middle seem like clutter. I suppose there is value in knowing affirmatively that nothing is in those categories. What do you think about it?

drgrice1:
Yeah, those are not supposed to be shown. When I converted it to JavaScript that was lost. I will fix that.

Alex-Jordan:
Other than that question (where the answer may be it's fine the way it is) I think this is great.

Alex-Jordan (approved):

somiaj:
This looks really good, I need to test it out.

I sometimes like to view tests separate from normal assignments, though probably don't want to split off more categories for this. What about some check boxes, 'Show assignment types:' and have a checkbox for normal assignments, tests, and just in time assignments (of course only include checkboxes for assignment types that are used and not show any checkboxes if only a single assignment type is use)?

drgrice1:
Empty sections are no longer shown.  Thanks for pointing that out.

somiaj:
Or maybe radio buttons, so users can show `all types`, `only assignments`, `only tests`, `only just in time`.

drgrice1:
@somiaj:  I don't see the need.  If you want to view tests only, then select "Show By Type" and just collapse the other sections.  As to JITAR assignments, it was decided that they should not be separated in our discussion.

somiaj:
@drgrice1 oh thanks, (what I get for not fully testing this out), didn't realize show by type did that...

drgrice1:
@somiaj:  The same thing happens with the develop branch.  That often happens with tooltips.

pstaabp (approved):
This looks quite good now.

PR #2391, opened 2024-04-06
Use an unpadded session.

drgrice1:
This changes the serialization method used by `Mojo::Sessions` to merely JSON encode the session without padding.  The default serialization method JSON encodes the session and then pads it with the letter `Z` to a length of 1025 characaters.

This results in much smaller session cookies (roughly one fifth the previous size).  Thus it is possible to be signed in to more courses at a time without hitting the cookie size limit.

This should perhaps only be considered if the limit on the number of courses that can be signed into at one time is deemed to be a big enough problem that something needs to be done.

### DIFF
--- a/lib/Mojolicious/WeBWorK.pm
+++ b/lib/Mojolicious/WeBWorK.pm
@@ -24,6 +24,8 @@ Mojolicious::WeBWorK - Mojolicious app for WeBWorK 2.
 
 use Env qw(WEBWORK_SERVER_ADMIN);
 
+use Mojo::JSON qw(encode_json);
+
 use WeBWorK;
 use WeBWorK::CourseEnvironment;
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
With this, I am able to log in to at least 12 courses without any issues. Not saying that is the cap I experienced, just that's how many I tried. Without this, it was usually 4. Sometimes it felt like I could enter a 5th, but then not succeed at navigating within that 5th.

drgrice1:
I think that we should go with this pull request.  To be honest, I don't understand what the advantage of padding the session is, and I don't see any real downside to removing the padding.  It seems that the padding just artificially inflates the cookie size.

pstaabp (approved):
I didn't test with many courses, but I think let's merge.

PR #2392, opened 2024-04-07
Make the LMS Context ID configurable via the course configuration page.

drgrice1:
This setting requires its own special `ConfigObject` submodule.  To make this work well, the `ConfigObject` setup needed to be changed a bit. Settings from the course settings table also now have their own `ConfigObject` submodule (the `courseTitle` is the only setting that uses this though).

A new series of permissions controls all configuration settings.  Each setting has its own permission.  Each of these permissions have the form `change_config_[var]`. Note that the user must also have the permission `modify_problem_sets` to change configuration settings. If a configuration option is not specifically set for a setting, then the `modify_problem_sets` permission alone is sufficient to change a configuration setting. The only one of these permissions that is set by default is `change_config_lms_context_id`, and that is set to `admin`.

So the previous `setting_table_courseTitle` would now be `change_config_courseTitle`.  That is not set by default since that is the same as `modify_problem_sets`, and so there is no point in it being set.  However, there is an example of it in `defaults.config`.

### DIFF
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -915,10 +915,10 @@ sub getConfigValues ($ce) {
 			var  => 'lms_context_id',
 			doc  => x('LMS Context ID'),
 			doc2 => x(
-				'This must be set in order to utilize LTI content selection. This must be configured in the LMS '
-					. 'first. If content selection from the LMS is attempted once the LMS is configured to use it, '
-					. 'then you will be shown the LMS context ID. Enter the context ID shown here, and then you '
-					. 'will be able to select assignments from this course, and import them into the LMS.'
+				'This must be set in order to utilize LTI content selection. The WeBWorK content item URL must be '
+					. 'set for the external tool in the LMS first. Then if content selection from the LMS is '
+					. 'attempted, you will be shown the LMS context ID. Enter the context ID shown here, and then '
+					. 'you will be able to select assignments from this course, and import them into the LMS.'
 			),
 			type => 'lms_context_id'
 		}


lib/WeBWorK/ConfigValues.pm
```diff
@@ -910,6 +911,17 @@ sub getConfigValues ($ce) {
 			),
 			type => 'boolean'
 		},
+		lms_context_id => {
+			var  => 'lms_context_id',
+			doc  => x('LMS Context ID'),
+			doc2 => x(
+				'This must be set in order to utilize LTI content selection. This must be configured in the LMS '
```
Alex-Jordan, line 918: What is meant by "This must be configured in the LMS first"? My understanding is that you discern what this is while in the LMS, then use that value to configure things over in WW.

Alex-Jordan (approved):
I tested that I could go to a course and change the context ID there, and see that reflected in the admin course. I have that one question about code commentary, but I'm marking for approval.

lib/WeBWorK/ConfigValues.pm
drgrice1: Yeah, too much pronoun usage, and it is not clear what "this" refers to.  I mean that first the webwork content selection URL needs to be set in the LMS.  Once that is done then the instructor can attempt content selection and will be shown the context id.  I will work on the wording.

drgrice1: I made a tweak to the wording.  Is that better?

Alex-Jordan: Yeah, I get it now.

pstaabp (approved):
Looks good now.  I think there is now plenty of room for the help pages on short problems.

drgrice1:
There was one change that didn't get into #2393.  That is a change to the help text for this setting.  So this pull request is now really only that fix.

pstaabp:
Reviewed the help text.  Looks good.  Will merge.

PR #2393, opened 2024-04-07
Make it so that changing tabs on the course configuration page does not need page reload.

drgrice1:
Instead all tabs are present always and are in bootstrap collapses.  Clicking a different tab just toggles visibility.  I.e., the usual nav tabs setup.
    
This means that changing tabs does not need a page reload.  The main advantage of this change is that when settings are saved, all settings are saved. Even those that are not in the active tab. So you can make changes in the "General" tab, and without clicking "Save Changes", select the "Optional Modules" tab and make changes there. Then select another tab and make changes there.  Then if you click "Save Changes" all of those changes are saved.
    
This is much more intuitive behavior, and what most users expect when seeing a tab layout such as is utilized on this page. As a result, those new to WeBWorK will often attempt to do what is described in the paragraph above, and then be surprised to see that the settings from the not visible tabs are all reset to what they were before the changes the instructor believed were just changed.
    
Although this is natural behavior I added a note in the help about this behavior.  This note is probably more for those that have been using WeBWorK, and expect the old behavior.
    
Note that this is built on top of #2392 to avoid needing to resolve conflicts with that pull request later.


pstaabp:
This looks good.  I'm wondering if someone makes a change, then changes tab (no save).  

How hard would it be to include some js to detect any changes and open a message that you must click save to make changes.

Alex-Jordan:
One thing about this. Certain config items affect the Course Config page. For example the theme (but there are others that affect what is present in certain dropdown menus). If you change the theme and save, you need to reload the page again to see that take effect. Before this PR, changing tabs is one way to reload the page again. With this PR, when changing tabs, you still don't see the change take effect. So it can feel like your change really did not take effect if you are not in the know about this.

So with this change, can we simultaneously make the "Save" button force the page to reload using the newly saved params?

drgrice1:
Unfortunately, that is still not possible.  In order for the changes to take effect, the course environment has to have been saved before a page load occurs since the theme is determined from the course environment when the request is received.

Alex-Jordan:
This might be a naive question, but it possible to submit the form (saving the new params) and then delay the page load so that page load process starts after the saving?

drgrice1:
A delay won't help.  However, it may be possible to save the configuration, and then redirect.  I am not sure though.

drgrice1:
Settings now appear to have immediate effect when saving from the course configuration page.  For example, if you change the theme and click "Save", when the page reloads the new theme is active.  This is accomplished by redirecting after saving.

Alex-Jordan (approved):
This is great, and thanks for working out the redirect. That's something that has bothered me for years.

drgrice1:
This now also asks the user if they really want to leave the config page if they attempt to do so when there are unsaved changes.

drgrice1:
Wait, that still needs some work.  It asks for confirmation when saving also.  That is not desirable.  I rushed that a bit.

drgrice1:
Okay.  That is fixed now.

pstaabp (approved):
This looks great now.

Alex-Jordan:
Perhaps related to this PR. In 2.18, when you are on some tab in the Course Config page and you save you will still be on that tab. Now in the 2.19 branch, if I'm on a tab and save, the page reloads on the "General" tab instead of whichever tab I was on. Is it possible to preserve the state of which tab is open?

PR #2395, opened 2024-04-11
Make the "Show Problem Grader" checkbox a button instead.

drgrice1:
The button is below the other usual problem/test submit buttons.  When clicked problem graders are shown, and the button becomes the "Hide Problem Grader" button.  Obviously when that is clicked the problem graders are hidden.

For now nothing is done with the check boxes for showing the various PG debugging information elements.  In thinking about this further, there really isn't a better way of handling those with the current structure.  Those are only shown for admin users (by default) so I don't think this is something that will be changed for this release.  I think that these should be removed from the problem page entirely, and instead only shown (for those with the permission to see them) on the PG editor page.  The PG code for handling these things should be completely revised and cleaned up.  It really is a hack for them to work with these check boxes to begin with.  On the PG editor page a modal dialog could be made available to view this data.

As to showing correct answers in a list after the problem (or at the bottom inside the probem?), I am not sure that is the best idea either.  We can talk about this more in a meeting, but I think it will end up being a rather ugly addition to the page.  We might as well go back to the results table.  This needs more thought.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -704,6 +704,9 @@ async sub pre_header_initialize ($c) {
 		return;
 	}
 
+	# Unset the showProblemGrader parameter if the "Hide Problem Grader" button was clicked.
+	$c->param(showProblemGrader => undef) if $c->param('hideProblemGrader');
+
 	# What does the user want to do?
 	my %want = (
 		showOldAnswers     => $user->showOldAnswers ne '' ? $user->showOldAnswers : $ce->{pg}{options}{showOldAnswers},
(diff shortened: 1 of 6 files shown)


Alex-Jordan (approved):
This works, and is a nice change.

pstaabp:
> As to showing correct answers in a list after the problem (or at the bottom inside the probem?), I am not sure that is the best idea either. We can talk about this more in a meeting, but I think it will end up being a rather ugly addition to the page. We might as well go back to the results table. This needs more thought.

First, since this looks like the previous problem grader, I take this as a comment as how to improve this and somewhat unrelated to this PR.  It might be nice to integrate into the feedback buttons or perhaps have a separate grader button instead.

drgrice1:
I should not have made that comment here.  That is unrelated to this pull request entirely.  However, it was part of the discussion that led to this pull request.   The comment belongs in https://github.com/openwebwork/pg/issues/1047.

drgrice1:
Note also, that this does not change the problem grader at all. It just changes the mechanism that activates it.

pstaabp:
Going to go ahead a merge.

Issue #2394, opened 2024-04-07
Session issue with problem editor


Alex-Jordan:
In develop, if I clear cookies and then log in and go to the PG editor, I can:

- enter `pi` as the answer
- click "Check Answers" over and over.
 
By my count on the 17th time, my session is killed and I have to log back in. This was noticed when exploring https://github.com/openwebwork/pg/issues/1050. 

Maybe this has to do with #2391.

drgrice1:
Just to clarify, are you saying that this is with #2391?  Or are you using the develop branch.  I haven't tested with #2391 yet, but I can not reproduce this on the develop branch.

drgrice1:
I can't reproduce this with #2391 either.

Alex-Jordan:
I see it with develop. Other details are I turned off 2fa, the os is oracle
8.7, locally I'm on a Mac with Firefox. Can't do more testing until late
tonight. Not sure now which session management is in play. But to be clear,
you tried checking 20+ times?

drgrice1:
Easily more than 20+ times.  I just kept clicking.

Alex-Jordan:
This is happening for me on `develop`. (Since I sort of have `develop` in production, I should clarify this is a "pure" `develop` branch. There are no uncommitted edits.) I've run `npm ci` in both repos just in the off chance that has anything to do with this. In `localOverrides.conf`, nothing relevant is changed from the default, except maybe `$twoFA{enabled} = 0;` which I don't think is relevant, but I mention it since it has to do with logging in.

The server is behind a firewall and I access it on PCC's internal network after activating a VPN client. It's name is `something-internal.pcc.edu` and I'm starting by deleting all cookies for `.pcc.edu`. To test this issue, I'm making a new course just to be sure there's nothing in a particular course causing trouble. I create the new course from the admin course, then enter it.

I go to the problem editor. I check in on cookies, and I have two at the moment. One of size 1460 (I think for the admin course) and one of size 1468. I type `pi` into the default problem's answer blank and click `Check Answers`. In the browser developer tools, the "1468" cookie flashes as if it was rewritten or otherwise accessed. Its size is still 1468. On the 17th click of "Check Answers", the problem no longer renders and I get this pink error:

`Rendering error: NetworkError when attempting to fetch resource.`

The "1468" cookie is now gone.

Now if I navigate anywhere, I find that I'm logged out. So I log in again. I try the same steps and this time I get bumped off after just 10 tries.

[...]

PR #2396, opened 2024-04-11
Wait for render completion before saving tmp file when clicking "Check Answers" in the PG editor.

drgrice1:
It seems that if the problem rendering request does not complete before the request to save the temporary file is sent, the streams can get crossed.  We all know how bad that is!

This seems to only occur when proxying hypnotoad in some way or another. Somehow hypnotoad manages to keep the streams straight when proxying is not involved.

This is really not specific to the "Check Answers" button, but could occur with any of the submit buttons in the problem.

Also, don't add the `passwd` and `key` to the form when rendering via the `render_rpc` endpoint unless cookies are disabled for the request, and in the case that `session_management_via` is "session_cookie" don't add the key to the form.

This fixes issue #2394.

### DIFF
--- a/htdocs/js/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/PGProblemEditor/pgproblemeditor.js
@@ -312,7 +312,7 @@
 		if (!problemForm) return;
 
 		for (const button of problemForm.querySelectorAll('input[type="submit"]')) {
-			button.addEventListener('click', (e) => {
+			button.addEventListener('click', async (e) => {
 				e.preventDefault();
 
 				// FormData does not support the characters in raw problem source.  URLSearchParams does.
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I'll check this out soon and see. But about this:
> This seems to only occur when proxying hypnotoad in some way or another. 

My servers are using hypnotoad with no proxying.

drgrice1:
Your load balancer counts as a proxy.  It is technically not a proxy, but I believe it is causing the same issue.

Alex-Jordan (approved):
This fixes the problem for me. Where it previously booted me after 17 clicks of the Check Answers button, I can go at least up to 40 now (and I'm sure indefinitely)

pstaabp (approved):
I'll approve based on code changes and @Alex-Jordan 's approval.  I was never able to see this problem.

PR #2397, opened 2024-04-12
Make the PGLM lab problem open via the `render_rpc` endpoint.

drgrice1:
Instead of opening the problem using the `editMode` hack of the usual problem page, render the problem directly in its own page via the `render_rpc` endpoint.  This means the style hacks in the PGML-lab problem are also not needed since the elements that were being hidden are not present in the page to begin with.
    
The PGMLLab link in a course's templates directory is not needed as the problem source is read directly from the assets directory, and submitted to the `render_rpc` endpoint via the `uriEncodedProblemSource` parameter.

Also, add examples of using PGML tags to the PGML lab problem.  The examples are in the "Substitutions" drop down menu of "Examples".  In particular there is an exmaple demonstrating how to use PGML tags for placement of the feedback button with MultiAnswer problems that use `singleResult => 1`.


templates/ContentGenerator/Instructor/PGProblemEditor.html.ep
```diff
@@ -114,24 +116,18 @@
...
+			<%= tag 'button', type => 'button',
+				title  => maketext(
+					'PG markdown syntax used to format WeBWorK questions. '
+					. 'This interactive lab can help you to learn the techniques.'
+				),
+				id => 'pgml-lab',
+				class  => 'reference-link btn btn-sm btn-info',
+				data   => { source => url_escape($pgml_lab_source), bs_toggle => 'tooltip', bs_placement => 'top' },
+				begin =%>
+				<%= maketext('PGML') =%>
```
pstaabp, line 129: Does this need to be in maketext? Is this translated?

drgrice1: It was in `maketext` before this pull request.  Although, it probably is not translatable.

drgrice1: I removed the `maketext` call.

pstaabp (approved):

Alex-Jordan (approved):
It's nice to have all the screen real estate available when rendering the sample PGML now.

PR #2398, opened 2024-04-13
Fix an overzealous `->c` addition from #2388.

drgrice1:
That pull request changed to passing the controller object to the ListingDB methods directly instead of using the WebworkWebservice's pass through methods that make it look like a controller object.  Thus `$self` needed to be change to `$self->c` in those calls.  However `$path` was not supposed to be changed to `$path->c`.

Just try to open the tags from any problem or in the library browser with the develop branch to see what this broke.  Oops.


Alex-Jordan:
I'm just going to merge this, since it is tied to #2388.

PR #2399, opened 2024-04-13
define set_to_display as empty string when undefined

Alex-Jordan:
Without this, if you go to the Library Browser, Set Definition Files button, and just click to view problems with no set selected, you get a perl error because of line 553 (the debug line) using an undefined string. It seems like if the string is undefined, it's OK to define it as `''` in the previous line to avoid this. And likewise with viewing problems from "Course Sets".


drgrice1 (approved):
Looks good.

drgrice1:
This is so minor, I will go ahead and merge it.

PR #2400, opened 2024-04-14
Don't try to read tags for unreadable files in the library browser.

drgrice1:
This is what causes the error when the problems in a set definition file are viewed and the problems in the set definition file can not be found.


drgrice1:
You can test this by trying to view the problems in the only OPL set definition file.

Alex-Jordan (approved):
Works as advertised.

pstaabp (approved):

PR #2401, opened 2024-04-14
Fix focusing the "name" input in the file manager.

drgrice1:
The form element does not have a `getElementsByName` method.  The `document` needs to be used for that.

This causes a console log error when the file manager page loads, and the "name" element is never focused.


Alex-Jordan (approved):
Looks good.

PR #2402, opened 2024-04-17
Add a "Use score from last check" button to the single problem grader.

drgrice1:
If the result score from the last time that "Check Answers" (or "Submit Answers") is different from the recorded score, then a new button is shown to the right of the "Problem Score" input (if the percent score is shown) and another new button is shown to the right of the "Point Value" input (if the point score is shown).  These buttons read "Use score from last check: score%" and "Use points from last check: points", respectively, where the score and points are the result that pg reported for the last time that "Check Answers" or "Submit Answers" was clicked. If one of those buttons is clicked, then the last checked score and points are inserted into the "Point Value" and "Problem Score" inputs.

The idea here is that the instructor may change answers in a student problem (for example to fix a typo), and then click "Check Answers" to see the result of those answers.  After doing so the scores from that check are now shown in the answer part scores above (if there are multiple parts).  Currently to make those scores take effect in the problem score that will be saved you need to change one of those part score inputs away and back.  This just adds a button that will take care of that for you.  Note that this last checked score must not be put into the "Point Value" and "Problem Score" inputs, since those must show the currently recorded score.

I am not sold on the wording shown on the buttons.  Does anyone have a better suggestion?

Note that the javascript for the single problem grader and the other grader has been separated into different files.  There is no overlap between the javascript used by the two anymore, so there is no reason for this to be in the same file.

Also prevent the default click behavior on 'help-popup's.  This can make links to '#' scroll to the top of the page.

This pull request is the result of a suggestion by @dlglin.


pstaabp:
This works well. About wording, I think it is fine,  "Use previous score" could be a shorter alternative.  I think it might be helpful to include a help popup to clarify what's going on to instructors who haven't seen it before (I'm think of myself next fall, when I forget about this new feature).

drgrice1:
I think that "Use previous score" is less clear than the current wording as to what the button does.  Which previous score does it refer to?  The current wording indicates that the score it will insert is the score from the last time that the problem was checked.  In the case that the instructor changed the answers in the problem and clicked "Check Answers" that would be the score from doing so.

In the case that the new button is present there is an additional message in the existing help popover explaining the button.  I don't think that adding a new help icon and popover is a good idea as it won't look good in the existing layout.

drgrice1:
After thinking about this more I realized that "Use previous score" is not only less clear, it is incorrect.  The score that is being offered by the new button is not a previous score at all. It is in fact the most current score.  The two scores presented are the currently recorded score for the student (in the problem score input), and the current score from grading the answers that are now shown in the problem (entered by the instructor).  So both are really current scores in a sense, but certainly neither are previous, and in any case the recorded score already in the input came chronologically before the score being offered by the new button.

Alex-Jordan (approved):
Marking this approved. My only observations is that the help popovers have multiple paragraphs, but are not structured with HTML`p`. There is a single `br` to separate paragraphs, but without any indentation or additional spacing between the paragraphs, it looks a bit like a bad line break rather than separation of paragraphs. A second `br` is a cheap fix that looks better imho. Or it could be structured with `p` and styled with css.

To be clear, I am marking for approval even if it stays as it is now though.

drgrice1:
I had noticed the same issue with the paragraphs.  I added `p` tags now with css classes to give the right spacing between paragraphs.

pstaabp (approved):
I'm good with this now.  I think I missed the help popover earlier--it's on the score/%, not on the button, so a little hidden, but I think helps.

PR #2403, opened 2024-04-18
Force scaffolds open when checking for a show me another variant.

drgrice1:
Currently when show me another checks for a new problem variant it does not force scaffolds to be open.  As such, if the first scaffold is static, later scaffolds are not, and the later scaffolds happen to have the same answers as the original problem, then show me another will claim that a new variant is not possible.

Here is a minimal test case:

```
DOCUMENT();
loadMacros('PGstandard.pl', 'PGML.pl', 'scaffold.pl', 'PGcourse.pl');

$a = Compute(random(1, 5));

Scaffold::Begin();
Section::Begin('Static Scaffold');
BEGIN_PGML
Enter [`5`]: [_____]{5}
END_PGML
Section::End();

Section::Begin('Dynamic Scaffold');
BEGIN_PGML
Enter [`[$a] + [@ 5 - $a @]`]: [_____]{5}
END_PGML
Section::End();
Scaffold::End();

ENDDOCUMENT();
```

With the develop branch (or any previous version of webwork2) this will fail to generate a new version.  With this pull request a new version will be generated.

Also update the comparison when checking answers or previewing from the show me another page to match the prior comparison.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ShowMeAnother.pm
+++ b/lib/WeBWorK/ContentGenerator/ShowMeAnother.pm
@@ -23,7 +23,6 @@ WeBWorK::ContentGenerator::ShowMeAnother - Show students alternate versions of c
 =cut
 
 use WeBWorK::Debug;
-use WeBWorK::Utils qw(wwRound);
 use WeBWorK::Utils::JITAR qw(jitar_id_to_seq);
 use WeBWorK::Utils::Rendering qw(getTranslatorDebuggingOptions renderPG);
 use WeBWorK::Utils::Sets qw(format_set_name_display);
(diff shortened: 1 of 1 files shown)


drgrice1:
To demonstrate why the  comparison for when answers are checked or previewed needs to be updated perform the following test.

Use this minimal example:
```
DOCUMENT();

$a = random(1, 5);

BEGIN_PGML
[! default icon !]{'defaulticon.png'}{100}

Enter [`[$a]`]: [_]{$a}
END_PGML

ENDDOCUMENT();
```

Add that problem to a set, and give your user the seed 1889 in the assignment.  Then open the show me another page for this problem in the set.  Then in the developer console enter `document.getElementsByName('problemSeed')[0].value = '804'`.  That is a seed that generates the exact same problem.  Now click "Preview My Answers" or "Check Answers".  With the develop branch the page will reload with the problem that is the same as your user's assigned problem.  With this pull request, the page will reload, but no problem will be shown since the check detected that the problem is the same.

pstaabp (approved):
Works as described and fixes the issue.

drdrew42 (approved):
👍

PR #2405, opened 2024-04-20
Fix more issues with show me another.

drgrice1:
There is a security vulnerability that allows a student to bypass the limit on the number of times that show me another can be used ($pg{options}{showMeAnotherMaxReps}) or to open a show me another problem before the required number of attempts at the original problem have been made.  The ShowMeAnother package adds a hidden showMeAnotherCheckAnswers field, and if this parameter is set it is assumed that the user is allowed to use ShowMeAnother. That can easily be hacked.  To do so construct the url `https?://server.edu/webwork2/course_id/set_id/problem_id/show_me_another?showMeAnotherCheckAnswers=1&checkAnswers=1&problemSeed=6543` where `problem_id` is a valid problem number for the set `set_id`.  Change the seed at will, and you will get a new version of the problem.  If that happens to generate the same problem as the seed for the assigned problem, it will not render a problem, but in no case will the showMeAnotherCount in the database be incremented, nor will the number of attempts at the original problem be checked. You can preview answers and check answers and do anything except click the "Show Me Another" page with this.
    
That issue was fixed by saving the problem seed found that generates a new problem variant in the session as well as the set id and problem id.  Thus it is not possible for a user to change these things. One advantage to this is that it is no longer necessary to verify that the problem is not the same as the assigned problem when checking or previewing answers.  The seed saved in the session can safely be used.
    
Another issue that was seen is that when the "Show Me Another" button is used on the show me another page until all uses are exhausted (assuming that `$pg{options}{showMeAnotherMaxReps}` is set to limit uses), the last time that it is used the tooltip on the button says "You can use this feature as many times as you want on this problem."  If the button is then clicked a new problem is not given and it says the uses are exhausted as it should, but the button shouldn't say that.

### DIFF
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -448,15 +448,16 @@ sub getConfigValues ($ce) {
 				var  => 'pg{options}{showMeAnother}',
 				doc  => x('List of options for Show Me Another button'),
 				doc2 => x(
-					'<ul><li><b>SMAcheckAnswers</b>: enables the Check Answers button <i>for the new problem</i> when '
-						. 'Show Me Another is clicked</li><li><b>SMAshowSolutions</b>: shows walk-through solution '
-						. '<i>for the new problem</i> when Show Me Another is clicked; a check is done first to make '
-						. 'sure that a solution exists </li><li><b>SMAshowCorrect</b>: correct answers <i>for the new '
-						. 'problem</i> can be viewed when Show Me Another is clicked; note that <b>SMAcheckAnswers</b>'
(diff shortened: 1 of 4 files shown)


drgrice1:
After fixing the issues in #2403 I kept seeing other issues with the implementation which led to this pull request.

drgrice1:
Also note that this is on top of #2403.

pstaabp (approved):
Looks good. It seems that adding problemSeed to the query string will update the problem in showMeAnother mode, but will also update the database.

drgrice1:
If you use the url with the parameters I gave it will not even update the database with the develop branch.  With this pull request none of that added to the query will have any effect.

PR #2406, opened 2024-04-20
Add a "Show Correct Answers" button to the problem and test pages.

drgrice1:
When this button is used correct answers will be shown with no other content in the feedback popover, scaffolds will all be open, and the feedback popovers will open immediately on page load. This is an instructor only option for now.  I don't think this will be a good idea to enable for students.
    
Furthermore, the "Show Correct Answers" buttons on other instructor pages (pg problem editor, problem library, set detail page, statistics, and manual problem grader) now use this for showing correct answers.
    
This works well for most problems, but there are some cases where answers are tightly grouped particularly when there are on the same line that the popovers can overlap in not the best way.
    
This is one potential resolution of https://github.com/openwebwork/pg/issues/1047.

Note that this requires https://github.com/openwebwork/pg/pull/1065 to work.

Also, this is built on top of #2395, and will have conflicts with #2405 that will need to be resolved.


pstaabp:
I think this is nice for all correct answers.  Here was a problem I was looking at:

image

and I roughly cut the padding in half on the answers a bit to get: 

image

Needs a bit of work, but might be nice to have answers that are 1 line high to be similar in height to the buttons so there is less overlap.  (I know this won't work with all problems.)

drgrice1:
Cutting the padding barely helps.  Generally speaking there is a problem authoring issue here.  Even without the feedback shown you can see in your screen shots that the inputs are smashed together from one line to the next.  That does not look good, and a problem author should do a better job of insuring spacing between lines.  Bad problem authoring practices that make problems not look good in general is a large part of the issue with the feedback popovers overlapping to begin with.

drgrice1:
I reduced the padding on the popover in the special case that correct answers are the only thing shown.  It helps a little with problems that smash things together.  But doesn't help the generally poor appearance of those problems to begin with, and doesn't help if the answers themselves have more height or the case of tightly grouped answers on the same line.

Alex-Jordan:
Those conflicts you predicted are here now.

drdrew42 (approved):
👍

pstaabp (approved):

Alex-Jordan:
I checked out the 2.19 release candidate branch, then created a new branch from that and merged this PR branch into it. I ran `npm ci` and restarted `webwork2`.

Now when I go to a set, I do see the "Show Correct Answers" button. However when I click it, I am not seeing the correct answers appear. Furthermore, I see the red incorrect feedback button. So it seems that something is not working right, or it's been long enough since I was doing testing that I'm forgetting something about testing protocols.

Here is what it looks like immediately following clicking that button.

Screenshot 2024-05-31 at 1 54 23 PM

drgrice1:
Just to make sure, you checked out https://github.com/openwebwork/pg/pull/1065 also, right?

Alex-Jordan:
Nope, I did not. That will explain it all.

PR #2413, opened 2024-04-26
The fallback original_student_ans can not be in math mode text in hardcopy.

drgrice1:
There are many answers for which the `original_student_ans` does not work inside `$\displaystyle \text{...}$` which is what is currently used for student answers if `preview_latex_string` is not defined or is empty.

For example, consider the following MWE.

```
DOCUMENT();
loadMacros('PGstandard.pl', 'PGML.pl', 'contextFraction.pl', 'PGcourse.pl');
$b   = random(2, 4);
$ans = Compute("3^($b)");
Context()->operators->undefine('+', '-', '*', ' *', '* ', '^', '**');
BEGIN_PGML
Simplify [`3^[$b]`]: [_]{$ans}
END_PGML
ENDDOCUMENT();
```

Add that problem to a set and submit the answer `3^$b` (whatever `$b` is for you) without actually simplifying.  Since exponents are disabled for this problem, that answer does not parse into a MathObject value, and so `preview_latex_string` is undefined.  Now if you generate a hardcopy for this set including "Student answers" hardcopy will fail since the `^` character must be in math mode and not inside `\text`.

So I see no alternative, but to go back to using verbatim in this case.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -1317,7 +1317,10 @@ async sub write_problem_tex ($c, $FH, $TargetUser, $MergedSet, $themeTree, $prob
 			} elsif (defined $pg->{answers}{$ansName}{original_student_ans}
 				&& $pg->{answers}{$ansName}{original_student_ans} ne '')
 			{
-				$stuAns = "\\begin{verbatim}$pg->{answers}{$ansName}{original_student_ans}\\end{verbatim}";
+				$stuAns =
+					"\\begin{verbatim}"
+					. ($pg->{answers}{$ansName}{original_student_ans} =~ s/\\end\{verbatim\}//gr)
+					. "\\end{verbatim}";
 			} else {
 				$stuAns = "no response";
 			}
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -1313,15 +1313,15 @@ async sub write_problem_tex ($c, $FH, $TargetUser, $MergedSet, $themeTree, $prob
 			if (defined $pg->{answers}{$ansName}{preview_latex_string}
 				&& $pg->{answers}{$ansName}{preview_latex_string} ne '')
 			{
-				$stuAns = $pg->{answers}{$ansName}{preview_latex_string};
+				$stuAns = "\$\\displaystyle $pg->{answers}{$ansName}{preview_latex_string}\$";
 			} elsif (defined $pg->{answers}{$ansName}{original_student_ans}
 				&& $pg->{answers}{$ansName}{original_student_ans} ne '')
 			{
-				$stuAns = "\\text{" . $pg->{answers}{$ansName}{original_student_ans} . "}";
+				$stuAns = "\\begin{verbatim}$pg->{answers}{$ansName}{original_student_ans}\\end{verbatim}";
 			} else {
-				$stuAns = "\\text{no response}";
+				$stuAns = "no response";
 			}
-			$stuAnswers .= "\\item\n\$\\displaystyle $stuAns\$\n";
+			$stuAnswers .= "\\item\n$stuAns\n";
 		}
 		$stuAnswers .= "\\end{itemize}}$corrMsg\\par\n";
 		print $FH $stuAnswers;


lib/WeBWorK/ContentGenerator/Hardcopy.pm
```diff
@@ -1313,15 +1313,15 @@ async sub write_problem_tex ($c, $FH, $TargetUser, $MergedSet, $themeTree, $prob
 			if (defined $pg->{answers}{$ansName}{preview_latex_string}
 				&& $pg->{answers}{$ansName}{preview_latex_string} ne '')
 			{
-				$stuAns = $pg->{answers}{$ansName}{preview_latex_string};
+				$stuAns = "\$\\displaystyle $pg->{answers}{$ansName}{preview_latex_string}\$";
 			} elsif (defined $pg->{answers}{$ansName}{original_student_ans}
 				&& $pg->{answers}{$ansName}{original_student_ans} ne '')
 			{
-				$stuAns = "\\text{" . $pg->{answers}{$ansName}{original_student_ans} . "}";
+				$stuAns .= "\\begin{verbatim}$pg->{answers}{$ansName}{original_student_ans}\\end{verbatim}";
```
Alex-Jordan, line 1320: Why have the `.=` operator here?

Alex-Jordan:
This made me realize that by printing student answers at all in the PDF, we are open to insertions.

Turn off MathQuill (since that complicates the following examples). Without this PR, I could enter an answer like `}$ Hello $\text{`. With this PR, I could use `\end{verbatim} Hello \begin{verbatim}`. The `Hello` could be a chunk of malicious code. Maybe there is not too much that could be done, but I think a person could at least cause trouble with some LaTeX code that is a drain on the CPU.

If we want to show the student answers int he PDF, one way to prevent an insertion would be to scan the student answer for an unused character, and use that as the delimiter in `\verb=...=`, in place of the `=` (if the student answer contains an `=`). We could also just not show the student answer in the PDF. Does it add that much value?

lib/WeBWorK/ContentGenerator/Hardcopy.pm
drgrice1: Thanks for pointing that out.  It should not have been.  That should give warnings in fact because `$stuAns` would not have been initialized.

drgrice1:
That is certainly a problem.  With MathQuill disabled submit `\end{verbatim}\newcount\X\X=1\loop\the\X\endgraf\advance \X by 1\unless\ifnum \X<0\repeat\begin{verbatim}` (or without this pull request submit `}$\newcount\X\X=1\loop\the\X\endgraf\advance \X by 1\unless\ifnum \X<0\repeat$\text{`).  Then generate hardcopy showing student answers.  The pdflatex process runs for a long time.  Presumably, tex is smart enough to end it eventually, but I didn't wait long enough.  I waited a full minute before killing the process.

I don't think that not allowing student answers in hardcopy is a good option.  I in fact need that.  We are required to submit student artifacts for assessment, and the only way to do that with WeBWorK assignments is to generate hardcopy showing student answers.  So we need to find a way to sanitize answers.  Note that essay answers already do this.  If you try to enter the above strings for an essay answer, the loop does not occur, and hardcopy generation succeeds.  So we may need to implement that more generally.  That is done on the PG side (on lines 94-99 of PGessaymacros.pl).  We probably can't do that to the `original_student_ans` in PG for other answers, so this may be tricky.

drgrice1:
Maybe your suggestion of finding a character not in the student answer to use for the verb delimiter will work.  The problem is, how do you find such a character?

Alex-Jordan:
We do this in pretext for something else. There is a list of potential
delimiter characters and it looks through the string until it finds an
available one. If it reaches the end of the list (which should never happen
in practice) the output becomes a message about not finding a delimiter.
But here, maybe the student answer could be truncated and then do the
search again.

drgrice1:
How about just removing any `\end{verbatim}`'s found in the `original_student_ans`.  Then the student's answer can't break out of the `\begin{verbatim}`.  I added that to this pull request.

drgrice1:
There is not reason that a student's original answer should ever have a `\end{verbatim}` in it unless a student is attempting to do something malicious.

Alex-Jordan:
What about like
\end{  verbatim  }
and other variants? All I can think of is whitespace surrounding 'verbatim'
and after 'end'. Are there other ways to end the environment? I'm not sure
what can be done with low level tex.

drgrice1:
I thought of that, and tested with spaces. They don't work to end the environment. LaTeX is particular about that. I don't think there is a low level TeX way to break out of a verbatim either.  If there were, then I don't see that using a `\verb=` would work either.

Alex-Jordan:
I just read about it and `verbatim` is one of a few special environments that parses ahead to find exactly `\end{verbatim}`, so it seems safe.

drdrew42 (approved):
👍

pstaabp (approved):

PR #2414, opened 2024-04-30
Don't show last answers when a new ShowMeAnother problem is shown.

drgrice1:
When I made the "Show Me Another" button an actual submit button in #2405, and unintended consequence is that the answers from the original problem or previous show me another problem persist when show me another is initialized with a new problem. I don't know why I thought that would work.  So that will need to be a link again.


drdrew42 (approved):
~~This should probably be included in the 2.19-RC~~ LOL, it already is

pstaabp (approved):

PR #2420, opened 2024-05-17
Fix warning that occurs on perl 5.36+ for using @_ in function signatures

pstaabp:
This fixes the warning and seems to function identical to the previous version.  I'm not sure about handling the initializers of `undef` in the previous version.


drgrice1 (approved):
Your code is correctly handling the initialization of `$name` and `$value` to `undef`.  That happens implicitly.  The only reason those were needed before was to allow not passing the parameters at all.  So if you call `$c->param` or `$c->param('param_name')`, then it will not be an error.

This looks good.

Issue #2419, opened 2024-05-17
Disappearing quiz versions when updating the date for a user.


pstaabp:
In short, if the date of a quiz is changed via the "Sets assigned to user" page can remove the versions of that quiz.  This is the RC version of 2.19 -- just updated our production server.  Here's the steps:

1. Create a quiz, assign it to yourself, take the quiz
2. Go to the "Sets assigned to YOU" page (via Accounts Manager) and locate the quiz.  You should also see the version:
<img width="684" alt="image" src="https://github.com/openwebwork/webwork2/assets/879338/a5d0dc17-aa6f-4dfb-87d8-d895b5ed24ee">

3. Change the date on the quiz (not the versioned quiz) and click "Save Changes".
4.  The versioned quiz will disappear:

<img width="654" alt="image" src="https://github.com/openwebwork/webwork2/assets/879338/25c99a34-f729-4b26-910f-013286a1cfe9">

stassenm:
I will add that it is not necessary to change any dates.  It is sufficient for step 3 to be simply "Save Changes." Even with no changes made, all versioned quizzes for that user are deleted.

pstaabp:
Fixed by #2421

Alex-Jordan:
IIRC, I started the redesign of that page and Glenn made it better. I'm sorry if it was my doing!

We've been using the develop branch since late March (or essentially what it was at that time) and had not run into this. Yesterday I applied the patch @drgrice1 made. This morning I have a report of missing quizzes, so naturally at first I thought I messed something up with applying the patch. But it seems clear now that one instructor lost some quizzes a little while ago and only reported it this morning. Testing directly, I am not able to make quizzes get deleted in this manner so I think the patch is definitely working.

drgrice1:
It is as much my fault as yours.  I knew about the importance of the hidden inputs for versioned sets on that page, and that they need to be after the check boxes to work right.  I should have seen this when those changes were made.

Hopefully your instructor did not lose important student quiz data.

PR #2421, opened 2024-05-21
Fix unintentional versioned set deletion on the user detail page.

drgrice1:
The issue is that with the restructuring of the page the hidden input corresponding to the "assigned" checkbox for a versioned set was not moved with the checkbox.  As a result it now comes earlier in the DOM. This means that the order of the values for the hidden input and checkbox by the same name are now reversed in the submitted form.  Since the WeBWorK::Controller param method returns the first one the code now gets the hidden input value of "delete" instead of the check box value of "assigned" even when the check box is checked.

The fix is simply to move the hidden input back to after the checkbox.

This fixes issue #2419.

### DIFF
--- a/templates/ContentGenerator/Instructor/UserDetail/set_row.html.ep
+++ b/templates/ContentGenerator/Instructor/UserDetail/set_row.html.ep
@@ -17,9 +17,6 @@
 						params => { editForUser => $userID }
 					) =%>
 			</span>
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):

pstaabp (approved):
Looks good.

PR #2423, opened 2024-06-07
Change double quotes to single quotes in minion db connect strings

dlglin:
If you use double quotes as shown in the example it tries to interpolate the `@`.  This was the case for mysql.  I haven't tested with Postgres, but expect the same behaviour.

Also, should the sample have `dbuser:dbpasswd` for Postgres as well, or is there some reason to not include a password?

### DIFF
--- a/conf/site.conf.dist
+++ b/conf/site.conf.dist
@@ -320,8 +320,8 @@ $job_queue{backend} = 'SQLite';
 # Database dsn for the Minion job queue. Some examples of settings for the
 # respective backends follow. The postgres and mysql examples will need to be
 # modified to work. The default sqlite setting will work as is.
-#$job_queue{database_dsn} = "postgresql://dbuser@/webwork2_job_queue";
-#$job_queue{database_dsn} = "mysql://dbuser:dbpasswd@localhost/webwork2_job_queue";
+#$job_queue{database_dsn} = 'postgresql://dbuser@/webwork2_job_queue';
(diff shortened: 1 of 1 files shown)


pstaabp:
Looks good. It wouldn't hurt to include the message to use single quotes in the comment above these lines.

drgrice1:
Good catch.  I have been using the mysql backend in testing for a while, and I had to switch to single quotes for that.  I probably should have thought to make a pull request to change this at that time.

As to the postgres example setting, postgres works differently than mysql and does not require a password or hostname. If connections are only allowed via sockets and the server can only be signed into directly by administrative users, then there really is no need for those things.  So the example as given will actually work in those cases.  Note the example was also taken directly from the Minion Pg backend documentation (see https://docs.mojolicious.org/Minion/Backend/Pg) modified only with the generic username and webwork2_job_queue database name.

drgrice1:
Actually, even my comment about the server only being able to be signed into directly by administrative users is not needed.  Since postgres requires that users are actual system users and you can only access the database as that system user via sockets, it is still secure to not require a password if there are non-administrative users that can sign into the system.  Also, even if remote access is allowed (it is by default), when accessing the datbase via sockets a password is still not needed.  It is needed if accessing it when a hostname is specified (even if the hostname is localhost).  Generally, postgres just uses a completely different security model than we are accustomed to with mysql.

Alex-Jordan (approved):

drgrice1 (approved):

drgrice1:
This has two approvals, so merging.

PR #2425, opened 2024-06-08
Make the max score for LTI content item requests be the webwork max score

drgrice1:
This is done for both versions of LTI.  However, Canvas does not respect the setting for LTI 1.1.  It does for LTI 1.3.  Moodle respects both. As usual, I am not sure what D2L will do.


Alex-Jordan (approved):
This works for me with D2L and LTI 1.3. I don't have a content selection tool set up for LTI 1.1, and would need the D2L admins to set that up. So I can't test that, at least not right now.

PR #2430, opened 2024-06-11
Prevent protected files from being modified by unprivileged users.

drgrice1:
By "protected files" I mean the files listed in the `$uneditableCourseFiles` list in the course environemt that is originally defined in `defaults.config`.  This usually contains the files `course.conf` and `simple.conf`.

As before, only users with the "edit_restricted_files" permission can edit these files.  The only change here is that `Mojo::File::realpath` is used instead of the `File::Spec::canonpath`.  This is a more reliable method since it properly resolves to a path without `..` (canonpath does not do this). It is possible to use `..` to create a filename that wouild still resolve to one of the protected files.  For instance, `templates/../course.conf` could be used to overwrite `course.conf`. The canonpath method would not correctly realize that those are the same file paths.

If a user that does not have the permission to "edit_restricted_files", and the user attempts to upload a file in the list, then the user will be forced to choose a new name for the upload, or cancel.  The overwrite option will not be given.  Users with the "edit_restricted_files" permission can still overwrite the file.

If a user that does not have the permission to "edit_restricted_files" extracts an archive that would overwrite a file in the list, then that file will be skipped and will not be extracted.  A message will notify the user that this file from the archive is not extracted.

Deletion or renaming of one of the files in the list is not allowed. Not even for users that have the "edit_restricted_files" permission. These files are required, and there is no reason that this should ever be done.

Finally, required course directories (all directories in the `courseDirs` hash in the course environment) are not allowed to be deleted or renamed by any user.  These directories are required, and so there is not reason that even a user with the "edit_restricted_files" permission should be allowed to do this.

This is intended to resolve discussion #2427.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -754,10 +754,12 @@ sub Upload ($c) {
 			&& grep { Mojo::File->new("$dir/$name")->realpath eq Mojo::File->new("$c->{courseRoot}/$_")->realpath }
 			@{ $c->ce->{uneditableCourseFiles} };
 
-		if (($invalidUploadFilenameMsg || $isProtected)
-			|| !$c->param('overwrite')
-			|| $action ne 'Overwrite'
-			|| $action ne $c->maketext('Overwrite'))
+		if (
+			($invalidUploadFilenameMsg || $isProtected)
+			|| (!$c->param('overwrite')
+				&& $action ne 'Overwrite'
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
I tested all the things mentioned in the description and they all work.

drgrice1:
I added a second commit to this that addresses #2437.  I also slightly tweaked how protected file uploads are handled.  Instead of just refusing to do it, it now gives the user a chance to rename it.

Alex-Jordan:
The changes to `checkName` here are related to #2487. It seems `checkName` is used two places outside of this file that will need adjustment.

PR #2431, opened 2024-06-11
Fix hidden sets being shown to students.

drgrice1:
Currently hidden sets are still listed for students.  I am not sure when this regression was added, but this fixes it.


somiaj (approved):
Looks good.

dlglin (approved):
Works for me.

Issue #2429, opened 2024-06-11
Is UTF-8 broken in infoboxes after upgrade to 2.19?


robert-marik:
Hello, after the upgrade to 2.19 I have problems that the texts from infoboxes (the text on the right margin of the page) to the server and to the course are not displayed properly. The non-ASCII letters are broken. The infoboxes to the homeworks are OK, but load as corrupted if I try to edit. See the following image. The text should be "Toto WeBWorK zadání 10_Shrnuti z předmětu Aplikovaná matematika má datum odevzdání pátek 31. května 2024 v 20:00 hodin (podle CEST). Informace o systému WebWorK v češtině nalezenete zde. Webová stránka předmětu je zde."

![image](https://github.com/openwebwork/webwork2/assets/2342862/b6c32f5d-b22f-4262-82eb-fc07954f960e)

I used https://webwork.maa.org/wiki/Release_notes_for_WeBWorK_2.19#Upgrade_Instructions for upgrade, nothing else has been changed. The problem also appears if I create new course and write something to the Course information.

robert-marik:
Thank you very much for checking this issue.

* Attached the file our_site_info.txt [our_site_info.txt](https://github.com/user-attachments/files/15787919/our_site_info.txt) which renders as follows
  ![image](https://github.com/openwebwork/webwork2/assets/2342862/16aa9419-79a7-4e42-8aec-dee65098b310)

* Attached the file course_info.txt [course_info.txt](https://github.com/user-attachments/files/15787999/course_info.txt) which renders as follows.
  
  ![image](https://github.com/openwebwork/webwork2/assets/2342862/d7c84052-6766-49b5-937c-07409f8bd4e7)

* Attached the pg file with header to the set. The pg file appears fine on the webpage with the assignment, but turns broken when opened in the editor (the txt suffix added to allow upload to github). 
[ASimpleCombinedHeaderFile.pg.txt](https://github.com/user-attachments/files/15788070/ASimpleCombinedHeaderFile.pg.txt)
  
  ![image](https://github.com/openwebwork/webwork2/assets/2342862/93e61fa1-40ff-4e9a-933b-4ea7a6fbe025)

* Observed another issue: with essay question. Compare how it appears when the student works with the question and how it appears if I click "Grade problem". The file is https://github.com/robert-marik/hw-webwork/blob/master/AM_10/15.pg
   
[...]

PR #2432, opened 2024-06-12
Increase the required version of Mojolicious to 9.34.

drgrice1:
This version added the encoding option to the `Mojo::File::slurp` method that is used in the `WeBWorK::Utils::Files::readFile` method.

An alternate approach to this pull request would be to not use the optional parameter to the `Mojo::File::slurp` method, and instead do the decoding ourselves.  Then version 9.22 of Mojolicious (the version in the Ubuntu 22.04 repositories) could still be used.


drgrice1:
By the way, this is the fix for issue #2429.

Alex-Jordan (approved):

taniwallach (commented):
I also got hit by the problem editor having problems with problems with UTF-8 characters (Hebrew) on a local install of 2.19.

Installing Mojolicious from CPAN fixed that for me. I ended up with Mojolicious-9.37, and now the problem editor of 2.19 works properly.

The downside of installing simply from CPAN is that the version installed will be depended on what is the current version of Mojolicious. Hopefully that will not cause different problems in the future.

drgrice1:
If at some point a newer version causes issues, then we will need to document installing a specific version.  You can do this using `cpanm` with `cpanm Mojolicious@9.37`.  However, at this point any version of Mojolicious that is version 9.34 or newer will work.

We also would need to implement a maximum version check in `check_modules.pl`.

drgrice1:
I also updated the docker build to use the latest version of Mojolicious from CPAN in this pull request now.

pstaabp:
Also, requiring newer version of Mojolicious might be beneficial for other features. I don't know any right now, but this could be a reason to go this route instead of #2433.

pstaabp (approved):

PR #2433, opened 2024-06-12
Don't have Mojolicious UTF-8 decode files. Instead do it ourself.

drgrice1:
In the `WeBWorK::Utils::Files::readFile` method UTF-8 decode the file contents after calling the `Mojo::File::slurp` without passing the UTF-8 decode parameter.

This is an alternate approach to #2432 that makes it so that version 9.22 of Mojolicious (the version in the Ubuntu 22.04 repositories) can still be used.

### DIFF
--- a/lib/WeBWorK/Utils/Files.pm
+++ b/lib/WeBWorK/Utils/Files.pm
@@ -19,6 +19,7 @@ use Mojo::Base 'Exporter', -signatures;
 use File::Spec::Functions qw(canonpath);
 use File::Find;
 use Mojo::File qw(path);
+use Mojo::Util qw(decode);
 
 our @EXPORT_OK = qw(
 	surePathToFile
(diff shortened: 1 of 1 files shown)


drgrice1:
Pick between #2432 and this pull request.  Would we rather require a newer version of Mojolicious, or allow the older version to still work?  Now taking votes.  To rephrase the vote in terms of deprecation: Do we deprecate usage of Mojolicious 9.22 for webwork2?

dlglin:
It's funny how you can affect the outcome of a vote just by the way you phrase the question.

RHEL 8 packages Mojolicious 8.42.
Ubuntu 24.04 currently packages 9.35.

That means that RHEL users will need to install Mojolicious from CPAN regardless, and Ubuntu 24 users should be able to use the packaged version.  Based on this I'm fine with requiring the newer version of Mojolicious.

somiaj (approved):
I'd prefer this approach, as I find installing system packages easier to manage than CPAN. Debian stable uses 9.31, so would also have issues. Seems this would be a fairly easy change to revert in a year or two if we wanted to wait a bit before requiring a newer version of mojolicious.

PR #2435, opened 2024-06-12
Allow access to the instructor_rpc route without revoking proctor authorization.

drgrice1:
This fixes a (rather serious) bug in which proctor authorization is revoked when the instructor_rpc endpoint is used by students during a test to get the server time.  This means that every single time that a student changes pages or hits preview during a proctored test, the student needs to re-renter the proctor username and password.

The only command that the instructor_rpc allows for student users is this command that gets the server time.  All other commands are instructor only, and should be fine to allow without revoking proctor authorization.


Alex-Jordan (approved):

somiaj (approved):
Tested that this fixes the issue.

PR #2436, opened 2024-06-12
Synchronize the "add user" page with the "edit user" page.

drgrice1:
That is the same fields are shown in the same order on the "add user" page as on the "edit user" page.  This means that "status" and "permission" can now be set when a user is created, instead of needing to first create a user, and then edit the user to set those.

The defaults in the status and permission dropdowns that are selected are the defaults from the course environment.  This means that if those are not changed, the behavior will be the same as before.

The order being different on this page than on the "add user" page and the general user list page has always annoyed me, and is in my opinion quite confusing to new webwork users.  Old webwork users will need to adapt to the new order of course.

Note that this is related to (and part of) discussion #1785.  I want to also deal with the password which is the main point of that discussion.  What I would like to do is have the password be another field that is shown on both the add user and edit user(s) pages, and remove the "Password" tab from the main "Accounts Manager" page.  This is not hard to do, but to do it easily in this way would mean dropping the behavior of setting the default password to being the student id.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/AddUsers.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/AddUsers.pm
@@ -32,6 +32,10 @@ sub initialize ($c) {
 
 	my $user = $c->param('user');
 
+	# Make sure these are defined for the template.
+	$c->stash->{statusValues}     = [];
+	$c->stash->{permissionValues} = [];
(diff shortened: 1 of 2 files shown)


Alex-Jordan (approved):
This is nice.

pstaabp (approved):

PR #2440, opened 2024-06-19
Remove instructions about moving the admin course after creation.

dlglin:
It seems easier to just create the admin course with the desired name.  The 2.19 installation instructions on the wiki now say to do it this way as well.


drgrice1 (approved):
Looks good to me.  We will need to make sure to update the installation instructions to advise installers to change this setting.

pstaabp (approved):

PR #2444, opened 2024-06-25
Remove mention of LTIConfigValues.config file from comment.

somiaj:
This file no longer exists, so remove mention of it from the configuration comment.


drgrice1 (approved):
Good catch.

pstaabp (approved):

PR #2445, opened 2024-06-25
Switch the default latex image svg method to dvisvgm.

drgrice1:
It is time for this.  On most systems dvisvgm is better.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -1198,7 +1198,7 @@ $pg{specialPGEnvironmentVars}{use_javascript_for_live3d} = 1;
 
 # Binary that the PGtikz.pl and PGlateximage.pl macros will use to create svg images.
 # This should be either 'pdf2svg' or 'dvisvgm'.
-$pg{specialPGEnvironmentVars}{latexImageSVGMethod} = "pdf2svg";
+$pg{specialPGEnvironmentVars}{latexImageSVGMethod} = "dvisvgm";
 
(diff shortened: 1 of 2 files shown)


somiaj:
Should probably mention `pgfsys-dvisvmg-bbox-fix.patch` as well in a comment. I also don't see mention of this in the 2.19 install instructions.

drgrice1:
I added the patch to the 2.19 instructions earlier (right before I submitted this pull request).  I also removed the other patch, since that is not needed for Ubuntu 24.04 anymore (which the instructions are for now).

somiaj:
@drgrice1 Ahh, missed you just changed that. Thanks.

Alex-Jordan:
Should this have also removed `pdf2svg` from `check_modules`?

drgrice1:
Previously `check_modules.pl` checked for both, so I didn't change that.  However, I agree that only `dvisvgm` should be checked for.  It would be nice to update `check_modules.pl` to actually check for the configured external program.  Of course there is the catch 22 of needing to have certain modules installed that we are checking for before the course environment can be loaded to read those configuration values.  But I was thinking we could have the `check_modules.pl` script work in phases.  It could first check for the base modules needed to get to the point that the course environment could be loaded.  If that fails it would exit with an additional message stating that not everything was checked due to previous failures, and that the script should be run again once those are fixed.  On success it would "runtime load" those modules and continue with the actual course environment values.

PR #2449, opened 2024-06-25
Update the docker build to Ubuntu 24.04.

drgrice1:
Note that the imagemagick-allow-pdf-read.patch is no longer needed. Ubuntu 24.04 allows PDF read rights by default.  The reason this restriction was previously applied was due to a security vulnerability in ghostscript.  That vulnerability has long been fixed (even for Ubuntu 22.04 although the policy was not updated).  The patch is left in the repository for Ubuntu 22.04 users.

Also, the Mojo::SQLite version downgrade is not needed anymore with the version of Mojo::SQLite in Ubuntu 24.04.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN echo Cloning branch $PG_BRANCH branch from $PG_GIT_URL \
 
 # We need to change FROM before setting the ENV variables.
 
-FROM ubuntu:22.04
+FROM ubuntu:24.04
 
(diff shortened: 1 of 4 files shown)


taniwallach (approved):
I tested this to the extent of creating new Docker images (manually setting it to use the 2.19 branches for now), and accessing the admin course. I also made sure that `bin/check_modules` does not report any problem.

Since the admin course comes up, and the modules all seem OK - I think this is good.

pstaabp (approved):

PR #2451, opened 2024-06-28
Add username and password option for SMTP authentication

dlglin:
This adds the ability to set a username and password for authentication with the SMTP server.

It also adds the ability to set `ssl_options` when creating the `Email::Sender::Transport::SMTP` object, which is necessary if the SMTP server uses a self-signed certificate.

I have tested this with our institutional SMTP server, and with smtp.gmail.com.  The only way I was able to authenticate to smtp.gmail.com was to create an app password for a gmail account that has 2FA enabled.

The gmail smtp server overwrites the "from" address in the message with the address of the user that authenticated.  My smtp server will only accept messages where the "from" address in the message matches the user that authenticated, so this PR won't be very useful without some implementation of setting "reply-to" addresses on messages, as discussed in #2383.  I can either add that to this PR or start a new PR for that.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -42,6 +42,16 @@ include("VERSION");  # get WW version
 # Mail Settings
 ################################################################################
 
+# $generic_sender_name will be used as the "From:" name on all feedback emails
+# sent without a defined user.
+$generic_sender_name = '';
+
+# The following variables will override the "From:" address in messages sent by
+# the named feature.
+$feedback_sender_email = ''; # For student feedback
+$instructor_sender_email = ''; # For instructors emailing students
+$jitar_sender_email = ''; # For notifications of incomplete JiTaR sets
+
 # By default, feedback is sent to all users who have permission to
 # receive_feedback in a course. If this list is non-empty, feedback is also sent
 # to the addresses specified here.
--- a/lib/Mojolicious/WeBWorK/Tasks/SendInstructorEmail.pm
+++ b/lib/Mojolicious/WeBWorK/Tasks/SendInstructorEmail.pm
@@ -79,17 +79,13 @@ sub mail_message_to_recipients ($job, $ce, $db, $mail_data) {
 			$mail_data->{merge_data}
 		);
 
-		my $email;
+		my $email = Email::Stuffer->to($user_record->email_address)->subject($mail_data->{subject})->text_body($msg)
+			->header('X-Remote-Host' => $mail_data->{remote_host});
 		if ($ce->{instructor_sender_email}) {
-			$email =
-				Email::Stuffer->to($user_record->email_address)
-				->from($mail_data->{from_name} . ' <' . $ce->{instructor_sender_email} . '>')
-				->reply_to($mail_data->{from})->subject($mail_data->{subject})->text_body($msg)
-				->header('X-Remote-Host' => $mail_data->{remote_host});
+			$email->from($mail_data->{from_name} . ' <' . $ce->{instructor_sender_email} . '>')
+				->reply_to($mail_data->{from});
 		} else {
-			$email =
-				Email::Stuffer->to($user_record->email_address)->from($mail_data->{from})
-				->subject($mail_data->{subject})->text_body($msg)->header('X-Remote-Host' => $mail_data->{remote_host});
+			$email->from($mail_data->{from});
 		}
 
 		eval {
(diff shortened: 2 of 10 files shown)


drgrice1:
I tested this with a gmail account, and I can see that this works.  I also see that the "from" address is filled in by the gmail smtp server with the email address  of the user that authenticated.  The gmail smtp server moves the original "from" address set by WeBWorK to the "X-Google-Original-From" field in the email source, but that is not shown anywhere in any email clients (including the gmail client in the browser).  So I think that adding a "reply-to" implementation is certainly needed with this.  If you already have that coded, then I think you should add it to this pull request.

dlglin:
This now implements the "reply-to" functionality.

There are four new variables added to the config:

- `$generic_sender_name` to set a name in a feedback email if no user is defined (not sure if this ever gets triggered)
- `$feedback_sender_email` for emails send using the "email instructor" feature
- `$instructor_sender_email` for emails sent from the Instructor Email page
- `$jitar_sender_email` for automated emails sent from jitar problems

If one of these is set, then it will be used as the "From:" address in any emails sent from that page.  The user's email will be set as the "Reply-to:" address for the email.

This also changes the Instructor email page so that instructors can no longer manually set the "From:" or "Reply-to:" address in messages that they send to students.  The "Reply-to:" was already ignored when the emails were sent, and now the "From:" address (or the "Reply-to" if `$instructor_sender_email` is set) is always set to the current user's name and email.

This changes the behaviour of instructor emails taken from a .msg file.  Previously the file would determine the "From:" and "Reply-to:" (though "Reply-to" didn't actually make it to the sent message).  Now those are not even read from the file.  The only option is for the message to come from the current user.

There are two other existing "From:" addresses that can be set for other features: `$twoFA{email_sender}` and `$mail{achievementEmailFrom}`.  Those should work the same as before.

drgrice1 (approved):
I think this is good.  Although there is one case that is a bit off to me.  That is the case of an instructor that does not have an email address in the database who attempts to send an email.  In that case the "From:" field is blank and can not be edited.  If the instructor clicks "Send Email" the page responds with "From field must contain one valid email address."  Yet, there is no way for the user to fix that.  Aside from adding an email address in the user settings, of course.

conf/defaults.config
```diff
@@ -42,6 +42,16 @@ include("VERSION");  # get WW version
 # Mail Settings
 ################################################################################
 
+# $generic_sender_name will be used as the "From:" name on all feedback emails
+# sent without a defined user.
+$generic_sender_name = '';
```
Alex-Jordan, line 47: Do we want this (and the three below) to be empty strings in `defaults.config`?

lib/Mojolicious/WeBWorK/Tasks/SendInstructorEmail.pm
```diff
@@ -79,9 +79,14 @@ sub mail_message_to_recipients ($job, $ce, $db, $mail_data) {
 			$mail_data->{merge_data}
 		);
 
-		my $email =
-			Email::Stuffer->to($user_record->email_address)->from($mail_data->{from})->subject($mail_data->{subject})
-			->text_body($msg)->header('X-Remote-Host' => $mail_data->{remote_host});
+		my $email = Email::Stuffer->to($user_record->email_address)->subject($mail_data->{subject})->text_body($msg)
+			->header('X-Remote-Host' => $mail_data->{remote_host});
+		if ($ce->{instructor_sender_email}) {
+			$email->from($mail_data->{from_name} . ' <' . $ce->{instructor_sender_email} . '>')
```
Alex-Jordan, line 85: I think this is from the diff I sent you. But I think this could be:
```suggestion
			$email->from("$mail_data->{from_name} <$ce->{instructor_sender_email}>")
```

although I am not sure which is more readable, since the `<` might appear to pair with the `>` in the `->`.

conf/defaults.config
drgrice1: @dlglin asked about this, and I told him to add these.  The default values should be represented in `defaults.config`, and in this case the defaults need to be not used.  Would it be better that they be `undef` or the empty string?

lib/WeBWorK/ContentGenerator/Feedback.pm
```diff
@@ -185,10 +185,14 @@ $emailableURL
 			$msg .= "***** Data about the environment: *****\n\n" . Dumper($ce) . "\n\n";
 		}
 
-		my $email =
-			Email::Stuffer->to(join(',', @recipients))->from($sender)->subject($subject)->text_body($msg)
+		my $from_name = $user ? $user->full_name : $ce->{generic_sender_name};
```
drgrice1, line 188: This line is not used outside the scope of the `if` statement below.  So it should be moved inside the `if` statement's block.

PR #2455, opened 2024-07-04
when grade passback only happens on mass update intervals, let studen…

Alex-Jordan:
…ts know

This is mainly so that when there is an LTI grade passback mode (`course or homework`) but `$LTIGradeOnSubmit` is false, there will be a message about how frequently grades are updated. 

We have `$LTIGradeOnSubmit` false because for some students, having this true leads to D2L pinging them through their phone every time they submit a correct answer. But then I have been getting many messages from students this term who think something is broken because their grade in D2L is not matching the WW grade. (They just haven't waited the hour that we have `$LTIMassUpdateInterval` set to.)

This message will only help so much, since many students won't read it. But it is something.

### DIFF
--- a/lib/WeBWorK/Utils/ProblemProcessing.pm
+++ b/lib/WeBWorK/Utils/ProblemProcessing.pm
@@ -248,25 +248,34 @@ async sub process_and_log_answer ($c) {
 					$c->param('startTime', '');
 				}
 
-				my $LTIGradeResult = -1;
-
-				# Try to update the student score on the LMS if that option is enabled.
-				if ($ce->{LTIGradeMode} && $ce->{LTIGradeOnSubmit}) {
-					$LTIGradeResult = 0;
-					my $grader = $ce->{LTI}{ $ce->{LTIVersion} }{grader}->new($c);
-					if ($ce->{LTIGradeMode} eq 'course') {
-						$LTIGradeResult = await $grader->submit_course_grade($problem->user_id);
-					} elsif ($ce->{LTIGradeMode} eq 'homework') {
-						$LTIGradeResult = await $grader->submit_set_grade($problem->user_id, $problem->set_id);
+				# Messages about passing the score back to the LMS
+				if ($ce->{LTIGradeMode}) {
+					my $LMSname        = $ce->{LTI}{ $ce->{LTIVersion} }{LMS_name};
+					my $LTIGradeResult = -1;
+					if ($ce->{LTIGradeOnSubmit}) {
+						$LTIGradeResult = 0;
+						my $grader = $ce->{LTI}{ $ce->{LTIVersion} }{grader}->new($c);
+						if ($ce->{LTIGradeMode} eq 'course') {
+							$LTIGradeResult = await $grader->submit_course_grade($problem->user_id);
+						} elsif ($ce->{LTIGradeMode} eq 'homework') {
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
It also changes the generic "the LMS" to the actual name of the LMS when that name is not the default `'e.g., Blackboard, Canvas, Moodle, etc.'`.

drgrice1 (commented):
This looks good except there is a translation issue.

lib/WeBWorK/Utils/ProblemProcessing.pm
```diff
@@ -248,26 +248,38 @@ async sub process_and_log_answer ($c) {
...
+								$c->tag('br') . $c->maketext('Your score was successfully sent to [_1].', $LMSname);
+						}
+					} elsif ($ce->{LTIMassUpdateInterval} > 0) {
+						my $massUpdateTime = "$ce->{LTIMassUpdateInterval} seconds";
+						$massUpdateTime = int($ce->{LTIMassUpdateInterval} / 60 + 0.99) . " minutes"
+							if ($ce->{LTIMassUpdateInterval} >= 120);
+						$massUpdateTime = int($ce->{LTIMassUpdateInterval} / 36000 + 0.9999) . " hours"
+							if ($ce->{LTIMassUpdateInterval} >= 7200);
+						$scoreRecordedMessage .= $c->tag('br')
+							. $c->maketext('Scores are sent to [_1] every [_2].', $LMSname, $massUpdateTime);
```
drgrice1, line 280: The second, minute, hour strings are not translated in this.  I suggest changing this to

```suggestion
						$scoreRecordedMessage .= $c->tag('br');
						if ($ce->{LTIMassUpdateInterval} < 120) {
							$scoreRecordedMessage .= $c->maketext('Scores are sent to [_1] every [quant,_2,second].',
								$LMSname, $ce->{LTIMassUpdateInterval});
						} elsif ($ce->{LTIMassUpdateInterval} < 7200) {
							$scoreRecordedMessage .= $c->maketext('Scores are sent to [_1] every [quant,_2,minute].',
								$LMSname, int($ce->{LTIMassUpdateInterval} / 60 + 0.99));
						} else {
							$scoreRecordedMessage .= $c->maketext('Scores are sent to [_1] every [quant,_2,hour].',
								$LMSname, int($ce->{LTIMassUpdateInterval} / 36000 + 0.9999));
						}
```

to fix this.

drgrice1: That was supposed to be with the last comment.

Alex-Jordan:
I don't love how this code here does a check against the string `'e.g., Blackboard, Canvas, Moodle, etc.'`.

One idea is to change the default to '`the LMS`', or `'the Learning Management System'`. I'm not sure if that's a good change for other places the name is used.

Another idea is to just allow the messages here so that they might come out like `Your score was not successfully sent to e.g., Blackboard, Canvas, Moodle, etc..` (Although, note the double period there.)

Are either of these better than leaving this how it is?

drgrice1:
I didn't notice the check against the default string that was added.  Yeah, I don't like that either.  I would be fine with changing the default to "the LMS".

Alex-Jordan:
I updated this to change the default LMS name the be "the LMS".

drgrice1 (approved):
Looks good now.

somiaj (approved):

PR #2457, opened 2024-07-09
add problem_grader permission

Alex-Jordan:
Currently the `score_sets` permission is doing too much. It is used for things that report scores (Scoring Tools, LTI Grade Update) and for things that let a user edit scores (Problem Grader). This PR creates `problem_grader` for the latter thing.

But there is an issue. In `lib/WebworkWebservice.pm`, the permission to use the `putUserProblem` command is `modify_student_data`. This command is needed for someone using the problem grader to actually save a score. I can't change the permission to use this command to the new `score_sets` permission because other things may need to use this command. And I can't scrap the new `score_sets` permission and just use `modify_student_data` for access to the grader, because you might want to grant a `ta` the permission to grade problems but not grant them the ability to modify other things about student data. I'm not sure what to do here.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -793,6 +793,7 @@ $authen{admin_module} = ['WeBWorK::Authen::Basic_TheLastOption'];
 	become_student                 => "professor",
 	access_instructor_tools        => "ta",
 	score_sets                     => "professor",
+	problem_grader                 => "professor",
 	send_mail                      => "professor",
 	receive_feedback               => ['ta', 'professor', 'admin'],
 
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -517,6 +517,15 @@ sub getConfigValues ($ce) {
 				type => 'permission'
 			},
 			{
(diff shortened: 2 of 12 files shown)


drgrice1:
One solution is to change the permission for the WebworkWebservice `putUserProblem`, `putProblemVersion`, and `putPastAnswer` methods to the new `problem_grader` permission.  Then in the methods check permissions again and only allow setting the `status` and `comment` for a user that does not have the `modify_student_data` permission.

I implemented the WebworkWebservice permission system when it was still using the XMLRPC approach.  With that approach the usual authentication and permission system wasn't available in the methods themselves.  It was only available outside of the XMLRPC region.  But now that that system is gone (good riddance), all of that can be done in the methods and finer grained permissions can be used.

drgrice1:
I added a pull request to this branch that does what I mentioned in my last comment.

Alex-Jordan:
I merged that PR, and this is ready for anyone else who would like to test.

drgrice1 (approved):
This looks good to me.  Although, some of the code was contributed by me.

pstaabp (approved):

PR #2458, opened 2024-07-10
Replace narrow no-break spaces in dates with regular spaces for PG.

drgrice1:
This is done for the dates explicitly passed to hardcopy in Hardcopy.pm as well as the dates passed to PG in general in Rendering.pm.

The DateTime::Locale package now uses a narrow no-break space in many of its en-US date time formats.  This fixes those characters breaking hardcopy without requiring a downgrade of the DateTime::Locale package.

A better solution will eventually be needed.  Perhaps now it is time to consider switching to XeTeX or LuaTeX which can handle UTF-8 characters, and droppping support for PDFLaTeX?

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Hardcopy.pm
+++ b/lib/WeBWorK/ContentGenerator/Hardcopy.pm
@@ -996,7 +996,7 @@ async sub write_set_tex ($c, $FH, $TargetUser, $themeTree, $setID) {
 		if ($MergedSet->{$_}) {
 			print $FH '\\def\\webwork'
 				. underscore_to_camel($_) . '{'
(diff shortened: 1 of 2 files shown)


Alex-Jordan (approved):
Looks good.

FWIW, in production we have had `$externalPrograms{pdflatex}` set to `xelatex` for a long time, and have not noticed any issues. Mainly I just like that we can put π in text with a simple option-p (on a MacOS keyboard) if places where math mode is overkill. I suspect that it also helps when students use special characters in answers for whatever reason, and the hardcopy prints their last answer.

somiaj (approved):

PR #2461, opened 2024-07-22
case insensitive sorting by name

Alex-Jordan:
Various things are currently sorted lexically with capitals before lowercase. This changes sorting to be case-insensitive. So for example in the file manager, what would currently be:

```
Banana.txt
apple.txt
```
would now be:
```
apple.txt
Banana.txt
```

If there is a tie, like `apple.txt` and `AppLE.txt` or `APPLE.TXT`, there may be inconsistent sorting. On a letter by letter basis, it would be easy to say make capitals come before lowercase. But then you end up with things like `Apricot` come before `apple`. I'm not sure how to modify this to put some consistent sorting when the words are a case-insensitive match. If that is a problem, this PR can be closed. But I thought it would be nice to have the File Manager (and other things) finally sort alphabetically instead of a character code based sorting.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -177,6 +177,8 @@ sub sortByName ($field, @items) {
 	}
 
 	my @sKeys = sort {
+		return $a cmp $b if (uc($a) eq uc($b));
+
 		my @aParts = split m/(?<=\D)(?=\d)|(?<=\d)(?=\D)/, $a;
 		my @bParts = split m/(?<=\D)(?=\d)|(?<=\d)(?=\D)/, $b;
 
@@ -195,8 +197,8 @@ sub sortByName ($field, @items) {
 				next if $aPart == $bPart;    # check next pair
 				return $aPart <=> $bPart;    # compare numerically
 			} else {
-				next if $aPart eq $bPart;    # check next pair
-				return $aPart cmp $bPart;    # compare lexicographically
+				next if uc($aPart) eq uc($bPart);    # check next pair
+				return uc($aPart) cmp uc($bPart);    # compare alphabetically
 			}
 		}
 		return +1 if @aParts;    # a has more sections, should go second


drgrice1 (approved):
I think this is fine.

Note that a case sensitive sort is consistent with how some operating systems do it (linux at least).  Of course Windows (and perhaps OSX) probably use a case insensitive sort.

Also note that the `sortByName` method is used many other places in the code (to sort sets on the problem sets page, most records shown in the scrolling record list, header files on the set detail page, etc).  So this is also changing the behavior for those sorts as well.  That is probably okay for those sorts, but worth noting.

Alex-Jordan:
I updated this to do something consistent for sorting `apple` versus `APPLE` etc. Can you confirm if the logic makes sense?

Alex-Jordan:
Well, the change I made to this will sort the following in this order:

```
ABC2
abc1
```

because the "parts" `ABC` and `abc` will be compared, and since they are equivalent without considering case, it will then compare them lexicographically before even considering the digits part that comes next.

So that's not ideal. I think that what is needed is to just initially compare the entire expressions without parsing into digit and non-digit parts. And if they fully match without considering case (like `apple1banana` and `Apple2Banana`) then do a simple lexicographical comparison. But otherwise do what I had before. What do you think?

drgrice1:
I forgot that this was a natural sort algorithm.  Achieving a semi case-insensitive natural sort seems to be challenging.  I think it will take two passes as you suggest.

Alex-Jordan:
This seems like something there would be an established perl library for, like https://metacpan.org/pod/Sort::Naturally. What do you think about I move to use that instead? (Or another option if that one has issues. I haven't investigated.)

Alex-Jordan:
I changed this back to how it was at first, but now also there is an initial check to see if the entire item is a case-insensitive match with the entire other item. I believe this makes a natural sort that sorts consistently even when the items are like `apple` and `ApPle`.

drgrice1:
Yeah, it seems that should work.

I am not opposed to using the `Sort::Naturally` package, if we want to go that way.

somiaj (approved):

PR #2462, opened 2024-07-24
Switch the default latex variant to xelatex.

drgrice1:
The course environment variable is now `$externalPrograms{latex2pdf}`. This was changed as discussed in the development meeting today because it is odd for it to be named pdflatex if we are not using pdflatex, and furthermore it forces those upgrading to check their configuration and make the change.

Note that a change to PG will also be needed due to the change in the environment variable name.  This is done in https://github.com/openwebwork/pg/pull/1087.

### DIFF
--- a/bin/check_latex
+++ b/bin/check_latex
@@ -46,20 +46,20 @@ my $temp_dir = eval { tempdir('check_latex_XXXXXXXX') };
 
 die $@ if $@;
 
-my $pdflatex_cmd =
+my $latex_cmd =
 	"cd $temp_dir && "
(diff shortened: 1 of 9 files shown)


Alex-Jordan:
Looks good and works.

There is a mention of `pdflatex` in `doc/devel/pre-bugzilla-TODO-file` which of course is very outdated. It reminds me that we may want to review the `doc` folder and remove or archive old irrelevant things.

Alex-Jordan (approved):

somiaj (approved):

PR #2465, opened 2024-07-27
don't show achievement points when there aren't any

Alex-Jordan:
When you earn an achievement, the notification message tells you that you earned so many points along with it. But the achievement might be worth 0 points, or not even have a point value entered. This change will suppress the point display in those situations.


drgrice1 (approved):
Looks good.

somiaj (approved):

PR #2467, opened 2024-07-28
adjustments to date extension achievement items

Alex-Jordan:
This addresses some issues I've had in practice with these achievement items.

* A student is aware they have an item that can extend a due date 24 (or 48) hours. They have a set that closes at, say, 10:00pm. Their expectation is often that they can let the due date pass, then use the extension item. (They are not thinking to abuse the system by viewing answers; they are just thinking it shouldn't matter.) But that is not how it currently works. So the first change here is that if it is within the 24 (or 48) hour window after a close date, you can still use the reward item. [The only reason not to allow them to use it indefinitely is to save them from wasting it.]
* This then raises the issue of a set closing, and possibly the answer date passes too, and the student can see all the answers. So if they use the reward item in that 24 (or 48) hour window, the set's seeds are all re-randomized. (No attempt is made to leave the seeds alone where the student already had status 1.)
* Currently, these reward items always push back the answer date along with the close date. But that can leave them in a position where the set has closed for them (as well as for all other students), and other students have access to answers, but this student does not. So the last change here is to only push the answer date back to whatever the new close date is, or leave it alone, whichever is later.

And then there is `ExtendDueDateGW`.  This is not used by the default set of achievements, so I wonder if anyone even uses it. If anyone does use it, they are likely power users with extensive understanding of achievements. I started editing this one to make the same changes, but this reward item makes little sense to me in the first place. It adds 24 hours to a test template set, which seems perfectly reasonable to me. But it also adds 24 hours to every test version. So a student who activated a test that was only supposed to be 1 hour long could use this and have the test open for 25 hours. And it seems like that is not the best idea. Along with extending the template set by 24 hours, maybe just granting the student an additional test version would be good? Anyway, aside from introducing the constant for 24 hours in seconds, I did not change that one.

### DIFF
--- a/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
@@ -79,7 +79,7 @@ sub use_item ($self, $userName, $c) {
 		my @probIDs = $db->listUserProblems($userName, $setID);
 		for my $probID (@probIDs) {
 			my $problem = $db->getUserProblem($userName, $setID, $probID);
-			$problem->problem_seed($problem->problem_seed + 100);
+			$problem->problem_seed($problem->problem_seed % 2**31 + 1);
 			$db->putUserProblem($problem);
 		}
 	}
--- a/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
@@ -28,7 +28,10 @@ sub new ($class) {
 	return bless {
 		id          => 'ExtendDueDate',
 		name        => x('Tunic of Extension'),
-		description => x('Adds 24 hours to the close date of a homework.')
+		description => x(
+			'Adds 24 hours to the close date of a homework. '
+				. 'This will randomize problem details if used after the original close date.'
+		)
 	}, $class;
 }
 
--- a/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
@@ -79,9 +79,7 @@ sub use_item ($self, $userName, $c) {
 
 	# Change the seed for all of the problems if the set is currently closed.
 	if (after($set->due_date)) {
-		my @probIDs = $db->listUserProblems($userName, $setID);
-		for my $probID (@probIDs) {
-			my $problem = $db->getUserProblem($userName, $setID, $probID);
+		for my $problem ($db->getUserProblemsWhere({ user_id => $userName, set_id => $setID })) {
 			$problem->problem_seed($problem->problem_seed % 2**31 + 1);
 			$db->putUserProblem($problem);
 		}
--- a/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
@@ -19,7 +19,7 @@ use Mojo::Base 'WeBWorK::AchievementItems', -signatures;
 # Item to extend a close date by 24 hours.
 
 use WeBWorK::Utils           qw(x nfreeze_base64 thaw_base64);
-use WeBWorK::Utils::DateTime qw(between);
+use WeBWorK::Utils::DateTime qw(after between);
 use WeBWorK::Utils::Sets     qw(format_set_name_display);
 
 use constant ONE_DAY => 86400;
--- a/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
@@ -37,7 +37,7 @@ sub print_form ($self, $sets, $setProblemIds, $c) {
 
 	for my $i (0 .. $#$sets) {
 		push(@openSets, [ format_set_name_display($sets->[$i]->set_id) => $sets->[$i]->set_id ])
-			if (between($sets->[$i]->open_date, $sets->[$i]->due_date + ONE_DAY())
+			if (between($sets->[$i]->open_date, $sets->[$i]->due_date + ONE_DAY)
 				&& $sets->[$i]->assignment_type eq 'default');
 	}
 
@@ -85,11 +85,11 @@ sub use_item ($self, $userName, $c) {
 	}
 
 	# Add time to the reduced scoring date if it was defined in the first place
-	$userSet->reduced_scoring_date($set->reduced_scoring_date + ONE_DAY()) if $set->reduced_scoring_date;
+	$userSet->reduced_scoring_date($set->reduced_scoring_date + ONE_DAY) if $set->reduced_scoring_date;
 	# Add time to the close date
-	$userSet->due_date($set->due_date + ONE_DAY());
+	$userSet->due_date($set->due_date + ONE_DAY);
 	# This may require also extending the answer date.
-	$userSet->answer_date($userSet->due_date) if ($userSet->due_date > $set->answer_date);
+	$userSet->answer_date($userSet->due_date) if $userSet->due_date > $set->answer_date;
 	$db->putUserSet($userSet);
 
 	$globalData->{ $self->{id} }--;
--- a/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
@@ -22,6 +22,8 @@ use WeBWorK::Utils           qw(x nfreeze_base64 thaw_base64);
 use WeBWorK::Utils::DateTime qw(between);
 use WeBWorK::Utils::Sets     qw(format_set_name_display);
 
+use constant ONE_DAY => 86400;
+
 sub new ($class) {
 	return bless {
 		id          => 'ExtendDueDate',
@@ -35,7 +37,8 @@ sub print_form ($self, $sets, $setProblemIds, $c) {
 
 	for my $i (0 .. $#$sets) {
 		push(@openSets, [ format_set_name_display($sets->[$i]->set_id) => $sets->[$i]->set_id ])
-			if (between($sets->[$i]->open_date, $sets->[$i]->due_date) && $sets->[$i]->assignment_type eq 'default');
+			if (between($sets->[$i]->open_date, $sets->[$i]->due_date + ONE_DAY())
+				&& $sets->[$i]->assignment_type eq 'default');
 	}
 
 	return unless @openSets;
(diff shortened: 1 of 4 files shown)


Alex-Jordan:
I noticed I'd overlooked `ReducedCred.pm`. It's not in the default collection that gets used. I made similar updates to that one.

drgrice1:
@Alex-Jordan:  You will need to retarget this for develop.

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -35,7 +37,8 @@ sub print_form ($self, $sets, $setProblemIds, $c) {
 
 	for my $i (0 .. $#$sets) {
 		push(@openSets, [ format_set_name_display($sets->[$i]->set_id) => $sets->[$i]->set_id ])
-			if (between($sets->[$i]->open_date, $sets->[$i]->due_date) && $sets->[$i]->assignment_type eq 'default');
+			if (between($sets->[$i]->open_date, $sets->[$i]->due_date + ONE_DAY())
+				&& $sets->[$i]->assignment_type eq 'default');
```
somiaj, line 41: Using `ONE_DAY()` for a constant vs just `ONE_DAY` seems odd to me, I think both work, I just find it odd to think of a constant as a function. Unsure what is preferred in perl code. The examples I saw on perldoc don't seem to include the parenthesis.

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -71,10 +74,22 @@ sub use_item ($self, $userName, $c) {
...
-	# Add time to the reduced scoring date, due date, and answer date.
-	$userSet->reduced_scoring_date($set->reduced_scoring_date() + 86400) if $set->reduced_scoring_date;
-	$userSet->due_date($set->due_date() + 86400);
-	$userSet->answer_date($set->answer_date() + 86400);
+	# Change the seed for all of the problems if the set is currently closed.
+	if (after($set->due_date)) {
+		my @probIDs = $db->listUserProblems($userName, $setID);
+		for my $probID (@probIDs) {
+			my $problem = $db->getUserProblem($userName, $setID, $probID);
+			$problem->problem_seed($problem->problem_seed + 100);
```
somiaj, line 82: Is it possible this causes an overflow issue if `$problem->problem_seed` is already close to the max value for an `INT`, or does the database correctly deal with that already?

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -71,10 +74,22 @@ sub use_item ($self, $userName, $c) {
...
+			$db->putUserProblem($problem);
+		}
+	}
+
+	# Add time to the reduced scoring date if it was defined in the first place
+	$userSet->reduced_scoring_date($set->reduced_scoring_date + ONE_DAY()) if $set->reduced_scoring_date;
+	# Add time to the close date
+	$userSet->due_date($set->due_date + ONE_DAY());
+	# This may require also extending the answer date.
+	$userSet->answer_date($userSet->due_date) if ($userSet->due_date > $set->answer_date);
```
somiaj, line 92: I don't think the parenthesis are needed here, just `if $userSet->due_date > $set->answer_date;` There are a few other places I think the parenthesis could be removed in below too.

somiaj (commented):
Looking at the code I noticed a few things that I think could be cleaned up.

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
somiaj: Thinking about this a bit more, I don't think overflow is an issue here as I don't think the seeds are ever large enough to be close to the max value of an `INT`.

Alex-Jordan: I think I went with adding 100 for no particular reason. I was thinking it helps to have something from which a prior seed can be determined, rather than actually rerandomizing the seed. But adding 1 would work just as well. Should I change it to that? I could first mod the current seed by 2^31 to make sure the result is less than 2^32-1, which I guess is the max int size?

Alex-Jordan: I changed it to `$problem->problem_seed % 2**31 + 1`

drgrice1: The current code for randomizing a problem in default sets when problem randomization is enabled uses the formula `($problem->{problem_seed} + $problem->num_correct + $problem->num_incorrect) % 10000`.  It might be a good idea to use a modulus like that just in case.  I don't think there is a need to use such a large number as 2^31.

Alex-Jordan: Why add `$problem->num_correct + $problem->num_incorrect`? That could just be zero right? Here the concern is that maybe answers have been visible. And there may have been zero attempts.

5000 is the modulus used for initialization of a new assignment of a set to a user. Should we maybe pick a standard modulus to use in all these places?

somiaj: I think that is part of re-randomization after a certain number of attempts, so that won't be zero.

The randomize seed in the problem editor goes to 10,000. Wonder if it would be worth having a configuration value for this just to try to standardize it.

drgrice1: The `$problem->num_correct + $problem->num_incorrect` is in fact guaranteed to be a non-zero multiple of `$problem->{prPeriod}` at the time that it is called.

We could switch to something like 2^31 in all of the places that we do this randomization.  I do not think it is worth having a configuration value for this.  We could however, define a constant somewhere for this.  We need a better place for global constants.  Many of the things already in the course environment should not be, and should be global constants (or something filled in to the course environment after the configuration files are executed).

lib/WeBWorK/AchievementItems/ReducedCred.pm
```diff
@@ -79,12 +82,24 @@ sub use_item ($self, $userName, $c) {
...
 	return "Couldn't find that set!" unless $set && $userSet;
 
-	# Enable reduced scoring on the set and add the reduced scoring period to the due date.
-	my $additionalTime = 60 * $ce->{pg}{ansEvalDefaults}{reducedScoringPeriod};
+	# Change the seed for all of the problems if the set is currently closed.
+	if (after($set->due_date)) {
+		my @probIDs = $db->listUserProblems($userName, $setID);
+		for my $probID (@probIDs) {
+			my $problem = $db->getUserProblem($userName, $setID, $probID);
+			$problem->problem_seed($problem->problem_seed % 2**31 + 1);
```
somiaj, line 90: Shouldn't the modulus happen after adding 1, not before? Also after some quick searches, it seems that perl might just switch to floating point once it hits the max size of an int vs overflowing. Though I doubt the seed will ever get that high if it starts out < 10,000.

somiaj: Oh wait, I missed your comment you are trying to keep the number less than 2^32, so that is why you mod before.

drgrice1 (approved):
Looks okay to me.

Alex-Jordan:
I fixed the conflicts and added to the descriptions.

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -71,10 +77,22 @@ sub use_item ($self, $userName, $c) {
 	my $userSet = $db->getUserSet($userName, $setID);
 	return q{Couldn't find that set!} unless $set && $userSet;
 
-	# Add time to the reduced scoring date, due date, and answer date.
-	$userSet->reduced_scoring_date($set->reduced_scoring_date() + 86400) if $set->reduced_scoring_date;
-	$userSet->due_date($set->due_date() + 86400);
-	$userSet->answer_date($set->answer_date() + 86400);
+	# Change the seed for all of the problems if the set is currently closed.
+	if (after($set->due_date)) {
```
drgrice1, line 81: The `after` method needs to be added to those imported from `WeBWorK::Utils::DateTime` in this file.

lib/WeBWorK/AchievementItems/ReducedCred.pm
```diff
@@ -79,12 +83,24 @@ sub use_item ($self, $userName, $c) {
 	my $userSet = $db->getUserSet($userName, $setID);
 	return "Couldn't find that set!" unless $set && $userSet;
 
-	# Enable reduced scoring on the set and add the reduced scoring period to the due date.
-	my $additionalTime = 60 * $ce->{pg}{ansEvalDefaults}{reducedScoringPeriod};
+	# Change the seed for all of the problems if the set is currently closed.
+	if (after($set->due_date)) {
```
drgrice1, line 87: The `after` method also needs to be added to those imported from `WeBWorK::Utils::DateTime` in this file.

lib/WeBWorK/AchievementItems/SuperExtendDueDate.pm
```diff
@@ -71,10 +77,22 @@ sub use_item ($self, $userName, $c) {
 	my $userSet = $db->getUserSet($userName, $setID);
 	return q{Couldn't find that set!} unless $set && $userSet;
 
-	# Add time to the reduced scoring date, due date, and answer date.
-	$userSet->reduced_scoring_date($set->reduced_scoring_date() + 172800) if $set->reduced_scoring_date;
-	$userSet->due_date($set->due_date() + 172800);
-	$userSet->answer_date($set->answer_date() + 172800);
+	# Change the seed for all of the problems if the set is currently closed.
+	if (after($set->due_date)) {
```
drgrice1, line 81: The `after` method also needs to be added to those imported from `WeBWorK::Utils::DateTime` in this file.

drgrice1 (commented):
I was testing using these achievement rewards with this pull request and got exceptions because the `after` method is not imported.

Alex-Jordan:
Oh, I wonder if it was there before, but I resolved those `qw` conflicts without enough care.

Alex-Jordan:
OK, `after` is imported in those three places now.

drgrice1:
Yeah, I think it was.  I though this worked when I tested it before.

drgrice1:
I noticed another issue when testing the `ReducedCred` reward (the Ring of Reduction).  It is not due to this pull request, and doesn't need to be fixed here, but I am just noting it.  I have reduced scoring enabled for the course, but I tested using it on a set that does not have reduced scoring enabled and received the message "This item won't work unless your instructor enables the reduced scoring feature. Let your instructor know that you received this message."  It seems that the form should only present sets that have reduced scoring enabled.

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -71,10 +77,22 @@ sub use_item ($self, $userName, $c) {
...
+		my @probIDs = $db->listUserProblems($userName, $setID);
+		for my $probID (@probIDs) {
+			my $problem = $db->getUserProblem($userName, $setID, $probID);
```
drgrice1, line 84: This could be done more efficiently with
```suggestion
		for my $problem ($db->getUserProblemsWhere({ user_id => $userName, set_id => $setID })) {
			say $problem->problem_id . ' old seed: ' . $problem->problem_seed;
```
Usually, the procedure of listing ids for a database object, and then getting them in a loop is one of the inefficiencies that I have been working to remove.

Alex-Jordan: OK, I committed your suggestion directly here.

drgrice1: Note that the same thing should be done in the `ReducedCred.pm` file at lines 88-90 and the `SuperExtendDueDate.pm` file at lines 82-84.

Alex-Jordan: Ah, sorry. That is done now.

lib/WeBWorK/AchievementItems/ReducedCred.pm
```diff
@@ -79,12 +83,23 @@ sub use_item ($self, $userName, $c) {
...
+		for my $problem ($db->getUserProblemsWhere({ user_id => $userName, set_id => $setID })) {
+			my $problem = $db->getUserProblem($userName, $setID, $probID);
```
drgrice1, line 89: It looks like you didn't quite go far enough here.  Line 89 should be deleted.

lib/WeBWorK/AchievementItems/SuperExtendDueDate.pm
```diff
@@ -71,10 +77,21 @@ sub use_item ($self, $userName, $c) {
...
+		for my $problem ($db->getUserProblemsWhere({ user_id => $userName, set_id => $setID })) {
+			my $problem = $db->getUserProblem($userName, $setID, $probID);
```
drgrice1, line 83: Also here line 83 should be deleted.

lib/WeBWorK/AchievementItems/ReducedCred.pm
Alex-Jordan: Got these too now.

drgrice1:
Looks good now.  Thanks for making those changes.

somiaj (approved):

drgrice1:
I will merge this with two approvals.

PR #2468, opened 2024-07-28
achievement to grant extensions

Alex-Jordan:
This is a bit of a hack, but I thought I'd put it out here to see if it's worth contributing to the distribution. We use something like this at PCC.

The goal is to give students some sort of ability to grant their own extensions. Either to the whole class, or just to students with certain accommodations about due date flexibility. So what you can do is import the `axp` file here. And then the first time a student answers a question correctly, they will be awarded ten 24-hour extensions, five 48-hour extensions, and three 24-hour set resurrections. (The 10, 5, and 3, can be edited in the `at` file.)

The image here is public domain. I just looked for something random from https://freesvg.org.


somiaj (approved):

drgrice1 (approved):
I am fine with this.

PR #2469, opened 2024-07-28
let certain achievements apply to closed sets

Alex-Jordan:
The achievements affected by this PR currently only apply to open sets. But in these cases, to me it seems that students should be allowed to use them after the close date. So this would change it so that these can be used on sets after their close date. 

This may be controversial if people like these the way they are. But I find that students (a) get anxious about if they should use these when they don't know ahead of time what future sets will be like and (b) other students aren't aware they have these rewards until it's too late to use them.


somiaj (approved):

drgrice1 (approved):
I think this is fine.

PR #2483, opened 2024-07-30
Update the docker base image name.

drgrice1:
Forgotten detail.


drgrice1:
This is an alternate to #2449 if it is not merged.  One of these should be in the release.

drgrice1:
Closing this.  We will go with #2449.

PR #2486, opened 2024-08-04
also clear tmpEdit folder when 'cleaning' a newly unarchived course

Alex-Jordan:
This should have been included with #2291, which lets you clear things from a course when you unarchive it. This is useful for when you are creating a new course by unarchiving, and you are really just using the archived course as a template.

So this PR also clears the tmpEdit folder when you do that.


somiaj (approved):
Sounds reasonable to me.

Just a thought, the tmpEditFileDir also contains problem backups. Would it be useful to have removing this be its own option that can be selected, vs combining this with the clean course as it is done here? I'm fine with either solution, just mentioning this possibility.

Alex-Jordan:
Since the course archive file is retained, nothing is lost when these
things are deleted. And the use case is you are using an archived course as
a template toake a new course. So I think I'd just keep it simple and not
separate this.

Issue #2487, opened 2024-08-05
error in Scoring Tools


Alex-Jordan:
I am getting an error on 2.19 if I try to visit Scoring Tools:

```
Too few arguments for subroutine 'WeBWorK::ContentGenerator::Instructor::FileManager::checkName' at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/Scoring.pm line 51. 
```

Alex-Jordan:
`checkName` is expecting two arguments, the controller and a file name to check. But I tried using it as a method on the controller and that leads to warnings that suggest `checkName` is not functioning:

```
Your file name is not valid! A file name cannot begin with a dot, it cannot be empty, it cannot contain a directory path component and only the characters -_.a-zA-Z0-9 and space are allowed.
Warning: There may be something wrong with this question. Please inform your instructor including the warning messages below.
```

pstaabp:
Fixed in #2488

PR #2488, opened 2024-08-05
update usage of checkName

Alex-Jordan:
This is meant to address #2487.


drgrice1 (approved):
Good catch.  This fix will work for now.

However, the real problem is that ScoringDownload.pm and Scoring.pm are improperly using a method from another package that is not exported.  That is what caused this issue to begin with.  It is unexpected when working with the FileManager.pm module to need to be concerned about usage in other files like this.  That shouldn't be done.  Of course this is something that has been there for a long time.

somiaj (approved):
Double checked it fixes the issue with scoring tools.

PR #2490, opened 2024-08-06
Don't require $authen{admin_module} is a subset of $authen{user_module}.

somiaj:
This changes the logic of how `$authen{admin_module}` is used to restrict modules for the admin course. Instead of looping over all user_modules, and requiring that module also be listed in admin_modules, only loop over the modules instead in admin_modules when logging into the admin course.

This fixes an issue with LDAP (and maybe other authen modules), where if they are listed in the user_module list, they have to also be listed in the admin_module list, otherwise the error message generated for this causes the authentication to fail.

The issue that this fixes is if `WeBWorK::Authen::LDAP` is listed in `$authen{user_module}`, then it must also be listed in `$authen{admin_module}` or the check for that module fails causing an error to propagate to the other modules. The cause of the error is lines 157-169 in `Authen.pm`:

```perl
	return $self->call_next_authen_method if !$self->request_has_data_for_this_verification_module;
	my $authen_ref = ref($self);
	if ($c->ce->{courseName} eq $c->ce->{admin_course_id}
		&& !(grep {/^$authen_ref$/} @{ $c->ce->{authen}{admin_module} }))
	{
		$self->write_log_entry("Cannot authenticate into admin course using $authen_ref.");
		$c->stash(
			authen_error => $c->maketext(
				'There was an error during the login process.  Please speak to your instructor or system administrator.'
			)
		);
		return $self->call_next_authen_method();
	}
```

What is happening is LDAP doesn't contain request information so the first test fails since there is no request data. Then the test to see if this is a valid admin module happens, which also fails, and an `authen_error` is added to the stash. This error persists even if the next authen method succeeds.  Removing placing the `authen_error` message in the stash also fixes the LDAP issue described above, but then there could be silent failures logging into the admin course.

I chose to not require that the admin_module is a subset of the user_modules, as I have a use case in which this is not the case. If it would be preferred to not change this requirement, another fix of not issuing an error will fix the LDAP issue.


somiaj:
It occurred to me my understanding of the logic above is probably slightly flawed, but the issue is the `authen_error` message either needs to not be set or cleared with future modules.

Alex-Jordan:
I'm not following everything. In `lib/WeBWorK/Authen.pm`, there is a stub for `request_has_data_for_this_verification_module` and it always returns 1. So wouldn't that be used when authenticating with LDAP? So line 157 would not execute?

somiaj:
@Alex-Jordan I didn't notice that, and it appears that only the LTI authen methods redefine that method. But my testing is showing that with the LDAP module, that test fails. In my admin course login.log file I see the following line:

```
Cannot authenticate into admin course using WeBWorK::Authen::LDAP. user_id= login_type= credential_source= host=<some ip> port=35746 UA=Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0
```

This is being logged by the line `$self->write_log_entry("Cannot authenticate into admin course using $authen_ref.");`.

I can try to debug this further.

somiaj:
Oh I see, the value always returns `1`, but then that test negates the value to be `0`, which then fails, so the return doesn't happen. I think this line is only to move to the next auth module when working with the LTI request headers, but other modules, like LDAP, just ignore this test.

Alex-Jordan:
I still don't understand this, but if something is being changed with this where the one thing is not required to be a subset of the other thing, the commentary about this in `defaults.config` needs to change.

To make this concrete to help me understand it better, which authentication modules are you wanting to use for general courses? And which ones for the admin course?

somiaj:
My use case is I want to disable local passwords with `Basic_TheLastOption` for my courses and only use LTI or LDAP. But since my admin user is just a local user, I want to enable `Basic_TheLastOption` only for the admin course with two factor auth setup (no need to use LDAP here since it will just be a failed call to the LDAP server for a local user).

The issue that arises currently is when logging into the admin course, most modules in `$authen{user_module}` that aren't also in `$authen{admin_module}`, get logged and `authen_error` gets defined before moving on to the next authen module in `$authen{user_module}`, which can cause the login page to error out by just going to it (before any login happens) or when trying to login.

This issue needs to be fixed, I can make another PR fixing it without changing current sub list behavior if preferred.

I'll update `defaults.conf` comments.

Alex-Jordan:
Everything that is currently happening is by design, under the assumption that the `admin_module` options should be a subset of the `user_module` options. That's why all the comments say the `admin_module` options should be a subset of the `user_module` options. It's supposed to fail if the `admin_module` options are not a subset of the `user_module` options.

I was confused because you are describing the issue as if something unexpected happens, but I think it is just expected with the current code, when the `admin_module` options are not a subset of the the `user_module` options.

But you would like to change things so that the `admin_module` options and `user_module` options are managed independently. That sounds like it would be OK. Scanning for `user_module` and `admin_module` elsewhere in the code, there is one use of `user_module` in line 156 of `lib/WeBWorK.pm`. In case the `admin_module` options are not a subset of the `user_module` options, are there any implications for that line of code too?

somiaj:
@Alex-Jordan there is currently a bug with LDAP and maybe other modules.

I have the following set:

```
$authen{user_module} = [
        { '*' => 'WeBWorK::Authen::LTIAdvanced' },
        { '*' => 'WeBWorK::Authen::LDAP' },
        { '*' => 'WeBWorK::Authen::Basic_TheLastOption'},
];
$authen{admin_module} = [
        'WeBWorK::Authen::Basic_TheLastOption'
];
```

This fails, and I cannot log into the admin course.

That is because when looping through `$authen{user_module}`, it gets to the LDAP module before the basic module, it then tests that the LDAP module is not in `$authen{admin_module}`, and both logs this attempt and sets `authen_error`, which prevents logging into the admin course.

I just would prefer more flexibility on setting which admin_modules are used, so I'm fixing this by changing behavior. I am making a new PR that fixes the issue by removing the error messages.

somiaj:
Note, @pstaabp also ran into a similar issue, and brought it up, it reminded me that I too had to work around this issue, but a fix is probably best at least to get the situation I described working.

somiaj:
Closing in favor of #2493

PR #2491, opened 2024-08-06
Allow instructors with no email saved to enter an email address.

drgrice1:
If an instructor does not have an email address in the database, then currently the from field on the instructor "Email" page is disabled, and there is no way to enter an email address.  Attempting to send an email address gives the message "From field must contain one valid email address", and yet there is still no way to enter an email address.

This makes it so that in the case that an instructor does not have an email address in the database the from field is not disabled, and the instructor can enter an email address.

The message was changed because I find it a bit confusing even if the instructor can change the email address, if the message says the "From field must contain one valid email address".  It seems to me that means that it would be valid to enter more than one email address as long as one is valid.  The message now reads "From field must contain a single valid email address."

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -185,10 +185,10 @@ $emailableURL
 			$msg .= "***** Data about the environment: *****\n\n" . Dumper($ce) . "\n\n";
 		}
 
-		my $from_name = $user ? $user->full_name : $ce->{generic_sender_name};
-		my $email     = Email::Stuffer->to(join(',', @recipients))->subject($subject)->text_body($msg)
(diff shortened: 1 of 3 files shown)


somiaj (approved):
This should also have a PR targeted against develop.

Alex-Jordan:
I have users with empty string email addresses, and on `main`, the input field on this page is not disabled. Do I need to test this by directly nullifying an email address field in the database?

somiaj:
@Alex-Jordan My testing on main has the from address disabled. Are you on the email page or sending a feedback email? The sending the instructor an email I can edit the from address, but the email page for sending mass emails I cannot.

This branch does allow me to edit that email address as stated.

somiaj (approved):

Alex-Jordan:
I see, yes, I was thinking this was about Email Instructor. Will test now.

Alex-Jordan (approved):
Still needs a third approval.

pstaabp (approved):

PR #2492, opened 2024-08-07
Don't log out items in $authen{user_module} not in $authen{admin_module}

somiaj:
When looping through $authen{user_module} for the admin course, modules not in $authen{admin_module} can be logged and `authen_error` is set before attempting the next module. This can cause the logins to the admin course to fail under certain configurations.

This removes both the logging and error message, and just attempts the next authentication method.

This is an alternative to fix the issue described in #2490.

One side affect of this is logging into the admin course if `$authen{admin_module}` does not contain any elements of `$authen{user_module}`, the login will silently fail. I just couldn't figure out any logic when looping through the login methods, that `$authen{admin_module}`, to test if no matches were found and creating an error message as a result.

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -154,17 +154,13 @@ sub verify {
 
 	debug('BEGIN VERIFY');
 
-	return $self->call_next_authen_method if !$self->request_has_data_for_this_verification_module;
 	my $authen_ref = ref($self);
-	if ($c->ce->{courseName} eq $c->ce->{admin_course_id}
-		&& !(grep {/^$authen_ref$/} @{ $c->ce->{authen}{admin_module} }))
+	if (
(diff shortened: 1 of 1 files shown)


drgrice1:
I am looking into setting up a local LDAP server to figure out the correct way to fix this.  I don't think that either #2490 nor this pull request are the correct way to fix this.  The problem is that the LDAP.pm authentication module is not correctly designed to fit into the webwork2 authentication system (much like most of the other remote authentication modules other than the LTI modules). 

If you have the LDAP authentication module in your user_module list, then there is no reason to ever have anything after it in the list including Basic_TheLastOption because nothing after it will ever get called.  Also, the LDAP authentication module is the basic authentication module except on initial login when the password is verified.  In that case it checks the password with the LDAP server, and then falls back to using basic password authentication from the base Authen.pm module.

somiaj:
Closing in favor of #2493

PR #2493, opened 2024-08-07
Make the `admin_module` separate from the `user_module` for authentication.

drgrice1:
Make the `admin_module` separate from the `user_module` for authentication.

When signing in to the admin course the course `$authen{admin_module}` is used, and for all other courses the `$authen{user_module}` is used. So the two lists are completely separate.  Neither needs to be a subset of the other.

Also fix the `net_ldap_options` course environment variable usage.  It is `net_ldap_options` in the `authen_ldap.conf` file, but was `net_ldap_opts` in the code.  Furthermore, the variable was dereferenced as an array, but is actually a hash which causes an error.

Note that this is built on top of #2490.  @somiaj I hope you don't mind me taking this over.  This is also rebased onto main and put in as a hotfix.

Note that the `$authen{user_module}` that is demonstrated in `localOverrides.conf.dist` is nonsensical.  The `LDAP.pm` authentication module is terminal.  In other words, if it is encountered in the list, it never allows anything after it to be used.  In fact the only authentication modules that actually allow later elements in the list to be used are the LTI authentication modules.  All others are terminal.   I removed the `Moodle` authentication module from the demonstrated list because that module should be considered deprecated (and I am absolutely certain does not work anymore), and removed the database config name (which also needs to be dropped) for now.  This needs to be reworked quite a bit.

### DIFF
--- a/conf/authen_CAS.conf.dist
+++ b/conf/authen_CAS.conf.dist
@@ -13,7 +13,7 @@ $authen{user_module} = {
 };
 
 # List of authentication modules that may be used to enter the admin course.
-# This should be a non-empty sublist of whatever is in $authen{user_module}.
+# This is used instead of $authen{user_module} when logging into the admin course.
 # Since the admin course provides overall power to add/delete courses, access
 # to this course should be protected by the best possible authentication you
 # have available to you.
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -46,19 +46,10 @@ $authen{user_module} = [
 ];
 
 # List of authentication modules that may be used to enter the admin course.
-# This should be a non-empty sublist of whatever is in $authen{user_module}.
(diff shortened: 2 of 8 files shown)


somiaj:
Tested this works, and I'll close my two PRs in favor of this.

I did run across an issue testing this, but unrelated to this PR. I'll open a new issue for that.

pstaabp (approved):
This also works for me testing on our server with ldap enabled.   Since @somiaj  and I may be the only testers with LDAP, I'll merge this.

pstaabp:
Except that @somiaj only has a shadow approval.

drgrice1:
Since it is targeted for main, one more approval is needed.  Do you have the checkbox option to bypass branch protections?  If you check that, then you can merge it.  I guess I can do it if needed.

Alex-Jordan (approved):

Issue #2494, opened 2024-08-07
Adding a new user without a password breaks two factor auth with LDAP.


somiaj:
Using the accounts manager to add a new user with a blank password breaks two factor authentication with LDAP. This is because if no password is supplied a row in the table `COURSEID_password` is not created, so if the user tries to login via LDAP an error happens because the row doesn't exist.

```
Can't call method "otp_secret" on an undefined value at /home/webwork/webwork2.19/lib/WeBWorK/ContentGenerator/TwoFactorAuthentication.pm line 51. 

46: 
47: 	# Note that this user has already authenticated with username and password,
48: 	# so this and the $user below should exist.
49: 	my $password = $c->db->getPassword($c->authen->{user_id});
50: 
51: 	if (!$password->otp_secret) {
52: 		my $totp =
53: 			WeBWorK::Utils::TOTP->new(
54: 				$c->authen->session->{otp_secret} ? (secret => $c->authen->session->{otp_secret}) : ());
55: 		$c->authen->session(otp_secret => $totp->secret);
56: 
```

somiaj:
Fixed in #2495.

PR #2495, opened 2024-08-08
Handle user's with no password record for two factor authentication.

drgrice1:
It makes sense for a system configured for LDAP authentication to also want to use two factor authentication, and a user does not need to have a local password defined for LDAP authentcation.  However, this is not something that I realized when I implemented two factor authentication. So I assumed a password record would exist.

This fixes the OTP verification for the case that a password does not exist in the database for a user.  I.e., this fixes issue #2494.

Should this be a hotfix?


somiaj (approved):
Tested that this fixes the issue.

I could use this hot fixed so I can use two factor with LDAP.

Alex-Jordan (approved):
Approving this, but I will not merge it while it's still under discussion to open a hotfix variant. (Which is fine with me, btw.)

Issue #2496, opened 2024-08-08
Warning shown when using LDAP auth with empty password


somiaj:
In testing LDAP auth, I tried to enter in no password with a valid account whose password in the data base was NULL (created testing two factor auth with LDAP), and a warning is sent to the user UI. The warning is:

+ AUTH LDAP: bind error 48: The server requires the client which had attempted to bind anonymously or
+ without supplying credentials to provide some form of credentials
+  .

Seems LDAP should just error out and not try to send a blank password to the server in this case.

drgrice1:
Note that this has nothing to do with what webwork has in the database as that is not used in this case.  Only the `passwd` from the submitted form is used for this.  The same happens if you don't enter a password for a user that does have a password saved in the webwork2 database.

Also, this will happen with all versions of webwork.  The code has never considered the case that the password is not given.  Apparently this is something that doesn't happen frequently, and so no one has noticed.

That being said, this is very easy to fix.

somiaj:
Fixed in #2499

drgrice1:
Fixed by #2494.

PR #2499, opened 2024-08-08
Skip checking a password for the LDAP authentication module if not given

drgrice1:
This fixes issue #2496.


somiaj (approved):

Alex-Jordan (approved):
Approving this. I would merge, except, should this also have a hotfix variant?

PR #2501, opened 2024-08-08
Allow instructors with no email saved to enter an email address. (#2491 for develop)

drgrice1:
If an instructor does not have an email address in the database, then currently the from field on the instructor "Email" page is disabled, and there is no way to enter an email address.  Attempting to send an email address gives the message "From field must contain one valid email address", and yet there is still no way to enter an email address.
    
This makes it so that in the case that an instructor does not have an email address in the database the from field is not disabled, and the instructor can enter an email address.
    
The message was changed because I find it a bit confusing even if the instructor can change the email address, if the message says the "From field must contain one valid email address".  It seems to me that means that it would be valid to enter more than one email address as long as one is valid.  The message now reads "From field must contain a single valid email address."

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -185,10 +185,10 @@ $emailableURL
 			$msg .= "***** Data about the environment: *****\n\n" . Dumper($ce) . "\n\n";
 		}
 
-		my $from_name = $user ? $user->full_name : $ce->{generic_sender_name};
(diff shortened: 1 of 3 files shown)


somiaj:
The target needs to be updated from main to develop in the PR.

drgrice1:
@somiaj Thanks. I apparently picked the wrong target.

somiaj (approved):

Alex-Jordan (approved):
Approved, but not merging until #2491 is merged as well.

Issue #2507, opened 2024-08-12
ordering of sets in the Sets Manager


Alex-Jordan:
It is not uncommon to have multiple sets with the same dates. This happens when sets are organized by topic or by textbook section.

When you visit the Sets Manager, they are sorted by Close Date and then by Open Date. Then within blocks where all the dates are tied, the sets are ordered by however they are in the database. For example:

<img width="1401" alt="Screenshot 2024-08-12 at 4 33 19 PM" src="https://github.com/user-attachments/assets/a0c60c39-48b8-4bde-944c-67bbebd68c39">

Since the sets are named by section number here, it looks strange to see them out of order like this.

I don't think adding a third column toggle to sort on is a great idea because of the clutter. But I wonder if we could just always secretly sort on the set name as a third sorting level. As it stands, when the first two levels tie, we get things in the database order which is not great.

somiaj:
My solution to the problem when I have multiple stuff closed on the same date is to alter that open date by one minute, so I would say have 1-1 at 7:00am, 1-2 at 7:01am, and so forth. Since students rarely see the open date anyways, I do this to get the right order.

Though I do agree, having a hidden way to control the order in this case could be nice

Alex-Jordan:
So, that's exactly where this started for me. I'm trying to end the need for that hack.

somiaj:
Fallback sorting add in #2509.

PR #2509, opened 2024-08-13
Add a fallback sort to the problem set list (the Sets Manager)

drgrice1:
If the `set_id` field is neither the primary sort field nor the secondary sort field, then it is added as the tertiary sort order (always in ascending order).

This was suggested in #2507.


pstaabp (approved):
Looks like this fixes it.

Alex-Jordan (approved):

Issue #2286, opened 2023-12-13
Feature - Shibboleth authentication for WeBWorK 2.18


jrchamp:
Hi all! We're planning to upgrade to WeBWorK 2.18 in our environment and we are planning to continue using Shibboleth authentication.

Don't worry! We'll figure out how to make it work (for our own documentation) and then share that back with the community. We are confident in our ability to update the existing Shibboleth integration file for WeBWorK 2.18, especially as we already support a few Shibboleth-adjacent login plugins. The goal will be to keep it as simple as possible to minimize any maintenance concerns (for everyone involved, us included).

For now, please assign this issue to me (@jrchamp) and my colleague (@smbader). If someone, such as @jmacdone or @xcompass, get to it before we can, even better. Teamwork! :heart:

xcompass:
Great! Our timeline for the upgrade is in the summer. So most likely you will get it before us. Let us know if there is any problem and anyway we can help troubleshooting!

drgrice1:
That is great.  If you need any advice concerning Mojolicious usage, feel free to ask.

jrchamp:
Thanks @xcompass and @drgrice1!

Our current progress:

 - Using httpd as a reverse proxy with mod_shib and shibd
 - It seems like the only way to get per-request information from httpd to Mojolicious is by setting a request header
 - Modify the Authen::Shibboleth to check the request header information

Does this seem like a reasonable plan? Is there a better way to send information from httpd to Mojolicious?

drgrice1:
That seems to be the only way this can be done since the webwork2 app no longer runs inside the apache2 process.  That is unless there is a shibboleth library that the app could directly use.  I don't think there is one though.

jrchamp:
Excellent! We'll do some more testing on our side and open a pull request when ready. Plus add some documentation on the httpd configuration.

drgrice1:
That would be great.  Thanks for your work on this.

juu7nitw23:
Any update on this?  Our EDU has been using Shibboleth with WW for a while (now looking to upgrade WW) and we might have to look for alternatives to do our Math placement exams if we cannot find a solution.  Thanks

jrchamp:
Yes! It is working for us and I'm hoping to open a pull request very soon.

jrchamp:
[...]

PR #2511, opened 2024-08-14
Add SAML2 auth

ionparticle:
Hello, this is a SAML2 authen module. It has two components, a regular Webwork Authen module and a
Mojolicious plugin for SAML2. The plugin uses the Net::SAML2 library which claims to be tested against many SAML2 flavours (see list in description [here](https://metacpan.org/pod/Net::SAML2)).

As Shibboleth is also based on SAML2, this is an alternate way to address #2286 that doesn't require Apache mod_shib.

Usage info is documented in lib/Mojolicious/Plugin/Saml2/README.md

We've been running this code in prod for the summer term with no issues so far.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -190,6 +190,7 @@ RUN cpanm install -n \
 	DBD::MariaDB \
 	Perl::Tidy@20220613 \
 	Archive::Zip::SimpleZip \
+	Net::SAML2 \
 	&& rm -fr ./cpanm /root/.cpanm /tmp/*
 
 # ==================================================================
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN echo Cloning branch $PG_BRANCH branch from $PG_GIT_URL \
 FROM ubuntu:24.04
 
 ENV WEBWORK_URL=/webwork2 \
-	WEBWORK_ROOT_URL=http://localhost::8080 \
+	WEBWORK_ROOT_URL=http://localhost:8080 \
 	WEBWORK_SMTP_SERVER=localhost \
 	WEBWORK_SMTP_SENDER=webwork@example.com \
 	WEBWORK_TIMEZONE=America/New_York \
--- a/DockerfileStage1
+++ b/DockerfileStage1
@@ -152,6 +152,7 @@ RUN cpanm install -n \
 	DBD::MariaDB \
 	Perl::Tidy@20220613 \
 	Archive::Zip::SimpleZip \
+	Net::SAML2 \
 	&& rm -fr ./cpanm /root/.cpanm /tmp/*
 
 # ==================================================================
--- a/lib/Mojolicious/Plugin/Saml2/Controller/MetadataController.pm
+++ b/lib/Mojolicious/Plugin/Saml2/Controller/MetadataController.pm
@@ -1,8 +1,6 @@
 package Mojolicious::Plugin::Saml2::Controller::MetadataController;
 
-use Mojo::Base 'Mojolicious::Controller', -strict, -signatures, -async_await;
-
-use WeBWorK::Debug;
+use Mojo::Base 'Mojolicious::Controller', -signatures, -async_await;
 
 async sub get ($c) {
 	my $sp = $c->saml2->getSp();
--- a/lib/Mojolicious/Plugin/Saml2/Controller/MetadataController.pm
+++ b/lib/Mojolicious/Plugin/Saml2/Controller/MetadataController.pm
@@ -1,0 +1,12 @@
+package Mojolicious::Plugin::Saml2::Controller::MetadataController;
+
+use Mojo::Base 'Mojolicious::Controller', -strict, -signatures, -async_await;
+
+use WeBWorK::Debug;
+
+async sub get ($c) {
+	my $sp = $c->saml2->getSp();
+	return $c->render(data => $sp->metadata(), format => 'xml');
+}
+
+1;
--- a/lib/Mojolicious/Plugin/Saml2/Router.pm
+++ b/lib/Mojolicious/Plugin/Saml2/Router.pm
@@ -1,0 +1,19 @@
+package Mojolicious::Plugin::Saml2::Router;
+
+use Mojo::Base -strict, -signatures;
+
+use WeBWorK::Debug;
+
+sub setup ($app, $conf) {
+	my $subRouter =
+		$app->routes->any($conf->{sp}{route}{base})->to(namespace => 'Mojolicious::Plugin::Saml2::Controller')
+		->name('saml2.base');
+	$subRouter->get($conf->{sp}{route}{metadata})->to(controller => 'MetadataController', action => 'get')
+		->name('saml2.metadata');
+	$subRouter->get($conf->{sp}{route}{error})->to(controller => 'ErrorController', action => 'get')
+		->name('saml2.error');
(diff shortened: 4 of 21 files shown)


Dockerfile
```diff
@@ -190,6 +190,8 @@ RUN cpanm install -n \
 	DBD::MariaDB \
 	Perl::Tidy@20220613 \
 	Archive::Zip::SimpleZip \
+	SQL::Abstract::Classic \
```
drgrice1, line 193: `SQL::Abstract::Classic` should not be installed.  Remove this.  Webwork2 no longer uses it.

DockerfileStage1
```diff
@@ -152,6 +152,8 @@ RUN cpanm install -n \
 	DBD::MariaDB \
 	Perl::Tidy@20220613 \
 	Archive::Zip::SimpleZip \
+	SQL::Abstract::Classic \
```
drgrice1, line 155: Make sure to also remove `SQL::Abstract::Classic` here.

lib/WeBWorK/Authen/Saml2.pm
```diff
@@ -0,0 +1,102 @@
...
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+package WeBWorK::Authen::Saml2;
+use Mojo::Base 'WeBWorK::Authen', -strict, -signatures;
```
drgrice1, line 17: `-strict` is not needed here.  If any flags are given (like `-signatures`), then `strict`, `warnings`, and `utf8`, are automatically imported by the `Mojo::Base` module.  The only time you actually need to give the `-strict` flag to get `strict`, `warnings`, and `utf8` is if no other flags are used.

lib/Mojolicious/Plugin/Saml2/Controller/MetadataController.pm
```diff
@@ -0,0 +1,12 @@
+package Mojolicious::Plugin::Saml2::Controller::MetadataController;
+
+use Mojo::Base 'Mojolicious::Controller', -strict, -signatures, -async_await;
+
+use WeBWorK::Debug;
```
drgrice1, line 5: This line should be removed since there are no `debug` calls in this file.

lib/Mojolicious/Plugin/Saml2/Router.pm
```diff
@@ -0,0 +1,19 @@
+package Mojolicious::Plugin::Saml2::Router;
+
+use Mojo::Base -strict, -signatures;
+
+use WeBWorK::Debug;
```
drgrice1, line 5: Please remove this line since there are no `debug` calls in this file.

lib/Mojolicious/Plugin/Saml2/Saml2Plugin.pm
```diff
@@ -0,0 +1,156 @@
+package Mojolicious::Plugin::Saml2::Saml2Plugin;
+use Mojo::Base 'Mojolicious::Plugin', -strict, -signatures;
+# external libs
+use Data::Dumper;
```
drgrice1, line 4: Please remove this module.  It is not used in this file.

lib/Mojolicious/Plugin/Saml2/Saml2Plugin.pm
```diff
@@ -0,0 +1,156 @@
...
+use File::Temp qw/ tempfile /;
+use JSON;
```
drgrice1, line 6: This should be `use Mojo::JSON qw(encode_Json);`.

lib/Mojolicious/Plugin/Saml2/Saml2Plugin.pm
```diff
@@ -0,0 +1,156 @@
...
+use Mojolicious;
+use Mojolicious::Plugin::NotYAMLConfig;
+use Net::SAML2::IdP;
+use Net::SAML2::SP;
+use URN::OASIS::SAML2 qw(BINDING_HTTP_POST BINDING_HTTP_REDIRECT);
+# external libs for NotYAMLConfig
+use CPAN::Meta::YAML;
+use Mojo::Util qw(decode encode);
+# webwork modules
+use WeBWorK::Debug;
```
drgrice1, line 16: This should be `use WeBWorK::Debug qw(debug);`.  I know that this is done incorrectly all over in our code, and that needs to be fixed.  Since this is new code, please do it correctly now so it doesn't need to be fixed later.  Explicitly import what is needed.

drgrice1 (commented):
I see some initial code changes that are needed, and I commented on them.

Note that the `lib/Mojolicious/Plugin/SAML2/README.md` file needs a little clean up.  Try to adhere to a 120 character limit on lines as much as possible, and fix linting errors shown by markdownlint.  I see errors for missing blank lines around lists, hard tabs, duplicate headings, trailing spaces, and headings that increment to far.  If you don't know how to run markdownlint, I can help you with this.

The php files use a four space tab instead of hard tabs.  The project contains an `.editorconfig` file and this conflicts with the settings in that file.  Note that only YAML files use a soft tab.

This will take some time to review thoroughly since there is a lot of code.  Also, most of us will not be able to test this, so ideally someone else that can test SAML would provide a review.

In any case, thanks for the work.  This will make a nice addition to the current authentication modules.

drgrice1:
Also, I would recommend that you rebase this onto the develop branch.  I see that this is missing some of the recent changes.  It would be best for you and others to test this with everything in develop in case there are regressions caused by other changes.

ionparticle:
Thanks for the speedy review, I've rebased latest develop and hopefully fixed all the issues. 

One exception though is regarding the php files. The stuff in docker-config/idp is a separate service, it's meant to run an instance of SimpleSAMLphp configured as an Identity Provider to allow easier testing by other developers (instructions in README.md). It's an external development tool, not part of the SAML2 plugin itself and should not be present in any kind of prod use. I would request they be considered external to Webwork and exempt. They're following SimpleSAMLphp standards because they're from that project.

Another note, when I ran markdownlint, it was enforcing 80 characters per line, so my fixes followed that standard.

drgrice1:
If a file is included in the webwork2 repository it is part of the webwork2 project, and can not be exempted from any coding standards of the project.  It will not be allowed to follow any external projects coding standards, and can not be considered external.  External files are files not in the webwork2 repository.  If the files are directly from another project, then they should be used from that project instead of being included in the repository.  For php, composer should be used.  Although, this is another thing that I wanted to bring up.  Is there any way that could be done with Perl instead of using php to begin with?

For the `README.md` file a 80 character limit is acceptable.  Although if you use markdownlint correctly, it will honor the settings in the `.editorconfig` file and use a 120 character limit.

ionparticle:
No problem, reformatted docker-config/idp to follow editorconfig rules.

Though, I want to clarify that git and composer are already used to grab source files from SimpleSAMLphp during the docker build of the idp. The files checked in here are just the configuration files which necessarily needs to be customized.

drgrice1:
I see.  Thanks for the clarification.  If those are files that need to be customized, then they should be considered part of the webwork2 project.  Thanks for reformatting them.

I will try to take another deeper look at this when I have more time.

PR #2513, opened 2024-08-15
when copying sets from a new course and adding an instructor, assign …

Alex-Jordan:
…all sets to that instructor

In 2.19, when you are creating a new course, you have the option to select an existing course and copy all of the sets into the new course. You also (as you've always been able to do) have the option of adding one "professor" level user to the new course. As things stand, the new professor will not have all of the copied sets assigned to them.

This commit changes it so that they will get all of those sets assigned to them.

I can't see a scenario where you want to copy sets from another course, but would not want them to be assigned to the new instructor. Maybe there is such a scenario. But I think it would be much more common to want that instructor to have everything assigned to them.

### DIFF
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -405,6 +405,10 @@ sub addCourse {
 				assignSetsToUsers($db, $ce, \@user_sets, [$user_id]);
 			}
 		}
+		if (defined $options{initial_userID}) {
+			my ($initialUser) = grep { $_->[0]{user_id} eq $options{initial_userID} } @users;
+			assignSetsToUsers($db, $ce, \@set_ids, [ $initialUser->[0]{user_id} ]) if $initialUser;
+		}
 	}
 
 	# add achievements
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -201,7 +201,7 @@ boolean options:
 =cut
 
 sub addCourse {
-	my (%options) = @_;
+	my (%options) = (initial_userID => '', @_);
 
 	for my $key (keys(%options)) {
 		my $value = '####UNDEF###';
@@ -216,6 +216,8 @@ sub addCourse {
 	my @users         = exists $options{users} ? @{ $options{users} } : ();
 
 	debug \@users;
+
+	my ($initialUser) = grep { $_->[0]{user_id} eq $options{initial_userID} } @users;
 
 	# get the database layout out of the options hash
 	my $dbLayoutName = $courseOptions{dbLayoutName};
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I have a few more changes coming that have been motivated by recent experiences using 2.19 with PCC and Runestone. They could all have hotfix counterparts, but as with something before, I'm unsure if we are still doing that. Are we just rolling with more hotfixes through August? (That's fine with me.)

drgrice1:
If the changes are not too great and are something that we want to get in, then we can still hotfix.

Alex-Jordan:
I reviewed the list of other things I have, and they are all either too big to accomplish anytime soon, or I think will be controversial. So please forget when I said:
> I have a few more changes coming

lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -405,6 +405,12 @@ sub addCourse {
 				assignSetsToUsers($db, $ce, \@user_sets, [$user_id]);
 			}
 		}
+		for my $userTriple (@users) {
+			if ($userTriple->[0]{user_id} eq $options{initial_userID}) {
+				assignSetsToUsers($db, $ce, \@set_ids, [ $userTriple->[0]{user_id} ]);
+				last;
+			}
+		}
```
drgrice1, line 413: This `for` loop should be wrapped in `if (defined $options{initial_userID}) { ... }`.  It is possible that the `addCourse` method could be called with the `copySets` option set, but the `initial_userID` option not set which would result in a warning.  This won't happen when the `addCourse` method is used by webwork2, but could happen if it is used by some other script.  This module is designed to be used externally by local scripts.

Also, in this case you are just looking for a single element of the @users array.  There is a standard way of doing this in Perl, and it is more succinct.  This is by using a grep as follows:
```suggestion
	   	if (defined $options{initial_userID}) {
			my ($initialUser) = grep { $_->[0]{user_id} eq $options{initial_userID}} @users;
			assignSetsToUsers($db, $ce, \@set_ids, [ $initialUser->[0]{user_id} ]) if $initialUser;
		}
```

drgrice1 (approved):
Looks good.

Alex-Jordan:
Thanks, both this and the hotfix branch have the changes you suggested now.

somiaj (approved):
Looks good and works in my tests, and assigning sets to the initial professor will probably help out, especially newer professors.

somiaj:
Just some comments in my testing of this. First I figured out that the first name, last name, and email address of this new prof that is being added are required (as I didn't include them and got an error), though the email address isn't validated for form, it just needs to be non blank. Not a big deal, but in comparison to adding a user inside a course only the username is needed, and nothing else is required.

Second, What about assigning achievements to this new professor that are copied.  I am unsure what would be preferable but am slightly on the side of not copying them, I just thought I would mention this as well.

Alex-Jordan:
I made it so the achievements are also assigned to the new instructor.

Issue #2515, opened 2024-08-15
Error occurs on course copy


pstaabp:
I'm setting up courses for the semester and using the new course copy options.  I'm only adding a new course_id, and the selecting all options available (except the course.conf).  

Upon creating the course and then logging in, I get the error:
```
Use of uninitialized value in concatenation (.) or string at template layouts/system.html.ep line 126.
```
indicating that `page_title` is empty.   

Looking in the database a the `_setting` table shows both the `courseTitle` and `courseInstitution` fields are null and the course I copied from does not have these in the database.  Where are these stored if not in the db?

somiaj:
Could this be related to #2497, though that was initially only reported for the admin course?

drgrice1:
I am unable to reproduce this.  I created a course with no title or institution.  Logging into that course gives no errors or warnings.  Then I copying that course again not providing a title or institution.  Logging into the copy course also gives no errors or warnings.  I see in the database that neither course has a title or institution in the setting table.  Note that the `page_title` method checks to see if the `courseTitle` is set before using it and returns the route title for the current route if it does not.  So I don't see how it could give a warning.

pstaabp:
The error occurred on my production machine.  I also can't reproduce on my development machine.  

@drgrice1 Doesn't `->can` just determine if there is a code ref for that (in the case of line 122 on `system.html.ep`) or does it also check if the result is undefined or empty string?

drgrice1:
`$c->can('page_title')` checks to see if the `$c` object can call a method named "page_title".  By default all content generator modules can call that method since it is in the base class.  However, a derived class could override the `can` method and make it return false for a method, and thus preventing it from being called in the templates.

It does not check for anything being undefined or the empty string.

somiaj:
Okay, looking a little closer, it appears that if there is a row in `courseID_setting`, that sets `(name, value) = (courseTitle, NULL)`, it causes the warning. Deleting this row so there is no `courseTitle` or setting the course title to an empty string or something non NULL makes the warning go away.

The issue appears is because in my case the admin course has no `courseTitle` setting, so when it is copied over, it is set to NULL because it doesn't exist in the database.

I'm unsure if the issue is when copying the course title/institution in adding a course, if it should just ignore copying settings that are not set or NULL, or if there needs to be a fall back to deal with the course title being set to NULL in the database.

drgrice1:
[...]

Issue #2497, opened 2024-08-08
No title for admin course


Alex-Jordan:
It seems that my admin course has no course title. (There is no `courseTitle` row in the `admin_setting` table. Nor is there a `courseInstitution` row.) This is an admin course that propagated from earlier versions of WW (although the admin upgrade script was applied each time), and I am unsure if a new admin course on a new installation would not have a title.

This is not a problem, because when you are in the admin course, nothing tries to access its title.

But now it will be common to copy the "admin" course to some new course with a different name that is destined to be the new admin course (just with a new name). When you copy the "admin" course, it would be natural to "select all" so that you copy everything possible. This is leading to the newly created course also having no title. And now if you try to enter that course, there are perl warnings about the title being undefined. Soon you will toggle the config setting to make this the new admin course, and then it won't matter again. But it's unsettling if you go to this course to check on things and get these warnings.

Once logged in to the new course, you can go to Course Config and give it a title. That is, if you understood the warnings well enough to know to try that.

You cannot do this in the original admin course while it is still the active admin course, because we don't give access to Course Config. (Well, you can still make your browser go to `webwork2/admin/instructor/config` but we don't offer a TOC link to the Config page.)

Long story short: I think that the update admin script should give the admin course a title now, if it does not already have one. This means the script would first check if the course has a title, and I don't know sql scripting well enough to do that. Also, if no title exits, what title should it get? An empty string? `Admin`? `Course Administration`?

Alex-Jordan:
And if it is not already done, the course integrity check (like what is run when you are in the admin course and go to Upgrade Courses) should be looking to see that there is a `courseTitle` row in the `settings` table. Since without such a row, there will be this perl warning.

somiaj:
This was fixed in #2516.

PR #2516, opened 2024-08-15
Prevent a NULL title from being saved in the database when a course is created/copied.

drgrice1:
Also don't use an undefined course title on the course home page.  Note that a NULL setting in the database becomes undef when converted to a Perl scalar value.

This fixes issue #2515 and probably #2497 as well.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -300,9 +300,9 @@ sub do_add_course ($c) {
 	my $db    = $c->db;
 	my $authz = $c->authz;
 
-	my $add_courseID          = trim_spaces($c->param('new_courseID'))          // '';
-	my $add_courseTitle       = trim_spaces($c->param('add_courseTitle'))       // '';
-	my $add_courseInstitution = trim_spaces($c->param('add_courseInstitution')) // '';
+	my $add_courseID          = trim_spaces($c->param('new_courseID')) // '';
+	my $add_courseTitle       = ($c->param('add_courseTitle')       // '') =~ s/^\s*|\s*$//gr;
+	my $add_courseInstitution = ($c->param('add_courseInstitution') // '') =~ s/^\s*|\s\*$//gr;
 
 	my $add_initial_userID          = trim_spaces($c->param('add_initial_userID'))          // '';
 	my $add_initial_password        = trim_spaces($c->param('add_initial_password'))        // '';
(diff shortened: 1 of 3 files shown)


somiaj (approved):
Fixes the issue. Tested that it fixes the warning if `courseTitle` is `NULL` and also prevents copying over the `courseTitle` in the settings table if it doesn't exist in the source course.

drgrice1:
By the way, with the approach I used a course title or institution of `0` is not allowed.  Should it be?

Alex-Jordan:
I think it should be allowed. Just to be consistent since we let other things be "0" (like set ids).

drgrice1:
Okay, so a course title or institution of `0` is now also allowed.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -300,9 +300,9 @@ sub do_add_course ($c) {
 	my $db    = $c->db;
 	my $authz = $c->authz;
 
-	my $add_courseID          = trim_spaces($c->param('new_courseID'))          // '';
-	my $add_courseTitle       = trim_spaces($c->param('add_courseTitle'))       // '';
-	my $add_courseInstitution = trim_spaces($c->param('add_courseInstitution')) // '';
+	my $add_courseID          = trim_spaces($c->param('new_courseID')) // '';
```
Alex-Jordan, line 303: This brings to my attention that we also cannot have a course named "0". Is that what we want? I don't think I would ever name a course "0", but I could see someone doing that as some sort of orientation or training course. Maybe.

If you try to create a course named "0" right now using the admin course, you are met with "You must specify a course ID.". And for anyone who is not a perl cognoscente, you would think "But I did!".

I feel like `trim_spaces` is part of what is broken here and it should not turn `0` into  ``. That is not consistent with what the name of the macro suggests it does.

drgrice1: I agree that `trim_spaces` is not doing the right thing.  I thought about changing that, but if we do so we need to review all of its uses.  There aren't so many of them, so that isn't that hard.  Most of the uses are in this file, there is one usage in `lib/WeBWorK/Utils.pm` (for the password), and the rest are in `lib/WeBWorK/ContentGenerator/Instructor/AddUsers.pm`.

As to allowing a course id of "0", that may be problematic.  I considered changing it when I did this, but then thought that there may be problems with that.  The course title is only used in a few places, so it doesn't take much to make sure that is okay.  The course institution isn't actually used at all, and is purely something saved in the database (not sure why we have it).  The course id, on the other hand, is used everywhere.  So it would take some time to verify that everything will work okay with it being "0".

Alex-Jordan (approved):

PR #2518, opened 2024-08-16
Stop adding a "contact" user to the admin course when creating a course.

drgrice1:
There is no need for this. These users are not used for anything and have no real purpose.  Furthermore, the users are just clutter in the admin course accounts manager.

There are also issues that can be caused by these users existing.  One issue occurs if you create a course with an initial user, then delete that course, then later create a course with the same course id but with a different initial user.  Then if you delete the second course it dies with the error "Incorrect number of contacts for the course courseID".

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -422,47 +422,6 @@ sub do_add_course ($c) {
 				$add_initial_email,
 			)
 		);
-		# Add contact to admin course as student?
-		# FIXME -- should we do this?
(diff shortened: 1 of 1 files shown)


somiaj (approved):
:+1: 

Can we make this a hotfix too?

drgrice1:
I would be okay with making this a hotfix.

Alex-Jordan (approved):
I'm OK with this. In theory this gave a way to email all of the current instructors. For example to issue an alert about upcoming maintenance. But it doesn't work in practice for a number of reasons.

somiaj:
`# FIXME -- should we do this?`....why yes we should. Two approvals so I'll merge.

PR #2519, opened 2024-08-16
Show newly imported achievements to the Achievement List.

somiaj:
This ensures that any imported achievement is shown in the achievement list when they are imported (currently you have to reload the page or apply a new filter to see them). This also honors any currently filtered list by adding the newly imported achievements to the current filtered list (vs just showing all achievements after an import).

### DIFF
--- a/templates/ContentGenerator/Instructor/AchievementList/filter_form.html.ep
+++ b/templates/ContentGenerator/Instructor/AchievementList/filter_form.html.ep
@@ -7,7 +7,7 @@
 					[ maketext('all course achievements')              => 'all' ],
 					[ maketext('selected achievements')                => 'selected' ],
 					[ maketext('enter matching achievement IDs below') => 'match_ids', selected => undef ],
-					[ maketext('enter matching category below')        => 'match_category' ],
+					[ maketext('select matching category below')       => 'match_category' ],
 					[ maketext('enabled achievements')                 => 'enabled' ],
 					[ maketext('disabled achievements')                => 'disabled' ]
 				],
@@ -27,10 +27,7 @@
 	<div id="filter_text_err_msg" class="alert alert-danger p-1 mb-2 d-inline-flex d-none">
 		<%= maketext('Please enter a list of IDs to match.') %>
 	</div>
-	% my @categories = $c->db->getAchievementCategories;
-	% for (@categories) {
-		% $_ = [$_ => $_];
-	% }
+	% my @categories = grep { defined && $_ ne '' } $c->db->getAchievementCategories;
 	<div id="filter_category_elements" class="row mb-2">
 		<%= label_for 'filter_category', class => 'col-form-label col-form-label-sm col-sm-auto', begin =%>
 			<%= maketext('Match on which category?') =%>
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I haven't looked, but do you know if this is the same approach used in the Accounts Manager and Sets Manager when an import is made? It would be nice if these were handled consistently.

somiaj:
When importing homework sets, they appear in the list of sets right after the import. I'm unsure on the subtly I have of adding them to the currently filtered items vs just showing all items though.

I also noticed there is another similar issue, when creating a new achievement, it won't appear in the achievement list on the submit after it is created, and the page also needs to be reloaded. I'll look into that and add it to this PR too.

somiaj:
Update the patch to also add newly created achievements to the `visibleAchivementIDs` list so they appear on the page after the action happens.

drgrice1 (approved):
Looks good.

drgrice1:
It is somewhat confusing when you import an achievement, and it says the achievement was imported, and yet it does not appear in the list.  Similarly for creating an achievement.

When testing this I noticed that there is something wrong with creating a new empty achievement (not caused by this pull request -- it also happens with develop).  The new achievement is created with no name, number, or category, and there is a warning "Use of uninitialized value in hash element at /usr/local/share/perl/5.38.2/Mojolicious/Plugin/TagHelpers.pm line 149."    Furthermore, that warning comes up every time the achievement list is loaded after creating the empty achievement.

drgrice1:
It seems that it is the lack of a "category" that causes the warning.

somiaj:
The error seems to happen if the `category` in the database is NULL or the empty string. The error comes from the drop down menu to filter based on category, so I added some logic to deal with NULL or emptystring categories in the dropdown.

somiaj:
Note you can also create this same warning by editing an achievement and setting the category to the empty string.

templates/ContentGenerator/Instructor/AchievementList/filter_form.html.ep
```diff
@@ -29,7 +29,7 @@
 	</div>
 	% my @categories = $c->db->getAchievementCategories;
 	% for (@categories) {
-		% $_ = [$_ => $_];
+		% $_ = (defined($_) && $_ ne '') ? [$_ => $_] : ["\x{27E8}" . maketext('blank') . "\x{27E9}" => ''];
 	% }
```
drgrice1, line 33: Replace lines 30-33 with
```
	% my @categories = grep { defined && $_ ne '' } $c->db->getAchievementCategories;
```
There is no need to make an array of array references when the text an value of the options are the same.  A string is sufficient.

somiaj: Though wouldn't that make it so a user couldn't filter by the blank category? Although I can't think of a use case for that, I'll update as you suggest.

somiaj: Nevermind, filtering by blank didn't work anyways.

drgrice1: Filtering by a blank category would require more changes than just adding it to the list.  I was going to open a separate pull request with this change, and then saw that you added it here, but this change should be a hotfix since it is a bug.  On the other hand, the changes needed to make filtering by the blank category I think are too much for a hotfix.

Alex-Jordan (approved):

PR #2520, opened 2024-08-16
Fix spelling typos in Student Orientation problems.

somiaj:


drgrice1 (approved):

Alex-Jordan (approved):
Sorry for all the embarrassing typos here.

Although, "spacebar" versus "space bar" is a style choice. This ngram is interesting too: https://books.google.com/ngrams/graph?content=spacebar%2C+space+bar&year_start=1800&year_end=2022&corpus=en&smoothing=3

Alex-Jordan:
Do we have anything yet that in `develop` versus `main` where there is a real difference? Maybe instead of all these hotfix PRs, we could just do another merge from one into the other.

somiaj:
@Alex-Jordan I was on the fence about modifying spacebar, but left it in the changes in the end. Since these are your problems, I can revert it to your previous style.

Alex-Jordan:
It's fine as "space bar", I am not requesting it to be changed :)

somiaj:
I'll make a hotfix PR for this too, assuming others would agree with that.

PR #2521, opened 2024-08-16
Also set sub_status when giving credit with achievement items.

somiaj:
This ensures that the progress bar reflects they have 'earned' full credit on the problems.

### DIFF
--- a/lib/WeBWorK/AchievementItems/FullCreditProb.pm
+++ b/lib/WeBWorK/AchievementItems/FullCreditProb.pm
@@ -100,8 +100,9 @@ sub use_item ($self, $userName, $c) {
 	my $problem = $db->getUserProblem($userName, $setID, $problemID);
 	return 'There was an error accessing that problem.' unless $problem;
 
-	# Set the status of the problem to one.
+	# Set the status and sub_status of the problem to one.
 	$problem->status(1);
+	$problem->sub_status(1);
 	$db->putUserProblem($problem);
 
 	$globalData->{ $self->{id} }--;
--- a/lib/WeBWorK/AchievementItems/HalfCreditSet.pm
+++ b/lib/WeBWorK/AchievementItems/HalfCreditSet.pm
@@ -26,7 +26,7 @@ sub new ($class) {
 	return bless {
 		id          => 'HalfCreditSet',
 		name        => x('Lesser Tome of Enlightenment'),
-		description => x('Gives half credit on every problem in a set.')
+		description => x('Increases the score of every problem in an assignment by 50%, to a maximum of 100%.')
 	}, $class;
 }
 
@@ -41,7 +41,9 @@ sub print_form ($self, $sets, $setProblemIds, $c) {
 	return unless @openSets;
 
 	return $c->c(
-		$c->tag('p', $c->maketext('Please choose the set for which all problems should have half credit added.')),
+		$c->tag(
+			'p', $c->maketext('Please choose the assignment for which all problems should have half credit added.')
+		),
 		WeBWorK::AchievementItems::form_popup_menu_row(
 			$c,
 			id         => 'hcs_set_id',
--- a/lib/WeBWorK/AchievementItems/HalfCreditSet.pm
+++ b/lib/WeBWorK/AchievementItems/HalfCreditSet.pm
@@ -73,11 +73,11 @@ sub use_item ($self, $userName, $c) {
 		my $problem = $db->getUserProblem($userName, $setID, $probID);
 
 		# Add .5 to grade with max of 1.
-		if ($problem->status < .5) {
-			$problem->status($problem->status + .5);
-		} else {
-			$problem->status(1);
-		}
+		my $new_status = $problem->status + 0.5;
+		$new_status = 1 if $new_status > 1;
+		$problem->status($new_status);
+		$problem->sub_status($new_status);
+
 		$db->putUserProblem($problem);
 	}
 
--- a/lib/WeBWorK/AchievementItems/FullCreditSet.pm
+++ b/lib/WeBWorK/AchievementItems/FullCreditSet.pm
@@ -72,8 +72,9 @@ sub use_item ($self, $userName, $c) {
 	for my $probID (@probIDs) {
 		my $problem = $db->getUserProblem($userName, $setID, $probID);
 
-		# Set status to 1.
+		# Set status and sub_status to 1.
 		$problem->status(1);
+		$problem->sub_status(1);
 		$db->putUserProblem($problem);
 	}
 
--- a/lib/WeBWorK/AchievementItems/HalfCreditProb.pm
+++ b/lib/WeBWorK/AchievementItems/HalfCreditProb.pm
@@ -28,7 +28,7 @@ sub new ($class) {
 	return bless {
 		id          => 'HalfCreditProb',
 		name        => x('Lesser Rod of Revelation'),
-		description => x('Gives half credit on a single homework problem.')
+		description => x('Increases the score of a single problem by 50%, to a maximum of 100%.')
 	}, $class;
 }
 
@@ -60,7 +60,7 @@ sub print_form ($self, $sets, $setProblemIds, $c) {
 		$c->tag(
 			'p',
 			$c->maketext(
-				'Please choose the set name and problem number of the question which should be given half credit.')
+				'Please choose the assignment name and problem number of the question to add half credit to.')
 		),
 		WeBWorK::AchievementItems::form_popup_menu_row(
 			$c,
--- a/lib/WeBWorK/AchievementItems/HalfCreditProb.pm
+++ b/lib/WeBWorK/AchievementItems/HalfCreditProb.pm
@@ -101,12 +101,10 @@ sub use_item ($self, $userName, $c) {
 	return 'There was an error accessing that problem.' unless $problem;
 
 	# Add .5 to grade with max of 1
-
-	if ($problem->status < .5) {
-		$problem->status($problem->status + .5);
-	} else {
-		$problem->status(1);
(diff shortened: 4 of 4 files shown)


Alex-Jordan (approved):

drgrice1 (approved):

somiaj:
@drgrice1 and @Alex-Jordan, you will need to confirm your approval again as I noticed the same issue with the half credit achievement item. This also updates the `sub_status` when using that item as well.

somiaj:
I thought the description "Gives half credit on a single homework problem." was ambiguous and updated it. I can move that to another PR or drop it if it is too much for this PR.

drgrice1 (approved):
Still all looks fine to me.

lib/WeBWorK/AchievementItems/HalfCreditSet.pm
```diff
@@ -26,7 +26,7 @@ sub new ($class) {
 	return bless {
 		id          => 'HalfCreditSet',
 		name        => x('Lesser Tome of Enlightenment'),
-		description => x('Gives half credit on every problem in a set.')
+		description => x('Increases the score of every problem in a homework set by 50%, to a maximum of 100%.')
```
Alex-Jordan, line 29: By adding the word "homework" here, are you implying that this can't be applied to a test set? I thought it was possible to enable achievements to be applied to a test. I don't use tests in my courses, so I'm less familiar with all that.

drgrice1: Yeah, it should say "in an assignment" instead of "in a homework set".  I didn't read that closely enough. Good catch.

Alex-Jordan:
I'm confused about this. Are you talking about this progress bar?

Screenshot 2024-08-16 at 6 49 51 PM

That's based on the "attempted" property of user-problem records. 

I'm having trouble looking at the code and understanding what `sub_status` is actually for, but it seems to have something to do with how reduced scoring is implemented. Can you explain how this affects a progress bar somewhere?

somiaj:
@Alex-Jordan `sub_status` is the amount of points earned during the full credit period, outside of the reduced scoring period `status = sub_status`, but inside the reduced scoring period, only the difference between the new score and `sub_status` is penalized. @drgrice1 added a patch that uses a method `compute_unreduced_score()` to determine if the problem is complete or not taking `sub_status` into account.

In short if we don't also increase the `sub_status`, `compute_unreduced_score()` won't be accurate, so make `status = sub_status` makes it so it was like the problem was completed before any reduced scoring was applied. That help?

drgrice1:
The `sub_status` is the score a student has on a problem before the reduced scoring period begins.  Until the reduced scoring period begins the `sub_status` and `status` are both set to the same thing.  Once the reduced scoring period begins, the `sub_status` is never set again (usually).  After the reduced scoring period begins the `sub_status` is used to compute the reduced score that the `status` is set to when a student answers a problem.  It is also used to recover the "unreduced" score later (i.e., the equivalent full score the student would have received if not in the reduced scoring period).  That is how the progress bar (the one you referred to) determines that a student has attained the maximum possible score for a problem in the reduced scoring period and no longer shows that the problem is in progress.  The computations are done in the `compute_reduced_score` and `compute_unreduced_score` methods in `lib/WeBWork/Utils/ProblemProcessing.pm`.

lib/WeBWorK/AchievementItems/HalfCreditSet.pm
somiaj: Most achievement items only can be applied to homework sets, all four in question in this PR only work on sets that meet the condition, `$set->assigment_type eq 'default'`. So these can only work on homework. As someone who has both homework and tests, this feature is what I would prefer.

There have been some achievement items that have been made to work on tests, such as `ExtendDueDateGW.pm`.

Really this should be configurable if achievement items work on homework or tests, and also ensure that achievement items honor the setting of which sets should be ignored for earning achievement points. But that is beyond what I was wanting to do here.

drgrice1: That's right.  I forgot there are specific achievements that work for tests.  So then the wording should be changed to "in an "assignment" instead of "in a homework set".  In other words, it doesn't matter what the achievements apply to.  We are trying to get away from the terminology of "homework" and "set".  Both words have issues.  Just because a set is a "default" set in the code does not mean that it is "homework" for the students.  They may work the assignment in class.  The word "set" is more of a programmer word for these things than what students call it.

somiaj: Updated the language to use "assignment" instead of "homework set".

Alex-Jordan:
Thanks for the explanations. I get it.

In these achievement items that increase one or more problems' status (and possibly all the way up to 1) is there any reason to also set the problem's "attempted" value to 1? Does that also have a role in the progress bar?

somiaj:
My testing shows that this makes it so the progress bar correctly shows for problems these items are used with. I originally tried to also set `num_correct = 1`, but that didn't fix the issue. I'm unsure if setting the `num_correct` or `attempted` values to 1 is needed, but I've only tested this with a single set and the progress bar which now functions as I would expect it to.

Alex-Jordan:
Before I merge this, do you want to open a hotfix for it too? (It helps me keep track of things if one of these PRs and its parallel hotfix are merged at the same time.)

somiaj:
I can open a hotfix, one moment.

somiaj:
@drgrice1 I noticed a similar issue with the `sub_status` and the single problem grader (probably the manual problem grader) as well. I give a student a 100% to deal with a bug, and the progress bar has the same issue as here, of not registering the 100% correctly.

I am just mentioning this here since it related, unsure if/what needs to be fixed. Maybe `compute_unreduced_score` would just return `1` in the case `status` is already 1, or should the problem graders update `sub_status` as well, basically any manual grade change overrides and removes any reduced scoring difference making `sub_status = status`.

drgrice1 (changes requested):
The problem graders definitely should not set the `sub_status` to be the same thing as the `status`.

Now that you mention this I am starting to question the change in this pull request (at least for the half-credit problem and set achievement items).

I think that this pull request should be dropped.  It seems that the real issue is with the `compute_unreduced_score` method.  The problem is that if the `status` is increased beyond the possible maximum score in the reduced scoring period, then the `compute_unreduced_score` method is returning a number greater than 1.  The progress bar code checks that the unreduced score is exactly 1 (this should probably also be changed to a greater than equal to), and if it is not 1, then it assumes the problem is not completely correct.  That would solve the issues you are seeing from these achievement items and the issues from the problem grader.

Really, the `sub_status` should never change again after the reduced scoring period has started.

somiaj:
I agree this should be side lined and probably discuss what to do with `sub_status` when something overrides a grade, such as achievement items and manual graders. If it is still before the reduced scoring date when the student uses the half credit item, should `sub_status` be updated then? What if a student has 40%, uses the half-credit item to increase their grade to 90%, but then `compute_unreduced_score` sees this as a number > 1, so the progress bar says the problem is complete?

Though since this all comes from trying to make the progress bar better reflect students current grade, the hypothetical I described up there isn't that much of an issue if they progress bar shows complete, compared to the current state.

drgrice1:
If the achievement is used before the reduced scoring date, then yes, the `sub_status` should be updated.  Before the reduced scoring date, the `status` and `sub_status` should always be the same.

The case you described where a student has 40%, uses the half-credit item to increase their grade to 90%, will work correctly with #2530 (or #2531).  If a student has 90% and the unreduced score was less than that, then the problem is complete, and the student can not work the problem further to increase the score.  So the progress bar should show that the problem is complete.

drgrice1:
So this pull request should be changed to only set the `sub_status` to be the same as the `status` only in the case that it is before the reduced scoring date.  Then in combination with #2530 the issue is resolved.

somiaj:
With your view point of the status bar check saying if a student can increase their score or not, show the check in my hypothetical would be fine.

I updated this to only update `sub_status` outside of the reduced scoring period. Currently I'm only checking if the set is configured to use reduced scoring, but should I also check if reduced scoring is enabled in the course environment as well? I think it is.

drgrice1:
Your code is only checking if reduced scoring is enabled in the course environment, it isn't checking if reduced scoring is enabled for the set.  That also needs to be checked.  Reduced scoring could be enabled in the course environment, but not for the particular set the problem is in.

lib/WeBWorK/AchievementItems/FullCreditProb.pm
```diff
@@ -102,6 +102,13 @@ sub use_item ($self, $userName, $c) {
 
 	# Set the status of the problem to one.
 	$problem->status(1);
+
+	# Update the sub_status unless inside the reduced scoring period.
+	$problem->sub_status(1)
+		unless ($ce->{pg}{ansEvalDefaults}{enableReducedScoring}
+			&& $problem->enable_reduced_scoring
```
drgrice1, line 109: There is no `enable_reduced_scoring` column for the `user_problem`.

drgrice1:
I think for simplicity, maybe just revert this to the way it was before.  For the full credit achievements there is no issue with setting both the `status` and `sub_status` to 1.  For the half credit problem if both are increased by 50%, it does mean that the student now has a higher maximal reduced credit score possible, but that is probably not that serious.  The student is using an achievement after all.  Make that a hidden benefit of the achievement.

somiaj:
Grr, I was checking the problem and not the set. Anyways, I agree that it makes a nice hidden bonus and simplifies things so I removed the commit.

drgrice1:
Cool .  `0 commits` again!

drgrice1 (approved):
Re-injecting my approval.

PR #2523, opened 2024-08-17
Add achievement items that modify reduced scoring.

somiaj:
This adds achievement items that modify the reduced scoring time period while leaving other settings alone. These are designed for classes which use reduced scoring and keep assignments open in the reduced scoring period for a long time (such as the end of the course).

+ ExtendReducedDate -> Extends the reduced scoring date 24 hours.
+ SuperExtendReducedDate -> Extends the reduced scoring date 48 hours.
+ NoReducedCred -> Removes the reduced scoring flag on an assignment to allow full credit after the reduced scoring date.

In short these provide extensions for a reduced scoring setup.

### DIFF
--- a/lib/WeBWorK/AchievementItems.pm
+++ b/lib/WeBWorK/AchievementItems.pm
@@ -27,11 +27,14 @@ use constant ITEMS => [ qw(
 	HalfCreditProb
 	FullCreditProb
 	ReducedCred
+	NoReducedCred
 	ExtendDueDate
+	ExtendReducedDate
 	DoubleSet
 	ResurrectHW
 	Surprise
 	SuperExtendDueDate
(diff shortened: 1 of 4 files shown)


somiaj:
Updated with the suggestions from the meeting. I now correctly don't let the extension to go beyond the due date, and don't list sets whose due date and reduced scoring date are the same. I also added a statement about the reduced scoring date cannot be extended beyond the due date in the items description.

drgrice1 (approved):
This looks good.

somiaj:
I'm working on modifying how achievement items are used, and it would be nice if this PR and the other one modifying AchievementItems were merged before I get to far into the updates, since all AchievementItems will need updated.

PR #2526, opened 2024-08-18
allow users to manage their first/last name from Account Settings

Alex-Jordan:
This is mostly so that a user can go to Account Settings and change their first and last name. That is, if they have the new permission allowing them to do that.

This also now prints the effective user ID in the page title, to be clear which user is being altered. The individual items already said the first and last name of the user being edited, but now that we could be editing first and last names, it seemed appropriate to put the effective user ID front and center.

A few cosmetic changes:

- remove "Change" from all the sections of the Course Config page. It seems unnecessary and would now be even more repetitive than it already is.
- change the button to use "Save" instead of "Change"
- make the read-only fields visually appear as plain text 

Overall that page could probably look a little better with some layout design work. But that is for another time.

This may need changes, so I'm not opening a parallel hotfix for now.

### DIFF
--- a/templates/ContentGenerator/Options.html.ep
+++ b/templates/ContentGenerator/Options.html.ep
@@ -18,18 +18,56 @@
 %= form_for current_route, method => 'POST', begin
 	<%= $c->hidden_authen_fields =%>
 	%
+	% if ($authz->hasPermissions($userID, 'change_name')) {
+		<h2><%= maketext('Name') %></h2>
+		<div class="row mb-2">
+			<div class="col-lg-8 col-md-10">
+				<div class="row mb-2">
+					<%= label_for currFirstName => maketext("Current First Name"),
+						class => 'col-form-label col-sm-6' =%>
+					<div class="col-sm-6">
+						<%= text_field currFirstName => $c->{effectiveUser}->first_name, readonly => undef,
+							id => 'currFirstName', class => 'form-control bg-light', dir => 'ltr' =%>
+					</div>
+				</div>
+				<div class="row mb-2">
+					<%= label_for currLastName => maketext("Current Last Name"),
+						class => 'col-form-label col-sm-6' =%>
+					<div class="col-sm-6">
+						<%= text_field currLastName => $c->{effectiveUser}->last_name, readonly => undef,
+							id => 'currLastName', class => 'form-control bg-light', dir => 'ltr' =%>
+					</div>
+				</div>
+				<div class="row mb-2">
+					<%= label_for newFirstName => maketext("New First Name"),
+						class => 'col-form-label col-sm-6' =%>
+					<div class="col-sm-6">
+						<%= text_field newFirstName => '', id => 'newFirstName', class => 'form-control', dir => 'ltr' =%>
+					</div>
+				</div>
+				<div class="row mb-2">
+					<%= label_for newLastName => maketext("New Last Name"),
+						class => 'col-form-label col-sm-6' =%>
+					<div class="col-sm-6">
+						<%= text_field newLastName => '', id => 'newLastName', class => 'form-control', dir => 'ltr' =%>
+					</div>
+				</div>
+			</div>
+		</div>
+	% }
+	%
 	% if ($authz->hasPermissions($userID, 'change_password')) {
-		<h2><%= maketext('Change Password') %></h2>
+		<h2><%= maketext('Password') %></h2>
 		<div class="row mb-2">
 			<div class="col-lg-8 col-md-10">
 				<div class="row mb-2">
 					<%= label_for 'currPassword', class => 'col-form-label col-sm-6', begin =%>
 						<%= maketext(
-							q{[_1]'s Current Password},
-							$c->{user}->first_name . ' ' . $c->{user}->last_name
+							q{[_1]'s Current Password}, $eUserName
 						) =%>
 					<% end =%>
 					<div class="col-sm-6">
+						<%= text_field dummyUsername => '', class => 'd-none' =%>
 						<%= password_field 'currPassword', id => 'currPassword', class => 'form-control', dir => 'ltr',
 							$c->{has_password} ? () : (disabled => 1) =%>
 					</div>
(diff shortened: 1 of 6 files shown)


drgrice1:
I think that this sort of thing is starting to go beyond hotfix territory.  This isn't a bug, but a new feature really.

Alex-Jordan:
That sounds good. I was wondering when/where we would draw a line.

somiaj:
I do see you mention the issue with this and the LTI updating the users name. Would it be worthwhile to just disable the ability to change a user's name if `$LMSManageUserData` is set vs making instructors configure things properly?

Alex-Jordan:
`$LMSManageUserData` might be set, but the course could not be using LTI at all.

somiaj:
If it isn't fully possible to determine if users are using an LMS, then it is fine to let instructors deal with issues that arise of a student changing their name only to be reset to the LMS.

Alex-Jordan:
Personally I feel that `$LMSManageUserData` should be off by default (but it is on by default). If something changes with a user's metadata in the LMS, then immediately changing that in the WW course without notifying the instructor seems like it could cause confusion. So I have it turned off for PCC. Maybe that informs my opinion on this too much.

dlglin:
I personally want student metadata to stay in sync with the LMS, so I would leave `$LMSManageUserData` on regardless of the default.

It might also be good to add a note in the LTI config reminding people that by default users can change their email address in WW, so they may want to review the permissions for changing email (and now name).

Alex-Jordan:
This week there was a Runestone student whose last name in the LMS has a
"wide character" (although in this case, I suspect it's either been
corrupted or it was a data entry error). Having the LMS enforce that would
lead to seeing errors about a wide character. We changed their name in
WeBWorK to make the error clear up. (Of course, maybe WeBWorK should be ok
with that kind of name though.)

Alex-Jordan:
I updated this so that the help test for `$LMSManageUserData` explains that you should review the permissions for changing one's name and email address.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -693,6 +693,11 @@ sub page_title ($c) {
 		return $courseTitle if defined $courseTitle && $courseTitle ne '';
 	}
 
+	# If the current route name is 'options' then show the effective user id alongside the regular title.
+	if ($c->current_route eq 'options') {
+		return $c->maketext("[_1] for [_2]", route_title($c, $c->current_route, 1), $c->param('effectiveUser'));
+	}
+
```
drgrice1, line 700: This shouldn't be done here.  Instead add
```
sub page_title ($c) {
	return $c->maketext("Account Settings for [_1]", $c->param('effectiveUser'));
}
```
to `lib/WeBWorK/ContentGenerator/Options.pm`.  When a ContentGenerator sub module needs special handling for a `ContentGenerator` method like `page_title` this is how it should be done.  This is consistent with how the other modules do it except for the `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetList.pm` module.  That module really should also do it this way, but it has always been this way.

templates/ContentGenerator/Options.html.ep
```diff
@@ -54,15 +91,15 @@
...
 		<div class="row mb-2">
 			<div class="col-lg-8 col-md-10">
 				<div class="row mb-2">
 					<%= label_for currAddress => maketext("[_1]'s Current Address", $eUserName),
 						class => 'col-form-label col-sm-6' =%>
 					<div class="col-sm-6">
-						<%= text_field currAddress => $c->{effectiveUser}->email_address,
-							readonly => undef, id => 'currAddress', class => 'form-control', dir => 'ltr' =%>
+						<%= text_field currAddress => $c->{effectiveUser}->email_address, readonly => undef,
+							id => 'currAddress', class => 'form-control-plaintext', dir => 'ltr' =%>
```
drgrice1, line 102: I guess I am okay with changing from `form-control` to `form-control-plaintext`, but I personally prefer the appearance with `form-control`.  It just seems like things align better that way and so it looks better to me.

templates/ContentGenerator/Options.html.ep
```diff
@@ -18,15 +18,52 @@
...
 			<div class="col-lg-8 col-md-10">
 				<div class="row mb-2">
 					<%= label_for 'currPassword', class => 'col-form-label col-sm-6', begin =%>
 						<%= maketext(
-							q{[_1]'s Current Password},
-							$c->{user}->first_name . ' ' . $c->{user}->last_name
+							q{[_1]'s Current Password}, $eUserName
 						) =%>
 					<% end =%>
 					<div class="col-sm-6">
```
drgrice1, line 69: Add `<%= text_field dummyUsername => '', class => 'd-none' =%>` here, and also add `autocomplete => 'new-password' to both the `newPassword` and `confirmPasssword` inputs below.  This is to fix a browser auto fill issue that I think @somiaj mentioned, and that I am seeing too.

The dummy not displayed input fixes the issue in Firefox.  Anytime Firefox sees a password input, it looks for a text input preceding it in the DOM.  If it finds that, then it assumes that input is the username input, and autofills it with the username, and the password field with the password for the site.  Since the `newLastName` field is the preceding input, that gets filled with the username.

The `autocomplete => 'new-password` attribute fixes the issue for Google Chrome (and probably other browsers too -- not sure why Firefox doesn't honor this?).  Chrome basically does the same thing and assumes that the text input preceding a password field is the username.   However, Chrome ignores auto filling a password field (and preceding input) if it has the `autocomplete => 'new-password'` attribute.

This wasn't an issue without this pull request simply because there was no text input preceding the password fields.

drgrice1 (commented):
I think this generally looks good other than changing how the custom options `page_title` is done.

templates/ContentGenerator/Options.html.ep
Alex-Jordan: I changed it back, but also applied `bg-light` to the readonly inputs. So it aligns like before, but there is still a visual indication that something is different about the usability of these ones.

Alex-Jordan:
OK, I made the changes described here.

The name where that happened had some sort of character modification character in it. IIRC, it was a character that places a small vertical line under the preceding character. So some kind of "accent" character. It was clearly wrong, as the student's name in other places had a regular ASCII letter in place of that special character.

I don't have the details written down though, this was over a Zoom session with a Runestone subscriber.

The original reason that we prefer `$LMSManageUserData` to be 0 at PCC, is that some faculty use the section and recitation fields in special ways. They manually put the students in groups using these fields. And we found that when `$LMSManageUserData` is 1, the instructor's section/recitation entries would get obliterated.

drgrice1 (approved):

dlglin:
> The original reason that we prefer `$LMSManageUserData` to be 0 at PCC, is that some faculty use the section and recitation fields in special ways. They manually put the students in groups using these fields. And we found that when `$LMSManageUserData` is 1, the instructor's section/recitation entries would get obliterated.

Is it worth adding functionality to customize which user data fields are managed by the LMS?  Perhaps have `$LMSManageUserData` accept an array of fields which should be updated from data in the LMS, and any fields not in the array are left alone?

somiaj (approved):

PR #2528, opened 2024-08-18
fix some validation issues on the achievements page

Alex-Jordan:
This fixes some HTML validation errors I happened to see in the Achievements page.

* some stray accidental mojolicious control characters coming out as text in the raw HTML
* the id for the item that extends a date by 24 hours is the same as that for the one that extends by 48 hours. So if both are on the page, there is a duplicate HTML id. So this changes one of those item's id.

I noticed this while looking at that page to try and make it more clear that an item button applies to the item described above the button, not the item below the button. The current spacing makes it hard to tell:

 
Screenshot 2024-08-18 at 11 21 52 AM

So as long as anyone is looking at this page, I will ask:  is there any objection to me structuring this differently? Currently each item is an `h3`, but I am wondering if a `dl` structure to this list would be more semantically correct. Either way, I will do something to clarify which item a given button applies to.

### DIFF
--- a/lib/WeBWorK/AchievementItems/SuperExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/SuperExtendDueDate.pm
@@ -44,7 +44,7 @@ sub print_form ($self, $sets, $setProblemIds, $c) {
 		$c->tag('p', $c->maketext('Choose the set whose close date you would like to extend.')),
 		WeBWorK::AchievementItems::form_popup_menu_row(
 			$c,
-			id         => 'ext_set_id',
+			id         => 'super_ext_set_id',
 			label_text => $c->maketext('Set Name'),
 			values     => \@openSets,
 			menu_attr  => { dir => 'ltr' }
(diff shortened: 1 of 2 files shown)


somiaj:
What if the button said `Use Scroll of Resurrection` (item name) instead of `Use Reward`? This would probably help a bit to link the button to its action?

Alex-Jordan:
Yes, that's one of the things I have on a branch. On that branch, I only changed the button text in that way and slightly adjusted the vertical spacing. But then it occurred to me that maybe this list of item titles followed by descriptions would be better built as a `dl` with `dt` and `dd`. Even then, labeling the button more fully might be better.

But that will be handled separately. Did you take a look at the small corrections that are here now? That's all this PR will be doing.

somiaj:
Those look good (and even made me update my PR adding new items to set the drop down menu id correctly). I'm okay with changing away from headers. I'd be interested in using what you come up with.

somiaj (approved):
Looks good.

drgrice1 (approved):
Good finds.

drgrice1:
If you have good restructuring ideas, then do it!

somiaj:
These changes are simple and look ready to go, though I'll wait to merge in case a hotfix is wanted for this (seems reasonable to me).

PR #2530, opened 2024-08-18
Cap the computed unreduced score at 1.

drgrice1:
If an instructor increases a problem score using a problem grader, or an achievement item is used that increases the score, then the computed unreduced score can be greater than 1.  This causes issues when that computed score is used at various places in the code (for example the progress bar).  So this caps the computed score at 1.

This is an alternate and better solution than #2521 and #2525.


somiaj (approved):
Thanks. Fixes the issue.

Alex-Jordan (approved):

PR #2532, opened 2024-08-19
correct the confirmation message when selecting sets for export

Alex-Jordan:
In the Sets Manager, when you select some sets and click to export, that does not actually export those sets. It takes you to a page where you must confirm the selected sets first. So the message here saying "Selected sets were exported." is not correct. This changes that message.


drgrice1 (approved):
I remember being quite confused by this when I first started using webwork.  I would select files for export and click on the old "Take Action" button, and yet nothing was exported even though the message said sets were exported.  Eventually I figured it out.  I forgot about that confusion, and never fixed it.  This is good.

somiaj (approved):
Looks good.

Should we be cautious of hotfixes for language changes, since this could take stuff that was translate and break it? Or maybe we should do a language hotfix a bit later if translators have time to catch up?

Issue #2536, opened 2024-08-21
Error creating admin course


miralgj:
On a fresh new install of 2.19 using MySQL Community Server v8.4, when attempting to create the admin course using the command `$WEBWORK_ROOT/bin/addcourse admin --db-layout=sql_single --users=$APP_ROOT/courses/adminClasslist.lst --professors=admin` I am getting the following error:

```
DBD::mysql::db do failed: BLOB, TEXT, GEOMETRY or JSON column 'session' can't have a default value at /opt/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm line 930.
```

I knew the installation instructions for 2.19 state that MariaDB is recommended but it also states that MySQL is ok to use. Is this error due to an incompatible database?

drgrice1:
What version of MySQL is that?  Also, are you using the `DBD::mysql` or `DBD::MariaDB` perl database driver module?

I believe that recent versions of MySQL do support default values for TEXT columns, but may be mistaken.  However, the `DBD::mysql` driver may not.  You could try the `DBD:MariaDB` driver instead.

If MySQL still does not support default values for TEXT columns, then perhaps it is time to say that we stop supporting use of MySQL.

dlglin:
From the MySQL docs (https://dev.mysql.com/doc/refman/8.4/en/data-type-defaults.html):
The [BLOB](https://dev.mysql.com/doc/refman/8.4/en/blob.html), [TEXT](https://dev.mysql.com/doc/refman/8.4/en/blob.html), GEOMETRY, and [JSON](https://dev.mysql.com/doc/refman/8.4/en/json.html) data types can be assigned a default value only if the value is written as an expression, even if the expression value is a literal:

This is permitted (literal default specified as expression):
```
CREATE TABLE t2 (b BLOB DEFAULT ('abc'));
```
This produces an error (literal default not specified as expression):
```
CREATE TABLE t2 (b BLOB DEFAULT 'abc');
```

drgrice1:
[...]

PR #2538, opened 2024-08-22
Use an "expression" to specify the `session` column default in the `key` table. (hotfix of #2537)

drgrice1:
MySQL requires that the default value for a `TEXT` column be written as an expression.  See https://dev.mysql.com/doc/refman/8.4/en/data-type-defaults.html. MariaDB apparently works fine without this, but works fine with it as well.

This fixes issue #2536.

### DIFF
--- a/lib/WeBWorK/DB/Record/Key.pm
+++ b/lib/WeBWorK/DB/Record/Key.pm
@@ -30,7 +30,7 @@ BEGIN {
 		user_id   => { type => "VARCHAR(100) NOT NULL", key => 1 },
 		key       => { type => "TEXT" },
 		timestamp => { type => "BIGINT" },
-		session   => { type => "TEXT NOT NULL DEFAULT '{}'" },
+		session   => { type => "TEXT NOT NULL DEFAULT ('{}')" },
 	);
 }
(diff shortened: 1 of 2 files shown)


somiaj (approved):
I created a new course without issue with this, so works on MariaDB 10.11.6 here. Anything else to test?

dlglin:
On mysql 8.0 I can create the admin course, but when I try to visit the newly created course I get

DBD::mysql::db do failed: BLOB, TEXT, GEOMETRY or JSON column 'data' can't have a default value at /opt/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm line 930.

In `LTILaunchData.pm` the `data` field is also of type "TEXT NOT NULL DEFAULT '{}'".  Making the same change to that entry fixed things for me.

drgrice1:
Good catch.  I added that to this pull request (and #2537).  I also checked all of the other database tables to ensure this wasn't done anywhere else, but there weren't any other cases.

dlglin (approved):
Looks good now.

somiaj (approved):

PR #2539, opened 2024-08-24
Update webwork2.sty for older LaTeX versions

maharper:
The definition of `\webworkSetCopyRightFooter` in webwork2.sty breaks in pre-November 2021 versions of LaTeX.

This modifies the definition to work in current and in older versions of LaTeX.

@Alex-Jordan suggested this as a hotfix in the discussion on the [Moodle forum](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8407).


somiaj (dismissed):

Alex-Jordan:
FWIW, it is not an oversight in the original that the link is not
functional. This is from the perspective that a WW hardcopy is only for
print. And also if you have internet access, why look at the PDF instead of
being in the assignment online? And then using \url is just a semantic way
to achieve monospace font, which I think is helpful for translating a URL
on paper to the keyboard.

I am not requesting any changes, just explaining why it was that way.

drgrice1:
I recall discussing that now that you point it out.  So that means it should be changed back from using `\href` to using `\url`.

maharper:
I changed the `\href` back to `\url`.  I had forgotten that I made that change, else I would have pointed it out the change, sorry

I think I noticed that webwork.org was not a working link, as my pdf previewer tries to follow it anyway, ending up at a missing file. I thought I was just fixing the reference.

somiaj (approved):
Thanks for updating openwebwork.org, the output looks the same to me after applying this and I can still confirm the changes work here.

drgrice1 (approved):
Looks good.  Thanks @maharper.

Issue #2543, opened 2024-08-28
check_latex can't find pg.sty


dlglin:
When I run `check_latex` I get "! LaTeX Error: File `pg.sty' not found."

A bit of digging shows that the value of `$ce->{pg}{directories}{assetsTex}` is "/assets/tex", which is not correct.

I believe that the problem is that `$pg_dir` is defined in webwork2.mojolicious.yml (or the .dist version), but when a WW CE is invoked from the command line the mojolicious config is not loaded, so `$pg_dir` is never set, which means that any paths that are built from that will be incorrect.

If I'm right this means that `check_latex` won't work for anyone in 2.19.  Can someone confirm whether it works for them?

Alex-Jordan:
I confirm on the two 2.19 servers that I checked, that I get the same error about `pg.sty`.

Alex-Jordan:
When I check this on my production servers (PCC and RS) `check_latex` fails.  But just now I was testing Glenn's PR to address this. And I found that on my devleopment server, on the develop branch, `check_latex` runs with success. (On the develop branch, not just with Glenn's fix.)

drgrice1:
You probably have the `PG_ROOT` environment variable set.  If you have set to the PG directory, then the script will still succeed, because it finds the PG directory that way rather than the new way of getting it from the configuration file.

drgrice1:
By the way, when @dlglin first posted this issue and I tested it, I saw the same thing.  Then I `unset $PG_ROOT` and tried again.  Then I saw the failure.

somiaj:
This was fixed by #2545 and hotfix #2546.

PR #2545, opened 2024-08-28
Fix the `check_latex` script.

drgrice1:
The addition of the requirement of PG files for hardcopy (namely the `pg.sty` file), means that the `check_latex` script needs to know where the pg directory is.  So find that the same way the webwork2 app does.

This fixes issue #2543.

### DIFF
--- a/bin/check_latex
+++ b/bin/check_latex
@@ -27,9 +27,17 @@ use feature 'say';
 
 BEGIN {
 	use Mojo::File qw(curfile);
-	use Env qw(WEBWORK_ROOT);
+	use YAML::XS qw(LoadFile);
+	use Env qw(WEBWORK_ROOT PG_ROOT);
 
-	$WEBWORK_ROOT = curfile->dirname->dirname;
+	$WEBWORK_ROOT = curfile->dirname->dirname->to_string;
+
+	# Load the configuration file to obtain the PG root directory.
+	my $config_file = "$WEBWORK_ROOT/conf/webwork2.mojolicious.yml";
+	$config_file = "$WEBWORK_ROOT/conf/webwork2.mojolicious.dist.yml" unless -e $config_file;
+	my $config = LoadFile($config_file);
+
+	$PG_ROOT = $config->{pg_dir};
 }
 
 use File::Temp qw(tempdir);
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
On my develop server `check_latex` runs successfully whether I am on the develop branch or this PR branch. I am not sure why. Approving this, but I don't feel like I tested it since I had no change in behavior.

somiaj:
I am also not able to reproduce the error, even if I unset `PG_ROOT` environment variable in develop.

I did notice that the output of the script changed, before it would only output `Compilation Success!`, now it outputs the line `/opt/webwork/pg/assets/tex` (or wherever pg is located). Any reason to have that line outputted without any context?

On a side note the script did fail with `bidi.sty` missing if I didn't have `texlive-lang-arabic`. I didn't notice any issues with not having this, so is this really required or optional? Anyways I'm okay either way, just something I noticed.

somiaj:
note, just because I don't think I was clear, it still outputs `Compilation Success!`, it just also outputs the line of the pg tex directory with this patch applied.

drgrice1:
The output of the pg directory was a debugging line that I added and forgot to remove.  It has been removed now.

drgrice1:
If the `PG_ROOT` environment variable is not set, then I don't know how the `check_latex` script in develop could possibly succeed.

I also have noticed the `bidi` package issue.  That should not be in the `check_latex_exam.tex` file.  The only hardcopy themes that use the `bidi` package are the Hebrew themes.  So the only time you need the package is if you are using those themes.  Since those themes are not enabled by default, we shouldn't check for it.  Perhaps we could check to see if those themes are enabled for the site, and only check that for that package in that case.

somiaj:
I still cannot get `check_latex` to fail in main/develop. I even added your debug line to `check_latex` in develop and I'm getting that `$pg{directories}{assetsTex}` is set to the pg directory correctly from the setting in `defaults.conf`. The only way I can get `check_latex` to fail is to modify `defaults.conf` and make `$pg{directories}{assetsTex}` point to a wrong location.

Further I even tested setting `$PG_ROOT=/bad/dir` (an invalid location), and it has no effect, so in my testing `$PG_ROOT` is not being used by `check_latex`. What could be different with my system that `$PG_ROOT` isn't being used at all here?

somiaj:
Okay, I tracked it down, in `site.conf` I have `$pg_dir = /opt/webwork/pg`, this is no longer is `site.conf.dist`, so for users with older `site.conf` that haven't removed this setting, things can still find the pg dir and ignore `PG_ROOT` altogether. If I remove this setting from `site.conf`, I can reproduce the error if `PG_ROOT` is unset.

Furthermore, the patch fixes the issue.

somiaj (approved):

drgrice1:
Ahh, that makes sense.  If you still have the old setting in `site.conf`, then the course environment would get it from there.

PR #2547, opened 2024-08-30
Improvements to SAML2 authentication.

drgrice1:
This fixes some issues with #2511.  This pull request is built on that one. That pull request is a nice start, but has some issues.  Thanks to @ionparticle for starting this work.  One of the most important things introduced in that pull request is the development identity provider via SimpleSAMLphp.  That gave me a way to work with this.
    
Although I liked the idea of using a Mojolicious plugin, that approach does not fit into the webwork2 course environment system.  So, unfortunately, that had to be changed.  So this rewrites the SAML2 authentication module to use the usual approach that all of the other authentication modules use.  There is a `authen_saml2.conf.dist` file that should be copied to `authen_saml2.conf` to use SAML2 authentication.  All settings for the authentication module are in that file.
    
The primary limitation of the plugin approach was that it is not possible for different courses to use different identity providers.  Or for one course to use SAML2 authentication and another only use the basic password authentication (without the need to add a URL parameter).  This is something that is expected of the authentication modules, and is desirable for multi-institutional servers.
    
Another problem with the implementation is that is does not work with `$session_management_via = "key"` set.  The reason that doesn't work is because the implementation broke the rule of creating and accessing a session before authentication is completed.  So this reimplementation uses the database via the "nonce" key hack much like LTI 1.1 authentication does.
    
The previous implentation also used the `WEBWORK_ROOT_URL` environment variable in the code.  That is not an environment variable that is defined for webwork2.  It is only defined for the docker build, and can not be used in the code.
    
The previous implementation also does not work with two factor authentication.  If an identity provider does not provide two factor authentication, then webwork2's two factor authentication should be used.  Of course, if the identity provider does provide two factor authentication, then the two factor authentication can be disabled in `localOverrides.conf`.
    
An option to enable service provider initiated logout has also been added. It is off by default, but if enabled, when the user clicks the "Log Out" button, a request is sent to the identity provider to also end its session.
    
The `README.md` file that was with the plugin code has been moved to the `docker-config/idp` directory, and is now purely for instructions on how to use the simpleSAMLphp development instance.  The readme also includes instructions on how to set up a simpleSAMLphp development instance without docker.  The documentation on how to configure and use the SAML2 authentication module is now all in the `authen_saml2.conf.dist` file.
    
Note that the webwork2 service provider certificate files are no longer directly in the configuration file. Instead file locations are to be provided.  It didn't really make sense to have the actual certificates in the configuration file, then write them to a file, which the Net::SAML2 module would read.  Furthermore, it would be very easy to add the certificates incorrectly to the configuration file.  With the YAML file approach if indentation were not correct, then the configuration file would fail to load.


drgrice1:
@ionparticle:  Please test this, and ensure that this still meets your needs and still works for you.  Also, if you have any suggestions for improvement or see any changes that are needed, please comment here.

drgrice1:
Also, if you know of anyone else that can test this as well, see if they could try it out also.

lib/WeBWorK/Utils/Routes.pm
```diff
@@ -222,6 +231,32 @@ my %routeParameters = (
 		action => 'content_selection'
 	},
 
+	# This route also ends up at the login screen on failure, and the title is not used anywhere else.
+	saml2_acs => {
```
ionparticle, line 235: We're implementing the HTTP-POST binding version of ACS, so ACS must only accept POST requests. $route->any breaks the spec here.

docker-config/idp/certs/saml.crt
```diff
@@ -0,0 +1,29 @@
+-----BEGIN CERTIFICATE-----
```
ionparticle, line 1: As these are SP certs and key, they probably shouldn't be under idp/ to reduce confusion.

docker-config/docker-compose.dist.yml
```diff
@@ -252,6 +251,28 @@ services:
     #ports:
     #  - "6311:6311"
 
+  # SimpleSAMLphp Saml2 identity provider for development use only. This is a separate profile from the other services
+  # so it doesn't start in normal usage. Use "docker compose --profile saml2dev up" to start, and "docker compose
+  # --profile saml2dev down" to stop.
```
ionparticle, line 256: Docker compose `stop` should be used instead of `down`. `down` stops *and* removes the containers, it is usually used when you want to blow things away for a fresh start.

docker-config/idp/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+Stop all docker services with
+
+```bash
+docker compose --profile saml2dev down
+```
+
+## Development IdP test instance without docker
+
+Effective development is not done with docker. So it is usually more useful to
```
ionparticle, line 46: Containerized development environments are pretty common nowadays.

ionparticle (commented):
Hey @drgrice1,

Good news is everything appears to work, I also tested the newly implemented SLO and that works too!

Unfortunately, the user flow used by us is no longer possible due to conflicting configuration though:

1. SAML2 login does NOT require internal MFA
2. Internal login can only be accessed with the bypass query.
3. Internal login IS protected by internal MFA

Setting `$external_auth = 1` in `authen_saml2.conf` satisfies 1 & 3, but breaks 2.
Disabling internal MFA satisfies 1 & 2 but breaks 3.

I recommend removing the internal MFA support for SAML2. It'll make the system a bit simpler and external SAML2 systems are expected to take responsibility for additional security anyways. They normally have much more robust and secure MFA implementations than anything we can do.

This and ACS POST are the only breaking changes for me.

lib/WeBWorK/Utils/Routes.pm
drgrice1: `$route->any` still accepts POST requests.  So it doesn't really break anything of any real significance.

docker-config/idp/certs/saml.crt
drgrice1: There really isn't a better place for these files.  They are really only used for the docker build, and it is best to keep all of the files for that together.

docker-config/docker-compose.dist.yml
drgrice1: When working with a development build, you usually need to do exactly what `down` does here.

docker-config/idp/README.md
drgrice1: No, not really.  Furthermore, they are not productive.

drgrice1: The problem with working in a container is the extra effort required to modify code, and really test things.  I have never seen a good development system in a container to be quite frank.  No one does it.

drgrice1:
The internal MFA can't be removed.  Just disable it if you don't want it.  Not all identity providers offer two factor authentication.  In that case, the two factor authentication provided by webwork should be used.  SAML2 systems that do not offer two factor authentication are no more secure than any username/password system.  If you are seeing an issue with it, then it needs to be fixed.  I will look into what can be done about the setting that you describe.

Your claim that SAML2 systems offer more robust and secure MFA implementations than anything we can do is simply false.  You don't know what we are capable of, nor do you really know what the SAML2 systems are capable of.

docker-config/docker-compose.dist.yml
drgrice1: The point here is that most of use are only going to be using this for testing.  So we want a quick up and down.  We don't want the container left hanging around.  This is not for production use.

lib/WeBWorK/Utils/Routes.pm
drgrice1: I will make it so that the ACS route only accepts POST requests.

drgrice1: This is done.

docker-config/idp/certs/saml.crt
drgrice1: These files are going to have to stay here.  We don't want to clutter the rest of the webwork directory structure with things that aren't used anywhere else.

ionparticle:
Sorry, I think you're taking this the wrong way, no insult on capability is meant. It's a statement regarding resources. These SAML2 implementations (from Google, Microsoft, Shibboleth, etc) have had far more resources and scrutiny invested in them.

drgrice1:
Yeah, your original statement was not a comment on resources.  Notwithstanding, don't put to much faith in such things.  They are not as reliable as you might think.

drgrice1:
There is now another Saml2 authentication variable `$saml2{twoFAOnlyWithBypass}`.  This option is for the case that the identity provider offers multi factor authentication, and yet the $saml2{bypass_query} is also allowed.  In this case you would not want webwork2's two factor authentication to be used when signing in via the identity provider.  However, two factor authentication should be used if the bypass query is used.  Setting $saml2{twoFAOnlyWithBypass} to 1 makes it so that webwork2's two factor authentication is skipped for users signing in via the identity provider, but still required for users signing in with a username/password. If this is set to 0, then webwork2's two factor authentication will always be required.

drgrice1:
@ionparticle:  The new `$saml2{twoFAOnlyWithBypass}` option address the setup that you gave.

ionparticle:
thank you very much for both fixes, I've tested and confirms that this will meet our needs.

On another note, we would like to add LTI 1.3 Names and Roles Provisioning Service for classlist sync, how would you like to see this done architecturally?

docker-config/docker-compose.dist.yml
drgrice1: I added to the comment about using both "stop" and "down" now.

drgrice1:
I am not sure.  I would have to look at the spec to get a better idea. If you have a start on implementing it, you could create a draft pull request for advice.

Issue #2550, opened 2024-08-30
The Problem Sets (landing) page looks incomplete if there are no sets assigned to a user


dlglin:
If there are no sets assigned to a user then the Problem Sets page looks incomplete.  It's not clear to me that there are no sets available, or whether I have to navigate somewhere else to find my homework.

Perhaps in this case we could display something like "There are no sets available to you at this time".

somiaj:
This was addressed by #2551 and hotfix #2555.

PR #2551, opened 2024-08-31
Add message for users with no visible sets.

somiaj:
Show a message for users viewing the "Assignments" page when they have no visible sets. This also hides the show by buttons when no sets are visible.

This addresses #2550.

### DIFF
--- a/templates/ContentGenerator/ProblemSets.html.ep
+++ b/templates/ContentGenerator/ProblemSets.html.ep
@@ -22,34 +22,40 @@
 % # Regular sets and gateway template sets are merged, but sorted either by name or urgency.
 % # Versions are not shown here. Instead they are on the ProblemSet page for the gateway quiz.
 %
-<div class="d-flex justify-content-end gap-2 mb-3">
-	<%= hidden_field courseID => $ce->{courseName} =%>
-	<%= hidden_field userName => param('user') =%>
-	<div class="btn-group" role="group" aria-label="<%= maketext('Display choice') %>">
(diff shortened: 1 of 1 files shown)


templates/ContentGenerator/ProblemSets.html.ep
```diff
@@ -21,35 +21,41 @@
...
-		   	>
-			<%= maketext('Show By Date') =%>
-		</button>
-		<button id="show-by-type-btn" type="button" class="btn btn-outline-primary"
-			data-default-title="<%= maketext('Regular Assignments') %>"
-			data-test-title="<%= maketext('Tests/Quizzes') %>"
-			>
-			<%= maketext('Show By Type') =%>
-		</button>
+% #
```
drgrice1, line 24: There is no comment here so the `#` is not needed.

drgrice1 (approved):
Looks good.

templates/ContentGenerator/ProblemSets.html.ep
somiaj: Thanks, I was going to put a comment about only showing the sets when they exist, but changed my mind, though the if statement was clear enough.

Alex-Jordan (approved):

Alex-Jordan:
Perhaps this should have a hotfix variant. It is addressing a shortcoming of a feature that was introduced with 2.19, not really adding a new feature.

drgrice1:
A hot fix would be fine for this.

Issue #2552, opened 2024-09-01
Feature request: Customize date format separately from the language


szhorvat:
This is a feature request to allow customizing the date format that instructors use when setting deadlines, independently of the WeBWorK language.

Currently this date format cannot be customized independently of the language. For as long as WeBWorK is set to English, it will use the US format, i.e. month/day/year. This format is almost unheard of outside of North America, and will be confusing for many people. I certainly find it confusing to have to mentally switch to it, and it caused me to set incorrect dates multiple times in the past.

Note that this refers to how we set dates in the WeBWorK interface, not how dates are displayed to students.

(This is a distilled down version of #2204.)

drgrice1:
I can not approve this feature request.  It is incorrect (and would be challenging and annoying to implement) to have the date format not match the locale that is set.

However, one thing that could be done is to add other English language variants.  For example, a UK variant could (and should) be added.  That would have the date/time format that you desire.

szhorvat:
I respectfully, and _very strongly_ disagree, not only because of obvious practicality of the situation, but also because of my experience of being a speaker/member of a minority language/culture in my birth country. Treating the concept of "locale" in an inflexible way as you are here very much reminds me of being forced into rigid systems designed without an understanding of minority needs. On modern computer systems, languages, date formats, number formats, calendar systems, and other "localization" aspects are independently adjustable. In many places, there are multiple of these used at the same time. For example, in many places, both d-m-y and y-m-d are commonly used date format (in fact I use the latter on  my devices).

What I asked for was not British English, or adhering to British customs, but the flexibility to choose the date format independently of the language. Please keep in mind that English is used in higher education in many more countries than just the US and the UK.

Please consider giving this more though before rejecting the request, or consulting with users who are directly affected by this decision.

That's all I have to say on the matter, I'll leave it to you from here.

szhorvat:
First of all, thank you for implementing a practical solution. The way you implemented it (dd/mm/yyyy and HH:mm) addresses my use case perfectly. I assume this will be part of WeBWorK 2.20 next year?

> Your statement that localization of languages, date formats, number formats, etc., be made customizable separately simply isn't going to happen, and is far more rare than you seem to be claiming. Even browsers don't allow such separate customization (neither Firefox nor Google Chrome support that at least).

[...]

PR #2553, opened 2024-09-01
Add an en-GB localization option.

drgrice1:
This adds the Great Britain (or United Kingdom) locale.  Translation isn't really needed in the `.po` file.  However, if someone wants to do so it could be done.  They do have different spellings of some words. The most important thing with this is that this adds the time format for the Great Britain locale which is quite different from the US format.

Note that the `en.po` file was updated as well.  This file hasn't been updated for a while since it really doesn't need to be.

If there are other time formats that are desired, then other en variants could be added as well.

This is to address issue #2552 (although not in the requested way which I do not approve of).


somiaj (approved):
Works for me in my testing various date/time output.

Note to others testing this, make sure to update the javascript side of things too, `npm ci`, if you want the date selectors to work.

pstaabp (approved):
Looks good.  Since there are no translations, mainly tested that the time/date format looks right.

PR #2557, opened 2024-09-03
in Course Config, don't claim database settings are defaults

Alex-Jordan:
IIRC when I added the course title as something you can change in the config page (and "setting" values in general), the "Default" column was left blank for that row. Somewhere along the line things changed so now you see the current course title in the "Default" column. But that doesn't make sense. It's not like you could clear the entry in the "Current" column and it would revert to what it says in the "Default" column.

So this change makes it so that entry is once again empty.

Before:

Screenshot 2024-09-02 at 11 22 39 PM

After:

Screenshot 2024-09-02 at 11 22 07 PM

### DIFF
--- a/lib/WeBWorK/ConfigObject/setting.pm
+++ b/lib/WeBWorK/ConfigObject/setting.pm
@@ -18,7 +18,14 @@ use Mojo::Base 'WeBWorK::ConfigObject', -signatures;
 
 # Configure settings in the course's setting table.
 
-sub get_value ($self, $ce) { return $self->{c}->db->getSettingValue($self->{var}) // '' }
+sub get_value ($self, $ce) {
+	# If the course name of the controller's course environment is the same as the passed course environment, then
+	# return the value from the database.  Otherwise this is the default site wide setting. So return the empty string
+	# (there is no default).
+	return $self->{c}->ce->{courseName} eq $ce->{courseName}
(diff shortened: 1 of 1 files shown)


drgrice1 (changes requested):
I don't like this approach to fixing the issue.  Each config object should be responsible for returning the correct thing for these methods using the override in the config object module.  This approach makes the base module handle it.  If instead the `get_value` method of `lib/WeBWorK/ConfigObject/setting.pm` were changed to
```perl
sub get_value ($self, $ce) {
	# If the course name of the controller's course environment is the same as the passed course environment, then
	# return the value from the database.  Otherwise this is the default site wide setting. So return the empty string
	# (there is no default).
	return $self->{c}->ce->{courseName} eq $ce->{courseName}
		? ($self->{c}->db->getSettingValue($self->{var}) // '')
		: '';
}
```
then the module is handling it by itself.  This works on the principle that the course environment of the controller is always the current course's course environment, while the course environment that is passed is the generic course environment with the courseName '___' in the case that the site default value is requested.  Any other time the `get_value` method is called the courseName's will match.

Alex-Jordan:
I changed this to how @drgrice1 posted. The change is now 100% Glenn's code so I changed the commit author :)

Alex-Jordan:
I can live with this not having a hotfix variant. But if anyone wants that, shout out.

drgrice1 (approved):

Issue #2567, opened 2024-09-09
checkboxes in File Manager get unchecked


Alex-Jordan:
When I enter the File Manager, the two checkboxes controlling behavior of archive file uploads, "Unpack archives automatically" and "then delete them" are checked.

If I delete a file, there is a confirmation page. After confirming and following through with the delete, those checkboxes are no longer checked.

This is just a little strange. And if you aren't looking at those checkboxes because you have internalized the default behavior, you will miss that now the boxes are unchecked.

drgrice1:
This was fixed with #2569.

PR #2569, opened 2024-09-09
Preserve check and radio state when performing a submit action in the File Manager.

drgrice1:
This was an error in the conversion from mod_perl2 to Mojolicious.  The hidden form fields set in the `HiddenFlags` method of `WeBWorK::ContentGenerator::Instructor::FileManager` need to set the value of each field to the paramater (done via the `getFlag` method). That is how it was before the conversion to Mojolicious, but I made a mistake and dropped it in the conversion.

This fixes issue #2567.


Alex-Jordan (approved):
I tested and it fixes #2567.

pstaabp (approved):
Fixes the issue.

PR #2570, opened 2024-09-10
Remove code to strip spaces from pg filenames when imported .def files

### LINKED ISSUE #2563: Importing a set definition file removes spaces from source_file
When importing a set definition file all spaces are removed from the `source_file` field.  This causes issues if a pg file genuinely has spaces in the filename as the imported set will not have the correct `source_file`.

To reproduce:

1. Create a pg file with a space in the filename and add it to a set.
2. Export the set.  The .def file will include the spaces in the filename.
3. Import the set into a course.  The spaces have been removed from the filename, and the Set Detail page will report that the source file can't be found (unless there happens to be a file with same name with no spaces).

This looks to be done deliberately:
https://github.com/openwebwork/webwork2/blob/647d317f126c4295bc9ef05865598091191f90e4/lib/WeBWorK/File/SetDef.pm#L565
In earlier versions of WW there were problems with including spaces in filenames, but as far as I can tell everything else now works with them.  Can this line be deleted, or might that break something else?

dlglin:
This is to address #2563.  I have tested with def files exported from 2.17.

I haven't checked if this works on the old format for .def files, but they haven't been used in many years.

### DIFF
--- a/lib/WeBWorK/File/SetDef.pm
+++ b/lib/WeBWorK/File/SetDef.pm
@@ -562,7 +562,6 @@ sub readSetDef ($ce, $fileName) {
 					$problemData->{$item} = $value if defined $value;
 				} elsif ($item eq 'problem_end') {
 					# Clean up and validate values
-					$problemData->{source_file} =~ s/\s*//g;
 					push(@errors, [ 'No source_file for problem in "[_1]"', $fileName ])
 						unless $problemData->{source_file};
 


Alex-Jordan:
This works, and I did testing of other things with spaces in a PG filename and found no issues. But I have two comments:

- Should you remove the comment right above this line too? I think that comment applies only to this line that is being removed.
- I haven't tried a version 1 set definition file, but as I read the code, it seems that removing lines 521-523 would be the corresponding change.

drgrice1:
@Alex-Jordan:  I think that the comment applies to the other things that are validated and cleaned up as well on the lines that follow.

As to doing the same for version 1 set definition files, if that is done, then should the file name still be trimmed?  That is leading and trailing spaces removed?  That is done for version 2 set definition files on lines 552 and 553.

Although, perhaps it is time to think about removing support for version 1 set definition files.  The only version 1 set definition files that I know of have now been removed from the OPL.

pstaabp (approved):
I think this looks good.

somiaj (approved):

PR #2571, opened 2024-09-10
Fix the `$webworkDirs{valid_symlinks}` course environment option.

drgrice1:
This was broken in the recent rewrite of the file manager.  Any symlink to a directory listed in the `$webworkDirs{valid_symlinks}` array should be allowed to be followed.  This is a quick fix that makes those symlinks look and behave like they are actually directories in the file manager.


Alex-Jordan (approved):

pstaabp (approved):

somiaj (approved):
Should this be a hotfix too?

drgrice1:
I don't really think that this environment variable is used much.  So a hotfix is probably not necessary.

Issue #2572, opened 2024-09-18
setting Default Time that the Assignment is Due leads to an error


Alex-Jordan:
In a course, go to Course Config, and set Default Time that the Assignment is Due to '12:30pm', and save. Then try to create a new set, either in the Sets Manager or Library Browser. This leads to an error:

```
Validation failed for type named Hour declared in package DateTime::Types (/usr/lib/x86_64-linux-gnu/perl5/5.34/DateTime/Types.pm) at line 101 in sub named (eval) with value 24
```

It sounds like something is converting the 12 (pm) to 24, and I'm guessing it should be 12 instead. And I'm not sure if 12 (am) is handled right, but that should be examined as well.

somiaj:
I can confirm the issue here and the following fixes it for me.

```diff
diff --git a/lib/WeBWorK/Utils/DateTime.pm b/lib/WeBWorK/Utils/DateTime.pm
index f1e13977f..3b055f917 100644
--- a/lib/WeBWorK/Utils/DateTime.pm
+++ b/lib/WeBWorK/Utils/DateTime.pm
@@ -59,7 +59,7 @@ sub getDefaultSetDueDate ($ce) {
        my ($hour, $minute, $ampm) = $ce->{pg}{timeAssignDue} =~ m/\s*(\d+)\s*:\s*(\d+)\s*(am|pm|AM|PM)?\s*/;
        $hour   //= 0;
        $minute //= 0;
-       $hour += 12 if $ampm && $ampm =~ m/pm|PM/;
+       $hour += 12 if $ampm && $ampm =~ m/pm|PM/ && $hour != 12;
+       $hour = 0 if $ampm && $ampm =~ m/am|AM/ && $hour == 12;

        my $dt = DateTime->from_epoch(epoch => time + 2 * 60 * 60 * 24 * 7);
```

pstaabp:
Fixed in #2573 and #2576

PR #2573, opened 2024-09-18
Fix an issue with converting AM/PM time to 24 hour time.

somiaj:
The logic to convert to 24 hour time didn't consider 12am and 12pm correctly, this fixes that.

This fixes issue #2572


dlglin (approved):
Fixes the problem

Issue #2556, opened 2024-09-03
Problem editor cannot find images when previewing/checking/showing answers.


somiaj:
The image is shown just fine when initially rendered. But when previewing, checking, or showing the correct answer, the image is not found and a warning is issued.  Here is a test problem.

```
DOCUMENT();      
loadMacros(qw(PGstandard.pl PGML.pl PGcourse.pl));

BEGIN_PGML
    [!Alt Text!]{'imagename.png'}

END_PGML
ENDDOCUMENT();
```

The warnings issued are:

+ Unable to find file "imagename.png".
+ Unable to form a URL for the auxiliary file "imagename.png" used in this problem.

somiaj:
@drgrice1 I have a possible fix to this, but I have a feeling it might not be correct. Check out the above commit as I'm unsure if there is a case where we would want the `$problemRecord->source_file` to be empty. The problem appears to be that the hidden field `fileName` is set to an empty string and `PGalias` is no longer able to find images relative to the location of the file being edited.

I have tested the various places problems are rendered (PGProblemEditor, Library browser, set editor, and stats for a single problem) and haven't noticed any issues with this.

I think it would be better if the rendered problem in the PGProblemEditor set the `fileName` hidden field to the temporary file name, but I wasn't able to track down where to do that.

drgrice1:
I believe this is fixed. @somiaj: Please confirm, and close this if it is.

somiaj:
It is fixed, I've been using the fix for a while. I'll close now that it is merged.

PR #2574, opened 2024-09-18
Fix rendering static images when checking answers in the PG problem editor.

drgrice1:
This fixes issue #2556 essentially as @somiaj suggests (although there is no need to first check that `$rh->{fileName}` is defined).

The reason that `$rh->{fileName}` is defined is because the `default.html.ep` template always adds the `fileName` hidden field even when the `fileName` parameter is not defined in the original request. So if you click "Check Answers" it will be defined in `RenderProblem.pm` on that request.

### DIFF
--- a/lib/WebworkWebservice/RenderProblem.pm
+++ b/lib/WebworkWebservice/RenderProblem.pm
@@ -189,13 +189,13 @@ async sub renderProblem {
 	my $r_problem_source;
 	if ($rh->{problemSource}) {
 		$r_problem_source = \(decode_utf8_base64($rh->{problemSource}) =~ tr/\r/\n/r);
-		$problemRecord->source_file(defined $rh->{fileName} ? $rh->{fileName} : $rh->{sourceFilePath});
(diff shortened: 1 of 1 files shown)


somiaj (approved):
Thanks for looking into this and double checking there wasn't a better fix.

Could this also be a hotfix?

Alex-Jordan:
I haven't followed this issue, have not tried to understand the relevant code, and have not tested anything :)

Just purely from looking at the code change from boolean testing `$rh->{fileName}` to `defined $rh->{fileName}`, is it never the case the the `$rh->{fileName}` is `'0'`? Sounds like a bad choice for a filename, but I try not to judge.

And is it never the case that `$rh->{fileName}` is defined, but has somehow wound up being empty?

drgrice1:
A filename of `0` also does not work in assignments.  So I don't think it needs to (or even should be) supported here.  Note that a filename of `0.pg` does work in assignments and will work here.

pstaabp (approved):

PR #2579, opened 2024-09-24
Add leaderboard for achievements.

somiaj:
This adds a leaderboard for achievements, which ranks users from the greatest to the least number of achievement points along with showing the badges of all earned achievements.

The default use of this is to provide a summary page for professors to see how many achievement points students have earned along with which badges they have earned. The default permission level to view the leaderboard and see usernames on the leaderboard is professor.

The permission level for viewing the leaderboard and viewing names on the leaderboard can be changed under course configuration to allow students to see the leaderboard. It is noted that since achievement points are often closely related to grades, that this should be considered before allowing students access.

Note, this is currently a little slow since it loops over all students, and for each student loops over all achievements making multiple database calls to get the user achievement records. In my class of 40 students with 80 achievements, it takes about 5 seconds to generate the page, and could be slower for much larger classes. I would like to speed this up, but unsure on how to more efficiently make database calls to get all the achievement data needed to build the leaderboard.


lib/WeBWorK/ContentGenerator/Leaderboard.pm
```diff
@@ -0,0 +1,101 @@
...
+
+		my @badges;
+		for my $badge (@allBadges) {
+			# Skip level achievements and only show earned achievements.
+			last if $badge->category eq 'level';
+			next unless $db->existsUserAchievement($user->user_id, $badge->achievement_id);
+
+			my $userBadge = $db->getUserAchievement($user->user_id, $badge->achievement_id);
+			push(@badges, $badge) if $badge->enabled && $userBadge->earned;
+		}
```
drgrice1, line 70: There is no need to call `existsUserAchievement` and then `getUserAchievement`.  In fact it is more expensive to do that since it takes two database queries, than to just call `getUserAchievement` and check if that is truthy which means that the achievement exists.

So change this as suggested below.  This also incorporates the other suggested variable changes.  I left the local `@badges` for now, but those are also not badges.  They are user achievement database records.

Note that this still leaves the `getUserAchievement` database call in the for loop.  You could get all user achievements before the for loop and then just look them up here, but there is a trade off in this case that didn't occur in the others.  That is that in this case you don't actually need all of the user achievements.  You only need those that aren't of the 'level' category.  So there is the trade off of loading unneeded data into memory versus the inefficiency of querying the database for every user and non-level achievement.  It would be possible to go to a lower level in the database layer and use a join on the achievement and achievement_user tables to accomplish this, but that is messy to do here and I guess it is okay to leave this one database query repeated in the loop for now.  It is still expensive to do this though.

```suggestion
		for my $achievement (@achievements) {
			# Skip level achievements and only show earned achievements.
			last if $achievement->category eq 'level';

			my $userBadge = $db->getUserAchievement($user->user_id, $achievement->achievement_id);
			next unless $userBadge;
			push(@badges, $achievement) if $achievement->enabled && $userBadge->earned;
		}
```

drgrice1 (commented):
There are several changes that are needed here.  I also suggested some changes that should improve loading the needed data from the database some.

drgrice1:
Also, to take the "Leaderboard" to "Achievement Leaderboard" comments further, I think you should rename `Leaderboard.pm` and both `Leaderboard.html.ep` files to start with "Achievement".  This makes it clear that these are associated with achievements for developers.  Who knows, there may be some other kind of leaderboard created for something else in the future.

drgrice1:
Yeah, I did note that "Achievements Leaderboard" does become the longest name in the site navigation.  But it seems to still fit okay.

lib/WeBWorK/ContentGenerator/Leaderboard.pm
```diff
@@ -0,0 +1,101 @@
...
+WeBWorK::ContentGenerator::Leaderboard - Leaderboard for achievements,
+which lists the total number of achievement points, level, and badges
+earned for each user with the 'include_in_stats' status.
+
+Only users with the 'view_leaderboard' permission can see the Leaderboard, and only
+users with the 'view_leaderboard_usernames' permission can see user names.
+
+=cut
+
+use WeBWorK::Utils qw(sortAchievements);
```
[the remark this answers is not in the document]
somiaj: This was used on line 46 below, I think you found it, but might have not removed this comment.

lib/WeBWorK/ContentGenerator/Leaderboard.pm
```diff
@@ -0,0 +1,101 @@
...
+	my $ce = $c->ce;
+
+	# Get user Data
+	$c->{userName}    = $c->param('user');
+	$c->{studentName} = $c->param('effectiveUser') // $c->{userName};
+
+	return unless $c->authz->hasPermissions($c->{userName}, 'view_leaderboard');
+
+	# Get list of all users (except set-level proctors) and achievements.
+	my @allUsers    = $db->getUsersWhere({ user_id => { not_like => 'set_id:%' } });
```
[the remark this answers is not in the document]
somiaj: Seems I only used this in one place too, I will just call the database there in the start of the for loop instead of storing it in a variable.

lib/WeBWorK/ContentGenerator/Leaderboard.pm
somiaj: The `next unless $userBadge` line doesn't seem needed here, I just put the check in the next if statement.

somiaj:
@drgrice1 thanks again, your suggestions improve load speed quite a bit here. Still takes a couple of seconds, but 3-5 seconds is faster than the 15 or seconds I was getting.

lib/WeBWorK/ContentGenerator/Leaderboard.pm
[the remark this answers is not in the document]
drgrice1: There was no usage of the `sortAchievements` method in the code that I was looking at.

lib/WeBWorK/ContentGenerator/Leaderboard.pm
drgrice1: The `next unless $userBadge` is needed.  If the achievement has not been assigned to a user, then it will be undefined, and then the `$userBadge->earned` will cause an error.

somiaj: It is, I put the check in the next if statement, `push(@badges, $achievement) if $userBadge && ....`.

lib/WeBWorK/ContentGenerator/Leaderboard.pm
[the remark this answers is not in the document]
drgrice1: :+1:

lib/WeBWorK/ContentGenerator/AchievementsLeaderboard.pm
```diff
@@ -0,0 +1,104 @@
...
+		my $level = $globalData->level_achievement_id ? $achievementsById{ $globalData->level_achievement_id } : '';
+
+		my @badges;
+		for my $achievement (@achievements) {
+			# Skip level achievements and only show earned achievements.
+			last if $achievement->category eq 'level';
+
+			my $userBadge = $db->getUserAchievement($user->user_id, $achievement->achievement_id);
+			push(@badges, $achievement) if $userBadge && $achievement->enabled && $userBadge->earned;
+		}
```
drgrice1, line 73: There is another improvement that should speed this up a little more.  Change this to
```suggestion
		my %userAchievements = map { $_->achievement_id => $_ } $db->getUserAchievementsWhere({
			user_id        => $user->user_id,
			achievement_id => [ map { $_->achievement_id } grep { $_->category ne 'level' } @achievements ],
		});

		my @badges;
		for my $achievement (@achievements) {
			# Skip level achievements and only show earned achievements.
			last if $achievement->category eq 'level';

			push(@badges, $achievement)
				if $userAchievements{ $achievement->achievement_id }
				&& $achievement->enabled
				&& $userAchievements{ $achievement->achievement_id }->earned;
		}
```

This gets all user achievements except those in the "level" category at once instead of getting them one at a time in the loop.  You can add `&& $_->category ne 'one_time'` if you like to also exclude the "extensions" achievement that isn't used.

I think that this code and the code in `lib/WeBWorK/ContentGenerator/Achievements.pm` needs to be revisited though.  Calling `last` as is done here on line 69 and on line 143 of that file is messed up.  The assumption that "level" achievements are last is really unintelligent design.  The id's of the achievements can be edited by the instructor, and if the instructor changes the id of a "level" achievement to be smaller than the id of the first achievement, then it messes everything up.

somiaj: I agree when I saw that as well, but that would require a bit of a redesign on how level achievements work, since they always need to be processed last. I wonder if we could just update the `sortAchievements` method to always put the level achievements at the end of the list independent of their number.

I cant think of any reason why one would want to process any other achievement after the level achievements, since if a student earned any of those achievements and got enough points to pass a level boundary, the level won't be awarded until the next time the level achievements are processed.

drgrice1: The problem is that an instructor might not understand how achievements work, and change the id's in such a way that it messes things up. It is just a poor design. It shouldn't be hard to rework things to ensure level achievements are processed last.  I'm my opinion there shouldn't even be a numerical id that is visible and even worse editable.

somiaj: The level achievements should be ordered within their group though, so I don't know how making the id hidden and not editable would let instructors add/remove levels. Though I guess there could be a separate list of level achievements that are managed and ordered in a different way. That would take a bit of work beyond just updating the sort to always include them last.

drgrice1: In general achievements need to be completely redesigned for webwork3 anyway.  Using actual Perl code for these, and another "safe compartment" is not the way to go to begin with.

drgrice1 (approved):
I think this looks good now.

Alex-Jordan:
This is a casual question, not a request to change anything. But would it be difficult to make it so that there is just "Achievements" in the menu, and then if you go to that page, you can toggle between seeing your own achievements and seeing the leaderboard? (Assuming you have permission to see the leaderboard.)

Alex-Jordan:
In the leaderboard, are the badges handled well in terms of accessibility? Is it right that they are in an `a` with role `button`?

Alex-Jordan:
There are some bootstrap styling options you could consider for the table for readability. I like the following but I defer to your opinion.

- The `table` could have `.table-striped` or `.table-hover`.
- The `thead` could have `.table-primary`.

Alex-Jordan:
This lays out the rows ordered by rank. Would it be helpful to be able to sort the rows by the user name too?

somiaj:
> This is a casual question, not a request to change anything. But would it be difficult to make it so that there is just "Achievements" in the menu, and then if you go to that page, you can toggle between seeing your own achievements and seeing the leaderboard? (Assuming you have permission to see the leaderboard.)

This could be done. I don't think we would want to use tabs, and load the leaderboard each time a user visits the achievement page, as this would require loading all achievements for all users each time. So the another option is to have a link to the leaderboard on the achievements page vs have a link in the main menu.

somiaj:
> In the leaderboard, are the badges handled well in terms of accessibility? Is it right that they are in an `a` with role `button`?

Here is what is rendered. I'm unsure how the `--data-bs-title` popup content gets dealt with by screen readers, but it appears the javascript is adding an `--aria-describedby` tag with the popover with the content, so everything looks good to me compared to other popovers.

```
<a class="help-popup" role="button" tabindex="0" data-bs-placement="top" data-bs-toggle="popover" data-bs-html="true" data-bs-content="<strong>Night Owl</strong><br>Finish a homework set between 12PM and 2AM.">
		<img alt="Night Owl Icon" src="<course>/achievements/night_owl.png" width="50">
</a>
```

templates/ContentGenerator/AchievementsLeaderboard.html.ep
```diff
@@ -0,0 +1,66 @@
...
+					% for my $badge (@$badges) {
+						<a class="help-popup" role="button" tabindex="0"
+							data-bs-placement="top" data-bs-toggle="popover" data-bs-html="true"
+							data-bs-content="<strong><%= $badge->{name} %></strong><br><%= $badge->{description} %>">
+							<%= image $badge->{icon}
+								? "$ce->{courseURLs}{achievements}/$badge->{icon}"
+								: "$ce->{webworkURLs}{htdocs}/images/defaulticon.png",
+								alt   => $c->maketext('[_1] Icon', $badge->{name}),
+								width => 50 %>
+						</a>
```
drgrice1, line 60: I think that @Alex-Jordan is right that these should not use an `a` tag with the role `button`.  I think that instead this should be
```suggestion
						<button class="btn btn-sm btn-link help-popup" type="button"
							data-bs-placement="top" data-bs-toggle="popover" data-bs-html="true"
							data-bs-content="<strong><%= $badge->{name} %></strong><br><%= $badge->{description} %>">
							<%= image $badge->{icon}
								? "$ce->{courseURLs}{achievements}/$badge->{icon}"
								: "$ce->{webworkURLs}{htdocs}/images/defaulticon.png",
								alt   => $c->maketext('[_1] Icon', $badge->{name}),
								width => 50 %>
						</button>
```
At the very least, that looks better.  The popovers align better relative to the icon, and the focus appearance is nicer.  The buttons are a little bigger, but I think that looks fine for these.  Custom styling could be applied if desired to make them smaller.

drgrice1:
Hmm.   I just saw something interesting looking at both the timing log and the morbo trace log.  The achievements leaderboard is loading faster on my machine than the achievements page.  The leaderboard takes about half the time to load.  Admittedly I don't have a lot of achievements data on my local machine (really only data for my instructor user and one test student), but there is still more that the leaderboard is loading.  So it stands to reason that the leaderboard should take longer.  It seems the achievement page needs to be cleaned up.  Looking at the code I see that that is one of the pages that didn't get as much attention when I did some of the heavy database access revision stuff several years ago.  Many of the tips I gave you for speeding this page up need to be applied to the achievements page also.

templates/ContentGenerator/AchievementsLeaderboard.html.ep
somiaj: Updated. The only thing I didn't like was the default padding making the badges space out a bit more, so I added the `p-0` class to remove the padding and it looks better to me.

somiaj:
> This lays out the rows ordered by rank. Would it be helpful to be able to sort the rows by the user name too?

I initially thought about this, but wasn't sure how useful it would end up being. To me the usernames should be hidden from students anyways, and it might add a minor use for instructors to find a particular student, but wasn't sure it was worth the extra complication due to this.

> There are some bootstrap styling options you could consider for the table for readability. I like the following but I defer to your opinion.
> 
>     * The `table` could have `.table-striped` or `.table-hover`.
> 
>     * The `thead` could have `.table-primary`.

I am not the biggest fan of `.table-striped`, but I applied the other two. I can apply `.table-striped` if others think it is better, I just don't prefer that look.

drgrice1:
The `table-hover` class reveals a problem with some of the icons.  The level two icon in particular, but there are others.  That icon has a white background instead of a transparent background.  As such when the mouse cursor is over a row with that icon in it it doesn't look the best.

drgrice1 (approved):
This looks pretty good to me at this point.

Alex-Jordan:
I am going to merge this. I hope that the "Achievements" and "Achievements Leaderboard" menu items can be merged later though. 

I opened #2665 to address the issues with the png images and transparency.

Issue #2581, opened 2024-09-26
Due date must be after reduced scoring date even if reduced credit is not enabled for a set


dlglin:
To reproduce:

Enable reduced scoring for a course.  Create a set with reducing scoring disabled, and set the reduced scoring date to October 1.  Now try to set the due date to September 30.  When you click "done" on the date picker the due date reverts to October 1 since the javascript forces the due date to be no earlier than the reduced scoring date.

I almost got burned by this.  If you're not paying attention it's easy to miss that the date you just set got changed to something else.

This might be fixable with some more complicated logic in the js (if reduced scoring is set to no, then don't check that the due date is after the reduced scoring date, but then if reduced scoring is subsequently changed to yes, then check the dates).

somiaj:
This is behavior I have gotten use to as well, so I agree with your suggestion with one modification. I think the JS should honor the user's input independent of reduced scoring being enabled.

This should also apply to any date. I think the JS should always adjust the other dates to match the user's input instead of reverting dates that are outside of the range set by other dates. I have often wanted to set the answer date and have it adjust both the close date and reduced scoring date, but instead I have to set the reduced scoring date, then copy/paste to the other dates.

drgrice1:
Ideally, the reduced scoring date would only be visible if reduced scoring is enabled both for the course and the set.  If reduced scoring is enabled for the course, then JavaScript should toggle visibility of the reduced scoring date to match the "Reduced Scoring Enabled" setting.   Of course if reduced scoring is disabled for the set, the JavaScript would also not consider its value when changing other dates.

@somiaj:  As to adjusting the other dates to match the user's input, it does do that in some cases.  If you change the reduced scoring date to a date that is later than the close or answer date, then those dates are adjusted.  However, the adjustments that you suggest really don't make sense (at least not as you stated them).  Think about it carefully and you will see that that would cause problems.

somiaj:
Fixed in #2583.

PR #2583, opened 2024-09-27
Change behavior when changing dates in a datepicker group to be more intuitive.

drgrice1:
Previously when any of the dates were changed, the open, reduced scoring, close, and answer dates were ensured to be in the correct order by always changing the input later in the last to a later date if needed.

This instead acts on the input that was modified.  That input value is not changed from what the user just changed it to.  Any earlier inputs in the list are set to the new date in the changed input if they are later than that date, and any later inputs in the list are set to the new date in the changed input if they are earlier than that date.

This is to address issue #2581.

### DIFF
--- a/htdocs/js/DatePicker/datepicker.js
+++ b/htdocs/js/DatePicker/datepicker.js
@@ -110,9 +110,9 @@
 								selectedDate.setFullYear(today.getFullYear());
 								selectedDate.setMonth(today.getMonth());
 								selectedDate.setDate(today.getDate());
-								fp.setDate(selectedDate);
+								fp.setDate(selectedDate, true);
 							} else if (index === 1) {
-								fp.setDate(new Date());
+								fp.setDate(new Date(), true);
 							}
 						}
 					})
--- a/htdocs/js/DatePicker/datepicker.js
+++ b/htdocs/js/DatePicker/datepicker.js
@@ -56,7 +56,7 @@
 				groupRules[activeIndex][1];
 
 			for (let i = 0; i < groupRules.length; ++i) {
-				if (i == activeFieldDate) continue;
+				if (i == activeIndex) continue;
 				const thisFieldDate =
 					groupRules[i][0]?.parentNode._flatpickr.selectedDates[0]?.getTime() || groupRules[i][1];
 				if (i < activeIndex && thisFieldDate > activeFieldDate)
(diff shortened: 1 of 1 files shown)


somiaj:
Overall this is what I was asking for in #2581, though I may have not been clear enough there. I just wanted user input to always be honored and the other dates updated based on that.

In testing, this doesn't seem to be fully working for me, as the open date is never being adjusted under any of my tests when it should. My tests are to set all the dates (open, reduced, close, answer) to be the same. Then when I set the answer, closed, or reduced date to be an earlier date, the open date isn't updated (all the other dates are updated appropriately though).

drgrice1:
The open date updates when I test it. :stuck_out_tongue_winking_eye:

somiaj (approved):
Works fine now, I must have been hallucinating.

drgrice1:
You did see that I added another commit after you made the comment?

somiaj:
Yea, I saw the fix, I tried to figure out why the open date was being ignored, but missed the loop index needed changed.

somiaj:
I seeing the following behavior with the date picker and setting the dates for individual user(s). Go override the dates for a single user, set all the dates to the same date in the future (I'm using a full year). Then delete the 'close' date. When this happens the reduced scoring and open date are reset, but they aren't rest to a valid date corresponding to the set close date. Here is a screen shot.

image

To me the dates should have reverted to `9/30/24, 12:00 PM` (the set close date), but for some reason `10/2/24, 12:00 AM` gets used.

somiaj:
Note, setting all the dates to some date over a year in the past, and deleting the open date, does reset the reduced, close, and answer date to equal the open date of the set, so I wasn't able to reproduce any issue with updating the dates in the other direction.

drgrice1:
Are you testing with this pull request?  I just tested with this pull request and used the exact dates shown in your screen shot, and I don't see that.  When I delete the close date, the open and reduced scoring dates are then set to `9/30/24, 12:00 PM`.

somiaj:
Yes, I'm testing with this PR. Wonder what other settings in my set could be causing the problems.

somiaj (approved):

drgrice1:
@somiaj: I pushed a change that should address the issue that you saw.  Basically, instead of falling back to the user's previous override when a date is deleted, it falls back to the course's date.

Alex-Jordan:
Tested and I get the desired behavior.

I'm not entirely following the finer point that was discovered if there is a previous date override. I will approve this and invite @somiaj to merge it if that isue is also resolved.

Alex-Jordan (approved):

somiaj:
Great, this works well in this case now. Thanks for looking into that last minor issue.

PR #2584, opened 2024-09-28
Use reduced credit when checking if problem is correct in Grades table.

somiaj:
Use `compute_unreduced_grade` to determine if a problem is correct or not on the student's grade table to highlight correct problems.

### DIFF
--- a/lib/WeBWorK/Utils/Sets.pm
+++ b/lib/WeBWorK/Utils/Sets.pm
@@ -103,7 +103,7 @@ sub grade_set ($db, $set, $studentName, $setIsVersioned = 0, $wantProblemDetails
 		$status = 1 if $status > 1;
 
 		if ($wantProblemDetails) {
-			push(@$problem_scores,             $problemRecord->attempted ? 100 * wwRound(2, $status) : '&nbsp;.&nbsp;');
+			push(@$problem_scores, $status || $problemRecord->attempted ? 100 * wwRound(2, $status) : '&nbsp;.&nbsp;');
 			push(@$problem_incorrect_attempts, $problemRecord->num_incorrect || 0);
 		}
 
--- a/lib/WeBWorK/Utils/Sets.pm
+++ b/lib/WeBWorK/Utils/Sets.pm
@@ -114,7 +114,8 @@ sub grade_set ($db, $set, $studentName, $setIsVersioned = 0, $wantProblemDetails
 	}
 
 	if (wantarray) {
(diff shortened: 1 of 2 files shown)


lib/WeBWorK/Utils/Sets.pm
```diff
@@ -309,7 +309,8 @@ In list context this returns a list containing the total number of correct
 problems, and the total number of problems in the set.  If
 C<$wantProblemDetails> is true, then a reference to an array of the scores for
 each problem, and a reference to the array of the number of incorrect attempts
-for each problem are also included in the returned list.
+for each problem are also included in the returned list. An array reference
+to the array of problem records is last in the returned list.
```
drgrice1, line 313: Rather than adding a sentence to the end, add the new returned element to the existing list.  Furthermore, make the type of problem record returned specific.  There are numerous types of problem records that can be obtained from the database.  There are global problem records, pure user problem records, and merged problem records (which are really an abstracted record as they are not actually in the database but are a merge of the global and user problem records).

So make the last three lines read

```
each problem, a reference to the array of the number of incorrect attempts for
each problem, and a reference to an array of merged problem records are also
included in the returned list.
```

drgrice1 (approved):
I think this looks good.  I only have a wording suggestion on the POD change for the `grade_set` method.

I am debating if it would be better if the `grade_set` handled the `compute_unreduced_score` computation, and instead of returning the array of problem records, only returned an array of the unreduced scores.  I suppose leave it as it is for now, and if the unreduced score becomes something used more frequently elsewhere in combination with the `grade_set` results, then we can change it at that time.

somiaj:
POD updated as suggested.

somiaj:
It appears that #2617 modifies `grade_set` in a similar way to what is needed in this PR. Since this is a much smaller change, this should be delayed until after #2617 is merged, and I can update all the conflicts. I should be able to just use the changes to `grade_set` directly and only have to update the grades table.

Also should I include the suggested update in #2627, so non-zero grades to unattempted problems are also shown in the grade table?

drgrice1:
You can add the non zero change to this.

drgrice1:
Also if you update the `grade_set` method so that your changes are identical to the change in #2617, then it shouldn't conflict.

somiaj:
Okay updated my changes to `grade_set` to match #2617, and this also will show non-zero status on unattempted problems.

drgrice1:
Looks good.  In theory it shouldn't matter on the merge order of this or #2617.  Assuming the other changes are far enough away.  Usually more than 5 lines away is enough.

PR #2586, opened 2024-09-28
Add LTI configuration info to the LTI course map page.

drgrice1:
I have had this branch sitting here for a while (since March).  Time to put it in. 

This adds LTI configuration information to the course admin LTI course map page.  This information is probably useful when managing the course map.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -2223,11 +2223,38 @@ sub do_unhide_inactive_course ($c) {
 # LTI Course Map Management
 
 sub manage_lti_course_map_form ($c) {
-	my $ce        = $c->ce;
+	my $ce = $c->ce;
+
 	my @courseIDs = listCourses($ce);
 	my %courseMap = map { $_->course_id => $_->lms_context_id } $c->db->getLTICourseMapsWhere;
 	for (@courseIDs) { $courseMap{$_} = '' unless defined $courseMap{$_} }
-	return $c->include('ContentGenerator/CourseAdmin/manage_lti_course_map_form', courseMap => \%courseMap);
+
+	my %ltiConfigs = map {
+		my $ce = eval { WeBWorK::CourseEnvironment->new({ courseName => $_ }) };
+		$_ => $@
+			? undef
+			: {
+				LTIVersion => $ce->{LTIVersion},
+				$ce->{LTIVersion}
+				? (
+					$ce->{LTIVersion} eq 'v1p1'
+					? (ConsumerKey => $ce->{LTI}{v1p1}{ConsumerKey})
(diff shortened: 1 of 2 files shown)


somiaj (approved):

Alex-Jordan (approved):
Do I understand right that this loads a course environment for each course and gets the relevant values, and then displays them?

It shouldn't hold up this PR, but I wondered about displaying more that could be relevant to an admin.  Like `preferred_source_of_username` and other things. The config chain can be such that these things don't have the site settings for them, and an admin could see some of those things without digging, turning on LTI debugging, etc.

drgrice1:
I was thinking eventually more could be added.  I want sure what would be beneficial to add though, and didn't want to just add everything.

Perhaps a general course environment inspector page might be something to think about adding to the admin course.  A page that shows all course environment​ resolved values for a course.  I am not sure how feasible that is though.

drgrice1:
I realized that I chose the things that are shown because they are the distinguishing LTI parameters that are used to determine the correct course to map to when an LTI content generation request is received.  I.e., the things that are important for determining the correct course to use in the LMS course map.  So only the `ConsumerKey` is shown for LTI 1.1 (and not the `ConsumerSecret`), and only the `PlatformID`, `ClientID`, and `DeploymentID` for LTI 1.3 (and not the `PublicKeysetURL`, `AccessTokenURL`, `AccessTokenAUD`, and `AuthReqURL`).

Note that the `ConsumerKey` for LTI 1.1 will be unset in most cases.  The only time it needs to be set is if there are multiple courses from different LMS's that have the same LMS course id and both use a course on the same webwork2 server.

However, this could be changed to show all of the parameters listed above as well as perhaps others (like the `preferred_source_of_username`).

somiaj:
There are two approvals, I'll merge this for now, discussion of showing more stuff could be in a different PR, or wait a bit?

Alex-Jordan:
I beat you to it :)

Having other stuff could be a different discussion. Some page like @drgrice1 describes, where you would click a course name and then see all its environment variables might be nice.

drgrice1:
Merging now is fine.  More can possibly be added later.

It also occurred to me that adding other things like the other LTI parameters for LTI 1.1 and 13 that I listed before may obsure the reason that the parameters that are currently shown are the ones that are shown.  That is what I said before, in that they are the important factors for determining the correct course to map to.

Also, I want to correct my comment on the LTI 1.1. consumer key.  That is that the ConsumerKey for LTI 1.1 will be unset in most cases. The only time it needs to be set is if there are multiple courses from different LMS's that have the same LMS "context" id (not "course" id as I said before) and both use a course on the same webwork2 server.  I copied that from the comment in the authen_LTI_1_1.conf file.  Apparently when I switched from course id to context id (due to @Alex-Jordan's point that the course id may not be unique even from the same LMS), I forgot to change the comment in the authen_LTI_1_1.conf file.

PR #2589, opened 2024-10-01
configuration option to allow students to work on sets that are missi…

Alex-Jordan:
…ng the set sourced ID even in homework grade passback mode

This is what I proposed in #2582.

I'm open to changing the name of the config variable. Also since it only makes sense to do this with LTI 1.3, I put it in `authem_LTI_1_3.conf` but maybe that is not the best place.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSets.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSets.pm
@@ -191,12 +191,14 @@ sub getSetStatus ($c, $set) {
 			$link_is_active = 0 unless $canViewUnopened;
 			push(@$other_messages, $c->restricted_progression_msg(0, $set->restricted_status * 100, @restricted));
 		} elsif (!$canViewUnopened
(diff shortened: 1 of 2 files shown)


lib/WeBWorK/ContentGenerator/ProblemSets.pm
```diff
@@ -191,12 +191,14 @@ sub getSetStatus ($c, $set) {
 			$link_is_active = 0 unless $canViewUnopened;
 			push(@$other_messages, $c->restricted_progression_msg(0, $set->restricted_status * 100, @restricted));
 		} elsif (!$canViewUnopened
+			&& ($ce->{LTIVersion} ne '1p3' || !$ce->{LTI}{v1p3}{ignoreMissingSourcedID})
```
drgrice1, line 194: This should be
```suggestion
			&& ($ce->{LTIVersion} ne 'v1p3' || !$ce->{LTI}{v1p3}{ignoreMissingSourcedID})
```

somiaj (approved):

drgrice1 (approved):
Looks good.

Issue #2588, opened 2024-10-01
Authentication issue with grader in 2.19


Alex-Jordan:
In 2.19, I made a test. I'm not sure if it matters, but my example of the issue has these settings:

<img width="524" alt="Screenshot 2024-09-30 at 8 22 08 PM" src="https://github.com/user-attachments/assets/2186d373-7e4d-474f-a098-57e204541742">

Again, this probably is not relevant, but I started the quiz, and answered one question correctly, then submitted the test to be graded.

I went back tot he test and clicked "Show Problem Graders". I went to the second exercise, and tried to award myself 100%. When I click Save, I get an authentication error:

<img width="762" alt="Screenshot 2024-09-30 at 8 24 50 PM" src="https://github.com/user-attachments/assets/46e6b9f7-d421-4c26-b90a-a627b6c37ea2">

This was reported by a Runestone subscriber. For them, the test settings are the same, but they sent me this different authentication error screenshot:

![image(3)](https://github.com/user-attachments/assets/70aa3f90-414c-451c-a582-c335e8431d7c)

PR #2590, opened 2024-10-01
Fix multiple loading of javscript for the single problem grader.

drgrice1:
This is the cause of issue #2588.

If a gateway quiz has mulitple problems on the same page, then the single problem grader template is loaded once for each problem.  Each time the template is loaded, it adds the single problem grader javacript to the page.  So as many instances of the javascript are active on the page as there are problems on the page.  As such, the save button event handlers are added once for each problem on the page as well.  So when the save button is clicked, the score and comment for that problem are submitted once for each problem on the page.  All of those requests firing at the same time can cause authentication to fail because the session key can be deleted and put back in again by different processes, and the timing of things can mess things up.  Note that this does reveal some deeper database issues though.

### DIFF
--- a/templates/HTML/SingleProblemGrader/grader.html.ep
+++ b/templates/HTML/SingleProblemGrader/grader.html.ep
@@ -1,8 +1,11 @@
 % use WeBWorK::Utils 'wwRound';
 %
-% content_for js => begin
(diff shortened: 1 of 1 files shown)


somiaj (approved):
Tested this fixes the issue, and I can see that only a single action is now fired when the save button is hit.

This should probably be a hotfix too.

Issue #2587, opened 2024-09-30
Users with multiple LMS roles


Alex-Jordan:
See this forum post for background:
https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8565#p21511
The issue also happened once with a Runestone-hosted course, and I suggested a manual one-off way to address the issue to that instructor.

Summary: a user in an LMS can have multiple roles. I *think* they would only ever have one "membership" role, indicating their role in that particular LMS course. But they could also have multiple "institution/person" roles. The troublesome case is when a student has a "Learner" membership role but also they have what would be a higher level "institution/person" role, perhaps because once at some other time, in some other course, they were a TA. Or even an instructor (consider an English professor using tuition credit to take a math course).

We want this person to be identified as a WeBWorK "student", but they have an "institution/person" role that is higher so WeBWorK won't automatically create the person's account using LTI.

1. Do we really need to be including "institution/person" roles? I don't know enough to rule out that we need those. But it seems like if there is a "membership" role, there would only be one. Maybe I am wrong about that, but it would reduce instances of this issue.
2. Could there ever be a situation where there is no "membership" role at all? What I can think of is some LMS admin who entered a course and has no actual role in that course. They would have some high level "institution/person" role (or several) and then what? Actually in practice, I would like it if such people would get WW accounts initialized (as WW "student"s) because they are often checking to see that some LTI setup is working or not. When they get the message about WW not creating an account for them, they are unsure if things are working. But I digress, are there other situations where there would be no "membership" role?

Assuming I am answering my own questions correctly, I propose that we:
1. only look at "membership" roles, and use the highest one to decide if the WW account will be created
2. if there is no "membership" role, go ahead and create a WW "student"

dlglin:
If it helps here's what D2L sends when I use LTI 1.3 as an instructor:
```
  "https://purl.imsglobal.org/spec/lti/claim/roles" => [
    "http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor",
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student",
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Learner"
  ],
```
When I log in as a user who is a student in the current D2L course, but is an instructor in other courses I get:
```
  "https://purl.imsglobal.org/spec/lti/claim/roles" => [
    "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner",
[...]

PR #2591, opened 2024-10-02
For LTI 1.3 only consider roles in the context for automatically created users.

drgrice1:
Currently the context  and institution roles are are considered.  This makes it so that only context roles are considered by default.  If someone needs the institution roles, then the new $LTI{v1p3}{AllowInstitutionRoles} option defined in conf/authen_LTI_1_3 can be set to 1, and then those roles will also be considered.  For now system roles are always ignored (as they were before).

This is to fix the issue discussed in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8565#p21514.

### DIFF
--- a/conf/authen_LTI_1_3.conf.dist
+++ b/conf/authen_LTI_1_3.conf.dist
@@ -144,6 +144,12 @@ $LTI{v1p3}{LMSrolesToWeBWorKroles} = {
 	'Grader'                  => 'ta',
 };
 
+# The LMS reports roles context (or membership), instititution, and system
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
This also relates to #2587, which should be closed if this is merged.

drgrice1:
@Alex-Jordan: Thanks for referencing the issue.  I was looking for that when I made this pull request, but for some reason, I couldn't find it.

Alex-Jordan (approved):

PR #2593, opened 2024-10-03
Improvements to SAML2 authentication.

drgrice1:
This fixes some issues with #2511.  This pull request is built on that one. That pull request is a nice start, but has some issues.  Thanks to @ionparticle for starting this work.  One of the most important things introduced in that pull request is the development identity provider via SimpleSAMLphp.  That gave me a way to work with this.
    
Although I liked the idea of using a Mojolicious plugin, that approach does not fit into the webwork2 course environment system.  So, unfortunately, that had to be changed.  So this rewrites the SAML2 authentication module to use the usual approach that all of the other authentication modules use.  There is a `authen_saml2.conf.dist` file that should be copied to `authen_saml2.conf` to use SAML2 authentication.  All settings for the authentication module are in that file.
    
The primary limitation of the plugin approach was that it is not possible for different courses to use different identity providers.  Or for one course to use SAML2 authentication and another only use the basic password authentication (without the need to add a URL parameter).  This is something that is expected of the authentication modules, and is desirable for multi-institutional servers.
    
Another problem with the implementation is that is does not work with `$session_management_via = "key"` set.  The reason that doesn't work is because the implementation broke the rule of creating and accessing a session before authentication is completed.  So this reimplementation uses the database via the "nonce" key hack much like LTI 1.1 authentication does.
    
The previous implentation also used the `WEBWORK_ROOT_URL` environment variable in the code.  That is not an environment variable that is defined for webwork2.  It is only defined for the docker build, and can not be used in the code.
    
The previous implementation also does not work with two factor authentication.  If an identity provider does not provide two factor authentication, then webwork2's two factor authentication should be used.  Of course, if the identity provider does provide two factor authentication, then the two factor authentication can be disabled in `localOverrides.conf`.
    
An option to enable service provider initiated logout has also been added. It is off by default, but if enabled, when the user clicks the "Log Out" button, a request is sent to the identity provider to also end its session.
    
The `README.md` file that was with the plugin code has been moved to the `docker-config/idp` directory, and is now purely for instructions on how to use the simpleSAMLphp development instance.  The readme also includes instructions on how to set up a simpleSAMLphp development instance without docker.  The documentation on how to configure and use the SAML2 authentication module is now all in the `authen_saml2.conf.dist` file.
    
Note that the webwork2 service provider certificate files are no longer directly in the configuration file. Instead file locations are to be provided.  It didn't really make sense to have the actual certificates in the configuration file, then write them to a file, which the Net::SAML2 module would read.  Furthermore, it would be very easy to add the certificates incorrectly to the configuration file.  With the YAML file approach if indentation were not correct, then the configuration file would fail to load.

This is #2547 again.  That branch was accidentally deleted.


somiaj (approved):
I suggest we merge this so it can be used.

@ionparticle can you still confirm this meets your needs and you haven't noticed any other issues with it?

ionparticle:
@somiaj regarding issues, the code has been rewritten, it's very different from what we run in prod, so our experience isn't really applicable. However, this does still meet our needs, as this implementation should work for us and we would prefer upstream maintenance of the auth system.

Alex-Jordan (approved):
I'm approving after skimming through the code. I don't know how to test this and I'll leave it to someone more involved here to decide to merge it.

somiaj:
I am going to merge this, if SAML users have issues they can report them.

PR #2596, opened 2024-10-07
Remove the special case in the `getThirdPartyAssetURL` method for MathQuill.

drgrice1:
That special handling is no longer needed now that MathQuill is installed via npm from the npm repositories.  Furthermore, jsdelivr provides the files from the package that can be used when the `$options{thirdPartyAssetsUseCDN}` option is set to 1.


somiaj (approved):
Confirmed that mathquill is loaded from the CDN with this.

pstaabp (approved):

pstaabp (approved):

PR #2603, opened 2024-10-14
Fix checkboxes for permission levels on Config page

dlglin:
This is to address #2602.

As suggested by @Alex-Jordan it orders the permissions by their numerical levels, and skips the 'nobody' permission if it is defined.  Is it okay to hard-code skipping 'nobody'?

### DIFF
--- a/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
+++ b/lib/WeBWorK/ConfigObject/permission_checkboxlist.pm
@@ -52,8 +52,8 @@ sub comparison_value ($self, $val) {
 }
 
 sub entry_widget ($self, $default, $is_secret = 0) {
-    my $c = $self->{c};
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks good.

I don't know why I didn't see this when this was implemented this way.  The roles shouldn't have been hard coded in that way.

I think it is fine to hard code skipping 'nobody'.  That is a special role that shouldn't really be modified.

somiaj (approved):

somiaj:
Compared to the hot fix, this has two commits, would merging these commits and keeping the commits the same help when merging branches next release?

drgrice1:
It doesn't really help.  I usually merge main into develop, and then reset all of the code to what is in develop to fix merge conflicts.  And there are always merge conflicts due to the hotfixes.

PR #2608, opened 2024-10-22
Guessing at fixes for Shibboleth issues.

drgrice1:
There are two problems with the Shibboleth authenticaiton that I have identified in discussion with @glarose together with analyzing code and the issues posted in the forums (see
https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8534#p21568).

The first is that the Shibboleth module overrides the WeBWorK::Authen check_session method, and does not properly initialize the database session.  This causes anything that uses the database session (which includes test proctor authentication) to fail.

The second is that the WeBWorK::Controller session method returns undefined in the case that cookies are disabled (which the Shibboleth authentication module does).  The Mojolicious::Plugin::DefaultHelpers _validation method then calls the session method expecting it to be an object or a hash reference.

I have no way to test that this fixes the issues that have been reported, and can only write code that should work.  So someone with a Shibboleth identity provider will need to test this.

### DIFF
--- a/lib/WeBWorK/Authen/Shibboleth.pm
+++ b/lib/WeBWorK/Authen/Shibboleth.pm
@@ -167,20 +167,24 @@ sub check_session {
 	if ($ce->{shiboff}) {
 		return $self->SUPER::check_session(@_);
 	} else {
-		my $Key = $db->getKey($userID);    # checked
+		my $Key = $db->getKey($userID);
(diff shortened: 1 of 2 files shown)


drgrice1:
I am closing this.  This is not the full solution.  An upcoming pull request will resolve this completely.

PR #2610, opened 2024-10-23
Tests, disable start button when acting as another user.

somiaj:
~~Hide the "Start New Test" button when acting as another user unless the user has permissions to do so.~~

**Edit:**  This now just disables the "Start New Test" button and displays a tool tip informing the user they don't have permission to start new tests when acting as another user.

Fix some logic where the warning about creating new test version wouldn't show with the record_answers_when_acting_as_student permission, and instead the version would be created without warning.

Translate the error messages that are created for an invalidSet.

Remove the test for invalidProblem, which isn't used in tests.

Update error messages to state 'test' instead of 'set'.

Note, to me it wasn't clear from the comment in defaults.conf that the `record_answers_when_acting_as_student` also applied to creating tests and recording answers for tests, but from the code it appears that is the intent, so that is what I went with here.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -331,7 +331,8 @@ async sub pre_header_initialize ($c) {
 		$c->{assignment_type} = 'gateway';
 
 		if (!$authz->hasPermissions($userID, 'modify_problem_sets')) {
-			$c->{invalidSet} = 'You do not have the authorization level required to view/edit undefined sets.';
+			$c->{invalidSet} =
+				$c->maketext('You do not have the authorization level required to view/edit undefined sets.');
 
 			# Define these so that we can drop through to report the error in body.
 			$tmplSet = fake_set($db);
@@ -340,8 +341,8 @@ async sub pre_header_initialize ($c) {
 		} else {
 			# In this case we're creating a fake set from the input, so the input must include a source file.
 			if (!$c->param('sourceFilePath')) {
-				$c->{invalidSet} =
-					'An Undefined_Set was requested, but no source file for the contained problem was provided.';
+				$c->{invalidSet} = $c->maketext(
+					'An Undefined_Set was requested, but no source file for the contained problem was provided.');
 
 				# Define these so that we can drop through to report the error in body.
 				$tmplSet = fake_set($db);
@@ -529,9 +530,11 @@ async sub pre_header_initialize ($c) {
 				&& (
 					$effectiveUserID eq $userID
 					|| (
-						$authz->hasPermissions($userID, 'record_answers_when_acting_as_student')
-						|| ($authz->hasPermissions($userID, 'create_new_set_version_when_acting_as_student')
-							&& $c->param('createnew_ok'))
+						(
+							$authz->hasPermissions($userID, 'record_answers_when_acting_as_student')
+							|| $authz->hasPermissions($userID, 'create_new_set_version_when_acting_as_student')
+						)
+						&& $c->param('createnew_ok')
 					)
 				)
 				)
@@ -586,26 +589,33 @@ async sub pre_header_initialize ($c) {
 				$currentNumAttempts = 0;
 
 			} elsif ($maxAttempts != -1 && $totalNumVersions > $maxAttempts) {
-				$c->{invalidSet} = 'No new versions of this assignment are available, '
-					. 'because you have already taken the maximum number allowed.';
+				$c->{invalidSet} = $c->maketext('No new versions of this test are available, '
+						. 'because you have already taken the maximum number allowed.');
 
-			} elsif ($effectiveUserID ne $userID
-				&& $authz->hasPermissions($userID, 'create_new_set_version_when_acting_as_student'))
+			} elsif (
+				$effectiveUserID ne $userID
+				&& ($authz->hasPermissions($userID, 'record_answers_when_acting_as_student')
+					|| $authz->hasPermissions($userID, 'create_new_set_version_when_acting_as_student'))
+				)
 			{
 
-				$c->{invalidSet} =
-					"User $effectiveUserID is being acted "
-					. 'as.  If you continue, you will create a new version of this set '
-					. 'for that user, which will count against their allowed maximum '
-					. 'number of versions for the current time interval.  IN GENERAL, THIS '
-					. 'IS NOT WHAT YOU WANT TO DO.  Please be sure that you want to '
-					. 'do this before clicking the "Create new set version" link '
-					. 'below.  Alternately, PRESS THE "BACK" BUTTON and continue.';
+				$c->{invalidSet} = $c->maketext(
+					'User [_1] is being acted as.  If you continue, you will create a new version of '
+						. 'this set for that user, which will count against their allowed maximum '
+						. 'number of versions for the current time interval.  IN GENERAL, THIS IS NOT '
+						. 'WHAT YOU WANT TO DO.  Please be sure that you want to do this before clicking '
+						. 'the "Create New Test Version" button below.  Alternately, PRESS THE "BACK" BUTTON '
+						. 'and continue.',
+					$effectiveUserID
+				);
 				$c->{invalidVersionCreation} = 1;
 
 			} elsif ($effectiveUserID ne $userID) {
-				$c->{invalidSet} = "User $effectiveUserID is being acted as.  "
-					. 'When acting as another user, new versions of the set cannot be created.';
+				$c->{invalidSet} = $c->maketext(
+					'User [_1] is being acted as. When acting as another user, '
+						. 'new versions of the test cannot be created.',
+					$effectiveUserID
+				);
 				$c->{invalidVersionCreation} = 2;
 
 			} elsif (($maxAttemptsPerVersion == 0 || $currentNumAttempts < $maxAttemptsPerVersion)
@@ -615,16 +625,16 @@ async sub pre_header_initialize ($c) {
 					$versionIsOpen = 1;
 				} else {
 					$c->{invalidSet} =
-						'No new  versions of this assignment are available, because the set is not open or its time'
-						. ' limit has expired.';
+						$c->maketext('No new versions of this test are available, because the test is '
+							. 'not open or its time limit has expired.');
 				}
 
 			} elsif ($versionsPerInterval
 				&& ($currentNumVersions >= $versionsPerInterval))
 			{
 				$c->{invalidSet} =
-					'You have already taken all available versions of this test in the current time interval.  '
-					. 'You may take the test again after the time interval has expired.';
+					$c->maketext('You have already taken all available versions of this test in the current '
+						. 'time interval. You may take the test again after the time interval has expired.');
 
 			}
 
@@ -643,7 +653,7 @@ async sub pre_header_initialize ($c) {
 		}
 
 	} elsif (!$c->{invalidSet} && !$requestedVersion) {
-		$c->{invalidSet} = 'This set is closed.  No new set versions may be taken.';
+		$c->{invalidSet} = $c->maketext('This test is closed.  No new test versions may be taken.');
 	}
 
 	# If the proctor session key does not have a set version id, then add it.  This occurs when a student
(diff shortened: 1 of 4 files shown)


somiaj:
What about still showing the button, but disabling it and maybe a popup with the info about it is disabled because they are currently acting as another user?

drgrice1:
That would work.

somiaj:
@drgrice1 I cannot seem to get the tooltip to work as described in the bootstrap docs. First 5.3 states that `bs-data-title` is used, but that is ignored, instead I had to use `title`, second it seems to ignore `data-bs-placement="top"` setting and just floats the tooltip next to the mouse.

Outside of that, the start new test button is just disabled and provides a tooltip informing the user as to why.

templates/ContentGenerator/ProblemSet/version_list.html.ep
```diff
@@ -126,9 +126,22 @@
...
+				% 'You are acting as another user and do not have permission to start a new test for other users.')
+			% . '"';
+	% }
 	<div class="mb-3">
-		<%= link_to maketext('Start New Test') => $c->systemLink(url_for($routeName, setID => $set->set_id)),
-			class => 'btn btn-primary' =%>
+		<span<%== $disabled_tooltip =%>>
+			<%= link_to maketext('Start New Test') => $c->systemLink(url_for($routeName, setID => $set->set_id)),
+				class => "btn btn-primary$disabled" =%>
+		</span>
```
drgrice1, line 144: This has a problem.  There is a reason that in the Bootstrap documentation for this that they use a `button` instead of an `a` tag with an `href`.  A disabled `button` is not in the tab order.  However, an `a` tag with an `href` even if disabled is in the tab order.  So this adds an unnecessary tab stop.  That means that if you are navigating through the page with the keyboard and "tab", then first focus stops on the `span` (since that has `tabindex="0"` added), but then hitting "tab" again brings focus uselessly to the `a` tag.

So change this to
```suggestion
		% if ($disable_start_new_test) {
			<span class="d-inline-block set-id-tooltip" tabindex="0" data-bs-toggle="tooltip" data-bs-placement="top"
				data-bs-title="<%= maketext(
					'You are acting as another user and do not have permission to start a new test for other users.'
				) %>">
				<button class="btn btn-primary" type="button" disabled><%= maketext('Start New Test') %></button>
			</span>
		% } else {
			<%= link_to maketext('Start New Test') => $c->systemLink(url_for($routeName, setID => $set->set_id)),
				class => 'btn btn-primary' =%>
		% }
```
and delete lines 129-139 above.  In addition this is actually less code, and much more readable than the mess of putting attributes into temporary variables.

Note that I used the `set-id-tooltip` class in this, and switched to the `data-bs-title` attribute as per my other comments.

drgrice1 (commented):
Some Bootstrap components (for instance modals) are initialized automatically.  If you add the appropriate `data-bs-toggle` attribute and other necessary markup for those, then you don't need to do anything else for them to work.  However, other Bootstrap components (for instance tooltips) are opt in (for performance reasons).  That means that even if you add the appropriate `data-bs-toggle` attribute, it won't work unless you also initialize them via JavaScript.  See https://getbootstrap.com/docs/5.3/components/tooltips/#enable-tooltips.

So one option is to add code to initialize the tooltips.  This would be appropriate to add in system.js since it is the only specific JavaScript for this page at this time.

Another option is to just add the `set-id-tooltip` class to the span.  There is already code in system.js to enable tooltips on any element with this class that has a `data-bs-title` attribute.  Note that tooltips with this class are really general purpose tooltips contrary to the class name.  See my comment on lines 88-89 of system.js.  Although, these tooltips are forced to be in the set position (in your case on the top) with no fallback position.  That may not be appropriate for these tooltips.

Note that by switching to the `title` attribute, you are not making this work as a Bootstrap tooltip, and the `data-bs-toggle="tooltip"` attribute is still ignored and is useless.  In that case the native browser `title` handling is in effect and you get the usual hover behavior for a `title` which is different.

templates/ContentGenerator/ProblemSet/version_list.html.ep
somiaj: Thanks. I did notice that a button was being used, but even testing the bootstrap example didn't work, as I was missing the `set-id-tooltip` class (must have overlooked that somewhere). I have updated the button to your suggestions and things work as expected for me now.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -47,6 +47,11 @@ async sub initialize ($c) {
 	my $userID  = $c->param('user');
 	my $eUserID = $c->param('effectiveUser');
 
+	# Don't show "Start New Test" button when acting as another user, unless user has permissions to do so.
+	$c->{disable_start_new_test} = $userID ne $eUserID
```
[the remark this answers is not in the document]
somiaj: Thanks.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -529,9 +530,11 @@ async sub pre_header_initialize ($c) {
...
 					$effectiveUserID eq $userID
 					|| (
-						$authz->hasPermissions($userID, 'record_answers_when_acting_as_student')
-						|| ($authz->hasPermissions($userID, 'create_new_set_version_when_acting_as_student')
-							&& $c->param('createnew_ok'))
+						(
+							$authz->hasPermissions($userID, 'record_answers_when_acting_as_student')
+							|| $authz->hasPermissions($userID, 'create_new_set_version_when_acting_as_student')
+						)
+						&& $c->param('createnew_ok')
```
drgrice1, line 537: Note that this has been the logic used for this since 2008.  I agree that it was not correct, but I am not entirely convinced this change is correct either.  I have questioned this for a while now.  I do no believe that the `record_answers_when_acting_as_student` permission should be considered here at all.  Either that or the `create_new_set_version_when_acting_as_student` should be removed from the code, and have this just be covered by the other permission.  This is the only thing the latter permission is ever used for, so it really does not make sense to have both of these permissions apply here.

drgrice1: Actually looking at the only other place that `create_new_set_version_when_acting_as_student` is used, you could argue that the current code is correct, and your correction is wrong.  The only other place it is used is to show the warning message about creating a new version for another user.  The `record_answers_when_acting_as_student` is not considered there.  So one could argue that the `record_answers_when_acting_as_student` is a stronger permission that allows you to create new teset versions without a warning, while `create_new_set_version_when_acting_as_student` allows you to create new test versions, but with a warning.

In fact, looking even closer to the code I see that if this change is made, you might as well delete the check of the `record_answers_when_acting_as_student` permission here, because it is impossible for the `createnew_ok` parameter to also be set with that permission (at least without the user manually adding it to the request themself).  The `createnew_ok` parameter is only ever added to the page if the `create_new_set_version_when_acting_as_student` permission is satisfied.

somiaj: When I looked at `defaults.conf` it was not clear to me that `record_answers_when_acting_as_student` should include `create_new_set_version_when_acting_as_student`, and I agree with you that it probably shouldn't, but the code used it.

What I didn't like is the only permission a user can toggle in "Course Configuration" is `record_answers_when_acting_as_student`, and toggling this bypasses the warning, which I didn't think was a good idea, as otherwise that means most likely that warning is never seen. So I went with ensuring the warning was being shown.

To me we should either only check the `create_new_set_version_when_acting_as_student` permission here (but would this be something we would also want an instructor to be able to toggle in the "Course Configuration"?). If we want it to be something an instructor can toggle on the "Course Configuration" page, probably removing the `create_new_set_version_when_acting_as_student` permission altogether would be simpler, and the warning with `record_answers_when_acting_as_student` should be enough to ensure instructors don't accidentally start a test version.

drgrice1: In any case a change like this needs to be discussed, and a consensus reached. The rest of this pull request is fine though.

somiaj: Agreed, we should discuss it at a future meeting to decide how to move forward.

drgrice1: I see now that you also added `record_answers_when_acting_as_student` on line 597 so that a user with this permission can still start a test version.

So my opinion is that how you have changed it in this pull request is the correct way to go.  No one should be able to start a test version for another user without at least some warning about the consequences of that action.  Someone with that permission should be able to create a test version for another user.  That is a stronger permission than being able to just create a test version, which does not allow recording answers.

somiaj: That was my intent, to not change what permissions could start a test version as another user, but only ensure the warning is always shown.

somiaj: Your final statement did make the permissions here make more sense to me, and I now see why the `record_answers_when_acting_as_student` was being used here, even though it was initially confusing as to why this permission was being used.

drgrice1 (approved):

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -586,26 +589,33 @@ async sub pre_header_initialize ($c) {
...
-					. 'below.  Alternately, PRESS THE "BACK" BUTTON and continue.';
+				$c->{invalidSet} = $c->maketext(
+					'User [_1] is being acted as.  If you continue, you will create a new version of '
+						. 'this set for that user, which will count against their allowed maximum '
+						. 'number of versions for the current time interval.  IN GENERAL, THIS IS NOT '
+						. 'WHAT YOU WANT TO DO.  Please be sure that you want to do this before clicking '
+						. 'the "Create New Test Version" button below.  Alternately, PRESS THE "BACK" BUTTON '
+						. 'and continue.',
+					$effectiveUserID
+				);
```
Alex-Jordan, line 610: I'm nitpicking about old wording here, but since it is now going to be translated, it seems best to change it now.

* `User [_1] is being acted as.` could be `You are acting as user [_1].`. Less passive.
* Using capital letters to emphasize something is not a good idea. It's not clear that will be understood as emphasis by a screen reader. I think it is fine to just use regular case.
* I think 'Alternately' should be 'Alternatively'.
* In 'PRESS THE "BACK" BUTTON', is that referring to the web browser's back button? If so, is that apparent and always meaningful, like on a small screen? Maybe this whole 'Alternately...' sentence should just be removed.

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -586,26 +589,33 @@ async sub pre_header_initialize ($c) {
...
 				$c->{invalidVersionCreation} = 1;
 
 			} elsif ($effectiveUserID ne $userID) {
-				$c->{invalidSet} = "User $effectiveUserID is being acted as.  "
-					. 'When acting as another user, new versions of the set cannot be created.';
+				$c->{invalidSet} = $c->maketext(
+					'User [_1] is being acted as. When acting as another user, '
```
Alex-Jordan, line 615: Also here I would change the passive voice in ''User [_1] is being acted as.'

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
drgrice1: Perhaps in addition to the "Create New Test" button, there should be a "Cancel" button which takes the user back to the test version list page.

drgrice1: Note that this isn't the only place that webwork uses the "use the browser back button" wording.  The email preview page also does this (and there may be other places as well).

somiaj: I added a "Cancel" button, removed the caps, and updated the language via the above suggestions and a few more edits I added myself.

Alex-Jordan (approved):

drgrice1:
I think this looks good, and I will merge.

PR #2612, opened 2024-10-25
Fix (and completely revamp) the Shibboleth authentication module.  (hotfix of #2611)

drgrice1:
This updates the Shibboleth authentication module to fit into the new scheme of the general webwork2 authentication process.  The module is set up to work just like all of the other up to date webwork2 authentication modules.

It has its own configuration file (conf/authen_shibboleth.conf.dist) that should be used instead of adding a buch of variables to localOverrides.conf.  The include statement in localOverrides.conf should be uncommented, and the dist file copied and modified. Furthermore, relatively complete instructions on how to use the authentication module are in the comments in the configuration file.

The variables in the configuration file are all the same as before, except that there is one new one.  That is the
`$shibboleth{bypass_query}`.  Previously "bypassShib" was hard coded for this purpose.  Now that can be configured.  If that variable is not set (and for those using this module before it wouldn't be), then the bypass parameter will not work.  So this is the only real change from before.

The issues that were causing webwork2's session not to work have been fixed.  This means that proctored test access will work again.

The library browser, pg problem editor, and everything else that uses the rpc endpoints will work correctly.  There simply is nothing special that the authentication module needs to do here, and most importantly it needs to not do anything special (like reverting to the base authentication module).  The rpc enpoints now use the usual authentication methods, and that does work with mod_shib.

### DIFF
--- a/conf/authen_shibboleth.conf.dist
+++ b/conf/authen_shibboleth.conf.dist
@@ -1,0 +1,143 @@
+#!perl
+################################################################################
+# Configuration for using Shibboleth authentication.
+#
+# To enable Shibboleth authentication, copy this file to
+# conf/authen_shibboleth.conf and uncomment the appropriate lines in
+# localOverrides.conf.
+#
+################################################################################
+
+# Note that Shibboleth authentication will only work if webwork2 is proxied via
+# apache2 and a Shibboleth service provider (mod_shib) is installed and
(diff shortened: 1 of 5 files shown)


somiaj (approved):
Same as #2611, we should merge this for the shibboleth users.

Alex-Jordan (approved):

PR #2613, opened 2024-11-01
Fix a rounding error with SingleProblemGrader point values.

somiaj:
When computing the point value in the SingleProblemGrader, the database saves 6 decimal digits in the status field, while the single problem grader only saved 2. This difference in rounding can cause the computation of the point value with a stepSize of 0.25 to get get different values between the computed score and the score saved in the database, causing the "Use points from last check" button to show up and present a different rounded value.

This fixes the issue by saving the current score in a `$rawCurrnetScore` variable that isn't rounded and using that for the point computation. Note, `$rawCurrentScore` can have more than the 6 decimal digits in the saved database score, but rounding this to match the database won't improve the accuracy of the point value rounded to the nearest stepSize.

Note I noticed this issue with questions with multiple parts with 12-14 points causing a 0.25 stepSize.

### DIFF
--- a/templates/HTML/SingleProblemGrader/grader.html.ep
+++ b/templates/HTML/SingleProblemGrader/grader.html.ep
@@ -10,7 +10,8 @@
 <div class="problem-grader">
 	<hr>
 	<div class="problem-grader-table">
-		% my $currentScore = 0;
+		% my $currentScore    = 0;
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks good.

Alex-Jordan (approved):

PR #2614, opened 2024-11-05
Control when sets with score 0 are included in LTI grade passback

Alex-Jordan:
The first thing I want to say about this is that I have not tested it actually works. It's too risky to try it with my production LMS. So when time permits, if anyone can test with their local Canvas or Moodle, and if that works, I'll risk trying it with my instance of D2L.

This adds a config variable `$LTISendZeroScores`. Its purpose is to control when sets with score 0 are included in passing grades data to the LMS. Its value can be `open', 'reduced', 'close', 'answer', or 'never'. The first four correspond to dates for a set. The idea is that you only include a zero-score set in grades data to the LMS if it is after this setting. So even if all sets open when the term begins, they do not necessarily contribute to tanking a student's overall grade in the LMS simply because the scores are all initially zero.

In "course" grade passback mode, zero-score sets are omitted from the overall calculation if it is presently earlier than that set's date corresponding to this setting. For example if this setting is "close", open sets with no work completed will be omitted from the scoring calculation until we reach the close date for those sets.

In "homework" grade passback mode, scores for zero-score sets are not sent to the LMS it is later than that set's date corresponding to this setting. There may already be null grade items in the LMS, and they will stay null until either we reach the appropriate date, or there is a nonzero score for the set.

This is configurable, because some instructors will always put close dates at the end of the term and want to use the reduced scoring date as the "due date". And other variations.

Note that prior to this commit, all sets are included in grade data sent to the LMS if it is after the open date. That is why I set the default for this to be `'open'`. If it's better, we could change the default behavior.

Also note that with this commit, if the score on a set is nonzero, that set will be included in grades data sent to the LMS as long as we are past the set's open date. Using the open date is hard coded. I have often thought we might want to change that behavior or make it configurable too. But I think it is a separate question from when to include sets with score zero.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
@@ -128,7 +128,8 @@ async sub submit_course_grade ($self, $userID) {
 	$self->warning("lis_source_did is not available for user: $userID")
 		if !$user->lis_source_did && ($ce->{debug_lti_grade_passback} || $self->{post_processing_mode});
 
-	return await $self->submit_grade($user->lis_source_did, scalar(grade_all_sets($db, $userID)));
+	return await $self->submit_grade($user->lis_source_did,
+		scalar(grade_all_sets($db, $userID, $ce->{LTISendZeroScores})));
 }
 
 # Computes and submits the set grade for $userID and $setID to the LMS.  For gateways the best score is used.
@@ -147,14 +148,23 @@ async sub submit_set_grade ($self, $userID, $setID) {
 	$self->warning('lis_source_did is not available for this set.')
 		if !$userSet->lis_source_did && ($ce->{debug_lti_grade_passback} || $self->{post_processing_mode});
 
-	return await $self->submit_grade(
-		$userSet->lis_source_did,
-		scalar(
-			$userSet->assignment_type =~ /gateway/
+	my $score =
+		scalar($userSet->assignment_type =~ /gateway/
 			? grade_gateway($db, $userSet, $userSet->set_id, $userID)
-			: grade_set($db, $userSet, $userID, 0)
-		)
+			: grade_set($db, $userSet, $userID, 0));
+
+	my %dates = (
+		open    => $userSet->open_date(),
+		reduced => $userSet->reduced_scoring_date(),
+		close   => $userSet->close_date(),
+		answer  => $userSet->answer_date()
 	);
+
+	if ($score == 0) {
+		return if ($ce->{LTISendZeroScores} eq 'never' || before($dates{ $ce->{LTISendZeroScores} }));
+	}
+
+	return await $self->submit_grade($userSet->lis_source_did, $score);
 }
 
 # Submits a score of $score to the lms with $sourcedid as the identifier.
--- a/lib/WeBWorK/Authen/LTIAdvantage/SubmitGrade.pm
+++ b/lib/WeBWorK/Authen/LTIAdvantage/SubmitGrade.pm
@@ -207,7 +207,8 @@ async sub submit_course_grade ($self, $userID) {
 	$self->warning('LMS user id is not available for this user.')   unless $user->lis_source_did;
 	$self->warning('LMS lineitem is not available for the course.') unless $lineitem;
 
-	return await $self->submit_grade($user->lis_source_did, $lineitem, grade_all_sets($db, $userID));
+	return await $self->submit_grade($user->lis_source_did, $lineitem,
+		grade_all_sets($db, $userID, $ce->{LTISendZeroScores}));
 }
 
 # Computes and submits the set grade for $userID and $setID to the LMS.  For gateways the best score is used.
@@ -225,10 +226,23 @@ async sub submit_set_grade ($self, $userID, $setID) {
 	$self->warning('LMS user id is not available for this user.') unless $user->lis_source_did;
 	$self->warning('LMS lineitem is not available for this set.') unless $userSet->lis_source_did;
 
+	my ($totalRight, $total) = (grade_set($db, $userSet, $userID, 0))[ 0, 1 ];
+
+	my %dates = (
+		open    => $userSet->open_date(),
+		reduced => $userSet->reduced_scoring_date(),
+		close   => $userSet->close_date(),
+		answer  => $userSet->answer_date()
+	);
+
+	if ($totalRight == 0) {
+		return if ($ce->{LTISendZeroScores} eq 'never' || before($dates{ $ce->{LTISendZeroScores} }));
+	}
+
 	return await $self->submit_grade($user->lis_source_did, $userSet->lis_source_did,
 		$userSet->assignment_type =~ /gateway/
 		? grade_gateway($db, $userSet, $userSet->set_id, $userID)
-		: (grade_set($db, $userSet, $userID, 0))[ 0, 1 ]);
+		: ($totalRight, $total));
 }
 
 # Submits scoreGiven and scoreMaximum to the lms with $sourcedid as the identifier.
--- a/lib/WeBWorK/Utils/Sets.pm
+++ b/lib/WeBWorK/Utils/Sets.pm
@@ -20,7 +20,7 @@ use Carp;
 
 use PGrandom;
 use WeBWorK::Utils qw(wwRound);
-use WeBWorK::Utils::DateTime qw(after);
+use WeBWorK::Utils::DateTime qw(after before);
 use WeBWorK::Utils::JITAR qw(jitar_id_to_seq jitar_problem_adjusted_status);
 
 our @EXPORT_OK = qw(
@@ -146,7 +146,7 @@ sub grade_gateway ($db, $set, $setName, $studentName) {
 	}
 }
 
-sub grade_all_sets ($db, $studentName) {
+sub grade_all_sets ($db, $studentName, $LTISendZeroScores) {
 	my @setIDs     = $db->listUserSets($studentName);
 	my @userSetIDs = map { [ $studentName, $_ ] } @setIDs;
 	my @userSets   = $db->getMergedSets(@userSetIDs);
@@ -156,17 +156,24 @@ sub grade_all_sets ($db, $studentName) {
 
 	for my $userSet (@userSets) {
 		next unless (after($userSet->open_date()));
+		my $totalRight;
+		my $total;
+		my %dates = (
+			open    => $userSet->open_date(),
+			reduced => $userSet->reduced_scoring_date(),
+			close   => $userSet->close_date(),
+			answer  => $userSet->answer_date()
+		);
 		if ($userSet->assignment_type() =~ /gateway/) {
-
-			my ($totalRight, $total) = grade_gateway($db, $userSet, $userSet->set_id, $studentName);
-			$courseTotalRight += $totalRight;
-			$courseTotal      += $total;
+			($totalRight, $total) = grade_gateway($db, $userSet, $userSet->set_id, $studentName);
 		} else {
-			my ($totalRight, $total) = grade_set($db, $userSet, $studentName, 0);
-
-			$courseTotalRight += $totalRight;
-			$courseTotal      += $total;
+			($totalRight, $total) = grade_set($db, $userSet, $studentName, 0);
 		}
+		if ($totalRight == 0) {
+			next if ($LTISendZeroScores eq 'never' || before($dates{$LTISendZeroScores}));
+		}
+		$courseTotalRight += $totalRight;
+		$courseTotal      += $total;
 	}
 
 	if (wantarray) {
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -135,6 +135,12 @@ $LTIGradeMode = '';
 #$LTIGradeMode = 'course';
 #$LTIGradeMode = 'homework';
 
+# This can be set to 'open', 'reduced', 'close', 'answer', or 'never'. It controls when a set
+# with score 0 will be included in grades that are passed back to the LMS. For example, setting
+# this to 'close' means that a set with score 0 is only included in grades sent to the LMS once
+# it is past the set's close date.
+$LTISendZeroScores = 'open';
+
 # When set this variable sends grades back to the LMS every time a user submits an answer.  This
 # keeps students grades up to date but can be a drain on the server.
 $LTIGradeOnSubmit = 1;
@@ -170,6 +176,7 @@ $LTIMassUpdateInterval = 86400;    #in seconds
 	#'LTI{v1p3}{LMS_url}',
 	#'external_auth',
 	#'LTIGradeMode',
+	#'LTISendZeroScores',
 	#'LTIGradeOnSubmit',
 	#'LTIMassUpdateInterval',
 	#'LMSManageUserData',
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -912,6 +912,25 @@ sub getConfigValues ($ce) {
 			labels => { '' => 'None', 'course' => 'Course', 'homework' => 'Homework' },
 			type   => 'popuplist'
 		},
+		LTISendZeroScores => {
+			var  => 'LTISendZeroScores',
+			doc  => x('When to send zero scores to the LMS'),
+			doc2 => x(
+				'If the grade passback mode is "homework", this controls when a set with score 0 will have '
+					. 'that score sent to the LMS, as opposed to leaving the score null in the LMS. If the grade '
+					. 'passback mode is "course", this controls when a set with score 0 will be included in the '
+					. 'overall grade calculation that is sent to the LMS.'
+			),
+			values => [qw(open reduced close answer never)],
+			labels => {
+				open    => 'After Open Date',
+				reduced => 'After Reduced Scoring Date',
+				close   => 'After Close Date',
+				answer  => 'After Answer Date',
+				never   => 'Never'
+			},
+			type => 'popuplist'
+		},
 		LTIGradeOnSubmit => {
 			var  => 'LTIGradeOnSubmit',
 			doc  => x('Update LMS Grade Each Submit'),


somiaj:
I think the logic should be a bit stronger than just checking if the score is zero. I think we should test if the set has any attempts, and zeros should still be sent on sets students have attempted but still have a zero on.

somiaj:
And if we go with the above suggestion, update the option to reflect this to something like `$LTISendUnattemptedGrades`.

drgrice1:
I agree that attempts should be considered.  Particularly for tests.  If a student has completed a test attempt and actually gets a score of 0, then that should be submitted.

Alex-Jordan:
In the case of a test, under what conditions would would we actually want a zero score to be passed to the LMS prior to the date set by the new `$LTISendZeroScores`? Given all the variations one could have with test configuration, when I start to think this through it gives me a headache.

In the case of a regular set, I'm not sure I agree. If I'm using this new feature, say I use the setting where zeros will be sent to the LMS once the close date passes. And there is an assignment where a student looks at the first question, submits a wrong answer, realizes they don't know what they are doing, and then leaves the set for days while they read, attend class, etc. I don't see myself wanting to pass that zero score to the LMS, just because of that one incorrect attempt. The zero showing up in the LMS will cause stress for the student, and then stress for me when the student reaches out to ask why they have a zero in the LMS when then due date has not yet come.

What is the scenario where you would actually want that zero to be sent to the LMS, just because one or more problems have unsuccessful attempts? Given that eventually we reach the date where the zero will be sent anyway?

somiaj:
Another hypothetical, take your situation, student opens up an assignment, gets one of ten parts correct on the first problem's attempt, giving a score of 1% on the assignment, then leaves the set alone for days. Why should this student get the 1% sent back after their single attempt, but the student who did a single attempt and got a zero doesn't get the grade sent back, since they have both attempted the set and got a very similar score. (The only situation I can think of I wouldn't want the zero sent back is if they hit grade problem without entering in any answers, but I don't think we can test for this and it won't be worth the effort if we could).

Though I will conceded for homework sets, as most common use of of homework gives students multiple attempts and it is much easier to test if the score is zero vs looping over all the problems to see if they have attempted it. So I don't think even if someone would prefer all attempts get sent to the LMS even if they had a zero score, that this distinction is useful enough to have to loop though the problems in a set checking for attempts.

As for a quiz/test. I would check if either multiple versions exists or a single version exists with at least one attempt on the first problem, and send the grade back in this case. To get to this point a student would have to have both started a test and then graded the test at least once, which I think is enough to want to send the grade back to the LMS even if it is zero.

drgrice1:
In thinking about this more, I think this needs a different change.  That is instead of never sending a zero score, this should only not send zero scores when a mass grade pass back is being used.  If a student submits an answer for an assignment, the score should be sent back regardless of if it is zero, and if a test is submitted the same.

drgrice1:
Also, I think that submitting zero scores of assignments is not needed at all for a mass grade pass back even if a student has attempted problems.  As to tests, I think that it would be sufficient to check the close dates of the versions.  If there is a version that is completed with no more attempts, then the close date of the version is the completion time.  So you don't need a complicated check for that.

All that is really important to me on this is that zero scores are not sent by mass pass back, but when a student submits an answer on an assignment or submits a test, I want grades always sent then.  I don't really care about zero scores of a completed test, because those are extremely rare.  The real issue for me is when a student opens a test front page and doesn't even start the test, and then a mass grade pass back happens later and submits a zero.  The student has never submitted anything.  In fact this is a case where there aren't even any test versions.  So for me, a sufficient check would even be if there are no test versions, then don't submit the grade.

Alex-Jordan:
Once an assignment closes, I do want zero scores sent with the mass pass
back. Are you saying you wouldn't want that? Otherwise an instructor has to
manually enter zeros into the LMS.

When you are thinking about tests, are you including thinking about the
ones that have no time limit? I'm guessing that means their close date is
the template close date

Alex-Jordan:
>  If a student submits an answer for an assignment

This makes me wonder if you are only considering when `$LTIGradeOnSubmit` is set. I have that turned off, so grade passback only happens at all when there is a mass passback.

drgrice1:
> Once an assignment closes, I do want zero scores sent with the mass pass back. Are you saying you wouldn't want that? Otherwise an instructor has to manually enter zeros into the LMS.

Once an assignment closes mass grade pass back should send all grades always regardless of this setting.

>  When you are thinking about tests, are you including thinking about the ones that have no time limit? I'm guessing that means their close date is the template close date

I am saying that when using dates, the dates for the versioned set should be used (if set), instead of the dates for the user set.  So if you have the test set for one submission, then the grade will be passed back once it that submission has occurred.

> This makes me wonder if you are only considering when `$LTIGradeOnSubmit` is set. I have that turned off, so grade pass back only happens at all when there is a mass pass back.

I do have `$LTIGradeOnSubmit` on.  That is the default setting, and using that is consistent with how proprietary systems do things with LTI grade pass back.  Grade pass back is intended to be immediate. However, if you have that turned off, then what I am suggesting is exactly what you want.  I am saying this setting should only apply to mass grade pass back, and not when an grade is submitted by a student submitting an answer to an assignment that is incorrect (or otherwise) or submitting a test (even if the score is zero).  This is easy to detect in the code.  For a mass grade pass back the `post_processing_mode` flag is set, and when a student submits an answer and the `$LTIGradeOnSubmit` option is on, then that flag is not set.

drgrice1:
Also, you should remove the "never" option from this setting.  There is no point in that.  You can not use grade pass back if you want that.

lib/WeBWorK/Utils/Sets.pm
```diff
@@ -146,7 +146,7 @@ sub grade_gateway ($db, $set, $setName, $studentName) {
 	}
 }
 
-sub grade_all_sets ($db, $studentName) {
+sub grade_all_sets ($db, $studentName, $LTISendZeroScores) {
```
drgrice1, line 149: The parameter should not be call `$LTISendZeroScores` here.  Instead it should be something like `$includeZeroScores`.  This method is not specific to LTI grade pass back.  At this point this method is only called in that case, but this is in the general WeBWorK/Utils and in the future could be used elsewhere.  Also, this should be given a default value so the method can still be called without passing that parameter.  Furthermore, the method should be implemented so that if that default value is used, then the behavior is the same as before.  Also, since this is a general utility method in a general utility file, this method may actually already be used in its current form by someone in a script.  As I have already mentioned the "never" value for this parameter should be eliminated, and instead there should be an "always" value which would work like before and would be the default value of this parameter.

Alex-Jordan:
> Also, you should remove the "never" option from this setting. There is no point in that. You can not use grade pass back if you want that.

Can you explain this? If I'm using grade passback, and if this new config variable is set to "never", the intention is that *nonzero* scores still get passed back to the LMS like they do now. So I don't understand "You can not use grade pass back" if this is set to "never".

Setting this to `never` would prevent zero scores from ever being passed back. This could be desirable if that leaves the LMS grade item score null, and excluded from the LMS grade averaging. I can think of a few reasons an instructor would want that.

drgrice1:
Although, looking at the code, what you say is not true.  If the option value is "never", then it never sends zero scores.  There is no option to ignore dates and always send zero scores.

lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm
```diff
@@ -147,14 +148,23 @@ async sub submit_set_grade ($self, $userID, $setID) {
...
+	my %dates = (
+		open    => $userSet->open_date(),
+		reduced => $userSet->reduced_scoring_date(),
+		close   => $userSet->close_date(),
+		answer  => $userSet->answer_date()
 	);
+
+	if ($score == 0) {
+		return if ($ce->{LTISendZeroScores} eq 'never' || before($dates{ $ce->{LTISendZeroScores} }));
+	}
```
drgrice1, line 165: The nested if here is not needed.

lib/WeBWorK/Authen/LTIAdvantage/SubmitGrade.pm
```diff
@@ -225,10 +226,23 @@ async sub submit_set_grade ($self, $userID, $setID) {
...
+	my %dates = (
+		open    => $userSet->open_date(),
+		reduced => $userSet->reduced_scoring_date(),
+		close   => $userSet->close_date(),
+		answer  => $userSet->answer_date()
+	);
+
+	if ($totalRight == 0) {
+		return if ($ce->{LTISendZeroScores} eq 'never' || before($dates{ $ce->{LTISendZeroScores} }));
+	}
```
drgrice1, line 240: This nested if is also not needed.

Alex-Jordan:
A change I am thinking about with this is to not make it tied to the score "0".  If there were a score threshold (which could be 0, or 70%, or 100%, etc) then that could be used instead of 0.

For example in my classes, I'd want grade passback to be postponed until either the score threshold is met, or a critical date like the reduced scoring date comes. Or in the case of a test (which I don't actually use) I guess there are other conditions where I'd want the grade to be passed back too.

The idea is to prevent early delivery to the LMS of a low score simply because the student hasn't gotten around to it yet. But also guarantee delivery to the LMS eventually. (That is, unless the instruction wants low scores to never be automatically passed back. I think this make sense in certain situations.)

So please pause on any testing if that is happening. I've see the code review comments and I'll act on those with the next round.

drgrice1:
What I am saying is that the "never" value for `$LTISendZeroScores` currently never sends zero scores.  I am saying change that.   Eliminate the "never" value from consideration as it is useless.  Instead use an "always" value, and make it so that `$LTISendZeroScores` is set to "always" then a score of zero is always sent.  That is, make an option that maintains the current behavior.  Your implementation does not have such an option at this point.

Alex-Jordan:
> What I am saying is that the "never" value for $LTISendZeroScores currently never sends zero scores.

Right.

>  I am saying change that. 

But I want that behavior if "never" is set.

>  Eliminate the "never" value from consideration as it is useless.

No it's not useless. An instructor might legitimately want zero scores to never be sent to the LMS automatically.

>  That is, make an option that maintains the current behavior.

The "open" option for this setting maintains the current behavior. What you are wanting for "always" is achieved by "open". The current behavior does not send grades data for sets that have not yet opened.

drgrice1:
> Setting this to `never` would prevent zero scores from ever being passed back. This could be desirable if that leaves the LMS grade item score null, and excluded from the LMS grade averaging. I can think of a few reasons an instructor would want that.

What reasons do you see for an instructor not wanting zero scores passed back ever?  Even after the last dates of the set have passed?  I don't see any real reason for that.

Alex-Jordan:
So an "always" option as the default would actually change the behavior from how it has always worked, if it started sending scores for sets that have not yet opened. I set the default for the new setting to "open" in the config file to preserve the current behavior.

Alex-Jordan:
> What reasons do you see for an instructor not wanting zero scores passed back ever? 

For starters, the entire WeBWorK course could be an optional extra credit thing. Students who do not engage would then not get 0s populating their LMS gradebook. The grade columns would remain null, and omitted from the LMS overall grade calculation.

But a student who engages and has nonzero scores (or later, scores above the threshold I am planning) would get their scores sent automatically.

There are other situations, but this one is easy to describe. Is it persuasive that there is a use case?

Alex-Jordan:
> So an "always" option as the default would actually change the behavior from how it has always worked, if it started sending scores for sets that have not yet opened.

Am I wrong about that? I am nearly certain that in "course" grade passback mode, sets that have not yet opened are excluded from the grade calculation. 

But maybe it's not like that in "homework" grade passback mode.

drgrice1:
I suppose.  I don't really see the point of an entire course that is for optional extra credit.  If you must have the option, then it is fine.

The current implementation of LTI grade submission methods do not consider dates at all.  When grades are submitted as a result of a student submitting an answer or a test (when $LTIGradeOnSubmit is true), then dates are indirectly considered by the calling modules (Problem.pm or GatewayQuiz.pm), and the submission can't even occur.  Also, since students generally can't open the assignment before the open date, grade pass back usually can't happen because webwork won't get the grade pass back information (actually for LTI 1.3 if the LMS provides grade passback information for an instructor, then it could).

However, when mass grade pass back occurs, it doesn't consider set dates at all.  One scenario that would make it so that grades are passed back for a set before the set open date is if an instructor opens a set by setting the open date before the current date, and students start working on the set.  Then the instructor changes the open date to a later date to prevent students from continuing to work on the assignment.  The mass grade pass back would then continue to send grades for those students even though it is before what is now the open date.

Alex-Jordan:
Here is the function `grade_all_sets` in `Utils/Sets.pm` which only considers sets for which the open date has passed. It is ultimately used by `SubmitGrade.pm`, although only for "course" grade passback mode.

https://github.com/openwebwork/webwork2/blob/c4b8039416fd6f99db0c677ffd226be481a57c30/lib/WeBWorK/Utils/Sets.pm#L158

Alex-Jordan:
I've long been familiar with how not-yet-open sets are excluded from the overall "course" grade that is passed back. I've wrongly assumed that behavior applied to "homework" passback mode too.

With LTI 1.3, I do tell instructors here to initialize everything by opening up links once. They would be doing this at the start of the term even if the set opens later. We have `$LTICheckPrior` set, so the LMS grade items for these sets would be null, and would never get updated to "0" anyway. So that explains why I have not heard complaints about 0 scores showing up before a set has opened, even though the instructor has initialized the LTI data.

Alex-Jordan:
I think I understand enough now to redesign this. Thanks for all the help talking through it. Evening "recreation" time, I will give this another go.

drgrice1:
So course grade pass back mode does consider the open date then.  I didn't see that line.  But I don't see anywhere that homework grade pass back mode does.

drgrice1:
Thanks for working on this.  It is something that I think is needed.

somiaj:
> A change I am thinking about with this is to not make it tied to the score "0". If there were a score threshold (which could be 0, or 70%, or 100%, etc) then that could be used instead of 0.

I haven't fully been following so if this has been discussed my apologizes. I personally don't see the use for this, but it could be a something other users want so I'm okay with it. Though I still think there should be a way to make this not score based but attempted base. Could allowing the value of the score to be negative, (like -1), mean attempts are used vs score?

Alex-Jordan:
@somiaj Earlier you described a situation where a student only has a 1% score but will come back days later to complete the assignment. You questioned why that 1% score should be sent back to the LMS. And I agree, it shouldn't. If this gets up and working. I'd want something like a 70% threshold for sets before the LMS gets an update. (Of course, the LMS would also get an update later at the close date or reduced credit date, regardless of the student meeting the threshold.

It's something I would want in order to reduce stress when an LMS score pops up as low and the student panics. But the threshold can be 0 and then it's not different from how this PR started.

I also want to respect what you are asking for with recognizing what sets are attempted or not. I haven't reached the point where I understand what that should actually mean though, especially with tests. It's been discussed in the thread and maybe I just need to read and reflect on that.

Alex-Jordan:
Whatever the ultimate definition of "attempted" becomes, would the following make sense? There is a threshold config variable. If it is a number like 0, 0.7, 1, then it is a score threshold a set can meet and then grades get passed back. But it could also take string value "attempted", and then the threshold is whether or not an assignment has been attempted. Not the score that the set has.

Seeking balance between the complexity of that attempted-ness feature and having too many config settings.

Alex-Jordan:
I'm about ready to push a new attempt at all this. There is one last thing to decide.

Suppose you declare the config variables so that you will force grades be passed back after the close date, and prior to that grades will be sent if a certain threshold is met. What should be done with a test? Regarding the threshold, I think if the threshold is met for any of the student's versions, that is enough to meet the threshold.

But what about the date? Should the close date of the user's test template set be used for that purpose? The earliest close date among all of the versions? Something else?

somiaj:
For tests, I would use the close date of the first version (as this should also be the earliest date), if it exists, otherwise use the template set date. To me the grade should be sent back after the first version is completed.

I do understand part of the design is to not push grades back to the LMS too soon to lesson anxiety of students, and if someone has tests setup that students can take multiple test versions maybe sending back the grade after the first version is closed might not be desired. One compromise might be have two options for the configuration variable, 'close' vs 'close-template'. Where 'close-template' would only ever check the template close date. Another option is the LMS (Canvas allows this, I suspect this is common to all LMSes) to hide grades from students (This is what I do) until they are ready to release them (WeBWorK doesn't have to solve things that can be done in the LMS in my opinion).

somiaj:
Note, I'm avoiding the logic of testing if all possible versions have been taken and then using the close date of the last possible version (if it exists). I think for instructors who want to keep test grades from being pushed back to the LMS the simplicity of a 'close-template' option should at least allow delaying sending back low test scores until the end of the test window.

Alex-Jordan:
I think there's a difference between earliest close date and close date of the first version, considering that the instructor could change dates in places in between takings, and the student might use an achievement item. It would improve efficiency to only examine the date for the first version, but would it be safer to actually hunt the earliest close date?

My feeling is that it is fine to start sending test scores back once there is one version that has closed. Even if there are (or will be) additional versions. The student is more likely to understand that OK that was my score, and I can improve it with a new version. It's different than with a homework set where there is never a clear "I'm done now" moment. That's when students get grades showing up in the LMS and for them, it can reasonably feel too early and the grade is unfairly lowered.

drgrice1:
It is not safe to use the first test version. It is not guaranteed to even be the first version take. If a test version is deleted, then it may be replaced with a test version taken later.

somiaj:
Oh yea, forgot about deleting versions or a user overriding a specific version. Also if a student starts a version right before the template is closed, the version close date could be after the template close date (something to consider if you decide to add a 'close-template' option). I was hoping you could avoid looping over the versions, but guess not. Thanks for the clarification.

Alex-Jordan:
In working on this, I noticed that `grade_gateway` in `lib/WeBWorK/Utils/Sets.pm` has an unused second argument, `$set`. Do I need to leave that there in case external scripts use this, and call it with the second argument?

somiaj:
Since `$set` is in the middle of the args, removing it would require updating all places `grade_gateway` is called. I can't say if external scripts would be using this though. Or if this is something we should worry about, none of our scripts in `bin/` seem to use it.

Alex-Jordan:
Yes, I got that much. I was halfway through patrolling all the places it is used by `webwork2`, removing those second arguments. But then I realized something external might call it, which led to the post.

Alex-Jordan:
Closing this, because with the new version I will push, so much here is now a distraction. A new PR coming soon and I will reference this one.

somiaj:
@Alex-Jordan I say just remove it. Here is the commit that created the `grade_gateway` method, and it never used the `$set` even back then (2016). I doubt third party scripts are using this method, and `$set` was only there as it was missed when the original method was split apart.

https://github.com/openwebwork/webwork2/commit/c100f790c2f59301ddd4b14f576bdcf808bb18b3

Alex-Jordan:
I will, but only after @drgrice1 weighs in. There was something similar in this thread already where Glenn implied such changes should not be made.

Alex-Jordan:
OK, I'll remove it in #2617 then.

PR #2615, opened 2024-11-06
Split up macros on PODViewer index page.

somiaj:
This splits up the macros on the PODViwer index page to be grouped by their location in the `pg/macros` directory. In addition macros are listed before the libraries. The two macros not in a subdirectory, `PG.pl` and `PGcourse.pl`, are grouped with the pod in `pg/doc`. This makes finding the POD for a specific macro easier.

### DIFF
--- a/bin/dev_scripts/PODtoHTML.pm
+++ b/bin/dev_scripts/PODtoHTML.pm
@@ -31,11 +31,21 @@ use POSIX qw(strftime);
 use WeBWorK::Utils::PODParser;
 
 our @sections = (
-	bin    => 'Scripts',
-	conf   => 'Config Files',
 	doc    => 'Documentation',
+	bin    => 'Scripts',
+	macros => 'Macros',
 	lib    => 'Libraries',
-	macros => 'Macros'
+);
+our %macro_names = (
+	answers    => 'Answers',
+	contexts   => 'Contexts',
+	core       => 'Core',
+	deprecated => 'Deprecated',
+	graph      => 'Graph',
+	math       => 'Math',
+	misc       => 'Miscellaneous',
+	parsers    => 'Parsers',
+	ui         => 'User Interface'
 );
 
 sub new {
@@ -43,21 +53,16 @@ sub new {
 	my $class = ref $invocant || $invocant;
 
 	my @section_list  = ref($o{sections}) eq 'ARRAY' ? @{ $o{sections} } : @sections;
-	my @macros_list   = ref($o{macros}) eq 'ARRAY'   ? @{ $o{macros} }   : ();
 	my $section_hash  = {@section_list};
-	my $macros_hash   = {@macros_list};
 	my $section_order = [ map { $section_list[ 2 * $_ ] } 0 .. $#section_list / 2 ];
-	my $macros_order  = @macros_list ? [ map { $macros_list[ 2 * $_ ] } 0 .. $#macros_list / 2 ] : [];
 	delete $o{sections};
-	delete $o{macros};
 
 	my $self = {
 		%o,
 		idx           => {},
 		section_hash  => $section_hash,
 		section_order => $section_order,
-		macros_hash   => $macros_hash,
-		macros_order  => $macros_order,
+		macros_hash   => {},
 	};
 	return bless $self, $class;
 }
@@ -140,12 +145,7 @@ sub update_index {
 	if ($subdir eq 'macros') {
 		$idx->{macros} = [];
 		if ($pod_name =~ m!^(.+)/(.+)$!) {
-			my $macros = $self->{macros_hash};
-			if ($macros->{$1}) {
-				push @{ $idx->{$1} }, [ $html_rel_path, $2 ];
-			} else {
-				warn "no macro for '$pod_name'\n";
-			}
+			push @{ $self->{macros_hash}{$1} }, [ $html_rel_path, $2 ];
 		} else {
 			push @{ $idx->{doc} }, [ $html_rel_path, $pod_name ];
 		}
@@ -171,7 +171,8 @@ sub write_index {
 			sections      => $self->{section_hash},
 			section_order => $self->{section_order},
 			macros        => $self->{macros_hash},
-			macros_order  => $self->{macros_order},
+			macros_order  => [ sort keys %{ $self->{macros_hash} } ],
+			macro_names   => \%macro_names,
 			date          => strftime('%a %b %e %H:%M:%S %Z %Y', localtime)
 		}
 	);
--- a/bin/dev_scripts/PODtoHTML.pm
+++ b/bin/dev_scripts/PODtoHTML.pm
@@ -43,15 +43,21 @@ sub new {
 	my $class = ref $invocant || $invocant;
 
 	my @section_list  = ref($o{sections}) eq 'ARRAY' ? @{ $o{sections} } : @sections;
+	my @macros_list   = ref($o{macros}) eq 'ARRAY'   ? @{ $o{macros} }   : ();
 	my $section_hash  = {@section_list};
+	my $macros_hash   = {@macros_list};
 	my $section_order = [ map { $section_list[ 2 * $_ ] } 0 .. $#section_list / 2 ];
+	my $macros_order  = @macros_list ? [ map { $macros_list[ 2 * $_ ] } 0 .. $#macros_list / 2 ] : [];
 	delete $o{sections};
+	delete $o{macros};
 
 	my $self = {
 		%o,
 		idx           => {},
 		section_hash  => $section_hash,
 		section_order => $section_order,
+		macros_hash   => $macros_hash,
+		macros_order  => $macros_order,
 	};
 	return bless $self, $class;
 }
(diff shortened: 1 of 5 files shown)


lib/WeBWorK/ContentGenerator/PODViewer.pm
```diff
@@ -30,15 +30,37 @@ use WeBWorK::Utils::PODParser;
...
 
-	return $c->render('ContentGenerator/PODViewer', sections => $sections, sidebar_title => $c->maketext('Categories'));
+	my $macros = {};
+	for (sort keys %$macroFiles) {
+		my $macro = $macroFiles->{$_} =~ s!^$pgRoot/macros/(.+)/.+$!$1!r;
+		if ($macro =~ /^$pgRoot/) {
+			push(@$docs, $macroFiles->{$_} =~ s!^$pgRoot/!!r);
+		} else {
+			push(@{ $macros->{$macro} }, $macroFiles->{$_} =~ s!^$pgRoot/macros/$macro/!!r);
+		}
```
drgrice1, line 49: Using a substitution to find a match is a bit off.  I would suggest instead using a match instead, since that is what it is made for.
```suggestion
		if ($macroFiles->{$_} =~ m!^$pgRoot/macros/(.+)/.+$!) {
			push(@{ $macros->{$1} }, $macroFiles->{$_} =~ s!^$pgRoot/macros/$1/!!r);
		} else {
			push(@$docs, $macroFiles->{$_} =~ s!^$pgRoot/!!r);
		}
```

drgrice1 (changes requested):
There are two issues with this pull request.

The first is that this changes how the POD is organized inconsistent with how the pod generated by the `bin/dev_scripts/generate-ww-pg-pod.pl` script is organized (that is the pod displayed at https://webwork.maa.org/pod/).

The second is that the reorganization of the files into sub directories was not intended to be a visible categorization of macros.  At least not initially.  It was done more for development purposes to keep the `pg/macros` directory from being a completely unorganized mess with one long list of files.  The organization that was done was not the most intuitive in terms of what macros are actually used for.  For example, all macros that started with "parser" were just put in the "parsers" subdirectory, and all macros that started with "context" into the "contexts" subdirectory.   Sorting purely by what the macro started with is not the best way to do that.  The intent was to sort things better eventually.  For example. the `parserGraphTool.pl` macro was later moved to the "graph" subdirectory.  I am not sure that the organization is ready to be publicized in this way.

So this needs discussion first.

somiaj:
I think we are already publicizing the macro organization this way. The macros are currently sorted by what directory they are in, further that directory is printed in the link name. So knowing what directory a macro is in helps one find it in the long list of macros on the POD index (this mostly just splits that list apart). Though I do welcome more discussion on this to have a way that users can easily find the POD of a macro they are using or find a new macro they may want to use.

drgrice1:
We are currently listing them with their directory for ease of location.  The sorting will then list those in the same directory together.  That is not exactly the same thing as a categorization.  As I said, this needs discussion.  In any case, my first point is more pressing.  The dev script will need to be updated to match.  That is if we decide we want this organization.

lib/WeBWorK/ContentGenerator/PODViewer.pm
somiaj: Thanks, I updated the code here.

somiaj:
Updated `dev_scripts/generate-ww-pg-pod.pl` to match splitting up the macros link.

bin/dev_scripts/pod-templates/category-index.mt
```diff
@@ -23,18 +23,45 @@
...
-	% my ($index, $content) = ('', '');
+	% my ($index, $macro_index, $content, $macro_content) = ('', '', '', '');
+	% for my $macro (@$macros_order) {
+		% next unless defined $pod_index->{$macro};
+		% my $new_index = begin
+			<a href="#macro-<%= $macro %>" class="nav-link"><%= $macros->{$macro} %></a>
+		% end
+		% $macro_index .= $new_index->();
+		% my $new_content = begin
+			<h3><a href="#_podtop_" id="<%= $macro %>"><%= $macros->{$macro} %></a></h3>
```
drgrice1, line 34: Either `macro-` needs to be deleted from the `href` anchor, or `macro-` added to the `id` on line 34.  The links in the sidebar are not working.

somiaj: Thanks. Updated.

bin/dev_scripts/generate-ww-pg-pod.pl
```diff
@@ -96,18 +96,39 @@ sub process_dir {
...
+	$dest_dir =~ s/^$webwork_root/$output_dir\/webwork2/ unless $is_pg;
+	$dest_dir =~ s/^$pg_root/$output_dir\/pg/ if $is_pg;
 
 	remove_tree($dest_dir);
 	make_path($dest_dir);
 
+	my $sections =
+		$is_pg
+		? [ doc => 'Documentation', macros => 'Macros',        lib => 'Libraries' ]
+		: [ bin => 'Scripts',       doc    => 'Documentation', lib => 'Libraries' ];
```
drgrice1, line 110: The scripts should appear on the PG POD page for the wiki.  This is one place that the POD in webwork2 differs from the wiki.

drgrice1: The POD in webwork2 is intended to be for problem authors working in the PG problem editor of webwork2.  The scripts don't need to be shown for that.  However, they are something that authors could use, and should be represented on the wiki.

somiaj: Thanks again, I overlooked they were included on the wiki when I was comparing things.

drgrice1:
I don't really like that the macro subdirectories are hard coded into this (in both `templates/ContentGenerator/PODViewer.html` and the `generate-ww-pod.pl` script.  That is a maintenance hassle.  Every time a directory is added, removed, or renamed, then this code has to be updated.

bin/dev_scripts/generate-ww-pg-pod.pl
somiaj: Do you have an opinion of if the scripts should go first or not. I felt doc should be first here, but can make this match current behavior.

drgrice1: They can be after the documentation.

somiaj:
I realized that, but also felt that alphabetical order wasn't the best way to order the macros by, so I went with this so the order can be better controlled. So to me the question is the hassle worth having a different ordering than alphabetical.

drgrice1:
Actually, I think it would be better to go with alphabetical anyway.  I didn't realize you were ordering them in a particular way.  Why have you chosen that order anyway?  The only subdirectory that would be nice to move down is the "deprecated" directory.  There is no real reason to order the rest.

somiaj:
Okay, updated to not have a pre defined list for the order and macros included. There is still a hash of macro names that will include the translated and/or custom name, so "User Interface" is shown instead of "ui". Any new or renamed directory not included in that hash will still be included, just won't have a custom name attached to it.

bin/dev_scripts/PODtoHTML.pm
```diff
@@ -31,11 +31,22 @@ use POSIX qw(strftime);
...
 
 our @sections = (
-	bin    => 'Scripts',
-	conf   => 'Config Files',
 	doc    => 'Documentation',
+	bin    => 'Scripts',
+	macros => 'Macros',
 	lib    => 'Libraries',
-	macros => 'Macros'
+	conf   => 'Config Files',
```
drgrice1, line 38: You can delete `conf` from this list.  At one point there was something in that directory that had POD (I think), but whatever it was is gone, and nothing in the `conf` directory should ever have POD again.

bin/dev_scripts/PODtoHTML.pm
```diff
@@ -131,7 +143,15 @@ sub update_index {
 	$subdir =~ s|/.*$||;
 	my $idx      = $self->{idx};
 	my $sections = $self->{section_hash};
-	if (exists $sections->{$subdir}) {
+	if ($subdir eq 'macros') {
+		$idx->{macros} = [];
+		if ($pod_name =~ m!^(.+)/(.+)$!) {
+			my $macros = $self->{macros_hash};
+			push @{ $macros->{$1} }, [ $html_rel_path, $2 ];
```
drgrice1, line 150: This could just be
```suggestion
			push @{ $self->{macros_hash}{$1} }, [ $html_rel_path, $2 ];
```

drgrice1 (approved):
This is better now.  At least all directories will be represented.  I am still not sure of the benefits of providing alternate labels for the directories and even translating them. That tacitly assumes the directory categorization has more meaning than the haphazard sorting of macros that it mostly is at this point.

bin/dev_scripts/PODtoHTML.pm
somiaj: Suggested updates applied.

Alex-Jordan (approved):
I like that this gives you one level deeper of navigation in the left.

PR #2616, opened 2024-11-08
Remove mention of adding instructors to admin course in help.

somiaj:
Instructors are no longer added to the admin course when creating a new course, so remove mention of this from the admin help.


Alex-Jordan (approved):

drgrice1 (approved):
Thanks.  I missed that.

PR #2618, opened 2024-11-09
Add course admin tool to manage OTP secrets.

somiaj:
Adds a page in the admin course that can be used to copy OTP secrets from one user to another or to reset OTP secrets. This allows copying an OTP secret from a single user in one course to a different user in the same or different course. Copying multiple secrets from users in a single course to one or more other courses, provided the user names are the same. Or to reset one or more OTP secrets for users in a single course.

### DIFF
--- a/htdocs/js/CourseAdmin/manage_otp_secrets.js
+++ b/htdocs/js/CourseAdmin/manage_otp_secrets.js
@@ -1,0 +1,34 @@
+(() => {
+	// Save user menus to be updated.
+	const sourceSingleUserMenu = document.getElementById('sourceSingleUserID');
+	const destSingleUserMenu = document.getElementById('destSingleUserID');
+	const sourceMultipleUserMenu = document.getElementById('sourceMultipleUserID');
+	const destResetUserMenu = document.getElementById('destResetUserID');
+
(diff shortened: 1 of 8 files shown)


drgrice1 (approved):
I think that this looks good, and will be a useful tool for system administrators in some cases.

There is a situation that is perhaps not so good.  Basically, when this new page loads it reads all users for all courses from the database and the user dropdowns then contain all of those users.  If there are a large number of users and courses that is troubling.  The database read is still rather quick, but the dropdown loading is not.  I have a course with more than 20,000 users, and the page loads relatively quickly, but if I select that course, and open the users dropdown I see the CPU usage spike rather heavily as the browser loads the contents.

somiaj:
We could put some protection on that (or at least a warning), similar to the accounts page, only show some users (maybe just any user that isn't a student) and have a toggle they can hit so they include all users. In general having 20,000 students, this page would be a bit annoying to have to search for and find the few students you wanted to copy anyways.

somiaj:
I added a check to only include smaller courses (less than 200 users) by default, and a warning and a button to show all courses is shown if any courses are ignored/skipped.

PR #2619, opened 2024-11-09
Updates to adding users to newly created courses.

somiaj:
When adding a course on the course admin page, list all users in the admin course and let the admin select which users to copy over to the new course (by default all admin users are selected). This way an admin can select an existing user to copy over which can include any password or OTP secrets setup for that use (currently the admin will have to copy that over to the admin course manually). This also allows only copying a subset of admins to a newly created course instead of all.

When creating a new user to add to a course, be consistent with adding a user on the accounts management page, which doesn't require a password (useful when using external auth), email, etc. Now the only required field is the new user ID.

Flag the new user password input as 'new-password', so webbrowsers don't try to auto fill it in. Add the student ID as a field they can fill out instead of making it equal to the user ID. Since users can be copied from the admin course, add an option to add the new user to the admin course (as a dropped user so they can't login) so admins can select the same user to add to future courses as needed.

Remove the empty list of sets to assign users to when adding new users on the accounts management page in the admin course.

Make it so all users are shown on the Email page in the admin course, since most instructors are "Dropped" to still allow sending them emails.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -259,22 +259,15 @@ sub add_course_validate ($c) {
 	}
 
 	for (1 .. $number_of_additional_users) {
-		my $userID          = trim_spaces($c->param("add_initial_userID_$_"))          || '';
-		my $password        = trim_spaces($c->param("add_initial_password_$_"))        || '';
-		my $confirmPassword = trim_spaces($c->param("add_initial_confirmPassword_$_")) || '';
+		my $userID = trim_spaces($c->param("add_initial_userID_$_")) || '';
 
-		if ($userID ne '') {
-			unless ($userID =~ /^[\w-.,]*$/) {
-				push @errors,
-					$c->maketext(
-						'User ID number [_1] may only contain letters, numbers, hyphens, periods, commas, '
-						. 'and underscores.',
-						$_
-					);
-			}
-			if ($password ne '' && $password ne $confirmPassword) {
-				push @errors, $c->maketext('Pasword number [_1] and its password confirmation must match.', $_);
-			}
+		unless ($userID =~ /^[\w-.,]*$/) {
+			push @errors,
+				$c->maketext(
+					'User ID number [_1] may only contain letters, numbers, hyphens, periods, commas, '
+					. 'and underscores.',
+					$_
+				);
 		}
 	}
 
@@ -337,21 +330,21 @@ sub do_add_course ($c) {
 		my $PermissionLevel = $db->getPermissionLevel($userID);
 		my $User            = $db->getUser($userID);
 		my $Password        = $db->getPassword($userID);
-		$User->status('O');    # Add admin course user as an observer.
+
+		# Enroll student users, and make all other users observers.
+		$User->status($PermissionLevel->permission == $ce->{userRoles}{student} ? 'C' : 'O');
 
 		push @users, [ $User, $Password, $PermissionLevel ];
 	}
 
 	# add additional instructors if desired
 	for (1 .. $number_of_additional_users) {
-		my $userID          = trim_spaces($c->param("add_initial_userID_$_"))          // '';
-		my $password        = trim_spaces($c->param("add_initial_password_$_"))        // '';
-		my $confirmPassword = trim_spaces($c->param("add_initial_confirmPassword_$_")) // '';
-		my $firstName       = trim_spaces($c->param("add_initial_firstName_$_"))       // '';
-		my $lastName        = trim_spaces($c->param("add_initial_lastName_$_"))        // '';
-		my $email           = trim_spaces($c->param("add_initial_email_$_"))           // '';
-		my $studentID       = trim_spaces($c->param("add_initial_studentID_$_"))       // '';
-		my $permissionLevel = trim_spaces($c->param("add_initial_permission_$_"));
+		my $userID          = trim_spaces($c->param("add_initial_userID_$_"))    // '';
+		my $password        = trim_spaces($c->param("add_initial_password_$_"))  // '';
+		my $firstName       = trim_spaces($c->param("add_initial_firstName_$_")) // '';
+		my $lastName        = trim_spaces($c->param("add_initial_lastName_$_"))  // '';
+		my $email           = trim_spaces($c->param("add_initial_email_$_"))     // '';
+		my $permissionLevel = $c->param("add_initial_permission_$_");
 		my $add_user        = $c->param("add_initial_user_$_") // 0;
 
 		if ($userID =~ /\S/) {
@@ -360,8 +353,7 @@ sub do_add_course ($c) {
 				first_name    => $firstName,
 				last_name     => $lastName,
 				email_address => $email,
-				student_id    => $studentID,
-				status        => 'O',
+				status        => $permissionLevel == $ce->{userRoles}{student} ? 'C' : 'O',
 			);
 			my $Password = $db->newPassword(
 				user_id  => $userID,
@@ -385,7 +377,7 @@ sub do_add_course ($c) {
 					$db->addUser($User);
 					$db->addPassword($Password);
 					$db->addPermissionLevel($PermissionLevel);
-					$User->status('O');
+					$User->status($permissionLevel == $ce->{userRoles}{student} ? 'C' : 'O');
 				}
 			}
 		}
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -293,7 +293,6 @@ sub do_add_course ($c) {
 	my $add_initial_firstName       = trim_spaces($c->param('add_initial_firstName'))       // '';
 	my $add_initial_lastName        = trim_spaces($c->param('add_initial_lastName'))        // '';
 	my $add_initial_email           = trim_spaces($c->param('add_initial_email'))           // '';
-	my $add_initial_studentID       = trim_spaces($c->param('add_initial_studentID'))       // '';
 	my $add_initial_user            = $c->param('add_initial_user')                         // 0;
 
 	my $copy_from_course = trim_spaces($c->param('copy_from_course')) // '';
@@ -337,7 +336,6 @@ sub do_add_course ($c) {
 			user_id       => $add_initial_userID,
 			first_name    => $add_initial_firstName,
 			last_name     => $add_initial_lastName,
-			student_id    => $add_initial_studentID,
 			email_address => $add_initial_email,
 			status        => 'O',
 		);
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -230,17 +230,18 @@ sub pre_header_initialize ($c) {
 }
 
 sub add_course_form ($c) {
+	$c->param('number_of_additional_users', ($c->param('number_of_additional_users') // 0) + 1)
+		if $c->param('add_another_instructor');
+
 	return $c->include('ContentGenerator/CourseAdmin/add_course_form');
 }
 
 sub add_course_validate ($c) {
 	my $ce = $c->ce;
 
-	my $add_courseID                = trim_spaces($c->param('new_courseID'))                || '';
-	my $add_initial_userID          = trim_spaces($c->param('add_initial_userID'))          || '';
-	my $add_initial_password        = trim_spaces($c->param('add_initial_password'))        || '';
-	my $add_initial_confirmPassword = trim_spaces($c->param('add_initial_confirmPassword')) || '';
-	my $add_dbLayout                = trim_spaces($c->param('add_dbLayout'))                || '';
+	my $add_courseID               = trim_spaces($c->param('new_courseID'))  || '';
+	my $number_of_additional_users = $c->param('number_of_additional_users') || 0;
+	my $add_dbLayout               = trim_spaces($c->param('add_dbLayout'))  || '';
 
 	my @errors;
 
@@ -257,11 +258,24 @@ sub add_course_validate ($c) {
 		push @errors, $c->maketext('Course ID cannot exceed [_1] characters.', $ce->{maxCourseIdLength});
 	}
 
-	if ($add_initial_userID ne ''
-		&& $add_initial_password ne ''
-		&& $add_initial_password ne $add_initial_confirmPassword)
-	{
-		push @errors, $c->maketext('The password and password confirmation for the instructor must match.');
+	for (1 .. $number_of_additional_users) {
+		my $userID          = trim_spaces($c->param("add_initial_userID_$_"))          || '';
+		my $password        = trim_spaces($c->param("add_initial_password_$_"))        || '';
+		my $confirmPassword = trim_spaces($c->param("add_initial_confirmPassword_$_")) || '';
+
+		if ($userID ne '') {
+			unless ($userID =~ /^[\w-.,]*$/) {
+				push @errors,
+					$c->maketext(
+						'User ID number [_1] may only contain letters, numbers, hyphens, periods, commas, '
+						. 'and underscores.',
+						$_
+					);
+			}
+			if ($password ne '' && $password ne $confirmPassword) {
+				push @errors, $c->maketext('Pasword number [_1] and its password confirmation must match.', $_);
+			}
+		}
 	}
 
 	if ($add_dbLayout eq '') {
@@ -283,17 +297,10 @@ sub do_add_course ($c) {
 	my $db    = $c->db;
 	my $authz = $c->authz;
 
-	my $add_courseID          = trim_spaces($c->param('new_courseID')) // '';
-	my $add_courseTitle       = ($c->param('add_courseTitle')       // '') =~ s/^\s*|\s*$//gr;
-	my $add_courseInstitution = ($c->param('add_courseInstitution') // '') =~ s/^\s*|\s\*$//gr;
-
-	my $add_initial_userID          = trim_spaces($c->param('add_initial_userID'))          // '';
-	my $add_initial_password        = trim_spaces($c->param('add_initial_password'))        // '';
-	my $add_initial_confirmPassword = trim_spaces($c->param('add_initial_confirmPassword')) // '';
-	my $add_initial_firstName       = trim_spaces($c->param('add_initial_firstName'))       // '';
-	my $add_initial_lastName        = trim_spaces($c->param('add_initial_lastName'))        // '';
-	my $add_initial_email           = trim_spaces($c->param('add_initial_email'))           // '';
-	my $add_initial_user            = $c->param('add_initial_user')                         // 0;
+	my $add_courseID               = trim_spaces($c->param('new_courseID')) // '';
+	my $add_courseTitle            = ($c->param('add_courseTitle')       // '') =~ s/^\s*|\s*$//gr;
+	my $add_courseInstitution      = ($c->param('add_courseInstitution') // '') =~ s/^\s*|\s\*$//gr;
+	my $number_of_additional_users = $c->param('number_of_additional_users') || 0;
 
 	my $copy_from_course = trim_spaces($c->param('copy_from_course')) // '';
 
@@ -314,12 +321,17 @@ sub do_add_course ($c) {
 			));
 			next;
 		}
-		if ($userID eq $add_initial_userID) {
-			$c->addbadmessage($c->maketext(
-				'User "[_1]" will not be copied from the [_2] course as it is the initial instructor.', $userID,
-				$ce->{admin_course_id}
-			));
-			next;
+		for (1 .. $number_of_additional_users) {
+			my $add_initial_userID = trim_spaces($c->param("add_initial_userID_$_")) // '';
+
+			if ($userID eq $add_initial_userID) {
+				$c->addbadmessage($c->maketext(
+					'User "[_1]" will not be copied from the [_2] course as it is the same as additional user '
+						. 'number [_3].',
+					$userID, $ce->{admin_course_id}, $_
+				));
+				next;
+			}
 		}
 
 		my $PermissionLevel = $db->getPermissionLevel($userID);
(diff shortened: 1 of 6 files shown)


Alex-Jordan:
This sounds nice. I have "math" admin users in the admin course, as well as "chemistry" admin users. It will be nice to have a way to easily exclude one set or the other from new courses.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -349,19 +337,35 @@ sub do_add_course ($c) {
...
+				$c->addbadmessage($c->maketext(
+					'User "[_1]" will not be added to [_2] course as it already exists.', $add_initial_userID,
+					$ce->{admin_course_id}
+				));
+			} else {
+				$User->status('D');    # By default don't allow user to login.
+				$db->addUser($User);
+				$db->addPassword($Password);
+				$db->addPermissionLevel($PermissionLevel);
+				$User->status('O');
```
Alex-Jordan, line 366: Is this line meant to be here?

Alex-Jordan:
In Firefox, I am seeing a scroll border at the bottom even though horizontal scrolling is not in play:

Screenshot 2024-11-10 at 3 47 42 PM

Does changing the `overflow-scroll` to `overflow-scroll-y` make sense?


Can we change "Add new user to admin course." to something like "Also add this new user the ___ course." where the blank is the actual name of the admin course?

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
somiaj: Yes, the status needs to be set back to observer before that user is copied into the new course (otherwise they would be copied into the new course as dropped).

Though I have been thinking that maybe dropping users by default isn't the best approach here (I just don't want them able to log into the admin course). Instead the admin course could probably be configured that only admins can log into it via a course configuration, but this would have to be done in `course.conf`.

somiaj:
Strange, in firefox here I only see the vertical scroll bar, do you have a user with a really long name down below causing the scroll bar to appear. I guess I could do `overflow-scroll-y`, but then if there is a really long `user name (role)` it won't be possible to see it.

Yea, I'll update it to use the admin course name.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
somiaj: I think I'll change it to not make users in the admin course dropped by default (maybe an admin wants to allow these users to login to the course so they can set passwords or setup two factor auth, to then be copied to new courses). Though we should probably add something about making a course.conf for the admin course that prevents anyone except admins from logging in as default if they decide to add non admin users to it.

drgrice1: I don't think that is is wise.  I think better would be to make the user dropped by default.  That can always be changed from the accounts manager if desired.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
```diff
@@ -39,13 +39,20 @@
...
 			<%= maketext(
-				'To add the WeBWorK administrators to the new course (as administrators) check the box below.') =%>
+				'Select admin course users to add to the new course (as the stated permission) below.') =%>
 		</div>
-		<div class="form-check">
-			<label class="form-check-label">
-				<%= check_box 'add_admin_users' => 1, class => 'form-check-input' =%>
-				<%= maketext('Add WeBWorK administrators to new course') =%>
-			</label>
+		<div class="col-lg-4 col-md-5 col-sm-6 overflow-scroll border rounded px-2" style="height: 100px;">
```
drgrice1, line 44: I think you should use `max-height` here instead of `height`.  Then if there are less than 5 users in the admin course (my admin courses never have more than 1) the box isn't so large.

drgrice1:
Change the `overflow-scroll` to `overflow-auto`.  Then the scroll bars will only show up if needed.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
```diff
@@ -39,13 +39,21 @@
...
+			% for my $user ($db->listUsers) {
+				% my $val   = $db->getPermissionLevel($user)->permission;
+				% my $role = maketext((grep { $ce->{userRoles}{$_} eq $val } keys %{ $ce->{userRoles} })[0]);
+				<div class="form-check">
+					<label class="form-check-label">
+						<%= check_box 'add-admin-users' => $user, class => 'form-check-input',
+							$val == 20 ? (checked => undef) : () =%>
```
drgrice1, line 52: You can't add `checked => undef` here like this.  This means that if the page is submitted (because the "Add Course" button is clicked) then when the page reloads whatever the user had selected will be forgotten, and the selections will be changed back to the defaults based on permission.  If there is an error in the provided data on the page, then the page reloads with the messages about what needs to be corrected, and these check boxes are then reset.  So now the user has to redo the selection.

You will need to set `add-amin-users` via the `param` method, and that should only be done on initial page load.  So this will take some work to fix.

drgrice1: Note that you can't just check if the `add-admin-user` param is previously defined either.  That is because if no check boxes are checked because that is what the user selected, then the parameter will also not be defined.

somiaj: Would adding a hidden field to check for be okay for this? I also found something odd with mojolicious, if I use `add_admin_users` with an underscore, any `checked => undef` or `selected => undef` get ignored. I wasn't sure what was going on, that is why I changed to dashes here compared to the id's used elsewhere.

somiaj:
Updated with the suggested changes. I reverted then enabled the dropping new users by default (they are still dropped by default). Though I did leave the comment about using `$permissionLevel{login}='admin'` in the `course.conf` file as a way to control access. That might be useful since there is no "Course Configuration" in the admin course.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
drgrice1: Adding a hidden field is the usual approach.  However, I am not sure if that will work here because this parameter has an array value.  You will have to test that.

somiaj: I just tested, if you add a new course successfully there is no link back to the add course page (do we want that?), so someone would have to hit back to in which the check boxes are still selected.

Further, the only way I could think of was to cause an error, so I created a course ID with a space in it, `New Course`, I selected a non-admin user, when I come back to the add course page with an error, that user is still selected, so I don't see the issue you are describing.

somiaj: Okay, I found a case that isn't covered, If I uncheck all admin users, and then cause an error on the page, when I come back the admin users are selected again. But that is the only case this happens.

drgrice1: There are lots of ways to cause an error.  Don't give a course id at all.  Add a user and give a password, and give a different confirmation password.

drgrice1: Note the issue only occurs if you uncheck all checkboxes.

drgrice1:
Do we have to add the student ID to the page?  That makes that part of the page look rather messy and unbalanced.

somiaj:
@drgrice1 I'm okay with it not being there, I also though about adding a comment field that could be used if adding the user to the admin course. But I'm fine with just removing it. I mostly didn't like the old behavior of making the studentID = userID, I would rather just leave it blank.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
somiaj: Okay, I found a simple solution that works in my tests. Since there isn't a link back to this page after the course is added, I'm unsure if it needs to be more robust, but any error that occurs won't change which boxes are checked in my testing.

Alex-Jordan:
I agree that having studentID default to userID is undesirable.

Also I rarely actually know the "studentID" of other instructors and I would not be able to enter it here, and would just leave it blank.

Not to explode the scope of this, but perhaps this whole device (where new instructors are added to a new course) could use a makeover. Why only be able to add one new user at this place? It's not uncommon to have two instructors, or one plus one or more ta~s. There could be something closer to what we see when adding a new user to a course using the Accounts Manger, all in one row, with the potential to click a button to add another row for a second additional user.

somiaj:
@Alex-Jordan this PR does make a way to do that, they just have to first add multiple users in the Accounts Manager page, then just select them. Might not cover the solution where the admin doesn't want users in the admin course. I'm unsure if having a way to add more rows would be overall worth it on this page or not.

somiaj:
I removed it, I too thought it looked odd, but thought maybe someone would want to use it. In general we don't have to do everything here, the accounts manager can be used to adjust things beyond what this page initially sets.

drgrice1:
I also don't like the student ID being set to the user ID.  I think it would be fine just to leave it unset here.

I also agree that being able to add multiple users might be nice.  That can be done later though.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
```diff
@@ -54,7 +62,7 @@
 				. 'The user ID may contain only numbers, letters, hyphens, periods (dots), commas,and underscores.'
 		) =%>
 	</div>
-	<div class="row mb-3">
+	<div class="row">
```
drgrice1, line 65: I think that instead of completely removing the margin here, the margin should just be made smaller.  I think that `mb-1` looks much better.  With no margin it looks to compact.  The "Also add this user to the admin course" checkbox looks to close to the inputs above.  All of the other places on the page `mb-1` or `mb-2` are used to still provide some separation, but `mb-3` is used to provide a larger visual separation.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
drgrice1: The hidden field that you added should work.  There is no need to worry about a link back to the page after the course is added.  There isn't such a link, and even if there was the form should not still have the data for the course that was already added in that case.  Going back to the add course page would be to add another course, and so the form should be reset.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
```diff
@@ -39,13 +39,21 @@
...
+				<div class="form-check">
+					<label class="form-check-label">
+						<%= check_box 'add-admin-users' => $user, class => 'form-check-input',
+							$val == 20 && !param('last_page_was_add_course') ? (checked => undef) : () =%>
+						<%= "$user ($role)" %>
+					</label>
+				</div>
+			% }
 		</div>
 	</div>
```
drgrice1, line 58: This is structurally incorrect.  Any `col` class element needs to be inside a div with the `row` class.  The `row` class should be added to the div on line 39 above.  However, that will mess with the border and padding that you have here.  Although, that isn't a good idea anyway.  The `col` class could potentially cause issues with the other classes and styles added here.  Furthermore, the current structure and styles are mismatched with the rest of the page.

I suggest that you change lines 39-60 to:
```perl
	<div class="row mb-3">
		<div class="mb-1">
			<%= maketext(
				'Select [_1] course users to add to the new course (as the stated permission) below.',
				$ce->{admin_course_id}) =%>
		</div>
		<div class="col-lg-4 col-md-5 col-sm-6">
			<div class="overflow-auto border rounded px-2" style="max-height: 100px;">
				% for my $user ($db->listUsers) {
					% my $val   = $db->getPermissionLevel($user)->permission;
					% my $role = maketext((grep { $ce->{userRoles}{$_} eq $val } keys %{ $ce->{userRoles} })[0]);
					<div class="form-check">
						<label class="form-check-label">
							<%= check_box 'add-admin-users' => $user, class => 'form-check-input',
								$val == 20 && !param('last_page_was_add_course') ? (checked => undef) : () =%>
							<%= "$user ($role)" %>
						</label>
					</div>
				% }
			</div>
		</div>
	</div>
```
That looks much better to me, and is structurally correct for bootstraps `row` and `col` classes.
```

**Review by drgrice1 — APPROVED** [review]
Other than the style issues that I commented on, this looks good.

**Comment by somiaj** [comment]
Okay, I have updated it so more than a single initial user can be added. There is now a button that can be used to add another instructor to the course, which extends the form. I have also added the ability to select the permission level
of the instructor, so TAs, etc can also be added this way (to match that TAs, etc can be copied from the admin course).

In doing this I noticed a small oversight. @Alex-Jordan had set it up so initial instructors would be assigned to sets and achievements if copping those over. Since now the instructor could be copied over, that would have skipped this, also there is now more than a single instructor that can be added. Instead I updated the logic to assign all sets and achievements to any user that is not an admin (as that appeared to be the original intention, since admins were being excluded from this).

If you would prefer I move this change to another PR, let me know, I don't have to included it here.

**Comment by somiaj** [comment]
Two things I'm unsure about, first is permission levels. I have hard coded that professor = 10, which I should probably avoid, but unsure how to best deal with this (what if someone renames the professor role). I do see that some places we do kinda assume that some roles are left alone, but not sure the best approach. But for admin's, I actually just check for the `admin` role, what if someone changes that?

Second, the action is written to the log file. It use to contain the first, last, and email of the new instructor. But now these are now optional and there could be multiple new instructors either copied from the admin course directly or created. I just went with not included that info in the log file. Unsure if I should list all the user IDs created and/or copied to the course or not.

**Comment by somiaj** [comment]
In testing this I notice another thing I don't know how to work around (maybe it is not possible). So when adding users and the "add another instructor" or an error occurs, any passwords are not saved and have to be reentered. I think this is probably desirable behavior, but it makes the idea of entering one user's info at a time then hitting 'add another instructor' not as convenient if you have to wait until the end to enter all their passwords.

**Comment by drgrice1** [comment]
Unfortunately I don't think that there is a work around.  Browsers are rather restrictive with password fields.

I think assigning sets to all but admin users should be fine.  Although, we could just make it assign sets to all users.  Is there any reason not to assign sets to admin users?  Often admin users need to access sets to deal with issues, so it may be more convenient to just do that.

**Comment by somiaj** [comment]
I thought the same thing, and I am often the only instructor and an admin user, so I wouldn't have sets assigned to me. I'll wait for @Alex-Jordan to respond, as I was just following what appeared to be their intent with only assigning sets to the newly created user.

**Comment by somiaj** [comment]
I got rid of the hardcoded 10 permission level, that was a mistake on my part, instead I assume that the key `professor` has not been changed, like `Problem.pm`. Not an issue for this PR, but we probably need a more robust way for this, as changing the roles of `student`, `professor`, or `admin` might not work correctly.

File: templates/ContentGenerator/CourseAdmin/add_course_form.html.ep [hunk 6]
```diff
@@ -36,71 +50,109 @@
 			</div>
 		</div>
 	</div>
-	<div class="mb-3">
+	<div class="row mb-3">
 		<div class="mb-1">
 			<%= maketext(
-				'To add the WeBWorK administrators to the new course (as administrators) check the box below.') =%>
+				'Select [_1] course users to add to the new course (as the stated permission) below.',
```
Alex-Jordan, line 56:
```suggestion
				'Select users from the [_1] course to add to the new course with the indicated permission.',
```

When the name of the admin course is "admin", the current wording is "Select admin course users..." and if you parse that sentence one way in your head it's asking for users with permission level "admin". And then it seems to not make sense with the list below that includes non-admin users.

I'm also suggesting a change to the tail of this sentence.

Alex-Jordan:
If I click the button twice to add an instructor, the layout feels a bit dense.

Screenshot 2024-11-13 at 9 58 55 AM

I tried changing the class on the checkbox to have a larger bottom margin with `mb-4`:

Screenshot 2024-11-13 at 10 01 45 AM

I also tried `mb-5`:

Screenshot 2024-11-13 at 10 06 02 AM

Do either of these seem better? There could be other solutions to the issue I'm perceiving. With the first image, I have a little uncertainty that the checkbox applies to the block above it versus below it.

Alex-Jordan:
I tried to break this, and I did. What I did was:

- checkbox a user that is in the admin course already
- add a new user by the same username
- copy non student users from the chosen template course, and that course also has a user by the same username

Some part of that created a big error message that starts `assignSetToGivenUsers: error during addMultipleUserSets: addMultipleUserSets errors: DB call error: [record exists] at /opt/webwork/webwork2/lib/WeBWorK/Utils/Instructor.pm line 398`.

somiaj:
Thanks for trying to break it. I did try to check for that case during validation, but maybe my logic isn't working. I'll look closer at this case.

somiaj:
As for the format, I also though about adding a `<hr>` between users to give some separation. I'm okay with adding some additional space, or go with an `<hr>` (maybe with some space `my-2` or `my-3` on it).

somiaj:
Oh wait, I see where the error is, I wonder if this issue isn't new, and if you would have added what you called an `initialUser` with the same name as a user that already existed, the error would have shown up. I will fix it here and go do an existence check.

Alex-Jordan:
I just tested and on a 2.19 server, and I don't get the error if I add a professor who is already an admin user in the admin course, as well as a user in the course where non student users are copied from.

In 2.19 code, there is:
https://github.com/openwebwork/webwork2/blob/beda37252233a43832837c6fde85fdd31a978a55/lib/WeBWorK/ContentGenerator/CourseAdmin.pm#L327
that prevents an admin course user with admin level permission from being copied to the new course if that's the same username that was entered to be the professor.

Then in `CourseManagement.pm`, this line:
https://github.com/openwebwork/webwork2/blob/beda37252233a43832837c6fde85fdd31a978a55/lib/WeBWorK/Utils/CourseManagement.pm#L365
skips over any users from the "copy from" course if they were already added because of either they were an admin from the admin course or they were the one new professor.

somiaj:
Admin users are protected against this, but the new initialUser is not, I just tested and in 2.19 I get a long line of errors if I try to add an initialUser that is the same as a user that exists starting with something like `addUserAchievement: user achievement exists (perhaps you meant to use putUserAchievement?)` so there is currently an issue in 2.19 very similar to this (should I look at a hotfix for 2.19 for that issue)?

I haven't fully tested, but that last commit should fix the issue, I remove any user from being assigned to sets or achievements that was copied from the old course. I also notice that there is protection against this already with eval { ... } and warns that probably only show up on the server logs (unsure where they would appear) if you try to add the user a second time.

somiaj:
Nevermind, just tested and I still get an error. Will dig deeper.

somiaj:
@Alex-Jordan Okay I think I fixed it. I also made it so all users (including admin users) now get assigned to everything when copying over the new course (I can remove admins from this if you would prefer).

One thing about the approach I went with, is if a user is being copied or added from the admin course, this means that users is assigned to every set and every achievement, and the sets/achievements they were assigned to in the old course is ignored.

Alex-Jordan:
That sounds good but I do think that this could not be done for admin users. I have some faculty who prefer to not see the admin users show up in the scoring tools output, and not assigning them assignments prevents that.

Does it also prevent that for "observer" users? If so, making sure they are added as observers would also be OK. Maybe better, since as an admin user helping someone in their course, it can frustrate things to not be assigned a particular set.

somiaj:
The observer status does exactly that. Observers are not included in any summary reports like scoring cvs files and the statistics page. Admin users added from the add course form have been made observers since that status was added (so 2.19 at least and maybe 2.18 too). So they are currently being made observers and nothing to do.

Note admin users will show up when they are on the page to assigns users to a set, and will get assigned new sets with assign all, but this would also be the case if they weren't assigned any sets. Since admins aren't taking quiz versions they won't show up when manually grading quizzes, but they might show up when manually grading assignments. One thing to do is put these users in a different section, and then one could filter by section if they don't want to have to skip over them if using the single problem grader. They will probably just have to skip over them if using the other problem grader.

templates/ContentGenerator/CourseAdmin/add_course_form.html.ep
somiaj: Thanks, I updated the language to your suggestion.

somiaj:
I played with adding a padded `<hr>` between users, would this look good, or think just a little more space is all that will be needed? (Note I only put the line between users, the bottom will just be the button as is)

image

Alex-Jordan:
I am not a fan of `hr` in places like this. The extra space is good enough for me.

somiaj:
What spacing should I increase it to? Of the two you gave, `mb-4` is probably what I would go with, I think the other is a bit too spaced out, but I don't have strong opinions.

somiaj:
@Alex-Jordan I added an "Enter information for additional instructor number X." to each additional instructor to clarify which inputs go to which instructor (also any errors will reference the number), which also breaks the inputs apart as well. Thoughts?

somiaj:
I was looking at the add user page under accounts manager, and passwords there are just normal text fields, not plain text. Should I make that the same here for the password fields, that way when the add another user button is pressed, the passwords are still there? Or should we also make the password fields passwords on when adding users in a normal course?

drgrice1:
I think that the best way to go is to make this page more like the add users page for the accounts manager.

Generally, a masked password field is for when you are creating a password for yourself, or at least for one user.  You really wouldn't want that for a long list of users.  The thing is that a masked password field needs to have an accompanying masked confirm password field.  That is really the point of the confirmation fields.  Since the password field is masked, you need to ensure that the created password is typed correctly.  You might have a typo when you type it the first time, and then if that is used you would get locked out of your account.  But it is unlikely you would have the same typo twice.  You really don't want to have to type the password and a confirmation password for each user in a long list of users.  Then if you have a typo in one and submit, you have to start all over again.

If you are editing a long list of users you should be doing that somewhere where you won't have anyone looking over your shoulder.

As such, that is probably what needs to be done here.  Say you add ten users when creating a course.  You now have to enter ten passwords twice with no typos.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -322,49 +313,86 @@ sub do_add_course ($c) {
...
-			password => cryptPassword($add_initial_password),
-		);
-		my $PermissionLevel = $db->newPermissionLevel(
-			user_id    => $add_initial_userID,
-			permission => '10',
-		);
+		my $PermissionLevel = $db->getPermissionLevel($userID);
+		my $User            = $db->getUser($userID);
+		my $Password        = $db->getPassword($userID);
+		$User->status('O');    # Add admin course user as an observer.
```
drgrice1, line 340: I think this should not set the status to observer for student users.  I just tested this and had created a student user in the admin course, and copied that user to the newly created course.  I saw that the student was added as an observer.  I don't think that should be.  Of course, usually student users should not be added in this way, but it is something that could be done.

This change does not need to be made for this pull request.  It is just something that I observed.

I still approve this pull request.

lib/WeBWorK/ContentGenerator/CourseAdmin.pm
```diff
@@ -260,24 +258,16 @@ sub add_course_validate ($c) {
...
-		}
-		if ($add_initial_lastName eq '') {
-			push @errors, $c->maketext('You must specify a last name for the initial instructor.');
-		}
-		if ($add_initial_email eq '') {
-			push @errors, $c->maketext('You must specify an email address for the initial instructor.');
+	for (1 .. $number_of_additional_users) {
+		my $userID = trim_spaces($c->param("add_initial_userID_$_")) || '';
+
+		unless ($userID =~ /^[\w-.,]*$/) {
```
drgrice1, line 264: I missed this before.  This is an invalid regular expression.  The `-` inside the braces needs to be at the end of all other characters.  Otherwise it is interpreted as a range, and the range from `\w` to `.` is not valid.

drgrice1: Note the `-` can also be at the beginning.

drgrice1: Or it can be escaped with a `\`.

somiaj: Thanks for spotting this.

PR #2620, opened 2024-11-11
Make the student nav filter selector not split toggle.

drgrice1:
A split toggle is implemented with two buttons.  A button with the text (in this case the filter name) and a button with the caret for the dropdown menu.  However, `student-nav-filter-selector` div only has one button and does not have the second button for the caret.  This causes issues with the border of the button.  Since the filter button has the `dropdown-toggle-split` class the `btn-group > .btn.dropdown-toggle-split` selector applies and the rounded borders on the right are removed.  That makes the button look wrong.

This just removes incorrect `dropdown-toggle-split` class that is causing the problem and makes it so the button is displayed correctly.

### DIFF
--- a/templates/ContentGenerator/GatewayQuiz/nav.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz/nav.html.ep
@@ -80,7 +80,7 @@
 			<div class="btn-group student-nav-filter-selector">
 				<%= link_to $filter ? $filters->{$filter}[0] : maketext('Showing all tests') => '#',
 						id              => 'testSelectorFilter',
-						class           => 'btn btn-primary dropdown-toggle dropdown-toggle-split',
+						class           => 'btn btn-primary dropdown-toggle',
 						role            => 'button',
 						data            => { bs_toggle => 'dropdown' },
 						'aria-expanded' => 'false' =%>


Alex-Jordan:
I don't use tests really and I'm not sure what filter this is referring to. Can you describe where I should go and what I should look at for testing?

drgrice1:
This is shown with the student nav when acting as a student.

Alex-Jordan:
I should know what you are talking about, but I don't.  This is only on a Gateway quiz page, or any page? What is "The student nav"?

drgrice1:
The student nav is added in the bar at the top of the page when acting as a student.  It is shown in both regular assignments and gateway tests.  It allows you to select which student you are acting as.  There is a filter for filtering which students are listed by section or recitation.  The button for regular sets does not have this issue though.

drgrice1:
Note that the button is not shown if you do not have sections or recitations.

somiaj (approved):
The dropdown for the section in gateways now match that of normal problems.

Alex-Jordan:
OK. There's one more thing. You have to have at least two users who have taken the test. I was trying this by making a test, having a different user take it, then returning to my admin user to act as them and see. But the filter button just was not there. Having a second user take the test changed that.

Now that I can see that button, it works as described.

PR #2621, opened 2024-11-13
Fix the comment about the LTI 1.1 consumer key in the authen_LTI_1_1.conf.dist file.

drgrice1:
The LMS course ID is not the issue.  It is the LMS context id.  That can be the same for two courses from different LMS's.   So the `ConsumerKey` is used in that case to distinguish.  The course id could actually even be the same for two courses from the same LMS.  However, that isn't used for this at all.


somiaj (approved):

somiaj:
I don't really see a reason to wait for a second approval, but @Alex-Jordan seems to be around, let's see if they can beat me to the merge again.

Alex-Jordan (approved):

PR #2622, opened 2024-11-13
Fix uninitialized variable warning when not using LTI.

somiaj:
Check that `$ce->{LTIVersion}` is defined and not empty before using it. This variable being initialized should ensure that the other LTI variables used here are also initialized.


drgrice1 (approved):
Good catch.  I don't know how I missed this when this change was added.  Probably the fact that I typically have LTI authentication enabled.

somiaj:
@drgrice1 Think we can remove the line `&& defined $ce->{LTIGradeMode}`, since if `$ce->{LTIVersion}` is defined, this not being defined is probably an error?

Issue #2598, opened 2024-10-09
No alt-text for graphs on the Statistics page


dlglin:
This came up from the accessibility review obtained from @glarose.

There is no alt-text for the graph images that are dynamically generated on the Statistics page.

Possible remedies:

1. Create alt-text for each graph.  This would be theoretically possible, but could be quite tedious to create.
2. Include the raw data used to generate the graph, either as alt-text for the image itself, or as additional content on the page (perhaps hidden by default).

Not sure how pressing this is since this is an instructor-only page and is not required for any core functionality.

somiaj:
Fixed in #2624

PR #2624, opened 2024-11-17
Add alt text to histograms on stats page.

somiaj:
This adds alt text to the histogram images on the stats page which fixes the issue mentioned in #2598.

In addition this fixes some issues I found with the stats page.
* Links on the problem bars weren't working so forced them to be a string.
* JitarSets problem stats - The adjusted status bars were reversed and didn't need to be and the adjusted scores were multiplet by 100 and didn't need to be.

Should the additional fixes be made into a hot fix, as these are present in 2.19 too?

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/Stats.pm
@@ -550,6 +550,7 @@ sub build_bar_chart ($c, $data, %options) {
 		mainTitle    => 'ERROR: This must be set',
 		xTitle       => '',
 		yTitle       => '',
+		descText     => '',
 		barWidth     => 22,
 		barSep       => 4,
 		barFill      => 'rgb(0,153,198)',
@@ -585,6 +586,7 @@ sub build_bar_chart ($c, $data, %options) {
 	);
 
 	# Main graph setup.
+	$svg->desc()->cdata($opts{descText});
 	$svg->rect(
 		id               => "bar_graph_window_$id",
 		x                => 0,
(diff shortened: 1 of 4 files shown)


somiaj:
I should note, it would be nice if someone who has actual Jitar set data can test this out. I only have normal set data that I transformed into a jitar set. It just felt like the bars on the problems page were backwards, maybe this won't be the case with real jitar data (not quite sure how jitar adjusted grade after review works). But the graphs I saw all seemed mirrored, so I removed the reverse. This way the adjusted data looks the same in my test case on both the set and problem stats page.

drgrice1:
Note that this is an inline `svg`, so the `img` patterns listed on that site are not an option.  It looks like the only options that work reliably are patterns 5 and 8.  Although those patterns do not have both a title and a description.  It seems that none of the patters with both a title and a description work reliably for all browsers.  Although pattern 11 works to give both, but the description is repeated twice in some cases.

The warning about the results with screen readers being outdated may apply though, and some cases may have been fixed (or broken?).

somiaj:
I moved the alt tag text into a `<desc>` tag for a long-text description of the image. Should I go with #11 mentioned above and also add this long description to the `aria-labeledby` of the image, or just leave it as a long-text description for browsers that support that?

drgrice1:
This works for me with Google Chrome and Gnome Orca.  I still haven't been able to get anything read for the SVG images in Firefox other than by adding `tabindex="0"` to the SVG.

I note that currently the SVG title is not a `title`.  Instead it is a `text` tag that the SVG is `aria-labelledby`.  Should that be moved to an actual `title` tag?  Basically that would put this at pattern #10, and if the `aria-labelledby` includes both the `title` and `desc` then it is pattern #11.

somiaj:
Changing the `text` tag to a `title` tag means the title won't be visible on the graph in my testing, so this would mean we would need both a `text` and a `title` tag, so probably leaving the `aria-labelledby` is better?

somiaj:
I added `tabindex="0"` to the svg tag.

drgrice1:
Yeah, it would mean that.  I think the current setup for the title should be fine.

drgrice1:
I am not sure on the `tabindex="0"` thing.  That may be an issue with Firefox or Gnome Orca.  It may be that we don't need to worry about it.  It should be tested without that on other systems where screen readers would be used more frequently.

somiaj:
Okay, reverted the `tabindex="0"` change.

somiaj:
Would remove the `aria-labeledby` and only including a description be worth considering, as the description contains the title information and more?

drgrice1 (approved):
I am going to go ahead and approve this.  It is at least a good start to making the images accessible.

Alex-Jordan (approved):
Well, I did not test with a JITAR set. But I think it's good!

Issue #1141, opened 2020-09-23
"Wide character in crypt" error message when passwords with certain "wide" characters are submitted in current/new password


taniwallach:
If certain Unicode characters (for example "א" but apparently most multi-byte UTF-8 characters) are entered in the password field of the login page and several other places, a "WeBWorK error" page is generated rather than a "nice response".  Several sample error messages are below.

See: https://perldoc.perl.org/5.32.0/functions/crypt.html (end of page).

This issue is somewhat complicated by the fact that some 8-bit characters do not trigger the error. For example, when the 3 character string     £¥±    is pasted into the password box no error occues. These are characters which in ISO-8859-1 used the high-bit and were copy+pasted from there to the string £¥± and into the password box. See: https://cs.stanford.edu/people/miles/iso8859.html . That is probably due to Perl's attempt to "downgrade (a copy of) the string back to an eight-bit byte string before calling crypt (on that copy)" when Unicode data is included. However, when that downgrade fails - the "Wide character in crypt" error is issued.

It would be nice to avoid these error pages, if necessary by just rejecting passwords entered in such places which contain wide characters which crypt cannot handle.

However, it would be nicer to handle wide characters inside the password in a manner which is invisible to the end user and which allows such characters as part of the passwords.

Since crypt does not support wide characters, we would need to "encode" the raw password string into a form which does not have any "wide characters" (for example to UTF-8) to avoid these types of problems. That sort of approach was suggested in the first patch proposed at https://bugzilla.mozilla.org/show_bug.cgi?id=453767 as a fix for similar issues with Bugzilla.

Most of WW seems to use `cryptPassword()` from  `lib/WeBWorK/Utils.pm` so the main change would be there. However, at least 2 places under `lib` call `crypt` directly to verify a user submitted password.
  - `checkPassword` in `lib/WeBWorK/Authen.pm`  (see also the sample `site_checkPassword`)
  - `body` in `lib/WeBWorK/ContentGenerator/Options.pm`   (section about updating the user's password)
Those locations also need to be modified.

I suspect that if the change is made carefully and using UTF-8 encoding of the new/provided password string, then any "plain" ASCII passwords (**at those only using characters without the high-bit set**) would continue to work as they did before. 

However, any change which enabled using more Unicode characters in the password might break the current behavior for passwords which have 8-bit characters which crypt currently managed to handle (as in the examples above), as such characters would be modified by the encoding into UTF-8 into a multi-byte character and thus the hashed password would be different.

[...]

PR #2628, opened 2024-11-21
Add support for wide characters in passwords.

taniwallach:
This is an attempt to address https://github.com/openwebwork/webwork2/issues/1141 about problems with wide characters in passwords.

Login attempts where the password field contained most wide Unicode characters would trigger an error page and not just a failed login attempt. So would pages where an attempt was made to set a password with such a character. The error pages was triggered as `crypt` would die on most wide Unicode characters. However, there were apparently some which would work as a result of special handling in crypt for wide characters which could be downgraded to 8-bit characters.

For example the character ג in passwords would trigger the error.

In order to support wide characters in passwords while not breaking any passwords which worked before this change, a password will be encoded into UTF-8 before being sent to crypt **only** when crypt dies on the original string.

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -633,8 +633,15 @@ sub checkPassword {
 	my $Password = $db->getPassword($userID);
 	if (defined $Password) {
 		# Check against the password in the database.
-		my $possibleCryptPassword = crypt $possibleClearPassword, $Password->password;
-		my $dbPassword            = $Password->password;
+		my $possibleCryptPassword = '';
+		# Wrap crypt in an eval to catch any "Wide character in crypt" errors.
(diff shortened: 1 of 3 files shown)


drgrice1 (commented):
This should use the `encode` method of `Mojo::Util` (which is really just the `Encode::encode` method), rather than the `Encode::utf8_encode` method.  The `Encode::utf8_encode` method should not be used anymore.

Also, the process that you have duplicated in three places should be made into a utility method in `lib/WeBWorK/Utils.pm`, and just called in all three places.  The `crypt` method can just be replaced with this utility method making much cleaner code.

I have put in a pull request to your branch making these changes.

drgrice1:
You could use "א".  That is such a character.  Change your password using that character (cut and paste).  Then check that you can log out and log back in again with that password.

pstaabp:
This has been replaced with #2659

Issue #2625, opened 2024-11-19
Accounts Manager breaks if a user has a permission level that is not defined in %userRoles


dlglin:
To reproduce: add a user to a course with a permission level that is not defined in the `%userRoles` hash for that course.

If you visit the Accounts Manager you get the error "maketext requires at least one parameter at /opt/webwork/webwork2/lib/Mojolicious/WeBWorK.pm line 110."

The way I came across this is that I had a course with a new role defined in course.conf, and I created a new course and copied instructors from that course.  Now I have a new course with a user with permission level 11, but the new course doesn't have a role associated with that permission level.

The other way this could occur is if you archive a course on a server with custom roles defined, and then unarchive it on a server that is missing those custom roles.

The issue appears to be at https://github.com/openwebwork/webwork2/blob/df4b848ebfb74973db226e5991ad831cc734852d/templates/ContentGenerator/Instructor/UserList/user_list_field.html.ep#L54 
If there is no entry in the hash associated with the permission level then you end up passing nothing to `maketext`.

somiaj:
Fixed in #2660.

PR #2629, opened 2024-11-21
Minimal support for custom permission levels in Accounts Manager which are not provided in the userRoles hash

taniwallach:
A minimal fix for https://github.com/openwebwork/webwork2/issues/2625.

This only prevents the error and allows the Accounts Manager to load. A fake permission level name "PermLevelN" will appear in the table. It is not a value which can be selected, and editing a user with such a value will force selecting a defined permission level.

### DIFF
--- a/templates/ContentGenerator/Instructor/UserList/user_list_field.html.ep
+++ b/templates/ContentGenerator/Instructor/UserList/user_list_field.html.ep
@@ -51,7 +51,7 @@
 			id => $fieldName . '_id', class => 'form-select form-select-sm w-auto flex-grow-0',
 			'aria-labelledby' => 'permission_header' =%>
 	% } else {
-		<%= maketext((grep { $ce->{userRoles}{$_} eq $value } keys %{ $ce->{userRoles} })[0]) %>
+		<%= maketext((grep { $ce->{userRoles}{$_} eq $value } keys %{ $ce->{userRoles} })[0] // "PermLevel$value") %>
 	% }
 % } elsif ($properties->{type} eq 'password') {
 	% # Note that this is only called if in editMode.


templates/ContentGenerator/Instructor/UserList/user_list_field.html.ep
```diff
@@ -51,7 +51,7 @@
 			id => $fieldName . '_id', class => 'form-select form-select-sm w-auto flex-grow-0',
 			'aria-labelledby' => 'permission_header' =%>
 	% } else {
-		<%= maketext((grep { $ce->{userRoles}{$_} eq $value } keys %{ $ce->{userRoles} })[0]) %>
+		<%= maketext((grep { $ce->{userRoles}{$_} eq $value } keys %{ $ce->{userRoles} })[0] // "PermLevel$value") %>
```
drgrice1, line 54: I think this should be
```suggestion
		% if (my $roleName = (grep { $ce->{userRoles}{$_} eq $value } keys %{ $ce->{userRoles} })[0]) {
			<%= maketext($roleName) %>
		% } else {
			<%= maketext("Permission Level [_1]", $value) %>
		% }
```
With this "Permission Level [_1]" will get translated.  With your code it will not.  I also don't like abbreviations shown in the user interface, so I would rather have "Permission Level" rather than "PermLevel".

Although, I was alternately thinking that maybe "Permission Level" in any form isn't needed.  This is the "Permission Level" column in the table.  So maybe instead just show the unknown permission level numerically?  So perhaps this should be
```suggestion
		% if (my $roleName = (grep { $ce->{userRoles}{$_} eq $value } keys %{ $ce->{userRoles} })[0]) {
			<%= maketext($roleName) %>
		% } else {
			<%= $value %>
		% }
```

Or maybe use `maketext('Unknown: [_1]', $value)`?

drgrice1:
I am closing this as it is contained in #2660.

PR #2630, opened 2024-11-22
Add hidden span labels for help links.

somiaj:
This adds a hidden span that gives a label for the help links for the main help link using the helpMacro and the help links on the course configuration page. There is now an aria-labelledby setting for this description which describes what the help link is for.

The page title is used for the description for the main help link. For the course configuration links, the setting short description is used for the title. A few of the short descriptions ended in a period, those were all removed to make the description "desc help." consistent and not have a period before the help statement.

I couldn't figure out a better label for the course configuration help statements than the short description. This seems a little long in some cases, but overall identifies what setting the help buttons is for.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -957,10 +957,11 @@ This method outputs a link that opens a modal dialog containing the results of r
 HelpFiles template.  The template file that is rendered is $name.html.  If that file does not
 exist, then nothing is output.
 
-The optional argument $args is a hash that may contain the keys label, label_size, or class.
-$args->{label} is the displayed label, $args->{label_size} is a font awesome size class and is
-only used if $args->{label} is not set, and $args->{class} is added to the html class attribute
-if defined.
+The optional argument $args is a hash that may contain the keys label, label_size, help_label,
+or class. $args->{label} is the displayed label. $args->{label_size} is a font awesome size class
+and is only used if $args->{label} is not set. $args->{help_label} is the hidden description of
+the help button, "help_label help.", which defaults to the page title, and is only used if
+$args->{label} is not set. $args->{class} is added to the html class attribute if defined.
 
 =cut
 
@@ -972,11 +973,16 @@ sub helpMacro ($c, $name, $args = {}) {
 		'i',
 		class         => 'icon fa-solid fa-circle-question ' . ($args->{label_size} // ''),
 		'aria-hidden' => 'true',
-		data          => { alt => ' ? ' },
 		''
-	);
+		)
+		. $c->tag(
+			'span',
+			class => 'visually-hidden',
+			$c->maketext('[_1] help.', $args->{help_label} // $c->page_title)
+		);
 	delete $args->{label};
 	delete $args->{label_size};
+	delete $args->{help_label};
 
 	return $c->include("HelpFiles/$name", name => $name, label => $label, args => $args);
 }
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -135,7 +135,7 @@ sub getConfigValues ($ce) {
 			},
 			{
 				var  => 'perProblemLangAndDirSettingMode',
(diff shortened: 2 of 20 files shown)


drgrice1 (commented):
When we talked about this in the meeting I think that you got mislead on a few things.

You mentioned that the visually hidden spans for the help icons on the course configuration page merely say "help", and that they should be more specific.  However, that isn't true.  Those should just say "help".  The thing is that when designing a page for a screen reader, it is important to control the verbosity and prevent text from being read multiple times.  If these buttons say, for instance, "Title for course displayed on the Assignments page help" and the text for the table also says "Title for course displayed on the Assignments page", then the screen reader reads both of those together making it rather verbose.  In the context since the text in the table will be read with together with the text for the icon, "help" is sufficient.

The same is true for the help links for the page by the page title.  Of course, those should not have the question mark for the content of the visually hidden span since that is not read by screen readers, but should have "help" for the content instead.

Also, the actual icons should are for sighted users only, and don't add anything for those using a screen reader.  So they should not have the `aria-labelledby` attribute, and instead should have the `aria-hidden` attribute.  This is so that the visually hidden span becomes what is read for link and the icon is completely ignored.  So the point is that the visually hidden spans as they were are structurally and semantically correct.

I actually also did some work on this after we discussed this in the meeting, and created a branch to fix the "?" help links, and removed the addition of the visually hidden spans by javascript.  I will go ahead and put that in as a pull request for comparison.

somiaj:
Ahh, I was considering the links in isolation too much, and thought having a lot of "Help link" would not be clear what the help was for, so added more to make that clear without considering where the link was being placed. Thanks.

Closing in favor of #2631.

drgrice1:
Actually, I am going to reopen this.  I would like @Alex-Jordan's opinion on this (or anyone else who has experience in this area).  I am certain that the `aria-labelledby` is not correct, but the more verbose text I am not certain of.  I tested with both Gnome Orca and Android Talkback.  With Gnome Orca the headers are read with the link, and it does seem rather redundant in that case.  However, with Android Talkback the link is read separately, although it is still the next thing that is read after the header.

drgrice1:
It seems that it is difficult to get screen reader support right.  Largely due to the fact that different screen readers seem to do rather different things.  It seems that screen reader support is not entirely standardized.

somiaj:
I removed the `aria-labelledby` tags and ids. If we do end up needing to with the more verbose text, I'll also update this to remove more of the alt text being added by javascript like #2631.

I am unsure what is best here, I do know with normal links the standard seems to be descriptive (don't use things like 'click me') and that links that do different things should have different text (which is partly why I thought all the links saying "help" might not be best). Anyways waiting to see if others know the best accessible standards here.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -972,11 +973,16 @@ sub helpMacro ($c, $name, $args = {}) {
...
 		class         => 'icon fa-solid fa-circle-question ' . ($args->{label_size} // ''),
 		'aria-hidden' => 'true',
-		data          => { alt => ' ? ' },
 		''
-	);
+		)
+		. $c->tag(
+			'span',
+			class => 'visually-hidden',
+			$c->maketext('[_1] help.', $args->{help_label} // $c->page_title)
```
drgrice1, line 981: I wanted to point out why I like my approach better.  The reason is mainly the change here.  My approach uses the help page layout title which already is translated and includes "help", and doesn't add a new optional argument to the `helpMacro` method.  So there is no need to go through all of the `helpMacro` calls and add this `help_label` argument.  For example, the set detail page help title is "Set Detail Help", and that title is used for the help link to that help.  I also moved the html generated here via the `tag` method into the `help_macro.html.ep` template which is cleaner because you can just use the `i` and `span` html tags directly.

somiaj: I didn't consider using those titles, and agree since they are already there they should be used. I think this also makes sense since in another view point, these are links to those headers, which should line up. Now that I understand the difference, I to like your approach better.

somiaj:
Closing in favor of #2631

PR #2631, opened 2024-11-22
Eliminate the visually hidden spans injected by javascript.

drgrice1:
The spans are simply added where needed.  No javascript is needed for them to be in the page.

~~The "?" help label is changed to just "Help" in most cases.  The "?" is usually not read by screen readers.  Note that these all go through the `helpMacro` method and a help file template is rendered.  The template can also set the ariaLabel stash value to override the default "Help" text.~~

~~Most pages shouldn't use the override though (I think).  For example, on the "Accounts Manager" page the screen reader says something like "accounts manager heading level 1 link help link" (at least that is what Gnome Orca says for this). I don't think it should be read as "account manager heading level 1 link accounts manager help link".  It increases the verbosity unnecessarily.~~

~~However, on the "Assignments" page the title of the page is the course title.  So the default aria label is overridden with "Assignments Help", and the screen reader says something like "course title heading level 1 link assignments help link".  Note the help title in the dialog was also changed to "Assigments Help".  It still said "Course Home Help". It seems we missed this when the page name was changed.~~

~~Of course mileage will vary as to what different screen readers say on different systems and with different browsers.~~

~~I worked on this after the discussion in the meeting this week, not realizing that @somiaj would also work on this.  See my comments on #2630.~~

Edit: This now sets the text in the hidden spans for the page help links to be the help dialog title.  So on the "Assignments" page the title of the help dialog is "Assignments Help" and that is now also the text for the hidden span that labels the link.  (Again, that was changed from "Course Home Help" in this pull request.)  On the "Account Settings" page the help dialog title and text for the hidden span is "Account Settings Help".  Of course the same goes for all pages.  Note that in some cases (notably the two I just specifically mentioned) the help dialog title is not the same as the page title.  On the "Assignments" page the page title is the course title, and on the "Account Settings" page the page title is "Account settings for userID".

This also now sets the text in the hidden spans for the course configuration settings/variables to be the return value of a new `ConfigObject` method `help_link_aria_label`.  For most configuration variables that is `maketext('Variable documentation for [_1]', $self->help_name)`.  For the database settings (`courseTitle` and `lms_context_id`) that is `maketext('Setting documentation for [_1]', $self->{var})`.

### DIFF
--- a/templates/ContentGenerator/Base/login_status.html.ep
+++ b/templates/ContentGenerator/Base/login_status.html.ep
@@ -8,7 +8,7 @@
 	%
 	<%= maketext('Logged in as [_1].', $userName) %>
 	<%= link_to $c->systemLink(url_for 'logout'), class => 'btn btn-light btn-sm ms-2', begin %>
-	<%= maketext('Log Out') %> <i class="icon fas fa-sign-out-alt" aria-hidden="true" data-alt="signout"></i>
+	<%= maketext('Log Out') %> <i class="icon fas fa-sign-out-alt" aria-hidden="true"></i>
 	<% end %>
 	%
 	% if ($effectiveUserID ne $userID) {
@@ -26,7 +26,7 @@
 		<%= maketext('Acting as [_1].', $effectiveUserName) %>
 		<%= link_to $c->systemLink(url_for, params => { effectiveUser => $userID }),
 			class => 'btn btn-light btn-sm ms-2', begin %>
-		<%= maketext('Stop Acting') %> <i class="icon fas fa-sign-out-alt" aria-hidden="true" data-alt="signout"></i>
+		<%= maketext('Stop Acting') %> <i class="icon fas fa-sign-out-alt" aria-hidden="true"></i>
 		<% end %>
 	% }
 % } else {
--- a/htdocs/js/System/system.js
+++ b/htdocs/js/System/system.js
@@ -117,18 +117,4 @@
 			messages.forEach((message) => bootstrap.Alert.getOrCreateInstance(message)?.close())
 		);
 	}
-
-	// Accessibility
-	// Present the contents of the data-alt attribute as alternative content for screen reader users.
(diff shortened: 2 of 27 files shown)


Alex-Jordan (approved):

templates/ContentGenerator/Base/login_status.html.ep
```diff
@@ -26,7 +26,7 @@
 		<%= maketext('Acting as [_1].', $effectiveUserName) %>
 		<%= link_to $c->systemLink(url_for, params => { effectiveUser => $userID }),
 			class => 'btn btn-light btn-sm ms-2', begin %>
-		<%= maketext('Stop Acting') %> <i class="icon fas fa-sign-out-alt" aria-hidden="true" data-alt="signout"></i>
+		<%= maketext('Stop Acting') %> <i class="icon fas fa-sign-out-alt" aria-hidden="true"></i>
```
drgrice1, line 29: Note that the changes in this file completely remove the visually hidden spans, and instead make the icons `aria-hidden`.  These icons are really purely decorative symbols that don't actually add anything for screen reader users.  The button text says what is needed.  It does not make sense to have the button text say "Log Out" and have the icon say "signout", so that the screen reader will say "Log Out signout".  In fact in one case (depending on how you arrive at this point in the page) Gnome Orca says "Log Out visited link signout visited link", making it seem like there are two links where there is only one.

drgrice1: Also, the "signout" text was incorrect for the "Stop Acting" button.

drgrice1:
Note that several of the `data-alt` attributes that were previously converted to a visually hidden span by JavaScript and now are replaced with direct visually hidden spans, previosly said "Help Icon" and now say "Help".  I will probably try to go through those and make them more specific also.  If anyone has suggestions for texts for those, please comment in the file here.

somiaj (approved):

PR #2632, opened 2024-11-22
Switch to using Perl::Tidy version 20240903 (the latest).

drgrice1:
Note that the `-vxl='q'` option has been removed.  This option disables vertical alignment of qw quotes.  Alignment of qw quotes was added in version 20220613 (the previous version), and this option to disable that was added to the webwork2 (and pg) perltidy configuration to avoid the alignment because it changes so many files. However, I don't think that alignment is bad, and I am tired of adding new options to prevent code changes.  With this newer version another option would be needed to prevent the new signed number alignment (although that affects PG much more than webwork2).

The `bin/dev_scripts/run-perltidy.pl` script now insists on this specific version of perltidy instead of this version or newer.  This is because every time a new version of perltidy comes out, it changes things.  So developers really need to be using the same version or it will not be consistent with the github workflow result.

Also, the version requirement in the check_modules.pl script has been removed.  Only developers need a specific version of perltidy.  Those running webwork2 can use other versions.  This is only used for formatting problems in the PG problem editor, and using other versions may result in slightly different formatting there, but that is not so consequential that we need to require a specific version. Note that this change depends on the corresponding pg pull request (https://github.com/openwebwork/pg/pull/1145), because in that pull request the `-vxl='q'` option has been removed.  Otherwise version 20220613 or newer of perltidy would still be needed.

The first commit makes the changes for the switch to the new version, and the second commit is merely the result of running `perltidy` with the new version and rules.

### DIFF
--- a/.github/workflows/check-formats.yml
+++ b/.github/workflows/check-formats.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Install dependencies
-        run: cpanm -n Perl::Tidy@20220613
+        run: cpanm -n Perl::Tidy@20240903
       - name: Run perltidy
         shell: bash
         run: |
--- a/.perltidyrc
+++ b/.perltidyrc
@@ -20,4 +20,3 @@
 -nlop   # No logical padding (this causes mixed tabs and spaces)
 -wn     # Weld nested containers
 -xci    # Extended continuation indentation
--vxl='q' # No vertical alignment of qw quotes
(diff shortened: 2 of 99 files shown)


somiaj (approved):
I don't have a strong opinion on if we should line up `qw` quotes or not, but we should probably go with what upstream suggests instead of including overrides to their default behavior.

pstaabp (approved):
I'm good with this.   Everyone will need to upgrade the perltidy version though.

PR #2633, opened 2024-11-22
Switch to using the latest Perl::Tidy (version 20240903) alternate approach.

drgrice1:
A new `--valign-signed-numbers` option has been added in the version of Perl::Tidy and is the default.  That option has been disabled to reduce the number of code changes.  Although it only changes one line of the webwork2 code. It changes a lot more in the pg code.

The `bin/dev_scripts/run-perltidy.pl` script now insists on this specific version of perltidy instead of this version or newer.  This is because every time a new version of perltidy comes out, it changes things.  So developers really need to be using the same version or it will not be consistent with the github workflow result.

Also, the version requirement in the check_modules.pl script has been removed.  Only developers need a specific version of perltidy.  Those running webwork2 can use other versions.  This is only used for formatting problems in the PG problem editor, and using other versions may result in slightly different formatting there, but that is not so consequential that we need to require a specific version. Note that this change depends on the corresponding pg pull request (https://github.com/openwebwork/pg/pull/1146), because in that pull request the `-vxl='q'` option has been removed.  Otherwise version 20220613 or newer of perltidy would still be needed.

The first commit makes the changes for the switch to the new version, and the second commit is merely the result of running `perltidy` with the new version and rules.

### DIFF
--- a/.github/workflows/check-formats.yml
+++ b/.github/workflows/check-formats.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Install dependencies
-        run: cpanm -n Perl::Tidy@20220613
+        run: cpanm -n Perl::Tidy@20240903
       - name: Run perltidy
         shell: bash
         run: |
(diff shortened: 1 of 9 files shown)


pstaabp (approved):

somiaj:
Closing as we went with #2632.

PR #2636, opened 2024-11-23
Fix PGML Emphasis in the PG CodeMirror editor.

drgrice1:
Really, the fix has been published in an updated pg-codemirror-editor package (actually the bug was in the codemirror-lang-pg package).  This just updates the version in the package.json file.

This was observed by @somiaj in #2595 (see https://github.com/openwebwork/webwork2/pull/2595#issuecomment-2495203173).  It turns out the problem was an error in converting from the Perl `substr` method to the JavaScript `substring` method.  The code actually uses the JavaScript `slice` method now.

### DIFF
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -8,7 +8,7 @@
             "license": "GPL-2.0+",
             "dependencies": {
                 "@fortawesome/fontawesome-free": "^6.5.2",
-                "@openwebwork/pg-codemirror-editor": "^0.0.1-beta.20",
+                "@openwebwork/pg-codemirror-editor": "^0.0.1-beta.25",
                 "bootstrap": "~5.3.3",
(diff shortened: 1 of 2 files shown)


somiaj:
This fixes the issue, thanks. I also notice that if using double `**` or `__` for strong/emphasis, that there is a bit of issue with the matching/formatting.

image

Also seems PGML doesn't require balance here, and will match any number of `*` or `_` on both ends, so this works but the syntax highlighting should probably also match all characters not just the first.

image

drgrice1:
I can fix that.  Note that PGML allows it, but it isn't documented that it should work.  Why would you do that anyway?

drgrice1:
I have fixed the emphasis delimiter length (for those who unintelligibly want to type extra characters).

somiaj (approved):

PR #2638, opened 2024-11-25
Remove a local link in the `package-lock.json` file.

drgrice1:
This is a remnant from testing the CodeMirror 6 stuff locally, that seems to have stayed there even after uninstalling the local link. This was never supposed to have been committed.


somiaj (approved):
Merging, this is an obvious fix.

PR #2639, opened 2024-11-25
Recover syntax highlighting for sample problems.

drgrice1:
This was an oversight when I made the switch to the PG CodeMirror editor with CodeMirror 6.  The sample problems also use CodeMirror for syntax highlighting.  Since the CodeMirror 5 javascript is no longer available that fails.

The pg-codemirror-editor package now exports a `runMode` method (available via the global `PGCodeMirrorEditor` object), that essentially does what the CodeMirror 5 `runMode` addon method did.  It is slightly different in that it only works for the PG language. At this point we don't use it for anything else, so that should be good for now.

### DIFF
--- a/htdocs/js/SampleProblemViewer/sample-problem.js
+++ b/htdocs/js/SampleProblemViewer/sample-problem.js
@@ -1,5 +1,5 @@
-for (const pre of document.body.querySelectorAll('pre.CodeMirror')) {
-	CodeMirror.runMode(pre.textContent, 'PG', pre);
+for (const pre of document.body.querySelectorAll('pre.PGCodeMirror')) {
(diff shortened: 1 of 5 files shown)


Alex-Jordan (approved):

drgrice1:
Note that the version of the `pg-codemirror-editor` that this updates to also includes some fixes for syntax highlighting in interpolation (the fixes are really in the `codemirror-lang-pg`, `codemirror-lang-perl`, and `codemirror-lang-mt` packages).

(I accidentally put this comment in https://github.com/openwebwork/pg/pull/1149, but it was meant for this pull request.)

pstaabp (approved):
👍

PR #2641, opened 2024-11-26
Updates to grading tests for other users.

somiaj:
When able to grade a test of another user, change the "Grade Test" button to "Grade Test for UserID" to make it clear the test is being graded for another user.

Show the "Check Test" button along side the "Grade Test" button for instructor who can use the problem grader but also have the permission to submit the past due test.

Allow instructors who can use the problem grader to grade unsubmitted tests for that user without needing the permission to submit test versions or answers for another user. This way if for some reason a student's test doesn't get submitted correctly (such as a user closing their browser without grading the test or after reaching a grade proctor screen), it can still be graded with their saved answers.

Update the timer when acting as another user if it is being shown, just only suppress warnings about running out of time in this case.

Update description of record_answers_when_acting_as_student to state that it can also start and grade test versions. Also include a warning about that permission should be disabled (set back to "nobody") after using it as it can interfere with tests, and should probably only be used temporarily.

Note this is just hopefully a good start to improving being able to grade tests that didn't get submitted and grading tests for other users, there are probably tweaks that need to be made.

I also initially wanted to only allow submitting an unsubmitted testing using a students saved answers, but due to how the GatewayQuiz module works, found this to be difficult to implement (or at least I didn't see an easy way to do that), so for now an instructor could change answers before submitting the unsubmitted test even if they don't have permission to record student answers. I suspect most instructors will just use the students saved answers in this case anyways.

### DIFF
--- a/lib/WeBWorK/Authen/Proctor.pm
+++ b/lib/WeBWorK/Authen/Proctor.pm
@@ -96,10 +96,17 @@ sub verify_normal_user {
 	# is 'No', then the verify method will have returned 1, and this never happens.  For an ongoing login session, only
 	# a key with versioned set information is accepted, and that version must match the requested set version.  The set
 	# id will not have a version when opening a new version. For that new proctor credentials are required.
-	if ($self->{login_type} eq 'proctor_login'
-		&& $c->stash('setID') =~ /,v\d+$/
+	if (
+		$self->{login_type} eq 'proctor_login'
 		&& $c->authen->session('proctor_authorization_granted')
-		&& $c->authen->session('proctor_authorization_granted') eq $c->stash('setID'))
+		&& (
+			(
+				$c->stash('setID') =~ /,v\d+$/
+				&& $c->authen->session('proctor_authorization_granted') eq $c->stash('setID')
+			)
+			|| $c->authen->session('acting_proctor')
+		)
+		)
 	{
 		return 1;
 	} else {
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -529,7 +529,14 @@ sub getConfigValues ($ce) {
 				var  => 'permissionLevels{record_answers_when_acting_as_student}',
 				doc  => x('Can submit answers for a student'),
 				doc2 => x(
-					'When acting as a student, this permission level and higher can submit answers for that student.'),
+					'When acting as a student, this permission level and higher can submit answers for that student, '
+						. 'which includes starting and grading test versions.  This permission should only be turned '
+						. 'on temporarily and set back to "nobody" after you are done submitting answers for a '
+						. 'student, as it can interfere with tests.  If you have this permission and are viewing a '
+						. 'test version for a student that is also working on that version, your answers will be '
+						. 'saved for that student when moving between pages, which could reset or change the answers '
+						. 'entered in by the student.'
+				),
 				type => 'permission'
 			},
 			{
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -532,10 +532,10 @@ sub getConfigValues ($ce) {
 					'When acting as a student, this permission level and higher can submit answers for that student, '
 						. 'which includes starting and grading test versions.  This permission should only be turned '
 						. 'on temporarily and set back to "nobody" after you are done submitting answers for a '
-						. 'student, as it can interfere with tests.  If you have this permission and are viewing a '
-						. 'test version for a student that is also working on that version, your answers will be '
-						. 'saved for that student when moving between pages, which could reset or change the answers '
-						. 'entered in by the student.'
+						. 'student.  Leaving this permission on is dangerous, as you could unintentionally submit '
+						. 'answers for a student, which can use up their total number of attempts.  Further, if you '
+						. 'are viewing an open test version, your answers on each page will be saved when you move '
+						. q/between pages, which will overwrite the student's saved answers./
 				),
 				type => 'permission'
 			},
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -233,7 +233,7 @@
 % # Remaining output of test headers.
 % # Display timer or information about elapsed time, output link, and information about any recorded score if not
 % # submitAnswers or checkAnswers.
-% if ($c->{can}{recordAnswersNextTime} || $c->{can}{gradeUnsubmittedTest}) {
+% if ($c->{can}{recordAnswersNextTime} && before($c->{set}->due_date + $ce->{gatewayGracePeriod}, $submitTime)) {
 	% my $timeLeft = $c->{set}->due_date - int($submitTime);    # This is in seconds
 	%
 	% # Print the timer if there is less than 24 hours left.
@@ -657,7 +657,7 @@
 					% && (!$c->{can}{recordAnswersNextTime} || $c->{can}{showProblemGrader})) {
 				<%= submit_button maketext('Check Test'), name => 'checkAnswers', class => 'btn btn-primary mb-1' =%>
 			% }
-			% if ($c->{can}{recordAnswersNextTime} || $c->{can}{gradeUnsubmittedTest}) {
+			% if ($c->{can}{recordAnswersNextTime}) {
 				<%= tag('input',
 					type  => 'submit',
 					name  => 'submitAnswers',
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -68,30 +68,33 @@
 % # If the set or problem is invalid, then show that information and exit.
 % if ($c->{invalidSet}) {
 	<div class="alert alert-danger mb-2">
-		<div class="mb-2">
-			% if ($c->{invalidVersionCreation}) {
-				<%= maketext(
-					'The selected test ([_1]) is not a valid test for [_2] (acted as by [_3]).',
-					$setID, $effectiveUserID, $userID
-				) =%>
-			% } else {
-				<%= maketext(
-					'The selected test ([_1]) is not a valid test for [_2].',
-					$setID, $effectiveUserID
-				) =%>
-			% }
-		</div>
+		% if (!$c->{confirmSubmitForStudent} || $c->{invalidVersionCreation}) {
+			<div class="mb-2">
+				% if ($c->{invalidVersionCreation}) {
+					<%= maketext(
+						'The selected test ([_1]) is not a valid test for [_2] (acted as by [_3]).',
+						$setID, $effectiveUserID, $userID
+					) =%>
+				% } else {
+					<%= maketext(
+						'The selected test ([_1]) is not a valid test for [_2].',
+						$setID, $effectiveUserID
+					) =%>
+				% }
+			</div>
+		% }
 		<div><%= $c->{invalidSet} %></div>
-		% if ($c->{invalidVersionCreation} && $c->{invalidVersionCreation} == 1) {
+		% if ($c->{confirmSubmitForStudent}) {
 			<div class="mt-3">
-				<%= link_to maketext('Create New Test Version') => $c->systemLink(
+				<%= link_to $c->{invalidVersionCreation}
+					? maketext('Create New Test Version') : maketext('View Test Version') => $c->systemLink(
 						url_for,
-						params => { effectiveUser => $effectiveUserID, user => $userID, createnew_ok => 1 }
+						params => { effectiveUser => $effectiveUserID, user => $userID, submit_for_student_ok => 1 }
 					),
 					class => 'btn btn-primary'
 				=%>
 				<%= link_to maketext('Cancel') => $c->systemLink(
-						url_for('problem_list', setID => $setID),
+						url_for('problem_list', setID => $setID =~ s/,v\d+$//r),
 						params => { effectiveUser => $effectiveUserID, user => $userID }
 					),
 					class => 'btn btn-primary'
@@ -409,6 +412,10 @@
 		% if ($numProbPerPage && $numPages > 1) {
 			<%= hidden_field newPage => '' =%>
 			<%= hidden_field currentPage => $pageNumber =%>
+		% }
+		% # Keep track that a user has confirmed it is okay to submit for a student.
+		% if (param('submit_for_student_ok')) {
+			<%= hidden_field submit_for_student_ok => 1 =%>
 		% }
 		%
 		% # Set up links between problems and, for multi-page tests, pages.
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -65,42 +65,46 @@
 % my $userID          = param('user');
 % my $effectiveUserID = param('effectiveUser');
 %
-% # If the set or problem is invalid, then show that information and exit.
+% # If the set is invalid, then show that information and exit.
 % if ($c->{invalidSet}) {
 	<div class="alert alert-danger mb-2">
-		% if (!$c->{confirmSubmitForStudent} || $c->{invalidVersionCreation}) {
-			<div class="mb-2">
-				% if ($c->{invalidVersionCreation}) {
-					<%= maketext(
-						'The selected test ([_1]) is not a valid test for [_2] (acted as by [_3]).',
-						$setID, $effectiveUserID, $userID
-					) =%>
-				% } else {
-					<%= maketext(
-						'The selected test ([_1]) is not a valid test for [_2].',
-						$setID, $effectiveUserID
-					) =%>
-				% }
-			</div>
-		% }
+		<div class="mb-2">
+			<%= maketext('The selected test ([_1]) is not a valid test for [_2].', $setID, $effectiveUserID) =%>
+		</div>
 		<div><%= $c->{invalidSet} %></div>
-		% if ($c->{confirmSubmitForStudent}) {
-			<div class="mt-3">
-				<%= link_to $c->{invalidVersionCreation}
-					? maketext('Create New Test Version') : maketext('View Test Version') => $c->systemLink(
-						url_for,
-						params => { effectiveUser => $effectiveUserID, user => $userID, submit_for_student_ok => 1 }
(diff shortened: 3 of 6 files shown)


templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -235,7 +235,7 @@
 % # Remaining output of test headers.
 % # Display timer or information about elapsed time, output link, and information about any recorded score if not
 % # submitAnswers or checkAnswers.
-% if ($c->{can}{recordAnswersNextTime}) {
+% if ($c->{can}{recordAnswersNextTime} || $c->{can}{gradeUnsubmittedTest}) {
```
drgrice1, line 238: I don't think that there is any benefit to showing the time for a user that can grade an unsubmitted test, but does not have permissions to record answers considering that you have set it so that an instructor can only do this when the test is overtime.  In testing this I see that while there is still time remaining the timer is not shown, and after the test is completed I get the message "This test version is past due, but has not been graded. You can still grade the test for this user.", but the timer just shows "Remaining Time: 00:00:00".  That isn't very useful.

lib/WeBWorK/ConfigValues.pm
```diff
@@ -529,7 +529,14 @@ sub getConfigValues ($ce) {
...
 				doc2 => x(
-					'When acting as a student, this permission level and higher can submit answers for that student.'),
+					'When acting as a student, this permission level and higher can submit answers for that student, '
+						. 'which includes starting and grading test versions.  This permission should only be turned '
+						. 'on temporarily and set back to "nobody" after you are done submitting answers for a '
+						. 'student, as it can interfere with tests.  If you have this permission and are viewing a '
+						. 'test version for a student that is also working on that version, your answers will be '
+						. 'saved for that student when moving between pages, which could reset or change the answers '
+						. 'entered in by the student.'
+				),
```
drgrice1, line 539: I am not sure that the wording "as it can interfere with tests" is very good.  I think I would say something indicating that it is dangerous because it is very easy to unintentionally submit answers for students.

The wording after that seems a bit odd as well.  The remark about answers being saved for that student when moving between pages and resetting answers seems confusing.  It doesn't really make sense to me at least.

templates/ContentGenerator/GatewayQuiz.html.ep
somiaj: I agree, thought the timer was a bit annoying even if I had the permission set to being able to submit grades for students and the due date is passed, I would prefer it was not there, but left that behavior as it was in that case and just added this case to that.

Seems just removing `|| $c->{can}{gradeUnsubmittedTest}` isn't enough, I must have added this before I added `can_gradeUnsubmittedTest` to the `can_recordAnswers` check (otherwise the answers weren't recorded), so I updated this to only show the timer before the due date.

lib/WeBWorK/ConfigValues.pm
somiaj: I had a lot of trouble trying to describe the behavior that the instructors answers are saved when moving between pages while working on an open test version. I have made an attempt to clean it up a bit, let me know of any suggested language you think would clear it up further.

drgrice1: The wording is much better now.

Although, I did not realize that the last sentence in your added documentation for this permission level was true.  I tested it and see that it is true.  If a student is working a test and at the same time an instructor with this permission acts as the student changes an answer but does not click "Grade Test" (or "Grade Test for ..."), then the answers do not immediately appear in the student's test (because the hidden form fields in the student's browser are not updated), but if the student leaves the test and then re-enters it, the instructors answers appear. This seems a bit concerning to me, but I am not sure what can be done about it.

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -233,7 +233,7 @@
 % # Remaining output of test headers.
 % # Display timer or information about elapsed time, output link, and information about any recorded score if not
 % # submitAnswers or checkAnswers.
-% if ($c->{can}{recordAnswersNextTime}) {
+% if ($c->{can}{recordAnswersNextTime} && before($c->{set}->due_date, $submitTime)) {
```
drgrice1, line 236: This is a problem.  The `gwTimer` div still needs to be displayed after the due date.  The javascript still acts based on this existing and uses the data from it after the due date.  I think that this would be okay if it included the `gatewayGradePeriod`.

lib/WeBWorK/ConfigValues.pm
somiaj: Yea, I realized this could happen when testing this, and I see why the answers have to be saved, and I don't see any way around this without changing how answers are saved. I did consider we could add another warning page (similar to when an instructor tries to start a test version) informing them of this if an instructor with permission to submit answers tries to view an open test version.

The possibility I was considering doesn't even require the instructor to change the student answers. If the instructor opens the same page the student is currently on, the instructor gets the answers for that student at that time. If the student changes their answers and moves pages, then the instructor moves pages, the old answers are then saved over the student's changes. Overall viewing an open test version with the permission to submit answers is very dangerous.

drgrice1: If the instructor and student are both navigating around within the quiz at the same time the last answer column in the database for all problems will always contain the answers from whatever is in the hidden form fields in the browser of whoever last changed pages or clicked preview answers.

I think that a warning page for this is justified.

templates/ContentGenerator/GatewayQuiz.html.ep
somiaj: Updated to include the `gatewayGracePeriod`.

lib/WeBWorK/ConfigValues.pm
somiaj: I added a warning page if a user is acting as another user tries to view an open test version with the permission to submit answers for another user.

somiaj:
@drgrice1 I decided to test this with a proctored test and it seems that it "mostly" works in the sense that you end up having to enter in proctor information twice to start or view the test, but otherwise things work as expected. I'm unsure of the best way to make it so you don't need to authenticate twice here, so I am asking for what you think would be best.

What I am seeing, is when starting or viewing a proctored test as a student with permissions to submit for them, the first page you are sent to is the proctor login, then after that is done you get the warning about starting or viewing open test versions, then after agreeing to the risks you get sent back to the proctor page, and finally after the second proctor authentication you get access to the test and can move around as expected. I don't think this behavior is new to this PR, but now it affects viewing test versions as well as starting them.

drgrice1:
Yeah, that certainly isn't good.  In no case should the proctor authentication be needed twice.  Although the behavior is not new with this pull request, it is new with #2610.  This does not happen with WeBWorK 2.18.  So this is something that was missed with #2610, and needs to be fixed.

somiaj:
Though maybe the issue is the fact that we are asking for a proctor authentication from a user who can authenticate themselves. Maybe something we can discuss at a meeting, should we really be asking an instructor (or other user) who has permissions to start a proctored test for a proctor login. They have already authenticated, so we don't need to ask them to authenticate again.

somiaj:
I don't think it is completely new with that PR, as I didn't mess with proctor behavior directly. Users with the permission `create_new_set_version_when_acting_as_student` probably had a similar issue since they had to agree to the warning, but now that we are warning users with the `record_answers_when_acting_as_student` permission it is affecting both.

My thoughts is the warning should appear before the proctor authentication page, but unsure how to best do it. The only other way I can think of to deal with it is turn the links on the warning page to a form submit, so we can keep all the hidden fields with the proctor authentication.

somiaj:
I can't reproduce your last comment, for both proctored and unproctored tests I get a warning when I click "Continue Open Test". Also the fact you don't see the "Grade Test for ....", are you still testing this branch?

drgrice1:
Unfortunately, the proctor authentication is going to have to come first.  Reworking things so that the warning page can come first would be messy, and could open potential security vulnerabilities.  Making the continue link on the warning page a submit button with hidden inputs may work, but better would be to use the session.  Hidden inputs and a form submission may open a loop hole for entering a proctored exam.

drgrice1:
I am on this branch, and not getting the warning page at all.  That is the warning for continuing an in progress test.

drgrice1:
I also never get the "Grade Test for ..." button in any situation.

somiaj:
I still can't reproduce this. Did you disable the permission to submit answers for students? You won't get the warning or the grade test button in that case.

drgrice1:
Wait, I was on the  other branch for #2610 accidentally.

somiaj:
In testing this I did run across another issue, though hopefully not something users are actually doing. If you have a proctored test, a student starts a proctored test version with authentication, you then change the template set from proctored to normal test, it is no longer possible to access that proctored test version with the error 'Requested set "Quiz07" is a proctored test, but no valid proctor authorization has been obtained.'. Not something I think is introduced here, and hopefully not something instructors are actually doing. But maybe something we should look into?

drgrice1:
That case is not something to concern ourselves with at this point.  That is something to defer for the future, and probably something that has existed for a long time.

drgrice1:
Another thing that I see.  The message that is shown when a user with permission to create a new set for another user attempts to do so is

The selected test (Testing_Proctor_Gateway) is not a valid test for tguy (acted as by grice1).
You are acting as user tguy. If you continue, you will create a new version of this test for that user, which will count
against their allowed maximum number of versions for the current time interval. In general, this is not what you want to
do. Please be sure that you want to do this before clicking the "Create New Test Version" button below. Alternatively,
click "Cancel".

That should not include the message that the set is not a valid test.  I is a valid test, it is just that the version doesn't exist.  That message needs to be changed to say that the test version does not exist instead.

somiaj:
I was looking into that too, and found the following change to not delete the key if waiting for a confirmation works.

```diff
diff --git a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
index 6a007a388..427fba093 100644
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -695,8 +695,10 @@ async sub pre_header_initialize ($c) {
        }
 
        # If the set or problem is invalid, then delete any proctor session keys and return.
-       if ($c->{invalidSet}) {
-               if (defined $c->{assignment_type} && $c->{assignment_type} eq 'proctored_gateway') {
+       # Don't delete key if asking for submit for student confirmation.
+       if (!$c->{confirmSubmitForStudent} && $c->{invalidSet}) {
+               if (defined $c->{assignment_type} && $c->{assignment_type} eq 'proctored_gateway')
+               {
                        delete $c->authen->session->{proctor_authorization_granted};
                }
                return;
```

I used `confirmSubmitForStudent` as the variable being set if confirmation is needed before either creating or viewing a test if the permission `record_answers_when_acting_as_student` (or one of the lesser permission are set). I just shorted the statement to be 'SubmitForStudent` vs `RecordAnswersWhenActingAsStudent`. I mostly tried to leave the `invalidVersionCreation` alone which was being set if trying to access a version that doesn't exist yet (i.e. when creating a new version) but it couldn't be created due to lack of permission or waiting for confirmation.

I agree, I thought that message about the set being invalid was also odd to appear when asking for confirmation, but left it alone as that was existing behavior. I can go clean this up to try to better describe what the cases are.

drgrice1:
The `invalidVersionCreation` is only ever set for a user acting as another user.  It is not set more generally for when an invalid version is attempted to be created.  So that flag needs to be reworked.

Instead drop both of those, and use two flags that make sense.  I suggest something like `actingConfirmation` and `actingConfirmationMessage`.  Then rework the code for this in the `GatewayQuiz.html.ep` file to use completely separate code for an `invalidSet` (it doesn't need much) and for these new flags.

drgrice1:
You can just use the `invalidSet` and go with your quick fix approach for now.  The method I am proposing will take more work.

somiaj:
@drgrice1 My simple fix (and even my more complicated fix I just pushed) only works for viewing open versions. When creating a new test version, even if I don't delete the `proctor_authorization_granted` key, it still requires two proctor authentication attempts. I have tried to track this down, but unsure why when trying to create a new version something in the proctor authorization fails, while viewing an open version works. I think it has something to do with having a version for the test to view, but I was not able to track down the difference.

drgrice1:
The problem is that proctor authorization only works if the session key `proctor_authorization_granted` is set to a versioned set (like `testId,v1`).  The only way this will work is to add another session key that is checked for that covers this case.  I am using `confirm_version_creation` in my testing.  So the `Proctor.pm` module needs to check for this key as well as the `proctor_authorization_granted` key, and not insist on a versioned set in the case that `confirm_version_creation` is set.

somiaj:
Ahh. Can you share your commit/diff? I was thinking that when `proctor_authorization_granted` was just set to `$setID` before appending the version would have covered this case, and was confused why it wasn't working, since all my testing showed the value didn't append the version before the version creation.

drgrice1:
What I have done is in the `grade-test-for-others-fixes` branch on my fork.  It doesn't have a valid commit message, since I wasn't finished with it, but that is what I have at this point.  I had a slightly different setup than you for some of the other things.

lib/WeBWorK/Authen/Proctor.pm
```diff
@@ -96,10 +96,16 @@ sub verify_normal_user {
...
-		&& $c->authen->session('proctor_authorization_granted')
-		&& $c->authen->session('proctor_authorization_granted') eq $c->stash('setID'))
+	if (
+		$self->{login_type} eq 'proctor_login'
+		&& (
+			(
+				$c->stash('setID') =~ /,v\d+$/
+				&& $c->authen->session('proctor_authorization_granted') // '' eq $c->stash('setID')
+			)
+			|| $c->authen->session('confirm_version_creation')
```
drgrice1, line 106: I changed this to `acting_proctor` in my branch.  This is used for both new version creation and entering an in progress test.  So `confirm_version_creation` isn't really correct.

somiaj:
Thanks, I adapted it to my previous changes, and everything seems to work in regard to not requiring two authentication attempts in my testing.

lib/WeBWorK/Authen/Proctor.pm
somiaj: Thanks. Updated.

lib/WeBWorK/Authen/Proctor.pm
```diff
@@ -96,10 +96,17 @@ sub verify_normal_user {
...
+		$self->{login_type} eq 'proctor_login'
+		&& (
+			(
+				$c->stash('setID') =~ /,v\d+$/
+				&& $c->authen->session('proctor_authorization_granted')
+				&& $c->authen->session('proctor_authorization_granted') eq $c->stash('setID')
+			)
+			|| $c->authen->session('acting_proctor')
+		)
+		)
```
drgrice1, line 109: This should be:

```suggestion
	if (
		$self->{login_type} eq 'proctor_login'
		&& $c->authen->session('proctor_authorization_granted')
		&& (
			(
				$c->stash('setID') =~ /,v\d+$/
				&& $c->authen->session('proctor_authorization_granted') eq $c->stash('setID')
			)
			|| $c->authen->session('acting_proctor')
		)
		)
```

The `proctor_authorization_granted` key still needs to be checked to ensure that it is set.  That is still what is set when authentication succeeds.

somiaj: Oh I misread in my  translation, thanks for catching.

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -65,39 +65,45 @@
...
+		<%= maketext(
+			'You are acting as user [_1] and do not have the permission to create a new test version when acting '
+				. 'as another user.',
+			$effectiveUserID
+		) =%>
+	</div>
+	% last;
+% }
+% # Get confirmation before creating new test version or working on an open test for another user.
+% if ($actingConfirmation) {
```
drgrice1, line 90: This needs to be

```suggestion
% if (stash->{actingConfirmation}) {
```
Either that or the stash value has to be defined at the beginning of the `pre_header_initialize` method.  This causes the exception `Global symbol "$actingConfirmation" requires explicit package name` if a gateway test is accessed when not acting.

drgrice1: The same for the other stash values below.

somiaj:
I had move the return when waiting for acting proctor confirmation to before deleting `proctor_authorization_granted` from the session, otherwise after I created a version I would view the first page, but moving to another page required authentication. I may have put this in the wrong place initially.

drgrice1:
Yeah, I think it is right now.

drgrice1 (approved):
I think this is looking good now.

Alex-Jordan:
Would it be possible to put in something like this?

Screenshot 2025-02-01 at 1 27 13 AM

That was a crude job just editing the DOM. If this is doable it could be done to look nicer. And attention would be needed on small screen too.

This is just a thought I had after reading the stern warning in the doc2. I don't trust faculty to read that, but if you put a warning in their face, they will see it. But if it's too much of a fiddle, I still approve the PR.

drgrice1:
I think it would be good to make the information that the user can submit answers for the user being acted as more prominent on the page, but I don't think it will look good with the "acting" button as in your fiddle.

Another possibility is to add this message to the "student nav" (that is the sticky nav bar that shows the dropdowns to change which user is being acted as and the buttons to change problems.  That is already shown anytime that you can actually submit answers anyway and would be easier to make work on all screen sizes.

somiaj:
> This is just a thought I had after reading the stern warning in the doc2. I don't trust faculty to read that, but if you put a warning in their face, they will see it. But if it's too much of a fiddle, I still approve the PR.

Instructors should now always encounter a warning screen if they try to access a test while acting as a student with this permission enabled telling them not to do that and making them click a I wish to proceeded button to actually access a students test when acting as a student with this permission. So I think this should classify as a warning in their face.

I can add an additional note to the UI, but currently the submit and grade test buttons how state 'Submit for username' and 'Grade Test for username' which is making it clear that you are acting as a student and have permission to submit for them as well.

somiaj:
If an additional warning is preferred, I would suggest putting it down next to the actual submit/grade test buttons in a `danger` alert to stand out.

drgrice1:
I agree that with the submit button now reading "Submit ... for ..." things are certainly much more clear than before.  I think this is good still.

PR #2656, opened 2025-01-15
Make JITAR disabled problems disabled without javascript.

drgrice1:
There is no reason to use javascript for this.  Just make the links disabled to begin with.  Furthermore, there is no reason to use jQuery for this.  I thought all webwork2 jQuery usage had already been removed, but this was missed.

To test this create a JITAR set with two or more problems (or really two or more nested sets of problems), and set "Restrict Problem Progression" to "Yes".  Assign the set to a student user, and then log in as that student user.  On the problem list page the latter prolems in the set should be grayed out, and clicking on them does nothing (that already works without javascript because these aren't even `a` tags, they are `span` tags).  If you open the first problem, then the latter problems in the prroblems list in the site menu should be grayed out, and clicking on them does nothing. Currently, that uses jQuery javascript to disable those links.  With this pull request, they are just disabled to begin with.

### DIFF
--- a/htdocs/js/Problem/problem.js
+++ b/htdocs/js/Problem/problem.js
@@ -4,9 +4,4 @@
 		const bsToast = new bootstrap.Toast(toast, { delay: 5000 });
 		bsToast.show();
 	});
-
-	// Prevent problems which are disabled from acting as links
-	$('.problem-list .disabled-problem')
-		.addClass('disabled')
-		.on('click', (e) => e.preventDefault());
 })();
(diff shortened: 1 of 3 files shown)


somiaj:
Tested this and looks good. Just some questions, why does the ProblemSet page when showing a list of problems in the table just use span's for the disabled links while the sidebar menu (Which is updated here) use disabled links? Any reason why these should be different? Also just a very minor thing, but in my testing there are no additional classes added via `<%= $linkClasses %>` so the html has an extra space, very very minor issue, just something I noticed, `<span class=" text-nowrap">`.

somiaj (approved):
Works as intended. My previous comment were just some things I noticed, but this is fine as is too.

drgrice1:
The `span` works in the problem page, but a `span` does not work in a `nav-item` as a `nav-link`.

The space in the class doesn't matter. It has always been that way.

pstaabp (approved):
This all seems to work.

PR #2658, opened 2025-01-22
Change the default value of `$CookieSecure` to 1.

drgrice1:
There is really no production scenario where this should be anything but 1.  If you are really serving without https, then you can change this. And good luck dealing with your users having to click through the warnings to get to your site.

Note that this just works if you are using `localhost` in development also.  Browsers have a built in exception for `localhost` and consider it secure without actual certificates.

I am tired of seeing forum issues because system administrators forget or don't know to change this.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -981,8 +981,7 @@ $session_management_via = "session_cookie";
 $CookieSameSite = "Lax";
 
 # Set the value of the secure cookie attribute.
-# The default is 0, as 1 will not work without https.
-$CookieSecure = 0;
+$CookieSecure = 1;
 
 # If $useSessionCookie is set to 1, then a "session" cookie will be used. This
 # means that the cookie will be deleted when the browser session ends.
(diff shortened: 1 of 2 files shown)


dlglin (approved):
As someone who was burned by this I wholeheartedly approve.
Is it worth adding a note to `localOverrides` telling people that changing this to 0 in a production environment is a bad idea?

drgrice1:
I think it is worth adding a note.  I almost did but for some reason held off.  I guess I wasn't sure exactly what to say other than "Don't do this."

drgrice1:
Historically there have been quite a few instances of webwork that were served without SSL, and we have lots of notes in the installation manual and such about "if you are serving with https...".  Nowadays, serving without SSL even for the most basic of sites is something you really just can't do anymore.

dlglin:
The only use case I see is for disabling this is test/dev environments.  That's why I suggested saying something about not changing it on production systems.
I agree that we should get rid of any instructions that describe https as optional.

somiaj (approved):

somiaj:
Oops, I missed you were complicating adding an additional note. I was mostly thinking you were discussing updating install instructions and adding notes there.

drgrice1:
That is okay. We can add a note later. I think that most will just leave this as it is.

PR #2659, opened 2025-01-23
Add support for wide characters in passwords. (replacement of #2628)

drgrice1:
This replaces #2628.  @taniwallach seems to not have time to make the requested changes, so this takes care of those for him.

The crypt function cannot handle most wide (Unicode) characters, but does handle some by attempting to downgrade to an 8-bit string.
    
In order to support wide characters in passwords while not breaking any passwords which worked before this change, a password will be encoded into UTF-8 before being sent to crypt only when crypt dies on the original string. This is handled by the new `utf8Crypt` method in `Utils.pm`.


somiaj:
Looks good. Do you have a suggestion of an easy way to test this. I should probably set up a test user and set a password with a wide character, but unsure which ones are considered wide.

drgrice1:
You could use "א".  That is such a character.  Change your password using that character (cut and paste).  Then check that you can log out and log back in again with that password.

somiaj (approved):
Confirmed this works with wide characters, while still working with old passwords.

pstaabp (approved):
Works well.  Now my password will have the poop and beer emojis  in it.

PR #2660, opened 2025-01-23
Support for custom permission levels not in the userRoles hash (replacement of #2629)

drgrice1:
This replaces #2629 and makes the changes requested for that pull request.

Currently an permission level that is not in the userRoles hash causes an exception to be thrown if the accounts manager is loaded.  This shows "Unknown: n" (where n is the unknown level) in that case and no exception is thrown.

The requested change is that the `maketext` call in #2629 was not valid.


drgrice1:
If you want to test this without going into the database, just add a custom role in localOverrides.conf such as `$userRoles{badRole} = 1`.  Then create a user with that role.  Then remove the custom role from localOverrides.conf, and load the accounts manager page.

Alex-Jordan (approved):

Issue #2661, opened 2025-01-24
PG debug messages are not shown in the problem editor.


dlglin:
The following MWE enables diagnostics for an answer checker.  The diagnostics are not displayed in the problem editor, but are displayed when the problem is viewed in a homework set.

```
DOCUMENT();

loadMacros(
    "PGstandard.pl",    # Standard macros for PG language
    "PGML.pl",          # PGML markup and Math Objects
    "PGcourse.pl",      # Customization file for the course
);

$f = Formula("x^2");

BEGIN_PGML
[_]{$f->cmp(diagnostics=>1)}
END_PGML
ENDDOCUMENT();
```

drgrice1:
The debugging messages are only shown for the `debug` format when using the rpc endpoints.  The problem editor uses the `simple` format.  So to get debugging messages we would need to switch the problem editor to use the `debug` format.

Also, there is a bug in the `formatRenderedProblem` method that wipes out all debugging messages that come from PG for LTI grade pass back (which shouldn't even have implemented in that method).

pstaabp:
Fixed in both #2662 and #2666

PR #2662, opened 2025-01-25
Fix a debugging messages in the `formatRenderedProblem` method.

drgrice1:
The `formatRenderedProblem` method of the `FormatRenderedProblem` module ensures that the `debug_messages` key of the rendered problem result is an array reference, but checks the wrong thing.  It checks if the result itself is an array reference instead of the `debug_messages` hash key of the result.  Since the result is always a hash reference and never an array reference, this means that the debugging messages from PG are always wiped out and replaced with a reference to an empty array.

This was discovered when investigating https://github.com/openwebwork/webwork2/issues/2661. If we want to switch the PG problem editor to using the `debug` output format of the `formatRenderedProblem` method we could, but in order to get the benefit of that change we need this.


drgrice1:
One way to test this is to get the MWE that @dlglin posted in #2661, and change line 399 of `htdocs/js/PGProblemEditor/pgproblemeditor.js` from `outputformat: 'simple'` to `outputformat: 'debug'`.  Then open the problem in the problem editor, enter an answer, and click "Check Answers".  With the develop branch and the change on line 399, debugging messages still will not be shown.  Of course, with this pull request they will be.

That also shows what would be needed to switch the PG problem editor to using the `debug` format if you want to test that out.  Note that shows the `debug_messages` and `internal_debug_messages` sent from PG if there are any.

drgrice1:
By the way, I don't think that anything ever sets the `internal_debug_messages` in PG anymore.  I think that is a left over relic from ages past.  I don't know why there is a `debug_messages` facility and an `internal_debug_messages` facility as well.

Alex-Jordan (approved):

somiaj (approved):

pstaabp (approved):

Issue #2646, opened 2024-12-17
issues with ExtendDueDateGW


Alex-Jordan:
`ExtendDueDateGW` is an achievement item (not included in the default item collection) that:

- adds 24 hours to a user's test template set.
- also adds 24 hours to every test version.

So for example a student who activated a test that was only supposed to be 1 hour long could use this and have the test be open for 25 hours. I think this second thing is not good and should be changed. Maybe it was not intentional.

I suggest that once a test version is open, we don't do anything at all to mess with its settings. While it would be good for accommodations to have something where some students get 1 hour, others 2 hours, etc, this should not be a reward earned from reaching a level. Not counting accommodations, timed tests should just use the same amount of time for all students in the class.

So this could be changed to only extend the date on that user's template set, not their versions. Also there should be a check so that only those tests for which a user has remaining versions are included in the list of sets to use this on.

drgrice1:
I just tested to make sure, and I was able to extend a 20 minute test version to 25 hour test.  That certainly shouldn't be.

Note that accommodations for accessibility purposes can be done by changing the test time limit for a particular user in the user's template set.  That should usually be done before a student ever starts a test, and not once a test version is in progress.

PR #2664, opened 2025-02-01
Move achievement item use form to ProblemSet page.

somiaj:
Move the achievement item use forms from the achievement page to the problem set page. The primary goal is to make less likely a student applies an achievement to the wrong set or problem when using drop down menus to select between multiple sets and problems on the achievements page, by placing the use item form on the set they wish to apply it to.

If achievement items are enabled, this adds a button to the bottom of the problem set page to "Use Achievement Reward" which opens a modal with a list of all available achievement items that can be used on the current set. Since the items now have access to the merged set and problem records, they can use this to both determine if the item
can be used, and include it in the form output. This makes the output more useful, stating what the extended due date is, or what the current and final grade would be when using an item.

If no achievement items are usable the button is disabled with a hover explaining why. Achievement items can also
no longer be used when acting as another user. In this case the button is also disabled. I decided to not setup a permission for this, because to me the items are for students, instructors can override the student data in other ways.

The achievement page now lists all earned items (including count), with a description that to use an item they must go to the set they wish to use it on and click the "Use Achievement Reward" button. When acting as a user this list shows up as well, so instructors can see what achievements a student still has available but not use them.

This is built on top of #2523 so I could update all achievement items and test them. So that should be merged first.

I have modified the gateway achievements temporarily. Due to the new setup I didn't implement applying extensions to the test versions, now the achievements only apply extensions to the template set. This is partly due to #2646, with comments that a better setup needs to be done (my thought is only allow extending open test versions that end at the same time as the template test vs ones that have a timer set).

Here is an achievement to give you 15 of all achievement items to test. If you want to test the unlimited use, set the count to a negative number.

```
$globalData->{AddNewTestGW} += 15;
$globalData->{DoubleProb} += 15;
$globalData->{DoubleSet} += 15;
$globalData->{DuplicateProb} += 15;
$globalData->{ExtendDueDate} += 15;
$globalData->{ExtendDueDateGW} += 15;
$globalData->{ExtendReducedDate} += 15;
$globalData->{FullCreditProb} += 15;
$globalData->{FullCreditSet} += 15;
$globalData->{HalfCreditProb} += 15;
$globalData->{HalfCreditSet} += 15;
$globalData->{NoReducedCred} += 15;
$globalData->{ReducedCred} += 15;
$globalData->{ResetIncorrectAttempts} += 15;
$globalData->{ResurrectGW} += 15;
$globalData->{ResurrectHW} += 15;
$globalData->{SuperExtendDueDate} += 15;
$globalData->{SuperExtendReducedDate} += 15;
$globalData->{Surprise} += 5;

return 1;
```


lib/WeBWorK/AchievementItems/DoubleSet.pm
```diff
@@ -30,62 +29,34 @@ sub new ($class) {
...
-		WeBWorK::AchievementItems::form_popup_menu_row(
-			$c,
-			id         => 'dub_set_id',
-			label_text => $c->maketext('Set Name'),
-			values     => \@openSets,
-			menu_attr  => { dir => 'ltr' }
-		)
-	)->join('');
+	return $c->tag('p',
+		$c->maketext('Increase this assignments total number of points from [_1] to [_2].', $total, 2 * $total));
```
Alex-Jordan, line 42:
```suggestion
		$c->maketext("Increase this assignment's total number of points from [_1] to [_2].", $total, 2 * $total));
```

lib/WeBWorK/AchievementItems/FullCreditSet.pm
```diff
@@ -30,59 +29,38 @@ sub new ($class) {
...
-		WeBWorK::AchievementItems::form_popup_menu_row(
-			$c,
-			id         => 'fcs_set_id',
-			label_text => $c->maketext('Set Name'),
-			values     => \@openSets,
-			menu_attr  => { dir => 'ltr' }
-		)
-	)->join('');
+sub print_form ($self, $set, $records, $c) {
+	return $c->tag('p', $c->maketext(q(Increase this assignments' grade from [_1]% to 100%.), $self->{old_grade}));
```
Alex-Jordan, line 48:
```suggestion
	return $c->tag('p', $c->maketext(q(Increase this assignment's grade from [_1]% to 100%.), $self->{old_grade}));
```

lib/WeBWorK/AchievementItems/HalfCreditSet.pm
```diff
@@ -30,64 +29,47 @@ sub new ($class) {
...
-			$c,
-			id         => 'hcs_set_id',
-			label_text => $c->maketext('Set Name'),
-			values     => \@openSets,
-			menu_attr  => { dir => 'ltr' }
+sub print_form ($self, $set, $records, $c) {
+	return $c->tag(
+		'p',
+		$c->maketext(
+			q(Increase this assignments' grade from [_1]% to [_2]%.),
```
Alex-Jordan, line 54:
```suggestion
			q(Increase this assignment's grade from [_1]% to [_2]%.),
```

lib/WeBWorK/AchievementItems/ResurrectHW.pm
```diff
@@ -30,70 +31,49 @@ sub new ($class) {
...
+	# This may require also extending the answer date.
+	if ($set->due_date > $set->answer_date) {
+		$set->answer_date($set->due_date);
+		$userSet->answer_date($set->answer_date);
+	}
+	$db->putUserSet($userSet);
 
-	return;
+	return $c->maketext(
+		'Closing date of this assignment extened by 24 hours to [_1].',
```
Alex-Jordan, line 74:
```suggestion
		'Closing date of this assignment extended by 24 hours to [_1].',
```

lib/WeBWorK/AchievementItems/SuperExtendDueDate.pm
```diff
@@ -35,69 +34,56 @@ sub new ($class) {
...
+	}
 	$db->putUserSet($userSet);
 
-	$globalData->{ $self->{id} }--;
-	$globalUserAchievement->frozen_hash(nfreeze_base64($globalData));
-	$db->putGlobalUserAchievement($globalUserAchievement);
-
-	return;
+	return $c->maketext(
+		'Closing date of this assignment extened by 48 hours too [_1].',
```
Alex-Jordan, line 84:
```suggestion
		'Closing date of this assignment extended by 48 hours too [_1].',
```

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -35,69 +34,56 @@ sub new ($class) {
...
+	}
 	$db->putUserSet($userSet);
 
-	$globalData->{ $self->{id} }--;
-	$globalUserAchievement->frozen_hash(nfreeze_base64($globalData));
-	$db->putGlobalUserAchievement($globalUserAchievement);
-
-	return;
+	return $c->maketext(
+		'Closing date of this assignment extened by 24 hours to [_1].',
```
Alex-Jordan, line 84:
```suggestion
		'Closing date of this assignment extended by 24 hours to [_1].',
```

lib/WeBWorK/AchievementItems/ResurrectGW.pm
```diff
@@ -33,63 +34,38 @@ sub new ($class) {
...
+		$set->reduced_scoring_date($set->reduced_scoring_date + ONE_DAY);
+		$userSet->reduced_scoring_date($set->reduced_scoring_date);
+	}
+	$set->due_date($set->due_date + ONE_DAY);
+	$userSet->due_date($set->due_date);
+	$set->answer_date($set->answer_date + ONE_DAY);
+	$userSet->answer_date($set->answer_date);
+	$db->putUserSet($userSet);
+
+	return $c->maketext('Close date of this test extened 24 hours to [_1].',
```
Alex-Jordan, line 67:
```suggestion
	return $c->maketext('Close date of this test extended 24 hours to [_1].',
```

lib/WeBWorK/AchievementItems.pm
```diff
@@ -41,65 +44,125 @@ use constant ITEMS => [ qw(
...
+		}
+		return \@new_items;
+	}
 	return \@items;
 }
 
+# Method that returns a string with the achievement name and number of remaining items.
+sub remaining_title ($self, $c) {
+	if ($self->count > 1) {
+		return $c->maketext('[_1] ([_2] remaining)', $c->maketext($self->name), $self->count);
```
Alex-Jordan, line 129: Is it not OK to say `(1 remaining)`? I think the `count == 1` case can be combined with the `count < 1` case.

lib/WeBWorK/AchievementItems/DoubleProb.pm
```diff
@@ -32,84 +29,49 @@ sub new ($class) {
...
-	return "You are $self->{id} trying to use an item you don't have" unless $globalData->{ $self->{id} };
-
-	my $setID = $c->param('dbp_set_id');
-	return 'You need to input a Set Name' unless defined $setID;
+sub print_form ($self, $set, $records, $c) {
+	return WeBWorK::AchievementItems::form_popup_menu_row(
+		$c,
+		id         => 'dbp_problem_id',
+		label_text => $c->maketext('Problem Number'),
+		first_item => $c->maketext('Choose problem to double.'),
```
Alex-Jordan, line 41: I'm not sure "Choose problem to double" has a clear meaning to a student. Double the score? Double the problem itself, adding a copy to the set? No, it means double the weight. ("value" in internal WW lingo but I think "weight" is better.)

Alex-Jordan: And there is another unclear use of "double" below.

lib/WeBWorK/AchievementItems/HalfCreditProb.pm
```diff
@@ -32,87 +29,60 @@ sub new ($class) {
...
-	$problem->status($new_status);
-	$problem->sub_status($new_status);
-
-	$db->putUserProblem($problem);
+sub print_form ($self, $set, $records, $c) {
+	return WeBWorK::AchievementItems::form_popup_menu_row(
+		$c,
+		id         => 'half_cred_problem_id',
+		label_text => $c->maketext('Problem Number'),
+		first_item => $c->maketext('Choose problem to increase 50%.'),
```
Alex-Jordan, line 46: This is similar to my comments about "double". Here it is also unclear, but the true meaning is different. This one is about increasing the score (internally, status) by 50%. The only language mentions adding "half credit" to which is one way to clear up the ambiguity.

There are possibly more like this for other items.

Alex-Jordan:
I made some typo suggested edits (which you could ignore in GH and just fold into your next commit if you like). And I left some other comments about certain wording. I haven't done a comprehensive review of all the wordings.

This will be better than what we have now. I think this came up in discussion at a recent meeting, do I remember right?

I think on the Achievements page, the Reward item titles no longer need the `fs-3` class. With that there, now that all the form content is gone, it looks too heavy.

Now there are three buttons:
```
Download Hardcopy for Current Set
Use Achievement Reward
Email Instructor
```

I wonder if it would look better (in a wide enough screen) like:
```
Download PDF                Use Reward                   Email Instructor
```

Perhaps beyond the scope of this PR, but if those buttons became icons and were placed more near the top of the page, I think it would be better. That includes "Email Instructor" on other pages besides a problem page. Maybe this is just my opinion.

lib/WeBWorK/AchievementItems.pm
somiaj: This was taken from the old code, so this is what was done before. I'm okay with simplifying this if desired.

lib/WeBWorK/AchievementItems/DoubleProb.pm
somiaj: The problem set page calls the column `Worth` if that matters. I think I initially though similar things, but couldn't think of a good way to phrase it at the time, then changing the vocabulary got lost when I was more focused on layout and making sure things worked. I'm okay with any suggested change here to clarify things.

Another thing I noticed I didn't update, is all the drop down forms are just labeled as 'Problem number', maybe we should make the labels reflect what the problem number is for as well. Just something that slipped by after I got the initial form up.

somiaj:
@Alex-Jordan thanks, I rebased the branch onto develop and updated all the typos.

somiaj:
@Alex-Jordan I updated a bunch of your suggestions. Leaving the button layout alone for now. If it is agreed to change I can wrap it in here, or it can be a future PR.

Alex-Jordan:
I believe all current achievements reward items are something you apply to a set. (Possibly to a problem or problems within a set.)

Would it make sense to ever have an achievement reward item that does not apply to a particular set? Like  a reward item where across the course, all exercises where you earned partial credit get elevated to 100%. Or an item that reopens all sets but reduced credit is in effect. Grant additional attempts to all problems across the course. Etc.

If such an item existed, how do you imagine giving students the ability to use it? Would it just always show up among the items in the modal dialog for the "Use Achievement Item" button in any set's problem list page? Or would such things be activatable from the Achievements page? Before we do away with the ability to activate these from the Achievements page, I wanted to ask if we are sure that would be OK.

somiaj:
My original approach was to allow this 'Use Achievement Reward' button to be used on multiple pages, and the `can_use` method just needs to be able to know what page to be able to adjust how it determines if it is usable or not. So it should be possible to extend things so they work better from any page. Initially I put the button on both set and problem pages, but the problem page led to lots of issues about using it with an undefined, or for show me another. So I just settled and moving the button only to the problem set page for now since it has access to the set record and all problem records without additional database calls.

That being said, if things a need arises to allow using achievements from different pages, it should be setup to make adding that feature not too difficult, but I would think about that for future work. I just stuck to a single page for now for simplicity and currently we don't have such achievements.

Alex-Jordan:
OK, it's good that however things are working for this, it is extensible to possible future situations.

drgrice1:
There is a problem with this pull request that needs to be fixed.  If `$achievementItemsEnabled` is false (i.e. if `Enable Achievement Rewards` is set to `False` in the course configuration), then the `Use Achievement Reward` button should not be shown as it currently is in this pull request.

somiaj:
@drgrice1 thanks for catching that, should be fixed now.

lib/WeBWorK/AchievementItems.pm
```diff
@@ -44,65 +44,128 @@ use constant ITEMS => [ qw(
...
-		)
-	)->join('');
+	unshift(@{ $params{values} }, [ $params{first_item} => '' ]) if $params{first_item};
+
+	my $row_contents = $c->tag(
+		'div',
+		class => 'form-floating',
+		$c->c(
+			$c->select_field($params{id} => $params{values}, %{ $params{menu_attr} }),
+			$c->label_for($params{id} => $params{label_text}, %{ $params{label_attr} })
```
drgrice1, line 164: This needs to be
```suggestion
			$c->select_field($params{id} => $params{values}, id => $params{id}, %{ $params{menu_attr} }),
			$c->label_for($params{id} => $params{label_text}, %{ $params{label_attr} })
```
The select must have an `id` for the label to refer to. The value of the `for` attribute of the `label` element must be the ID of a non-hidden form control.

lib/WeBWorK/AchievementItems/ResurrectGW.pm
```diff
@@ -33,63 +34,38 @@ sub new ($class) {
...
+		$set->reduced_scoring_date($set->reduced_scoring_date + ONE_DAY);
+		$userSet->reduced_scoring_date($set->reduced_scoring_date);
+	}
+	$set->due_date($set->due_date + ONE_DAY);
+	$userSet->due_date($set->due_date);
+	$set->answer_date($set->answer_date + ONE_DAY);
+	$userSet->answer_date($set->answer_date);
+	$db->putUserSet($userSet);
+
+	return $c->maketext('Close date of this test extended 24 hours to [_1].',
```
drgrice1, line 67: For consistency this should be
```suggestion
	return $c->maketext('Close date of this test extended by 24 hours to [_1].',
```

lib/WeBWorK/AchievementItems/ResurrectHW.pm
```diff
@@ -30,70 +31,48 @@ sub new ($class) {
...
+	# This may require also extending the answer date.
+	if ($set->due_date > $set->answer_date) {
+		$set->answer_date($set->due_date);
+		$userSet->answer_date($set->answer_date);
+	}
+	$db->putUserSet($userSet);
 
-	return;
+	return $c->maketext(
+		'Closing date of this assignment extended by 24 hours to [_1].',
```
drgrice1, line 73: Change this to
```suggestion
		'Close date of this assignment extended by 24 hours to [_1].',
```

lib/WeBWorK/AchievementItems/SuperExtendDueDate.pm
```diff
@@ -35,69 +34,58 @@ sub new ($class) {
...
+	}
 	$db->putUserSet($userSet);
 
-	$globalData->{ $self->{id} }--;
-	$globalUserAchievement->frozen_hash(nfreeze_base64($globalData));
-	$db->putGlobalUserAchievement($globalUserAchievement);
-
-	return;
+	return $c->maketext(
+		'Closing date of this assignment extended by 48 hours too [_1].',
```
drgrice1, line 86: Change this to
```suggestion
		'Close date of this assignment extended by 48 hours to [_1].',
```

lib/WeBWorK/AchievementItems/SuperExtendReducedDate.pm
```diff
@@ -36,75 +35,50 @@ sub new ($class) {
...
+	$userSet->reduced_scoring_date($set->reduced_scoring_date);
 	$db->putUserSet($userSet);
 
-	$globalData->{ $self->{id} }--;
-	$globalUserAchievement->frozen_hash(nfreeze_base64($globalData));
-	$db->putGlobalUserAchievement($globalUserAchievement);
-
-	return;
+	return $c->maketext(
+		'Reduced scoring date of this assignment exted by 48 hours to [_1].',
```
drgrice1, line 79: Change this to
```suggestion
		'Reduced scoring date of this assignment extended by 48 hours to [_1].',
```

templates/ContentGenerator/ProblemSet/use_achievement_items.html.ep
```diff
@@ -0,0 +1,56 @@
...
+									<%= $form %>
+									% unless ($item->{hideUseButton}) {
+										<input class="btn btn-primary" type="submit"
+											name="use_achievement_item_<%= $item->id %>"
+											value="<%= maketext('Use [_1]', $item->remaining_title($c)) %>">
+									% }
+								<%= end =%>
+							</dd>
+						% }
+					</dl>
```
drgrice1, line 36: This should be
```suggestion
					<div class="modal-body">
						<dl>
							% for (@$achievementItems) {
								% my ($item, $form) = @$_;
								<dt class="fs-4"><%= $item->name %></dt>
								<dd class="mx-3 mb-4">
									<%= form_for current_route, method => 'POST', name => 'use_reward', begin =%>
										<%= hidden_field 'use_achievement_item_id' => $item->id %>
										<%= $form %>
										% unless ($item->{hideUseButton}) {
											<input class="btn btn-primary" type="submit"
												name="use_achievement_item_<%= $item->id %>"
												value="<%= maketext('Use [_1]', $item->remaining_title($c)) %>">
										% }
									<%= end =%>
								</dd>
							% }
						</dl>
					</div>
```
Give the modal its proper structure.

Also it wouldn't be a bad idea to add
```perl
 <div class="modal-footer">
        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
 </div>
```
immediately after the above.  It is good to have a "Close" button at the bottom.

lib/WeBWorK/AchievementItems.pm
```diff
@@ -44,65 +44,128 @@ use constant ITEMS => [ qw(
...
-			$c->select_field($params{id} => $params{values}, id => $params{id}, %{ $params{menu_attr} })
-		)
-	)->join('');
+	unshift(@{ $params{values} }, [ $params{first_item} => '' ]) if $params{first_item};
+
+	my $row_contents = $c->tag(
+		'div',
+		class => 'form-floating',
+		$c->c(
+			$c->select_field($params{id} => $params{values}, %{ $params{menu_attr} }),
```
drgrice1, line 163: It would be good to change this to
```suggestion
			$c->select_field($params{id} => $params{values}, required => undef, %{ $params{menu_attr} }),
```
This makes it so the form won't submit if the user has not selected an item, and the message "Please select an item in the list." is shown when the user clicks the submit button without selecting a problem number.

lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -35,69 +34,58 @@ sub new ($class) {
...
+	}
 	$db->putUserSet($userSet);
 
-	$globalData->{ $self->{id} }--;
-	$globalUserAchievement->frozen_hash(nfreeze_base64($globalData));
-	$db->putGlobalUserAchievement($globalUserAchievement);
-
-	return;
+	return $c->maketext(
+		'Closing date of this assignment extended by 24 hours to [_1].',
```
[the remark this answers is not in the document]
somiaj: I'm assuming the extra `'` at the start and end are typos here and not adding them.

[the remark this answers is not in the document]
drgrice1: Yeah.

somiaj:
@drgrice1 Thanks for the review. I have committed all your suggested changes and removed the number of remaining items from the remaining title of the Surprise achievement item.

I am currently out of town so I am unable to test the changes, I just had some down time to commit them. I will test further when I return home next week.

drgrice1 (approved):
I think this is looking good now.  I think this will be a much nicer way to use achievement rewards.  Thanks for your work on this.

PR #2665, opened 2025-02-02
make only outer edge of achivements images transparent and fix a miss…

Alex-Jordan:
…pelling

In the thread for #2579, @drgrice1 observed issues with the transparency for certain achievement images. I think the changes here address that.

While doing this I saw "persistance" used repeatedly instead of "persistence", so I changed them all to "persistence".

One more thing. On MacOS, it turns out that when a directory has name with a `.dist` extension, MacOS thinks it is a certain special kind of file instead of thinking it is a directory. For example, I could not `cd` into this folder until I renamed it temporarily. Would there be any objection to renaming `courses.dist` to something else? And if not, what is a good name? I could add that change to this PR (although it is only loosely related and could be separate).


somiaj (commented):
What is this binary file `courses.dist/modelCourse/html/.DS_store` doing, probably something on your end that sneaked into the PR?

Alex-Jordan:
Thanks for seeing that. It's a MacOS file that (I think) stores metadata for a folder. It get created if you visit a folder using Finder. I did this on my local machine instead of on the PCC development server since it was easier to maniuplate images that way, and this snuck in.

drgrice1:
The `.DS_store` is usually listed in the `.gitignore` file to prevent MacOS users from accidentally committing those files.  You should add it to the beginning of the file in the first group of ignores.  Those are all this sort of thing.  Vim creates `.bak` and `.swp` files that we don't want committed (depending on your vim settings).  Something else probably creates a `.save` file.

Alex-Jordan:
For most of these, I had ImageMagick:

1. Convert all transparent background to white.
2. Then use a command that floodfills transparency around the border until reaching pixels that are different from white.

That did not work for a few of them like `complete_100_problems.png`. But for those it was easy to specify coordinates to floodfill white and that worked out, because the black in those images happens to go all the way to the edge in the right places.

drgrice1 (approved):
Looks good.  I marked this as needing a comment in release notes.  We need to make sure that the updated files get copied into the model course as well as any course using achievements.  Otherwise no one will see the benefits of this pull request!

Alex-Jordan:
For several reasons, I think that we could look into centralizing these image files and the `.at` and `.axp` files similar to how hardcopy themes are centralized.

For one thing, it's the point you make, that courses propagate with old versions of things. There's not much to be done about that since an old course can be unarchived at any time. The release notes will help but people will forget. (I have the same issue with the course `login_info.txt` file in courses at PCC, where I just can't eradicate an old version unless I went to the trouble of unarchiving all old courses.)

Also it's a bit wasteful to have these image files copied in course after course after course, then kept in archive after archive. I took a course archive file, unpacked it, deleted the achievements stuff, and re-zipped it. The archive file went down from 2.0MB to 1.1MB. That would be a big effect over lots of courses over the years.

drgrice1:
I was thinking the same thing, and completely agree.

Issue #2669, opened 2025-02-13
test with bad group still leads to version creation, except missing data


Alex-Jordan:
If you have a test named `myTest`, where a problem is defined like `group:mySet`. But suppose there is no set named `mySet`. You will get an error screen about no problems being available in `mySet`. (And maybe you would get the same message if `mySet` exists but has zero problems in it.)

OK, fine. But your user will still get assigned to `mySet,v1`. And `mySet,v1` has been partly created, but not completely. After navigating away from the above error screen, a user can try to go to version 1 of the test. Now they will get a different error message complaining about the set not have a close date. But the real issue is that lots of things about the version were not initialized. The initialization process was aborted midway without cleaning up after itself.

pstaabp:
Fixed in #2670

PR #2670, opened 2025-02-13
Actually handle failure to find a problem from a grouping set gracefully.

drgrice1:
Actually, this handles any failure to generate a new test version gracefully. This wraps the set version assignment call in an `eval` and catches any exceptions that occur in the process including issues from an instructor not setting up grouping sets correctly.  If any exceptions do occur the user is shown the usual invalid set screen with a message stating, "Unable to generate a valid test version. This is usually caused by invalid usage of grouping sets or a database error. Please speak to your instructor to fix the error. A system administrator can obtain more details on this error from the logs." The actual exception is logged to the webwork2 application log file (logs/webwork2.log) so an instructor can ask a system administrator for details of the error. Note this does not add to the general webwork2 issue of showing exceptions in the user interface.

Edit: I forgot to mention that if a test version is created in the process when an exception occurs, then it is immediately (and quietly) deleted.

This is one way to resolve issue #2669 and probably the best way without massively revising the GatewayQuiz module.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -550,55 +550,68 @@ async sub pre_header_initialize ($c) {
 				)
 				)
 			{
-				# Assign the set, get the right name, version number, etc., and redefine the $set and $problem for the
-				# remainder of this method.
+				# Attempt to assign the set.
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
Fixes the issue more gracefully than before.  And +1 for fixing a known issue that is 20 years old.

PR #2675, opened 2025-02-16
Switch the feedback emails to HTML.

drgrice1:
The email content is generated by the `templates/ContentGenerator/Feedback/feedback_email.html.ep` template. All of the same content is there, just arranged nicely into tables and such. The link to the page the user sent the email from is of course an `a href`.  For now it reads "To visit the page from which (user name) sent feedback, click here." where "click here" is the link. Note that it used to read "To visit the page from which the user sent feedback, long link."  We have the user name so why not use it here?

Also, when the sent message is displayed for the user in the browser, the message is formated much the same as it now is in the email.  That is it is not wrapped and placed in a `pre` tag.  Instead it is in a bordered `div` formatted with essentially the same style as in the email.

Here is a screen shot of the email with the default verbosity of 1.  At least what I could fit in.  Below what is show are the tables for "Data about the problem" and "Data about the homework set".  They look much the same as the tables shown.
feedback-email

I would like to reorder the tables.  I would like to have "Data about the problem" first (assuming the email is sent from a problem page) since that is the information that is most useful in the case that a user is asking about a problem.  Then "Data about the homework set" next.  Then "Data about the user".  The user data is almost all shown above in the other content or the subject anyway, so it doesn't need to be so high.  Then the "Data about the problem processor" last.  I generally consider that information mostly useless.  It definitely should not be at the top, and I would even vote for not even including it in the default verbosity.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Feedback.pm
+++ b/lib/WeBWorK/ContentGenerator/Feedback.pm
@@ -25,10 +25,9 @@ WeBWorK::ContentGenerator::Feedback - Send mail to professors.
 use Data::Dumper;
 use Email::Stuffer;
 use Try::Tiny;
-use Text::Wrap qw(wrap);
 
 use WeBWorK::Upload;
-use WeBWorK::Utils qw(decodeAnswers createEmailSenderTransportSMTP fetchEmailRecipients);
+use WeBWorK::Utils qw(createEmailSenderTransportSMTP fetchEmailRecipients);
 
 # request paramaters used
 #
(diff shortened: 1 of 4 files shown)


somiaj:
Isn't "click here" considered bad practice for an accessible link? To me I don't see needing two links, one to the webwork server and one to the page the email was sent from, could that be combined into a single link to the page the email was sent from?

drgrice1:
As I said, the "click here" is for now.  Yes, the "click here" is bad practice, and I was hoping someone would offer a better suggestion.

The webwork server is not technically a link.  It is just a URL.  However, most email clients present any URL as a link even if they are not `a href`ed.  That was in the email before.  I didn't change that.

Alex-Jordan:
This is great and I'll approve it when I can get to a computer. Just on
phone now. I'm sure we can evolve it in a subsequent PR.

somiaj:
Thanks, I am trying to think of suggestions but can't think of anything I'm sold on. But I am thinking something along the lines of combining the two lines into a single line something like:

```
Message from $first_name $last_name ($username) via <this WeBWorK page>.
```

I was also considering we could put the institution there, but still can't think of any phrasing I'm sold on.

Alex-Jordan:
Do we need to care about faculty who use an email client that won't process HTML in email? I notice that now what is sent is not very human readable. In addition to the HTML you get a bunch of `=09` in the message for the tabs that are used to indent the HTML tags.

drgrice1:
@Alex-Jordan: Thanks for reminding me about a text fallback.  I meant to do that, but forgot.  I have added that now.

The text message is generated by the `templates/ContentGenerator/Feedback/feedback_email.txt.ep` template. The text message is essentially the same as the prior text email. If your email client is set to show text, then you will see this message. If your client is set to show HTML, then you will see the new HTML message. If your client does not support HTML, then you will see the text message.

pstaabp:
First @drgrice1, love the feedback message sent from your fake student.  Pure honesty. 

Overall, this looks great.  How about something along the lines of: STUDENT NAME sent feedback about THIS WeBWorK PAGE

where the end is the link.  I was thinking "this WeBWorK problem", but there are non-problem related pages that students can email about.

drgrice1:
I changed the line with the "click here" link to
```
<%= $user->first_name %> sent feedback from <%= link_to 'this page' => $emailableURL %>.
```

Lets go with that for now.  The previous line is still the same.  This can be tweaked in a later pull request if desired, but I think for this pull request focus on the HTML/text functionality aspects.

pstaabp (approved):

somiaj (approved):
Here is my approval for the HTML switch. I wasn't able to test the text fall back for a non HTML email client, but assuming Email::Stuffer is doing its job correctly.

I still hope someone decides to come clean up the actual content of the feedback a bit as per suggestions here, but agree this is just about switching to HTML.

Alex-Jordan:
Extending some discussion here, perhaps inappropriately. We are using the new HTML emails in production and just in the first few days of the term, I've had enough feedback from faculty to suggest I need to make changes. Here is what things are like with develop:

Screenshot 2025-04-02 at 3 36 30 PM

-----

And here is what it looks like with a few changes I made:

Screenshot 2025-04-02 at 3 36 51 PM

One important change is there is only one link. Faculty kept thinking they should click the first link to go to the page where the student sent a problem (because they are not reading the message closely). And then I condensed the first three "p" into just one. And removed most borders that feel like clutter to me. But I made the table headers stand out more.

Are these good changes for a PR? Anything else to do or do differently?

somiaj:
@Alex-Jordan I think the discussion ended that the PR mostly just switched to HTML format and that a future PR would be good to modify the actual content and/or layout. I think some discussion about maybe we shouldn't call the link 'this page' could be added if you desire. But yea, add a PR with the changes you think based on the feedback you got, include anything else you may find useful is my suggestion.

drgrice1:
Looks good to me. Marie the pull request!

PR #2677, opened 2025-02-19
PG Problem Editor file selector

drgrice1:
This changes the PG Problem Editor to not assume a file type when opened without a file type specified.  So when the problem editor is opened from the site navigation link it does not directly open to the blank file template.  Instead it opens to a file chooser page. On that page, three options are presented.  The user may choose a "New Problem Template", a "File" and give a path relative to the course templates directory, or a "Sample Problem" and pick a sample problem file from a dropdown.

In order to support sample problems a new file type needed to be implemented for the problem editor.  The sample problem parsing module for PG is used to strip out the documentation when the problem is loaded into the editor.

However, I observed that the current code for that used by the sample problem viewer to offer a sample problem file for download is highly innefficient.  It currently reads all of the sample problem files, then reads all of the PG macros, and then parses the desired problem stripping out documentation, and finally serves the file. So the PG pull request https://github.com/openwebwork/pg/pull/1198 adds a `getSampleProblemCode` method that simply reads only the desired sample problem and efficiently strips out the documentation without needing the macros and sample problem metadata. So instead of needing to read 253 files (the number of macros plus the number of sample problems currently), it only reads 1.

To test this make sure to check out https://github.com/openwebwork/pg/pull/1198.

### DIFF
--- a/htdocs/js/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/PGProblemEditor/pgproblemeditor.js
@@ -1,4 +1,57 @@
 (() => {
+	const fileChooserForm = document.forms['pg-editor-file-chooser'];
+	if (fileChooserForm) {
+		const newProblemRadio = document.getElementById('new_problem');
+
+		const sourceFilePathInput = fileChooserForm.elements['sourceFilePath'];
+		const filePathRadio = document.getElementById('file_path');
+
+		const sampleProblemFileSelect = fileChooserForm.elements['sampleProblemFile'];
+		const sampleProblemRadio = document.getElementById('sample_problem');
+
+		newProblemRadio?.addEventListener('change', () => {
+			if (newProblemRadio.checked) {
+				sampleProblemFileSelect.required = false;
+				sourceFilePathInput.required = false;
+			}
+		});
+
+		if (filePathRadio && sourceFilePathInput) {
+			const filePathSelected = () => {
+				sampleProblemFileSelect.required = false;
+				sourceFilePathInput.required = true;
+				filePathRadio.checked = true;
+			};
+			filePathRadio.addEventListener('change', () => {
+				if (filePathRadio.checked) filePathSelected();
+			});
+			sourceFilePathInput.addEventListener('focusin', filePathSelected);
+		}
+		if (sampleProblemRadio && sampleProblemFileSelect) {
+			const sampleProblemSelected = () => {
+				sampleProblemFileSelect.required = true;
+				sourceFilePathInput.required = false;
+				sampleProblemRadio.checked = true;
+			};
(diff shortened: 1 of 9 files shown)


drgrice1:
I forgot to mention that I made the editor NOT show the "Append form for a new blank problem as it was before.  That is because that doesn't work.  If you use it, it adds the path `/opt/webwork/webwork2/assets/pg/newProblem.pg` to the set, and that path doesn't work quite right in a set.  It actually works when you view the problem (and really shouldn't -- that is a security vulnerability that needs to be investigated), but it shows "This source file does not exist!" when you view the set on the set detail page.  Really, the new problem should be saved to a new file in the course templates directory before adding it to a set.

pstaabp:
I'm getting the following error:
```
"getCleanSampleProblemCode" is not exported by the SampleProblemParser module Can't continue after import errors at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm line 23. BEGIN failed--compilation aborted at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/SampleProblemViewer.pm line 23. Compilation failed in require at (eval 744) line 1.
```

I think you mean to import "getSampleProblemCode".

drgrice1:
Oops.  I changed the method name on the PG side at some point and forgot to change it here also.

pstaabp:
I know we had talked about this.  Maybe we can add some help to the sample problems to explain how to use the problem editor to get the problem.  

Also, I'm thinking we might be able to add an option on the "Save As" in the Problem Editor to append to an existing set to simplify a bit more.  

Neither of these should be part of this PR, just brainstorming.

Alex-Jordan:
This is great. I'm just going to mention a thought, and defer to @drgrice1 about if it would be worth pursuing. What if there were a fourth option when you go to the Editor, that is basically like the File Manager?  So you could browse files, enter folders, and then select a file to edit? It would be redundant, since you can also just go to File Manager if that's what you want. It only occurs to be if I was of a mind to go edit a file in some folder in my course, and on autopilot, I might go to the Editor instead of the File Manager. And I might not be able to recall (or correctly type) the path.

Alex-Jordan (approved):
Approving, even with the comment/question I just left.

drgrice1:
I think that would be nice, and I also thought of something like that.  I think that will take a bit more effort to achieve though.  So perhaps that could be done in a future pull request.

As I mentioned in the meeting last week, I think there are other possibilities for options on this page as well.  Like opening a file on the user's local computer directly in the editor.

pstaabp (approved):
I think this is a good start and we can update the Problem Editor in a later PR.

As I mentioned I think we can do some other streamlining on the "Save As" tab as well.

PR #2682, opened 2025-03-13
Check paths before rendering static files.

drgrice1:
Test quickly, merge quickly. Don't discuss here.

See the email sent to you about this.


dlglin (approved):
Tested on my dev box.  With two approvals I will merge.

PR #2686, opened 2025-03-18
when copying a course including its sets, don't copy the lis_source_i…

Alex-Jordan:
…d of the sets

Before this change, if a new WW course was cloned from an old one (including its sets) then the new WW course may end up attempting to send grades to the old LMS course.


drgrice1 (approved):
Looks good.

Issue #2680, opened 2025-03-03
Error creating/accessing backups for pg filenames containing parentheses


dlglin:
In the problem editor, saving a file with parentheses in the name with the "create backup" box ticked leads to errors.

Steps to reproduce:
- Create a pg file with parentheses in the name.  e.g. in the problem editor, Save as `p(.pg`.  The file is created successfully.
- Now click "Save" with the Create backup box ticked.  You will get an error: Unmatched ( in regex; marked by <-- HERE in m//opt/webwork/courses/Parens/templates/tmpEdit/q) <-- HERE .pg.bak/ at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm line 604.

If instead you have a filename with matched parentheses (e.g. "p().pg") you get a different, uglier error.

pstaabp:
This was supposed to autoclose.

PR #2688, opened 2025-03-18
Properly quote file names in the regexp used to extract backup times.

drgrice1:
If a backup filename contains certain special characters, then an exception is thrown when used in the regular expression to extract the time from the backup filename. To protect against this the metaquoting escape sequence (`\Q...\E`) is needed.

This fixes issue #2680.

Edit: Also allow parenthesis in file names in the file manager.  I went to delete the file I created to test the first part of this pull request, and discovered that I could not because the FileManager `checkPWD` method considers parentheses (or really anything except `-_./A-Z0-9~, `) to be an illegal character.  This probably needs to be revisited.  There are lots of other valid characters in file names that this does not allow.


Alex-Jordan (approved):

pstaabp (approved):
fixed.

PR #2689, opened 2025-03-30
Rework the warning and exception handling in the config file safe compartment.

drgrice1:
This is so that evaluation of the configuration files gives better messages.

Currently if a file is included from another included file then exceptions in the inner included file are not shown, but rather the exception from the outer included file is shown.  For example, if `include('authen_LTI.conf')` is called in `localOverrides.conf` which is included from `defaults.config`, and `authen_LTI.conf` does not exist, then the current message says `Failed to read include file /opt/webwork/webwork2/conf/localOverrides.conf (has it been created from the corresponding .dist file?): No such file or directory` and now it says `Failed to read include file /opt/webwork/webwork2/conf/authen_LTI.conf (has it been created from the corresponding .dist file?): No such file or directory`. This was fixed by adding a local die signal handler to the `include` method defined in the safe compartment.  That pushes messages onto a `state` array variable.  When an exception occurs, only the last message is shown which is the one from the inner inclusion since that exception is pushed onto the array first (the exception from the outer inclusion happens second as a result of the inner exception).

In addition if warnings occur inside an included file, then this actually becomes an exception.  The reason for this is that the global warning signal handler defined in `lib/Mojolicious/WeBWorK.pm` is called inside the safe compartment used for evaluating the configuration files. This causes an exception when that handler attempts to access the controller `stash` which is not defined in the safe compartment. Note that this only occurs for a request, and not when the application first loads because that doesn't go through the global warning signal handler. Although, that isn't the exception you see because when that exception occurs Mojolicious attempts to wrap that exception with a `Mojo::Exception`. That causes another exception because that package is also not shared to the safe compartment. So you get an exception about the `new` method of the `Mojo::Exception` package not existing which is completely uninformative.  To fix this a local warning signal handler is used that pushes messages into a warning array.  Those warnings are rebroadcast to the global warning signal handler after the safe compartment evaluation of the configuration files is complete. A simple way to test this is to add a `warn` statement to `localOverrides.conf` and try to open a page in the browser.

### DIFF
--- a/lib/WeBWorK/CourseEnvironment.pm
+++ b/lib/WeBWorK/CourseEnvironment.pm
@@ -51,6 +51,7 @@ safe compartment into a hash. This hash becomes the course environment.
 
 use strict;
 use warnings;
+use feature 'state';
 
 use Carp;
 use Opcode qw(empty_opset);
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Tested the warn inside 'local overrides' and showing the warning like other warnings in WW instead of the whole process dying.

somiaj (approved):

drgrice1:
I converted this to a draft for now.  The signal handling is causing some of the same conflicst that were seen before with async/await.

drgrice1:
This is fixed now.  Instead of using a local `$SIG{__DIE__}` handler an array variable is shared to the safe compartment, and error messages pushed onto that array.  The first one in the array, if any, is used for the exception.

This issue is a little different than the `async/await` issue I observed with the `$SIG{__WARN__}` handler.  This was an issue with the `Mojo::Reactor::EV` package, and the local `$SIG{__DIE__}` handler ends up leaving the app in an unrecoverable state.  Even SIGINT (`Ctrl-c` in the terminal) won't close the app.  You have to resort to SIGQUIT (`Ctrl-\` in the terminal).

PR #2692, opened 2025-04-02
Fix an issue with conflicting JSON modules in use in FormatRenderedProblem.pm.

drgrice1:
Both the `JSON` module and `Mojo::JSON::encode_json` method are loaded in `lib/FormatRenderedProblem.pm`.  This causes first request issued to the `render_rpc` endpoint after each server process starts to issue the warning `Prototype mismatch: sub FormatRenderedProblem::encode_json ($) vs none at /usr/lib/x86_64-linux-gnu/perl-base/Exporter.pm line 63.`  Once the modules are loaded in the process they aren't loaded again,  so the warning only occurs once per process.

There is no need for both modules that both do essentially the same thing.  The `Mojo::JSON::encode_json` method also UTF-8 encodes the data as was done before.

Note that the only place the `JSON` module was used was for rendering the `raw` format.

This is just something I noticed while working on #2691.

### DIFF
--- a/bin/OPL-update-legacy
+++ b/bin/OPL-update-legacy
@@ -539,8 +539,6 @@ if($canopenfile) {
 }
 #### End of taxonomy/taxonomy2
 
-use JSON;
-
 #### Save the official taxonomy in json format
 my $webwork_htdocs = $ce->{webworkDirs}{htdocs};
 my $file = "$webwork_htdocs/DATA/tagging-taxonomy.json";
--- a/bin/OPLUtils.pm
+++ b/bin/OPLUtils.pm
@@ -17,7 +17,7 @@ use warnings;
 use File::Find::Rule;
 use File::Basename;
 use open qw/:std :utf8/;
-use JSON;
+use Mojo::JSON qw(encode_json);
 
 our @EXPORT = ();
 our @EXPORT_OK =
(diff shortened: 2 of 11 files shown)


somiaj (approved):
Didn't test, but looks straight forward and obvious change to remove an unneeded module.

I see we use `JSON` if a few other places, is it worth switching those to `Mojo::JSON` as well and then remove dependency on `JSON`?

drgrice1:
I decided to take this where @somiaj suggested.  I replace all `JSON` module usage with `Mojo::JSON` usage.

drgrice1:
And now this goes all the way.  In https://github.com/openwebwork/pg/pull/1226 PG is also switched to using `Mojo::JSON`.

Alex-Jordan (approved):

Alex-Jordan:
I tested this (too lightly perhaps) on my dev server. Then merged here. Then deployed to production server, and some things broke with graphTool exercises. I do believe I had run `npm ci` and restarted `webwork2`.

No time to investigate now, but I will later. I wanted to note that maybe there is an issue here.

drgrice1:
I am sorry, I forgot to mention in this pull request that this is paired with https://github.com/openwebwork/pg/pull/1226.  That would be why the graphTool exercises are now broken (among other things).

drgrice1:
I guess I should have waited with that part and made a separate pull request.

drgrice1:
For now you can change line 1366 of `conf/defaults.config` back to `[qw(JSON)]` and that will get PG back.  That is the only thing in this pull request (other than in `bin/check_modules.pl`) that PG depends on.

Alex-Jordan:
It's no problem, in production I just used git to undo the merge. We can take time to get pg#1226 in too.

Issue #2690, opened 2025-04-01
unusual characters in an uploaded file


Alex-Jordan:
On develop, a student used the Email Instructor button and attached a file with name `Screenshot 2025-04-01 at 3.22.24â€¯PM.png`. This led to them getting an error message about a wide character from `lib/WeBWorK/Upload.pm` line 117. (The image and message did make it to the instructor.)

somiaj:
This should have been fixed in #2693

PR #2693, opened 2025-04-03
Rewrite the WeBWorK::Upload module.

drgrice1:
This takes advantage of things available with Mojolicious.

This does not use `Data::UUID` anymore.  A comment stated that it was overkill.  It is and it was never a good idea to use that for this purpose.  This just uses the `Mojo::Asset::Memory::to_file` method which converts to a `Mojo::Asset::File` object and saves the file to a temporary file using `Mojo::File::tempfile` (which uses `File::Temp` under the hood).  That is guaranteed to give a safe filename that does not conflict with any existing files.  Setting the environment variable `MOJO_TMPDIR` locally to be the passed in directory (which is `$ce->{webworkDirs}{uploadCache}`) ensures that the file is saved in the usual webwork2 upload location (although there is no need for that, and we could drop that configuration variable and just use the system temporary directory for this).

The info file that is written is now UTF-8 encoded and decoded.  This fixes issue #2690.

The `dir` option which was previously required and the only "option" is now just a required argument.  It doesn't make sense to use an "option" hash things that are required.  Particularly if there is only one option and it is required.

Other than what is mentioned above the module behaves much the same.

### DIFF
--- a/lib/WeBWorK/Upload.pm
+++ b/lib/WeBWorK/Upload.pm
@@ -6,275 +6,202 @@ WeBWorK::Upload - store uploads securely across requests.
 
 =head1 SYNOPSIS
 
-Given C<$u>, an Mojo::Upload object
+Given C<$u>, a C<Mojo::Upload> object
 
- my $upload = WeBWorK::Upload->store($u,
- 	dir => $ce->{webworkDirs}->{DATA}
- );
- my $id = $upload->id;
- my $hash = $upload->hash;
+    my $upload  = WeBWorK::Upload->store($u, $ce->{webworkDirs}{uploadCache});
+    my $tmpFile = $upload->tmpFile;
+    my $hash    = $upload->hash;
 
 Later...
 
- my $upload = WeBWorK::Upload->retrieve($id, $hash,
- 	dir => $ce->{webworkDirs}->{uploadCache}
- );
- my $fh = $upload->fileHandle;
- my $path = $upload->filePath;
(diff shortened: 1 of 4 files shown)


lib/WeBWorK/Upload.pm
```diff
@@ -21,275 +21,202 @@ WeBWorK::Upload - store uploads securely across requests.
...
-An upload stored in the upload cache can be retrieved by supplying its ID and
-hash (accessible from the above C<id> and C<hash> methods, respectivly. The file
-can then be accessed by name or file handle, moved, and disposed of.
-
-=head2 CONSTRUCTOR
-
-=over
+An upload stored in the upload cache can be retrieved by supplying its temporary
+file name and hash (accessible from the above C<tmpFile> and C<hash> methods,
+respectively.  The file can then be accessed by name or file handle, moved, and
```
pstaabp, line 139: minor thing: close the ) after 'respectively' ?

pstaabp:
Thinking about testing this. There's clearly the email that came from the issue, but does this get called for a file upload to through the File Manager?  I notice the code in `FileManager.pm`, but unclear if that is enough to test.

lib/WeBWorK/Upload.pm
drgrice1: Thanks.  That is fixed.

drgrice1:
There are currently two places this is used, and you mentioned both of them.

pstaabp (approved):
Tested with uploading to the File Manager.  Didn't test with emails.

drgrice1:
I believe that the `Data::UUID` is still used elsewhere though.

somiaj (approved):

PR #2699, opened 2025-04-11
Fix random order page change issue.

drgrice1:
I bet you thought all of the random order gateway test issues were resolved.  Nope.

When a page change occurs in a random order test, the answers saved in the hidden inputs for the problems not on the current page are still not respecting the order.  As a result the answers get moved around to incorrect problems each time that a page change occurs.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -971,7 +971,7 @@ async sub pre_header_initialize ($c) {
 					create_ans_str_from_responses($c->{formFields}, $pg_result,
 						$pureProblem->flags =~ /:needs_grading/);
 			} else {
-				my $prefix         = sprintf('Q%04d_', $problemNumbers[$i]);
+				my $prefix         = sprintf('Q%04d_', $problemNumbers[ $probOrder[$i] ]);
 				my @fields         = sort grep {/^(?!previous).*$prefix/} (keys %{ $c->{formFields} });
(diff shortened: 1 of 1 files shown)


drgrice1:
To test this create a test, use all of the default test settings except change "Order Problems Randomly" to "Yes", and add at least 3 problems (more might make it easier to see the issue).  You can just use the default new problem for all problems.

Now start a test and enter answers for all problems (I entered the template id for the answers).  Then grade the test.  Now change the page a few times after doing that.  The answers stay in the right places at this point.  Now leave the test (go back to the test information page), and then enter the same version again.  Now change pages a few times again.  You might see the problem now.  Answer might disappear.   If not, then leave the test again, and re-enter it a third time.  Change pages again.  If you still don't see the problem.  Repeat.  Eventually you will.

It is easiest to test as an instructor that is shown the template id for the problem.  Note that if the template ids all happen to match with the random order (i.e., the random order happened to be the actual template order), then of course, the problem won't occur.  So as an instructor when testing you can verify that easily by looking at the template ids and start over.

drgrice1:
By the way if you look in the database at the `last_answer` column even before the test is first graded you can see that the last answers are going into the wrong problems.  You can use `SELECT problem_id,last_answer from courseID_problem_user where user_id='your_userId' and set_id='your_set_name';`.  The problem id and `Q00??` prefixes in the answers should match always, but they won't without this pull request.

Alex-Jordan (approved):

somiaj (approved):
I suspect we will have at least one more dragon hiding! :D

PR #2702, opened 2025-04-14
Convert the conf/database.conf.dist file to a Perl module.

drgrice1:
The database layout is now defined in the  Perl `WeBWorK::DB::Layout` package.  An instance of the `WeBWorK::DB` package is no longer instantiated by passing the layout.  Instead it is passed a course environment object.  An instance of the `WeBWorK::DB::Layout` package is constructed from that.

In addition, the `WeBWorK::DB::Driver`, `WeBWorK::DB::Driver::Null`, and `WeBWorK::DB::Driver::SQL.pm` modules have been deleted.  Instead there is the `WeBWorK::DB::Database` module that does everything those modules used to do.  It is now the only "driver" for all tables.  Although, that never should have been called a driver.  The driver is still optional and is `DBD::MariaDB` or `DBD::mysql`.  That is what is usually referred to as the driver, not the DBI connection handle which is what the `WeBWorK::DB::Database` module deals with, and what the previous "driver" modules did as well.  Furthermore, only one instance of this module is constructed when a `WeBWorK::DB` object is constructed, and that is passed to each table.  That is essentially the same as before in terms of DBI connection handles except that before there were multiple `WeBWorK::DB::Driver::SQL` instances that all shared the same DBI conneciton handle via the `connect_cached` call.

So the layout is simplified considerably.  The "driver", "source", "engine", and "character_set" are no longer options for a table, and there are no "%sqlParams" passed to each table. The only thing that is variable in the layout is the "$courseName" used for the tableOverride. The "source", "engine", and "character_set" are parameters of the `WeBWorK::DB::Database` object and retrieved from there as needed.

Since there is no choice of database layout there is no point in having that be part of the authentication module selection in the configuration files.  So the `$authen{user_module}` in the configuration files should be either a string naming a single authentication module or an array of strings.  For example,

```perl
$authen{user_module} = [
    'WeBWorK::Authen::LTIAdvantage',
    'WeBWorK::Authen::LTIAdvanced',
    'WeBWorK::Authen::Basic_TheLastOption'
];

```

The old format of `{ '*' => 'WeBWorK::Authen::LTIAdvantage' }` (where `*` meant use this authentication for all database layouts) is still allowed for now, but eventually support for that will be dropped.

Several related unused files were deleted.  Those are

* `bin/check_database_charsets.pl`
   (This should have never been in the repository.)
* `bin/wwdb`
   (This has been broken for a long time.)
* `lib/WeBWorK/DB/Driver/Null.pm`
  (This was mentioned above, but I don't think this ever was used or even made sense to be used.)
* `lib/WeBWorK/Utils/CourseManagement/sql_moodle.pm`
  (This is not used, I doubt this has worked for a long time, and is no longer supported in any case.)
* `lib/WeBWorK/Utils/CourseManagement/sql_single.pm`
  (Also not used and hasn't been working for a while now.)
* `lib/WeBWorK/Utils/DBImportExport.pm`
  (This hasn't worked for a while now and code that called this in `lib/WeBWorK/DB.pm` was removed.)
* `lib/WeBWorK/DB/Schema/NewSQL/VersionedMerge.pm`
   (This is unused.  At some point the `lib/WeBWorK/DB/Schema/NewSQL/Merge.pm` module was updated to handle its functionality directly.)

This is essentially a rework of #2338 except it does not move the database settings from site.conf, and it does not mess with the external program settings.


pstaabp (approved):
Tested by running through most of the pages.  Looks good. 

Seems like we should get this in soon so we can test by using it.

somiaj (approved):

PR #2704, opened 2025-04-16
Database code clean up.

drgrice1:
The main thing that this does is implement the `WeBWorK::DB::Ex` exceptions as the comments in `lib/WeBWorK/DB.pm` (that have been there since 2008) say they were intended to be used.

For `WeBWorK::DB::Ex::DependencyNotFound` and `WeBWorK::DB::Ex::RecordNotFound` the error message for the exception is the same as before. So for those the only difference is that they are now Exception::Class objects (which contain more information than just the message).

For `WeBWorK::DB::Ex::RecordExists` and `WeBWorK::DB::Ex::TableMissing` the error message for the exception is the error message that DBI gives. The previous messages for `WeBWorK::DB::Ex::RecordExists` like "addUserSet: user set exists (perhaps you meant to use putUserSet?)" no longer replace the DBI exception messages. So as the comments from 2008 stated don't check for the string "user set exists" in the exception message anymore.  Instead check if `WeBWorK::DB::Ex::RecordExists->caught` is true.

In addition the code for auto creation of password records in `getPasswords` and the auto creation of permission records in `getPermissionLevels` has been deleted simply because that code doesn't even work.  It hasn't worked for a long time if it ever did. In both methods the `gets` method is called to retrieve the requested records.  The `gets` method does not even include non-existent records in the list it returns.  So the auto creation loops were only looping over records that were guaranteed to exist.  So that code was purely an inefficient loop that did nothing.

Next, the `addMultipleUserSets` and `addUserMultipleProblems` methods of `lib/WeBWorK/DB.pm` were removed.  Instead what is needed from those methods is used directly in the `lib/WeBWorK/Utils/Instructor.pm` module where those methods were called.  Those methods added a high level of inneficiency to the process of assigning sets and problems to multiple users.  For example the methods in `lib/WeBWorK/Utils/Instructor.pm` check that the sets don't already exist and if so it skips them. Then the `addMultipleUserSets` method again checks if the sets exist. Note that the previous `lib/WeBWorK/DB.pm` methods did check for the existence of the users, and this no longer does.  That check slows things down considerably, and in all cases those users come from the database to begin with.  So it is not really a necessary check.  This rewrite speeds up the assignment of multiple sets and problems considerably.

Finally, the methods in `lib/WeBWorK/Utils/Instructor.pm` no longer return unused lists of failures.  In all of the uses of those methods in the webwork code, there was only one place where those return values were ever used, and even in that case it was not in an essential way. Furthermore, that one usage was in `lib/WeBWorK/SetActions.pm` (for the WebworkWebservice) which is not actually used by webwork2 anyway (or probably even by anyone).  Instead of catching and rethrowing exceptions, the exceptions are just let through.  The exceptions that were ignored before (but put into a return value that was later ignored) are still ignored.  The POD in this file is cleaned up as well.

### DIFF
--- a/lib/WeBWorK/DB.pm
+++ b/lib/WeBWorK/DB.pm
@@ -92,39 +92,24 @@ use WeBWorK::DB::Utils qw/make_vsetID grok_vsetID grok_setID_from_vsetID_sql
 use WeBWorK::Debug;
 use WeBWorK::Utils qw(runtime_use);
 
-=for comment
-
-These exceptions will replace the ones in WeBWorK::DB::Schema and will be
-allowed to propagate out to calling code. The following callers will have to be
-changed to catch these exceptions instead of doing string matching:
-
-lib/WebworkSOAP.pm:     if ($@ =~ m/user set exists/) {
-lib/WeBWorK/ContentGenerator/Instructor.pm:             if ($@ =~ m/user set exists/) {
-lib/WeBWorK/ContentGenerator/Instructor.pm:     if ( $@ =~ m/user set exists/ ) {
-lib/WeBWorK/ContentGenerator/Instructor.pm:             if ($@ =~ m/user problem exists/) {
-lib/WeBWorK/ContentGenerator/Instructor.pm:             if ($@ =~ m/user problem exists/) {
-lib/WeBWorK/ContentGenerator/Instructor.pm:                     next if $@ =~ m/user set exists/;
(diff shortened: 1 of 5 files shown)


drgrice1:
Note that this has no conflicts with #2702.

pstaabp (approved):
Tested by performing a lot of standard actions that use the DB.  Like the other DB change, I would argue to get this in soon for testing by just day-to-day using it.

somiaj (approved):

Issue #2706, opened 2025-04-17
Can't use Scroll of Resurrection


Alex-Jordan:
I'm on develop, and there is an issue with achievements. I have 10 Scrolls of Resurrection.

<img width="366" alt="Image" src="https://github.com/user-attachments/assets/ea395c50-016c-4802-93b2-ab701cabd8be" />

----

There is a set that closed on 4/12. 

<img width="403" alt="Image" src="https://github.com/user-attachments/assets/ab1340b8-0d45-4dc5-995e-aa8a734bbcfb" />

----

When I visit that set, the interface won't allow me to use a Scroll of Resurrection on it.

<img width="366" alt="Image" src="https://github.com/user-attachments/assets/273ea248-cf93-4915-93d7-4659ad33a6bd" />

drgrice1:
Is the due date of the set within 24 hours ago?  That award can only be used if that is the case.  I tested with an assignment that was due yesterday, and I see the "Scroll of Resurrection" available for use.

Alex-Jordan:
If that's how the code is right now, that is wrong. That limitation is only supposed to apply to the two extension achievements (Tunic of Extension, Robe of Longevity). The Scroll of Resurrection has always been something you can apply to any dead assignment, even one from weeks ago.

drgrice1:
I see that this is a change from how it worked before.  For previous versions of webwork... You just posted what I am saying.  I think you are correct.  This is from @somiaj's rework.  That can be easily changed back.

Alex-Jordan:
Yeah it looks like the 24-hour limitation slipped in with #2664. We should undo that. The check for can_use just needs to be that the current date is past the close date.

drgrice1:
For WeBWorK 2.19 the check was
```
			if $sets->[$i]->assignment_type eq 'default'
			&& (after($sets->[$i]->due_date)
				|| ($sets->[$i]->reduced_scoring_date && after($sets->[$i]->reduced_scoring_date)));
```

Also the `use_item` method needs to be changed.  For 2.19 it used `time + 86400`.  Now it uses the current due date for the set plus 86400 which will not work if it is more than one day after the due date.

pstaabp:
fixed in #2707

PR #2707, opened 2025-04-18
Restore previous behavior of the "Scroll of Resurrection" achievement item.

drgrice1:
With previous versions of WeBWorK the scroll of resurrection could be used more than a day after the due date (there was no restriction). This just makes that the behavior again.  It seems this was changed in #2664.

This fixes issue #2706.

### DIFF
--- a/lib/WeBWorK/AchievementItems/ResurrectGW.pm
+++ b/lib/WeBWorK/AchievementItems/ResurrectGW.pm
@@ -19,7 +19,7 @@ use Mojo::Base 'WeBWorK::AchievementItems', -signatures;
 # Item to extend the due date on a gateway
 
 use WeBWorK::Utils           qw(x);
-use WeBWorK::Utils::DateTime qw(between);
+use WeBWorK::Utils::DateTime qw(after);
 
 use constant ONE_DAY => 86400;
 
@@ -35,7 +35,8 @@ sub new ($class) {
 }
 
 sub can_use($self, $set, $records) {
-	return $set->assignment_type =~ /gateway/ && between($set->due_date, $set->due_date + ONE_DAY);
+	return $set->assignment_type =~ /gateway/
+		&& (after($set->due_date) || ($set->reduced_scoring_date && after($set->reduced_scoring_date)));
 	# TODO: Check if a new version can be created, and only allow using this reward in that case.
 }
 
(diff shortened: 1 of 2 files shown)


somiaj (approved):

Alex-Jordan:
It seems that the message `Close date of this assignment extended by 24 hours to [_1].` is not saying quite the right thing either.

Thanks for this. If it wasn't clear in the other thread, what I meant was that the way that this item was behaving was "wrong", not that your analysis was wrong.

drgrice1:
> It seems that the message `Close date of this assignment extended by 24 hours to [_1].` is not saying quite the right thing either.

What should that say then?

> Thanks for this. If it wasn't clear in the other thread, what I meant was that the way that this item was behaving was "wrong", not that your analysis was wrong.

Yeah, I got that!

drgrice1:
Should it say just `Close date of this assignment extended to [_1].`

Alex-Jordan:
Locally I just changed it to `This assignment is now open again with a due date of [_1]. Problems have been rerandomized.` @somiaj might have input.

Alex-Jordan:
Er, I guess I would prefer `close date` to `due date`. Maybe this:

`This assignment has been reopened and will now close on [_1]. Problems have been rerandomized.`

drgrice1:
I made that the message.

somiaj:
Looks good to me, seems I miss understood the item when I updated it. Thanks for catching this.

Alex-Jordan:
I think `ResurrectGW.pm` has the same issue. Although for that one, I am confused if the item applies to the test template or to individual versions. Something about that one needs to change, but what will we change it to? I think we can feel free to just do something that seems right, because (a) that item is not in the default collection of level achievement items and (b) IIRC this was doing something obviously bad before, like extending a test *version* by 24 hours.

drgrice1:
I also reverted some of the behavior of the "Necromancer Charm", i.e., `ResurrectGW.pm`.  Note that this does not change any dates for test versions.  It only changes the dates for the template set.  So if the student still has versions left to take then those can be taken, but previous version will remain closed.  The information states that new versions of the test will not be made available.

drgrice1:
Perltidy is doing some odd things here.

For the `ResurrectHW.pm` file the `);` after the last `return` statement in the file was previously on a line at the end by itself, but perltidy now moved it up onto the previous line at the end.

However, for the `ResurrectGW.pm` file for the `);` after the last `return` statement it did the exact opposite.  It was on the line before at the end, and now it is moved onto a line by itself.

It seems to look at the two lines before it and if the first of those two lines is longer than the second, then it puts the parentheses on the second line, but if the first line is shorter, then it puts the parentheses on the next line by itself.

I wish it would do what prettier does for this and always put it on the next line by itself.

Alex-Jordan (approved):

PR #2708, opened 2025-04-20
changes to email template for feedback

Alex-Jordan:
This changes several things about the HTML email feedback template. If it is desired, some parallel changes could also be made to the text email feedback template.

* The stylesheet in the head (and the head) is removed, and all styling is now inline. This is for two reasons. One is that some email clients that do understand HTML in email do not understand a style element in the head. Also, I was finding that I would receive an email and the HTML was styled well, but then if I forwarded that email or replied to it, the recipient no longer could see the styling. The head (or the style element within the head) was not included when the message was passed along. But now with inline styling, the styling persists.
* The styling is changed. See screenshot below.
* The message at the top no longer includes a link to the WeBWorK site. It only includes a link to the page where the user initiated the feedback.
* If the feedback is coming from a problem or from a set's problem list page, the link text describes that origin. Otherwise it still says "this page".
* If the feedback is from a test, there was an error where the "Attempts per version" row of one of the tables was showing the wrong thing (it was showing `$set->assignment_type`).
* The order of the tables was problem processor, user, problem set. I changed it to what I think puts more useful things higher up: problem, set, problem processor, user.
* The order of rows within the problem table is changed so that the "last answer" row comes last. This is because it is built in a special way and it just simplified the code to move it last.
* The rows that indicate Show Old Answers, Show Correct Answers, Show Hints, Show Solutions previously print out 0 or 1, and now print out "no" or "yes".

I may be missing additional things that I changed. This is the file that is in use on PCC's production server and I am unsure if there is some additional change I made that is too small for me to notice as I look over this diff.

Perhaps now (or soon after this PR) would be a good time to consider cutting some things from the tables. One I am wondering about is the "Comment" about a user. I wonder if an instructor might put something there that they would not want or expect a student to ever see.

Screenshot 2025-04-20 at 12 22 33 AM


templates/ContentGenerator/Feedback/feedback_email.html.ep
```diff
@@ -1,183 +1,183 @@
...
+		% } else {
+			<p>
+				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
+				<%= $ce->{institutionName} %> (sent from <%= link_to 'this page' => $emailableURL %>).
+			</p>
+		% }
 		% if ($feedback) {
-			<p><%= $user->full_name %> (<%= $user->user_id %>) wrote:</p>
-			<div class="user-message mb-1">
+			<div style="border: 1px solid black; padding: 1rem; margin-bottom: 1rem">
```
drgrice1, line 23: I take it you didn't like the `border-radius: 0.375rem;` style here?  I liked that, and I felt that it offset the student message in a nicer way.  Also, the same style is applied to the student message that is displayed for the student sending the email after it is sent in `templates/ContentGenerator/Feedback.html.ep` (on line 16 via the `rounded` Bootstrap class).

templates/ContentGenerator/Feedback/feedback_email.html.ep
```diff
@@ -1,183 +1,183 @@
...
+			</p>
+		% } elsif ($set) {
+				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
+				<%= $ce->{institutionName} %> (sent from <%= link_to format_set_name_display($set->set_id) %>)
+
+		% } else {
+			<p>
+				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
+				<%= $ce->{institutionName} %> (sent from <%= link_to 'this page' => $emailableURL %>).
+			</p>
```
drgrice1, line 20: The `$set` case is missing the URL for the link.  So the email shows a link that does not work.  In addition, that case is not wrapped in a `<p>` tag like the other cases.  It would be better to move the `<p>` tag outside of the `if ... then ... else` statement so that this sort of mistake doesn't happen.  So this should be
```suggestion
		<p>
			% if ($problem) {
				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
				<%= $ce->{institutionName} %> (sent from
				<%= link_to format_set_name_display($set->set_id) . ', #' . $problem->problem_id => $emailableURL %>).
			% } elsif ($set) {
				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
				<%= $ce->{institutionName} %>
			   	(sent from <%= link_to format_set_name_display($set->set_id) => $emailableURL %>)
			% } else {
				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
				<%= $ce->{institutionName} %> (sent from <%= link_to 'this page' => $emailableURL %>).
			% }
		</p>
```

drgrice1 (commented):
I think this generally looks good.  There is one issue that needs to be fixed.

I am not entirely sold on the non-bordered tables.  I kind of like the borders particularly for offsetting the last answer section.

drgrice1 (commented):
Actually, there is another thing that needs to be fixed.  You changed the content of the HTML for the email, so the text also needs to be change accordingly.

templates/ContentGenerator/Feedback/feedback_email.html.ep
Alex-Jordan: It's true, I did not like it much. For me, it's perhaps partly because of ambient Gmail styling when I open the message in Gmail. Just now I played with an old email (where the radius was still there) and changed the border color from black to lightgray, and it is less jarring for me. Do you have feelings about changing that color?

This discussion makes me wonder what happens with people reading these HTML emails using an email client that is using dark mode.

Alex-Jordan:
I see what you mean about the last answer section.

I prefer this overall without the clutter of the table borders everywhere. Abstractly, I'm not sure what we have here is really a "table". To me it feels more like a "description list", with titles and then info for each title. But we don't have access to use the `dl` element because that HTML element is not supported by email clients. Since it feels like a `dl` to me, that is probably the source of my aversion to the table borders.

If you really want the borders in general, I can add them back. But to be clear, a border for the whole table? For each cell? For each row (and no vertical lines in the middle)? What about just doing something to make the "Last answer" section more clearly stand out?

drgrice1:
Doing something for the last answer section would be sufficient.

Alex-Jordan:
This may need more cosmetic changes, but is ready to be looked at again.

- corrected the oversight with the GitHub suggested edit
- the text template has been updated to agree (as much as it can) with the html template
- "last answer" changed to "last submission"
- the alignment of "last submission" agrees with the top row of the last submission table
- there is a horizontal rule (top border for one row) separating the last submission row from the rows above
- restored the border-radius on the student message, but also softened the color. I am fine with further suggestions or requests to style that differently.

templates/ContentGenerator/Feedback/feedback_email.txt.ep
```diff
@@ -1,9 +1,7 @@
 % use WeBWorK::Utils qw(decodeAnswers);
 %
 Message from <%== $user->full_name %> (<%== $user->user_id %>) via WeBWorK at
-<%== url_for('root')->to_abs %>
-
-To visit the page from which <%== $user->first_name %> sent feedback, go to:
+<%== $ce->{institutionName} %> (sent from
```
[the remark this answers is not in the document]
Alex-Jordan: OK, I made this change.

drgrice1 (approved):
This looks good.

I put a pull request in to this branch with a little code formatting.

As to the border radius, if you don't like it, you don't have to leave it.  I don't feel strongly about that.

Alex-Jordan:
About the border radius: I'm not planning to change it unless others weigh in. So everyone should consider this up for review.

pstaabp (approved):

PR #2709, opened 2025-04-20
Improve loading time for the achievements leaderboard.

drgrice1:
Currently the user achievements are obtained from the database in the user loop.  This takes a long time when there are lots of users. So instead, this gets all user achievements for all users before the loop and references them by user id and then achievement id.

Testing this on a course for 5000 users shows this gives a significant speed up on load time for the page.  With the develop branch it takes around 25 seconds, and with this branch it takes around 3 seconds.

Note that there were also two redundant queries (one that listed all achievements and then one that fetched all achievements on lines 49 and 50 in the develop branch) for which the data from those queries was not even used.  Those were removed.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/AchievementsLeaderboard.pm
+++ b/lib/WeBWorK/ContentGenerator/AchievementsLeaderboard.pm
@@ -46,27 +46,30 @@ sub initialize ($c) {
 	my %globalUserAchievements =
 		map { $_->user_id => $_ } $db->getGlobalUserAchievementsWhere({ user_id => { not_like => 'set_id:%' } });
 
-	my @allBadgeIDs = $db->listAchievements;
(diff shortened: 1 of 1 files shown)


drgrice1:
To help with testing this here is  a class list with 5000 users.
[longDemoCourse.lst.txt](https://github.com/user-attachments/files/19825750/longDemoCourse.lst.txt)

Import that into a course (that will take a while) and assign all achievements to those users (this will take a long time -- about 10 minutes on my machine), and then open the achievements leaderboard page.  With the develop branch you will wait a while.  With this pull request it will be much faster.

Note that if you are using `hypnotoad` then you may need to increase the `inactivity_timeout` and `heartbeat_timeout` in `conf/webwork2.mojolicious.yml` for the achievement assignments to complete.  Although I think that the assignments will complete even if the request times out.  It will continue on the the server even though the request times out (not sure o this though).

Alex-Jordan (approved):

PR #2710, opened 2025-04-20
Speed up deleting of global user achievements.

drgrice1:
Instead of listing user achievements and then deleting them one by one in a loop, just delete them with one query using a `where` statement.

This improved the deletion of 5000 users all of which had the default set of achievements assigned to them from taking a minute and a half to less than 20 seconds.

### DIFF
--- a/lib/WeBWorK/DB.pm
+++ b/lib/WeBWorK/DB.pm
@@ -1467,17 +1467,8 @@ sub deleteGlobalUserAchievement {
 	# userAchievementID can be undefined if being called from this package
 	my $U = caller eq __PACKAGE__ ? "!" : "";
 	my ($self, $userID) = shift->checkArgs(\@_, "user_id$U");
-
-	my @achievements = $self->listUserAchievements($userID);
-	foreach my $achievement (@achievements) {
-		$self->deleteUserAchievement($userID, $achievement);
-	}
-
-	if ($self->{global_user_achievement}) {
-		return $self->{global_user_achievement}->delete($userID);
-	} else {
-		return 0;
-	}
+	$self->{achievement_user}->delete_where({ user_id => $userID });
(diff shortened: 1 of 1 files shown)


drgrice1:
You can test this using the class list provided in #2709 and in combination with testing that pull request.  Delete the users from imported and for which the achievements are assigned for testing that pull request.

Alex-Jordan:
The new line 1471 assumes `$self->{global_user_achievement}` will always be something that the `delete` method applies to. Is it for sure that there is no scenario where `$self->{global_user_achievement}` is simply undefined? (I am not up to speed on how the `global_user_achievement` property is initialized.)

drgrice1:
All of the hash keys of `$self` are references to `WeBWorK::DB::Schema` objects.  They are defined in the `WeBWorK::DB::new` method earlier in this file (or really the `WeBWorK::DB::init_table` method call by `new`), and correspond to the hash keys of the database layout.  So if you are using an alternate database layout that does not have a `global_user_achievement` table defined, then that would be undefined.  Then all of the methods from lines 1406 through 1465 that don't check to see `$self->{global_user_achievement}` is undefined will fail with an exception.  However, this method will succeed and always return 0 in that case.  So attempting to count, list, check exists, get, add, or put a global user achievement all will throw an exception if you don't have `global_user_achievement` defined in your database layout, but deleting a global user achievement is okay.

Basically this method is entirely inconsistent with every other method in this file.  If this were consistent, then every method in this file should perform this same check to see that the hash key for the corresponding schema is defined before doing anything.

Since we do not support alternate database layouts anymore, and furthermore, since there won't even be the option to have an alternate database layout soon (once #2702 is merged), this check is just not needed.

pstaabp (approved):
See the speed up.

PR #2712, opened 2025-04-21
Speed up achievement scoring.

drgrice1:
First get all of the necessary achievement data from the database, then process it and print it to the file.

Comparing scoring all course achievements for all users in a course with 5000 users shows a considerable speed improvement. It takes more than three minutes for this with the develop branch, and less than 3 seconds with this pull request.

Obviously the generated scoring files are identical.


pstaabp (approved):
more improved speed up.

drgrice1:
For testing here is the class list: https://github.com/user-attachments/files/19825750/longDemoCourse.lst.txt

drgrice1:
For testing here is that class list: https://github.com/user-attachments/files/19825750/longDemoCourse.lst.txt

Alex-Jordan:
Before this, scoring just one achievement for all 5000 users took 15 seconds. Now it's instant.

PR #2713, opened 2025-04-21
Speed up class list importing.

drgrice1:
Use a batch insertion and update for users, permissions, and passwords when importing a class list.

This improves the speed of imporing large class lists quite a bit.  For example, importing a classlist containing 5000 users improves from around 22 seconds to around 7 seconds in my testing.


pstaabp:
I tested the reloading of the large classlist you provided and skipped 5000 users.  The dialog box is many pages worth.    I was thinking that shortening the list if over 100 skipped user would be nice, but then again this is an extreme example and may not be worth the work.

drgrice1:

dlglin (approved):

PR #2714, opened 2025-04-21
Speed up loading of the "Users Assigned to Set" page.

drgrice1:
Load all user sets for the given set and all users before rendering the template instead of getting one set at a time in the for loop inside the template.  This speeds up loading of this page a lot with a large number of users.  With 5000 users it decreases the load time from 3.4 seconds to 0.14 seconds.


pstaabp:
I only noticed a few percent speedup with this.   I used your 5000 person course and then loaded the Demo set (from model course).  On develop, it took about 73 seconds and with this was about 67 seconds.  

The set was not assigned to the users.   I was just navigating to the "Users Assigned to Set Demo" page.  Is this the wrong place?

drgrice1:
That is the correct page.  I just tested doing the same thing.  I am still seeing a rather noticeable speed up.  I takes about 0.14 seconds for the app to render the page with this pull request, and 3.6 seconds with the develop branch.  This is after importing the Demo set, but not assigning it to all users.

Since your timings are so large to begin with, I suspect that the main delay you are experiencing is network latency.  You are seeing a 6 second decrease in time which is about right and is the improvement from this pull request.  The rest is delay in the network.

My tests are done with morbo on the same computer.  So there is no network latency.  That gives a good measure of direct efficiency of the app in rendering the page.  Furthermore, monitoring cpu usage you will see that the load is considerably less with this pull request.

pstaabp:
I'm on my laptop--no network at all.  

It seems to be when the set is assigned to users.   Without assignment, it was about a second of load.  (under develop, I got over a minute). 

When the set was assigned to all users, it was more than 1 minute on this PR.

drgrice1:
I am seeing the same timings that I mentioned before when the set is assigned to all users.  More than 4 seconds with develop, and a fraction of a section with this pull request.

This is just for loading the page.  Not assigning or unassigning the sets.

drgrice1:

pstaabp (approved):
Safari was terrible at rendering the page when the assignment was assigned, but chrome rendered it quite quickly.

dlglin (approved):

PR #2715, opened 2025-04-22
Speed up the achievement users page.

drgrice1:
The usual thing (by now with this series of pull request) is done.  That is removing database access from loops and reducing to single queries as much as possible.

In addition instead of using the Mojolicious tag helpers to render the "earned" checkbox and "counter" text field, direct html is used. I discovered that this renders much faster.  With the usual 5000 user test the rendering of the template alone takes 5 seconds after changing a checkbox or counter value and saving, while with the direct html it takes something like 0.2 seconds.  I believe that it has something to do with their code to set the values of the inputs that is causing the slow down.  I am not sure on that, but I know that the direct html is much faster.

That database changes are still the biggest part of the speed improvement here in any case.  This page was rendering quite slowly even on initial load before, and that part was fast with the tag helpers. After making the database changes things sped up considerably, but then timing parts of the code when saving revealed the tag helper issue.


drgrice1:
I forgot to mention that the "counter" text input was also switched to a number input with a minimum value of 0 and the default step size of 1.

pstaabp (approved):
This improves the speed.

drgrice1:

PR #2718, opened 2025-04-24
Remove the filtering of large courses on the course admin "Manage OTP Secrets" page and fix security vulnerability.

drgrice1:
This page loads quickly regardless of if these courses are filtered out. Furthermore the menus are generally responsive even when these large courses are included.  The only case where something is a bit slow is in the case that on the "Copy Single Secret" tab the selected "Source Course ID" is one of these large courses.  In that case if you click on the "Source User ID" it takes some time for the dropdown menu to appear and even then it has to be an astronomically large course for that to be slow (more than 20,000 users).  However, even if the large courses are included as long as a large course is not selected the menus are quick.

It is interesting that the multiple select elements on the "Copy Multiple Secrets" tab are still fast even with astronomically large courses.  It is just the single select elements on the "Copy Single Secret" tab that experience a noticeable slowdown.

In any case, the point is that there is no reason to filter these courses.  Doing so just adds unnecessary steps for the user.

@somiaj: This is probably not the direction you though this would go!

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -2403,23 +2403,14 @@ sub do_save_lti_course_map ($c) {
 
 # Form to copy or reset OTP secrets.
 sub manage_otp_secrets_form ($c) {
-	my $courses          = {};
-	my $dbs              = {};
-	my $skipped_courses  = [];
-	my $show_all_courses = $c->param('show_all_courses') || 0;
+	my $courses = {};
+	my $dbs     = {};
 
 	# Create course data first, since it is used in all cases and initializes course db references.
 	for my $courseID (listCourses($c->ce)) {
 		my $ce = WeBWorK::CourseEnvironment->new({ courseName => $courseID });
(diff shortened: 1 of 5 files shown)


drgrice1:
Note that I haven't tested copying a large number of OTP secrets.  Looking at the code I can see that that is going to need work.  I expect that this will be exceedingly slow in fact.  I see that there not only is heavy database access in for loops, but there is even construction of database objects inside user loops.  Unfortunately, getting a large number of users with OTP secrets to work with and test things is a problem in itself.

drgrice1:
I just noticed that there is a security vulnerability on this page.  The page should prevent the current user in the admin course from modifying there own OTP secret in the admin course in any way.  They should not be able to reset their own OTP secret and should not be able to overwrite their OTP secret with one for any user in another course by copying from that course.  It is probably okay to allow them to copy their OTP secret to another course I suppose.  Although that is certainly questionable.

drgrice1:
This now also prevents a user in the admin course from modifying their own OTP secret.  So this is now a security vulnerability pull request as well.

somiaj (approved):
Confirmed I cannot overwrite my admin OTP secret, thanks for catching that.

PR #2723, opened 2025-05-05
Make WebworkWebservice honor showEvaluatedAnswers

somiaj:
When rendering a problem via the `render_rpc` endpoint, honor the course environment `showEvaluatedAnswers` setting, instead of defaulting to `1`.


drgrice1 (approved):
Looks good to me.

PR #2711, opened 2025-04-20
Speed up assignments of achievements.

drgrice1:
If you have payed attention (or tested #2709 and #2710), then you know that this takes a long time if there are a large number of users.  Way to long to be at all reasonable.

This uses techniques like those used for assigning multiple sets to users, i.e., the `WeBWorK::DB::Schema::NewSQL::Std::insert_records` and `WeBWorK::DB::Schema::NewSQL::Std::update_records` methods, to speed up assignment of achievements to users.

This brings the assignment time for 5000 users down from more than 8 minutes, to less than 30 seconds.

PR #2728, opened 2025-05-14
Fix file uploads.

drgrice1:
The `MOJO_TMPDIR` attempt was not working.  The problem is that the `Mojo::Upload` object is already created before the `WeBWorK::Upload::store` method is called, and so the temporary file location is already determined and the temporrary file saved there.  So when the upload is later retrieved it is not where it is expected to be.

This does not happen with rather small files because a Mojo::Upload::Memory object is used instead of a Mojo::Upload::File object, and in that case the file is not saved to a temporary file to begin with.

Note: I just noticed that uploading tar files and automatically extracting them is broken and was broken in 2.19 as well.  That is another issue that needs to be fixed.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -18,7 +18,7 @@ use Archive::Zip::SimpleZip qw($SimpleZipError);
 
 use WeBWorK::Utils                   qw(sortByName min);
 use WeBWorK::Utils::CourseManagement qw(archiveCourse);
-use WeBWorK::Utils::Files            qw(readFile);
+use WeBWorK::Utils::Files            qw(readFile path_is_subdir);
 use WeBWorK::Upload;
 
 use constant HOME => 'templates';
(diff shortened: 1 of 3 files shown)


somiaj (approved):
Confirmed this fixed the issue I mentioned in stack.

For others testing, to reproduce the issue, try to upload a course `tar.gz` archive downloaded from the admin course (noticed this trying to copy a course from one server to another server via downloading/uploading a course archive). Without this branch I get an error, with this branch the error is fixed.

drgrice1:
@somiaj:  Note that this pull request actually has nothing to do with tar files, and was not created to address the issue you mentioned in Slack.  This pull request is about file size, and was created to address the issue that @Alex-Jordan mentioned in https://github.com/openwebwork/webwork2/pull/2711#issuecomment-2880899772.  I suspect that in your testing the tar files you were uploading were large enough to hit this issue, and the other files you tested just happened to be smaller.

The tar issue that I am seeing still occurs with this pull request.  If you attempt to extract a tar file and the tar file contains files inside a directory, and not the directory itself (which happens to be the structure of course archives -- at least those created with the WeBWorK-2.20 branch), then the extraction fails. A zip file with the same structure should cause the same problem.

drgrice1:
Two more changes were added to this pull request.

First, when a zip or tar file is extracted, instead of using the `Mojo::File::realpath` method, the `path_is_subdir` method is used.  The `realpath` method throws an exception if a path component does not exist on the disk, which causes extraction of tar or zip files to fail if they contain a files inside a directory, but not the directory itself.

Second, when a course archive is created, make sure that the course directory itself is actually included in the archive.  So it won't fail without the first change!

Simple tar and zip files are attached that have the troublesome structure for testing this.  If you attempt to extract either of them with WeBWorK 2.19 or 2.20, they will fail.  They will succeed with this pull request.
[test.tar.gz](https://github.com/user-attachments/files/20214179/test.tar.gz)
[test.zip](https://github.com/user-attachments/files/20214182/test.zip)

PR #2729, opened 2025-05-20
Speed up the assignment of achievments when importing achievments.

drgrice1:
Importing the `default_achievements.axp` file into a course with no achievements and 5000 users and assigning to all users took about 8 minutes with the WeBWorK-2.20 branch.  With this pull request the time decreased to about 20 seconds.


pstaabp (approved):
I honestly ran the assignment on WW-2.20, but didn't notice how long it took--It wasn't done within 5 minutes.  This PR speeds that up to about 15 seconds.

PR #2730, opened 2025-05-21
move helper tool buttons to above problem list

Alex-Jordan:
This moves the Email Instructor, Use Achievement Reward, and Download Hardcopy buttons to above the list of problems for a set, making it more likely students are aware of these buttons.

Screenshot 2025-05-20 at 9 27 04 PM

For a test, the Email Instructor and Use Achievement Reward buttons are likewise moved to the top, but no Download Hardcopy button will appear at the top. Instead, the current behavior is retained (for a student, there is no such button anywhere except row by row for one version at a time, and for higher level users there are checkboxes and a button at the bottom). Here is what it looks like for higher level users:

Screenshot 2025-05-20 at 9 29 39 PM


somiaj:
Overall looks good, my initial thought is put the 'Start New Test' button above the Email/Use Achievement buttons (swap their order).

templates/ContentGenerator/ProblemSet/auxiliary_tools.html.ep
```diff
@@ -0,0 +1,78 @@
+<div class="col-12 mb-2">
```
drgrice1, line 1: This should be

```suggestion
<div class="submit-buttons-container mb-3">
```

Any `div` with any of the `col` class variants should be a direct descendant of a `div` with the `row` class, and this is not.  The `submit-buttons-container` class is designed for exactly this setup with buttons.  It is also used by the submit buttons at the bottom of the problem in an assignment.

The `mb-3` is what all of the `div` elements above and below are using, and margins should be kept consistent.

drgrice1 (commented):
The `templates/ContentGenerator/Base/feedback_macro_email.html.ep` and `templates/ContentGenerator/Base/feedback_macro_form.html.ep` files need a change.  The `<p>` tags in those files wrapping the button need to be removed.  Those are on lines 9 and 12 of the first file, and lines 23 and 26 of the second.  Those `<p>` tags should not be there in the setting that this pull request creates, and probably shouldn't have been there before.  Note that to test the second file set `$courseURLs{feedbackFormURL} = "https://www.google.com";` in `localOverrides.conf` (or use whatever URL you want).

Also, in `lib/WeBWorK/ContentGenerator.pm` line 976 should be changed from
`$c->ce->{courseURLs}{feedbackURL});` to `$c->ce->{courseURLs}{feedbackURL}, class => 'btn btn-primary');`.  This would have looked better before anyway, but looks particularly bad with the changes in this pull request.  Set `$courseURLs{feedbackURL} = "https://www.google.com";` in `localOverrides.conf` to test this.

Note that without these changes things are not working so well on narrow screens.  Here is a screenshot: 
condensed-buttons
Admittedly that is for an extremely narrow screen.  Even narrower than most mobile phone screens would have.

Alex-Jordan:
> put the 'Start New Test' button above the Email/Use Achievement buttons

It could probably be done, but right now the "Start New Test" button is built into the template that includes the version list table below it. I haven't looked into how to decouple those.

Alex-Jordan:
@drgrice1 I made all the changes you mentioned above and force pushed.

Alex-Jordan:
The more I think about it, I lean toward these three buttons logically belonging *before* the "Start New Test" button. If you use the "Start New Test" button on this page, that is the last thing you will be clicking on this page before going off to do the test. Conversely:

- you might get an Email off to the instructor before starting the test
- you might apply an achievement reward item before starting a new test, assuming we ever have achievement reward items that apply to test templates in the first place. What we have now would only be used on a version, in which case your next click is probably going to be going that particular version, not clicking "Start New Test".
- you would only use the Hardcopy button on a version, and I'm not sure it matters how that one is ordered with respect to "Start New Test"

drgrice1 (approved):

PR #2733, opened 2025-05-27
Add a search box to the Sample Problems page and a script to build a search file

pstaabp:
This PR has two related features
1. A script in `bin/dev_scripts` to parse all of the sample problems and the POD to build up a "database" (JSON file) of non-common words for a given problem/POD page.  The JSON file contains information to be searched on within the browser using the package MiniSearch. 
2. Add search boxes to the sample-problems/POD pages to return pages that match the search criteria.  The searching is done using MiniSearch. 

There is a file `htdocs/DATA/search.json` that is provided that will be used while searching.  This is not complete but currently should give some sense about how this works.   The script will rebuild this file as needed. 

Note: a subsequent PR will be posted to update the POD in the macros to give the search tools needed terms.


dlglin:
When I start typing I get suggestions, but if I type more after that then the suggestions disappear.  Then if I type even more they reappear.

pstaabp:
Added an `update-ww.pl` script that will 
1. run `OPL-update` without the interactive loadOPLstats part of the script
2. run `build-search-json.pl` to generate the search-json for the POD/sample probs. 

Note: this will need to be added to the upgrade instructions. 

So now to test, run `update-ww.pl`.  To skip the OPL-update use the `-s` flag.  Then test the search box on the sample problems page.

bin/build-search-json.pl
```diff
@@ -0,0 +1,262 @@
...
+   -p|--pg-root          Directory containing  a git clone of pg.
+                         If this option is not set, then the environment
+                         variable $PG_ROOT will be used if it is set.
+   -f|--json-file        Location (relative to WW_ROOT) to store the resulting JSON file.
+                         Default value is htdocs/DATA/search.json
+   -s|--sample-prob-dir  Location (relative to $PG_ROOT) where the sample problems are located.
+                         Default value is tutorial/samples-problems
+   -b|--build            One of (all, macros, samples) to determine if the macros, sample
+                         problems or both should be scraped for data.
+   -v|--verbose          Setting this flag provides details as the script runs.
```
drgrice1, line 18: This script needs help.  I mean that literally.  Any script with options should have a `-h|--help` flag that shows how to use the script.

bin/build-search-json.pl
```diff
@@ -0,0 +1,262 @@
...
+	'b|build=s'           => \$build,
+	'w|show-warnings'     => \$show_warnings,
+	'v|verbose+'          => \$verbose
+);
+
+die "The build options must be one of (all, macros, samples). The value $build is not valid."
+	if ((grep { $_ eq $build } qw/all macros samples/) != 1);
+
+my $ww_root = $ENV{WW_ROOT};
+$ww_root = Mojo::File->curfile->dirname->dirname->realpath unless defined($ww_root);
```
drgrice1: Drop the usage of the `WW_ROOT` environment variable entirely.  Just use the correct path relative to the location of this file.  The correct environment variable that we have always used (but should not be used in general) is `WEBWORK_ROOT`, so definitely do not introduce a new one that no one will have set.  Also, there is no need for the `realpath` call on the next line.  So this should be
```suggestion
my $ww_root = Mojo::File->curfile->dirname->dirname;
```

bin/build-search-json.pl
```diff
@@ -0,0 +1,262 @@
...
+use feature "say";
+
+use Getopt::Long qw(:config bundling);
+use File::Find;
+use Mojo::JSON qw(encode_json);
+use Mojo::File qw(curfile);
+use Pod::Simple::SimpleTree;
+
+my $build   = "all";
+my $pg_root = $ENV{PG_ROOT};
```
drgrice1: Also drop the usage of the `PG_ROOT` environment variable.  Once the webwork2 root directory is determined relative to this script, then load the `webwork2.mojolicious.yml` file and get the PG root directory from the settings.  See how this is done in the `bin/check_latex` script for example (or in the `bin/webwork2` app script itself).

bin/download-OPL-metadata-release.pl
```diff
@@ -80,8 +80,12 @@ BEGIN
 die "The directory $libraryDirectory does not exist or is not writable."
 	if (!-d $libraryDirectory || !-w $libraryDirectory);
 
+warn $ENV{OPL_GIT_REPOSITORY};
```
drgrice1: Obviously this line needs to be deleted.

bin/update-ww.pl
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env perl
+
+=head1 NAME
+
+update-ww.pl -- run all needed scripts when webwork is updated.
```
drgrice1: Don't add this script.  The `OPL-update` script is already a script that just runs other scripts.  Now you are going to add another one that runs that script and one more?  Find another way to do it.

In addition the options for this script are not well explained, and there is no `-h` help flag.

bin/update-ww.pl
```diff
@@ -0,0 +1,72 @@
...
+use strict;
+use warnings;
+use feature 'say';
+
+use Mojo::File;
+use Getopt::Long qw(:config bundling);
+
+# BEGIN {
+# 	use Env        qw(WEBWORK_ROOT);
+# }
```
drgrice1: When lines like this are in new files in a pull request, that means this pull request is really not ready for review.  Look over the code, and clean it up.

drgrice1 (changes requested):
It seems that you need to run `npm install minisearch`, and then add the changed `package.json` and `package-lock.json` files to this pull request.

I haven't delved into this in much depth yet.  I can see that this pull request clearly needs a lot of work before it is really ready for review.

bin/update-ww.pl
pstaabp: @drgrice1 You had suggested that I add this.   I thought the idea was that this script would run anything that we needed when WeBWorK was upgraded.   Are you suggesting in the `OPL-update` to include the building of the search JSON?

drgrice1: Yes, that is what I actually meant with the suggestion.  That the `OPL-update` includes running this script.  Not the creation of a new script.  It is a bit of a stretch that this goes under the description of "updating the OPL", but the point is that everything be handled by a single script.  However, that is minor, and the point is to not need to change the instructions for installation/upgrade.

Although, I have also though about just removing the need for this new script entirely.  Have you actually benchmarked the script?  How bad would performance be if the files were dynamically combed at runtime, and the search actually done live?  If it isn't bad, or with some tweaking could be done efficiently at runtime, then that is the ideal way to go.

drgrice1: Another possibility is to effectively run the script the first time the server starts if the JSON file does not already exist.  Perhaps with a job queue task that handles this.

pstaabp: Running it seems to take 5-10 seconds which would be pretty bad to run on the page load.

I'll try to see if we can run it at server start and hand off to the job queue.  That sounds like a good idea.

drgrice1: Have you considered that the reason the script takes 5-10 seconds to run is that your code is highly inefficient?  And that improvements could be made to make it run in a fraction of the time?  I know this to be true in fact, and I made a small change that makes it so that running the script takes around 4 tenths of a second.  I am certain better can be done with more effort.

Have you also actually looked at the result of running the script and the words that are being listed in the JSON file?  That clearly needs more thought.  I am pretty sure that we don't want the word "`exppi" which is listed (for example from the ProvingTrigIdentities.pg sample problem) because "`exp(pi" appears in a file and the script removes the parenthesis.

pstaabp:
If you put in a PR to this branch, I'll deal with the conflicts.  I can made it run in the webwork2 script and probably is unnecessary and then can remove that.

drgrice1:
Most likely I am just going to take this over and put in a completely new pull request.  I am sorry, but there is a lot wrong with this pull request, and almost everything needs to be completely rewritten.  There aren't going to be any scripts.  Everything is going to be done efficiently and dynamically directly from the code.

pstaabp:
That's fine @drgrice1

PR #2734, opened 2025-06-01
Use tex-svg-full.js instead of tex-svg.js for MathJax.

somiaj:
Some MathObjects create TeX that uses `\boldsymbol`, which needs to be loaded to use in MathJax.

Side note, why is `noerrors` being loaded here? I actually find the errors MathJax gives when a complicated formula is malformed quite useful in debugging. Is it to limit memory/processing power of the clients?

### DIFF
--- a/lib/FormatRenderedProblem.pm
+++ b/lib/FormatRenderedProblem.pm
@@ -91,7 +91,7 @@ sub formatRenderedProblem {
 		[ 'node_modules/jquery-ui-dist/jquery-ui.min.js',                      0, {} ],
 		[ 'node_modules/iframe-resizer/js/iframeResizer.contentWindow.min.js', 0, {} ],
 		[ 'js/MathJaxConfig/mathjax-config.js',                     0, { defer => undef } ],
-		[ 'node_modules/mathjax/es5/tex-svg.js',                    0, { defer => undef, id => 'MathJax-script' } ],
+		[ 'node_modules/mathjax/es5/tex-svg-full.js',               0, { defer => undef, id => 'MathJax-script' } ],
 		[ 'node_modules/bootstrap/dist/js/bootstrap.bundle.min.js', 0, { defer => undef } ],
 		[ 'js/Problem/problem.js',                                  0, { defer => undef } ],
 		[ 'js/System/system.js',                                    0, { defer => undef } ],
--- a/templates/layouts/system.html.ep
+++ b/templates/layouts/system.html.ep
@@ -25,7 +25,7 @@
 % # JS Loads
 <%= javascript $c->url({ type => 'webwork', name => 'htdocs', file => 'js/MathJaxConfig/mathjax-config.js' }),
 	defer => undef =%>
-<%= javascript $c->url({ type => 'webwork', name => 'htdocs', file => 'node_modules/mathjax/es5/tex-svg.js' }),
+<%= javascript $c->url({ type => 'webwork', name => 'htdocs', file => 'node_modules/mathjax/es5/tex-svg-full.js' }),
 	id => 'MathJax-script', defer => undef =%>
 <%= javascript $c->url({ type => 'webwork', name => 'htdocs', file => 'node_modules/jquery/dist/jquery.min.js' }) %>
 <%= javascript $c->url({


somiaj:
Here is a problem to show case the error (reminder run `generate-assets.js` or `npm ci` to test this change).

```
DOCUMENT();

Context('Vector');
$v = Compute("x i + y j + z k");
BEGIN_PGML
[`[$v]`]
END_PGML
ENDDOCUMENT();
```

Without loading `boldsymbol`, MathJax won't render the unit vectors and shows a red `\boldsymbol` error.

Alex-Jordan:
We are using the [tex-svg combined component](https://docs.mathjax.org/en/latest/web/components/combined.html#tex-svg) which loads the `autoload` extension, which includes `boldsymbol`. So I'm not sure what could be preventing that from working for you.

somiaj:
I have MathJax on my system set to use the `CHTML` render, if I switch to `SVG` it goes away, so wonder why `CHTML` isn't autoloading that extension but `CHTML` requires it to be manually loaded.

Alex-Jordan:
Also if I load the page with CHTML in effect and change to SVG, that does not clear up the issue. Only after reloading the page with SVG in effect.

Alex-Jordan:
Does it also resolve things for you to explicitly load the `autoload` extension? If so, maybe we should do that. To catch not only this, but maybe more.

somiaj:
Trying to force `autoload` to load breaks `chtml` even worse, not only does it not work, I get errors in the console about "No version information available for component [tex]/autoload". So it is appearing that is some issue with `autoload` and `chtml` in MathJax, but I may not know mathjax well enough to know what.

somiaj:
@drgrice1 thanks. My tests show that loading `tex-svg-full.js` no longer has the issues I was experiencing with the `CHTML` math render. I found a second instance of just loading `tex-svg.js` and changed that to `tex-svg-full.js` as well.

somiaj:
Looking at this a bit closer, I'm unsure if using the `-full` to load all TeX extensions is the best solution here, it works but feels like it could have other consequences. I wonder what the underlying cause of autoload not working for chtml when using `tex-svg.js`. Though I just tested on my site which uses chtml by default, and if I switch to svg, autoload fails in that case too. Something about autoload makes it not work if someone manually chooses the math renderer not being loaded by the site. Wonder if this should just be reported to MathJax.

drgrice1:
It seems that it is a MathJax bug.  Switching to `tex-svg-full.js` is not exactly ideal.  It configures TeX to use all extensions except the `colorv2` and `physics` extensions, while the `tex-svg.js` only loads the `ams`, `newcommand`, `require`, `autoload`, `configmacros`, and `noundefined` extensions.

If you want to switch to only adding the `boldsymbol` extension as you had it before, that should be fine.  That seems to be working also.

somiaj:
I'll report the bug to MathJax, see what comes of it. The only thing with only loading `boldsymbol` is that is just one case I've ran across, any other extensions that pg uses by default might also be missing. But I think that solution is better than just blinding loading all extensions.

drgrice1:
I don't think it hurts to load all extensions though.  Note that the other extensions are loaded anyway as needed with `tex-svg.js`.  It is just that for some reason that is not working with the `boldsymbol` extension and chtml.

somiaj:
Well if you don't think there will be any downsides to just loading everything, we can leave it as is. Though I still think this is working around a bug in MathJax with autoload. I tested `\bbox` and `\verb` on my local site and have the same issue if I change the render. `boldsymbol` was just the extension I initially ran into.

Alex-Jordan:
@dpvc, are you following this? Is this a known issue with MathJax 3.2, or something specific to the WeBWorK/PG usage?

somiaj:
I posted an issue on MathJax for this, https://github.com/mathjax/MathJax/issues/3370.

somiaj:
I don't think the issue is specific to WeBWorK/PG, as I get the same thing on my website that uses MathJax as well for my course notes.

dpvc:
@Alex-Jordan, no I don't follow most of the WeBWorK issues unless I'm tagged, so I didn't see this.

I have answered @somiaj's issue on the MathJax tracker.  It includes a work-around configuration.

somiaj:
We could think about using the work around, or maybe since this has probably been an issue for a while, not worry about it and wait until version 4 which it should be fixed.

PR #2735, opened 2025-06-03
ability to cap the number of problems per page in a test

Alex-Jordan:
This arose from Runestone subscribers using tests with all the problems rendering on one page, and too many students starting a test at the same time, and the server crashed.

With this, the site admin can now declare a maximum for how many problems can be on one page of a test. So even if the instructor is trying to put 10 problems per page (or 0 for unlimited) the site admin might have imposed a cap of 2. And the 2 will be respected.

There are four places in `lib/WeBWorK/ContentGenerator/GatewayQuiz.pm` where code is in place to ensure the cap is respected. I am not sure these are all necessary. I put them all in for now because I couldn't think through cases where one or two may not be needed.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -385,6 +385,21 @@ async sub pre_header_initialize ($c) {
 			}
 		}
 	} else {
+		# If there is a cap on problems per page, make sure that is respected in the global set in
+		# case something higher snuck in.
+		if ($ce->{test}{maxProblemsPerPage}) {
+			my $globalSet = $db->getGlobalSet($setID);
+			if (
+				$ce->{test}{maxProblemsPerPage}
+				&& ($globalSet->problems_per_page == 0
+					|| $globalSet->problems_per_page > $ce->{test}{maxProblemsPerPage})
+				)
+			{
+				$globalSet->problems_per_page($ce->{test}{maxProblemsPerPage});
+				$db->putGlobalSet($globalSet);
+			}
+		}
+
 		# Get the template set, i.e., the non-versioned set that's assigned to the user.
 		# If this failed in authz->checkSet, then $c->{invalidSet} is set.
 		$tmplSet = $db->getMergedSet($effectiveUserID, $setID);
@@ -393,6 +408,13 @@ async sub pre_header_initialize ($c) {
 		# graded proctored tests.  If a set was not obtained from the database, store a fake value here
 		# to be able to continue.
 		$c->{assignment_type} = $tmplSet->assignment_type || 'gateway';
+
+		# If there is a cap on problems per page, make sure that is respected in case something higher snuck in.
+		if ($ce->{test}{maxProblemsPerPage}
+			&& ($tmplSet->problems_per_page == 0 || $tmplSet->problems_per_page > $ce->{test}{maxProblemsPerPage}))
+		{
+			$tmplSet->problems_per_page($ce->{test}{maxProblemsPerPage});
+		}
 
 		# next, get the latest (current) version of the set if we don't have a
 		#     requested version number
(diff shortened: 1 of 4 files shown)


somiaj:
Would a better approach to this be just make the configuration filed a drop down menu if this is set, or limit the range on it? That way users are informed about the cap if they try to go beyond it? This would also mean that when importing set definition files, the test for the cap needs to be set there. Unsure about how to deal with current tests if this is changed. So I guess probably both are needed, the cap in GatewayQuiz.pm along with the configuration page informing users of the cap and not letting them go beyond it.

Alex-Jordan:
I wanted to make sure if a high value was already in the database for some reason, that it will be replaced. For example if someone on Runestone currently has a test with unlimited problems per page ("0") then when I turn this on there, I want it to take effect.

How would a dropdown menu work when umlimited is an option?

There is a change in `lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm` to the `FIELD_PROPERTIES` constant. That is a constant in this file, and I am not sure I can access the course environment variables to assess whether or not a cap is in effect. If I could, I would have the help text be more specific and state the cap there.

somiaj:
And yes, as I was typing out my suggestion I realized that there is probably no way you can ensure the actual setting is within the set limit, so the restrictions tests in the gateway module are also needed.

Though this makes me wonder, could you put those restrictions in the database call methods, so the database method would apply any set restrictions there, vs you having to test in every place it is used?

lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -385,6 +385,20 @@ async sub pre_header_initialize ($c) {
 			}
 		}
 	} else {
+# If there is a cap on problems per page, make sure that is respected in the global set in case something higher snuck in.
```
drgrice1: Please wrap and indent this line.  So make it something like
```suggestion
		# If there is a cap on problems per page, make sure that is respected
		# in the global set in case something higher snuck in.
```
You probably had this indented, but perltidy removed the indentation to fit it into the 120 character limit.

Alex-Jordan:
This now has that comment indentation fixed. And it has the change that @somiaj suggested to use a dropdown when appropriate.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -365,7 +365,8 @@ use constant FIELD_PROPERTIES => {
 				. 'This situation can lead to their typed answers being lost and unrecoverable.  Additionally, having '
 				. 'many problems load at the same time on one page can put a strain on the server.  This is especially '
 				. 'worth considering if the test has many dynamically generated images, which can slow things down '
-				. 'significantly.'
+				. 'significantly. The site administrator may impose a cap on this number and trying to use a larger '
+				. 'value will automatically be converted down to the cap.'
```
somiaj: Could you add this message only if the admin has chosen a cap with your modification of of the proprieties below? Maybe instead you could just say, "The site administrator has imposed a cap of X problems per page." and append that to the string? Though unsure how to work this well with translations.

somiaj:
I like the drop down, but if the cap is set to '1' a dropdown with one item is a bit awkward, would it be possible to just put an uneditable field in this case? Though if this is a pain, a dropdown with a single item is fine, this is mostly a side thought.

Alex-Jordan:
This has a few changes.

- If the cap on problems per page has been set to 1, the problems per page option simply doesn't appear.
- The help text message is different than the default help text if there is a cap.
- A large (and perhaps controversial) change is that I added a new "type" for each field. If you give it something like `[0, undef, 1]` (as opposed to "edit", "choose", "hidden", or view") then it is understood to mean an input field with type "number". And the three things in the array are `[min, max, step]` for the number input field. I made this for the `problems_per_page` field and then also applied it everywhere it seems appropriate. I could remove it from all those other places if it is controversial to use it there.

Alex-Jordan:
Initially I applied the `[min, max, step]` type to more fields. @somiaj pointed out that using the number input fields with step set means there will be validation and disallow users from entering things like decimal values or otherwise in between steps. So I reverted except for fields where it really has to be an integer.

drgrice1:
My approval still stands on this with the latest changes.

somiaj (approved):

PR #2736, opened 2025-06-03
Implement suggested workaround for erroneous extension loading with CHTML renderer.

drgrice1:
This is an alternate for #2734, and implements the workaround suggested by @dpvc in https://github.com/mathjax/MathJax/issues/3370#issuecomment-2928277846.

We just need to decide which approach is best.

### DIFF
--- a/htdocs/js/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/MathJaxConfig/mathjax-config.js
@@ -98,6 +98,24 @@ if (!window.MathJax) {
 					AM.symbols.splice(i, 0, { input: trigger, ...newTriggers[trigger].symbols });
 				}
 
+				// The following is a workaround for a bug in MathJax when the math renderer is changed.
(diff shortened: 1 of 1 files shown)


dpvc:
You should probably note that if you move to v4, you will need to remove that code.  Or you could put it in an if-then that checks the version of MathJax before making the changes.

drgrice1:
Since we lock in the version of MathJax used by webwork (via the package-lock.json file), we will just remove this code when we update.  I added a note in the comment.

Alex-Jordan:
Looks like MJ4 will use `pnpm` instead of `npm`. Not sure what that means wrt the package lock file, but there will be a lot to look into when we go to MJ4.

https://github.com/mathjax/Mathjax-src/releases

drgrice1:
That is for their package management.  That does not affect what we use.

somiaj (approved):
This works in my tests. My gut feeling is this is probably the better approach.

Issue #2740, opened 2025-06-12
Should checl_modules.pl use site.conf?


Alex-Jordan:
It seems as though `check_modules.pl` looks for various utilities in the `$PATH` environment variable. Should it instead use the explicit paths that are set in `site.conf`?

I'm working on an installation where LaTeX was installed but not added to my path. So I noticed all these things flagged by `check_modules` even though the paths I set in `site.conf` are good.

drgrice1:
`check_modules` cannot access the course environment or `site.conf`.  That would create a catch 22 situation.  `check_modules` is checking that the Perl modules necessary for the webwork2 app to run are installed and at the correct versions.  The course environment will not work unless those modules are installed and present.

So the only way to accomplish what you are asking is to split the `check_modules` script into two scripts.  Once that checks the Perl modules, and another that checks the system binaries that are needed.

Alex-Jordan:
That makes sense. Although, could it just not even try to check system tools until critical perl modules pass? Except then we'd have to do a better job with things like checking `DBD::mysql` versus `DBD::MariaDB`.

It would also be nice to have a `check_modules_for_pg`. For local PG processing for PreTeXt. It's hard to know which modules in `check_modules` are needed for PG, which for webwork2, and which for both.

Issue #1917, opened 2023-03-22
Unused modules in check_modules?


dlglin:
Looking through `check_modules.pl` there are a number of modules that are not referenced anywhere in the WW or PG code:

- Array::Utils
- Email::Sender::Simple
- HTML::Tagset
- HTML::Template
- IO::Socket::SSL
- Net::LDAPS
- Net::SMTP
- Net::SSLeay
- PadWalker
- Path::Class
- Template

Are these no longer required?  Is it possible that other perl packages depend on some of these?  If so, do we need them in `check_modules.pl`?

If these truly aren't being used then I can put in a PR to remove them from `check_modules.pl`.

drgrice1:
I think those need to be deleted.  If a package depends on them, then there is no need for them to be checked as long as the package that depends on them is checked.

pstaabp:
`Net::LDAP` is used, but not in `check_modules`.   Looks like the `LDAPS` version is for secure, but `Net::LDAP` seems to handle security as well. Maybe replace `Net::LDAPS` with `Net::LDAP`

pstaabp:
Also, if we're going to remove these, check the packages in Docker as well.  I just did a search for padwalker which is in the Dockerfiles.

drgrice1:
Docker doesn't need to install libpadwalker-perl.  It also doesn't need to install libarray-utils-perl, libhtml-template-perl, libtemplate-perl, and libpath-class-perl.  There may be others too.

pstaabp:
> Docker doesn't need to install libpadwalker-perl. It also doesn't need to install libarray-utils-perl, libhtml-template-perl, libtemplate-perl, and libpath-class-perl. There may be others too.

@dlglin if you put in a PR for this, let's clean up all of the Docker stuff as well.

dlglin:
Does Docker take care of dependencies, or do they need to be explicitly listed?

dlglin:
node is in `check_modules`, but npm isn't.  Should npm be there?

drgrice1:
Which dependencies are you referring to for Docker?  If you mean the Ubuntu packages, then that isn't a Docker controlled thing.  That would be apt.  apt will pull in any necessary dependencies.  They do not need to be explicitly listed.

Yes, `npm` should be in `check_modules`.  It is possible to have `node` installed, but not have `npm` installed.  So really both should be listed.

pstaabp:
[...]

PR #2741, opened 2025-06-13
Rework check_modules.pl.

drgrice1:
The script now uses the settings in the course environment for external programs and to check the database driver (DBD::mysql or DBD::MariaDB). In order for that to work, first the script checks that all of the needed modules are found, load, and are at the sufficient version. That is except for the database driver module.  If any are not found, then the script exits with a message stating that it is unable to continue without the required modules.

If all modules are good, then the script checks the database driver module next. For this it "runtime" loads the course environment and its dependencies (as well as a runtime detection of the webwork2 root and lib directories).

This is to address issue #2740.

In addition the following modules are removed that are no longer needed:

* Array::Utils
* Email::Sender::Simple - still used but is a dependency of Email::Stuffer and doesn't need to be checked separately
* HTML::Tagset
* HTML::Template
* IO::Socket::SSL
* Net::LDAPS
* Net::SMPT
* Net::SSLeay
* PadWalker
* Path::Class
* Safe - we use our local version (WWSafe.pm) and still check Opcode which it uses
* Statistics::R::IO
* Template

`npm` is added to the list of required applications.

Most of these things were noted in issue #1917.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -1,53 +1,39 @@
 #!/usr/bin/env perl
-#
 
 =head1 NAME
 
-check_modules.pl - check to ensure that all applications and perl modules are installed.
+check_modules.pl - Check to ensure that applications and perl modules needed by
+webwork2 are installed.
 
 =head1 SYNOPSIS
 
 check_modules.pl [options]
 
  Options:
-   -m|--modules          Lists the perl modules needed to be installed.
-   -p|--programs       	 Lists the programs/applications that are needed.
-   -A|--all         		 checks both programs and modules (Default if -m or -p is not selected)
+   -m|--modules          Check that the perl modules needed by webwork2 can be loaded.
+   -p|--programs         Check that the programs needed by webwork2 exist.
+
+Both programs and modules are checked if no options are given.
 
 =head1 DESCRIPTION
 
-Lists all needed applications for webwork as well as a perl modules.
+Checks that modules needed by webwork2 can be loaded and are at the sufficient
+version, and that applications needed by webwork2 exist.
 
 =cut
 
 use strict;
 use warnings;
 use version;
+use feature 'say';
+
 use Getopt::Long qw(:config bundling);
 use Pod::Usage;
-
-my @applicationsList = qw(
-	convert
-	curl
-	dvisvgm
-	mkdir
-	mv
-	mysql
-	node
-	tar
-	git
-	gzip
-	latex
-	pandoc
-	xelatex
-	dvipng
-);
 
 my @modulesList = qw(
 	Archive::Tar
 	Archive::Zip
 	Archive::Zip::SimpleZip
-	Array::Utils
 	Benchmark
 	Carp
 	Class::Accessor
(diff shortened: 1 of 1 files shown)


drgrice1:
I also forgot to mention that the script also looks up which of `dvisvgm` or `pdf2svg` the site is configured to use and only checks for the configured executable.

somiaj:
Note, I normally don't install `Statistics::R::IO`, which is only an optional module (unsure if this is the only optional one), but I don't think the script should fail if that module is missing.

drgrice1:
I removed `Statistics::R::IO`.  That is no longer a dependency at all, optional or otherwise.

somiaj:
What happens if someone uses the `-p` flag, but modules are missing so the site.conf cannot be loaded? Seems that flag would bypass the initial module test.

drgrice1:
Yeah, I was going to point that out.  If you use the `-p` flag, then you skip the module check, and there is nothing that can be done about that.  So if you use that flag the modules need to be installed first.

drgrice1:
If someone has already checked modules, and know they are all good, then they might just want to only check programs.  So I think there is no reason to remove the option.

somiaj (approved):
Works in my tests.

One thing I noticed is if for some reason a key doesn't exist in `$externalPrograms`, it does error properly, but the error I found just slightly off, `** latex2pdf not found in $PATH`, since there really isn't a bin called `latex2pdf`, would something like `** latex2pdf is not defined in site.conf` be more useful? Not really an issue, just something I noticed in my testing.

drgrice1:
@somiaj: Without rather heavily restructuring the script there is no way to do what you are asking.

drgrice1:
@Alex-Jordan: The `Email::Stuffer` package lists `Email::Sender::Simple` as a dependency regardless of the Perl version.  See https://metacpan.org/pod/Email::Stuffer.

drgrice1:
By the way, we don't actually directly use `Email::Sender::Simple`.  We only use it indirectly via `Email:Stuffer`.  So it really doesn't matter if it is a dependency or not.  It should not be in `check_modules.pl`.

Alex-Jordan:
OK, that's a good reason to remove it.

In general, suppose perl module X depends on module Y. But not consistently across perl versions. Then suppose a person uses cpan or cpanm to install module X and their perl version does not have X depending on Y. I would wonder whether cpan or cpanm would automatically install Y. My guess is no, and whatever the list is here, we should not trim the list based solely on one module being a dependency of another. That's just a comment on how we could think about this list.

drgrice1:
`cpanm` does install dependencies.  I am not sure for `cpan`.

drgrice1:
We recommend `cpanm` in any case.  We don't give instructions for `cpan` anymore.

Alex-Jordan:
That's all good, makes sense. I am just making a larger point beyond the Email packages. For example, at least one perl version has `File::Path` depending on `File::Spec` and `Carp`. But just because we check for `File::Path`, that does not mean we should drop the checks for those other two modules.

drgrice1:
By the way `Carp` is also a core Perl module.

PR #2745, opened 2025-06-19
Add back the "real" upload check.

drgrice1:
This was removed when the uploads were reworked, but I see now that it should not have been.  Without this a warning about the empty string not being a valid upload occurs if there is not an upload.


Alex-Jordan (approved):
This works for me on the PCC server.

PR #2747, opened 2025-06-20
give hints about password fallback

Alex-Jordan:
This makes it so when you go to add users, the password fields have placeholder text that indicates what the fallback will be.


somiaj (approved):
Looks good. Would it be worth adding `$fallback_password_source` to `localOverides.conf.dist` to make the option more visible to site admins?

PR #2749, opened 2025-06-20
Change LTI update to allow selecting multiple sets.

somiaj:
Update the LTI Mass Update page to allow updating either all users or a single user for multiple sets with a single update.

### DIFF
--- a/htdocs/js/LTIUpdate/ltiupdate.js
+++ b/htdocs/js/LTIUpdate/ltiupdate.js
@@ -1,33 +1,52 @@
 (() => {
-	// Store select options in dictionary to rebuild the user/set menus from.
-	const userOptions = {};
+	// Store select options in dictionary to rebuild the set select from.
 	const setOptions = {};
 	const userMenu = document.getElementById('updateUserID');
 	const setMenu = document.getElementById('updateSetID');
-	for (let i = 1; i < userMenu.length; i++) {
-		userOptions[userMenu[i].value] = userMenu[i];
-	}
-	for (let i = 1; i < setMenu.length; i++) {
-		setOptions[setMenu[i].value] = setMenu[i];
+	const setSelectAll = document.getElementById('selectAllSets');
+	for (const option of setMenu.options) {
+		setOptions[option.value] = option;
 	}
 
-	// Update user and set drop down menus to only include valid user/set combinations.
-	document.getElementById('updateUserID')?.addEventListener('change', (e) => {
-		const setList = e.target.options[e.target.selectedIndex].dataset.sets.split(':');
-		const selectedSet = setMenu.value;
-		while (setMenu.length > 1) setMenu.lastChild.remove();
+	// Update set select to only include valid sets for the current selected user.
+	const updateSetMenu = () => {
+		const setList = userMenu.options[userMenu.selectedIndex].dataset.sets.split(':');
+		const selectedSets = {};
+		let allSelected = true;
+		for (const option of setMenu.options) {
+			if (option.selected) selectedSets[option.value] = true;
+		}
+		while (setMenu.length > 0) setMenu.lastChild.remove();
 		setList.forEach((set) => {
 			setMenu.append(setOptions[set]);
+			const option = setMenu.lastChild;
+			if (option.value in selectedSets) {
+				option.selected = true;
+			} else {
+				option.selected = false;
+				allSelected = false;
+			}
 		});
-		setMenu.value = selectedSet;
+		setSelectAll.checked = allSelected;
+	};
+
+	// Deal with the select all sets checkbox to either select all sets or uncheck if a set is unselected.
+	userMenu?.addEventListener('change', updateSetMenu);
+	setMenu?.addEventListener('change', (e) => {
+		let allSelected = true;
+		for (const option of setMenu.options) {
+			if (!option.selected) allSelected = false;
+		}
+		setSelectAll.checked = allSelected;
 	});
-	document.getElementById('updateSetID')?.addEventListener('change', (e) => {
-		const userList = e.target.options[e.target.selectedIndex].dataset.users.split(':');
-		const selectedUser = userMenu.value;
-		while (userMenu.length > 1) userMenu.lastChild.remove();
-		userList.forEach((user) => {
-			userMenu.append(userOptions[user]);
-		});
-		userMenu.value = selectedUser;
+	setSelectAll?.addEventListener('change', () => {
+		for (const option of setMenu.options) {
+			option.selected = setSelectAll.checked;
+		}
+	});
+
+	// Update the set list on page load.
+	window.addEventListener('load', () => {
+		if (userMenu.value !== '') updateSetMenu();
 	});
 })();
--- a/lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm
+++ b/lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm
@@ -7,7 +7,7 @@ use WeBWorK::CourseEnvironment;
 use WeBWorK::DB;
 
 # Perform a mass update of grades via LTI.
-sub run ($job, $userID = '', $setID = '') {
+sub run ($job, $userID = '', $setIDs = '') {
 	# Establish a lock guard that only allows 1 job at a time (technically more than one could run at a time if a job
 	# takes more than an hour to complete).  As soon as a job completes (or fails) the lock is released and a new job
 	# can start.  New jobs retry every minute until they can acquire their own lock.
@@ -39,10 +39,16 @@ sub run ($job, $userID = '', $setID = '') {
 
 	# Determine what needs to be updated.
 	my %updateUsers;
-	if ($setID && $userID && $ce->{LTIGradeMode} eq 'homework') {
-		$updateUsers{$userID} = [$setID];
-	} elsif ($setID && $ce->{LTIGradeMode} eq 'homework') {
-		%updateUsers = map { $_ => [$setID] } $db->listSetUsers($setID);
+	if ($setIDs && $ce->{LTIGradeMode} eq 'homework') {
+		if ($userID) {
+			$updateUsers{$userID} = ref($setIDs) eq 'ARRAY' ? $setIDs : [$setIDs];
+		} else {
+			if (ref($setIDs) eq 'ARRAY') {
+				%updateUsers = map { $_ => $setIDs } $db->listUsers;
+			} else {
+				%updateUsers = map { $_ => [$setIDs] } $db->listSetUsers($setIDs);
+			}
+		}
 	} else {
 		if ($ce->{LTIGradeMode} eq 'course') {
 			%updateUsers = map { $_ => 'update_course_grade' } ($userID || $db->listUsers);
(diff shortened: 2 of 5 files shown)


lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm
```diff
@@ -39,10 +39,16 @@ sub run ($job, $userID = '', $setID = '') {
 
 	# Determine what needs to be updated.
 	my %updateUsers;
-	if ($setID && $userID && $ce->{LTIGradeMode} eq 'homework') {
-		$updateUsers{$userID} = [$setID];
-	} elsif ($setID && $ce->{LTIGradeMode} eq 'homework') {
-		%updateUsers = map { $_ => [$setID] } $db->listSetUsers($setID);
+	if ($setIDs && ref($setIDs) eq 'ARRAY' && $ce->{LTIGradeMode} eq 'homework') {
```
Alex-Jordan: All blocks except the last `else` fall under `$setIDs && $ce->{LTIGradeMode} eq 'homework'`, so this might be more human readable if they were wrapped in one conditional for that.

```
	if ($setIDs && $ce->{LTIGradeMode} eq 'homework') {
		if (ref($setIDs) eq 'ARRAY' && $userID) {
			$updateUsers{$userID} = $setIDs;
		} elsif (ref($setIDs) eq 'ARRAY') {
			%updateUsers = map { $_ => $setIDs } $db->listUsers;
		} elsif ($userID) {
			$updateUsers{$userID} = [$setIDs];
		} else {
			%updateUsers = map { $_ => [$setIDs] } $db->listSetUsers($setIDs);
		}
	} else {
	...
```

Alex-Jordan: It might simplify the logic to precede all of this with:

```
$setIDs = [$setIDs] unless (ref $setIDs);
```

and just always act as if it is an array, even with just one entry.

somiaj: Thanks, I tried to simplify the logic a bit. I don't want to do `$setIDs = [$setIDs]` as if it is just a single set, that string is needed to keep old behavior of choosing just a single test for messages.

Alex-Jordan (approved):
I have not actually tested, but once this is merged and I pull into production, I'll test there.

lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm
```diff
@@ -63,10 +69,22 @@ sub run ($job, $userID = '', $setID = '') {
...
 
-	if ($setID && $userID && $ce->{LTIGradeMode} eq 'homework') {
-		unshift(@messages, $job->maketext('Updated grades via LTI for user [_1] and set [_2].', $userID, $setID));
-	} elsif ($setID && $ce->{LTIGradeMode} eq 'homework') {
-		unshift(@messages, $job->maketext('Updated grades via LTI all users assigned to set [_1].', $setID));
+	if ($setIDs && ref($setIDs) eq 'ARRAY' && $ce->{LTIGradeMode} eq 'homework') {
+		if ($userID) {
+			if (ref($setIDs) eq 'ARRAY') {
+				unshift(@messages,
+					$job->maketext('Updated grades via LTI for user [_1] for [_2] sets.', $userID, @$setIDs));
```
drgrice1: This is not correct.  I believe the intent was to give the message `Updated grades via LTI for user (userID) for (number of sets updated) sets.`  However, that isn't what this code does.  Since the arguments to the `maketext` method are in list context (as arguments always are for any method -- barring certain prototypes), this instead will give the message `Updated grades via LTI for user (userID) for (first set id in list) sets.`  So this should be
```suggestion
					$job->maketext('Updated grades via LTI for user [_1] for [_2] sets.', $userID, scalar(@$setIDs)));
```

lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm
```diff
@@ -63,10 +69,22 @@ sub run ($job, $userID = '', $setID = '') {
...
+			} else {
+				unshift(@messages,
+					$job->maketext('Updated grades via LTI for user [_1] and set [_2].', $userID, $setIDs));
+			}
+		} else {
+			if (ref($setIDs) eq 'ARRAY') {
+				unshift(@messages, $job->maketext('Updated grades via LTI for all users for [_1] sets.', @$setIDs));
```
drgrice1: The same things happens here.  This should be
```suggestion
				unshift(@messages, $job->maketext('Updated grades via LTI for all users for [_1] sets.', scalar(@$setIDs)));
```

lib/WeBWorK/Authen/LTI/GradePassback.pm
```diff
@@ -43,20 +43,27 @@ sub massUpdate ($c, $manual_update = 0, $userID = undef, $setID = undef) {
...
-		} elsif ($setID && $ce->{LTIGradeMode} eq 'homework') {
-			warn "LTI Mass Update: Queueing grade update for all users assigned to set $setID.\n";
+		if ($setIDs && ref($setIDs) ne 'ARRAY' && $userID && $ce->{LTIGradeMode} eq 'homework') {
+			warn "LTI Mass Update: Queueing grade update for user $userID and set $setIDs.\n";
+		} elsif ($setIDs && ref($setIDs) ne 'ARRAY' && $ce->{LTIGradeMode} eq 'homework') {
+			warn "LTI Mass Update: Queueing grade update for all users assigned to set $setIDs.\n";
+		} elsif ($setIDs && $userID && $ce->{LTIGradeMode} eq 'homework') {
+			warn "LTI Mass Update: Queueing grade update for $userID for @$setIDs sets.\n";
+		} elsif ($setIDs && $ce->{LTIGradeMode} eq 'homework') {
+			warn "LTI Mass Update: Queueing grade update for all users for @$setIDs sets.\n";
```
drgrice1: A similar problem to the `maketext` calls happens here.  When an array is interpolated into a string it is in list context, and the entire array is joined with the value of `$LIST_SEPARATOR` (or `$"` for short).   See https://perldoc.perl.org/perlvar#$LIST_SEPARATOR.  So to make this show the number of elements in the list this should be
```suggestion
			warn "LTI Mass Update: Queueing grade update for $userID for " . scalar(@$setIDs) . " sets.\n";
		} elsif ($setIDs && $ce->{LTIGradeMode} eq 'homework') {
			warn "LTI Mass Update: Queueing grade update for all users for " . scalar(@$setIDs) . " sets.\n";
```

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
```diff
@@ -23,36 +23,32 @@ sub initialize ($c) {
...
 	} else {
-		$c->addgoodmessage($ce->{LTIGradeMode} eq 'homework'
-			? $c->maketext('LTI update of all users and sets queued.')
-			: $c->maketext('LTI update of all users queued.'));
+		$c->addgoodmessage($c->maketext('LTI update of all users and sets queued.'));
 	}
 
-	# Note that if somehow this point is reached with a setID and grade mode is "course",
-	# then the setID will be ignored by the job.
+	# Note that if somehow this point is reached with a setIDs and grade mode is "course",
```
drgrice1: Now that "setID" is plural, the "a" in "a setIDs" needs to be dropped.

lib/WeBWorK/ContentGenerator/Instructor/LTIUpdate.pm
```diff
@@ -23,36 +23,32 @@ sub initialize ($c) {
 
 	return unless ($c->param('updateLTI'));
 
-	my $setID       = $c->param('updateSetID');
-	my $userID      = $c->param('updateUserID');
-	my $prettySetID = format_set_name_display($setID // '');
+	my @setIDs  = ($c->param('updateSetID'));
+	my $userID  = $c->param('updateUserID');
+	my $allSets = $c->param('selectAllSets');
```
drgrice1: This needs to be reworked quite a bit.  The `selectAllSets` parameter really should be a JavaScript only parameter that is entirely ignored server side.  The list given in the `updateSetID` parameter has everything that is needed server side.  Note that one consequence of this incorrect usage is that if a single user is selected and some selection of one or more but not all sets are selected, then the message that is shown regarding the queued job reads something like "LTI update of user userID queued.", but gives no indication for which set(s) the update was queued for.

somiaj: Looks like my logic was flawed, I was trying to use this to keep the old message of 'all sets queued'. Is there a way I can tell if all items were selected or not, for some reason I only though I got the list of what was selected in the multiple select, and was looking for a way to tell if everything was selected or not. I guess I could compare it to the number of possible sets.

somiaj:
Update via @drgrice1 review. I do forget how to do plural in maketext, for instance the messages may say `1 sets` vs `1 set`, since in all cases I chose to just list the number of sets. I could list all the set IDs, I just thought that might get a bit long, but if that is preferred I could change `3 sets`, to 'sets: id1, id2, id3', and list all the actual set IDs, though this may get long when all sets are selected (which was I decided to just list the number of sets).

lib/Mojolicious/WeBWorK/Tasks/LTIMassUpdate.pm
```diff
@@ -63,10 +69,27 @@ sub run ($job, $userID = '', $setID = '') {
...
-		unshift(@messages, $job->maketext('Updated grades via LTI for user [_1] and set [_2].', $userID, $setID));
-	} elsif ($setID && $ce->{LTIGradeMode} eq 'homework') {
-		unshift(@messages, $job->maketext('Updated grades via LTI all users assigned to set [_1].', $setID));
+	if ($setIDs && ref($setIDs) eq 'ARRAY' && $ce->{LTIGradeMode} eq 'homework') {
+		if ($userID) {
+			if (ref($setIDs) eq 'ARRAY') {
+				unshift(
+					@messages,
+					$job->maketext(
+						'Updated grades via LTI for user [_1] for [_2] sets.', $userID, scalar(@$setIDs)
```
drgrice1: The syntax for a plural is `[plural,_1,set]`.  Note that there is fourth argument that can be given, but for English an `s` is appended by default if the fourth argument is not provided.  In the `po` files the translators generally will need to use the fourth argument.  So here this should be
```suggestion
						'Updated grades via LTI for user [_1] for [_2] [plural,_2,set].', $userID, scalar(@$setIDs)
```

drgrice1:
By the way, aren't you supposed to still be out climbing mountains, and not working on WeBWorK development?

somiaj:
@drgrice1 Is there a way to get the list of set IDs assigned to all users put into a hash `userID => $setIDs`, with a single database call. This way I don't have to call the database for each user in a loop (what is currently being done in the LTIUpdate templates file) and also have access to the number of sets assigned to each user, so I can determine if I should say 'all sets' or '5 sets'. Basically I want to the put the following in a single call (or maybe two, one to get user list, and one to get set list for each user).

```
	my @allUsers    = $db->listUsers;
	my %allUserSets = map { $_ => [ sort $db->listUserSets($_) ] } @allUsers;
```

somiaj:
Mountains have been climbed, lots of southern utah has been seen, now at the air b and b with some time to work.

drgrice1:
Wait.  That isn't right.  Just a second.

somiaj:
Oh wait, that is a single database call, missed that. Thanks.

somiaj:
@dlglin I think you need to restart the `webwork2-job-queue.service` after switching to this branch, as I am not able to reproduce what you are seeing. This is what I see when I select two sets:

```
Preparing to submit grade to LMS for user XXXXXXXX and set SETID1
LMS user id is not available for this user.
Preparing to submit grade to LMS for user XXXXXXXX and set SETID2
LMS user id is not available for this user.
```

I don't have fully functional LTI setup on my dev setup, so that second message is just failing due to that, but I am seeing both setIDs begin processed for all users in my logs in the "Job Manager".

dlglin:
Yup.  I forgot to restart the job queue.  It's working as expected now.

lib/WeBWorK/DB.pm
```diff
@@ -1388,7 +1388,7 @@ sub getUserSet {
 }
 
 sub getUserSets {
-	my ($self, @userSetIDs) = shift->checkArgsRefList(\@_, qw/user_id set_id/);
```
drgrice1: This is what is causing the issue that @dlglin noticed.  This shouldn't be changed.

somiaj:
@dlglin Could you retest things, I updated the logic. I removed testing if `$setIDs` is an array reference, since this function is only run in one place and it will always be an array reference, the logic now just assumes `$setIDs` is an array reference and no longer tests for that case.

I also updated the messaging a bit. If a single set is selected, I use the old message of 'LTI update of set SETID for user USERID queued' in that case, and only list the count of sets if more than one is selected.

htdocs/js/LTIUpdate/ltiupdate.js
```diff
@@ -1,33 +1,48 @@
...
 		});
-		setMenu.value = selectedSet;
+		setSelectAll.checked = allSelected;
+	};
+
+	// Deal with the select all sets checkbox to either select all sets or uncheck if a set is unselected.
+	userMenu?.addEventListener('change', updateSetMenu);
+	setMenu?.addEventListener('change', (e) => {
+		if (!e.target.selected) setSelectAll.checked = false;
 	});
```
drgrice1: This isn't working.  `e.target.selected` is always undefined.  So this might as well just always set `setSelectAll.checked = false`.  The problem is that `e.target` is always the select element, and not one of the options.  A select element does not have the `checked` property.  I suggest changing this to
```suggestion
	setMenu?.addEventListener('change', () => {
		let allSelected = true;
		for (const option of setMenu.options) {
			if (!option.selected) allSelected = false;
		}
		setSelectAll.checked = allSelected;
	});
```
With that the select all check will accurately reflect when all sets are selected or not.

drgrice1 (approved):
Other than the select all issue, I think this looks good.

htdocs/js/LTIUpdate/ltiupdate.js
somiaj: Updated.

I see, I wasn't trying to make it so the checkbox would be checked if someone manually selected everything, but only to remove the check if someone selected all then removed an item, so in my tests always setting it to false worked. But this is better, now the checkbox automatically becomes selected when all things are selected.

drgrice1: Yes, it worked, but for bad reasons!

drgrice1:
I will merge this now.

PR #2750, opened 2025-06-20
when updating courses, check for standard sylminks

Alex-Jordan:
As I open this PR, it only checks that the standard symlinks (Library, Contrib, capaLibrary, and Student_Orientation) are present in a course, have the appropriate permissions, and point to the right place. It does not yet repair them if they are missing or bad. But this PR should evolve to have that too. I will mark this as draft.

Test by trying to upgrade courses, rename a course's ID, or archive a course.

Note that in defaults.config, I need to move the library settings earlier in the file so they can be referenced as paths for the library links.


lib/WeBWorK/Utils/CourseManagement.pm
```diff
@@ -50,6 +50,7 @@ use constant {    # constants describing the comparison of two hashes.
 # 	checkCourseTables
 # 	updateCourseTables
 # 	checkCourseDirectories
+# 	checkCourseLinks
```
drgrice1: Instead of adding the new `checkCourseLinks` method just delete lines 48 through 53 here.  None of it belongs here.

PR #2752, opened 2025-06-23
Show Deprecated macros warning in PG editor and library Browser

pstaabp:
This uses code from @somiaj to take the list of deprecated macros in the PG object and display only for instructors in the PGeditor and library browser.  

Note: this needs openwebwork/pg#1256.

### DIFF
--- a/lib/WeBWorK/Utils/Rendering.pm
+++ b/lib/WeBWorK/Utils/Rendering.pm
@@ -268,11 +268,18 @@ sub renderPG ($c, $effectiveUser, $set, $problem, $psvn, $formFields, $translati
 			flags            => $pg->{flags},
 		};
 
+		# get a list of the deprecated macros
+		my @deprecated_macros;
+		for (keys %{ $pg->{pgcore}{PG_loadMacros}{macroFileList} }) {
+			my @dirs = split('/', $_);
+			push(@deprecated_macros, $dirs[$#dirs]) if $dirs[ $#dirs - 1 ] eq 'deprecated';
+		}
+
 		if (ref($pg->{pgcore}) eq 'PGcore') {
 			$ret->{internal_debug_messages} = $pg->{pgcore}->get_internal_debug_messages;
 			$ret->{warning_messages}        = $pg->{pgcore}->get_warning_messages();
 			$ret->{debug_messages}          = $pg->{pgcore}->get_debug_messages();
-			$ret->{deprecated_macros}       = $pg->{pgcore}{PG_loadMacros}{deprecated_macros};
+			$ret->{deprecated_macros}       = \@deprecated_macros;
 			$ret->{PG_ANSWERS_HASH}         = {
 				map {
 					$_ => {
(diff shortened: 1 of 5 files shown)


pstaabp:
Updated to use the `macroFileList` within `PG_loadMacros` field of the PGcore object.

lib/WeBWorK/Utils/Rendering.pm
```diff
@@ -253,10 +253,18 @@ sub renderPG ($c, $effectiveUser, $set, $problem, $psvn, $formFields, $translati
 			flags            => $pg->{flags},
 		};
 
+		# get a list of the deprecated macros
+		my @deprecated_macros;
+		for (keys %{ $pg->{pgcore}{PG_loadMacros}{macroFileList} }) {
+			my @dirs = split('/', $_);
+			push(@deprecated_macros, $dirs[$#dirs]) if $dirs[ $#dirs - 1 ] eq 'deprecated';
```
drgrice1: This should be
```suggestion
			push(@deprecated_macros, $dirs[-1]) if $dirs[-2] eq 'deprecated';
```

pstaabp:
Merged @drgrice1 's PR as noted above.

somiaj (approved):
Looks good.

One very minor thing, it seems that when chain loading macros, the chain loaded macros are listed before the originally loaded macro. For instance `loadMacros('littleneck.pl')` gives the following list of deprecated macros:

+ problemRandomize.pl
+ littleneck.pl

So even though the only macro loaded was `littleneck.pl`, a macro that was never loaded got listed first. I can't think of a way to fix this, just something I noticed.

drgrice1:
@somiaj: The order that the macros are displayed does not depend on if one macro loads the other at all.  It also doesn't depend on the order that the macros are loaded at all either.  The macros are stored in the `macroFileList` hash of the `PGloadfiles` object.  Since it is a hash it is not ordered.  So the order the macros are displayed will be completely random.  One time you will see `problemRandomilze.pl` first and `littleneck.pl` second.  Another time you will see the opposite order.  I just tried a problem that uses the `littleneck.pl` macro, and saw both orders in several renderings.

drgrice1 (approved):
I think this looks good at this point.

Although, if we merge this, then there are now going to be warnings in the library browser, on the set details page, on the statistics page, on the problem grader page, and in the PG problem editor.  So these message will be quite prevalent for instructors, and there are thousands of problems in the OPL that will give these warnings.  Notably, many problems use the `AnswerFormatHelp.pl` macro.  So we need to start working on those problems.  Particularly replacing `AnswerFormatHelp.pl` usage with `helpLink` usage.

dlglin:
Peter submitted https://github.com/openwebwork/webwork-open-problem-library/pull/1279 to replace `AnswerFormatHelp` with `helpLink`, though I still don't like pushing out these warnings to all instructors.  There are many problems in the OPL which use deprecated macros for which there aren't alternative problems available, and most instructors using these problems aren't in a position to update the problem code.

I think it's at least as likely that instructors see these warnings that the problems should not be used, and conclude that WeBWorK can't provide them with enough working problems.

drgrice1:
I agree that we may not be ready for this yet.  I think that work on OPL problems to fix deprecated macro usage should come first.

drgrice1:
I think that the pull request is ready.  However, the primary concern with this one is that it is going to make warnings appear for a large number of OPL problems.  I think that is not something that we want.  So we are waiting with this until we get the OPL into better  shape.

Alex-Jordan:
Would it make sense to make this configurable, and (for now) be off by default? If we merged it into develop, it would perhaps facilitate reviewing OPL problems to make improvements there.

drgrice1:
Yeah, that would work.   @pstaabp: Could you add a configuration option for this?

Perhaps we could even remove the configuration option later, once the OPL is in better shape.

Issue #2754, opened 2025-06-24
Save As allows users to create illegal file names


PhilipLoewen:
Here in a stock installation of WW2.19, I tried to use the problem editor on a problem from the Library. The editor balked, and suggested using the "Save As" button to create a local copy with a new name. I did this, forming the new name by changing xxx.pg to xxx+PDL.pg. My goal was to respect the provenance of the original while also indicating the source of its local version. The results were "interesting".

(In addition to using the "Save As" button, I had chosen the option to replace the current problem in the set with the new one I was saving.)

It seems that including the character "+" in a file name causes lots of trouble. Some things work (e.g., the problem renders OK), but others crash hard. First, trying to load the editor fills the screen with an error that I won't even try to describe. Second, trying to recover by using the WeBWorK File Manager to rename the file doesn't work either. The File Manager displays the file name OK, but when I select the file and click the Rename button, the system complains that the file name is invalid and refuses to do anything with it. Please go ahead and try these things.

I worked around this by logging in to the server with SSH, navigating the tree that starts with `/opt/webwork/courses/` to find the poorly-named file, and renaming it using standard file system commands. Then I could go back into WeBWorK to update the problem set definition with the correct file name using the GUI. So I am back in business, a little better-informed about how things work. However, for the sake of others, ...

I suggest that the "Save As" button should activate some logic that checks proposed names for acceptability before using them.

Thanks a lot.

dlglin:
I'm not sure if you automatically get notifications when a pull request references an issue you created, but #2755 has been opened to address this.  Until that is merged into the code you're stuck with the workaround you described.

Alex-Jordan:
I merged #2755, so this is fixed for version 2.20.

PR #2755, opened 2025-06-24
Check the file name for valid characters when using "Save As" in the problem editor.

drgrice1:
This ensures that the given file name has the same valid characters that the file manager requires.

Also, show the message "The text box now contains the source of the original problem. You can recover lost edits by using the Back button on your browser." any time that a save is not going to be done in the save_as handler, since it applies to any of those cases.

Finally, only set the `editFilePath` when a save really will occur.

This fixes issue #2754.


drgrice1:
Note that really, the redirect even in the case that a save does occur should be eliminated.  That takes quite a bit of restructuring of the problem editor code though.

pstaabp (approved):

PR #2757, opened 2025-06-25
Create missing symbolic links in a course templates directory when upgrading.

drgrice1:
This finishes @Alex-Jordan's work in #2750.  It checks for missing symbolic links in a course's templates directory when checking if courses need upgrading, and creates the missing links when upgrading the course.  If a symbolic link points to an incorrect location (as defined in the course environment), then the upgrade process attempts to delete the symbolic link and recreate it pointing to the correct location.
    
Note that the check for permissions of the symbolic links has been removed.  For symbolic links there are two permissions in consideration.  Those of the link itself, and those of the link target.  The permissions of the link itself cannot be changed.  Even the linux command line `chmod` utility cannot do this (contrary to what Google AI wants to tell you, there is no `-h` option for `chmod`).  Furthermore, linux completely ignores symbolic link permissions, and doesn't use them at all.  So the only permissions that really matter are those of the link target which linux does use.  However, the directories that are the target of these links do not belong to a particular course.  They are system directories (or part of the webwork2 installation), and setting the correct permissions on those is part of the installation of webwork2. It should not be part of a course upgrade process. Including these permissions in the upgrade report when not correct would actually mean that every course would show that it has the incorrect permission, or no courses would. Upgrading would never fix this, because the permissions for these directories cannot be changed from the user interface in any case.  What could be done here is to make another admin page that does a system status check, and lets the admin user know what directory permissions in the webwork installation are not correct what needs to be fixed. Of course the `setfilepermissions` script can be used to fix these permissions (it already handles most of the directories in question).

@Alex-Jordan: Hopefully you are okay with me opening a new pull request instead of making a pull request to your branch.  I thought this would be easier for me to maintain, since you said you didn't have time now to finish that pull request.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -449,10 +449,10 @@ $courseDirs{tmpEditFileDir}  = "$courseDirs{templates}/tmpEdit";
 $courseDirs{mailmerge}   = "$courseDirs{DATA}/mailmerge";
 
 # course links that are regulated by webwork2
-$courseLinks{Library}             = [$problemLibrary{root},                         "$courseDirs{templates}/Library"];
-$courseLinks{Contrib}             = [$contribLibrary{root},                         "$courseDirs{templates}/Contrib"];
-$courseLinks{capaLibrary}         = ["$contribLibrary{root}/CAPA",                  "$courseDirs{templates}/capaLibrary"];
-$courseLinks{Student_Orientation} = ["$webworkDirs{assets}/pg/Student_Orientation", "$courseDirs{templates}/Student_Orientation"];
+$courseLinks{Library}             = [ $problemLibrary{root},                         "$courseDirs{templates}/Library" ];
+$courseLinks{Contrib}             = [ $contribLibrary{root},                         "$courseDirs{templates}/Contrib" ];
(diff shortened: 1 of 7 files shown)


Alex-Jordan:
Thanks for picking this up. I will test it soon, perhaps this evening.

pstaabp (approved):
Works well. I agree with @Alex-Jordan if it's not a heavy lift to handle the rename course to fix the links as well.

drgrice1:
With using the directories from the course environment it is not possible to use relative paths for the links.

Actually, not using relative paths for the symbolic links is better than the current links in the model course anyway.  Perhaps you like relative links, but using absolute paths solves a problem.  That is that theoretically the location of the courses directory is configurable, as is the location of the OPL.  So if someone uses a non-standard location for the courses directory or for the OPL, then the links in the model course won't work.  However, the links created by this method will.  It seems that what we really should do is remove  the links from the model course, and when a course is created the links should be created from the configuration.  Perhaps when unarchiving a course the links could even be created/fixed at that time.  Then there would be no need for handling this in the course update, and the links would always be there and work.

PR #2759, opened 2025-06-27
Rework the search box for the sample problems page.

drgrice1:
This is a rework of #2733.
    
This does not add any scripts that need to be run when webwork is installed or upgraded.  Instead the search data is dynamically generated.  The first time that the sample problems home page is loaded all PG macros and sample problems are parsed and the search data extracted.  Then the data is saved to the `DATA/sample-problem-search-data.json` file. This takes less than half a second on my computer, but that will probably vary some depending on the server's processing capabilities. The last modified time of each file is also saved in that search data.  On subsequent requests the last modified time is checked, and a file is only parsed again if it has been modified since the last time the page was loaded.  If a file is deleted from PG, the data is removed from the file.  So even in development the data will always reflect the current state of the PG repository on the server.  Subsequent requests when no changes to PG occur usually complete in around 1/25th of a second. A change of a single file increases that time only slightly.
    
Note that the search data is accessed by JavaScript via the new `/sampleproblems/search_data` route.
    
The layout of the sample problems home page is changed considerably from #2733.  The layout now works on narrow screens and some accessibility and html validation issues were addressed.
    
There is also quite a bit of improvement in the search data that is saved.  There is no attempt to single out methods or functions for macros from the POD headers.  Unfortunately, that was never really going to work.  Instead the words from all headers (that are not stop words) are used.  So the methods or functions that are in a POD header will be in the search data, but other things will also be there.  There are some custom stop words added that are not desirable in the search data, like "description", "synopsis", "podlink", and "problink".
    
Also, macros that don't have POD in them are not indexed at all since there isn't anything to show for these files.


pstaabp (approved):
Looks good.  Thanks for shortening up the processing time and seems like a good workflow.

PR #2760, opened 2025-06-30
Move LTI login JavaScript out of html template

taniwallach:
I am trying to move JavaScript out of the the HTML files, in order to allow the core WW2 functionality to work with a Content-Security-Policy which does not require `script-src` to include `unsafe-inline` and without a need to add specific hash values to get mission-critical JavaScript functionality to work.

I am aware that several advanced features of PG will currently not work with such a CSP.

This PR allows the LTI 1.3 login process to work without needing either `unsafe-inline` or the relevant hash value in the CSP.

~~One small typo I noticed in an unrelated file was corrected.~~

A unneeeded comment with a type in an unrelated file was deleted.

These same changes were tested on my production WW 2.19 server

### DIFF
--- a/htdocs/js/LTILogin/ltilogin.js
+++ b/htdocs/js/LTILogin/ltilogin.js
@@ -1,0 +1,1 @@
+document.getElementById('ltiRepost')?.submit();
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -36,7 +36,6 @@
 % end
 %
 % content_for js => begin
-	% # Page specfific javascript
 	<%= javascript getAssetURL($ce, 'js/GatewayQuiz/gateway.js'), defer => undef =%>
 	%
 	% # Output javascript for jquery-ui for problems to use.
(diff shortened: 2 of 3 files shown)


templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -36,7 +36,7 @@
 % end
 %
 % content_for js => begin
-	% # Page specfific javascript
+	% # Page specific javascript
```
drgrice1: I would be even happier with deleting this comment that is really pointing out the obvious.

htdocs/js/LTILogin/ltilogin.js
```diff
@@ -0,0 +1,6 @@
+document.addEventListener('DOMContentLoaded', function() {
+	var form = document.getElementById("ltiRepost");
+	if (form) {
+		form.submit();
+	}
+});
```
drgrice1: The contents of this file should just be
```suggestion
document.getElementById('ltiRepost')?.submit();
```
This JavaScript file is deferred, and so it is loaded after the `DOMContentLoaded` event has already fired.  You could also just literally use the exact code you deleted from the `templates/ContentGenerator/LTI/self_posting_form.html.ep` file that was inside `javascript begin/end`.

templates/ContentGenerator/GatewayQuiz.html.ep
taniwallach: Ok - deleted.

htdocs/js/LTILogin/ltilogin.js
taniwallach: Thanks! Changed to what you suggested.

drgrice1 (approved):
Looks good.  There is still a long way to go to make a content security policy without `unsafe-inline` work, but this is a start.

pstaabp (approved):

PR #2763, opened 2025-07-03
Fix a potential XSS vulnerability in the past answers table (the Answer Log).

drgrice1:
If MathQuill is enabled and a student types `"<script>alert(1);</script>` and submits the answer, then that script is executed if an instructor views the answer on the past answers page.  If MathQuill is not enabled, and a student simply enters `<script>alert(1);</script>` the same thing happens.

This is due to a change in #1899 to make the special character `&#1970;` used to separate array answers (usually coming from checkbox answers) display correctly.

To fix this better handling of answers with that character is needed. So this splits the student answers on that character, and now DOES html escape all other parts of the student answer again, but directly inserts the `&#9070` characters without escaping. The effect is that answers will still be displayed the same as they were before other than scripts in the answer. Those will now actually appear as text. Note that is the same as it would have been before #1899.

### DIFF
--- a/templates/ContentGenerator/Instructor/ShowAnswers/past-answers-table.html.ep
+++ b/templates/ContentGenerator/Instructor/ShowAnswers/past-answers-table.html.ep
@@ -53,7 +53,11 @@
 									% if ($answer eq '') {
 										<small><i><%= maketext('empty') %></i></small>
 									% } else {
-										<%== $answer =%>
+										% my @parts = split("&#9070;", $answer);
+										% for (0 .. $#parts - 1) {
+											<%= $parts[$_] =%>&#9070;\
+										% }
+										<%= $parts[-1] =%>
 									% }
 								</td>
 							% }


templates/ContentGenerator/Instructor/ShowAnswers/past-answers-table.html.ep
```diff
@@ -53,7 +53,11 @@
 									% if ($answer eq '') {
 										<small><i><%= maketext('empty') %></i></small>
 									% } else {
-										<%== $answer =%>
+										% my @parts = split("&#9070;", $answer);
+										% for (0 .. $#parts - 1) {
+											<%= $parts[$_] =%>&#9070;<% =%>
```
taniwallach: Note: I'm not exactly sure why you used `<% =%>` at the end of this line, and again 2 lines below.

drgrice1: That is to prevent spaces and new lines in the injected html.  Without those the HTML will have text nodes injected that cause extra spacing.

taniwallach: Thanks. Eventually, I'll understand more of the fine points of Mojolicious.

I was curious about how this worked and if I understand https://docs.mojolicious.org/Mojo/Template it seems that putting a single `\` at the end of those lines should also work. I don't know if that is more efficient, but it may be as it may not "invoke" Perl processing of the empty interior of the current approach.

drgrice1: I realized that the `<% =%>` on the last instance on line 60 was certainly unnecessary because it already ends in `=%>`.  So that was removed.

I also switched to using `\` on the first instance on line 58 as that seems to be sufficient here.  Note that the difference between ending a line in a backslash and ending a template interpolation with `=%>` is that the backslash only removes the newline, and a template interpolation that ends with `=%>` removes all surrounding white space.  The reason that the backslash is enough on line 58 is that the earlier expression on that line does end with `=%>` which strips the white space on that line.  Also, the surrounding lines start with only `%` which are perl expressions only, and automatically have surrounding white space stripped.

drgrice1: I wish that Mojolicious templates would reverse the behavior of `=%>`.  If those are not given I would like white space to be stripped, and if they are given then white space would be preserved.  The point is that now days HTML should be minimized (all unnecessary white space should be removed).  Gone are the days where it is allowable to attempt for HTML to be human readable.  Use things like prettier if you want that.

PR #2764, opened 2025-07-11
record/update user lis_source_did even when grade passback not in use

Alex-Jordan:
Just because a course has LTIGradeMode set to false, we should still record a user's lis_source_did in case LTIGradeMode gets toggled later. This has happened both at PCC and with Runestone instructors, and they don't understand why some students are not having grades passed back even when the instructor initiates an LTI Grade Update. If they understood what was going on, they would know they can wait until those students visit WeBWorK again from the LMS and things will be fine. But they usually don't understand that level of detail about this process.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvantage.pm
+++ b/lib/WeBWorK/Authen/LTIAdvantage.pm
@@ -326,11 +326,9 @@ sub authenticate ($self) {
 		$self->{initial_login} = 1;
 	}
 
-	# If we are using grade passback then make sure the data we need to submit the grade is kept up to date.
-	my $LTIGradeMode = $ce->{LTIGradeMode} // '';
-	if ($LTIGradeMode eq 'course' || $LTIGradeMode eq 'homework') {
-		WeBWorK::Authen::LTIAdvantage::SubmitGrade->new($c)->update_passback_data($self->{user_id});
-	}
+	# In case we will use grade passback at some point,
+	# make sure the data we need to submit the grade is kept up to date.
+	WeBWorK::Authen::LTIAdvantage::SubmitGrade->new($c)->update_passback_data($self->{user_id});
 
 	return 1;
 }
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
perltidy is doing something odd with the comment in `lib/WeBWorK/Authen/LTIAdvantage.pm`. But this is what it takes to pass the test.

Alex-Jordan:
OK, nevermind. The comment is long, and was all on one line, so perltyidy replaced the tab with two spaces so it could still fit on one line under 120 characters. I broke it up over two lines.

drgrice1 (changes requested):
This should not be done for LTI 1.1.  There is just no point in that.  There is no user specific information saved if grade pass back is not enabled for the course.  If grade pass back is enabled later, then the global data will be obtained when one user signs in.  If grade pass back is enabled later and no users sign in, there still won't be sufficient data to pass grades back even if this is done.

Alex-Jordan:
OK, I reverted the change for LTI 1.1.

drgrice1:
There is one other change that should be made for this.  On line 72 of `lib/WeBWorK/Authen/LTIAdvantage/SubmitGrade.pm` you should add `return unless $ce->{LTIGradeMode}`.  There is nothing after that that should be done if grade mode is not enabled.  Most importantly the `$self->get_access_token` line should not be executed.  It is possible that someone does not intend to use grade pass back ever, and in that case they would not need to configure the LMS for grade pass back.  In that case webwork should not attempt to obtain the access token from the LMS.  The access token is obtained if needed anytime grade pass back occurs, so the access token is not needed at that point anyway even if grade pass back is enabled.  It is only there to get it out of the way then when grade pass back is enabled so that when grade pass back does occur it will be faster.

drgrice1 (approved):
Looks good now.

somiaj:
The code looks reasonable to me, but I don't have a way to test this.

PR #2748, opened 2025-06-20
Add the pg-critic problem analyzer to the PGEditor

pstaabp:
This adds the ability to run the `pg-critic` analyzer (see openwebwork/pg#1254 for details on the script) to run inside the "Format Code" tab of the PGeditor. 

The analyzer also produces a report on how to improve the problem in the editor.  

To use this, you'll need to install openwebwork/pg#1254.

PR #2768, opened 2025-07-16
Add "PG Critic" to the PG problem editor.

drgrice1:
This adds the PG critic results from `WeBWorK::PG::Critic` to the PG problem editor.  Note that the "Format Code" tab in the editor is now the "Code Maintenance" tab.  This is because this new option is not a code formatter.  All of the options fit into the category of code, so this is a better name.

This is dependent on https://github.com/openwebwork/pg/pull/1278 (so make sure to check that PG branch out to test this), and is intended to replace https://github.com/openwebwork/webwork2/pull/2748.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -88,6 +88,7 @@ my @modulesList = qw(
 	Net::OAuth
 	Opcode
 	Pandoc
+	Perl::Critic
 	Perl::Tidy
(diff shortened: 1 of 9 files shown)


Alex-Jordan (approved):
I think we should get this in, and then we can test it more organically as we are doing other things.

somiaj (approved):

PR #2788, opened 2025-08-14
Update third party npm dependencies.

drgrice1:
The following dependencies are updated:

* @fortawesome/fontawesome-free: 6.52 -> 7.0.0
* bootstrap: 5.3.3 -> 5.3.7
* iframe-resizer: 4.3.11 -> @iframe-resizer/parent and @iframe-resizer/child 5.5.2
* jquery-ui-dist: 1.13.2 -> 1.13.3
* luxon: 3.4.4 -> 3.7.1
* sortablejs: 1.15.2 -> 1.15.6
* autoprefixer: 10.4.19 -> 10.4.21
* chokidar: 3.6.0 -> 4.0.3
* cssnano: 6.1.2 -> 7.1.0
* postcss: 8.4.38 -> 8.5.6
* prettier: 3.2.5 -> 3.6.2
* rtlcss: 4.1.1 -> 4.3.0
* sass: 1.75.0 -> 1.90.0
* terser: 5.30.4 -> 5.43.1
* yargs: 17.7.2 -> 18.0.0

All libraries are now at their latest versions.

Note that the there are two changes to the the generate-assets.js script that were needed. There is a change in how yargs is used, and an option needed to be added to the `sass.compile` call to disable warnings about deprecated sass usage in Bootstrap's scss.

The `iframe-resizer` library has changed considerably and upgrading to the latest version requires quite a few changes.

First, the `iframe-resizer` package has been replaced with the `@iframe-resizer/parent` and `@iframe-resizer/child` packages. The latest version is 5.5.2 which this upgrades to.
    
Second, the `data-iframe-height` and `data-iframe-width` options are deprecated (and in fact don't work) and now there is only the `data-iframe-size` option. This requires a small change to PG as well (done in the last commit of https://github.com/openwebwork/pg/pull/1306).  Also, the `iFrameResize` method is deprecated and replaced with the `iframeResize` method (annoying change here, did you really need to change the case of one letter). The `iFrameResizer` property set on the iframe still exists but is not documented. The `iframeResizer` property replaces it. I assume the `iFrameResizer` is also deprecated (although this is not documented), so that switch was made.

Furthermore, they now have a commercial license version (which requires purchase) and an open source license version. We can use the open source version, but there are some annoying things that they do with that (like console logs that you can't prevent).

The first commit in this pull request upgrades iframe-resizer to version 4.4.2 which is the last version before things went haywire ans started doing stupid things.  The last commit in this pull request further upgrades to 5.5.2 and the new packages.  We can revert the last commit if we decide we can't tolerate these changes.

### DIFF
--- a/htdocs/generate-assets.js
+++ b/htdocs/generate-assets.js
@@ -2,7 +2,8 @@
 
 /* eslint-env node */
 
-const yargs = require('yargs');
+const yargs = require('yargs/yargs');
+const { hideBin } = require('yargs/helpers');
 const chokidar = require('chokidar');
 const path = require('path');
 const { minify } = require('terser');
@@ -14,7 +15,7 @@ const postcss = require('postcss');
 const rtlcss = require('rtlcss');
 const cssMinify = require('cssnano');
 
-const argv = yargs
+const argv = yargs(hideBin(process.argv))
 	.usage('$0 Options')
 	.version(false)
 	.alias('help', 'h')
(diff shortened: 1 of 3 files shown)


drgrice1:
I now removed the last commit with the update to the latest version of iframe-resizer as we discussed today.

Issue #2790, opened 2025-08-19
webwork 2.20 course admin prohibits email user_ids


glarose:
It appears that when ContentGenerator::CourseAdmin adds a course it prohibits added user_ids from being email addresses.  In add_course_validate (line 270), user_ids are required to match /^[\w.,-]*$/ .  I think we should allow "@" signs so that user_ids can be email addresses.

somiaj:
Fixed in #2791 #2792

PR #2791, opened 2025-08-19
Allow @ in user id when adding new course.

somiaj:
This is in response to #2790.

Should this also be a hotfix, emails as usernames can be a common?


drgrice1 (approved):
Yes, this should be made into a hotfix pull request as well.

glarose (approved):

PR #2793, opened 2025-08-20
Fix a `realpath` failure when resolving course symlinks.

drgrice1:
If one of the course links that are now maintained by webwork points to a location inside a non existent directory, then the `realpath` calls in `lib/WeBWorK/Utils/CourseDirectoryIntegrityCheck.pm` throw exceptions. This occurs both when checking for course upgrades and when upgrading courses.  So those calls are wrapped in evals to prevent the exceptions.

I have identified this as the cause of the issue discussed in https://forums.openwebwork.org/mod/forum/discuss.php?d=8757#p22321.

To test this delete the Contrib link in a course's templates directory, and create a bad link with something like

```bash
sudo ln -s /bad/location /opt/webwork/courses/courseId/templates/Contrib
```

Then go to the "Upgrade Courses" page in the admin course.  It will give an error with the current develop or main branches.  With this pull request it will show that the link structure of the course needs repair. Furthermore, repairing the link will work.


somiaj (approved):
Fixes the issue.

Alex-Jordan (approved):

PR #2797, opened 2025-08-28
Make LTI 1.3 launch redirection honor query parameters in the target_link_uri.

drgrice1:
This is useful if you want to make unpublished instructor links in the LMS that go to specific pages in webwork with query params to make certain display options active immediately.  For example, if you use a link such as `https://webwork.server.edu/webwork2/instructor/progress/set/setID?returning=1&show_date=1&show_testtime=1&show_problems=1` then the test dates, times, and problems will be shown when the page opens.

Currently URL parameters are dropped.  This was not intentional when I implemented this, it is merely a side effect of calling `systemlink` and `url_for` on the `target_link_uri` that is given.

This just gets the query parameters that are passed in the given `target_link_uri` and then passes them to `systemLink` so that it adds them back to the generated URL that gets redirected to.


somiaj:
I will approve and merge this, but I don't have a way to test. I trust that both @drgrice1 and @Alex-Jordan have both tested it.

PR #2799, opened 2025-09-02
Don't create users for LTI users that do not have permission to login.

drgrice1:
Currently if `$permissionLevels{login} = 'professor'` and a user signs in via LTI that would be assigned the role of "student", then webwork2 creates the user and signs the user in.  However, on subsequent LTI logins authentication fails.  This refuses to create a user if the requested role would not have permission to login.  @somiaj pointed this out recently in Slack.

Clean up the error messages some.  There is a lot of work left to do on this. The LTIAdvance.pm module has an extremely poor design for error handling and messaging to go with those errors.  The LTIAdvantage.pm module is only a tad better (I largely just copied the poor design of the LTIAdvanced.pm module).  The `log_error` key is set and appended to numerous times, frequently resulting in a long run on message that doesn't really make sense. Also, there were some of these errors that were adding "LOGIN FAILED".  That was removed because The `Authen.pm` code always prepends that and that resulted in logs with "LOGIN FAILED LOGIN FAILED ...".

The `authenticate` method is expected to return either 1 or a message indicating the failure.  Currently it returns either 1 or 0.  As a result the messages that are set in the `authenticate` method go into the abyss.  Those messages should be returned instead of setting `$self->{error}`. Note that the method can still return 0 if no message should be set (as in the case of the OAuth token failing to verify for LTI 1.1).

For LTI 1.3 make sure that the fallback_source_of_username is set before attempting to use it. Otherwise the claim extraction fails and it results in a database error later.

### DIFF
--- a/lib/WeBWorK/Authen/LTIAdvanced.pm
+++ b/lib/WeBWorK/Authen/LTIAdvanced.pm
@@ -117,10 +117,9 @@ sub get_credentials {
 	# Determine the WW user_id to use, if possible
 
 	if (!$ce->{LTI}{v1p1}{preferred_source_of_username}) {
-		warn
-			"LTI is not properly configured (no preferred_source_of_username). Please contact your instructor or system administrator.";
-		$self->{error} = $c->maketext(
-			"There was an error during the login process.  Please speak to your instructor or system administrator.");
+		$self->{error} = $c->maketext("There was an error during the login process.  "
+				. "Please speak to your instructor or system administrator.");
+		warn "LTI is not properly configured (no preferred_source_of_username).\n" if $ce->{debug_lti_parameters};
 		debug("No preferred_source_of_username in "
 				. $c->ce->{'courseName'}
 				. " so LTIAdvanced::get_credentials is returning a 0\n");
@@ -228,8 +227,8 @@ sub get_credentials {
 			warn "================================\n";
 		}
 		if (!defined($self->{user_id})) {
-			croak
-				"LTIAdvanced was unable to create a username from the data provided with the current settings. Set \$debug_lti_parameters=1 in authen_LTI.conf to debug";
+			croak "LTIAdvanced was unable to create a username from the data provided with the current settings. "
+				. "Set \$debug_lti_parameters=1 in authen_LTI.conf to debug";
 		}
 
 		$self->{login_type}        = "normal";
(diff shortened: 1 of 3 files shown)


lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -285,9 +285,8 @@ sub check_user {
 
 		foreach my $key (keys(%options), ($use_lis_person_sourcedid_options ? @lis_person_sourcedid_options : ())) {
 			if (defined($c->param($key))) {
-				debug(
-					"User |$user_id| is unknown but may be an new user from an LSM via LTI. Saw a value for $key About to return a 1"
-				);
+				debug("User |$user_id| is unknown but may be an new user from an LSM via LTI. "
+						. "Saw a value for $key About to return a 1");
```
somiaj: Typo? "LSM"

lib/WeBWorK/Authen/LTIAdvanced.pm
```diff
@@ -117,10 +117,10 @@ sub get_credentials {
...
-			"LTI is not properly configured (no preferred_source_of_username). Please contact your instructor or system administrator.";
-		$self->{error} = $c->maketext(
-			"There was an error during the login process.  Please speak to your instructor or system administrator.");
+		warn "LTI is not properly configured (no preferred_source_of_username). "
+			. "Please contact your instructor or system administrator.";
+		$self->{error} = $c->maketext("There was an error during the login process.  "
+				. "Please speak to your instructor or system administrator.");
 		debug("No preferred_source_of_username in "
 				. $c->ce->{'courseName'}
 				. " so LTIAdvanced::get_credentials is returning a 0\n");
```
somiaj: Do we really need these warn statements (here and below)? Wouldn't it be better to just show the user `$self->{error}` and put all releveant info in the `debug()` message. No need to state the page has errors, users who haven't enabled debugging shouldn't see that.

lib/WeBWorK/Authen/LTIAdvanced.pm
drgrice1: I fixed that typo.

lib/WeBWorK/Authen/LTIAdvanced.pm
drgrice1: I don't know.  It has always been this way.  I agree that a warning is not a good way to deal with this, but this is a warning that on a properly configured system will not be seen.  Students should never see it.  By the time students start signing in, the system will usually have been tested and properly configured.

drgrice1: I changed the warning and made it so that it is only shown if `debug_lti_parameters` is set.  I agree that this should not be shown to the end user, and that the `$self->{error}` is sufficient.

drgrice1: I also did the same with the warn statement "below".  It took me a bit to figure out which warn statement you were talking about "below".  It is way below with several other warn statements in between.

PR #2802, opened 2025-09-07
is this a concern ?

jeis4wpi:
is messsage a concern on this line of code ?

### DIFF
--- a/htdocs/js/TagWidget/tagwidget.js
+++ b/htdocs/js/TagWidget/tagwidget.js
@@ -48,7 +48,7 @@
 
 	// Load the library taxonomy from the JSON file.
 	const response = await fetch(tagWidgetScript.dataset.taxonomy).catch(
-		(err) => `Could not load the OPL taxonomy from the server: ${err.messsage ?? err}`
+		(err) => `Could not load the OPL taxonomy from the server: ${err.message ?? err}`
 	);
 	if (typeof response === 'string') return showMessage(response);
 	if (!response.ok) return showMessage('Could not load the OPL taxonomy from the server.');


lib/WeBWorK/Localize/webwork2.pot
```diff
@@ -585,7 +585,7 @@ msgid "ANSWERS ONLY CHECKED -- ANSWERS NOT RECORDED"
 msgstr ""
 
 #: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm:1483
-msgid "ATTEMPT NOT ACCEPTED -- Please submit answers again (or request new version if neccessary)."
+msgid "ATTEMPT NOT ACCEPTED -- Please submit answers again (or request new version if necessary)."
```
drgrice1: The `webwork2.pot` file should not be directly modified.  It is generated by the `bin/dev_scripts/update-localization-files` script.  Once the string is corrected in the code, then this will be fixed when that script is run again.

drgrice1:
This was a fine pull request. Why did you close it?

jeis4wpi:
Okay I will reopen.


  *
John

jeis4wpi:
I found other typos in the repo.  Happy to propose more corrections with another pull request


  *
John

bin/dev_scripts/run-perltidy.pl
```diff
@@ -66,7 +66,7 @@ =head1 OPTIONS
 my %options;
 my $err = Perl::Tidy::perltidy(argv => \@args, dump_options => \%options);
 exit $err                                               if $err;
-die "The -pro option is not suppored by this script.\n" if defined $options{profile};
+die "The -pro option is not supported by this script.\n" if defined $options{profile};
```
drgrice1, line 69: Due to the addition of a letter on line 69, another space is needed on line 68 to make the `if`s line up.  That is why the perltidy action is failing.

jeis4wpi:
Should I change or revert this file?


  *
John

drgrice1:
Fix the spacing. The spelling error should be fixed.

jeis4wpi:
I will repair the space soon.  Thanks for the pertidy education.  I will get better at this.


  *
John

drgrice1:
Ideally you would have your text editor set up to automatically run perltidy anytime you save a Perl file. That way you don't have to think about it.

lib/WeBWorK/Utils.pm
```diff
@@ -610,7 +610,7 @@ Returns C<$float> rounded to C<$places> decimal places.
 
 Usage: C<cryptPassword($clearPassword)>
 
-Returns the crypted form of C<$clearPassword> using a random 16 character
+Returns the encrypted form of C<$clearPassword> using a random 16 character
```
drgrice1: This is not actually a typo.  Crypted is a slangy way of saying encrypted.  Also, the Perl `crypt` method is used to encrypt the password, and so crypted is the past tense of that method name.

jeis4wpi: Thank you for fixing.

Alex-Jordan (approved):
This looks good, although I have no opinion on whether "encrypted" should go back to "crypted". I'm marking as approved so we can get this in, and any future typo fixes could be on a smaller PR.

drgrice1:
Everything else looks good.

dlglin:
I think that it should remain as `crypt`, as that is the name of the underlying function.

PR #2806, opened 2025-09-08
Fix an error that makes the `NoReducedCred` achievement item unusable.

drgrice1:
The `can_use` method of the `NoReducedCred.pm` achiievent item returns 0 unless the condition for which it can be used is true, and then there is no follow up return value for the case that the condition is true.  As a result the method always returns a false value.  Thus the achievement item can never be used.

Since the result of the `can_use` method is used in a purely boolean fashion it should just return the condition result.


somiaj:
Since this item is broken, this should probably be a hot fix.

drgrice1:
I though that would be requested.  It does make sense to hotdog this.

Alex-Jordan (approved):

PR #2811, opened 2025-09-09
Fix checkSet to honour the ignoreMissingSourcedID setting

dlglin:
Currently if `$LTI{v1p3}{ignoreMissingSourcedID}` is set to 1, students are able to click on a set on the Problem Sets page if the `lis_source_did` is not set, but when they click on the set they get an error because `checkSet` does not check for that setting.  This fixes that.


drgrice1 (approved):
Looks good to me.

somiaj (approved):
I can't test this, but code looks obvious.

Issue #2814, opened 2025-09-09
Test "next page" and "previous page" navigation


Alex-Jordan:
In a test, what might we do to make the existence of other pages more apparent to students? Would big fat "next page" and "previous page" buttons placed somewhere be helpful?

This and #2813 are motivated by Runestone users. To avoid Runestone courses with many students starting a quiz with many questions that are all on one page, all at once, I've set the cap so that at most 2 exercises can appear on a page at a time. The occasional student is still submitting the test after only answering questions on the first page.

dlglin:
I think that part of the problem is that page navigation is links and "submit" is a button.  Students look for a button after entering an answer, so I also think having buttons to navigate between pages would help.

I think a well thought-out layout of the buttons will also help. i.e. physically separate the submit button - put it below and separate from everything else to try to make it clear that it's a final action.

Would it help to change the text of the submit button from "Grade Test" to something that clearly states that you are submitting all answers?  My first thought is "Grade ENTIRE Test" or "Grade ALL Questions", but there is probably better wording.

PR #2818, opened 2025-09-15
Add previous and next buttons to tests.

drgrice1:
The buttons go to the next or previous page.  If there is only one problem per page, then the buttons are "Previous Problem" and "Next Problem" buttons.  Otherwise they are "Previous Page" and "Next Page" buttons.  These buttons are shown below the page/problem navigation links and like those links are shown both at the top and bottom of the page.  This means that these buttons are above the "Grade Test" button at the bottom of the page, making it much clearer to students that there are more problems than those shown on the page.

Note these buttons are not shown if all problems are on one page.

This was requested in issue #2814.

Also fix a couple of HTML validation issues that I observed.

First, the "preview answer" buttons had a `data_page_number` attribute. The underscores are not the valid format for data attributes.  They should be hyphens. I.e., kebab case, not snake case.  This was caused by using the incorrect format for the data attribute argument to the `link_to` method.  This was my fault and was done when I initially converted to Mojolicious.

The other issue only occurs when the number of problems in the test is not evenly divisible by the number of problems per page.  This resulted in the wrong number of columns in the problem/page navigation links table for the `colgroup` definition of the last column.

### DIFF
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -448,9 +448,10 @@
 		% end
 		% if ($numProbPerPage && $numPages > 1) {
 			% content_for 'gw-navigation-cols' => begin
-				% for (1 .. $numPages) {
+				% for my $pageNum (1 .. $numPages) {
 					<colgroup class="page">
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Looks good.

PR #2819, opened 2025-09-15
Add a message to the grade test confirmation dialog when questions are unanswered.

drgrice1:
When the "Grade Test" button is clicked, JavaScript parses the `probstatus` inputs and determines what problems are in the test. Then it finds the answer inputs in the page and determines if there are any questions in a problem that are unanswered.  If there are any problems with unanswered questions, then a message stating that is added to the confirmation dialog.

This is to address the feature request in issue #2183.

Generally, this will work for most answers.  However, there are a couple of cases that don't work right.  These are cases where answer inputs are non-empty even if a student has not actively entered an answer.

Most notable is the case of an answer to a `draggableProof.pl` or `draggableSubset.pl` question. The JavaScript for those always fills in a non-empty answer as soon as the problem is loaded into the page.

Another case is if a problem uses the deprecated `PopUp` method from the `parserPopUp.pl` macro or the deprecated methods from `PGchoicemacros.pl`.  For example, if `PopUp([ '?', 'a', 'b', 'c' ], 1)` is used.  That will have the non-empty value `?` for the default answer.

Unanswered questions that use the `draggableProof.pl` and `draggableSubset.pl` macros could be determined server side, but the only way to do so would be to process the problem via PG.  There were several suggestions in issue #2183 to use a server side set "flag" or have other server queries involved in this.  Those really don't help other than in this case, and generally shouldn't be done.  We really do not want to have a submission to the server that processes all problems in a test to determine if there are unanswered questions before submitting a test for grading in which all problems will again be processed.  There was a suggestion to change the "preview test" button into a "save answers" button, but even that doesn't really help.  It would not be reliable, because the information would only be there if the button were used first. There is no guarantee it would be used first, and this needs to have the ability to process the answers regardless.

Note that attempting to correctly determine if a question that uses `parserPopUp.pl` or `PGchoicemacros.pl` is unanswered cannot even be done by processing the problem.  This just adds emphasis that those should not be used anymore.

There may be some other cases that have non-empty answers without students actively answering a question, but I can't think of any at this point.

One thing to note is that if a problem has never even been displayed on the page (for example a `draggableProof.pl` problem that is on a page of the test the student never goes to) this approach will work correctly even for the above cases.


somiaj (approved):
Works under my tests, messages seem reasonable.

Hopefully this helps improves the situation, but people will still hit yes without reading. Mostly random thought, would changing the order of the buttons so No is first and Yes is second help? Though, I guess that would only really change keyboard users having to hit tab one more time.

drgrice1:
You do have to have `attempts_per_version` greater than zero.  The dialog is not new, this just changes the message in the dialog.  Do you have any console errors?

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -718,6 +718,15 @@
 							: maketext(
 							'This is your only submission.  If you say yes, then your answers will be final, '
 								. 'and you will not be able to continue to work this test version.'
+							),
```
Alex-Jordan, line 721: I'm thinking that the alert that you have unanswered questions should come first, before the alert about it being your only/last submission. Given how people (don't) read things, it's easy for me to see a student reading the alert about using up their last submission and feeling confident, then clicking to grade. Without reading the alert about the unanswered questions. If they appeared in the other order, I don't see that sort of thing happening. If they only read the first part that they have unanswered questions, and they are still confident they want to submit anyway, it won't matter to them that it's the only/last submission.

drgrice1: I moved the part about unanswered questions to the top.

Issue #2820, opened 2025-09-17
Current values not honored when set/problem settings are changed on the problem set detail page.


drgrice1:
This is a rather serious regression that is caused by changes in pull request #2735.  That pull request changed the inputs that were previously text inputs into number inputs using place holders for those textual type values.  However, it also sets the value to the actual textual value in those cases.  However, browser's do not ever submit textual values ever for number inputs.  The result is that if one of those textual type values is not the default value for the field, then it will be set to the default value anytime the form is submitted.

For example, the "Show Me Another" field has two textual values.  Those are "Never" and "Course Default", corresponding to the numeric values of -1 and -2, respectively.  The default for this happens to be -1 or "Never" (that is really wrong though, it should be -2 or "Course Default").  So if the value for a problem is set to -2 (i.e., "Course Default"), then when the problem set detail page is loaded the place holder reads "Course Default" as it should, and the value of the input from the server is the text "Course Default".  So now click the "Save Changes" button without making any changes.  The browser submits the empty string for that parameter since the original value from the server is not valid for the number input (in fact the browser immediately changes the invalid text value from the server to the empty string as soon as the page loads).  So the server now gets the empty string from the form submission and then uses the default value for the field.

One of the worst manifestations of this is if you have all of the problems set for show me another to use the course default, and then you select the option to add a problem to the set, and then save changes.  Now all of the problems in the set have show me another disabled (set to "Never").

There are several possible fixed, but one must be made and this certainly needs a hotfix.

One possible fix is to try to make the number type input work.  It may work that anytime the server receives an undefined value and the current value is one of these textual type values, then the setting is not change.  This is probably the ideal way to go, but I am not entirely sure that it will fully work.  I haven't tried to implement it yet.

Another possible fix is to stop using these textual values at all.  Instead just show -2 to mean "Course Default" and -1 to mean "Never", and count on the help for the input being sufficient to inform the user what that means.  This is certainly the least transparent way to go.

The only other fix I can think of is to go back to how it was before with a text input and actual values of "Course Default" or "Never" again.  This way we know works, but it is not exactly ideal.

somiaj:
Funny enough I was just noticing this, and thought I was going mad as the values of 'show me another' kept changing.

I don't have a strong feeling about which one to use, but the final one with just using a text input is fine by me, I always type in the number and don't use the arrows/scrolling to change the number. Maybe if we revert to just a text input, put some error checking so it only accepts integer values greater than or equal to -2.

[...]

PR #2822, opened 2025-09-17
Revert number inputs for fields with labels on the problem set detail page. (hotfix)

drgrice1:
This is a quick fix that will make things function for now, and is good for a hotfix.  This just reverts all of the fields that were switched to number inputs and that have labels for special negative values back to text inputs and the functionality prior to WeBWorK 2.20.

The fields that do not have those labels can still be number inputs. That will work fine for those.

This resolves issue #2820 for the current release.

We can try to do better for the next release, but I don't see any other easy way to fix this that is appropriate for a hotfix.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -480,7 +480,7 @@ use constant FIELD_PROPERTIES => {
 	},
 	max_attempts => {
 		name     => x('Max Attempts'),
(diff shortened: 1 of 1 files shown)


drgrice1:
Note that this also does not set the `placeholder` for a number input to the value.  That was a bad idea.  The idea was that it would be one of these labels.  However, it didn't only set the placeholder in that case.  It also set the placeholder to a numeric value.  So if the value was 3, and then you delete the 3 in the input, the input still shows 3.  That is really confusing.

somiaj (approved):

PR #2836, opened 2025-10-29
Add the hidden authentication fields to the LTI update form.

drgrice1:
The content generator `hidden_authen_fields` method should be called in all page forms, but was not added ot the form in the LTI update page. As a result if you are acting as a student, go to that page, and click "Update Grades", then the acting ceases. Even worse, if `session_management_via` is set to "key", and you click "Update Grades", then you are sent to the login page, and have to login again in order for the form to submit.  I missed this when this page was created. @somiaj observed the issue when acting as another user on the page.


somiaj:
Due to the session management via key issue, should this be a hot fix as well?

PR #2837, opened 2025-10-29
Make MathJax show errors for bad TeX.

drgrice1:
There have been requests to either remove this extension or at least make it so that those editing problems do not have it loaded, as it makes it easier to determine what is wrong with TeX in a problem.

This pull request makes it so that these errors are only shown for users that have the `view_problem_debugging_info` permission (which is those with the ta role or higher by default).  So students will still see the usual faulty TeX code in yellow.

Another option would be to just remove the `noerrors` MathJax extension instead of optionally loading it. I am not sure why this package was added. It seems that I added it when I upgraded from MathJax version 2 to version 3, but I don't remember why.  Perhaps it was just in the configuration that @dpvc recommended, or maybe I added it for some reason.  Perhaps it was just about maintaining compatibility with version 2 of MathJax.  For version 2 (as I understand it), the `noerrors` extension was included by default, but with MathJax version 3 it must be explicitly loaded.

Even if we decide to drop the `noerrors` extension, the change from the `webwork_url` to the `webwork_js_config` method in the `WeBWorK::ContentGenerator` module should be made anyway.  There is also a `webwork_url` method in the `Mojolicious::WeBWorK` module that is already available for all controller modules (since it is a Mojolicious helper method), and having this other one overrides that one and it is confusing to have both that return almost the same value.  The only difference is that `WeBWorK::ContentGenerator` method called the `location` helper which returns the empty string if the root URL is '/', and the `webwork_url` helper returns '/' in that case. I don't know what I was thinking creating the `WeBWorK::ContentGenerator` method which was really just an alies for the `location` helper method anyway.

### DIFF
--- a/htdocs/js/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/MathJaxConfig/mathjax-config.js
@@ -1,6 +1,6 @@
 if (!window.MathJax) {
 	window.MathJax = {
-		tex: { packages: { '[+]': ['noerrors'] } },
+		tex: { packages: { '[+]': webworkConfig?.showMathJaxErrors ? [] : ['noerrors'] } },
 		loader: { load: ['input/asciimath', '[tex]/noerrors'] },
 		startup: {
 			ready() {
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -32,6 +32,7 @@ use MIME::Base64;
 use Scalar::Util qw(weaken);
 use HTML::Entities;
 use Encode;
+use Mojo::JSON qw(encode_json);
 
 use WeBWorK::File::Scoring qw(parse_scoring_file);
 use WeBWorK::Localize;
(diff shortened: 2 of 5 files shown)


htdocs/js/MathJaxConfig/mathjax-config.js
```diff
@@ -1,7 +1,6 @@
 if (!window.MathJax) {
 	window.MathJax = {
-		tex: { packages: { '[+]': ['noerrors'] } },
-		loader: { load: ['input/asciimath', '[tex]/noerrors'] },
+		loader: { load: ['input/asciimath'] },
```
dpvc: Note that it is OK to load `[tex]/noerrors' even if you don't add it to the `packages` array (it is small and so it low cost).  That would mean you can use the

``` js
		tex: { packages: { '[+]': webworkConfig?.mathjax_show_errors [] : ['noerrors'] } },
		loader: { load: ['input/asciimath', '[tex]/noerrors'] },
```

instead of the new lines 145-149.

drgrice1: Okay.  I will switch it to that.  Although it would become more complicated if there were other packages that need to be added to the `packages` array.  At this point there aren't any, so that can be dealt with later if it ever happens.

PR #2838, opened 2025-10-29
Make MathJax show errors for bad TeX. (alternate approach)

drgrice1:
There have been requests to either remove this extension or at least make it so that those editing problems do not have it loaded, as it makes it easier to determine what is wrong with TeX in a problem.

This pull request just removes the `noerrors` MathJax extension. So TeX errors are shown for all users.

I am not sure why this package was added. It seems that I added it when I upgraded from MathJax version 2 to version 3, but I don't remember why.  Perhaps it was just in the configuration that @dpvc recommended, or maybe I added it for some reason.  Perhaps it was just about maintaining compatibility with version 2 of MathJax.  For version 2 (as I understand it), the `noerrors` extension was included by default, but with MathJax version 3 it must be explicitly loaded.

The change from the `webwork_url` to the `webwork_js_config` method in the `WeBWorK::ContentGenerator` module is not needed for this, but should be made anyway.  There is also a `webwork_url` method in the `Mojolicious::WeBWorK` module that is already available for all controller modules (since it is a Mojolicious helper method), and having this other one overrides that one and it is confusing to have both that return almost the same value.  The only difference is that `WeBWorK::ContentGenerator` method called the `location` helper which returns the empty string if the root URL is '/', and the `webwork_url` helper returns '/' in that case. I don't know what I was thinking creating the `WeBWorK::ContentGenerator` method which was really just an alies for the `location` helper method anyway.

### DIFF
--- a/htdocs/js/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/MathJaxConfig/mathjax-config.js
@@ -1,7 +1,6 @@
 if (!window.MathJax) {
 	window.MathJax = {
-		tex: { packages: { '[+]': ['noerrors'] } },
-		loader: { load: ['input/asciimath', '[tex]/noerrors'] },
+		loader: { load: ['input/asciimath'] },
 		startup: {
 			ready() {
 				const AM = MathJax.InputJax.AsciiMath.AM;
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -32,6 +32,7 @@ use MIME::Base64;
 use Scalar::Util qw(weaken);
 use HTML::Entities;
 use Encode;
+use Mojo::JSON qw(encode_json);
 
 use WeBWorK::File::Scoring qw(parse_scoring_file);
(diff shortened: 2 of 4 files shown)


dpvc:
The original reason for the `noerrors` extension is that, for most people reading the page, the TeX errors are of no use at all, and if they can see the original TeX, they at least have some chance of understanding what the author had originally intended, even if it contains errors.

The TeX errors are really only useful for page authors, and the other PR at least tried to handle that by using the permissions to determine who sees them.  It would be better if the PG editor could control the presentation of the messages, rather than trying to use the permission as a proxy for that, or always showing errors.  I haven't looked at the editor code in decades, but I wonder if it could insert some javascript to set the needed value to prevent the `noerrors` package from being used.

> Perhaps it was just in the configuration that @dpvc recommended

You probably used the v2-to-v3 configuration conversion tool to convert the old configuration to the new one.  Since the v2 default has `noerrors`, it would have kept that in the new configuration that you obtained.

Alex-Jordan (approved):

PR #2842, opened 2025-11-08
Changes needed to accomodate changes to the PG POD and sample problem generation.

drgrice1:
First, there are some minor tweaks to the POD and sample problem viewers:

* Fix the wording on the sample problem viewer index page for "Sample Problems by Macro".  It current says, "For many macros, this lists all sample problems used by the macro."  However, macros do not use sample problems.  Sample problems use macros.  So the wording is now, "For many macros, this lists all sample problems that use the macro."

* Update the version of bootstrap in the POD templates to 5.3.7 which is currently the latest.

* Update the links in the pod templates to wiki.openwebwork.org.

* Remove the `favicon` links.  Those are not explicitly needed.  Browser's now load the favicon even if the page does not contain the link. The link only needs to be added if the file is not named favicon.ico.

The following are the actual changes needed to accomodate changes to the PG POD and sample problem generation:

* Remove files that are now used from the PG repository.

* Remove the PODViewer js and css files, the PODParser.pm and PODtoHTML.pm modules, and category-index.mt and pod.mt templates, all of which are now provided by PG.

* Make the necessary changes to the generate-ww-pg-pod.pl script so that it can use those files from the PG repository.

* Update the SampleProblemParser module usage for its changes and namespace change.

* The `WeBWorK::PG::SampleProblemParser` module now handles generation of the actual search data with its `getSearchData` method.  So the `WeBWorK::ContentGenerator::SampleProblemViewer` just calls that method.

* The `assets/stop-words-en.txt` file is now in the PG repository so that the `WeBWorK::PG::SampleProblemParser::getSearchData` method can get at it easily.

* The `htdocs/js/SampleProblemViewer/documentation-search.js` file is now in the PG repository, and webwork2 uses it from there.  PG also copies it to the PG docs github pages site in its new workflow. The options are passed in via the script in `templates/ContentGenerator/SampleProblemViewer.html.ep`.

Note that this depends on https://github.com/openwebwork/pg/pull/1346.


somiaj (approved):
POD and Sample problem generation works without issue.

PR #2844, opened 2025-11-08
Add permissions to render problems with WebworkWebservice.

somiaj:
First, this adds the permission `webservice_render_problem` used to determine if a user can render a problem with the WebworkWebservice, instead of using the `proctor_quiz_login` permission for this.

Second, this adds an additional permission `webservice_render_source` used to determine if a user can render problems using the problem provided with the request. The use case for this is to allow users which can render problems only using a problem filename, but not by providing the problem's source.

These permissions are both set to `login_proctor` to match current behavior and are provided to allow server admins to change which users can render problems. These permissions are not added to the course configuration page as they are permissions that should not be modified by most users, only server admins via `localOverrides.conf` or `course.conf`.

One way to test this is to set `webservice_render_source` to `nobody`, and see that pages like the library browser still work since rendering is done via filenames, but the problem editor no longer works.

### DIFF
--- a/lib/WebworkWebservice.pm
+++ b/lib/WebworkWebservice.pm
@@ -257,7 +257,7 @@ sub command_permission {
 		convertCodeToPGML => 'access_instructor_tools',
 
 		# WebworkWebservice::RenderProblem
-		renderProblem => 'proctor_quiz_login',
+		renderProblem => 'webservice_render_problem',
 
 		# WebworkWebservice::SetActions
 		listGlobalSets        => 'access_instructor_tools',
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -780,6 +780,13 @@ $authen{admin_module} = ['WeBWorK::Authen::Basic_TheLastOption'];
 	modify_tags                    => "admin",
 	edit_restricted_files          => "admin",
 
+	# Permission to render problems using the WebworkWebservice.
+	# Users with only webservice_render_problem can render problems with a provided filename.
+	# Users with both permissions can also render problems with providing the problem source.
+	# Note the Problem Editor requires having both permissions.
+	webservice_render_problem      => "login_proctor",
+	webservice_render_source       => "login_proctor",
+
 	##### Behavior of the interactive problem processor #####
 	show_correct_answers_before_answer_date         => "ta",
 	show_solutions_before_answer_date               => "ta",
--- a/lib/WebworkWebservice/RenderProblem.pm
+++ b/lib/WebworkWebservice/RenderProblem.pm
@@ -26,6 +26,14 @@ async sub renderProblem {
 	# is enabled.  That is an expensive method to always call here.
 	debug(pretty_print_rh($rh)) if $WeBWorK::Debug::Enabled;
 
+	# If the problem source is provided, check user is allow to render problem source.
+	if (!$ws->authz->hasPermissions($rh->{user}, 'webservice_render_source')
+		&& ($rh->{problemSource} || $rh->{rawProblemSource} || $rh->{uriEncodedProblemSource}))
+	{
+		$ws->error_string(__PACKAGE__ . ": User $rh->{user} does not have permission to render problem source.");
+		return {};
+	}
+
 	my $problemSeed = $rh->{problemSeed} // '1234';
 
 	my $beginTime = Benchmark->new;


lib/WebworkWebservice.pm
```diff
@@ -108,6 +109,12 @@ async sub rpc_execute {
 	return $self->error_string(__PACKAGE__ . ": User $user_id does not have permission for the command $command")
 		unless $c->authz->hasPermissions($user_id, $permission);
 
+	# If rendering a problem and problem source is provided, check user is allow to render the provided source.
+	return $self->error_string(__PACKAGE__ . ": User $user_id does not have permission to render problem source.")
+		unless $command ne 'renderProblem'
+		|| $c->authz->hasPermissions($user_id, 'webservice_render_source')
+		|| !($inputs_ref->{problemSource} || $inputs_ref->{rawProblemSource} || $inputs_ref->{uriEncodedProblemSource});
+
```
drgrice1: I don't like this code being here.  This should either be done in the `pre_header_initialize` method of the `WeBWorK::ContentGenerator::RenderViaRPC` package, or at the beginning of the `renderProblem` method in the `WebworkWebservice::RenderProblem` package.  The code here should not be dependent on the command that is being called.  Everything that is needed to do this is available in both of the above locations, and can be done in a relatively equivalent way in both of those locations.

Also, don't use `unless` for this.  `unless` should only be used for simple conditions, and definitely never in the case that the conditional involves a negation anywhere.  If the conditional involves a negation, then the `unless` statement becomes a double negative, and that makes the code harder to interpret.  This is a case where `if` should be used.  This is easier to understand as
```perl
	if $command eq 'renderProblem'
		&& !$c->authz->hasPermission($user_id', 'webservice_render_source')
		&& ($inputs_ref->{problemSource} || $inputs_ref->{rawProblemSource} || $inputs_ref->{uriEncodedProblemSource});
```

drgrice1: Also note that both of those code locations will be executed if and only if the command is `renderProblem`.

drgrice1: In case it helps, here is code that does the same thing in the `renderProblem` method of the `WebworkWebservice::RenderProblem` package (other than changing the package name in the error string):
```perl
	# If the problem source is provided, check user is allow to render problem source.
	if (!$ws->authz->hasPermissions($rh->{user}, 'webservice_render_source')
		&& ($rh->{problemSource} || $rh->{rawProblemSource} || $rh->{uriEncodedProblemSource}))
	{
		$ws->error_string(__PACKAGE__ . ": User $rh->{user} does not have permission to render problem source.");
		return {};
	}
```
Place that code starting after line 24.

Or here is the code in the `pre_header_initialize` method of the `WeBWorK::ContentGenerator::RenderViaRPC` package:
```perl
	my $user = $c->authen->{user_id};

	# If the problem source is provided, check that the user is allow to render problem source.
	if (!$c->authz->hasPermissions($user, 'webservice_render_source')
		&& ($c->param('problemSource') || $c->param('rawProblemSource') || $c->param('uriEncodedProblemSource')))
	{
		$c->{output} =
			$c->{wantsjson}
			? { error => $c->maketext("User $user does not have permission to render problem source.") }
			: $c->maketext("User $user does not have permission to render problem source.");
		return;
	}
```
Place that code starting after line 53.

Both cases assume that all changes in the `WebworkWebservice` package are reverted except the `renderProblem` command permission.

somiaj: I originally wanted to put this at the start of the `renderProblem` method of `WebworkWebservice::RenderProblem`, but that method returns a hash with the rendered problem and was unsure how to return an error that would be processed and reported from that method without causing other issues. Since this method was already reporting errors, I put it here since I saw how I could gracefully report an error.

So I will move this to the start of the `renderProblem` method, but need to know what to return so an error gets reported correctly.

somiaj: Ahh looks like as long as the ref of the returned command is not 'Future', it gets returned the same way, guess I wasn't exactly sure what the code was doing and didn't want to break things, so I'll just move this code there, and test if it works.

drgrice1: Technically what is returned from the `renderProblem` method is always a `Mojo::Promise`.  You don't get a choice on that.  The `renderProblem` method is async, and so the return value is always wrapped in a promise.   However, when that promise is awaited it resolves to what you do return.   All that matters is that the return value is a reference so that the code at the end of the `rpc_execute` method of the `WebworkWebservice` package gets what it considers a valid result, and then passes it on to the caller.  In this case that is the `pre_header_initialize` method of the `WeBWorK::ContentGenerator::RenderViaRPC` package, and that formats the error string for rendering.

somiaj: Thanks, I initially didn't see that you gave code suggestions, ignore the previous replies, they are from cross communication. I was actually updating the problem when I got this code.

> Place that code starting after line 24.

My original thought was to replace this after line 27, so any debugging messages still get printed? I will move the code to where you suggest. Thanks again.

drgrice1: After line 27 if fine also.

drgrice1 (approved):
I think that separating the `renderProblem` permission from the `proctor_quiz_login` permission is good, and further separating the capability of rendering problem source into another permission is also good.

I think that even better security for the `allow_unsecured_rpc` setting would also be good to implement in the future.  Although perhaps that is a bit silly since that is always going to be unsecured.  At least in the PreTeXt case, there is a user whose credentials are entirely public that is allowed to run untrusted code.  There probably is nothing that can really be done to protect that.

Alex-Jordan:
Local pg rendering could become the norm for static PTX production. And we
could develop a way to use a renderer local to the web server for pretext
live rendering. And then after a year or two of deprecation notices, we
could close the insecure access to a webwork2 server.

Alex Jordan
Mathematics Instructor
Portland Community College

PR #2845, opened 2025-11-10
Remove the label/value hack on the "Set Detail" page.

drgrice1:
This make those fields that have the special word values (such as "Unlimited", "Set Default", etc.) be numeric inputs again.  However, those special word values are separated into a select.  So now there is a select and a number input.  The select  gives the option to choose one of those word values, and it also has a numeric option (labeled appropriately for the field). When that numeric option is selected then the value in the number input is used.

With this change the old "undoLabels" hack is no longer needed. That is the hack that switched the word values back to the numeric values server side.  The select options already have the correct value.  The select also has a special "numeric" value that signals that the number in the number input is to be used instead.

This is an approach to replace the previous number input approach implemented in #2820 and reverted in #2823. Unfortunately that approach had some issues that could not be relegated purely with a number input.

I also noticed that there was an issue when the $test{maxProblemsPerPage} variable is set to 1.  In that case the `problems_per_page` setting would not be shown when editing the global set or editing a set for several users. However, when editing the set for a single user the setting would be shown, although it still couldn't be edited. It doesn't make any sense to show an option that can't be edited for the set as a whole, and isn't even shown in that case, when editing for a single user. In fixing that issue I noticed that the override "none" setting in the `FIELD_PROPERTIES` hash is rather messed up.  See the comment I added on line 84 of the `ProblemSetDetail.pm` file.  That setting is no longer used since I removed the `attempted`, `last_answer`, `num_correct`, and `num_incorrect` fields from the hash that were nonsensically included in that hash with the override "none" and type "hidden" values, which basically meant that those fields were ignored everywhere.

Note the `FIELD_PROPERTIES_GWQUIZ` constant was also remove because it was not used in actuality.  The only field in that hash was the `max_attempts` field, but since it is no included in the `GATEWAY_PROBLEM_FIELD_ORDER` array that hash key was never accessed.

### DIFF
--- a/htdocs/js/ProblemSetDetail/problemsetdetail.js
+++ b/htdocs/js/ProblemSetDetail/problemsetdetail.js
@@ -472,7 +472,7 @@
 	for (const numericSelect of document.querySelectorAll('.mixed-numeric-select')) {
 		const select = numericSelect.querySelector('select');
 		const numberInput = numericSelect.querySelector('input');
-		let currentNumberValue = numberInput.value;
+		let currentNumberValue = numberInput.value !== '' ? numberInput.value : numberInput.min;
 
 		const setNumericState = () => {
 			if (select.value === 'numeric') {
@@ -480,7 +480,7 @@
 				numberInput.disabled = false;
 				numberInput.required = true;
 			} else {
-				currentNumberValue = numberInput.value;
+				if (numberInput.value !== '') currentNumberValue = numberInput.value;
 				numberInput.value = '';
 				numberInput.disabled = true;
 				numberInput.required = false;


Alex-Jordan (approved):
This is good!

somiaj:
Just checked, didn't realize that it is currently possible to allow `0` attempts, so that is nothing new with this PR, but maybe something to consider.

drgrice1:
I went ahead and made it so that when switching from a non-numeric selection to the numeric selection, the number input now gets the minimum value.

pstaabp (approved):
This looks good.

PR #2846, opened 2025-11-10
Add an "accommodation time factor" to provide for extra time on timed tests.

drgrice1:
This accommodation time factor is a user property and is set when editing a student on the "Accounts Manager" page. The time that a student will have to complete any timed test is the product of the "Test Time Limit" set for the test on the "Set Detail" page, and this accommodation time factor for the student taking the test.  By default the accommodation time factor for each student is 1, but can be set to something like 1.5 to determine that a student is allowed time and a half to complete timed tests.

The point of this is that it is a bit tedious to need to go through all timed tests and change the test time limit for all of the students that need to be given extra time for accessibility accommodations. With this you only need to set one number, and it is rather convenient to do so for all of the students in the class that need it from one page.

Since this is a per user setting, this requires the addition of a new column to the user table in the database.

### DIFF
--- a/htdocs/js/ActionTabs/actiontabs.js
+++ b/htdocs/js/ActionTabs/actiontabs.js
@@ -6,6 +6,7 @@
 		actionLink.addEventListener('show.bs.tab', () => {
 			if (takeAction) takeAction.value = actionLink.textContent;
 			if (currentAction) currentAction.value = actionLink.dataset.action;
+			takeAction.formNoValidate = actionLink.dataset.noValidate ? true : false;
 		});
 	});
 
--- a/htdocs/js/UserList/userlist.js
+++ b/htdocs/js/UserList/userlist.js
@@ -58,7 +58,7 @@
 					}
 				} else {
 					element?.classList.remove('is-invalid');
-					if (element.id in event_listeners) {
+					if (element && element.id in event_listeners) {
 						element?.removeEventListener('change', event_listeners[element.id]);
 						delete event_listeners[element.id];
 					}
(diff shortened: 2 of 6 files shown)


Alex-Jordan:
When entering an accommodation time factor, is there any way to protect against invalid input? For example if I enter "1.5a", I get a a perl error.

drgrice1:
I will see what can be done.  It looks like that is going to take quite a bit of restructuring of the user list code to do anything about it though. There isn't currently a mechanism for data validation there.

Alex-Jordan:
A funny thing is that you can use a negative number as well, which is valid from the perspective of the database column type. So no error. But then the test version's close date ends up being earlier than its open date.

drgrice1:
I added validation of the "Accomodation Time Factor".

somiaj:
I wonder if there is any standard on to what the factors can be. The only ones I've seen are 1, 1.5, 2. I suspect those are the most common. Though making a drop down menu that only shows the ones we think might arise probably isn't best. Wonder if there is some standard that we could follow for this.

Alex-Jordan:
Also unlimited time might be a standard option.

Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
I am not sure why we are discussing switching to a drop down.  This works well and covers the possibilities (other than the unlimited time which a drop down won't help with the implementation of), so why make that change?

somiaj:
I was just brainstorming mostly, but didn't think a drop down menu was needed.

Tested the current setup and all is working as expected. Though I would suggest using a step of 0.5 instead of the default of 1, so the up/down arrows to change the number will be slightly more useful in common cases. I don't know if any places use factors of 1.25 or 1.75, but a step of 0.25 could also be useful. I am okay with how it is, but 1.5 is a very common factor in my use case, so having the arrows able to reach that would be convenient but not needed.

drgrice1:
Actually, this does not use the default step of 1. It uses a step of "any".  That is a special step value that allows any decimal number. Unfortunately the arrows perform an increment or decrement of 1, and there isn't any way to change that while allowing any decimal. Note that setting a step of 0.5 means that numbers like 1.25 are not allowed, and that is not what we want.

PR #2853, opened 2025-12-04
Fix the openwebwork.org link in hardcopy.

drgrice1:
The `webwork2.sty` file currently creates this link using the `\url` command.  That command attempts to detect the link type.  In this case that results in a file link.  Depending on the PDF viewer used, that results in different behavior.  In Firefox's PDF viewer it behaves as if it is not a link at all, but just text.  Clicking on it does nothing. In Google Chrome's PDF viewer or Evince it is a link, but clicking on it results in a message being displayed that the file does not exist.

So this switches to using the `\href` command instead. That requires that both the link URL and link text explicitly be given, but reliably results in a working link.


drgrice1:
I added `\nolinkurl` to the link text.  That does basically the same thing that `\texttt` does in this case.  Using `\nolinkurl` is how the hyperref package documents to do this.

Alex-Jordan (approved):

somiaj (approved):

PR #2864, opened 2025-12-18
Fix the MathQuill usage on the problem grader page.

drgrice1:
When MathQuill was switched to a published npm package instead of being used from the Github repository, the `node_modules` location changed. This was all updated on the PG side of things, but I forgot to do it for the problem grader (which is the only place MathQuill is used for webwork2).


somiaj:
I am just going to merge this, it is a tiny targeted obvious fix to use the openwebwork version of mathquill.

PR #2866, opened 2025-12-20
enhanced subject line formatting options for email

Alex-Jordan:
This is intended to improve feedback emails where the subject is like:
```
[WWfeedback] course:courseID user:alex.jordan set: prob: sec: rec:
```
where the `set: prob: sec: rec:` is empty. This would happen if the user triggers the email from the Assignments page where there is no setID or problemID. And if there is no section or recitation assigned to the sender.

Maybe there is a better way, but what this does is let the format string be like:
```
'[WWfeedback] course:%c user:%u%s{ set:}%s%p{ prob:}%p%x{ sec:}%x%r{ rec:}%r'
```
If something like `%s{ set:}` is in the format string, then the string " set:" will be in the subject line, but only if the setID is defined. So the above would make the subject line be:
```
[WWfeedback] course:courseID user:alex.jordan
```
if there is no section, recitation, set, or problem ID.

I believe the change here is backward compatible, so if anyone uses their own format string, nothing changes for them. As long as they were not doing something odd with braces.

### DIFF
--- a/lib/WeBWorK/Utils.pm
+++ b/lib/WeBWorK/Utils.pm
@@ -400,11 +400,9 @@ sub formatEmailSubject ($formatString, $courseID, $userID, $setID, $problemID, $
 	if (@braces) {
 		# for each brace pair, do substitutions, but leave %c etc when variable is empty
 		my %braces = map { $_ => $_ =~ s/%([$chars])/$subject_map{$1} ne '' ? $subject_map{$1} : "%$1"/egr } @braces;
-		# if there is an instance of %c, etc, nullify the whole thing
-		%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : $braces{$_} } keys %braces;
-		# remove outer braces
-		%braces = map { $_ => $braces{$_} =~ s/\{(.*)\}/$1/egr } keys %braces;
-		my $regex = join('|', keys %braces);
+		# If there is an instance of %c, etc, nullify the whole thing. Remove outer braces.
+		%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : substr($braces{$_}, 1, -1) } keys %braces;
+		my $regex = join('|', map {"\Q$_\E"} keys %braces);
 		$regex = qr/$regex/;
 		$subject =~ s/($regex)/$braces{$1}/g;
 	}


lib/WeBWorK/ContentGenerator/Feedback.pm
```diff
@@ -109,16 +109,27 @@ sub initialize ($c) {
...
+			c   => $courseID,
+			u   => $user    ? $user->user_id       : undef,
+			s   => $set     ? $set->set_id         : undef,
+			p   => $problem ? $problem->problem_id : undef,
+			x   => $user && $user->section    ? $user->section    : undef,
+			r   => $user && $user->recitation ? $user->recitation : undef,
 			'%' => '%',
 		);
 		my $chars   = join('', keys %subject_map);
 		my $subject = $ce->{mail}{feedbackSubjectFormat} || 'WeBWorK question from %c: %u set %s/prob %p';
```
somiaj: Seems odd to me that the fallback here is not the same as the default, I wonder if it is worth changing that too.

lib/WeBWorK/Utils/ProblemProcessing.pm
```diff
@@ -383,19 +383,15 @@ sub jitar_send_warning_email ($c, $userProblem) {
...
-		'p' => $problemID,
-		'x' => $user->section,
-		'r' => $user->recitation,
-		'%' => '%',
+	my $subject = formatEmailSubject(
+		$ce->{mail}{feedbackSubjectFormat},
+		$courseID,
+		$user    ? $user->user_id       : '',
+		$set     ? $set->set_id         : '',
+		$problem ? $problem->problem_id : '',
```
drgrice1: This should be
```suggestion
		$setID, $problemID,
```
The variables `$set` and `$problem` are not defined here.

lib/WeBWorK/Utils.pm
```diff
@@ -382,6 +383,33 @@ sub generateURLs ($c, %params) {
...
+	my @braces = $formatString =~ /(\{(?:[^{}]*|(?0))*\})/xg;
+	# for each brace pair, do substitutions, but leave %c etc when variable is empty
+	my %braces = map { $_ => $_ =~ s/%([$chars])/$subject_map{$1} ne '' ? $subject_map{$1} : "%$1"/egr } @braces;
+	# if there is an instance of %c, etc, nullify the whole thing
+	%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : $braces{$_} } keys %braces;
+	# remove outer braces
+	%braces = map { $_ => $braces{$_} =~ s/\{(.*)\}/$1/egr } keys %braces;
+	my $regex = join('|', keys %braces);
+	$regex = qr/$regex/;
+	my $subject = $formatString =~ s/($regex)/$braces{$1}/gr;
```
drgrice1: I am seeing warnings from this line if the `$mail{feedbackSubjectFormat}` string is used with this code.  These warnings will occur for any string that doesn't have any of these brace pairs in it.  The issue is that if a string doesn't have any of the brace pairs, then the `%braces` hash has no key/value pairs, and so the `$regex` will be the regex that matches the empty string string.  Since that matches the empty string between every successive pair of characters in the `$formatString` and `$braces{''}` is undefined, there are warnings.  One fix is to change this line to
```suggestion
	my $subject = $formatString =~ s!($regex)!$braces{$1} // ''!egr;
```
However, that is not the most efficient, because that regular expression will replace the empty string between every successive pair of characters with the empty string.  It would probably be better to restructure the code a bit to skip this regular expression substitution if the `%braces` hash is empty.

lib/WeBWorK/Utils.pm
```diff
@@ -382,6 +383,33 @@ sub generateURLs ($c, %params) {
...
+	);
+	my $chars = join('', keys %subject_map);
+	# extract the brace pairs
+	my @braces = $formatString =~ /(\{(?:[^{}]*|(?0))*\})/xg;
+	# for each brace pair, do substitutions, but leave %c etc when variable is empty
+	my %braces = map { $_ => $_ =~ s/%([$chars])/$subject_map{$1} ne '' ? $subject_map{$1} : "%$1"/egr } @braces;
+	# if there is an instance of %c, etc, nullify the whole thing
+	%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : $braces{$_} } keys %braces;
+	# remove outer braces
+	%braces = map { $_ => $braces{$_} =~ s/\{(.*)\}/$1/egr } keys %braces;
```
somiaj: Any reason you can't just leave the outer braces out from the start to avoid having to removing them later? Removing the braces from the grouping seems to work in my testing.

```perl
	my @braces = $formatString =~ /\{((?:[^{}]*|(?0))*)\}/xg;
```

lib/WeBWorK/Utils.pm
```diff
@@ -382,6 +383,33 @@ sub generateURLs ($c, %params) {
 	}
 }
 
+sub formatEmailSubject {
+	my ($formatString, $courseID, $userID, $setID, $problemID, $section, $recitation) = @_;
```
somiaj: Shouldn't this be more declarative now?

```perl
sub formatEmailSubject ($formatString, $courseID, $userID, $setID, $problemID, $section, $recitation) {
```

drgrice1: Yes.  Use signatures.

lib/WeBWorK/Utils.pm
Alex-Jordan: Suppose the string were something dumb like `%x taco{%x taco}` and there is no value for `%x`. If I immediately drop the braces, the array has `%x taco`. Later we will end up substituting away both instances of `%x taco`. Unless of course we look for braces as well at the place where that substitution happens. So it seemed cleaner to me to not have to remember to bring back the braces later.

somiaj: I see, I missed that you were using the braces to know what to replace in the original string.

Though I still think you can remove this line, and include this above. We know that we just need to remove the first and last character, and only in the case that we actually include the content. So I think line 402 could be (haven't tested).

```perl
%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : substr($braces{$_}, 1, -1) } keys %braces;
```

somiaj: Also looking at this, I know it is inconsistent in the codebase, but I think using sentences (capitalization and periods) on the comments is easier to read.

Alex-Jordan: I've lost track of what was being said here. It must have applied to a version before the current version, because line 402 is now where `%braces` is defined in the first place. But I believe it works now. It's a mild pain to test this email stuff, so unless there is a compelling reason to make more changes, I'd just leave it as it is.

If we have a standard for comment style, I'll edit the new comments to conform to that. There were nearby existing comments that are just lowercase, no periods. So I went with that style for now.

lib/WeBWorK/Utils.pm
```diff
@@ -382,6 +383,35 @@ sub generateURLs ($c, %params) {
...
+	my $subject = $formatString;
+	# extract the brace pairs
+	my @braces = $formatString =~ /(\{(?:[^{}]*|(?0))*\})/xg;
+	if (@braces) {
+		# for each brace pair, do substitutions, but leave %c etc when variable is empty
+		my %braces = map { $_ => $_ =~ s/%([$chars])/$subject_map{$1} ne '' ? $subject_map{$1} : "%$1"/egr } @braces;
+		# if there is an instance of %c, etc, nullify the whole thing
+		%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : $braces{$_} } keys %braces;
+		# remove outer braces
+		%braces = map { $_ => $braces{$_} =~ s/\{(.*)\}/$1/egr } keys %braces;
```
somiaj:
```suggestion
		# If there is an instance of %c, etc, nullify the whole thing. Remove outer braces.
		%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : substr($braces{$_}, 1, -1) } keys %braces;
```

I should learn the suggestion feature better. This is what I was suggesting, saves having to loop through the hash again and since you know the braces area always the first and last character, take advantage of that vs using regex.

I agree the email output is also hard for me to check so haven't fully tested this.

Alex-Jordan: OK, I get it. I'll commit the suggestion here, thanks.

somiaj (approved):

lib/WeBWorK/Utils.pm
```diff
@@ -382,6 +383,33 @@ sub generateURLs ($c, %params) {
...
+		s   => $setID,
+		p   => $problemID,
+		x   => $section,
+		r   => $recitation,
+		'%' => '%',
+	);
+	my $chars   = join('', keys %subject_map);
+	my $subject = $formatString;
+	# extract the brace pairs
+	my @braces = $formatString =~ /(\{(?:[^{}]*|(?0))*\})/xg;
```
drgrice1, line 399: I don't think that the recursive sub-pattern in this is right.  I think it should be removed, and this should be
```suggestion
	my @braces = $formatString =~ /(\{(?:[^{}]*)*\})/xg;
```
The problem with the recursive sub-pattern is that it matches balanced braces which means it will include matched braces inside a matched brace pair.  At first you might think that is correct, but it really isn't.  For example, consider the `$mail{feedbackSubectFormat}` that contains `[WWfeedback] course:%c user:%u{{ set:%s}}{ prob:%p}{ sec:%x}{ rec:%r}`.  With that subject format the recursive sub-pattern makes `@braces` contain `{{ set:%s}}`, `{prob:%p}`, `{sec:%x}`, and `{$rec:%r}`.  Then on line 406 `$regex` is `{{ set:%s}}|{ rec:%r}|{ prob:%p}|{ sec:%x}` and that causes the warning `Unescaped left brace in regex is passed through in regex`.

If you remove the recursive sub-pattern, then your current value for `$mail{feedbackSubjectFormat}` still works, and so does the modified value I gave in the example above.

In any case, there should never be any warnings regardless of what the `$mail{feedbackSubjectFormat}` is set to since it is string that can be set by a user (the instructor of the course in this case).

drgrice1: Another option if you really think the recursive sub-pattern is necessary is to change line 404 to `my $regex = join('|', map { $_ =~ s/([{}])/\\$1/gr } keys %braces);`.  I.e., escape the braces for the regex.

drgrice1: Hmm.  I am seeing other warnings in certain cases as well.  If a brace pair contains some other character that needs to be escaped in a regular expression, then there are still warnings.  For example, if the `$mail{feedbackSubjectFormat}` contains something like `{ [set:$s}`.  That is not very well formed, but could happen, and should not result in warnings.

drgrice1: So here is the best code that I have found so far.  It keeps the sub-pattern recursion and I haven't been able to get warnings from anything so far.
```perl
	my @braces = $formatString =~ /(\{(?:[^{}]*|(?0))*\})/xg;
	if (@braces) {
		# for each brace pair, do substitutions, but leave %c etc when variable is empty
		my %braces = map { $_ => $_ =~ s/%([$chars])/$subject_map{$1} ne '' ? $subject_map{$1} : "%$1"/egr } @braces;
		# If there is an instance of %c, etc, nullify the whole thing. Remove outer braces.
		%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : substr($braces{$_}, 1, -1) } keys %braces;
		my $regex = join('|', map { $_ =~ s/([{}])/\\$1/gr } keys %braces);
		eval {
			$regex = qr/$regex/;
			$subject =~ s/($regex)/$braces{$1}/g;
		};
	}
	$subject =~ s/%([$chars])/$subject_map{$1} ne '' ? $subject_map{$1} : ''/eg;
```
The `eval` prevents the warnings.  Although, it also prevents the brace substitutions from happening if there is something that causes issues with creating the regular expression.

drgrice1: Note I did see a reason to keep the recursive sub-pattern.  If the subject template contains `{{ rec:%r}}` and the recitation is not set, then that entire thing is removed from the end result with the recursive sub-pattern.  However, the outer braces is kept without the recursive sub-pattern.

drgrice1: You could add brackets to the escaping substitution if you want to (so `my $regex = join('|', map { $_ =~ s/([{}[\]])/\\$1/gr } keys %braces);`) if you want to make bad brackets work.  However, the `eval` is still needed, because there are lots of other cases where something in the user defined `$mai{feedbackSubjectFormat}` will cause issues.

drgrice1: Ah-hah.  Even better.  Use
```perl
	my @braces = $formatString =~ /(\{(?:[^{}]*|(?0))*\})/xg;
	if (@braces) {
		# for each brace pair, do substitutions, but leave %c etc when variable is empty
		my %braces = map { $_ => $_ =~ s/%([$chars])/$subject_map{$1} ne '' ? $subject_map{$1} : "%$1"/egr } @braces;
		# If there is an instance of %c, etc, nullify the whole thing. Remove outer braces.
		%braces = map { $_ => $braces{$_} =~ /%[$chars]/ ? '' : substr($braces{$_}, 1, -1) } keys %braces;
		my $regex = join('|', map {"\Q$_\E"} keys %braces);
		$regex = qr/$regex/;
		$subject =~ s/($regex)/$braces{$1}/g;
	}
```
I remembered the `\Q...\E` construct.  That says to disable pattern meta chars for anything inside.

Alex-Jordan: Thanks, that was a pretty significant investigation and I attributed the commit to you.

drgrice1 (approved):
I think this looks good.

I think that it would be nice to add the student sender's first and last name to the available substitutions even if they aren't in the default template line.  I would rather use that than the user id.

PR #2867, opened 2025-12-23
Remove Cosign authentication module

dlglin:
According to Wikipedia CoSign single sign on is no longer being updated.  As far as we can tell this is no longer being used by anyone, so it's just taking up space.


drgrice1 (approved):
You could also remove the Moodle authen module.  That is also obsolete and hopelessly outdated.

somiaj (approved):

PR #2868, opened 2025-12-24
prompt email recipient to use reply-all when there were multiple reci…

Alex-Jordan:
…pients

### DIFF
--- a/templates/ContentGenerator/Feedback/feedback_email.html.ep
+++ b/templates/ContentGenerator/Feedback/feedback_email.html.ep
@@ -11,12 +11,15 @@
 			% } elsif ($set) {
 				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
 				<%= $ce->{institutionName} %>
-			   	(sent from <%= link_to format_set_name_display($set->set_id) => $emailableURL %>)
+				(sent from <%= link_to format_set_name_display($set->set_id) => $emailableURL %>).
 			% } else {
 				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
 				<%= $ce->{institutionName} %> (sent from <%= link_to 'this page' => $emailableURL %>).
 			% }
 		</p>
+		% if ($numRecipients > 1) {
+			<p><%= maketext('Message delivered to multiple recipients. Consider using reply-all.') =%></p>
+		% }
 		% if ($feedback) {
 			<div style="border: 1px solid lightgray; padding: 1rem; margin-bottom: 1rem; border-radius: 0.375rem">
 				% for (split /\n\r?/, $feedback) {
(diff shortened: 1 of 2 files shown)


templates/ContentGenerator/Feedback/feedback_email.html.ep
```diff
@@ -11,12 +11,15 @@
...
 				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
 				<%= $ce->{institutionName} %>
-			   	(sent from <%= link_to format_set_name_display($set->set_id) => $emailableURL %>)
+				(sent from <%= link_to format_set_name_display($set->set_id) => $emailableURL %>).
 			% } else {
 				Message from <%= $user->full_name %> (<%= $user->user_id %>) via WeBWorK at
 				<%= $ce->{institutionName} %> (sent from <%= link_to 'this page' => $emailableURL %>).
 			% }
 		</p>
+		% if ($c->stash->{numRecipients} > 1) {
```
drgrice1: This should be
```suggestion
		% if ($numRecipients > 1) {
```
Stash values are converted into variables when templates are compiled.  The only time you need to be careful and actually get them from the stash is if the stash value might not be defined.  In that case the variable won't exist and the template will fail to compile.  In this case this stash value will always be defined.

templates/ContentGenerator/Feedback/feedback_email.txt.ep
```diff
@@ -12,6 +12,11 @@ Feedback sent from:
 % }
 <%== $emailableURL %>
 
+% if ($c->stash->{numRecipients} > 1) {
+
+<%= maketext('Message delivered to multiple recipients. Consider using reply-all.') =%>
```
drgrice1: This should be
```suggestion
% if ($numRecipients > 1) {

<%= maketext('Message delivered to multiple recipients. Consider using reply-all.') %>
```

The reasoning for the switch from `$c->stash->{numRecients}` to `$numRecipients` is the same as the other comment.

The rest is about spacing.  The empty line before the conditional together with the empty line after it result in two empty lines in the text email.  The `=%>` ending that you had at the end of the template substitution strips whitespace after the substitution which result in the empty line after it being removed.  The result with your code is:

```txt
Feedback sent from setID:
page url here


Message delivered to multiple recipients. Consider using reply-all.
Username (userId) wrote:
```

With my changes it becomes

```txt
Feedback sent from setID:
page url here

Message delivered to multiple recipients. Consider using reply-all.

Username (userId) wrote:
```

Also, in the case that there is only one recipient so that this new message is not shown, then the empty line before and the emtpy line after both are in the output (the `=%>` substition doesn't happen in this case), and so even in that case there are two empty lines.  The above changes cover this case also.

templates/ContentGenerator/Feedback/feedback_email.html.ep
Alex-Jordan: Thanks, I thought I had tried this at first and hit an error. But if I remember that right, I may have just had a typo.

templates/ContentGenerator/Feedback/feedback_email.txt.ep
Alex-Jordan: I confess to not looking at the test email. But I realize now I can just "show original" in Gmail to see that when I'm testing. Thanks for the cleanup!

drgrice1 (approved):

somiaj (approved):

PR #2869, opened 2025-12-24
Allow viewing ProblemSet page for any valid set.

somiaj:
Rework the ProblemSet page so users can always see the set header and additional set information for any set assigned to them. Instead of giving an error message when a user tries to access a set that is not open, lti restricted, conditional release, or ip restricted, checkSet will set a flag which is used to determine what information a user can see.

If a set not visible and a user cannot see hidden sets, only a warning message is shown. In all other cases the set header and any date information, such as when the set opens, closes, etc, is shown. If the set is restricted due to a conditional release, ip restrictions, or lti restrictions, a warning message is shown informing the user of the restriction.

When a set is restricted, never show the set problems and always show any previous taken set versions. This way users can still access set versions they have taken for a restricted set from a non restricted location.

Note, this is consistent with the permission `view_unoppend_sets` description, which states it only configures if a set problems can be seen. So this permission is no longer used to see if a user can see set information such as open date and set header.

This also makes it so the right info panel div on the ProblemSet page is only shown if the set header exists, and is not empty.

Translations were added to messages about IP restrictions.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ProblemSet.pm
+++ b/lib/WeBWorK/ContentGenerator/ProblemSet.pm
@@ -19,18 +19,23 @@ use WeBWorK::Localize;
 use WeBWorK::AchievementItems;
 use WeBWorK::HTML::StudentNav qw(studentNav);
 
+sub can ($c, $arg) {
+	if ($arg eq 'info') {
+		return $c->{pg} ? 1 : 0;
+	}
+	return $c->SUPER::can($arg);
+}
+
 async sub initialize ($c) {
 	my $db    = $c->db;
 	my $ce    = $c->ce;
 	my $authz = $c->authz;
 
-	# $c->{invalidSet} is set in checkSet which is called by ContentGenerator.pm
-	return
-		if $c->{invalidSet}
-		&& ($c->{invalidSet} !~ /^Client ip address .* is not in the list of addresses/
-			|| $authz->{merged_set}->assignment_type !~ /gateway/);
+	# $c->{invalidSet} is set in checkSet which is called by ContentGenerator.pm.
+	# If $c->{viewSetCheck} is also set, we want to view some information unless the set is hidden.
+	return if $c->{invalidSet} && (!$c->{viewSetCheck} || $c->{viewSetCheck} eq 'hidden');
 
-	# This will all be valid if checkSet did not set $c->{invalidSet}.
+	# This will all be valid if the above check passes.
 	my $userID  = $c->param('user');
 	my $eUserID = $c->param('effectiveUser');
 
@@ -106,6 +111,7 @@ async sub initialize ($c) {
 
 	$c->{pg} =
 		await renderPG($c, $effectiveUser, $c->{set}, $problem, $c->{set}->psvn, {}, { displayMode => $displayMode });
+	$c->{pg} = '' unless $c->{pg}{body_text} =~ /\S/;
 
 	return;
 }
@@ -169,10 +175,8 @@ sub siblings ($c) {
 	return $c->include('ContentGenerator/ProblemSet/siblings', setIDs => \@setIDs);
 }
 
-sub info {
-	my ($c) = @_;
-	return '' unless $c->{pg};
-	return $c->include('ContentGenerator/ProblemSet/info');
+sub info ($c) {
+	return $c->{pg} ? $c->include('ContentGenerator/ProblemSet/info') : '';
 }
 
 # This is called by the ContentGenerator/ProblemSet/body template for a regular homework set.
--- a/lib/WeBWorK/Utils/Sets.pm
+++ b/lib/WeBWorK/Utils/Sets.pm
@@ -17,6 +17,7 @@ our @EXPORT_OK = qw(
 	is_restricted
 	get_test_problem_position
 	list_set_versions
+	restricted_set_message
 );
 
 sub format_set_name_internal ($set_name) {
@@ -255,6 +256,53 @@ sub list_set_versions ($db, $studentName, $setName, $setIsVersioned = 0) {
 	return (\@allSetNames, $notAssignedSet);
 }
 
+sub restricted_set_message($c, $set, $status) {
+	my $ce    = $c->ce;
+	my $db    = $c->db;
+	my $authz = $c->authz;
+
+	if ($status eq 'conditional') {
+		my $user       = $c->param('effectiveUser') || $c->param('user');
+		my @restricted = $ce->{options}{enableConditionalRelease} ? is_restricted($db, $set, $user) : ();
+		return '' unless @restricted;
+
+		if (@restricted == 1) {
+			return $c->maketext(
+				'To access this set you must score at least [_1]% on set [_2].',
+				sprintf('%.0f', $set->restricted_status * 100),
+				$c->tag('span', dir => 'ltr', format_set_name_display($restricted[0]))
+			);
+		} else {
+			return $c->maketext(
+				'To access this set you must score at least [_1]% on the following sets: [_2].',
+				sprintf('%.0f', $set->restricted_status * 100),
+				join(', ', map { $c->tag('span', dir => 'ltr', format_set_name_display($_)) } @restricted)
+			);
+		}
+	}
+
+	if ($status eq 'lti') {
+		# Only show this message if unable to view unopened sets or acting as another user.
+		return ''
+			if $authz->hasPermissions($c->param('user'), 'view_unopened_sets')
+			&& (!$c->param('effectiveUser') || $c->param('effectiveUser') eq $c->param('user'));
+
+		if ($ce->{LTIVersion}
+			&& ($ce->{LTIVersion} ne 'v1p3' || !$ce->{LTI}{v1p3}{ignoreMissingSourcedID})
+			&& $ce->{LTIGradeMode} eq 'homework'
+			&& !$set->lis_source_did)
+		{
+			return $c->maketext(
+				'You must access this assignment from [_1] before you can start.',
+				$ce->{LTI}{ $ce->{LTIVersion} }{LMS_url}
+				? $c->link_to($ce->{LTI}{ $ce->{LTIVersion} }{LMS_name} => $ce->{LTI}{ $ce->{LTIVersion} }{LMS_url})
+				: $ce->{LTI}{ $ce->{LTIVersion} }{LMS_name}
+			);
+		}
+		return '';
+	}
+}
+
 1;
 
 =head1 NAME
@@ -353,5 +401,13 @@ reference to an array of names of set versions and whether or not the user is
 assigned to the set.  The list of names will be a list of set versions if the
 set is versioned (i.e., if C<setIsVersioned> is true), and a list containing
 only the original set id otherwise.
+
+=head2 restricted_set_message
+
+Usage: C<restricted_set_message($c, $set, $status)>
+
+Checks for and returns any restricted messages for the given set and status.
+C<$status> can be either 'conditional' for conditional release, or 'lti' to
+check for lis_source_did when using homework grade passback.
 
 =cut
--- a/templates/ContentGenerator/ProblemSet.html.ep
+++ b/templates/ContentGenerator/ProblemSet.html.ep
@@ -1,21 +1,27 @@
 % use WeBWorK::Utils::DateTime qw(before between);
 %
-% if (
-	% $c->{invalidSet}
-	% && ($c->{invalidSet} !~ /^Client ip address .* is not in the list of addresses/
-		% || $authz->{merged_set}->assignment_type !~ /gateway/)
-	% )
-% {
-	<div class="alert alert-danger">
-		<p class="mb-3">
-			<%= maketext(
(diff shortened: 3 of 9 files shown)


templates/layouts/system.html.ep
```diff
@@ -136,11 +136,15 @@
 				<%= content =%>
 			</div>
 			% if ($c->can('info')) {
-				<div id="info-panel-right" class="col-md-4 mb-md-0 mb-2 order-md-last order-first">
-					<div class="info-box bg-light">
-						<%= $c->info =%>
+				% # Only show info div if the content is not empty.
+				% my $info = $c->info;
+				% if ($info !~ /^\s*$/) {
```
drgrice1: This should be
```suggestion
				% if ($info =~ /\S/) {
```
That is the usual check that a string contains non white space characters and is a simpler regular expression.  Benchmarking shows it is generally faster too!

drgrice1: Actually, this change is not right.  The problem is that if this actually happens, then the result is an improperly formatted page.  The check on line 135 above and the case that the info is actually shown really should match up perfectly, and this is not doing that.

What case are you trying to catch here and avoid showing an empty info?  The first thing to check is if it really is a case to be concerned about.  If so, then there are better ways to do this.

somiaj: I missed the other check that made used the `col-8` vs `col-12` style above, but I didn't think it mattered, is it really not formatted correctly if the divs don't use all the columns in the grid? Anyways, if you go to a bad set or a set with an error, `https://server/webwork2/CourseName/BadSetName`, there is an empty gray div floating to the right with nothing in it. I was trying to prevent that from happening. I am unsure if any other pages that use this info div have similar issues where it is just empty.

drgrice1: In my opinion yes. Generally, the info box display needs work regardless of this pull request.  Not only for this case, but for others where the content may be empty.  For example, if your course has an empty `course_info.txt` file, then that also currently results in the empty gray box.  These cases should all not show the info box, and the page should be properly formatted as intended for the layout.

It seems that to make this work right, all modules that provide the info box and such that the info box might be empty should also override the `can` method (as the ProblemSets.pm module already does), and that method should not only return 0 for permissions reasons but for the empty info box case.

Note that there is another case where `$c->can('info')` is called that affects the layout.  That is in the `go` method of the base `ContentGenerator` module (line 113).  That sets the layout for the footer to match the main content (note that some modules override that setting later also).  So that also needs to choose the correct layout setting for these cases.

There is an efficiency issue with the current setup.  The `ProblemSets.pm` is already performing extra file access due to this approach.  The can method is called three times and each time it opens the course info file and reads its contents and compares it to the default.  Then the `info` method opens that file yet again and reads the contents to actually display (assuming the `can` method returned true).  So to do this well with the current setup, the `can` method really should do all of the work and stash the content, but with protections to ensure it doesn't execute inefficiency multiple times, and the `info` method should just use the stashed result.  Care also needs to be taken for the `ProblemSet.pm` module, because that doesn't know if it has content until after the `initialize` method runs (that is when the set header is currently rendered), and that runs after the current `can` call in the `go` method of the base `ContentGenerator` module.

To simplify all of this I think that really we should just do away with the info box entirely.  Students really should just be left in the dark anyway, and don't really need or even want instructions! :smiling_imp:

somiaj: Yea, the fact that `$c->can('info')` is called before `$c->initialize` does make things harder to just override the can method. Any way that call in the `go` method could be moved until after the `initialize` method? I found something that appears to work, I'll just update the setting for the `footerWidthClass` on a future `can` call which appears to work. Patch incoming.

drgrice1: The call in the `go` method could be moved after the `initialize` method.  It would just need to use `//=` instead of `=`.

somiaj: Any preference to moving the method vs updating the value of `footerWidthClass` at the end of the `initialize` method as I am currently doing?

drgrice1: If you are just updating it with the same value that the `go` method assignment uses based on the `can` call result, then I would prefer moving the `go` method assignment.

somiaj: Updated, the method has now been moved.

lib/WeBWorK/ContentGenerator/ProblemSet.pm
```diff
@@ -161,9 +166,7 @@ sub siblings ($c) {
 	return $c->include('ContentGenerator/ProblemSet/siblings', setIDs => \@setIDs);
 }
 
-sub info {
-	my ($c) = @_;
-	return '' unless $c->{pg};
+sub info ($c) {
```
drgrice1, line 178: I think that the line `return '' unless $c->{pg};` should remain in this method.  In the template where this is used the `can` method is called before it, but that would not necessarily need to be done.  If this method were called without the prior `can` call, then there would be issues.

somiaj: Okay, added the `return ''` back.

somiaj:
@drgrice1 I updated it to show a warning message about accessing the set via the LMS in the case you described. I don't have a way to test LTI in my development setup, so I'm mostly 'guessing' it works (sorry about that, I just don't have time to setup a way to test that at the moment).

somiaj:
I am not quite sure if the links should be active and/or the message should be shown on the `ProblemSets` page in this case. I will think about that a bit more (unless you have a suggestion), but currently the set info should be shown with a warning, and they won't be able to start any test version or access any homework problems.

templates/ContentGenerator/ProblemSet.html.ep
```diff
@@ -1,28 +1,53 @@
...
 <%= include 'ContentGenerator/Base/set_status', set => $set =%>
 %
+% # Shows warning about restricted IP settings.
+% if ($c->{viewSetCheck} && $c->{viewSetCheck} eq 'restricted') {
+	<div class="alert alert-warning"><%= $c->{invalidSet} %></div>
+% }
+% # Show a warning that the user must access the set via their LMS before they can start.
+% if ($c->{viewSetCheck} && $c->{viewSetCheck} eq 'lti_restricted') {
+	<div class="alert alert-warning">
+		<%= maketext('You must access this set via your Learning Management System ([_1]) before starting.',
```
drgrice1: This needs to be
```suggestion
		<%== maketext('You must access this set via your Learning Management System ([_1]) before starting.',
```
Otherwise the link is displayed literally as `<a href="https://myschool.edu/lms/">the LMS</a>` instead of being a link.

drgrice1 (commented):
I still really don't like this pull request as it is.  I guess I just really don't like that this is a permission at all.  I think we need to decide if set headers really need such a permission.  I really don't think there is such a need, and I think this is just yet another rather confusing permission.

The addition of this permission also adds a bunch of redundancy and now out of sync wording for text shown in different cases.  The wording on the problem sets page if a student is missing the requisite grade pass back data and the user does not have this new permission is `You must log into this set via your Learning Management System ([_1]).`, but the new wording if the user does have this new permission is `You must access this set via your Learning Management System ([_1]) before starting.`  Those should use the same wording.  Then there is also the message that is in the `Authz::checkSet` method that says `You must use your Learning Management System ([_1]) to access this set.  Try logging in to the Learning Management System and visiting the set from there.` and is shown if the user finds and uses a link that directly enters the set, but doesn't have the new permission.

Ultimately we need to decide if this needs a permission at all.  I vote no.  I don't see the need for the security for set headers.

Alex-Jordan:
All that is to say the goal of making the situation friendlier is good. And
I see it as an improvement just to always show headers.

Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
@somiaj: By the way, you don't need actual LTI authentication to test this for the LTI setup.  Just set up the course so that it uses LTI authentication with homework grade passback, but allow direct login.  That will still show the behavior that needs to be tested here.

drgrice1:
Yeah, I don't think that this was handled the best before, but is how it has been since restricted release sets were implemented.  I think that the important thing is that there are indicators letting the instructor know the restrictions that apply for the set.

somiaj:
I updated it so restricted set messages are not hidden from instructors. I put the message on both the ProblemSets and ProblemSet page.

This includes showing any LTI message (only when acting as a student) that way a instructor can see and inform a student they need to access the set via their LMS first if a student contacts them. The conditional release set message is also shown again (Still requires that the conditions are not met, I didn't change that, but could, though when acting as a student the message should be what a student would see, and most likely when not acting the conditions have not been met so it will be there). Also instructors will see IP restriction messages on sets that have IP restrictions on the ProblemSet page, but not the ProblemSets page.

lib/WeBWorK/Utils/Sets.pm
```diff
@@ -1,3 +1,4 @@
+## Please see file perltidy.ERR
```
drgrice1: Is this comment essential to understanding the code in this file? 🤪

lib/WeBWorK/Utils/Sets.pm
```diff
@@ -255,6 +257,53 @@ sub list_set_versions ($db, $studentName, $setName, $setIsVersioned = 0) {
...
+			if $authz->hasPermissions($c->param('user'), 'view_unopened_sets')
+			&& (!$c->param('effectiveUser') || $c->param('effectiveUser') eq $c->param('user'));
+
+		if ($ce->{LTIVersion}
+			&& ($ce->{LTIVersion} ne 'v1p3' || !$ce->{LTI}{v1p3}{ignoreMissingSourcedID})
+			&& $ce->{LTIGradeMode} eq 'homework'
+			&& !$set->lis_source_did)
+		{
+			return $c->maketext(
+				'You must access this set from your Learning Management System ([_1]) before starting.',
```
drgrice1: I don't like this wording.  Something doesn't seem right with the "before starting" part to me.

lib/WeBWorK/Utils/Sets.pm
somiaj: Unsure how that got in there. What I get for deviating from my usual approach to manually running perltidy.

lib/WeBWorK/Utils/Sets.pm
somiaj: Any suggestions, I felt just saying they must access from the LMS seemed incomplete since they are in the set. They only need to access from the LMS before they work on the homework problems or before they start a test version, and if they are seeing this before the open date, nothing will appear to change (except the message go away) if they access from the LMS. Maybe: "before working on the assignment", though for a test, "before starting a test version", then the message would depend on assignment type? I think telling them about the grade link might be too technical, "to link your grade with your LMS".

I'll try to think of something, I just don't have any ideas at the moment.

drgrice1: Yeah, I am not sure either.  This is why we all try to avoid having our students see this message.

I use the `navigation_allowed` permission so that my students never even see the "Assignments" page.

I think @Alex-Jordan uses LTI 1.3 and accesses all of the sets himself at the beginning of each term.  D2L sends the pass back data for instructors, so a student just has to enter one set and then all sets have what are needed for grade pass back.

somiaj: @drgrice1 @Alex-Jordan What about this? It does make the default 'the LMS' not as good, but I think makes any configuration useful:

"You must access this assignment from [_1] once before you can start it."

somiaj: So using 'the LMS' seems odd there, but 'Canvas', 'Blackboard', 'D2L Brightspace' would all seem reasonable there.

drgrice1: That seems okay to me.  

Perhaps you could use

"You must access this assignment from the external authentication system ([_1]) once before you can start it."

like in the other places where the `$LTI{v1p1}{LMS_name}` is used.  Although that seems to be a little long for in the list on the "Assignments" page.

somiaj: I went with, "You must access this assignment from [_1] before you can start.", a little shorter than my original suggestion and gets the same point across in my view.

somiaj: I can change it if anyone has a better suggestion, though as @drgrice1 mentioned, best to try to find ways to keep students from seeing this message, so probably only going to be an issue for us still using lti 1.1, as lti 1.3, only takes a single student or instructor to access, or you could even disable it altogether.

drgrice1: The `navigation_allowed` permission works with LTI 1.1.  As I mentioned, that is what I use.

Also, LTI 1.3 may not fix anything for Canvas in regard to this.  The issue is that Canvas does not send LTI grade pass back data for an instructor.  I believe that is true even for LTI 1.3.  So at least one student has to enter each set from the LMS.  Perhaps the Canvas test student would work though.  Not sure.

somiaj: I would assume the test student would work. in LTI 1.1 it sends enough info that grade pass back works just fine for the test student, and I would be surprised if that changed in 1.3.

pstaabp (approved):
Looks good.

PR #2872, opened 2025-12-29
Configurable feedback email links show problem grader.

drgrice1:
This adds a new course configuration option that allows the instructor to choose if the problem grader will be open or not when the links sent in a feedback email from a problem are followed.

I am finding that to be rather obtrusive, and not something that I ever want.  I know I am not alone on this.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -81,6 +81,11 @@ $mail{feedbackSubjectFormat} = '[WWfeedback] course:%c user:%u{ set:%s}{ prob:%p
 #  2: as in 1, plus the problem environment (debugging data)
 $mail{feedbackVerbosity}     = 1;
 
+# If this is 1, then the links included in feedback emails when feedback is sent
+# from a problem will open the problem grader when followed. If this is 0, then
+# the problem grader will not be open when those links are followed.
+$mail{linksOpenProblemGrader} = 1;
+
 # Should the studentID be included in the feedback email when feedbackVerbosity > 0:
 # The default is yes
 $blockStudentIDinFeedback = 0;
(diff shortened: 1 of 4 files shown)


Alex-Jordan:
This is fine, and I agree that it's not good how the problem grader can take up so much space that you can't even see the problem when the page loads. Is there any interest in changing the interface for the problem grader?

One thing I wonder about, is how the button to show/hide the grader re-renders the problem. Is that needed for some reason? If you have access to the grader, what if it just always appears, except that it has a collapse mechanism and starts out collapsed?

somiaj:
We have talked about the grader is at the top to be close to the answer table which is no longer there. I agree that with the new feedback the grader should at least be at the bottom.

drgrice1:
I started messing with implementing a collapse approach as you requested.  However, that is not going to work.  The problem render is necessary.  Not only to add the problem grader to the page, but to determine if correct answers should be shown to go with it.  Those are, in my opinion, and essential thing to show with the problem grader, and that can't be done with the collapse approach.  That would mean the correct answers would always be there, and that is certainly not desirable.  Basically, there are server side decisions that are made based on whether the problem grader is to be shown or not, and those decisions can't be made if the problem grader is always shown.

As @somiaj mentioned, we discussed moving the problem grader below the problem in a recent meeting.  Originally that is where I had put it in homework sets, and still is there for tests.  I moved it up in homework sets because that was requested, but that was when we still had the results table, and it made more sense to be closer to that.  Now that the results table is gone, it doesn't need to be above anymore.

Alex-Jordan:
Is it understood that when I suggested "always" showing the grader, I meant only under conditions where a user would currently have access to the Show Grader button. Not literally always.

Also, I didn't exactly mean to request it. But I wanted to start discussing alternative solutions in case we can brainstorm something better. Maybe moving it down is better.

drgrice1:
I realize that, but even in that case we don't "always" want the correct answers to be shown.

Issue #2873, opened 2025-12-29
strange score if I use grader and then submit


Alex-Jordan:
I noticed a few odd things while playing with the problem grader just now. In a course, set up a new assignment (I used the Student Orientation) and set the dates so that it is open. Now go to problem 1, and show the grader. Enter 50%, and click Save. Now click "Problem 1" in the left navigation panel to reload the problem.

Now that I have 50% for the exercise, I would expect to see that down in the `#score_summary`, but I don't. I haven't looked but I assume it's because I haven't attempted the exercise yet. If that's right, it should be easy to change. But next is where things get weird. Leave the answer blank, and click Submit. Now the score summary says I have 67%. Indeed, that is my score if I look elsewhere, like in Grades.

I repeated this on a different exercise with 5 answer blanks, again using the grader to give myself 50%. Then I correctly answered two questions before clicking Submit, so I should have 40% for that submission. But again my final score is set tat 67%. Or if use like 10% when setting my own score, clicking Submit with all empty answers makes 13% as the final score.

Alex-Jordan:
OK, I see this behavior if I check out `main`. Maybe relevant that I pulled including all the main hotfixes.

I don't see it with `develop`. I don't see it with the 2.20 tags. 

Have you tried it on `main`? If not, I will keep digging.

drgrice1:
I originally tested with develop.  However, I tested again now with main (with all hotfixes), and I still don't see the odd scores.

Alex-Jordan:
Alright, I'll start probing config settings.

Alex-Jordan:
OK, I have to stop now for local home stuff. But it's something to do with having a reduced scoring date. The last things I did that brought back the bad behavior are put these in `simple.conf`:

```
$pg{ansEvalDefaults}{enableReducedScoring} = 1;
$LTISendGradesEarlyThreshold = '0.7';
```
which did *not* bring back the bad behavior. But then for the set, set `Reduced Scoring Enabled` to "yes". And then the bad behavior returned.

This is on `main`, but now I don't have reason to suspect it won't also happen with `develop`. I'd keep looking tonight but I have to shut down computer stuff :)

somiaj:
Was mentioned it was fixed, but not closed, so closing.

PR #2875, opened 2025-12-30
Make the problem graders also save the sub_status when needed.

drgrice1:
Basically, anytime that reduced scoring is not enabled or if it is but it is before the reduced scoring period, then the sub_status needs to also be set to the same thing as the status.  Otherwise the "unreduced" score will not be computed correctly.  So the problem graders now take this into account.

This fixes issue #2873.

### DIFF
--- a/htdocs/js/ProblemGrader/singleproblemgrader.js
+++ b/htdocs/js/ProblemGrader/singleproblemgrader.js
@@ -125,6 +125,7 @@
 						version_id: saveData.versionId,
 						problem_id: saveData.problemId,
 						status: parseInt(scoreInput.value) / 100,
+						...(saveData.saveSubStatus === '1' ? { sub_status: parseInt(scoreInput.value) / 100 } : {}),
 						mark_graded: true
 					}),
 					signal: controller.signal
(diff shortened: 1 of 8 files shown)


drgrice1:
Perhaps this should be considered for a hotfiix.

drgrice1:
Also note that achievement item usage does not need a fix here.  Achievement items already set the `sub_status` regardless of the reduced scoring status.

Alex-Jordan:
I set out to test this, starting out by going to the develop branch (on both pg and sebwork2). I wanted to repeat the issue first. For both repos, I ran npm ci, and then restarted webwork2.

But now when I click to Save the score I enter, I get an error message
```
Error saving score.
Unknown server communication error.
```
This is any time I try to use the grader, not just the conditions from #2873.  Are you still able to produce the original issue on develop? Does this other error arise for you?

somiaj:
Do you have a configuration issue with your development server (maybe the rpc isn't trying to talk to the right location?)

drgrice1:
Do you have perlcritic installed?  That is a new dependency due to the merge of the pg critic pull requests.  I tested the develop branch on my remove production server and had the same issue, and the reason was that I did not have perlcritic installed.  On Ubuntu install the `libperl-critic-perl` package.  I don't think you use Ubuntu, so you will need to figure out the package for your distribution (Oracle I beleive?) or install from cpan.  The perl module is `Perl::Critic`.

Note that the server will work without having the package installed, but the webwork web service will fail to compile, particularly the `WebworkWebservice::ProblemActions` module, since it depends on the `WeBWorK::PG::Critic` package which uses the `Perl::Critic` module.

We need to remember to add this to the installation instructions and release notes.

drgrice1:
Note that the `PPI` package is also needed, but if you install `Perl::Critic` that should also be installed.  It is a dependency of `Perl::Critic`.

PR #2876, opened 2026-01-02
Add message to PG critic output if no issues are found.

somiaj:
I found it confusing when no message was given when there were no pg critic violations.

This adds a message that no violations were found. I couldn't think of much to say, so this is simple now, but was thinking maybe it could contain other information to help problem authors even though their problem code doesn't trigger any pg critic violations.

### DIFF
--- a/templates/ContentGenerator/Instructor/PGProblemEditor/pg_critic.html.ep
+++ b/templates/ContentGenerator/Instructor/PGProblemEditor/pg_critic.html.ep
@@ -2,7 +2,11 @@
 %
 <div class="m-3 overflow-auto">
 	<h2><%= maketext('PG Critic Violations') %></h2>
-	% my @pgCriticViolations = grep { $_->policy =~ /^Perl::Critic::Policy::PG::/ } @$violations;
+	% my @pgCriticViolations   = grep { $_->policy =~ /^Perl::Critic::Policy::PG::/ } @$violations;
+	% my @perlCriticViolations = grep { $_->policy !~ /^Perl::Critic::Policy::PG::/ } @$violations;
+	% unless (@pgCriticViolations || @perlCriticViolations) {
+		<p><%= maketext('Congratulations! No PG critic violations found.') %></p>
+	% }
 	% if (@pgCriticViolations) {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Looks good.

templates/ContentGenerator/Instructor/PGProblemEditor/pg_critic.html.ep
```diff
@@ -2,7 +2,13 @@
 %
 <div class="m-3 overflow-auto">
 	<h2><%= maketext('PG Critic Violations') %></h2>
-	% my @pgCriticViolations = grep { $_->policy =~ /^Perl::Critic::Policy::PG::/ } @$violations;
+	% my @pgCriticViolations   = grep { $_->policy =~ /^Perl::Critic::Policy::PG::/ } @$violations;
+	% my @perlCriticViolations = grep { $_->policy !~ /^Perl::Critic::Policy::PG::/ } @$violations;
+	% unless (@pgCriticViolations || @perlCriticViolations) {
+		<p><%= maketext('Congratulations! No PG critic violations found.') %></p>
+		</div>
+		% last;
```
somiaj: @drgrice1 I just realized these aren't needed, as the following two conditionals will fail and the only output after this is the trailing `div`. Any preference on if I should remove lines 9 and 10 and have two more if checks?

drgrice1: Yeah, please remove the `</div>` and `% last;` lines.  I really don't like unbalanced code in the templates, and would prefer to avoid it when it can be done.  You added a similar thing to the file `templates/ContentGenerator/ProblemSet/auxiliary_tools.html.ep` in pull request #2869.  I am allowing that for now, but I don't like it.  Here it isn't needed.

drgrice1: The problem I have with this sort of thing is that it can make it difficult to check HTML validity by analyzing the code.  In this case there is a `<div>` that starts, and then there are two corresponding `</div>` endings.  The old theme templates had a ton of this, and it was a nightmare to sort out.

Alex-Jordan (approved):

PR #2877, opened 2026-01-02
Rework the single problem grader interface.

drgrice1:
The problem grader is now always visible for users that have the permission to use it and in the case that they are acting for another user.  This does mean that there is no way to open the problem grader when viewing your own problem.  However, the problem grader is in a collapse.  The state of the collapse is stored in local storage, and whenever you open another problem or change effective users, the collapse goes back to the state that it was in the last time that you had a page open that showed the problem grader.

Correct answers in feedback are now always shown with the reveal button, even when the problem grader is on the page.  However, the reveal button is removed by JavaScript behind the scenes while the problem grader is expanded, and put back if the feedback button is not opened while the problem grader is open.  So if you open a feedback button while the problem grader is open, the reveal button is not shown, and the correct answer is immediately visible.  To summarize the reveal button visibility, the reveal button will not be shown anytime that a feedback button is opened while the problem grader is open, and in that case will never return until the page reloads, but any feedback button that is not opened while the problem grader is open will still show the reveal button, and as usual once the reveal button is used, it will never come back until the page is reloaded.

The problem grader is now below the problem in homework sets as it is in tests. With the collapse and the grader always in the page, I really do not want it above the problem as it currently is.

The original reason for the problem grader being on top was so that it would be close to the old results table with the answers.  With that gone, that reason no longer applies.

This is as requested by @Alex-Jordan in #2872, and is an alternate approach to that pull request.

Also remove the code for the `output_hidden_info` method in the `Problem.html.ep` template. This is because the answer to the question `$c->can('output_hidden_info')` is `$c` can't.  There is no such method anywhere in the code anymore.

### DIFF
--- a/htdocs/js/GatewayQuiz/gateway.scss
+++ b/htdocs/js/GatewayQuiz/gateway.scss
@@ -60,7 +60,7 @@ table.attemptResults {
 	border: 1px solid #ddd;
 	border-radius: 3px;
 
-	h2 {
+	h2.gw-problem-number {
 		display: inline-block;
 		font-size: 16px;
 		margin-right: 5px;
--- a/htdocs/js/ProblemGrader/singleproblemgrader.js
+++ b/htdocs/js/ProblemGrader/singleproblemgrader.js
@@ -216,4 +216,141 @@
 			}
 		});
 	}
+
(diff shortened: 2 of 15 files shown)


Alex-Jordan (approved):
I tried this out, and it's good! It works as described. I will try #2876 too now.

PR #2878, opened 2026-01-04
Update MathJax to 4.1.0.

somiaj:
I don't think anything else needs updated. This is the result of `npm install mathjax@latest`.


drgrice1 (approved):
I will go ahead and approve then.

PR #2880, opened 2026-01-07
Fix sets manager import form selection multiple default option selected issue.

drgrice1:
Currently when the sets manager page loads the "Import how many sets?" select has "a single set" initially selected and the "Import from where?" select has "Select filenames below" selected.  But then if you change the first select to "multiple sets" the "Import from where?" select still has "Select filenames below" selected.  Furthermore, if you then click on that select and use shift-down arrow to select multiple sets, that first disabled option stays selected.  Then form validation fails for that option since it has no value if you click the "Import" button.

This just makes it so that when you switch from the "Import how many sets?" select from "a single set" to "multiple sets", that first option in the "Import from where?" select is immediately unselected.

There is also a little clean up of this section of JavaScript code.  The elements with id `import_source_select` and name `action.import.number` are actually the same element, so no need to find them in the DOM twice. Also ensure the elements exist and are found before trying to work with them. More of this is needed in this file, but this is probably good enough for now.

### DIFF
--- a/htdocs/js/ProblemSetList/problemsetlist.js
+++ b/htdocs/js/ProblemSetList/problemsetlist.js
@@ -145,14 +145,23 @@
 			const number = parseInt(numSelect.options[numSelect.selectedIndex]?.value ?? '1');
 			const importSourceSelect = document.problemsetlist['action.import.source'];
 			if (importSourceSelect) {
-				if (number === 1 && !importSourceSelect.value) importSourceSelect.options[0].selected = true;
 				importSourceSelect.size = number;
-				importSourceSelect.multiple = number > 1 ? true : false;
-				if (number > 1) importSourceSelect.options[0].selected = false;
+				if (number === 1) {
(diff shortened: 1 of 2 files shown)


somiaj (approved):
Works well. Fixing the plural issue is very minor, so I approve either way.

drgrice1:
I found out that I was checking the value of the `action.import.source` select to late in the method.  If you check it before the `size` and `multiple` attributes are changed, then it does have the selected option when switching from `multiple` to `single`.  So now if you switch from `multiple` to `single` with nothing selected, the first default option is selected.

As to the plurality of the wording, note that this code also has an untranslated string the javascript uses.  Namely `(taken from filenames)` for the `action.import.name` input.  This code is very old, and is from before I started adding this data attribute translation approach.  That could be updated.

drgrice1:
I went ahead and fixed the plurality and translated that other string.

PR #2881, opened 2026-01-07
Fix the label for the import date shift input.

drgrice1:
The input that is initally inserted into the page has the id that the label points to.  However, that input is hidden and replaced with the visible input for the flatpickr date selector, and so the label does not work.  This moves the id of the original and now hidden input onto the visible flatpickr input.

This is the same technique used for other date/time inputs in the `htdocs/js/DatePicker/datepicker.js` file, and should have been implemented in this special case as well.


somiaj:
Closing now that #2882 was merged.

PR #2882, opened 2026-01-08
Add the capability to mass edit a particular set date when editing sets in the sets manager.

drgrice1:
When editing sets in the sets manager, there are now some new inputs and such above the set list table.  There is a select to choose which set date to change (the open date, reduced scoring date, close date, or answer date), and which date to set that to for all selected sets. Obviously the reduced scoring date is not shown if not enabled for the course.  Then there is an "Apply to Selected Sets" button which if clicked will set the selected date type to the chosen date for all selected sets.

There now needs to be the set check boxes on the edit page (these were not shown in edit mode before).  Internally those checks have a different input name so that they don't conflict with the behavior of the usual checks and since they are for a different purpose.

There is validation, and if the button is used when the inputs haven't been set or if no sets are selected, then messages are displayed informing the user of what is needed.

Note that the usual ordering of dates applies.  So if a set date is out of order then the other dates are adjusted just as if that date were edited directly for the particular set.

This is how I envision what was requested in https://forums.openwebwork.org/mod/forum/discuss.php?d=8795#p22520. I don't see the suggested way of this working in that forum post as actually being feasible or even really making sense.  The suggestion was to add a button that would copy a date to all sets below it, but the order of sets in the list is not a reasonable way of managing this since that order could depend on the set dates to begin with.

### DIFF
--- a/htdocs/js/ProblemSetList/problemsetlist.js
+++ b/htdocs/js/ProblemSetList/problemsetlist.js
@@ -7,12 +7,12 @@
 		for (const id of ids) elements.push(document.getElementById(id));
 		for (const element of elements) {
 			if (element?.id.endsWith('_err_msg')) {
-				element?.classList.remove('d-none');
-			} else {
-				element?.classList.add('is-invalid');
+				element.classList.remove('d-none');
+			} else if (element) {
+				element.classList.add('is-invalid');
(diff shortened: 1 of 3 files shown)


somiaj (approved):
This works as advertised and should address the feature request.

PR #2887, opened 2026-01-13
Add SECURITY.md

shriyanss:
There’s currently no SECURITY.md in the repository root. Since WeBWorK is used by multiple universities for graded assignments, having a clear security reporting path matters—publicly disclosed vulnerabilities could impact the integrity of assignments and course outcomes.

I recommend adding a standard SECURITY.md to enable GitHub vulnerability reporting and to provide clear guidance on how to responsibly report issues (what details to include, expected response times, and where to report). GitHub’s guidance is here:
https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/configure-vulnerability-reporting/adding-a-security-policy-to-your-repository

PS: The file changed is an example, and should be replaced with the actual security policy.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,11 +6,10 @@
 
 | Version | Supported          |
 | ------- | ------------------ |
-| 2.20   | :white_check_mark:  |
-| 2.19   | :x:                 |
-| 2.18   | :x:                 |
-| 2.17   | :x:                 |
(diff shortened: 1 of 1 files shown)


somiaj:
Could we convert this to an issue, since we would need to modify the template file added here?

Issue #2885, opened 2026-01-12
Section filters fail if section name contains a colon


dlglin:
If the section name for a user contains a colon character, then filtering by such a section returns no users.

To reproduce:

- Create a user, and enter text including a colon in the section field (e.g. "a:1").
- Go to the Assigner Tool or Instructor Tools page, select the section containing the colon in the filter, then Refresh List.  No students appear.

PR #2888, opened 2026-01-14
Allow colons in filter values.

drgrice1:
The format for a filter in the scrolling record lists is `filterName:filtervalue`.  When the filters are applied the name and value are slit on the colon.  Now if the filter value contains a colon, then the current code splits that value as well, and since only two variables are saved in the return value, the rest goes off into the abyss.  This just adds the optional limit to the `split` call so that the colon in the value is not split.

This fixes issue #2885.


dlglin (approved):
Works for me.  Thanks.

PR #2889, opened 2026-01-15
Make it much easier to act as a student.

drgrice1:
This is achieved by displaying the student nav on the problem set page for users that have the permission to act as a user as well as always showing the student nav in problems and in tests.

If you are acting as a user, then the student nav looks the same as before with the next and previous buttons and the name of the user that is currently being acted as shown on the button.  However, if you are not currently acting as another user, then the next and previous buttons are not shown and the button says "Select Student to Act As" (or in a test it says "Select a Test to Review").

Also fix the breadcrumb in a test when the set is not valid, but the setID does have the `setID,v?` format.  Currently if you are acting as a student user and reviewing a test version, say "test,v1", but you have not worked the test, and you click the "Stop Acting" button, then the message stating that the selected test is not valid for the user is shown (it would be nice to not show this even) and the breadcrumb ends with the inactive link "test,v1", and you can only go back to the assignments page. Now, the "setID" link is shown and works, and the inactive "v1" link is at the end.

This is built on top of #2875 since it would conflict rather heavily with that pull request if it were not, and I don't want to have to resolve the conflicts later!


somiaj (approved):
Looks good.

PR #2892, opened 2026-01-19
Fix three issues with display of answers on the problem grader page.

drgrice1:
First, if an answer is a checkbox answer with multiple parts checked, then the `&#9070;` character is not handled.  This needs the same processing as is done on the past answers page for this.

Second, the essay answers can not be put into a `Mojo::Collection` and joined with `<br>` tags.  The result of that is a `Mojo::ByteStream` which means that it is not escaped.  That was the original point since the `<br>` tags cannot be escaped.  However, the answers must be escaped so that answers like `<script>alert('xss attack')</script>` are not executed.  So a for loop similar to that used for the checkbox answers must be used.  Note that these answers were also wrapped in a redundant `<div>` tag with the same class as the containing `<div>` that is still there, and that was removed.

Third, there was a dangling end `</div>` tag for formula answers that was removed.

### DIFF
--- a/templates/ContentGenerator/Instructor/ProblemGrader.html.ep
+++ b/templates/ContentGenerator/Instructor/ProblemGrader.html.ep
@@ -178,16 +178,21 @@
 										) %>">
 											% if ($answerTypes[$i] && $answerTypes[$i] eq 'essay') {
 												% # If the answer is an essay answer then display it line by line.
-												<div class="past-answer">
-													<%= c(split /\n/, $answers[$i])->join('<br>') =%>
-												</div>
+												% my @lines = split /\n/, $answers[$i];
+												% for (0 .. $#lines - 1) {
+													<%= $lines[$_] =%><br>
+												% }
+												<%= $lines[-1] =%>
 											% } elsif ($answerTypes[$i] && $answerTypes[$i] eq 'Value (Formula)') {
 												% # If its a formula then mark it as tex for MathJax.
 												`<%= $answers[$i] %>`
-												</div>
 											% } else {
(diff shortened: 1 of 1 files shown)


templates/ContentGenerator/Instructor/ProblemGrader.html.ep
```diff
@@ -179,16 +179,21 @@
...
 											% } elsif ($answerTypes[$i] && $answerTypes[$i] eq 'Value (Formula)') {
 												% # If its a formula then mark it as tex for MathJax.
 												`<%= $answers[$i] %>`
-												</div>
 											% } else {
 												% # If it isn't an essay or a formula then show it as text.
-												<%= $answers[$i] %>
+												% my @parts = split("&#9070;", $answers[$i]);
+												% for (0 .. $#parts - 1) {
+													<%= $parts[$_] =%>&#9070;\
```
Alex-Jordan, line 193: Is the backslash after the `&#9070;` meant to be here?

drgrice1: Yes, the backslash is intended.  That prevents a newline from being injected, and thus creating a white space node in the HTML output.  Here are screenshots showing the difference (the first image is with the backslash, and the second without):

Image
<br>
Image

and here is the difference in the generated HTML (again first with the backslash, and second without):

```html
								<td style="color:#006600" class="px-3 ">
C&#9070;D&#9070;E&#9070;F								</td>
```

```html
								<td style="color:#006600" class="px-3 ">
C&#9070;
D&#9070;
E&#9070;
F								</td>
```

PR #2893, opened 2026-01-19
Refactor check_modules script

dlglin:
Update the the `check_modules.pl` script to add the name of the Debian and RedHat package for perl packages where available.
This adds another flag when calling the script, where the user can specify if they are on a deb- or rpm-based system, and the output will suggest a command for installing any missing packages.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -544,6 +544,8 @@ my %modulesList = (
 	}
 );
 
+my %cpanm_package = ('ubuntu' => 'cpanminus', 'rhel' => 'perl-App-cpanminus');
+
 my @programList = qw(
 	convert
 	curl
@@ -631,10 +633,10 @@ sub determine_distribution {
 			}
 		}
 		close $fh;
-		if ($os_attrs{'id'} =~ 'ubuntu' || $os_attrs{'id_like'} =~ 'ubuntu') {
+		if ($os_attrs{'id'} =~ 'ubuntu' || ($os_attrs{'id_like'} && $os_attrs{'id_like'} =~ 'ubuntu')) {
 			$packagetype = 'ubuntu';
 			return 1;
-		} elsif ($os_attrs{'id'} =~ 'fedora' || $os_attrs{'id_like'} =~ 'fedora') {
+		} elsif ($os_attrs{'id'} =~ 'fedora' || ($os_attrs{'id_like'} && $os_attrs{'id_like'} =~ 'fedora')) {
 			$packagetype = 'rhel';
 			return 1;
 		}
@@ -695,6 +697,8 @@ sub check_modules {
 			if (@missing_packages) {
 				say $packagemgrcommand{$packagetype} . join(' ', @missing_packages) . "\n";
 			}
+			say 'If the cpanm command is not installed, you can try installing it using the following command:' . "\n";
+			say $packagemgrcommand{$packagetype} . $cpanm_package{$packagetype};
 		}
 		say 'Exiting as this is required to check the database driver and programs.';
 		exit 0;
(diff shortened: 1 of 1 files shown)


bin/check_modules.pl
```diff
@@ -30,98 +33,492 @@ =head1 DESCRIPTION
...
-my %moduleVersion = (
-	'Future::AsyncAwait'   => 0.52,
-	'IO::Socket::SSL'      => 2.007,
-	'LWP::Protocol::https' => 6.06,
-	'Mojolicious'          => 9.34,
-	'SQL::Abstract'        => 2.000000
+my %modulesList = (
+	'Archive::Tar' => {
+		'package' => {
+			'deb' => 'perl-modules',
```
drgrice1: Change this to
```suggestion
			'deb' => 'perl',
```
There is no `perl-modules` package (at least for Ubuntu).  There is a `perl-modules-5.38` package on Ubuntu 24 which contains all of the packages you listed for the `perl-modules` pakcage.  However it would probably be a bad idea to list that here.  That would be specific to the version of Ubuntu.  The `perl` package would be on all versions of Ubuntu, and it depends on the `perl-modules-5.38` package (that would be a different Perl version for other Ubuntu versions though), and so it will be installed if the `perl` package is.

bin/check_modules.pl
```diff
@@ -30,98 +33,492 @@ =head1 DESCRIPTION
...
+	},
+	'Data::Dump' => {
+		'package' => {
+			'deb' => 'libdata-dump-perl',
+			'rpm' => 'perl-Data-Dump'
+		}
+	},
+	'Data::Dumper' => {
+		'package' => {
+			'deb' => 'libperl',
```
drgrice1: Change this to
```suggestion
			'deb' => 'perl',
```
There is also no `libperl` package, but there is `libperl5.38t64`.  The `perl` package also depends on that.

drgrice1 (approved):
I think this is a good idea.  However, I think that it may be necessary to have separate `packagetype` options for Debian and Ubuntu.  Usually packages are the same for the two, but not always.  Could that be an issue for Red Hat and Oracle?

bin/check_modules.pl
```diff
@@ -12,6 +12,12 @@ =head1 SYNOPSIS
  Options:
    -m|--modules          Check that the perl modules needed by webwork2 can be loaded.
    -p|--programs         Check that the programs needed by webwork2 exist.
+   -d|--distribution     Specify your linux distribution.  Currently supported options are
+                           'ubuntu' - tested on ubuntu 24. May work for other distributions
+				using the apt package manager
+                           'rhel' - for RedHat Enterprise Linux and equivalents with the 
+				EPEL and CodeReady Builder repositories enabled
+				(e.g. Rocky Linux, Oracle Linux)
```
drgrice1: I just noticed this.  Fix the indentation here.  There are tabs characters that are messing with things.

dlglin:
> Would it be possible to add `debian` as an alias of `ubuntu` that could be overridden in the few places the package names differ. That way I wouldn't need to add a `debian => 'package'` option to each dependency?

I don't like this.  It would be based on the assumption that if there isn't a Debian package specified, then the name would match the ubuntu package, but it's possible that there isn't a Debian package at all.

pstaabp (approved):
tested on ubuntu.  Found that I was missing perl::Critic

bin/check_modules.pl
```diff
@@ -9,9 +9,35 @@ =head1 SYNOPSIS
...
+
+=over 4
+
+=item C<ubuntu>
+
+Tested on ubuntu 24. May work for other distributions using the apt package manager 
+
+=item C<rhel>
+
+For RedHat Enterprise Linux and equivalents with the EPEL and CodeReady Builder repositories enabled (e.g. Rocky Linux, Oracle Linux)
```
drgrice1: Please wrap this long line.  Roughly place a newline so that this fits into a 80 character limit.  That is generally what I try to do with POD.

dlglin: Fixed

PR #2896, opened 2026-01-28
Update some of the help.

drgrice1:
The instructor links help is updated to order the links in the help in the same order that they are in the site navigation.  Also add help for the "Job Manager" which was missing.

Also fix the links the PG problem editor help.  The links above the CodeMirror editor window in the PG problem editor have changed in two ways that the links in the help were not updated to match.

First, the "Problem Techniques" button was replaced with the "Sample Problems" button.  So that Wiki link is now also removed from the help, and it is now the local server sample problems link.

The PGML lab was updated so that it does not need the PGML-lab.pg file in the templates directory of the course.  It now directly uses that problem from the assets/pg directory.  The link in the help file now does the same thing.

I removed the `reference-link` class from the links above the CodeMirror window because that class does not do anything.  I also removed the old webwork.maa.org links in the comments because those don't need to be there even if they are updated to the new locations.

### DIFF
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -545,8 +545,8 @@ sub getConfigValues ($ce) {
 				var  => 'permissionLevels{report_bugs}',
 				doc  => x('Can report bugs'),
 				doc2 => x(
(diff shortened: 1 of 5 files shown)


drgrice1:
Yes, there are other jobs.  There are also instructor emails and achievement notification emails.

PR #2900, opened 2026-02-01
Minor tweak to the mass date edit on the set list page.

drgrice1:
The default option should be disabled.


drgrice1:
Note that only happens if you enter the edit page by editing "selected sets".  It doesn't if you edit "all course sets".  The reason for it is that you had the checkbox by the same name on the set list page checked.  So the Mojolicious auto form filling handling automatically checks it again.  But then the `selectall.js` code sees that none of the check boxes in the select all's group are checked and unchecks it.

This is a rather minor visual change that most won't notice, and I don't care if a few that are more observant do.  It isn't worth trying to fix.

somiaj:
Yea, I only noticed it when using the select all feature, and it is minor visual issue, everything works fine, I just thought I would mention it.

PR #2904, opened 2026-02-04
Add completed time to test output.

somiaj:
Add the submit time next to the time taken on completed test versions. My use case for this is being able to see the time a test ended and compare it to the time work was submitted to my lms when grading written work with a test.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1309,7 +1309,8 @@ async sub pre_header_initialize ($c) {
 	} elsif ($endTime > $set->due_date) {
 		$c->{exceededAllowedTime} = 1;
 	}
-	$c->{elapsedTime} = int(($endTime - $set->open_date) / 0.6 + 0.5) / 100;
+	$c->{elapsedTime}   = int(($endTime - $set->open_date) / 0.6 + 0.5) / 100;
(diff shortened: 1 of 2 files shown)


lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
```diff
@@ -1310,6 +1310,7 @@ async sub pre_header_initialize ($c) {
 		$c->{exceededAllowedTime} = 1;
 	}
 	$c->{elapsedTime} = int(($endTime - $set->open_date) / 0.6 + 0.5) / 100;
+	$c->{submitTime}  = $c->formatDateTime($endTime, $ce->{studentDateDisplayFormat});
```
drgrice1: You definitely can not use the `submitTime` hash key here.  That will conflict with the `submitTime` set in `lib/WeBWorK.pm` via the `submitTime` method of `lib/WeBWorK/Controller.pm`.

drgrice1: That is a critical hash key that is of the most importance for tests in fact.

somiaj: Thanks, though I think I got lucky in my testing this on my server. I see `$c->submitTime` as a method but not a key. But thanks for catching that.

pstaabp (approved):
I didn't test this, but looks fine.

drgrice1 (approved):
This looks good.

PR #2906, opened 2026-02-06
Clarify messages in extension achievement items when reduced scoring is enabled

dlglin:
The messages for extension items only talked about the due date, which would be confusing to students, as they would probably associate "due date" with the full-credit deadline.  This clarifies what will happen to both the reduced credit date and the close date, trying to use similar terminology to the messages at the top of the set.

### DIFF
--- a/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
+++ b/lib/WeBWorK/AchievementItems/ExtendDueDate.pm
@@ -65,7 +65,8 @@ sub print_form ($self, $set, $records, $c) {
 				$c->tag(
 					'p',
 					$c->maketext(
-						'Because the deadline has already passed you will only receive reduced credit during this extension.'
+						'Because the deadline has already passed you will only '
+							. 'receive reduced credit during this extension.'
 					)
 				)
 			)->join('');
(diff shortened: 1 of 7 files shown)


lib/WeBWorK/AchievementItems/ExtendDueDate.pm
```diff
@@ -25,14 +25,63 @@ sub can_use ($self, $set, $records) {
...
+					$c->maketext(
+						'Extend the reduced credit deadline of this assignment to [_1] (an additional 24 hours). [_2]',
+						$c->formatDateTime($set->due_date + ONE_DAY, $c->ce->{studentDateDisplayFormat}),
+						$randomization_statement
+					)
+				),
+				$c->tag(
+					'p',
+					$c->maketext(
+						'Because the deadline has already passed you will only receive reduced credit during this extension.'
```
drgrice1: Please change this line to

```suggestion
						'Because the deadline has already passed you will only '
							. 'receive reduced credit during this extension.'
```

The line exceeds the 120 character limit.

Also make the same change in the `ExtendDueDateGW.pm` and `SuperExtendGDueDate.pm` files.

drgrice1 (approved):
Other than the minor issue with line length, this looks good.

somiaj (approved):

Issue #2910, opened 2026-02-12
'accommodation_time_factor' cannot be null error when adding new users


dlglin:
If I try to add a new user to a course on the current develop branch I get the error:
DBD::MariaDB::st execute failed: Column 'accommodation_time_factor' cannot be null at /opt/webwork/webwork2/lib/WeBWorK/DB/Database.pm line 137

This happens when using the Accounts Manager or if I manually try to add an additional user when creating a new course from the Admin course.

somiaj:
Fixed in #2911

PR #2911, opened 2026-02-12
Fix an issue with adding and updating users when the accomodation_time_factor is not set.

drgrice1:
The `gen_new` method is not sufficient to create the new method for database tables with default values for columns.  So the `newUser` method needs to be created manually and make sure the default is set.
    
This was reported in issue #2910, and fixes that issue and most likely other places where users are added or updated without explicitly setting the `accomodation_time_factor`.

### DIFF
--- a/lib/WeBWorK/DB.pm
+++ b/lib/WeBWorK/DB.pm
@@ -436,11 +436,17 @@ sub abort_transaction {
 
 BEGIN {
 	*User            = gen_schema_accessor("user");
-	*newUser         = gen_new("user");
 	*countUsersWhere = gen_count_where("user");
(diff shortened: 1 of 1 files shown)


somiaj (approved):

drgrice1:
I had to change this further.  Not only the `addUser` method is affected.  Generally,  anytime a database user object is obtained with the `newUser` method there can be issues if that is used to add or update a user.  So that method needs to be overridden to ensure the default is set.  This is the same approach I used for the LTI launch data and the session in the key table.  I should have realized this was needed before.

somiaj (approved):

PR #2912, opened 2026-02-17
make grades page show true problem number

Alex-Jordan:
Suppose there is an assignment with exercises that were originally numbered 1-10. Then an instructor deleted numbers 2 and 7, and did not renumber the problems in the assignment. Prior to the change here, the grades page shows scores for "number 1" through "number 8". But of course, "number 8" is labeled "number 10" when the student visits the set, and so on.

So this change makes it so that in the Grades page, the score for "number 10" is under the 10 column, and so on. In columns 2 and 7, there will be an empty cell.

So it will be with regular sets and jitar sets. But this PR should not change anything with tests. For those, it's too complicated to try to change the current behavior because of how exercises can be presented in a random order. I'm not even sure that other things don't break if an instructor deletes an exercise from a test template or test version. If someone thinks this should happen with tests as well, I will need help making that happen.

This is a preliminary PR so that later, I can make the cells link to the problem number in question.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Grades.pm
+++ b/lib/WeBWorK/ContentGenerator/Grades.pm
@@ -7,7 +7,7 @@ WeBWorK::ContentGenerator::Grades - Display statistics by user.
 
 =cut
 
-use WeBWorK::Utils                    qw(wwRound);
+use WeBWorK::Utils                    qw(wwRound max);
 use WeBWorK::Utils::DateTime          qw(after);
 use WeBWorK::Utils::JITAR             qw(jitar_id_to_seq);
 use WeBWorK::Utils::Sets              qw(grade_set format_set_name_display);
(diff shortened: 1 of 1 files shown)


drgrice1:
This definitely should not do this for tests.  Note that in tests the problems are ALWAYS displayed for the student in consecutive order in the actual test regardless of the `problem_id` in the database and set definition.  So they should be displayed that way on the grades page as well.  This is true for the case that a problem was deleted (for example, if in the set definition the problem ids are 1, 3, 4) or the case that the problems are displayed in random order.

drgrice1:
My wording was not concise, so I want to clarify further.  If the problem ids are 1, 3, 4 in the set definition, then in the test the problem numbers are shown as 1, 2, 3.  They are displayed not only consecutively, but incrementally starting from 1.

The point is that the current numbering on the grades page is already consistent with the numbering that is shown for students in the test itself.

lib/WeBWorK/ContentGenerator/Grades.pm
```diff
@@ -181,21 +181,21 @@ sub displayStudentStats ($c, $studentID) {
...
 			for my $problemID (@problemIDs) {
 				my @seq = jitar_id_to_seq($problemID);
-				$num_of_problems++ if ($#seq == 0);
+				$max_problem_id = $seq[0] if ($#seq == 0 && $seq[0] > $max_problem_id);
 			}
 		} else {
-			# For other sets we just count the number of problems.
-			$num_of_problems = $db->countGlobalProblems($setID);
+			# For other sets we just get the largest problem ID.
+			$max_problem_id = max($db->listGlobalProblems($setID));
```
drgrice1, line 195: There is going to need to be a special case for tests that does what the previous else case did for those.  For example, if a test has numbers 1, 3, 5, then this will get 5 for the largest problem id.  However, the problems will be displayed under numbers 1, 2, and 3 (on both the grades page and in the test).  So this may add extra problem numbers that are shown even though there may not be any sets with those problem numbers.

Alex-Jordan:
Yeah, I am not opposed to something larger scale. But I have very little
bandwidth for something comprehensive. Do any others think they would be
giving a go at an overhaul?

For a student in D2L at the grades page, they see grade items as blocks
stacked vertically, almost certainly needing to scroll down a lot to see
all items. Within each block there are some details about the grade item.
Would we want a layout like that?

Alex Jordan
Mathematics Instructor
Portland Community College

Alex-Jordan:
What I really want that started this. I want it to be easy for a student to
be surveying their grades, and see that (for example) question 7 on
assignment 5 was never completed, and they can get their in one click.

Alex Jordan
Mathematics Instructor
Portland Community College

somiaj:
@Alex-Jordan One of the ideas that came up was remove the grades page from students. The info you want is already on the ProblemSet page, but not in one place. Also the grade page is kinda ordered poorly for this.

so one though, was just put the percent grade on the main Assignments page, then students can see the % of what is currently open or in the reduced scoring period, and if it isn't 100%, they can click on it and then see exactly which problems they have left to finish.

Alex-Jordan:
I'm closing this since #2917 is under development.

PR #2917, opened 2026-02-25
Replace the student grades page.

somiaj:
This replaces the Grades page for students with a new layout designed for students. The old grades table is still available to instructors under "Student Progress". Being a grade page for students, instructors see the same info a student would (no hidden sets or grades are shown for instructors acting as a student). The only difference for an instructor is the student navigation menu is shown to switch which student to act as.

The assignments are split into categories. Open assignments, reduced scoring assignments (if reduced scoring is enabled), and closed assignments. Assignments are ordered using the same `byUrgency` logic as on the `ProblemSets` page.

The total grade, if configured to be shown, is shown at the top of the page for all sets that are past the open date.

Each assignment is a list item which shows the total score. For tests the best test version score is shown if the student can see the score. Then a table which includes the total score and status for each problem in the set. For just in time, only top level problems are shown. For tests, only the best test version is shown.

The old Grades page code is moved into `StudentProgress.pm` where it is still used.


Alex-Jordan:
Thanks @somiaj.

> Please let me know if there are any other features you would think would be useful for students on this page.

For me with #2912, this started because I wanted a quick way to be on the Grades page and jump directly to a particular exercise. So can you make it so that from each cell in a table for an assignment, there is a link to that problem?

Based on what I read here, are you no longer reporting the number of attempts used on a problem? (Or the number of incorrect attempts, as it were?) I find that helpful for a student to see in one place where they struggled, even if they eventually have all answers correct.

somiaj:
@Alex-Jordan Thanks for the suggestions, I too thought about adding a link directly to the problem attached to the problem number, then forgot I wanted to do that. I can make each problem number a link, though I'm a bit worried about accessibility here. Having multiple links called "1" that go to different locations isn't something that is considered accessible I thought.

You are correct, I'm not currently showing attempts on this page as I couldn't think of a use case for that information for students. Your use case sounds reasonable though it would add another row to the problem table for each set will end up taking more space. Also should this only be done on assignments, what about tests? I know tests are used in different ways, but to me it won't be as useful there. My experience is students (unless prompted to do so) probably won't be concerned about how many attempts they take, also with scaffold and multi part problems that info seems to not be as useful. I'm just brain storming the benefit of adding another row vs the overall use of it. Maybe let me know what you think about the layout of adding such a row once you get a look at this. Would me adding some screenshots of the current layout help?

somiaj:
@Alex-Jordan correct, for assignments, I show the `problem_id`, but for tests I show the problem count.

drgrice1:
I was just looking at this to see where it stands.  Here are the things that I observe at this point.

Did you forget to add a `grades.scss` file to this?  I am seeing a console warning about a `grades.css` file not being found (it is loaded from `templates/ContentGenerator/Grades.html.ep`), but see no such file in this pull request.  I also don't see anything like your screenshots.

Also, why is there a `assets/pg/myNewProblem.pg` file in this?

One thing I notice immediately when looking at the result is that there are structural problems.

The total grade shown at the top of the page should not be a list with a single item in it, and that can never have more than one item.  That is semantically incorrect.  I also don't think it is visually appealing to use a `list-group` for this.

The grade items need to be reworked.  A screen reader goes to the percent grade for the set first before announcing the set.  It is semantically out of order.

somiaj:
Fixed the last issues mentioned. I accidentally committed the wrong file, so fixed that and the `grades.scss` is now included.

I updated the structure to address @drgrice1 comments. The grade total is no longer a list. I also set up some custom flexbox css to make it so the items are listed as 'set name' -> 'grade %' -> 'problem table', but are still positioned in a similar fashion.

I also updated it so only the best test version is shown. And if there is more than one test version, the grade table informs the user that they are seeing `Version X of Y total versions`.

Probably still a bunch to do. If you notice any problems please submit a PR, as I am limited on time, but will work on this when I can.

somiaj:
I removed all the code/tooltips to decide if an assignment is complete or not to inform students to work on it. There is just too many conditions, such as should students who have resurrect items be told they can use them, what about reduced scoring removal items, and so on. Instead this just now lists the score.

Currently I am using the same `byUrgency` logic to sort the sets as the ProblemSets page. I don't know if this or just alphabetical is better. I guess it depends on what is being looked for. I'm open for suggestions on how to organize the sets.

At least for me this is getting closer to being ready. Since I decided to not go down the route of trying to indicate sets that students can improve on, I think only suggestions/fixes for any issues found are needed.

Alex-Jordan:
I'm looking at the page now, and just commenting on what I see there. I haven't looked at the code itself. Sorry for the large list, just reporting what I see.

- Something is wrong with how it displays test versions. In one test, I took it using three versions. My scores are 1/9, 1/9, and 0/9. So zero on the third version. The table has the correct scoring (11%, 1 out of 9, indicates the right scores per problem) but it says "Version 3 of 3 versions taken"). It should be either version 1 or version 2, which is where the scoring data is coming from.
- It might help with accessibility to put `scope="row"` on the `th` elements for the row headers.
- Not complaining but why not put links to problems when the set is a test version?
- In the status row, maybe it would be better to leave out all the % signs, but put that in the row header like "Status (%)".
- I don't like how the set titles are indented, sort of straddling over the tables. Why not have them left-aligned? Or left-aligned with the table?
- How easy is it to keyboard navigate to a set somewhere in the middle of the list? It might be easier if the set titles were `h3`.
- I do use the current grades page to see the number of excess incorrect attempts, and this changes things by not showing that data. I don't need to see "excess incorrect attempts", just showing total number of attempts would serve the same purpose. (Which is to see which exercises a particular student struggled with.) Could you add a row to these tables for that?
- The current grades page indicates which problems have not been attempted. Could we do that here? Instead of writing a 0 for the status, there could be an indicator for when there are no attempts at all.
- The 💯 character clashes stylistically with the other percentages. We used that in the current table to save horizontal space, but I think it may be better to use "100%" with this redesign.
- Once this is merged, I might propose some styling changes. But I don't want to get in the way of this for now.

somiaj:
> Not complaining but why not put links to problems when the set is a test version?

Mostly because this requires a bit of extra work due to problems per page, so the links will be per page vs per problem. Could be done, just didn't think it was worth the extra work.

>  I don't like how the set titles are indented, sort of straddling over the tables. Why not have them left-aligned? Or left-aligned with the table?

Mostly for space, the setup is something like this

```
|---|-----------------|
|   |        1        |
| 2 |-----------------|
|   |        3        |
|---|-----------------|
```

This way the tab order works, but this makes the rows a little more compact. Of course this can be changed, this is just the layout I went with.

> How easy is it to keyboard navigate to a set somewhere in the middle of the list? It might be easier if the set titles were `h3`.

I use the order described above so tab hits the title first, the status second and the problem/version table last.

> I do use the current grades page to see the number of excess incorrect attempts, and this changes things by not showing that data. I don't need to see "excess incorrect attempts", just showing total number of attempts would serve the same purpose. (Which is to see which exercises a particular student struggled with.) Could you add a row to these tables for that?

I know this is something you want, and we did talk about this a bit in the meeting. Overall this information can be found if students go to the set, and it just takes up extra space and I don't really think it will be used by most students. Instructors still have access to the old grades page under student progress so they can see that info compactly if needed. So after talking it over in the meeting we kinda just decided it wasn't worth adding another row for this use.

> The current grades page indicates which problems have not been attempted. Could we do that here? Instead of writing a 0 for the status, there could be an indicator for when there are no attempts at all.

To me this is more a grades page, no attempts is still a zero. So not something I thought of. Again students get direct links to the problem set if they want more info about why it is a zero, so I'm not convinced it is worth having that info here too. Of course use cases to vary.

> The 💯 character clashes stylistically with the other percentages. We used that in the current table to save horizontal space, but I think it may be better to use "100%" with this redesign.

Similar result here, `100%` will make everything wider. Though If we remove the `%` it could work. I kinda like the `%` (I do know you suggested not having it, but I think it looks better).

> Once this is merged, I might propose some styling changes. But I don't want to get in the way of this for now.

You can make a PR to my branch too, though I guess you have lots of ideas of info you would want to see, so that is probably a good idea, so your particular changes can be discussed separately of the initial update.

Alex-Jordan:
> Mostly for space, the setup is something like this
>
```
|---|-----------------|
|   |        1        |
| 2 |-----------------|
|   |        3        |
|---|-----------------|
```
>
> This way the tab order works, but this makes the rows a little more compact. Of course this can be changed, this is just the layout I went with.

The "1" and the "3" are not aligned. The "1" starts left of the "3":

Screenshot 2026-05-19 at 6 57 23 PM

If the "1" is not to be all the way to the left, can it start left-aligned with the table?

> I use the order described above so tab hits the title first, the status second and the problem/version table last.

That's not what I mean. Suppose there are 20 sets and I want to navigate to the 12th. The tab order you describe asks me to tab through 11*3 things before I reach the 12th set. I'm looking for a way to navigate past the first 11 sets more quickly, and one way (not sure it is the best) would be if the set titles were `h3`. Keyboard navigators can navigate quickly through `hN` elements.

> I know this is something you want, and we did talk about this a bit in the meeting. Overall this information can be found if students go to the set, and it just takes up extra space and I don't really think it will be used by most students. Instructors still have access to the old grades page under student progress so they can see that info compactly if needed. So after talking it over in the meeting we kinda just decided it wasn't worth adding another row for this use.

I ask my students to use that information. And it sounds like you are saying my students should navigate into each set one at a time to hunt the exercises where they used multiple attempts. But when they are prepping for an exam, I ask them to look at this one page to see the places where they struggled, even when they eventually got everything correct. And I trust that will work more when they don't have to track entering and checking every assignment.

I could be fine without this information on tests. And test tables here already use 4 rows. So adding a 4th row for homework sets would actually make everything have a more uniform look.

> To me this is more a grades page, no attempts is still a zero. So not something I thought of. Again students get direct links to the problem set if they want more info about why it is a zero, so I'm not convinced it is worth having that info here too. Of course use cases to vary.

This is the second thing where the new version of this page is losing information that the current version has. Well, I guess it's the same, since if the number of attempts is included in a row, then there would not need to be a separate "unattempted" indicator.

> Similar result here, 100% will make everything wider. Though If we remove the % it could work. I kinda like the % (I do know you suggested not having it, but I think it looks better).

I'm referring to the "2" in your diagram. Not the scores in table cells. And changing to use "100%" does not make anything wider. Here I changed one directly using the developer tools:


Screenshot 2026-05-19 at 7 08 25 PM

The scores in the table cells are actually not using the special 💯 character right now, and are using "100%". And so actually those are taking up more horizontal space at the moment.

somiaj:
I implemented some of those suggestions.

* Change the titles to `h3` and then made them smaller than `h3` is normally to help with key board navigation.
* Added `scope="row"` to table headers.
* Added attempts to the homework problems table.
* Use `100%` for the grade instead of UTF-8 character.
* Attempted to fix the bug where the wrong version number was being displayed.
* Removed table indent. I still like it better, but don't have strong feelings.

I looked at maybe adding links to test problems, and didn't see an easy way to do this, as the page  you are viewing is part of the parameters sent in the link. It is possible just not something I thought is important enough to figure out at the moment.

Alex-Jordan:
Thanks for those changes. About saving space, I have a few thoughts.

I already mentioned saving horizontal space by moving the "%" signs to the row header. So this:

Screenshot 2026-05-20 at 12 00 56 AM

would become this:

Screenshot 2026-05-20 at 12 01 55 AM

Although it might look better to set a uniform width on those cells:

Screenshot 2026-05-20 at 12 05 30 AM

For saving some vertical space, I am looking at the "Score: 5 out of 8" rows that don't really need to be part of the table. They give information about the set as a whole, not per problem which is what the table is for. They also probably make the table a challenge to navigate with a keyboard. And probably the problem number row should be full of `<th scope="col">` tags instead of `<td>`, so that if a user is in a lower cell, their screen reader can clearly identify something like "row Status, column problem 3" while keeping focus on that cell.

So what do you think about moving the "5 out of 8" to somewhere else? In some respects, it's redundant, since we already see the 63% overall score. Do we need "5 out of 8" at all? Would it make sense in tiny font right under the "63%"? Or just some place that doesn't take up vertical space?

Similarly with tests and the row that has "Version 1 of 3 versions taken". That could be a parenthetical adjacent to the title of the test.

If you don't like these ideas, feel free to reject them. I'm going to mark approval now. (Bad Alex, I haven't scrutinized the actual code...)

Alex-Jordan (approved):

somiaj:
* Used "Status (%)" for problem table to remove % from each score.
* Moved the version and score from the table into a small caption.
* Made the problem numbers table headers with `scope="col"`.
* Made the `min-width` of the table header `2rem`, so they look more uniform.

@Alex-Jordan What about those changes?

templates/HelpFiles/Grades.html.ep
```diff
@@ -2,13 +2,21 @@
 % title maketext('Grades Help');
 %
 <p>
-	<%= maketext(q{This page shows the student's current grades for all sets they are assigned to.  Only visible sets }
-		. 'are shown to the student, while invisible set names are italic when viewed as an instructor.  Students can '
-		. 'only see the per problem grades on open assignments.') =%>
+	<%= maketext(q{This page shows the student's current grade for all assignments they are assigned to.  This page }
```
Alex-Jordan, line 5: Some of what is written here in the new help text is addressed to an instructor. If a student is at their Grades page and clicks the help icon, some of this content won't mean anything to them. Like what is the "Student Progress" page, or the "student navigation menu at the top"?

drgrice1: The help icon is not shown for students.

Alex-Jordan: Oh. Are any of them shown for students?

drgrice1: Not at this point.

Alex-Jordan:
I opened a PR to @somiaj's branch:  https://github.com/somiaj/webwork2/pull/15

But evidently I need to clean up the scss code formatting. How do I do that again?

drgrice1:
Run `npm run prettier-format` from the `htdocs` directory.

Alex-Jordan:
I think it's good, and my previous approval tag stands. I only skimmed the code changes. But the behavior on the page seems good. I'm not seeing that issue with the test versions anymore, for instance.

htdocs/js/Grades/grades.js
```diff
@@ -0,0 +1,32 @@
...
+	};
+
+	toggleWeightsButton?.addEventListener('click', () => showHideRows(toggleWeightsButton));
+	toggleAttemptsButton?.addEventListener('click', () => showHideRows(toggleAttemptsButton));
+
+	document.addEventListener('DOMContentLoaded', () => {
+		for (const button of [toggleWeightsButton, toggleAttemptsButton]) {
+			if (button && localStorage.getItem(`WW.${courseID}.grades.show-${button.dataset.rowClass}`))
+				showHideRows(button);
+		}
```
drgrice1: There is no need to execute this code in the `DOMContentLoaded` event handler.  This script is deferred, and so the DOM is ready for this code to run as soon as the script loads.

templates/ContentGenerator/Grades/grade_items.html.ep
```diff
@@ -0,0 +1,33 @@
...
+					<div><em><%= $item->{message} %></em></div>
+				% }
+				% if (defined $item->{problems}) {
+					<%= include 'ContentGenerator/Grades/problem_table', item => $item %>
+				% }
+			</div>
+		</div>
+	</li>
+% }
+</ul>
```
drgrice1, line 32: Everything inside the `<ul>` and `</ul>` should be indented one more level.

templates/ContentGenerator/Grades/grade_items.html.ep
```diff
@@ -0,0 +1,33 @@
+<ul class="list-group">
+% my $n = 0;
+% for my $item (@$items) {
+	% $n++;
```
drgrice1: Always pre-increment unless post-increment is needed.  So this should be

```suggestion
	% ++$n;
```

drgrice1 (commented):
Generally this is looking nice.  Just a few things I noticed.

The margins don't look good between the different assignment groups.  Here is what I see:
Image
The margin on top of the header doesn't match the margin on the bottom, and so it is to close to the group above.

drgrice1:
Also, the wording for tests is a bit strange to me.  "Score: 16 out of 16 for version 4 of 5 versions taken" seems off for some reason.  Also, I think the link to the actual version should only be on "version 4", and not "for version 4 of 5 versions taken".

drgrice1:
Perhaps it would be better to make that two parts. Something like "Score: 16 out of 16 for version 4, 5 versions taken".  The current wording seems to say the score of 16 out of 16 is for 4 of the 5 versions.  It was just a bit confusing to me when I read it the first time.  Maybe it is the repetition of the word version in the sentence?  I am not sure, it just doesn't read well to me.

Alex-Jordan:
You can blame me for the wording. Jaimos had the two pieces of information on separate lines, and I merged them to save vertical space. Any improvement welcome.

drgrice1:
Perhaps a parenthetical would work well here.  Something like "Score: 16 out of 16 for version 4 (5 versions taken)"

somiaj:
* Add "mb-3" to give a bottom margin on the list group items+
* Indent code properly.
* Make versions taken statement a parenthetical.
* Remove use of DOM content loaded event.

I originally thought it might be useful to have versions taken information on this page. Though this is not something I use and unsure how useful it is, we could remove it if we don't think it is that useful to know number of versions taken and only just show the best version score.

drgrice1:
I think that being able to see all versions taken could be useful to students.  If a student has taken several attempts at a test, it could be useful for the student to identify problems that they are having trouble with for future attempts.  This does seem to be a subtraction from the previous grades page that is not the best.

somiaj:
I think the reason I put the number of versions taken so students know they can click on the test to see them. I originally was showing all versions, but that could take up a lot of space. We could also go with the hidden route and have a way to toggle seeing all versions but keep them hidden by default.

Alex-Jordan:
I like it reporting the number of versions. Maybe that is enough for now, but later we could explore a way to cycle through all test versions. Like the page could load showing version 3 scores (for example, because that is the high score version) with clickable left/right arrows to carousel through other versions. So it would not take extra space.

drgrice1:
> Also, I think the link to the actual version should only be on "version 4", and not "for version 4 of 5 versions taken".

This was not addressed.  The now parenthetical "(5 versions taken)" should not be part of the hyperlink.  The link is to the version that was taken, not all versions taken.

templates/ContentGenerator/Grades/problem_table.html.ep
```diff
@@ -0,0 +1,66 @@
...
+			% }
+			% if ($item->{version_count} > 1) {
+				<%= maketext('for version [_1] ([_2] versions taken)',
+					$item->{version_id}, $item->{version_count}) %>
+			% } else {
+				<%= maketext('for version [_1]', $item->{version_id}) %>
+			% }
+			% if ($item->{version_link}) {
+				</a>
+			% }
```
drgrice1: Here is one way to make the link only be on the version taken text:

```suggestion
			% my $takenVersion = maketext('for version [_1]', $item->{version_id});
			% $takenVersion = begin
				<a class="fw-bold" href="<%= $item->{version_link} %>"><%= $takenVersion %></a>
			% end->() if $item->{version_link};
			<%== $takenVersion %>
			% if ($item->{version_count} > 1) {
				<%== maketext('([_1] versions taken)', $item->{version_count}) =%>
			% }
```
It splits the parenthetical into a separate `maketext` call, but that is fine.  It is actually probably worse for translation to split the `[_1] out of [_2]` from the `for version [_1]`.

somiaj: I was just doing something similar. I also removed the 'for' from the version link.

somiaj: Updated to use your `begin .. end` approach, but still went with mine of keeping the two translations, I think it is okay to split just the version name from the full translation like I did, but I can update to your approach if preferred.

drgrice1 (commented):
Looks good.  You have addressed the primary issues.

It would be nice to make test problems links as well, but that can wait for now.

drgrice1 (approved):

PR #2920, opened 2026-02-26
Show set description on ProblemSet page.

somiaj:
This adds the set description to the ProblemSet page as discussed in #2919.

There were suggestions for a configuration option or a reveal button. In most cases I suspect the description would be short, so I just went with showing it if it is not empty. The description also cannot include any html or formatting. It will just be inserted into a single div.


drgrice1 (approved):

dlglin (approved):
Works for me.

PR #2921, opened 2026-02-26
Add point input to ProblemGrader.

somiaj:
Add a point input field to set the problem score in the ProblemGrader. This is the same as what is being used in the SingleProblemGrader, and honors the same setting to show it or not. The input uses JavaScript to update the actual score which is what is submitted when the grader is saved. If the percent score is not shown, a hidden field is used instead.

This also adds a step of 1 to the percent score and validation on both the percent score and point score values.

Note, I couldn't think of a way to nicely use the same javascript for both the ProblemGrader and SingleProblemGrader due to slight differences in how the pages are setup and the fact the SingleProblemGrader shows per part scores too, so there is some duplication of the javascript to update the point and percent values.

### DIFF
--- a/htdocs/js/ProblemGrader/problemgrader.js
+++ b/htdocs/js/ProblemGrader/problemgrader.js
@@ -1,6 +1,45 @@
 'use strict';
 
 (() => {
+	const setPointInputValue = (pointInput, score) =>
+		(pointInput.value = parseFloat(
+			(Math.round((score * pointInput.max) / 100 / pointInput.step) * pointInput.step).toFixed(2)
+		));
+
+	// Update problem score if point value changes and is a valid value.
+	for (const pointInput of document.querySelectorAll('.problem-points')) {
+		pointInput.addEventListener('input', () => {
+			const userId = pointInput.id.replace(/\.points$/, '');
+			if (pointInput.checkValidity()) {
+				const scoreInput = document.getElementById(`${userId}.score`);
+				if (scoreInput) {
+					scoreInput.classList.remove('is-invalid');
+					scoreInput.value = Math.round((100 * pointInput.value) / pointInput.max);
+				}
+				pointInput.classList.remove('is-invalid');
+			} else {
+				pointInput.classList.add('is-invalid');
+			}
+		});
+	}
+
+	// Update problem points if score changes and is a valid value.
+	for (const scoreInput of document.querySelectorAll('.problem-score')) {
+		scoreInput.addEventListener('input', () => {
+			const userId = scoreInput.id.replace(/\.score$/, '');
+			if (scoreInput.checkValidity()) {
+				const pointInput = document.getElementById(`${userId}.points`);
+				if (pointInput) {
+					pointInput.classList.remove('is-invalid');
+					pointInput.value = setPointInputValue(pointInput, scoreInput.value);
+				}
+				scoreInput.classList.remove('is-invalid');
+			} else {
+				scoreInput.classList.add('is-invalid');
+			}
+		});
+	}
+
 	const userSelect = document.getElementById('student_selector');
 	if (!userSelect) return;
 
--- a/templates/ContentGenerator/Instructor/ProblemGrader.html.ep
+++ b/templates/ContentGenerator/Instructor/ProblemGrader.html.ep
@@ -1,4 +1,4 @@
-% use WeBWorK::Utils qw(wwRound getAssetURL);
+% use WeBWorK::Utils qw(wwRound getAssetURL points_stepsize round_nearest_stepsize);
 % require WeBWorK::PG;
 %
 % content_for js => begin
@@ -122,11 +122,17 @@
 						<%= check_box 'select-all' => 'on', id => 'select-all', class => 'select-all form-check-input',
 							data => { select_group => 'mark_correct' } =%>
 					</th>
-					<th id="score-header"><%= maketext('Score (%)') %></th>
+					% unless ($ce->{problemGraderScore} eq 'Point') {
+						<th id="score-header"><%= maketext('Score (%)') %></th>
+					% }
+					% unless ($ce->{problemGraderScore} eq 'Percent') {
+						<th id="point-header"><%= maketext('Points (0 - [_1])', $problem->value) %></th>
+					% }
 					<th id="comment-header"><%= maketext('Comment') %></th>
 				</tr>
 			</thead>
 			<tbody>
+				% my $stepSize = points_stepsize($problem->value);
 				% for my $user (@$users) {
 					% my $userID = $user->user_id;
 					%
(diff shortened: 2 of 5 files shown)


htdocs/js/ProblemGrader/problemgrader.js
```diff
@@ -1,6 +1,43 @@
...
+
+	// Update problem score if point value changes and is a valid value.
+	for (const pointInput of document.querySelectorAll('.problem-points')) {
+		pointInput.addEventListener('input', () => {
+			const userId = pointInput.id.replace(/\.points$/, '');
+			if (pointInput.checkValidity()) {
+				const scoreInput = document.getElementById(`${userId}.score`);
+				pointInput.classList.remove('is-invalid');
+				scoreInput.classList.remove('is-invalid');
+				scoreInput.value = Math.round((100 * pointInput.value) / pointInput.max);
```
drgrice1: It would be best to change this to

```suggestion
				if (scoreInput) {
					scoreInput.classList.remove('is-invalid');
					scoreInput.value = Math.round((100 * pointInput.value) / pointInput.max);
				}
pointInput.max);
```

The score input should always be there, but sometimes things don't work right.  So this just protects against an exception in the event that for some odd reason the score input isn't there.

somiaj: Updated.

lib/WeBWorK/ContentGenerator/Instructor/ProblemGrader.pm
```diff
@@ -25,10 +25,11 @@ async sub initialize ($c) {
...
 	# Make sure these are defined for the template.
-	$c->stash->{set}          = $db->getGlobalSet($setID);
-	$c->stash->{problem}      = $db->getGlobalProblem($setID, $problemID);
-	$c->stash->{users}        = [];
-	$c->stash->{haveSections} = 0;
+	$c->stash->{set}           = $db->getGlobalSet($setID);
+	$c->stash->{problem}       = $db->getGlobalProblem($setID, $problemID);
+	$c->stash->{users}         = [];
+	$c->stash->{haveSections}  = 0;
+	$c->stash->{problem_value} = $c->stash->{problem}->value;
```
drgrice1: I don't exactly see the point of adding this stash value.  The `problem` is in the stash, and so the value can be accessed from there anywhere already.  Stashing the problem value additionally is unnecessary.

Although, there is perhaps a bigger issue with this in the pull request in general.  This is using the same problem value for all students.  Yes, usually this is the same for all students, but it does not have to be. The problem value can be set per student.  It is probably bad teaching practice to do so, but it is a supported override setting.  So the merged user problem should be used for this instead.  The single problem grader does use the user's specific value for the problem.

This probably means you cannot show the `Points (0 - n)` in the header for the column since `n` is not necessarily the same for all students.  Furthermore, the `stepsize` for the number input will need to be set for each student separately.

somiaj: I originally tried to use the user problem value, but it was undefined in my tests because the ProblemGrader currently isn't pulling in merged sets. Which lead me to just using the template assuming that if someone was going to make different users have different values, they would be able to adjust. But I agree, I probably should have dug deeper (mostly I was just trying to get something to work and forgot to go back and deal with this, so thanks for catching it).

Instead of using the merged set, I am still going to use the template set value unless the user set value is defined and different. So I left the header `Points (0 - n)`, since it probably relevant most the time.

In the case that a user's set value differs from the template set, I use it for the `max` value and the `stepsize`. I then write `Points (0 - m)` under the the number input to let the user know that the point value is different in this case.

I'm not fully sold on adding the `Points (0 - m)` there. I probably need to update the aria stuff for accessibility. Maybe a tooltip would be better. I am going to think on it a bit, but I have pushed my updates and any other suggestions are welcomed.

somiaj: I tried to put `0 - n` as a tool tip for each number input, and I am not liking that any better. Since it will probably be rare this case happens, probably best to show the different point range in the rare case it was different than the template set.

somiaj: I am thinking maybe a tooltip only on the number inputs that differ from the template set would work and be the least intrusive.

templates/ContentGenerator/Instructor/ProblemGrader.html.ep
```diff
@@ -206,14 +212,42 @@
...
-								<%= number_field "$userID.$versionID.score" =>
-										wwRound(0, $_->{problem}->status * 100),
-									class => 'score-selector form-control form-control-sm restricted-width-col',
-									style => 'width:6.5rem;', min => 0, max => 100, autocomplete => 'off',
-									'aria-labelledby' => 'score-header' =%>
-							</td>
+							% unless ($ce->{problemGraderScore} eq 'Point') {
+								<td class="restricted-width-col">
+									% param("$userID.$versionID.score", undef);
+									<%= number_field "$userID.$versionID.score" => wwRound(0, $_->{problem}->status * 100),
```
drgrice1, line 218: This is exceeding the 120 character limit.  Probably the cleanest way to fix this is to change it to

```suggestion
									% param("$userID.$versionID.score", wwRound(0, $_->{problem}->status * 100));
									<%= number_field "$userID.$versionID.score" => 0,
```

That is equivalent in behavior.  The same change would fix line 232 exceeding the character limit.

drgrice1 (approved):
This looks good now (other than the minor character limit issue).

I think that what you have done will work for the case that a user has the value for a problem overridden.  Although, there is one odd case.  If the value for the global problem is nonzero, and the user's value is overridden to be 0, then things don't quite work right.  You can ignore this for now if you don't want to mess with this probably extremely rare odd case.  Even with the single problem grader this case is a bit odd.  It shows points possible of `0-0` and if the user has a status of 0, then it shows the user has 0 points, but 100%.

templates/ContentGenerator/Instructor/ProblemGrader.html.ep
```diff
@@ -206,14 +212,42 @@
...
+										id    => "$userID.$versionID.score",
+										class => 'problem-score form-control form-control-sm restricted-width-col',
+										style => 'width:6.5rem;', min => 0, max => 100, step => 1,
+										autocomplete => 'off', 'aria-labelledby' => 'score-header' =%>
+								</td>
+							% }
+							% unless ($ce->{problemGraderScore} eq 'Percent') {
+								% my $problemValue = $_->{problem}->value || $problem->value;
```
drgrice1: To fix the user value 0 versus nonzero set value issue, I think it is sufficient to just change this to

```suggestion
								% my $problemValue = $_->{problem}->value ne '' ? $_->{problem}->value : $problem->value;
```

This exceeds the 120 character limit by 1, but we can let that slide.  Note that `$_->{problem}->value` will always be defined and is the empty string if `NULL` is in the database.  So there is no need to also check that it is defined.

somiaj:
```
% my $problemValue = ($_->{problem}->value ne '' ? $_->{problem} : $problem)->value;
```

Does the same thing and doesn't exceeded 120 characters so I went with that.

templates/ContentGenerator/Instructor/ProblemGrader.html.ep
somiaj: Thanks, unsure how I missed that exceeded 120 characters.

somiaj:
I added one more tweak, in the case the user value is different than the template value, I put the label `Points (0 - n)` in a div with an id, and updated the `aria-labeledby` of the number input to use that vs the set header for accessibility.

templates/ContentGenerator/Instructor/ProblemGrader.html.ep
drgrice1: Yep.  That works too.  Looks good.

dlglin (approved):

PR #2922, opened 2026-03-03
PGProblemEditor: Add option to copy auxiliary files.

somiaj:
When saving a problem in a new location in the PGProblemEditor, add an option to copy auxiliary files or not. Since it is not possible to determine what files are needed, all auxiliary files were copied in all cases. This allows the user to uncheck the "Copy auxiliary files" option to not copy any auxiliary files.

The default option is to copy auxiliary files which has not changed. This only allows the user to disable that if they are sure no auxiliary files are needed.

This also adds a check so the message about auxiliary files being copied only shows if at least one file was copied.

@drgrice1 I am unable to figure out how to keep this new option to stay unchecked once a user unchecks it. I used the same setup the "Create backup" option uses with a hidden field, but when I uncheck this new option, and save the file, the page checks it by default again. So I'm unsure what is going on in that case.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/PGProblemEditor.pm
@@ -704,14 +704,15 @@ sub saveFileChanges ($c, $outputFilePath, $backup = 0) {
 	}
 
 	# If the file is being saved as a new file in a new location, and the file is accompanied by auxiliary files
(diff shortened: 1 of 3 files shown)


somiaj:
Note, due to the fact `Save As` is a redirect, form parameters are lost in the redirect. This is a more complicated issue, so this PR is not going to address the issue about the checkbox staying unclicked, as `Save As` actually losses info about most the form anyways. The code is setup so if ever the redirects better deal with this, it will work.

drgrice1 (approved):
This looks good.

dlglin (approved):

PR #2925, opened 2026-03-03
ProblemSets: Place set description in info button.

somiaj:
Move the set description from a tooltip on the link to the set to a tooltip on an info button located just right of the set link if the set description is not empty.

This adds an additional option to the `set-id-tooltip` class JavaScript that if `data-fallback-placements` is set as a spaced separated list, that list will be passed to the bootstrap Tooltip on creation.

This was suggested by @dlglin.

### DIFF
--- a/templates/ContentGenerator/ProblemSets/set_list_row.html.ep
+++ b/templates/ContentGenerator/ProblemSets/set_list_row.html.ep
@@ -17,9 +17,16 @@
 	<div class="ms-3 me-auto">
 		<div dir="ltr">
 			<%= link_to $display_name => $c->systemLink(url_for('problem_list', setID => $set->set_id)),
-				class => 'fw-bold set-id-tooltip',
-				data  => { bs_toggle => 'tooltip', bs_placement => 'right', bs_title => $set->description }
+				class => 'fw-bold'
 			=%>
+			% if ($set->description =~ /\S/) {
+				<a class="set-id-tooltip" role="button" tabindex="0" data-bs-placement="right"
+						data-bs-toggle="tooltip" data-bs-title="<%= $set->description =%>"
+						data-fallback-placements="top bottom">
+					<i class="icon fas fa-circle-info" aria-hidden="true"></i>
+					<span class="visually-hidden"><%= maketext('Assignment Description') =%></span>
+				</a>
+			% }
 		</div>
 		<div class="font-sm"><%= $status_msg %></div>
 		% if (!$set->visible && $authz->hasPermissions(param('user'), 'view_unopened_sets')) {
--- a/htdocs/js/System/system.js
+++ b/htdocs/js/System/system.js
@@ -74,7 +74,8 @@
 	// FIXME: These are really general purpose tooltips and not just in the homework sets editor.  So the class name
 	// should be chosen to better reflect this.
 	document.querySelectorAll('.set-id-tooltip').forEach((el) => {
-		if (el.dataset.bsTitle) new bootstrap.Tooltip(el, { fallbackPlacements: [] });
+		if (el.dataset.bsTitle)
+			new bootstrap.Tooltip(el, { fallbackPlacements: el.dataset.fallbackPlacements?.split(' ') || [] });
 	});
 
 	// Hardcopy tooltips shown on the Problem Sets page.


templates/ContentGenerator/ProblemSets/set_list_row.html.ep
```diff
@@ -17,9 +17,15 @@
 	<div class="ms-3 me-auto">
 		<div dir="ltr">
 			<%= link_to $display_name => $c->systemLink(url_for('problem_list', setID => $set->set_id)),
-				class => 'fw-bold set-id-tooltip',
-				data  => { bs_toggle => 'tooltip', bs_placement => 'right', bs_title => $set->description }
+				class => 'fw-bold'
 			=%>
+			% if ($set->description =~ /\S/) {
+				<a class="help-popup" role="button" tabindex="0" data-bs-placement="right"
+						data-bs-toggle="popover" data-bs-content="<%= $set->description =%>">
```
drgrice1: This should be

```suggestion
						data-bs-toggle="popover" data-bs-title="<%= $set->description =%>">
```
Bootstrap uses the `data-bs-title` attribute for tooltip contents.  You can also directly set the `title` attribute and that will be used the same.  Apparently they also honor `data-bs-content` although it is not documented and seems to result in a lot of extra padding on the tooltip which does not look good to me.

somiaj: Thanks. I too noticed the extra padding and wasn't sure how to deal with it.

I took this code from `help-popup` use elsewhere in the code. Should I update those to use `data-bs-title` too?

somiaj: Actually I don't like `data-bs-title`, it puts the content in the `title` of a card, which looks off to me, is bold and has a grey background, with only rounded corners at the top.

https://getbootstrap.com/docs/5.3/components/popovers/

Looks like popovers actually do use cards, and `title` and `content` are both described there. Though removing the padding would be nice.

drgrice1: Ahh, I didn't look closely.  You switched from a tooltip to a popover.  That uses both `data-bs-content` and `data-bs-title`.  The difference is that the title is the header in the popover, and the content is the body of the popoever.  I am not sure that I like switching to a popover for this.

drgrice1: No, don't switch the other popovers to use the title.

somiaj: Yea, I was just matching other help popovers as that way I didn't have to worry about updating the fallback position since it was already included and matched other button popovers.

I could go back to a `set-id-tooltip`. I also notice that `data-bs-custom-class` can be used to give the content a class that could be used to remove any padding.

drgrice1: Why not still use a tooltip, but not the `set-id-tooltip`?

somiaj: I don't quite follow, if I don't use `set-id-tooltip`, I would need to write other javascript to cause the tooltip to popup? Anyways, how is thi? I did use `set-id-tooltip`, and updated it to allow fallback position data attribute which just splits a space separated list.

I wasn't sure if I should make this a `<a>` tag that is a `button` and has a `tabindex="0"` or just an icon like the assignment type icon, that uses a `tabindex="-1"` for accessibility. The change I just made here is a diff to the current setup that uses an icon inside a inline block div instead.

```diff
--- a/templates/ContentGenerator/ProblemSets/set_list_row.html.ep
+++ b/templates/ContentGenerator/ProblemSets/set_list_row.html.ep
@@ -20,12 +20,13 @@
                                class => 'fw-bold'
                        =%>
                        % if ($set->description =~ /\S/) {
-                               <a class="set-id-tooltip" role="button" tabindex="0" data-bs-placement="right"
+                               <div class="d-inline-block">
+                                       <i class="set-id-tooltip icon fas fa-circle-info" tabindex="-1" data-bs-placement="right"
                                                data-bs-toggle="tooltip" data-bs-title="<%= $set->description =%>"
-                                               data-fallback-placements="top bottom left">
-                                       <i class="icon fas fa-circle-info" aria-hidden="true"></i>
+                                               data-fallback-placements="top bottom left" aria-hidden="true">
+                                       </i>
                                        <span class="visually-hidden"><%= maketext('Assignment Description') =%></span>
-                               </a>
+                               </div>
                        % }
                </div>
                <div class="font-sm"><%= $status_msg %></div>
```

Both work in my testing, so unsure which is preffered.

drgrice1: Yes, you would have to add JavaScript to activate the tooltip if you don't use the `set-id-tooltip`, but that is minor and easy to do.

I guess the `data-fallback-placements` approach is okay.  As long as it doesn't cause conflicts for Bootstrap, since that is their actual data attribute for the setting.  Also, `left` should not be in the fall back placement list.

somiaj: Doesn't bootstrap require the `data-bs-foo` now? It works in my testing, various tooltips using this, and thought not having `bs` on the name would ensure no conflict.

I found that when the icon is at the far right of the screen due to a long file set name, that `left` looks better here otherwise the tooltip is pressed right against the edge of the screen. I can remove it if desired, but to me `left` looks better than `top` or `bottom` in that case.

drgrice1: I didn't see that you omitted the `bs` in the attribute.  Okay, then that won't conflict.

drgrice1: I think that probably the `a` tag is better here.  I am not certain though.

drgrice1: I don't like `left` because that places the tooltip over the set id.  I don't see a problem with it being above or below but against the right side of the screen.

somiaj: I don't have strong feelings to argue that `left` remain, I think it looks better, but that does mean it would have to be toggled again on mobile devices. I still like the look better, but not enough to keep it around, maybe @dlglin can chime in on that.

somiaj: I removed `left` unless others suggest I add it back.

pstaabp (approved):
Looks good.

dlglin (approved):
Works for me.

drgrice1:
I think that this pull request will need to be rethought.

This is giving a new warning with the Google Chrome "Lighthouse" accessibility checker.  It says that this icon and the link it is next to are touch targets without sufficient size or spacing.  It says touch targets with sufficient size and spacing help users who may have difficulty targeting small controls to activate the targets.

Since this icon does not actually do anything other than reveal a tooltip, it is debatable that it should be in the tab order at all.  The icon to the left of the link is not in the tab order.  That is the icon that is either an open book or list, and the tooltip reveals that it is a "Regular Assignment" or "Test/Quiz".  That icon has a visually hidden span that is still read for screen reader users (it does not need to be in the tab order for that).  So the new icon adds an unnecessary tab stop for screen reader users.

dlglin:
I was going to suggest a combination of the old behaviour and the new: Revert to showing the description when mousing over the set name, but also add the information icon next to the set name to indicate that the description exists.  I'm envisioning that the icon would just be part of the set link, so mousing over either the set name or the icon would give you the description.

The issue with this approach would be for touch screen users, as it isn't obvious how to get the information indicated by the icon.  Maybe that isn't a big deal since the description is now shown on the set page.

PR #2933, opened 2026-03-16
Make `check_modules.pl` check for node version 20 or newer.

drgrice1:
Node 18 is deprecated, and our dependencies now require node 20 or newer.


somiaj (approved):

dlglin (approved):
Not tested, but should work.

PR #2937, opened 2026-03-25
Fix an npm dependency security vulnerability.

drgrice1:
Yes, this is the same one the dependabot pointed out for the main branch.


somiaj (approved):
Going to merge this to match what was done in main.

PR #2938, opened 2026-03-26
Fix Resurrect Achievement Items

### LINKED ISSUE #2907: Error in resurrect achievement items when reduced scoring is enabled
There is an error in the logic for the ResurrectHW and ResurrectGW achievement items:
https://github.com/openwebwork/webwork2/blob/3678e212b09434cb5fc9a80cdf213c4054e6fee0/lib/WeBWorK/AchievementItems/ResurrectHW.pm#L19-L22
This says that if there is a reduced scoring period, then the student can use the item during that period.  This is problematic because it then sets the due date to be 24 hours from the current time.

Consider the following scenario:  the reduced scoring date was yesterday, and the close date is in a week.  Currently if a student uses a resurrect item on the set, it will set both the reduced scoring date and the close date to tomorrow, so although the student will get an extra 24 hours at full credit, they will lose the rest of the week that they used to have for reduced scoring.

What would be the desired behaviour here?  My first thought is to say that students should only be able to use resurrect items on sets that are completely closed, and if they do they will get 24 hours at full credit, with no additional reduced scoring period.

dlglin:
This fixes a couple of issues with the resurrect achievement items in the case where reduced scoring is enabled:
1. Checks whether reduced scoring is enabled rather than checking for the existence of a reduced scoring date.
2. Fixes the messages and the logic when using the item on a set that is in the reduced scoring period.  See #2907 for the discussion of how this is implemented.

This fixes #2907.


lib/WeBWorK/AchievementItems/ResurrectGW.pm
```diff
@@ -44,18 +87,31 @@ sub use_item ($self, $set, $records, $c) {
...
-	);
+	if ($set->enable_reduced_scoring && ($set->reduced_scoring_date != $set->due_date)) {
+		return $c->maketext(
+			'This assignment has been reopened and is due on [_1].  After that date any work '
+				. 'completed will count for '
+				. $c->ce->{pg}{ansEvalDefaults}{reducedScoringValue} * 100 . '% '
+				. 'of its value until [_2]',
+			$c->formatDateTime($set->reduced_scoring_date, $c->ce->{studentDateDisplayFormat}),
+			$c->formatDateTime($set->due_date,             $c->ce->{studentDateDisplayFormat})
+		);
```
somiaj, line 115: You shouldn't add a variable to a translated string directly, instead use a substitution. I haven't tested this is formatted correctly. Seems you did this properly below, you may have just missed fixing this too.

```suggestion
		return $c->maketext(
			'This assignment has been reopened and is due on [_1].  After that date any work '
				. 'completed will count for [_2]% of its value until [_3].',
			$c->formatDateTime($set->reduced_scoring_date, $c->ce->{studentDateDisplayFormat}),
			$c->ce->{pg}{ansEvalDefaults}{reducedScoringValue} * 100,
			$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat})
		);
```

dlglin: Yeah, I realized that and changed all the other calls.  Looks like I missed this one.  It's fixed now.

lib/WeBWorK/AchievementItems/ResurrectHW.pm
```diff
@@ -18,46 +18,84 @@ sub new ($class) {
...
+			$c->ce->{pg}{ansEvalDefaults}{reducedScoringValue} * 100,
+			$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat})
+		) . $rerandomizeMessage;
+	} else {
+		return $c->maketext(
+			'This assignment has been reopened and will now close on [_1]. ',
+			$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat})
+		) . $rerandomizeMessage;
+	}
 }
```
somiaj: `$rerandomizeMessage` should also be a positional argument inside of the `maketext` call.

somiaj: Unsure if that gets tricky or there should be a conditional to check if it should be added or not so the translated strings don't have extra space in them in the case `rerandomizeMessage` is empty.

dlglin: `$rerandomizeMessage` is the result of a `maketext` call, so that string will already be translated.  I don't see how having it as a positional argument is necessary.

Is it a problem to have an extra space at the end of a string to be translated?

somiaj: My understanding is not all languages are right to left, so just appending the string to the end may not be appropriate for some translations.

In terms of the html, an extra space won't matter, but what happens if a translation doesn't understand that space is important and leaves it off. Putting the translated string as a positional argument inside the maketext call makes the use of the space clearer and allows positioning the statement in the string appropriately for the translation.

There are various places thought out the code that a string is translated and then added as a positional argument in future translations to join things together. I think this should probably do the same. I think the space thing is very minor and probably won't matter if the positional argument is empty.

somiaj: Others may have a better knowledge, what you did may be okay, I just feel it isn't. I think a positional argument is better than string concatenation. I'm unsure if there is a better approach to avoid having to write lots of cases like you have been doing in most of this.

drgrice1: It is always best to give as much as possible together for translations.  However, generally concatenations of two sentences can be done.  Usually the bidirectional algorithm will make it work even for right to left languages.  However, the space should be inserted between the translations and not inside of one of them.  The translator may not add that space in the translation, and even if they do may not put it in the correct place, not knowing that it is going to be concatenated with another translation.

Perhaps making it a positional argument inside another translation is better though.

lib/WeBWorK/AchievementItems/ResurrectHW.pm
```diff
@@ -18,46 +18,84 @@ sub new ($class) {
...
+				. 'completed will count for [_2]% of its value until [_3]. ',
+			$c->formatDateTime($set->reduced_scoring_date, $c->ce->{studentDateDisplayFormat}),
+			$c->ce->{pg}{ansEvalDefaults}{reducedScoringValue} * 100,
+			$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat})
+		) . $rerandomizeMessage;
+	} else {
+		return $c->maketext(
+			'This assignment has been reopened and will now close on [_1]. ',
+			$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat})
+		) . $rerandomizeMessage;
```
somiaj:
```suggestion
		return $c->maketext(
			'This assignment has been reopened and will now close on [_1].',
			$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat})
		) . ($rerandomizeMessage ? " $rerandomizeMessage" : '');
```

dlglin:
This now handles the case where reduced scoring is enabled for a set but the feature is disabled for the course, which also applied to the `ExtendReducedDate` and `ReducedCred` items.

lib/WeBWorK/AchievementItems/ResurrectHW.pm
```diff
@@ -16,48 +16,86 @@ sub new ($class) {
 	}, $class;
 }
 
-sub can_use ($self, $set, $records) {
+sub can_use ($self, $set, $records, $c) {
 	return $set->assignment_type eq 'default'
-		&& (after($set->due_date) || ($set->reduced_scoring_date && after($set->reduced_scoring_date)));
+		&& (after($set->due_date) || ($set->enable_reduced_scoring && after($set->reduced_scoring_date)));
```
drgrice1: Now that you have the controller object, this should be

```suggestion
		&& (
			after($set->due_date)
			|| ($c->ce->{pg}{ansEvalDefaults}{enableReducedScoring}
				&& $set->enable_reduced_scoring
				&& after($set->reduced_scoring_date))
		);
```

lib/WeBWorK/AchievementItems/ResurrectHW.pm
```diff
@@ -16,48 +16,86 @@ sub new ($class) {
...
+
+	# Change the seed for all of the problems if the set is currently closed.
+	if (after($set->due_date)) {
+		my %userProblems =
+			map { $_->problem_id => $_ }
+			$db->getUserProblemsWhere({ user_id => $set->user_id, set_id => $set->set_id });
+		for my $problem (@$records) {
+			my $userProblem = $userProblems{ $problem->problem_id };
+			$userProblem->problem_seed($userProblem->problem_seed % 2**31 + 1);
+			$problem->problem_seed($userProblem->problem_seed);
```
drgrice1, line 67: Perhaps this re-randomization should only occur if it is also past the answer date.  Is there any need for this if the answer date has not passed?

lib/WeBWorK/AchievementItems/ResurrectHW.pm
```diff
@@ -16,48 +16,86 @@ sub new ($class) {
...
+			$set->answer_date($set->due_date);
+			$userSet->answer_date($set->answer_date);
+		}
 	}
 	$db->putUserSet($userSet);
 
-	return $c->maketext(
-		'This assignment has been reopened and will now close on [_1]. Problems have been rerandomized.',
-		$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat}));
+	if ($set->enable_reduced_scoring && ($set->reduced_scoring_date != $set->due_date)) {
```
drgrice1: This should be

```suggestion
	if ($c->ce->{pg}{ansEvalDefaults}{enableReducedScoring}
		&& $set->enable_reduced_scoring
		&& ($set->reduced_scoring_date != $set->due_date))
	{
```

Otherwise if reduced scoring is disabled for the course but enabled for the set, then this could give the wrong message.

lib/WeBWorK/AchievementItems/ResurrectGW.pm
```diff
@@ -44,18 +92,30 @@ sub use_item ($self, $set, $records, $c) {
...
+			$userSet->answer_date($set->answer_date);
+		}
 	}
 	$db->putUserSet($userSet);
 
-	return $c->maketext(
-		'This assignment has been reopened and will now close on [_1].',
-		$c->formatDateTime($set->due_date, $c->ce->{studentDateDisplayFormat})
-	);
+	if ($set->enable_reduced_scoring && ($set->reduced_scoring_date != $set->due_date)) {
```
drgrice1: This should be

```suggestion
	if ($c->ce->{pg}{ansEvalDefaults}{enableReducedScoring}
		&& $set->enable_reduced_scoring
		&& ($set->reduced_scoring_date != $set->due_date))
	{
```

for the same reason as for the `ResurrectHW.pm` reward.

drgrice1 (commented):
It looks like this is getting close.

lib/WeBWorK/AchievementItems/ResurrectHW.pm
dlglin: In the case where there are limited attempts allowed on a problem a student could check their answers as many times as they want to find the correct answers, and then resurrect the set and enter the correct answers using only one attempt.

drgrice1: Ahh, I forgot about the fact that once the set is closed the student can still use the "Check Answers" button, and thus can do what you say.

drgrice1 (approved):

somiaj (approved):

PR #2939, opened 2026-03-29
Fix an issue with MathJax CHTML mode.

drgrice1:
The `switchToBSStyle` method should be called on the `commonStyles` key of the `CHTML` object rather than the object itself (which is actually a function).


somiaj (approved):
I can confirm this fixes the issue I mentioned in slack. Since it is a simple one line fix, I'm just going to merge.

PR #2943, opened 2026-04-06
show desired course ID in messages for when course does not exist

Alex-Jordan:
If you try to visit a course that does not exist (including the situation where a course has been archived and closed), you see one of these messages affected here. The change is to display the desired course ID in the message. Of course, you could see the desired course ID in the address bar. But Safari doesn't show the address bar unless you make extra effort.

This is frustrating when someone sends me a screenshot from Safari, and they are in this situation. It's often the case that they are using LTI, and did not update LTI links to a new WeBWorK course when they copied the LMS course. It would be easier to diagnose and convince people this is the issue if the messages here included the desired courseID.

### DIFF
--- a/lib/WeBWorK.pm
+++ b/lib/WeBWorK.pm
@@ -140,10 +140,11 @@ async sub dispatch ($c) {
 		# This route could have the courseID set, but does not need authentication.
 		return 1 if $c->current_route eq 'saml2_metadata';
 
-		return (0, 'This course does not exist.')
+		return (0, "The course $routeCaptures{courseID} does not exist.")
 			unless (-e $ce->{courseDirs}{root}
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks fine.

Though at first I was concerned that this might be a potential XSS vulnerability, but it is not.  If, for example, the URL contained something like `https://server.edu/webwork2/badCourseId<script>alert('hello')<%2Fscript>` the `#courseID` route capture does not get that because it only allows word characters and hyphens.  In fact doesn't match any routes and gives a 404 not found due to the changes in https://github.com/openwebwork/webwork2/pull/2691.

drgrice1:
Did you want to target this to the release candidate branch?

somiaj:
I am just going to approve and put in the release candidate branch.

somiaj (approved):

PR #2946, opened 2026-04-09
Add horizontal padding to set info icon.

somiaj:
This adds some spacing between the set link and the info icon to make it accessible on small touch screens to give enough space to touch the icon without hitting the link instead.

### DIFF
--- a/templates/ContentGenerator/ProblemSets/set_list_row.html.ep
+++ b/templates/ContentGenerator/ProblemSets/set_list_row.html.ep
@@ -20,7 +20,7 @@
 				class => 'fw-bold'
 			=%>
 			% if ($set->description =~ /\S/) {
-				<a class="set-id-tooltip" role="button" tabindex="0" data-bs-placement="right"
+				<a class="set-id-tooltip px-1" role="button" tabindex="0" data-bs-placement="right"
 						data-bs-toggle="tooltip" data-bs-title="<%= $set->description =%>"
 						data-fallback-placements="top bottom">
 					<i class="icon fas fa-circle-info" aria-hidden="true"></i>


somiaj:
Added #2947 as an alternative to this approach and remove the set icon altogether. I'll let approvals vote for which approach others prefer.

templates/ContentGenerator/ProblemSets/set_list_row.html.ep
```diff
@@ -20,7 +20,7 @@
 				class => 'fw-bold'
 			=%>
 			% if ($set->description =~ /\S/) {
-				<a class="set-id-tooltip" role="button" tabindex="0" data-bs-placement="right"
+				<a class="p-1 set-id-tooltip" role="button" tabindex="-1" data-bs-placement="right"
```
drgrice1: This should use `px-1` instead of `p-1`.  There is no need for vertical padding here.  Only horizontal padding is needed.

drgrice1 (commented):
Adding `tabindex="-1"` is certainly incorrect.  At least in my testing with Gnome Orca that makes the set description completely inaccessible with the screen reader and only keyboard controls.

To be quite frank, I think that this should be reverted to the way it was with the tooltip on the set id link.  Yes, it does make the tooltip practically inaccessible for mobile devices, but for all other cases it works well.

My vote is for neither of pull requests #2946 or #2947, and to put this back to the way it was and has been for a long time.  The set description is now on the problem set page for mobile devices.

drgrice1:
Actually I think this pull request is the one to go with, but with it only adding `px-1` and leaving `tabindex="0"`.

In thinking about it more, I kind of like that the icon makes it clear that there is more information.

Also, testing with Gnome Orca, it isn't even reading the tooltips when they are on the links.  Although, I suspect that is some issue with Gnome orca and that the proprietary screen readers would.

templates/ContentGenerator/ProblemSets/set_list_row.html.ep
somiaj: I debated between `p-1`, `ps-1`, and `px-1`, and thought the extra padding didn't affect the layout of the page in a noticeable way, but made the button slightly bigger vertically to make it easier to click on a small phone. I will update as suggested, just explaining my thought process.

templates/ContentGenerator/ProblemSets/set_list_row.html.ep
```diff
@@ -20,7 +20,7 @@
 				class => 'fw-bold'
 			=%>
 			% if ($set->description =~ /\S/) {
-				<a class="set-id-tooltip" role="button" tabindex="0" data-bs-placement="right"
+				<a class="px-1 set-id-tooltip" role="button" tabindex="0" data-bs-placement="right"
```
drgrice1: Please move `px-1` to the end of the class list.  I think it is best to keep our own classes that generally have more meaning as to what an element is and that apply styles defined in or css at the front.  But generally I think Bootstrap styles for padding only have little significance and should be at the end.

drgrice1 (approved):

PR #2950, opened 2026-04-14
Fix the `change_user_id` script for the case that a user does not have a password.

drgrice1:
It is now the case that a user is not required to have a password record in the database.  However, the `change_user_id` script was never updated for that case.  So an exception is thrown if the script is used for a user that does not have a password record (for example a user created via LTI authentication).

Note that perltidy has also been run on the file. The script does not have the `.pl` extension and so the workflow doesn't check this file and the `run-perltidy.pl` script doesn't attempt to format it.  Hide whitespace changes to see the important change (there is also some clean up of the comments at the beginning of the file).

Also fix the check for the type of a database record in the `checkArgs` method of `lib/WeBWorK/DB.pm`.  Just because a variable is a `ref` does not mean that it is an object for which the `isa` method can be called. The proper check is `blessed $obj && $obj->isa('Package')`.  Note the defined check was unnecessary in any case. This prevented the error in the script from giving more useful information.  Instead it complained about the inability to call `isa` when it should have given the intended error message from this check.

### DIFF
--- a/bin/change_user_id
+++ b/bin/change_user_id
@@ -1,18 +1,19 @@
 #!/usr/bin/env perl
+
+# Sometimes a webwork user id changes.  This script transfers the webwork data for the old user_id to the new user_id.
 #
-#Sometimes a webwork user id changes.  This script transfers the webwork data for the old user_id to the new user_id
-# Update database tables
-#user
-#permission
-#password
-#key
-#set_user
-#problem_user
-#set_locations_user
-#global_achievement_user
-#achievement_user
+# The script updates the following database database tables
+#     user
+#     permission
+#     password
+#     key
+#     set_user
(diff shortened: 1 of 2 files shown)


somiaj:
Just double checking this script doesn't need any tweak to deal with the accommodation factor?

drgrice1:
It does not.  It uses a database object and obtains the database record for the original userId from the database before adding it for the new userId.  So the record will have the non null `accomodation_time_factor` in it to begin with, and it won't cause an issue.

somiaj (approved):

Alex-Jordan:
I did not know there even was such a script.

Alex-Jordan (approved):

PR #2952, opened 2026-04-17
Remove PRINT_FILE_NAMES_FOR and PRINT_FILE_NAMES_PERMISSION_LEVEL

dlglin:
This removes the special PG environment variables `PRINT_FILE_NAMES_FOR` and `PRINT_FILE_NAMES_PERMISSION_LEVEL`.

In defaults.config the value of `PRINT_FILE_NAMES_PERMISSION_LEVEL` is set to the value of the `print_path_to_problem` permission, so now that permission is just checked directly.

I don't see the use of `PRINT_FILE_NAMES_FOR`.  When a new course is created it is hardcoded to add all users with permission level of 'professor' or higher to the array, but those users will already be able to view file names via the `print_path_to_problem` permission.

The motivation behind this was to remove the one remaining setting that is added to course.conf when a new course is created.  Now that file contains no settings by default.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -797,7 +797,7 @@ $authen{admin_module} = ['WeBWorK::Authen::Basic_TheLastOption'];
 	can_check_and_submit_answers                    => "ta",
 	can_use_show_me_another_early                   => "ta",
 	create_new_set_version_when_acting_as_student   => undef,
-	print_path_to_problem                           => "professor", # see "Special" PG environment variables
+	print_path_to_problem                           => "professor",
 	always_show_hint                                => "professor",
 	always_show_solution                            => "professor",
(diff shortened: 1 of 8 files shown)


drgrice1 (commented):
With this pull request there are no longer any valid `$courseOptions`.  The only thing this has was used for was for writing the `course.conf` file, and the only thing that was left in it was the `PRINT_FILE_NAMES_FOR` option.  So it should be removed from the code entirely.  Both in the `CourseManagement.pm` file and in the `CourseAdmin.pm` file.

Furthermore, the `writeCourseConf` method in `CourseManagement.pm` doesn't need any arguments anymore other than the file handle to write to.

I have already done this, and will put a pull request in to this branch.

drgrice1 (approved):

somiaj (approved):

PR #2964, opened 2026-05-02
Better keep track of which sets are complete in achievements.

somiaj:
Store the setID of all completed sets in the `globalData` hash when evaluating achievements. This allows achievements to use this data vs just counting the number of completed sets. One use case is being able to exclude optional sets, such as review sets, from some achievements without completely excluding them from all achievements.

In addition saving all the setIDs can avoid a double counting completed sets, as there is currently no check to ensure a set is not counted multiple times, and a student can earn the complete N sets achievements by submitting a problem on the same completed set over and over.

This also cleans up some typos in the comments and formats them better.

### DIFF
--- a/lib/WeBWorK/AchievementEvaluator.pm
+++ b/lib/WeBWorK/AchievementEvaluator.pm
@@ -23,17 +23,17 @@ sub checkForAchievements ($problem_in, $c, %options) {
 	# Make a copy of the problem so that local modifications do not persist.
 	our $problem = $db->newUserProblem($problem_in);
 
-	# Date and time for course timezone (may differ from the server timezone)
-	# Saved into separate array
+	# Date and time for course timezone (may differ from the server timezone).
+	# Saved into separate array:
 	# https://metacpan.org/pod/DateTime
 	my $dtCourseTime = DateTime->from_epoch(epoch => time(), time_zone => $ce->{siteDefaults}{timezone} || 'local');
 
-	# Set up variables and get achievements
+	# Set up variables and get achievements.
 	my $cheevoMessage = $c->c;
 	my $user_id       = $problem->user_id;
 	my $set_id        = $problem->set_id;
 
-	# exit early if the set is to be ignored by achievements
+	# Exit early if the set is to be ignored by achievements.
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/AchievementEvaluator.pm
```diff
@@ -132,19 +132,24 @@ sub checkForAchievements ($problem_in, $c, %options) {
 		}
 	}
 
-	$globalData->{completeSets}++ if ($allcorrect);
+	# Store the setID of all completed sets so they can be used in achievements and not double counted.
+	$globalData->{completedSetIds} = {} unless defined $globalData->{completedSetIds};
```
drgrice1: This should be

```suggestion
	$globalData->{completedSetIds} //= {};
```

lib/WeBWorK/AchievementEvaluator.pm
```diff
@@ -132,19 +132,24 @@ sub checkForAchievements ($problem_in, $c, %options) {
...
+	if ($allcorrect) {
+		$globalData->{completeSets}++ unless $globalData->{completedSetIds}{$set_id};
```
drgrice1: Always use pre increment instead of post increment unless post increment is what is really needed.  I know it was post increment before, but time to fix that.  So make this

```suggestion
		++$globalData->{completeSets} unless $globalData->{completedSetIds}{$set_id};
```

Pre increment is more efficient because it doesn't need to save and return the original value.

drgrice1 (approved):
Other than minor code quibbles this looks good.

Alex-Jordan:
If a student completes a set, and then that setID is stored as completed because of this commit, and then the instructor unassigns/reassigns the set, will everything be working here as expected? I don't recall what's going on with the `globalData` hash.

somiaj:
The `globalData` is stored in the achievements table and is not modified by assigning/unassiging sets. If the set was unassigned, then reassigned, they would still be counted as having completed it if it a set with that id was completed before it was unassigned.

This was also true before this change, unassiging a set that was complete would not reduce the completed counter. Overall achievements are very fragile to stuff like this. This is just a little more robust than before because it won't allow double counting a set even in the case you describe.

Alex-Jordan (approved):

PR #2965, opened 2026-05-03
Add extension time factor setting.

somiaj:
Add `$achievementExtensionFactor` setting that is used to configure the length of extensions. This works as a multiplicative factor, by multiplying the base time (either 24 or 48 hours) by the factor. The extension time is always rounded to the nearest hour, and cannot be less than a single hour (two hours for the super extensions).

This affects all items that have an extension time.

* ExtendDueDate
* ExtendDueDateGW
* ExtendReducedDate
* ReducedCred
* RessurectGW
* RessurectHW
* SuperExtendDueDate
* SuperExtendReducedDate

### DIFF
--- a/lib/WeBWorK/AchievementItems/AddNewTestGW.pm
+++ b/lib/WeBWorK/AchievementItems/AddNewTestGW.pm
@@ -6,14 +6,16 @@ use Mojo::Base 'WeBWorK::AchievementItems', -signatures;
 use WeBWorK::Utils           qw(x);
 use WeBWorK::Utils::DateTime qw(between);
 
-sub new ($class) {
+sub new ($class, $c) {
 	return bless {
 		id          => 'AddNewTestGW',
 		name        => x('Oil of Cleansing'),
-		description => x(
-			'Unlock an additional version of a test.  If used before the close date of '
-				. 'the test this will allow you to generate a new version of the test.'
-		)
+		description => [
+			x(
+				'Unlock an additional version of a test.  If used before the close date of '
+					. 'the test this will allow you to generate a new version of the test.'
+			)
+		]
 	}, $class;
 }
 
--- a/lib/WeBWorK/Utils/DateTime.pm
+++ b/lib/WeBWorK/Utils/DateTime.pm
@@ -3,7 +3,8 @@ use Mojo::Base 'Exporter', -signatures;
 
 use DateTime;
 use DateTime::TimeZone;
-use Time::Zone qw(tz_offset);
+use Time::Zone     qw(tz_offset);
+use WeBWorK::Utils qw(max);
 
 our @EXPORT_OK = qw(
 	before
@@ -13,6 +14,7 @@ our @EXPORT_OK = qw(
 	getDefaultSetDueDate
 	parseDateTime
 	timeToSec
+	getExtensionTime
 );
 
 sub before ($time, $now = time) { return $now < $time }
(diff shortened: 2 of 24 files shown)


lib/WeBWorK/AchievementItems/AddNewTestGW.pm
```diff
@@ -3,14 +3,13 @@ use Mojo::Base 'WeBWorK::AchievementItems', -signatures;
...
 use WeBWorK::Utils::DateTime qw(between);
 
-sub new ($class) {
+sub new ($class, $c) {
 	return bless {
 		id          => 'AddNewTestGW',
-		name        => x('Oil of Cleansing'),
-		description => x(
+		name        => $c->maketext('Oil of Cleansing'),
+		description => $c->maketext(
```
drgrice1: Why did you make this change of translating the name and description at construction time instead of just marking them for translation now, and translating them later when used?  In terms of efficiency, this is not better.  This means that `maketext` is called to look up the translation even if the string is not actually used.

somiaj: I didn't consider that there was an efficiency difference. The main thing was my change made the `description` have an input variable to the translation, so instead of making that an array ref I changed those calls to use `maketext`. I just updated all of those without realizing it was an efficiency hit.

I'll go revert all that and make the `description =>` an array ref.

lib/WeBWorK/AchievementItems/AddNewTestGW.pm
```diff
@@ -6,14 +6,16 @@ use Mojo::Base 'WeBWorK::AchievementItems', -signatures;
 use WeBWorK::Utils           qw(x);
 use WeBWorK::Utils::DateTime qw(between);
 
-sub new ($class) {
+sub new ($class, $c) {
```
drgrice1, line 9: I don't think that `$c` needs to be passed in anymore, since you reverted the `maketext` call.

somiaj: The controller was added for the extension achievements to have access to the course environment for this new setting to compute extension time. I only started translating because I had the controller and didn't think about efficiency concerns.

Though this does mean a similar computation gets used a few times, it was most straight forward to add that computation to the `new` method, but I guess I could compute the times and time text once then pass it to the `new` methods.

drgrice1: I forgot you needed it to get the new setting.  That is fine.

lib/WeBWorK/Utils/DateTime.pm
```diff
@@ -311,4 +333,18 @@ and issue warning.
...
+Usage: C<getExtensionTime($c, $base)>
+
+Determines the length of an extension based on the course environment extension
+time factor setting and C<$base> time. The C<$base> time is the number of days
+for the base extension. The extension time is rounded to the nearest full hour
+and cannot be less than C<$base> hours.
+
+Returns an array consisting of the extension time in seconds and a translated
+string stating the length of the extension. The translated string will state
+the number of days and hours if greater than or equal to 3 days, otherwise
```
drgrice1, line 347: I think this should be

```suggestion
the number of days and hours if greater than or equal to 2 days, otherwise
```

based on the code.

drgrice1 (approved):
This looks good to me.

It is nice that the new time setting allows the `SuperExtendDueDate.pm` and `SuperExtendReducedDate.pm` packages to extend the `ExtendDueDate.pm` and `ExtendReducedDate.pm` package, respectively.  Nice code refactor!

lib/WeBWorK/Utils/DateTime.pm
somiaj: I did consider that, but went with "greater than equal to 3 days", because "2 days and 12 hours" will actually print as "60 hours".

Alex-Jordan (approved):

PR #2967, opened 2026-05-06
Fix colors on the proctor login page for dark mode.

drgrice1:
The colors for this were inline style.  Colors cannot be inline anymore since the server cannot detect if the user will have the browser in dark mode or not.  So classes are used instead.


somiaj (approved):

somiaj:
Would it be worth just using default bootstrap colors here, so 'warning' instead of the custom yellow, and so forth?

Alex-Jordan (approved):

PR #2968, opened 2026-05-06
Fix grading proctored tests that require proctor authorization to do so.

drgrice1:
The `can_recordAnswers` method of the `WeBWorK::ContentGenerator::GatewayQuiz` package can not call any of the `WeBWorK::ContentGenerator::GatewayQuiz` methods because it is called directly by the `WeBWorK::ContentGenerator::LoginProctor` module without a `WeBWorK::ContentGenerator::GatewayQuiz` object. Attempting to do so will cause an exception.

This is a bit annoying as the conditions of the `can_gradeUnsubmittedTest` method (including the conditions of the `can_showProblemGrader` method it calls) must be directly used by the `can_recordAnswers` method. There is no way around that though.

This definitely should be considered for a hotfix.


somiaj (approved):
Fixes the issue.

I too think a hotfix is worth it. Should #2969 be included in that?

drgrice1:
Yeah, that could be a hotfix as well.

drgrice1:
This is certainly more serious than #2968 though as this prevents grading of a proctored test at all in the case that proctor authorization is required at the end of the test.

Alex-Jordan (approved):

Issue #2962, opened 2026-04-29
Grading protocred test for student dosen't remove proctor status.


somiaj:
The feature I added to "Grade Test for <student>" for ungraded tests after close date needs to change test from proctored to normal with a proctored test.

I had a student just do a proctored test, but they didn't hit "Grade Test" afterwards and just closed their laptop and left. I was able to grade their tests using the "Grade Test for <student>" option, but now every time I access the test version it still requires proctored authentication.

Just putting this here as it could be a while until I can look at it, or maybe someone else can look at it too.

drgrice1:
There is an even bigger problem here.  I am assuming that you have the `Proctor Authorization Type` set to `Only start`, because if you have the test set to `Both start and grade` then there is an exception when attempting to grade a proctored test.  This is caused by the addition of the `$c->can_gradeUnsubmittedTest` call to the `can_recordAnswers` method of the `GatewayQuiz.pm` package.  The `LoginProctor.pm` package calls the `can_recordAnswers` method directly without a `WeBWorK::ContentGenerator::GatewayQuiz` object, passing itself (a `WeBWorK::ContentGenerator::LoginProctor` object) for `$c`.  The call `$c->can_gradeUnsubmittedTest` then causes and exception becuase the `LoginProctor` object doesn't have that method.  So this means that any proctored test that requires proctor authorization to grade can not be graded at all.

PR #2969, opened 2026-05-06
Make sure that a proctored test becomes unproctored when graded by instructor.

drgrice1:
When an instructor uses the "Grade Test for" button and the test is on its final submission, currently the grading occurs, but the test is still set as a proctored test.

To fix this the condition that determines if the assignment type should be changed needs to check the `$can{recordAnswers}` value in the case that the `$userID` and `$effectiveUserID` are different.  In that case `$can{recordAnswers}` is the result of the `can_recordAnswers` method which will be true if either the user has the
`record_answers_when_acting_as_student` permission or the user can grade an unsubmitted test.

This fixes issue #2962.

The `GatewayQuiz.pm` module is really so convoluted at this point that it is really becoming impossible to do anything with, and is in desperate need of a complete overhaul and rewrite. Any time any new feature is added to the module or change is made to the module it is almost impossible to go through all of the possibilities and ensure you haven't broken something.

Note that to test this make sure you have the test set with a `Proctor Authorization Type` of `Only Start`.  This will work if the type is `Both Start and Grade`, but you will need #2968 to test that.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1245,7 +1245,7 @@ async sub pre_header_initialize ($c) {
 					($userID eq $effectiveUserID && !$can{recordAnswersNextTime})
 					|| (
 						$userID ne $effectiveUserID
-						&& $authz->hasPermissions($userID, 'record_answers_when_acting_as_student')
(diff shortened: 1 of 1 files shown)


somiaj (approved):
Fixes the issue.

Alex-Jordan (approved):

PR #2971, opened 2026-05-11
Use custom btn-ww-info in UI.

somiaj:
Since `btn-info` is used for PG feedback buttons, modifying this button changes the colors seen in the PG feedback button. This configures and uses a custom `btn-ww-info` class instead of `btn-info` to allow theming and to not change the color of the PG feedback buttons.

Due to the two new variables `$ww-info` and `$ww-info-text` this will break the building of any custom theme. I have opted to do not put in default values for this so anyone who has created a custom theme will see the breakage and realize they need to update their `_theme-colors.css` override.


somiaj:
Closing in favor of updating PG to not use bootstrap colors for its buttons.

PR #2974, opened 2026-05-11
Fix colors on the proctor login page for dark mode. (Bootstrap alert aproach)

drgrice1:
The colors for this were inline style.  Colors cannot be inline anymore since the server cannot detect if the user will have the browser in dark mode or not.  So this switches to using Bootstrap alerts instead.

This is an alternate approach to #2967.

One minor problem with this approach is that in the `math4-yellow` theme the difference between an `alert-primary` and an `alert-warning` is somewhat subtle.  Here is a screenshot demonstrating the two together with this pull request:
proctor-login
The first alert in that image is an `alert-warning` and the second is an `alert-primary`.

### DIFF
--- a/templates/ContentGenerator/LoginProctor.html.ep
+++ b/templates/ContentGenerator/LoginProctor.html.ep
@@ -16,14 +16,14 @@
 % # Print a message about submission times if we're submitting an answer.
 % if (param('submitAnswers')) {
 	% my $dueTime = $userSet->due_date;
-	% my ($color, $msg) = ('#ddddff', '');
+	% my ($alertClass, $msg) = ('alert-primary', '');
 	% if ($dueTime + $ce->{gatewayGracePeriod} < $submitTime) {
-		% $color = '#ffffaa';
+		% $alertClass = ' alert-warning';
 		% $msg = maketext('The time limit on this assignment was exceeded. The assignment may be checked, '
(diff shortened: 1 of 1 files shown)


somiaj:
Seems that alerts looking similar might also be an issue on the ProblemSet page or other pages with multiple alerts.

My thoughts (Since I requested this) is this approach is preferable, let's stick with the bootstrap colors for consistency vs have some pages use their own.

somiaj (approved):

drgrice1:
This isn't really a problem elsewhere.  Just because we don't use `alert-primary` anywhere else.  This could be switched to using `alert-secondary` or `alert-info`, both of which are similar in the `math4-yellow` theme, but quite different from `alert-warning`.  The `math4-yellow` theme colors could also be adjusted.  At this point only the `primary` and `info` colors are changed from the Bootstrap defaults.

Alex-Jordan:
I would go with this one. But since @drgrice1 uses the yellow theme, I defer to you if the subtleness is too much of a concern. I'll approve both and let you merge whichever.

Alex-Jordan (approved):

drgrice1:
I think I also prefer this approach.  The subtle difference for the `math4-yellow` theme is enough for me.  Particularly for a page that isn't used much.

PR #2979, opened 2026-05-19
Remove the `Demo` set.

drgrice1:
This set has nothing of value in it anymore.  The problems are really not what we want problem authors to use as a model for new problems.

It is time for this set to go.


somiaj (approved):
Agreed. They should be using the sample problems.

pstaabp (approved):

PR #2981, opened 2026-05-19
Fix the test nav.

drgrice1:
I used the wrong thing in #2975.  Clearly I didn't test this.  Bad,  bad Glenn.


somiaj (approved):
Fixes the bug. Merging as this fixes a bug created by the previous merge.

PR #2984, opened 2026-05-21
Cleanup modelCourse achievements.

somiaj:
Cleanup and update the default achievements in the modelCourse.

* Remove the same comment from the bottom of all evaluators. This is hard to maintain, instead put all the information in achievements_readme.txt and point to this file from other evaluators.

* Cleanup the achievement evaluators:
  * Simplify the code, don't use else statements when there was a return inside the previous if. Clean up some of the logic.
  * Take advantage of the new $globalData->{completedSetIds} hash to avoid looping through all problems to determine if a set is completed.
  * Ran perltidy on them using an 80 character limit.

* Remove some evaluator duplication. The complete n problems or n sets can all be done using a single evaluator, since the only difference is the $maxCounter setting from the achievement.

* Remove all unused achievement evaluators. There are lots of these that are really just the same example achievement over and over again to see if particular problems from particular sets were completed. The images are still available for helping people think of names, but the evaluators don't need to be included. Instead provide four example evaluators for completing specific problems and/or sets.

* Add perltidy of .at achievement files to the run-perltidy dev_script and the github workflow.


somiaj:
Okay, I have tested the most of the achievements all work and fixed a few issues/typos I found. Only ones I didn't test were the "Finish after 8 hour break", "Finish over a 48 hour period", and and ones that require finishing late at night or early in the morning. So this is now fully ready for review.

Alex-Jordan:
Could we move the distribution achievements to the assets folder in some way?  Then set up symlinks in `templates/achievements` and `html/achievements` to the right place with the assets folder tree?

And then include these symlinks in the symlink verification business that is part of doing a course upgrade.

The goal is to make it so future impovements of achievement files automatically are experienced by all courses, even old courses or clones of old courses.

There would be something to work out for when an old course still has an actual achievement file in `templatex/achievements` that clashes with the central assest folder version...

drgrice1:
Ideally we could even entirely eliminate the model course!

somiaj:
One issue to work around with using symlinks, is then users cannot add their own evaluators easily because of now being linked to a shared directory. We would probably need some override system where achievements are looked for in both a system location for default ones and a local location for user achievements and overrides.

I think if we decide to change where achievements are saved, and setup some user override system, that should be a PR separate of this cleanup, and not be part of this cleanup.

drgrice1:
I agree.  Limit this pull request to the cleanup.

courses.dist/modelCourse/templates/achievements/achievement_readme.txt
```diff
@@ -1,89 +1,112 @@
...
+    This data will be accessible from *every* achievement and is unique to
+    the user.  There are three variables stored in this hash that are
+    maintained by the system and should not be modified.
+    - $globalData->completeSets:
+        This is the number of sets which the student has earned 100% on.
+    - $globalData->completedSetIds:
+        This is a hash of all set ids which the student has earned 100% on.
+    - $globalData->completeProblems:
+        This is the number of problems which the student has earned 100% on.
+    - $globalData->prev_level_points:
```
drgrice1: The hash key access syntax should be fixed in these.  I know that this has been incorrect for a long time, but the `completedSetIds`, `completeProblems`, and `prev_level_points` are not object methods, and so attempting to access these values as documented would cause an error.  The actual achievements that use them do use them correctly, but it should be documented correctly here.  The documentation in the `.at` files actually had these correctly shown.

courses.dist/modelCourse/templates/achievements/back_for_more.at
```diff
@@ -1,88 +1,26 @@
...
-
-# initialize the time thing
+# This achievement is awarded if a student has worked on one problem after
+# an eight hour break. Uses $localData to save last attempt time.
+#
+# See [templates]/achievements/achievement_readme.txt for details.
 
-$localData->{lastattempttime} = time() unless $localData->{lastattempttime};
+# Initialize saved $localData for this achievement.
+$localData->{lastattempttime} //= time();
```
drgrice1: The parentheses on `time` is not needed.  There are several achievement evaluators that do this.  Please, go through and remove the parenthesis from all of them.

courses.dist/modelCourse/templates/achievements/challenge_eight.at
```diff
@@ -1,88 +1,23 @@
...
+my %validproblems = ('Challenge' => { '8' => 1 });
 
+# Checks to see if a valid problem was solved.
+if ($validproblems{ $problem->set_id }
+	&& $validproblems{ $problem->set_id }{ $problem->problem_id }
+	&& $problem->status == 1
+	&& $problem->num_correct == 1)
+{
+	# Update challenger counter.
+	$globalData->{challengerCounter}++;
```
drgrice1: Change this to

```suggestion
	++$globalData->{challengerCounter};
```

Go through all of the achievement evaluators and make this change where post-incrementation is used and shouldn't be.

courses.dist/modelCourse/templates/achievements/challenger.at
```diff
@@ -1,78 +1,9 @@
...
-#  ($globalData) so that the achievement progress bar works correctly.  
+# This is a "solved n challenge problems" achievement.  Notice how we update
+# $counter (even though the appropriate variable is already given to use in
+# $globalData) so that the achievement progress bar works correctly.
+#
+# See [templates]/achievements/achievement_readme.txt for details.
 
-# Update counter from globalData
+# Update counter from $globalData.
 $counter = $globalData->{'challengerCounter'};
```
drgrice1: Remove the quotes on `challengerCounter`.  The aren't needed.

courses.dist/modelCourse/templates/achievements/complete_n_problems.at
```diff
@@ -0,0 +1,11 @@
+# This is a "solved n problems" achievement.  This evaluator can be assigned
+# to multiple different achievements since it uses $maxCounter to determine
+# if the desired number of problems to be completed. Notice how we update
+# $counter (even though the appropriate variable is already given to use in
+# $globalData) so that the achievement progress bar works correctly.
+#
+# See [templates]/achievements/achievement_readme.txt for details.
+
+# Update counter from $globalData.
+$counter = $globalData->{'completeProblems'};
```
drgrice1: Remove the quotes on `completeProblems`.  They aren't needed.

courses.dist/modelCourse/templates/achievements/complete_n_sets.at
```diff
@@ -0,0 +1,11 @@
+# This is a "complete n sets" achievement.  This evaluator can be assigned
+# to multiple different achievements since it uses $maxCounter to determine
+# if the desired number of problems to be completed. Notice how we update
+# $counter (even though the appropriate variable is already given to use in
+# $globalData) so that the achievement progress bar works correctly.
+#
+# See [templates]/achievements/achievement_readme.txt for details.
+
+# Update counter from $globalData.
+$counter = $globalData->{'completeSets'};
```
drgrice1: Remove the quotes on `completeSets`.  They aren't needed.

drgrice1 (commented):
This generally looks good, but there is some more clean up that should be done while we are at it.

It would be good to add `.at` files to those that are formatted by perltidy when the `bin/dev_scripts/run-perltidy.pl` file is run.  This can be accomplished by changing line 93 of that file to `return unless $path =~ /\.p[lm]$/ || $path =~ /\.t$/ || $path =~ /\.at$/;`.

It would also be good to check these in the formatting workflow.  Change line 29 of `.github/workflows/check-formats.yml` to `perltidy --pro=./.perltidyrc -b -bext='/' ./**/*.p[lm] ./**/*.t ./**/*.at && git diff --exit-code` to make this happen.

somiaj:
In adding the `.at` files to the perltidy scripts and workflow. What would be needed to add the `-l 80` for these files? Would a separate line be needed? I ran these with the 80 character limit to match what the pg perl tidy is doing and these files to be easier to see in the editor on small screens.

somiaj:
Applied all of @drgrice1 suggested changes.

drgrice1:
With the `run-perltidy.pl` script and the workflow, you really can't set a separate line width.  You would have to use a separate `.perltidyrc` file, and that isn't worth it.  You can already use the `run-perltidy.pl` script on these files by passing the file name, and I did so.  It doesn't change any of the current formatting.  `perltidy` generally doesn't consolidate onto one line if the code is already broken onto multiple lines.  So the 120 character limit is fine for this.

It is better that general formatting be enforced for repository files.

somiaj:
Okay, .at files added to the `run-perltidy` and github workflow.

drgrice1 (approved):
Looks good.

Alex-Jordan (approved):

PR #2989, opened 2026-05-27
Switch `package-lock.json` to lockfile version 3.

drgrice1:
Lockfile version 3 has been supported since node version 18 (and npm version 7).  So there is no need to stay with lockfile version 2 anymore. The only real change it seems is that the `dependencies` section is removed. This means that the file is about 1500 lines shorter!


somiaj (approved):
fwiw outside of the warnings node 20 is still working.

Alex-Jordan (approved):

drgrice1:
Yeah, node 20 supports lockfile version 3 too.

PR #2994, opened 2026-06-01
Add a course selection step in content item selection for LTI 1.3.

drgrice1:
If the new `$LTI{v1p3}{allowCourseSelection}` option is set to 1 and an instructor attempts to sign in via LTI 1.3 authentication from an LMS course that is not in the LTI course map for the site, then a list of webwork courses will be compiled that have matching LTI 1.3 parameters, that have a user that works for the LMS user with email_address set that matches the email address sent from the LMS, and such that the user has the access_instructor_tools and modify_problem_sets permissions. So it is important that the email_address be set in the webwork courses for the user. Also note that the LTI user id selection parameters (`preferred_source_of_username`, `fallback_source_of_username`, strip_domain_from_email`, and `lowercase_username`) have to be the same for all matching courses (effectively the settings from the first course found are used).
    
Then the user will be presented a list of courses to choose from (or if only one matching course is found that course will be used). When the user selects a course (or only on course matches), it will be automatically added to the LTI course map, and then content selection will continue as usual with that course.
    
For this to work, the user is authenticated into the first matching course. Then if the user chooses a different course, that authentication is transfered to the selected course. Authentication should work in any of them since all of the matching courses have the same LTI 1.3 parameters and the user in each course has the same user id, email address, and permissions.
    
If no matching courses are found or if the `$LTI{v1p3}{allowCourseSelection}` option is set to 0, then the context id is shown as before.

Note that the `$LTI{v1p3}{allowCourseSelection}` is necessarily a site wide option that cannot be configured per course.

Note that this is built on top of #2993.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -242,6 +242,8 @@ $LTIMassUpdateInterval = 86400;
 	#'LMSManageUserData',
 	#'LTI{v1p1}{round_score_digits}',
 	#'LTI{v1p1}{BasicConsumerSecret}',
+	#'LTI{v1p3}{ignoreMissingSourcedID}',
+	#'LTI{v1p3}{autoSyncSetDatesToLMS}',
 	#'LTI{v1p3}{PlatformID}',
 	#'LTI{v1p3}{ClientID}',
 	#'LTI{v1p3}{DeploymentID}',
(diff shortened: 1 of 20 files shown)


drgrice1:
There was an issue with the course selection that I just fixed.  Basically, if the user is shown the course selection page, but submits without actually selecting a course, then it just went on to content selection using the first found matching course.  This now requires a selection before proceeding.

Alex-Jordan (approved):

PR #2995, opened 2026-06-01
Restrict LTI authentication by LTI course map.

drgrice1:
This was requested by @dlglin, and restricts authentication into a webwork course to only users in the LMS course that matches the context id of the webwork course set in the LTI course map for the site.

If the new `$LTI{v1p3}{restrictAuthenticationByCourseMap}` option is set to 1, then this restriction is in effect.

This is built on top of #2993 and #2994.  So you can test all of these LTI pull requests with this one.  This is the final planned pull request for LTI 1.3 at this point.


Alex-Jordan (approved):

dlglin (approved):
Everything I've tested works.

PR #2958, opened 2026-04-22
Update the docker build.

drgrice1:
Update to node 24 in the build instead of node 20.
    
Update the Perl::Tidy version to the version currently used in development.

Don't install the `Statistics::R::IO` module.  That is not used anymore.

Update to the mariadb 10.11 docker image for the database volume. That should have been done before when I updated the docker build to Ubuntu 24.  That matches the version of mariadb in Ubuntu 24.

Remove the `--db-layout` option that was passed to the `addcourse` script.  That is not valid for the script anymore.

<s>Note that this depends on #2957, since the docker build will fail without the bug in the `addcourse` script fixed.</s>

PR #2996, opened 2026-06-01
Switch the docker build to Ubuntu 26.

drgrice1:
The `DBD::MariaDB` package is provided by the Ubuntu `libdb-mariadb-perl` package again. So that does not need to be installed from cpan anymore.
    
Also change the mariadb docker image to version 11.8 to match the version used by Ubuntu 26.
    
The `libio-compress-perl` Ubuntu package is a dependency of the `Archive::Zip::SimpleZip` Perl package that is needed since the Perl packages it provides fail to build when installed from cpan. It may be that they just need some other build dependency, but it is better to install the Ubuntu packages anyway.
    
Several dependencies of the Perl `Net::SAML2` package were added to speed up the build some and avoid these being installed from cpan.
    
Remove `install` from the `cpanm` command.  That isn't a command or option for the `cpanm` command and results in the `install` perl package being installed which isn't needed.
    
The `pgfsys-dvisvgm-bbox-fix.patch` file is not needed for Ubuntu 26.  Its changes are included in upstream file installed on the system.
    
Fix the comments about executing `docker compose`.  That is no longer executed as `docker-compose`.
    
Update the `docker-entrypoint.sh` to use `libpapers2`.  The `libpapers1` package no longer exists.
    
The `newgrp` command used in the `docker-entrypoint.sh` script is no longer is installed by default.  The `util-linux-extra` package is needed for that.

Note that this is built on top of #2958.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN echo Cloning branch $PG_BRANCH branch from $PG_GIT_URL \
 
 # We need to change FROM before setting the ENV variables.
 
-FROM ubuntu:24.04
+FROM ubuntu:26.04
 
 ENV WEBWORK_URL=/webwork2 \
 	WEBWORK_ROOT_URL=http://localhost:8080 \
@@ -79,12 +79,17 @@ RUN apt-get update \
 	libclass-tiny-antlers-perl \
 	libclass-tiny-perl \
 	libcpanel-json-xs-perl \
+	libcrypt-dev \
 	libcrypt-jwt-perl \
+	libcrypt-openssl-x509-perl \
 	libcryptx-perl \
 	libdata-dump-perl \
 	libdata-structure-util-perl \
 	libdatetime-perl \
+	libdatetime-format-xsd-perl \
+	libdatetime-hires-perl \
 	libdbd-mysql-perl \
+	libdbd-mariadb-perl \
 	libdevel-checklib-perl \
 	libemail-address-xs-perl \
 	libemail-date-format-perl \
@@ -98,12 +103,14 @@ RUN apt-get update \
 	libfile-copy-recursive-perl \
 	libfile-find-rule-perl-perl \
 	libfile-sharedir-install-perl \
+	libfile-slurper-perl \
 	libfuture-asyncawait-perl \
 	libgd-barcode-perl \
 	libgd-perl \
 	libhtml-scrubber-perl \
 	libhtml-template-perl \
 	libhttp-async-perl \
+	libio-compress-perl \
 	libiterator-perl \
 	libiterator-util-perl \
 	liblocale-maketext-lexicon-perl \
@@ -117,6 +124,11 @@ RUN apt-get update \
 	libmodule-pluggable-perl \
 	libmojolicious-perl \
 	libmojolicious-plugin-renderfile-perl \
+	libmoose-perl \
+	libmoosex-types-common-perl \
+	libmoosex-types-datetime-perl \
+	libmoosex-types-perl \
+	libmoosex-types-uri-perl \
 	libnet-https-nb-perl \
 	libnet-ip-perl \
 	libnet-ldap-perl \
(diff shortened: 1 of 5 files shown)


Alex-Jordan (approved):

drgrice1:
To facilitate testing this and since our github wiki docker instructions are so convoluted and out of date, here are instructions for doing so.  This will give you a minimal setup for basic testing, and is not for production usage.  You can use either docker or podman for this.  These instructions are for the 2 stage build since that is how the `docker-compose.dist.yml` file is setup by default.

- Either install docker or podman.  On Ubuntu execute `sudo apt install docker.io docker-compose-v2 docker-buildx` to install docker, and `sudo apt install podman podman-compose` to install podman.
- Get a clone of this branch with `git clone -b docker-build-update-ubuntu-26 https://github.com/drgrice1/webwork2.git`.
- `cd webwork2`
- `cp docker-config/docker-compose.dist.yml docker-compose.yml`
- `cp docker-config/env.dist .env`
- Edit the `docker-compose.yml` copy made above and set `- WEBWORK2_GIT_URL=https://github.com/drgrice1/webwork2.git` on line 79, `- WEBWORK2_BRANCH=docker-build-update-ubuntu-26` on line 88, and `- PG_BRANCH=PG-2.21` on line 89.
- Execute `docker build -t webwork-base:forWW221 -f DockerfileStage1 .` or `podman build -t webwork-base:forWW221 -f DockerfileStage1 .`.  (Make sure to include the dot at the end of both of those commands.) This will take about 10 minutes.
- Execute `docker compose build` or `podman-compose build`. This will take less than a minute.
- Now kick it all off with `docker compose up` or `podman-compose up`.  This will take a bit and probably give a lot of messages stating that the web app is waiting on the database (many more of those with docker than podman), but after a little while it should say something like `Web application available at http://127.0.0.1:8080`.
- Enter the URL `http://localhost:8080` in your browser on the same machine.  You can log into the admin course with the user `admin` and password `admin`.  You will need setup two factor authentication to login, but then you can create courses and such.
- Shut it down with `docker compose down` or `podman-compose down`.  (Change `down` to `stop` if you don't want to delete the container).  Podman seems to have trouble stopping the database.  So you need to run `podman container ls -a` (and not the container_id of the database container which should be the only container shown at this point), then `podman inspect container_id --format '{{.State.Pid}}'` and not the PID that gives, then `kill -9 PID`.

Note that general docker testing with a specific branch can always be done with the above by changing `docker-build-update-ubuntu-26` to whatever branch you want to test.  Also change the PG branch appropriately for testing PG branches with this.

somiaj:
Thanks for the instructions, that helped out a lot.

Overall the build worked, I was able to create a course. Anything using the RPC failed (problem editor, library browser), but if I used the problem editor to view problems in their own page they worked just fine. I assume this is just the step to configure the RPC correctly wasn't listed and I didn't debug it further.

In following your instructions with `podman` for the sage 1 build, I get `WARN[0000] missing "ADDITIONAL_BASE_IMAGE_PACKAGES" build argument.`. Is this something that we should worry about (I will assume not).

The build works, here is my approval. Should we merge this (and the other pull request first) now that there is two approvals or see if @taniwallach wants to do more in depth testing?

somiaj (approved):

drgrice1:
I don't think docker cares if an environment variable is not set, but podman warns about it.  It is inconsequential though.  If you want additional Ubuntu packages installed in the base image, then you set that variable with the list of packages you want (as a space separated string).  Otherwise leave it unset.

The RPC endpoint should work.  I will look into that.

somiaj:
I have ngnix running as a proxy here for my development setup, maybe the problem was with that, though as long as the port 8080 is being used I don't think that will conflict with ngnix running on 80/443. I did stop all my development servers in the testing, so if the RPC works for you, I'll look into my setup more.

drgrice1:
It wasn't working for me either.  It is just a missing dependency though.  Perl critic needs to be added to the packages that are installed.  I will push the changes soon, after the build completes and I test it.

drgrice1:
I also just noticed that it is still installing the wrong version of `Perl::Tidy`.  That doesn't matter too much since that is only a development requirement, but I usually keep it in sync.  Otherwise there is no reason to install it from cpan in the docker build.

drgrice1:
Everything should be fixed now.

somiaj:
Tested again, RPC works fine. Should we merge or wait for more testing?

drgrice1:
I think merging is okay.

PR #3000, opened 2026-06-04
add-on conf files

Alex-Jordan:
This allows for there to be a folder `webwork2/conf/addon/` (or some other name, overridden in `localOverrides.conf`) where you can store `.conf` files. This has to do with #2999.

When adding a new course from the admin course, you can select one or more of these config files to be included at the end of the new course's `course.conf` file. If the new course is copying a `course.conf` file from some other course, nothing happens with `.conf` files from the `addon` folder.

If there is no `addon` folder, or if it has no `.conf` files in it, there is no change in the Add Course page. Otherwise, the checkbox for copying an old course's `course.conf` file is gone, and there is one select. The select has three optgroups:

1. The first has only one option, to just make the default `course.conf` file. This is also what will happen if the form is submitted with no selection made.
2. The second has only one option: copy the `course.conf` file from the indicated older course.
3. The third has all of the `.conf` files from the `addon` folder.

The select allows multiple selections. This needs one more thing where I need javascript help from @drgrice1. If the option from the first or second optgroup is selected, all other options need to be de-selected. Googling, I see how to do this with javascript. But I'm not sure where to build that javascript properly for its use here. That is, I'm not sure what I should do instead of writing inline script.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -277,6 +277,9 @@ $webworkDirs{bin}           = "$webworkDirs{root}/bin";
 # Location of configuration files.
 $webworkDirs{conf}          = "$webworkDirs{root}/conf";
 
+# Location of add-on configuration files.
+$webworkDirs{addOnConf}     = "$webworkDirs{conf}/addon";
+
 # Location of assets (tex, pg, themes)
 $webworkDirs{assets}        = "$webworkDirs{root}/assets";
 
--- a/htdocs/js/CourseAdmin/restrict_select.js
+++ b/htdocs/js/CourseAdmin/restrict_select.js
@@ -1,31 +1,33 @@
 (() => {
 	const addOnConfSelect = document.getElementById('add_on_conf');
-	const addOnConfOptgroups = [...addOnConfSelect.querySelectorAll('optgroup')];
+	if (!addOnConfSelect) return;
+
+	const addOnConfOptgroups = addOnConfSelect.querySelectorAll('optgroup');
 
 	// Track previously selected options to identify the newly clicked option
 	let previousSelection = [];
 
-	addOnConfSelect.addEventListener('change', (event) => {
-		const currentSelection = Array.from(addOnConfSelect.selectedOptions);
-
+	addOnConfSelect.addEventListener('change', () => {
 		// Find the option the user just clicked/selected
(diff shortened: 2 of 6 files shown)


Alex-Jordan:
Here is what this looks like when there is an `addon` folder with `.conf` files in it. In this case, `test.conf` and `test2.conf`'

Screenshot 2026-06-03 at 10 43 23 PM

I should have mentioned more about selecting multiple `.conf` files to append. You might do this if you have one for a certain LTI setup, and another for an instructor's style preferences. Just for example, and you want to load both but keep them separate.

There is no control at this level for what order they get loaded. They will land in `course.conf` in a fixed order from how they were submitted by the form. If it matters that the order be changed, someone would have to manually edit the `course.conf` file.

Alex-Jordan:
I added inline javascript to this template, just so it can be tested that the functionality works. I will need some guidance with how to move the javascript somewhere else.

drgrice1:
The place to put a javascript file for this would be in `htdocs/js/CourseAdmin`.  The place to load that file would be in the `templates/ContentGenerator/CourseAdmin/add_course_form.html.ep` file.  See the `templates/ContentGenerator/CourseAdmin/manage_otp_secrets_form.html.ep` and how it loads the `htdocs/js/CourseAdmin/manage_otp_secrets.js` file for a similar case.

Alex-Jordan:
OK, thanks. This is ready to examine now.

drgrice1 (commented):
I think this generally looks good.  I am going to add a pull request with some minor clean up and issue fixes.  The things in the pull request are explained there rather than duplicating the comments in this review.

One thing that I noticed and that I did not address in the pull request to this branch is that the JavaScript handling of the `add_on_conf` select does result in some odd behavior if `Ctrl` or `Shift` are used in combination with the array keys, or if the mouse is used to click and drag to select options. I suppose it is fine for this case though.

Alex-Jordan:
Thanks for the changes. And I hadn't thought to try keyboard use of that select. It seems that if you are trying to make a legal selection, keyboard use works though, so I'm not too concerned. I only see the weirdness if I try to make an illegal selection.

drgrice1 (approved):

somiaj (approved):

PR #3001, opened 2026-06-05
run perltidy on defaults.config

Alex-Jordan:
This is simply the result of running `./bin/dev_scripts/run-perltidy.pl conf/defaults.config`.

As I read through `defaults.config`, I see several inaccuracies. Starting with the opening paragraph that mentions the Apache config. Before I attempt to edit those, I wanted to normalize the code here.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -14,7 +14,7 @@
 # file, you can put a directive in localOverrides.conf.
 
 include("conf/site.conf");
-include("VERSION");  # get WW version
+include("VERSION");    # get WW version
 # The version of PG is now obtained from the file pg/VERSION
 # using code added to CourseEnvironment.pm
 # with this one exception include can only read files under the webwork2 directory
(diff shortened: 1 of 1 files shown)


conf/defaults.config
```diff
@@ -1029,16 +1022,16 @@ $caliper{enabled} = 0;
 # in the homework sets editor.
 $pg{displayModes} = [
 	"MathJax",    # render TeX math expressions on the client side using
-	              # MathJax; we strongly recommend people install and use
-	              # MathJax, and it is required if you want to use MathView
+				  # MathJax; we strongly recommend people install and use
+				  # MathJax, and it is required if you want to use MathView
 	"images",     # display math expressions as images generated by dvipng
-#	"plainText",  # display raw TeX for math expressions
+	#	"plainText",  # display raw TeX for math expressions
```
drgrice1: I noticed that `perltidy` still wants to change this line.  To fix this add an empty line before it so that `perltidy` doesn't try to align this comment with the comments above.  It isn't the best, but at least then if `perltidy` is run on this file it won't change it further.

Also, please remove the tab character in between the `#` and the commented out content.  That is how this line appears in `localOverrides.conf` as well.

drgrice1 (approved):
This is something I have wanted to do for many of the configuration files for a while.   I have auto format on save set in vim, and I always have to use `:noautocmd w` to save them or they get perltidied by the auto save.

conf/defaults.config
Alex-Jordan: OK, I changed that line.

somiaj (approved):

PR #3007, opened 2026-06-08
Some odds and ends that I have had sitting here for a couple of years now.

drgrice1:
This is house cleaning for me.  I want to delete this branch.

Remove the bold tag from the unused sub-display error for the admin course.  The errors in the error list that message is added to are html escaped, and so it can not contain html.  The errors could be prevented from being escaped, but I see no reason to do so for that one message. None of the much more important errors added to that list use html.  This particular message will almost never be seen unless a user is manually editing the URL with an incorrect subDisplay parameter.

Remove an unused variable in `templates/ContentGenerator/ProblemSet/version_list.html.ep`, and fix a minor issue with the remaining time shown for tests when viewing the test information page and a timed test is in progress.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
+++ b/lib/WeBWorK/ContentGenerator/CourseAdmin.pm
@@ -226,7 +226,7 @@ sub pre_header_initialize ($c) {
 				$method_to_call = 'do_registration';
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
Didn't test this, but looks fine.

somiaj (approved):

PR #3008, opened 2026-06-08
Fix contrast when viewing the xml of a hardcopy theme in the PG editor in dark mode.

drgrice1:
The `bg-white` class is set for the containing panel for the view area in the PG editor.  So for the text to be dark so that it works even in dark mode.

This is the quick and easy fix.  Better probably would be to use a color scheme responsive class such as the `bg-light` class (as is done for course info files). Then in dark mode you would get a dark background with light colored text.  However, that would change the background color in light mode.  Currently the background is actual white in light mode, and the `bg-light` class is an off white (same as the background of the site navigation).


somiaj (approved):

dlglin (approved):
Works for me

PR #3010, opened 2026-06-08
Fix a major bug with the navigation that everyone has certainly seen.

drgrice1:
Well, @somiaj has at least.  If you have your window size set to exactly the breakpoint (which is 768 pixels), then the site navigation will not open.  The `=` needs to be on the other side of the check in `htdocs/js/System/system.js`.

### DIFF
--- a/htdocs/js/System/system.js
+++ b/htdocs/js/System/system.js
@@ -11,7 +11,7 @@
 			navigation_element.classList.remove('invisible');
 			content.classList.toggle('toggle-width');
 
-			if (currentWidth < threshold) {
+			if (currentWidth <= threshold) {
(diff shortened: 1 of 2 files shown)


somiaj (approved):
Fixes the issue for me.

It was just random happen stance I was working with the perfectly sized window for this to occur.

somiaj:
@drgrice1 seems the title is just your branch name, you want to keep it that way?

drgrice1:
Thanks.  I fixed the title to what it was supposed to be!

Also I added one small tweak to the sidebar when it is hidden that seems to give a cleaner appearance in some cases.

pstaabp (approved):

PR #3013, opened 2026-06-09
Add a case for incompatible Perl package versions in `check_modules.pl`.

drgrice1:
Any package can have an `incompatibleVersions` flag whose value should be a reference to an array of version strings.  If the version on the system of a package is one of these versions, then `check_modules.pl` will warn about it.

`Mojolicious` now has this set for versions 9.43, 9.44, 9.45, and 9.46.

Also remove unneeded quotes on left operands of a fat comma and hash keys.  I missed these when they were added in #2893.  I left the quotes on the single word package names which also don't need them, but this is consistent with the other package names that do.

Note that the primary changes (the ones not just removing unneeded quotes) are lines 682 - 690, and the addition of the key for the `Mojolicious` package on line 354.

### DIFF
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -60,493 +60,494 @@ use Pod::Usage;
 
 my %modulesList = (
 	'Archive::Tar' => {
-		'package' => {
-			'ubuntu' => 'perl',
-			'rhel'   => 'perl-Archive-Tar'
+		package => {
+			ubuntu => 'perl',
+			rhel   => 'perl-Archive-Tar'
 		}
 	},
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
I went to Mojolicious 9.45 and this caught it. Back to 9.42 and I passed the check.

somiaj (approved):
I am not able to test if it catches bad versions of mojolicious, but the script works just fine for me with a valid version.

somiaj:
Going to accept @Alex-Jordan that it correctly catches the bad versions.

PR #3014, opened 2026-06-10
Perltidy update

drgrice1:
Update to `perltidy` version 20260204.

Developers will need to install and use this version to match the workflow.

Also update the Github workflow to use node 24, and the actions in the workflow to their newest versions.

The first commit in this pull request sets everything up.  The second commit perltidies the Perl files in the code base.  Note that some of the files in `bin` that do not have the `.pl` extension were also perltidied.  That will not be enforced by the workflow, but this still should be done.


somiaj (approved):
We should probably get this in so other PRs can be tidied with the new version.

Alex-Jordan (approved):

Issue #3012, opened 2026-06-09
perl tidy errors persist in Problem Editor


pstaabp:
In the problem editor, use a problem and include a syntax error (like leave off a ;)

If you go to perltidy it from the Code Maintenance tab, on the right, there will be the output from the perltidy script showing the error location.

If you fix the error and rerun perltidy, the error will persist on the right side.  It should either return to the problem output or some indication that the perltidy was successful.

drgrice1:
I guess one side effect of this approach is that if you first use PG critic on a problem, and then run perltidy on the problem, the problem will also be rendered, replacing the PG critic results.  I am okay with that though.

somiaj:
Fixed in #3015.

PR #3015, opened 2026-06-10
Re-render a problem in the PG problem editor after a perltidy failure.

drgrice1:
When editing a problem and the code has a syntax error that prevents perltidy from working, if perltidy is then used it shows the errors from running perltidy.  If you then fix those errors and run perltidy again, the toast shows that pertidy was successful, but the errors are still shown in the render window.

This pull request makes it so that when perltidy is successfully executed after a previous executing failed, then the problem is re-rendered, thus removing the perltidy errors.  This is detected simply by checking if the render area contains an iframe or not.

This addresses issue #3012.


Alex-Jordan (approved):

somiaj (approved):
Fixes the issue.

PR #3018, opened 2026-06-14
Set default in achievement evaluator editor save as form.

### LINKED ISSUE #2795: Saving a new achivement evaluator doesn't select a default action.
When editing an achievement evaluator, then clicking on the 'Save As' tab, none of the three options are selected by default. This leads to the following error if the user doesn't select any of the options.

"Don't recognize saveMode: |no_save_mode_selected|. Unknown error."

Despite this error, the file is still saved, but the achievement evaluator isn't changed, and any edits appear to be lost as the editor reverts back to the original evaluator.

I think the best fix is just to ensure that one of the three radio buttons are selected by default, but putting this as an issue to see what others thing before trying to fix it myself.

somiaj:
Set the first option to use the achievement evaluator in the current achievement as the default selected option in the save as form of the achievement evaluator editor if it was not already set from a previous post. This fixes #2795.

In testing this, this only fixes the issue I ran across where nothing was set as default. There are still some oddities and that could be improved. If you choose to `Save as` a new achievement, the new file is created, but the editor reverts back to the evaluator of the current achievement. Overall this form needs improved and is not addressed here, as any issues are just current behavior.

### DIFF
--- a/templates/ContentGenerator/Instructor/AchievementEditor/save_as_form.html.ep
+++ b/templates/ContentGenerator/Instructor/AchievementEditor/save_as_form.html.ep
@@ -1,5 +1,6 @@
 % # There are three things you can do with a new achievement editor.
 % # You can replace the current achievement, use it in a new achievement, or not use it at all.
+% param('action.save_as.saveMode', 'use_in_current') unless param('action.save_as.saveMode');
 <div>
 	<div class="row align-items-center mb-2">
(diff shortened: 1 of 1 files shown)


drgrice1 (commented):
This looks fine.

I think that the "save as" tabs for this and for the PG problem editor need to be generally reworked.  The behavior should be similar to the "save as" functionality for most file editors.  That is if you choose a filename of an existing file, then you are asked if you want to replace that file or not.

drgrice1 (approved):

Alex-Jordan (approved):

Issue #2785, opened 2025-08-13
rounding LTI passback grades


ophelia-adams:
It appears that WeBWorK rounds the scores sent via LTI to an LMS to the nearest integer.

This happens in line 175 of lib/WeBWorK/Authen/LTIAdvanced/SubmitGrade.pm:
`$score = wwRound(2, $score);`

While this appears consistent with how rounding is done elsewhere (cf https://github.com/openwebwork/webwork2/pull/513) those instances of rounding are for display, rather than calculation. It seems to me that, for numbers which will be used in grade calculations, exactness is preferable, so that rounding occurs just one time when calculating and displaying grades to students through the LMS.

I believe LTI 1.3 allows sending exact scores (separate score & total score) but I am not familiar enough with the compatibility considerations involved in WW's choice to only report percentages. In any case, it seems like sending as many significant figures as the LTI standards allow would suffice. (I acknowledge that even the current rounding scheme almost surely has no effect on student grades, but I can see how it might set off some worry)

ophelia-adams:
I see -- I believe my institution is using 2.15, which probably precedes LTI 1.3. The wiki documentation seemed to contradict that info about LTI 1.3 (but that I can update). Thanks! Might still be advisable to remove rounding from the LTI 1.1 version.

kyleclaassen:
I'd like to request that this issue be reopened. I'm still using LTI 1.1 for grade passback, and rounding the score ratio to two decimal places generates obvious discrepancies in the LMS (Moodle for me). For example, suppose a set has 6 problems and the assignment in the LMS is set to be worth 6 points. If a student scores 5 out of 6 in WeBWorK, then "0.83" is passed back to the LMS, which shows as 4.98 out of 6 instead of 5 out of 6. Naturally, this generates student complaints.

I'm not sure why the rounding was implemented, but it doesn't seem to be required by the LTI 1.1 specificiation; see [Section 6.1.1: replaceResult in the LTI 1.1 implementation guide](https://www.imsglobal.org/specs/ltiv1p1/implementation-guide#toc-26). The example XML message uses a value of "0.92" (indeed two decimal places), but rounding isn't mentioned anywhere. I've been commenting out the `wwRound` line, and  the grade passback to Moodle still works.

I'd like to propose adding a configuration variable `$LTI{v1p1}{enable_score_rounding}` in `conf/authen_LTI_1_1.conf` with a default value of `1` for backward compatibility:
```perl
[...]

PR #3019, opened 2026-06-15
LTI 1.1 configure number of digits to round score.

somiaj:
This adds a configuration option `$LTI{v1p1}{round_score_digits}` which sets the number of digits to round the score to before sending it to the LMS. If this is set to a number less than 2 (nearest whole percent), then rounding of the score is disabled. The default is `2` which is current behavior.

This was requested in #2785.

### DIFF
--- a/conf/authen_LTI.conf.dist
+++ b/conf/authen_LTI.conf.dist
@@ -240,6 +240,7 @@ $LTIMassUpdateInterval = 86400;
 	#'LTISendGradesEarlyThreshold',
 	#'LTIMassUpdateInterval',
 	#'LMSManageUserData',
+	#'LTI{v1p1}{round_score_digits}',
 	#'LTI{v1p1}{BasicConsumerSecret}',
 	#'LTI{v1p3}{PlatformID}',
 	#'LTI{v1p3}{ClientID}',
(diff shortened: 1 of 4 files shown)


drgrice1 (approved):
Looks good to me.

drgrice1:
Actually, it might be a good idea to add this as a variable that can be configured from the LTI tab in the course configuration.  So add this variable to those in the `@LTIConfigVariables` array, and add the appropriate entry in the `$LTIConfigValues` hash in `lib/WeBWorK/ConfigValues.pm`.

somiaj:
Added this to the list of possibly configurable items on the LTI tab of the course configuration page.

pstaabp (approved):
Didn't test, but looks good.

kyleclaassen:
Awesome, thank you so much for implementing this! I just noticed a couple typos in ConfigValues.pm:
* "Note that there maybe" should be "Note that there **may be**"  (two words).
* "recives" should be "receives".

Minor suggestion: could ensure that `$digits` is an integer in SubmitGrade.pm, e.g. `my $digits = int(...)`. (It'd be dumb, but one could set `$LTI{v1p1}{round_score_digits} = 1.2345`, which would end up setting `$score = wwRound(1.2345, $score)`. Looking at the code for `wwRound`, the returned value isn't meaningful if the number of places isn't an integer.)

somiaj:
Thanks for the typo catches.

In general webwork configurations require the person to put in the correct value type. So it should be up to the sysadmin to know this has to be an integer value, and not something I think the code should try to enforce.

PR #3025, opened 2026-06-24
Disabled tabs for the PG editor.

drgrice1:
The "Save" tab will always shown for all file types, but will be disabled for any file not in the course's templates directory (symlinks do not count here, i.e., this tab will not be disabled for a file that is linked to from the course's templates directory assuming the server has write permission to the file), or the server does not have write permission to the file.

The "Append" tab will be shown for all PG file types (problems and set headers).  However, it will be disabled for any file not in the course's templates directory (again symlinks do not count here, i.e., this tab will not be disabled for a file that is linked to from the course's templates directory, even if the server does not have write permissions to the file).

Also the help has been updated. help for this. The help for each action tab now states when the tab is available and when it is active.

### DIFF
--- a/templates/HelpFiles/InstructorPGProblemEditor.html.ep
+++ b/templates/HelpFiles/InstructorPGProblemEditor.html.ep
@@ -127,67 +127,99 @@
 </dl>
 
 <h2><%= maketext('Actions:') %></h2>
-<i><%= maketext('(If an action cannot be executed it will not appear.)') %></i>
 <dl>
 	<dt><%= maketext('View/Reload') %></dt>
 	<dd>
-		<%= maketext('Render or re-render the problem, course info file, or set header being edited. This does not '
-			. 'change the permanent file on the disk. You can view different versions of the same problem by changing '
-			. 'the seed. You can also change the manner in which the mathematics is typeset. If "Open in new window" '
-			. 'is checked, then the problem will open in a new tab or window showing the problem as it will be '
-			. 'rendered for students in the set.') =%>
+		<p>
+			<%= maketext('Render or re-render the problem, course info file, or set header being edited. This does not '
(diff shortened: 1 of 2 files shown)


templates/HelpFiles/InstructorPGProblemEditor.html.ep
```diff
@@ -196,25 +228,42 @@
...
 		</p>
-		<p class="mb-0">
+		<p>
 			<%= maketext('If the original problem cannot be edited than the path name must be changed in order to be '
 				. 'allowed to save the problem. Adding "local/" to the beginning of the original path is the default '
 				. 'solution. All locally created and edited files will then appear in a subdirectory named '
 				. '"local".') =%>
 		</p>
+		<p>
+			<%= maketext('This tab is available for all file types except for the course information file.') =%>
```
Alex-Jordan, line 238: Maybe I misunderstand what is meant here. But if I go to Problem Editor and type the file as course_info.txt, it opens in the editor, tries to render it as PG, but pertinent to this claim in the help, the Save As tab is available.

Alex-Jordan: I see. If you open the file for editing from the actual set info panel, then it knows this is the course information file, and has that behavior.

Alex-Jordan (approved):

templates/HelpFiles/InstructorPGProblemEditor.html.ep
drgrice1: Yes, in general, the problem editor does not really know what type of file is being edited unless it is told.  In all cases that determination is made at the place that you open a file.  So when you open a file from the "File" input on the problem editor "chooser" page, there is an assumption that you are editing a problem.

pstaabp (approved):

PR #3026, opened 2026-06-24
configure mathjax to line break long display math

Alex-Jordan:


somiaj:
What about adding `displayOverflow: 'scroll'` to the `output` key as well? This makes it so formulas will scroll if they are longer than their container and there are no line breaks.

htdocs/js/MathJaxConfig/mathjax-config.js
```diff
@@ -132,6 +132,15 @@ if (!window.MathJax) {
...
 			},
 			ignoreHtmlClass: 'tex2jax_ignore'
+		},
+		output: {
+			displayOverflow: 'linebreak',
+			linebreaks: {
+				inline: false,
+				width: '100%',
+				lineleading: .2,
+				LinebreakVisitor: null,
```
drgrice1: The `prettier` issues are that this should be

```suggestion
				lineleading: 0.2,
				LinebreakVisitor: null
```

but you can run `npm run prettier-format` to take care of this.

drgrice1:
The default setting for inline math is for line breaking to be enabled.  So this disables line breaking for inline math.  Do we really want that?

drgrice1:
Testing this with

```
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl');

BEGIN_PGML
[`
    \int \frac{4x^4 + 5x^3 + 110x^2 + 128x + 175}{x^3 + 25x} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x+3}{x^2 + 25} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x}{x^2 + 25} + \frac{3}{x^2 + 25} \,dx
    = 2x^2 + 5x + 7\ln\left(\left|x\right|\right) + \frac{3}{2}\ln\left(x^2 + 25\right) + \frac{3}{5}\tan^{-1}\left(\frac{x}{5}\right) + C.
`]

[```
    \int \frac{4x^4 + 5x^3 + 110x^2 + 128x + 175}{x^3 + 25x} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x+3}{x^2 + 25} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x}{x^2 + 25} + \frac{3}{x^2 + 25} \,dx
    = 2x^2 + 5x + 7\ln\left(\left|x\right|\right) + \frac{3}{2}\ln\left(x^2 + 25\right) + \frac{3}{5}\tan^{-1}\left(\frac{x}{5}\right) + C.
```]
END_PGML

BEGIN_PGML_SOLUTION
[`
    \int \frac{4x^4 + 5x^3 + 110x^2 + 128x + 175}{x^3 + 25x} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x+3}{x^2 + 25} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x}{x^2 + 25} + \frac{3}{x^2 + 25} \,dx
    = 2x^2 + 5x + 7\ln\left(\left|x\right|\right) + \frac{3}{2}\ln\left(x^2 + 25\right) + \frac{3}{5}\tan^{-1}\left(\frac{x}{5}\right) + C.
`]

[```
    \int \frac{4x^4 + 5x^3 + 110x^2 + 128x + 175}{x^3 + 25x} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x+3}{x^2 + 25} \,dx
    = \int 4x + 5 + \frac{7}{x} + \frac{3x}{x^2 + 25} + \frac{3}{x^2 + 25} \,dx
    = 2x^2 + 5x + 7\ln\left(\left|x\right|\right) + \frac{3}{2}\ln\left(x^2 + 25\right) + \frac{3}{5}\tan^{-1}\left(\frac{x}{5}\right) + C.
```]
END_PGML_SOLUTION

ENDDOCUMENT();
```

I see that the first line breaks with the current MathJax configuration, but not with this pull request.  I don't think that is good.

Of course, the display mode math on the second line does not break with the current MathJax configuration, but does with this, and still does if `inline: true` is set.

So I think that it would be best to leave `inline: true`.

Alex-Jordan:
I didn't think too hard about `inline: false`. I have seen that when you do the same think in actual LaTeX (using a package that allows inline math to break) it can be bad. Because when reading a sentence with inline math and there is a line break, it really disrupts the readability of the sentence. But for WW exercises, that's less of a concern that expository textbook.

drgrice1:
line breaking of any sort is new with MathJax, so whatever is decided is fine.  I wouldn't actually put something like that example inline, and have always used things like the `aligned` environment to avoid long lines in any case

Alex-Jordan:
> What about adding displayOverflow: 'scroll' to the output key as well?

Yes, there is scroll, line break, and also the option to scale it down. I thought the line break for display math seemed the most pleasing to look at. We can set it to something else if it's what we want. Any of it can be changed by the user by accessing the MathJax menu.

Do you have a preference for one of those as our default? Do you, @drgrice1?

drgrice1:
I think that using line breaking is probably best, with scrolling a close second. I think scaling it down by default is probably not good.

drgrice1 (approved):
This looks good to me as is.  We can still debate on if inline math should be line broken or not, and if it is better to line break or scroll (or scale down) display math though.  I will go ahead an approve it conceptually.

pstaabp (approved):
Using @drgrice1 example problem above, adding line breaks is definitely better than without them. 

I think keeping the block/display mode unaffected is good.  As with others, I use aligned to format (add line breaks).

PR #3030, opened 2026-06-25
icon alert if a date has been overrided for a user on the Assignments…

Alex-Jordan:
On the Assignments page, if the date that is mentioned in the status message is actually an override date for that user, indicate this with an icon using the same structure as the set description.

This is motivated my instructors reaching out to me, because they do not understand why dates are not changing on a student's Assignment screen even though the instructor is changing the global set dates.

Here is a screenshot for a quick look.

Screenshot 2026-06-24 at 10 49 23 PM

Note that if there are additional dates mentioned in subsequent messages, such as `'Afterward reduced credit can be earned until [_1].'`, there is no icon like this. It is only for the current date of primary importance.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -752,6 +752,7 @@ $authen{admin_module} = ['WeBWorK::Authen::Basic_TheLastOption'];
 	access_instructor_tools => "ta",
 	score_sets              => "professor",
 	problem_grader          => "professor",
+	override_indicator      => "ta",
 	send_mail               => "professor",
 	receive_feedback        => [ 'ta', 'professor', 'admin' ],
 
--- a/lib/WeBWorK/ConfigValues.pm
+++ b/lib/WeBWorK/ConfigValues.pm
@@ -606,6 +606,17 @@ sub getConfigValues ($ce) {
 				type => 'permission'
 			},
 			{
+				var  => 'permissionLevels{override_indicator}',
+				doc  => x('Indicate each date overriding an assignment date'),
+				doc2 => x(
+					"On the Assignments page, if a primary date is shown for an assignment, and if the user's value of "
+						. 'that date is overriding the course assignment data, then show an indicator that this is the '
+						. 'case. This will happen for all users at permission level at or above the selected '
+						. 'permission level.'
+				),
+				type => 'permission'
+			},
+			{
 				var  => 'permissionLevels{view_unopened_sets}',
 				doc  => x('Allowed to view problems in sets which are not open yet'),
 				type => 'permission'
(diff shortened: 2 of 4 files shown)


drgrice1:
It seems odd to me for a student looking at their assignments  to have the message "This date has been overridden for this user from the course set date." next to an assignment date.  The odd thing is that the "for this user" in the message is the student viewing the set.  I think that this message should either not be shown at all when a student views their own set, or perhaps the message should be different in this case, perhaps just omit the "for this user" wording.

Alex-Jordan:
I changed the message to "This date overrides the assignment date."

drgrice1 (approved):
That is better. Looks good now.

somiaj:
So students see that their date was overridden? Can this be turned off? I guess to me I don't see the use case a student needs to know that their date is different than the default.

somiaj:
Would only showing this message when acting as another user be sufficient for your use case @Alex-Jordan?

drgrice1:
I sort of thought that too, but I don't think it hurts anything.   It does let the student know that an extension was granted.  So if a student asks the instructor for an extension and the instructor says that it is okay, then this gives them further verification that it was done.

somiaj:
In that case should this message also be added to the ProblemSet page as well, most students may only see that, and the due date is there too?

somiaj:
I guess it could also let students know they used an extension achievement item too. So maybe there is some use for students being able to see this. I still feel it is mostly extra info they don't need, but I may have to see it in use to see if students send me questions about it when they see it.

Alex-Jordan:
Yes, I was going to say that a student can see confirmation their date was changed after they used an achievement. But also if they had just asked for an extension. And generally, students will rarely have these date overrides. But I would argue that it is good to indicate to the student if/when their dates may differ from a classmate's.

Alex-Jordan:
> In that case should this message also be added to the ProblemSet page as well, most students may only see that, and the due date is there too?

Could be done with future work. But this alone is an improvement, I think.

somiaj:
One use case I use is I have multiple sections all in the same webwork course. For tests, they are all scheduled differently depending on section's class time. So all sections except the 'default section' will see this override, which may not have useful meaning. This is more than just a students will rarely have overrides use case. I'm okay with adding this as most my students won't see this page, but not all overrides from a template may have meaning as an extension, it could be section based due dates too.

Alex-Jordan:
A config setting for this would be trivial. Just accessed in one place in
the template file. And added to Course Config page. I'd prefer it be
enabled by default, and people doing what you describe could disable it. I
used to do that too, until I just made all sections have the same dates at
the end of a weekend. I feel like it's a hack to work around our poor
support for sectioning, but I get it and did do that for a few years.

I can't actually do any coding for a few weeks. Feel free to PR my branch
and I'll merge by phone.

Alex Jordan
Mathematics Instructor
Portland Community College

pstaabp:
Overall, I like this. 

I'd recommend it is on by default with an optional configuration.

Alex-Jordan:
This now has a permission level setting, and the indicator icon only appears if you are a user with that permission level or above. Including if you are acting as a user of a lesser permission.

pstaabp (approved):

PR #3031, opened 2026-06-25

Alex-Jordan:
These are typos identified by `github.com/crate-ci/typos`, except not typos in `lib`. Except a few that are in `lib` that I either already had on my typos branch, or that are connected to a typo in `templates` and can only be fixed at the same time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -258,7 +258,7 @@ COPY docker-config/docker-entrypoint.sh /usr/local/bin/
 
 ENTRYPOINT ["docker-entrypoint.sh"]
 
-# Add enviroment variables to control some things during container startup
+# Add environment variables to control some things during container startup
 ENV SSL=0 \
 	PAPERSIZE=letter \
 	SYSTEM_TIMEZONE=UTC \
--- a/DockerfileStage2
+++ b/DockerfileStage2
@@ -101,7 +101,7 @@ COPY docker-config/docker-entrypoint.sh /usr/local/bin/
 
 ENTRYPOINT ["docker-entrypoint.sh"]
 
-# Add enviroment variables to control some things during container startup
+# Add environment variables to control some things during container startup
 ENV SSL=0 \
 	PAPERSIZE=letter \
 	SYSTEM_TIMEZONE=UTC \
--- a/assets/pg/PGMLLab/PGML-lab.pg
+++ b/assets/pg/PGMLLab/PGML-lab.pg
@@ -286,7 +286,7 @@ TEXT(tag(
 						'External ANS' => [
 							<<~'ENDPG',
 								Context('Vector');
-								ANS(Vector(1, 2, 3)->cmp(showCoodinateHints => 0));
+								ANS(Vector(1, 2, 3)->cmp(showCoordinateHints => 0));
 								ENDPG
 							<<~'ENDPGML'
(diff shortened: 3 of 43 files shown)


Alex-Jordan:
`github.com/crate-ci/typos` identifies 1185 typos in `lib`. I'm sure some of them are false positives, but I would not be surprised if around 1000 of them are legit. I'll save that for some other time. Or anyone else, feel free to try this tool. I'll be offline for a few weeks now.

doc/Artistic
```diff
@@ -126,6 +126,6 @@ products derived from this software without specific prior written permission.
 
 10. THIS PACKAGE IS PROVIDED "AS IS" AND WITHOUT ANY EXPRESS OR
 IMPLIED WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED
-WARRANTIES OF MERCHANTIBILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
```
somiaj: This is a typo in the original license, https://perlfoundation.org/artistic-license-10.html, and should not be fixed.

somiaj:
I looked over the typos. Most are just comments/output text and look fine. A few did change some typos in variable names. In this case they appear to have found all cases, because it looks like the variable names were consistently spelled wrong in the code. Though I didn't do any testing and some should be done to ensure that code changes didn't break anything (initial look is most likely not).

doc/Artistic
Alex-Jordan: Should we upgrade to Artistic License 2.0? (https://opensource.org/licenses) Among other changes, it did correct that misspelling. Either way, I will remove this change from both this PR and the `pg` one once I get a chance to.

somiaj: Upgrading the license is not an easy thing. Since the Artistic License 1.0 doesn't give the option to use any future versions of the license, this would require consent from all authors (or some time window to give authors time to disagree). We have discussed issues around upgrading licenses, but unlike the GPL 2.0+ which gives us the option to use any later versions of the license, the Artistic License doesn't and there would need to be some great benefit to go through the process to upgrade.

Alex-Jordan: Hmm. The WeBWorK Project holds copyright, and can release version 2.21 using whatever licenses it wants to. If the concern is that past contributors did not give consent to some other license, the thing is that they gave consent to TWP to hold copyright over that content. I guess we developers should not decide, but I do hold that TWP can decide to make a change without much fuss. It would not invalidate old licenses applied to old versions. But it could be a reset starting with a certain version of webwork2/pg.

pstaabp:
Looking over this, it seems that there are only two files with code ramifications, `PGProblemEditor.pm` and the template `save_as_form.html` and since the typos were consistent in the previous versions and fixed in both places, this should be good. 

Should we also then revert the typo in `doc/Artistic` ?

doc/Artistic
Alex-Jordan: In any case, I reverted this change back to "MERCHANTIBILITY".

Alex-Jordan:
@pstaabp I just now reverted that license typo correction.

pstaabp (approved):

drgrice1 (approved):

Issue #2982, opened 2026-05-21
LTI 1.3 issue after a Moodle update - needed to temporarily hack Moodle code to force SameSite=None


taniwallach:
A Moodle site accessing a webwork server I manage ran into a total failure of the LTI 1.3 authentication process after the Moodle server pulled in a [recent commit](https://github.com/moodle/moodle/commit/f6090cf28524f0e27d021d9af62a2ad2428b6dc8) added to the `MOODLE_405_STABLE` branch (the site is on 4.5.11+ so got that commit on a periodic pull of updates).

The underlying problem was that webwork makes the call back to the "authentication request URL" as a POST, while Moodle's session cookie (after the change) has `SameSite=Lax` so was not sent. Setting `SameSite=Lax` is now done for almost all clients.

The Moodle commit dropped the code which forced the Moodle session cookie to be set with `SameSite=None` for all modern Chrome browsers, and now **forces** `SameSite=Lax` (except for a special case).

I think that this change to Moodle also overrides any setting of  `session.cookie_samesite = "None"` made in `php.ini` on the Moodle server. See [the PHP documentation on session-set-cookie-params](https://www.php.net/manual/en/function.session-set-cookie-params.php). 

**If that is correct, it is likely that any Moodle server which gets this recent Moodle code change will apparently run into problems with LTI 1.3 connections to webwork.** 

The issue is under discussion  [on the Moodle LTI forum](https://moodle.org/mod/forum/discuss.php?d=474279).

What is happening:
1. Moodle is now explicitly setting `SameSite=Lax` (except for a special case) on the session cookie.
2. After the webwork server handles the "login" phase, it send the browser the self posting form to trigger the authetication phase.
3. The browser send a `POST` to Moodle's `mod/lti/auth.php` as its handling of the self posting form. Since this is a `POST` and the referred is webwork, the `SameSite=Lax` cookie policy **forbids** the browser from sending the Moodle session cookie.
    * Part of the data is `repost` set to `1` (set explicitly by webwork).
4. Moodle receives the request **without** the session cookie.
5. Without the cookie, Moodle cannot complete the handshake (it stores critical data for the process in the session data) and reports an Invalid request. When debugging settings are enables, the traceback shows a failure to retrieve the data needed to continue from the session.

[...]

PR #3032, opened 2026-06-26
LTI self-posting form fix

taniwallach:
Drop the repost parameter form the LTI self-posting form used after the login stage to trigger the authentication stage on the LMS.

The parameter was is not part of the LTI standard, and is used by Moodle to prevent the authentication stage from entering a self-reposting loop, as a self-reposting mechanism is used to pull in the Moodle session cookie which is no longer provided by browsers when SameSite=Lax is set on the LMS side and the sites are considered cross-site.

Having this unneeded parameter set breaks the Moodle to WeBWorK login/launch and content selection mechanism after Moodle changed to defaulting to SameSite=Lax.

Fix for https://github.com/openwebwork/webwork2/issues/2982

### DIFF
--- a/lib/WeBWorK/ContentGenerator/LTIAdvantage.pm
+++ b/lib/WeBWorK/ContentGenerator/LTIAdvantage.pm
@@ -134,7 +134,6 @@ sub login ($c) {
 		'ContentGenerator/LTI/self_posting_form',
 		form_target => $c->ce->{LTI}{v1p3}{AuthReqURL},
 		form_params => {
-			repost           => 1,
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
As far as I can tell this parameter is only used by Moodle, and so this should not affect any other LMSs. So if it is causing problems for Moodle, is should be removed.

drgrice1:
Also, I have tested this with both Moodle and Canvas, and everything seems to be working.

Alex-Jordan (approved):

PR #3034, opened 2026-06-26
Remove the html2xml endpoint.

drgrice1:
This has been deprecated since the release of webwork 2.18. It is time for it to go.  I believe that everyone that used to use this endpoint have now updated to use the `render_rpc` endpoint. If not, then we can wait another release for this.

### DIFF
--- a/conf/webwork2.mojolicious.dist.yml
+++ b/conf/webwork2.mojolicious.dist.yml
@@ -49,14 +49,9 @@ JSON_ERROR_LOG: 0
 # origins without an explicit directive from the server that it is safe.
 # Uncomment the extra_headers setting as is to set the header to enable
 # Cross-Origin Resource Sharing (CORS) for url responses to requests to urls
-# that start with /webwork2/render_rpc, /webwork2/html2xml, and /webwork2_files.
-# This may be needed to use the client formats if they are accessed from a
-# *remote* server.
-#
-# Note that the /webwork2/html2xml url is deprecated, and code should be adapted
-# to use the /webwork2/render_rpc path instead.  They both do the same thing
-# that the /webwork2/html2xml endpoint did before, but they no longer pass
-# through the XMLRPC middle man.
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
There is one thing that PTX still uses this for. It's something about determining the WeBWorK version of a server it is contacting. I will have to refresh myself later. I suspect it is something that could be ironed out in time for PTX to support a 2.21 server, but might need to ensure there is a solid way for a webwork2 server to be broadcasting its version. I can look into it in a few weeks. But in the meantime, is there a simple way to just ask a webwork2 server to report its version?

drgrice1:
At this point there is no particularly clean way to make a webwork server report its version.  But there are ways to get it.

One way would be to try the html2xml endpoint, and if the request fails, then you know it is 2.21 or newer.

Another way would be to get the root page `https://your.server/webwork2` and parse the html.  The footer has the version.

drgrice1:
For testing to see if the `html2xml` endpoint exists, I guess you have to be a little careful.  The request won't exactly fail.  Instead it will render the text response `The course html2xml does not exist.` with a status of 404.  But that gives a rather definitive determination of version 2.21 or newer (assuming this pull request is merged).

drgrice1:
Using the `html2xml` endpoint to determine the webwork2 version was really not the correct thing to do.  What your code should do is first try the `render_rpc` endpoint, and if that works get the version it sends.  If that fails, then try the `html2xml` endpoint.  This wouldn't be hard to implement in the `pretext/lib/webwork.py` file where this version check is done.

Alex-Jordan (approved):

pstaabp (approved):

Alex-Jordan:
Noting here that I had to revert this when I went to 2.21 in production. There are two PreTeXt books that were built long ago, and still expect to find the html2xml endpoint. And it's not practical to consider rebuilding those books. But simply undoing the edits here worked for now. In the meantime I will look into search-and-replace in all the generated HTML to move those books to use render_rpc.

Issue #2568, opened 2024-09-09
symlinks in File Manager can be deleted


Alex-Jordan:
An instructor can unintentionally delete the symlinks in their templates/ folder, leading to an admin having to restore them. Is it possible to make it so that symlinks cannot be deleted using the File Manager page?

Alex-Jordan:
I think that with #3036, this can be closed. As mentioned in that thread, if there are symlinks that are not "protected", they could be added to the protected list as appropriate.

PR #3036, opened 2026-07-03
Protect the required course symbolic links from deletion in the file manager.

drgrice1:
This just adds the `courseLinks` in the course environment to the list of directories (and now links) that are checked when an instructor attempts to delete a directory in the file manager.

This is to address issue #2568.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/FileManager.pm
@@ -342,7 +342,8 @@ sub Delete ($c) {
 
 	my $dir = "$c->{courseRoot}/$c->{pwd}";
 
-	my @course_dirs = values %{ $c->ce->{courseDirs} };
+	my @course_dirs  = values %{ $c->ce->{courseDirs} };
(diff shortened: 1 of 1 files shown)


pstaabp:
Notice that this just pertains to symbolic links like `Library`.  I have a `Fitchburg` one that points to the `Fitchburg` directory in `Contrib` and it was happy to delete it for me.

pstaabp (approved):

drgrice1:
Yes, this only pertains to the required links. A local library link is not one of the required links.

drgrice1:
Note that if you add `$courseLinks{Fitchburg} = [ '/actual/path/to/Fitchburg', "$courseDirs{templates}/Fitchburg"];` to `localOverrides.conf`, then your `Fitchburg` link would also be protected by this.

drgrice1:
Note that if you check for upgrades in the admin course, it will tell you ` Link structure needs to be repaired.` for any course that does not have that added link, and if you upgrade the course it will create the link.

Alex-Jordan (approved):

Issue #2976, opened 2026-05-13
password reset


Alex-Jordan:
I've been wondering if we could implement a password reset flow. It feels like a medium-sized project, but also there should be good workflows to follow for making it happen.

drgrice1:
What exactly do you mean by a password reset flow?  Do  you mean a link on the login page that can be used to reset your password if you have forgotten it?

Alex-Jordan:
Yes, and by "flow" mean the mechanics of how that works for the user. An
option is that you could enter your username..but maybe you should actually
enter your email address. And then what happens? You could get an email
with a link that expires, and use it to reset the password. Or I can
imagine other things. I know what I've seen used, but I don't know what's
the most secure and/or the right thing for WeBWorK to do

Alex Jordan
Mathematics Instructor
Portland Community College

Alex-Jordan:
This is done now, thanks to #3041.

PR #3041, opened 2026-07-07
Implement a reset password flow.

drgrice1:
In order to enable the reset password flow, the course environment variable `$password_reset_sender_email` must be set. This can be set per course in the `course.conf` files.

If that is set, then a "Forgot Password" link will appear on the course login page to the right of the "Continue" button. When that link is used a "Forgot Password" page is shown. The user will then need to enter their username and click "Request Password Reset". Then an email is sent to the user (assuming the user has a valid email address in the database) containing a "Reset Password" link. A single use token is embeded in the link, and that token expires in 15 minutes. When that link is used a "Reset Password" page is shown. On that page the user must enter a new password, and confirm it by entering it again, and then click "Reset Password".  If the token in the link is valid, then the user's password is changed.

If two factor authentication is enabled, then when the reset link that is emailed to the user is used, the user must enter a one-time security code as well as a new password in order to reset their password. Also, if the user has not yet set up two factor authentication for their account, then password reset is not allowed for the user.
    
All requests to reset a user's password and the success or failure when using the reset password link are logged to the login log.

This is to address issue #2976.

Remember to upgrade the admin database and the course used when testing this.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/ForgotPassword.pm
+++ b/lib/WeBWorK/ContentGenerator/ForgotPassword.pm
@@ -1,0 +1,99 @@
+package WeBWorK::ContentGenerator::ForgotPassword;
+use Mojo::Base 'WeBWorK::ContentGenerator', -signatures;
+
+=head1 NAME
+
+WeBWorK::ContentGenerator::ForgotPassword - render the forgot password form
+
+=cut
+
+use Email::Stuffer;
+use Email::Sender::Transport::SMTP;
+use Digest::SHA          qw(sha256_hex);
+use Math::Random::Secure qw(irand);
+
+use WeBWorK::Debug       qw(debug);
+use WeBWorK::Utils       qw(createEmailSenderTransportSMTP);
+use WeBWorK::Utils::Logs qw(writeCourseLog);
+
+# Override the can method to disable links for the forgot password page.
+sub can ($c, $arg) {
+	return $arg eq 'links' ? 0 : $c->SUPER::can($arg);
+}
+
+sub initialize ($c) {
+	my $ce = $c->ce;
+
+	$c->stash->{footerWidthClass} = 'col-xl-5 col-lg-6 col-md-7 col-sm-8';
+
+	return unless $c->param('request_password_reset') && $ce->{password_reset_sender_email};
+
+	my $user = $c->db->getUser($c->param('user'));
+
+	# Send a password reset email to the user if the user exists, the user has an email address set, the user has the
+	# permission to change their password, and the user already has a password (the user cannot create a password via
+	# the password reset flow).
+	return
+		unless $user
+		&& $user->email_address =~ /\S/
+		&& $c->authz->hasPermissions($user->user_id, 'change_password');
+
+	my $password = $c->db->getPassword($user->user_id);
+	return unless $password && $password->password;
+
+	my $resetToken = join('', map { [ 0 .. 9, 'a' .. 'z' ]->[ irand(36) ] } 1 .. 64);
+	$password->reset_token(sha256_hex($resetToken));
+	$password->reset_token_expiration(time + 900);
+	eval { $c->db->putPassword($password) };
+	if ($@) {
+		$c->log->error('Unable to save password reset token to database for user ' . $user->user_id . ": $@");
+		return;
+	}
+
+	my $resetURL = $c->url_for('reset_password')->to_abs->query(user => $user->user_id, token => $resetToken);
+
+	my $email =
+		Email::Stuffer->to($user->rfc822_mailbox)
+		->from($ce->{password_reset_sender_email})
+		->subject($c->maketext('Reset Password'))
+		->text_body($c->maketext(
+			"Use the following link to reset your password: [_1]\n\n"
+			. "That link will only be valid for 15 minus (until [_2]).",
+			$resetURL,
+			$c->formatDateTime($password->reset_token_expiration, 'datetime_format_long')
+		))
+		->html_body($c->tag(
+			'html',
+			$c->tag(
+				'body',
+				$c->c(
+					$c->tag('div', $c->link_to($c->maketext('Reset password') => $resetURL)),
+					$c->tag(
+						'div',
+						$c->maketext(
+							'That link will only be valid for 15 minutes (until [_1]).',
+							$c->formatDateTime($password->reset_token_expiration, 'datetime_format_long')
+						)
+					)
+				)->join('')
+			)
+		))->header('X-Remote-Host' => $c->tx->remote_address || 'UNKNOWN');
+
+	eval {
+		$email->send_or_die({
+			transport => createEmailSenderTransportSMTP($ce),
+			$ce->{mail}{set_return_path} ? (from => $ce->{mail}{set_return_path}) : ()
+		});
+		debug 'Successfully sent password reset email to ' . $user->email_address;
+	};
+	if ($@) {
+		my $exception_message = ref($@) ? $@->message : $@;
+		$c->log->error('Error sending password reset email to ' . $user->email_address . ": $exception_message");
+	}
+
+	writeCourseLog($ce, 'login_log', 'PASSWORD RESET REQUEST user_id=' . $user->user_id);
+
+	return;
+}
+
+1;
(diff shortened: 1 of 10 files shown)


taniwallach:
I would recommend that requests to reset the password and successful resets be logged to the `login.log` file.

Maybe when a user has a MFA secret saved in the course, a password change should depend on providing an valid OTP result. That would make the system more secure against possibly malicious use.

drgrice1:
> I would recommend that requests to reset the password and successful resets be logged to the `login.log` file.

I will add this I suppose.

> Maybe when a user has a MFA secret saved in the course, a password change should depend on providing an valid OTP result. That would make the system more secure against possibly malicious use.

This is true, and I will see about implementing this.  This may take some effort though, and may need to wait.

taniwallach:
Don;t hold things back about the OTP side, it would be nice to have, as it protects from an email compromise becoming a compromise of webwork. The logging is in my opinion more important, in case issues come up with claims of account compromise via the password reset system, it provides some data on when it occurred.

drgrice1:
@taniwallach: I have added the two factor authentication requirement when it is enabled, and the login log logging.

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
```diff
@@ -0,0 +1,97 @@
...
+	my $ce = $c->ce;
+
+	$c->stash->{footerWidthClass} = 'col-xl-5 col-lg-6 col-md-7 col-sm-8';
+
+	return unless $c->param('request_password_reset') && $ce->{password_reset_sender_email};
+
+	my $user = $c->db->getUser($c->param('user'));
+
+	# Send a password reset email to the user if the user exists, the user has an email address set, the user has the
+	# permission to change their passworrd, and the user already has a password (the user cannot create a password via
```
Alex-Jordan: `s/passworrd/password/`

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
```diff
@@ -0,0 +1,97 @@
...
+
+	my $resetURL = $c->url_for('reset_password')->to_abs->query(user => $user->user_id, token => $resetToken);
+
+	my $email =
+		Email::Stuffer->to($user->rfc822_mailbox)
+		->from($ce->{password_reset_sender_email})
+		->subject($c->maketext('Reset Password'))
+		->text_body($c->maketext(
+			"Use the following link to reset your password: [_1]\n\nThat link will only be valid until [_2].",
+			$resetURL, $c->formatDateTime($password->reset_token_expiration)
```
Alex-Jordan: My testing server can't send emails, so for now I am just reading code more than actual testing. Can you confirm that the time which is presented here is either in the user's local time locale, or the course's? If it is the server's time, that could be confusing. There are some servers (notably RS) that are in one time zone, but users actually do all their business in a different one.

lib/WeBWorK/ContentGenerator/ResetPassword.pm
```diff
@@ -0,0 +1,110 @@
+package WeBWorK::ContentGenerator::ResetPassword;
+use Mojo::Base 'WeBWorK::ContentGenerator', -signatures;
+
+=head1 NAME
+
+WeBWorK::ContentGenerator::ResetPassword - reset a users password if requested to do so
```
Alex-Jordan: I'm either nitpicking or being thorough, but this should be "user's" instead of "users".

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
drgrice1: This will be displayed in the same time zone that all other times are displayed in webwork2 which would be server time.  Perhaps this could be configured to show the time zone of the server, but there is no way that it can be displayed in the users time zone.  The server would not know that.

lib/WeBWorK/ContentGenerator/ResetPassword.pm
drgrice1: This is fixed.

Alex-Jordan:
My testing server can't actually send emails, and for now I'd rather not put time into setting that up. I did read through the code. And I loaded the branch and see the link, and I see the page with the message after you attempt to start the reset process. That all looks good. I am inclined to just approve this now and I will remember to try everything out once this is in production. Are there any concerns with that?

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
drgrice1: This is fixed.  For some reason `crate-ci/typos` is not showing this one in my editor.

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
Alex-Jordan: What do you think about it using the course's time zone? That is, the time zone that might be specified in the course.conf file for the course in question?

drgrice1:
That is fine.  We can always hotfix if needed!

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
Alex-Jordan: For the RS instance, the server is at UTC time, with users all over the place. It might just be better to say "15 minutes from now" and not try to give the exact time.

drgrice1: Actually, that is what I meant.  The time will be in the course's time zone.

Alex-Jordan (approved):

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
drgrice1: The `formatDateTime` method in the the `WeBWorK::ContentGenerator` module which is used here always defaults to the timezone and language of the course.

Alex-Jordan:
OK, I marked it approved!

Thanks for this. It will reduce a fair amount of email traffic I get from faculty asking for password resets.

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
drgrice1: The problem with saying "now" in an email is that "now" is relative.  At the time that the user reads the email, 15 minutes from now may be 30 minutes ago.

drgrice1: I change it to say `That link will only be valid for 15 minutes (until July 11, 2026, 7:18:31 PM CDT).`  So now the 15 minute time limit is included as well as the time of expiration including the time zone.

lib/WeBWorK/ContentGenerator/ForgotPassword.pm
```diff
@@ -0,0 +1,99 @@
...
+
+	my $resetURL = $c->url_for('reset_password')->to_abs->query(user => $user->user_id, token => $resetToken);
+
+	my $email =
+		Email::Stuffer->to($user->rfc822_mailbox)
+		->from($ce->{password_reset_sender_email})
+		->subject($c->maketext('Reset Password'))
+		->text_body($c->maketext(
+			"Use the following link to reset your password: [_1]\n\n"
+			. "That link will only be valid for 15 minus (until [_2]).",
```
pstaabp, line 61:
```suggestion
			. "That link will only be valid for 15 minutes (until [_2]).",
```

pstaabp (approved):
This works great.  

BTW, I don't have email on my development machine, so I dumped out the link with 
```
$c->log->info('Sending password reset email to ' . $user->email_address . ' with reset URL: ' . $resetURL);
```
near line 55 of `ForgotPassword.pm`

and the url dumped out in the terminal if you're running morbo or some log file if you set that.

templates/ContentGenerator/ResetPassword.html.ep
```diff
@@ -0,0 +1,49 @@
...
+							class => 'col-form-label col-sm-6' =%>
+						<div class="col-sm-6">
+							<%= password_field 'confirmPassword', id => 'confirmPassword', class => 'form-control',
+								dir => 'ltr', autocomplete => 'new-password', required => undef =%>
+						</div>
+					</div>
+					% if ($ce->two_factor_authentication_enabled
+						% && $authz->hasPermissions($userID, 'use_two_factor_auth')) {
+						<div class="row mb-2">
+							<%= label_for confirmPassword =>
```
Alex-Jordan, line 35: Claude flagged this while I was doing something else. Do you agree that this should be:

```suggestion
							<%= label_for otp_code =>
```

?

drgrice1: Yes, that should be `otp_code`.  That is a cut and paste error.

PR #3044, opened 2026-07-08

Alex-Jordan:
These are typos in `lib`, but not in `lib/WeBWorK/`.

The misnaming of `ResourceIri` as `ResourseIri` was consistent. Someone who knows Caliper should weigh in about this, if the package is used outside of this code. @andrew-gardener, does changing this typo matter?

I have not thought hard about the typo in `lib/WebworkWebservice/RenderProblem.pm`. If that code gets used, certainly it is broken with the typo in place.

### DIFF
--- a/lib/Caliper/Actor.pm
+++ b/lib/Caliper/Actor.pm
@@ -19,7 +19,7 @@ sub generate_anonymous_actor {
 
 sub generate_default_actor {
 	my ($ce, $db, $user) = @_;
-	my $resource_iri = Caliper::ResourseIri->new($ce);
+	my $resource_iri = Caliper::ResourceIri->new($ce);
 
 	return {
 		'id'   => $resource_iri->actor_homepage($user->user_id()),
--- a/lib/Caliper/Entity.pm
+++ b/lib/Caliper/Entity.pm
@@ -17,7 +17,7 @@ use Caliper::Actor;
 
 sub webwork_app {
 	my ($ce, $db) = @_;
-	my $resource_iri = Caliper::ResourseIri->new($ce);
+	my $resource_iri = Caliper::ResourceIri->new($ce);
 
 	# $ce doesn't have WW_VERSION when doing login/logout for some reason
 	my $seed_ce    = WeBWorK::CourseEnvironment->new;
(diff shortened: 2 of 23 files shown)


drgrice1:
I do not know `Caliper` particularly well.  However, I do know that if the typo is fixed consistently, it will not matter, and the code will still work.

drgrice1:
In fact, I am not entirely certain that the code works with that typo since the file is named `ResourceIri.pm` without the typo.  The `Caliper` code is highly suspect and unmaintained, and unless someone steps up to maintain it, it should be removed.

Alex-Jordan:
I added more to this. I won't add any more after this though. Any more will be in separate PRs. This now includes `lib/WeBWorK/`, but not `lib/WeBWorK/ContentGenerator/`, `lib/WeBWorK/Localize/`, or `lib/WeBWorK/Utuils/`. Those each have have many more typos.

One place that should get extra inspection is the changes to `lib/WeBWorK/Authen/Saml2.pm`.

drgrice1:
Is there a spell checker for GitHub conversations? `lib/WeBWorK/Utuils`

Alex-Jordan:
Haha, yeah I saw that after I posted and left it for fun.

`lib/WeBWorK/Localize/` just has typos that are derived from other typos. So I'm not going in there. And it turns out `lib/WeBWorK/Utils/` has few enough typos that I will add it here. 

I'll save `lib/WeBWorK/ContentGenerator/` fir another day.

drgrice1 (approved):
Looks good, and some of the typos were causing issues.

The typo in `lib/WeBWorK/Authen/Saml2.pm` is a local typo, and fixing it won't break anything (or fix anything other than making the spelling correct).

The typo in `lib/WeBWorK/ConfigValues.pm` was preventing the `LTI{v1p3}{PlatformID}` variable from working in the LTI tab. So this is a bug fix.

The typo in `lib/WebworkWebservice/RenderProblem.pm` was clearly preventing the preview button from working as intended. Although, some might argue that the intended functionality of showing answer evaluator messages for an answer preview is not desirable. But that is another debate.

drgrice1:
I just noted that there were two typos in `lib/WeBWorK/Saml2.pm`.  The `$ceritificateOnly` typo was the local one I referred to.  The other typo certainly needs to be fixed, but would only cause an "uninitialized string" warning when debugging is enabled, and prevent the debugging message from being as useful as intended.

xcompass:
PR #3044 looks good to me. It includes all the changes in my removed commit as well.

xcompass (approved):

PR #3045, opened 2026-07-08

Alex-Jordan:
These are the typos from `lib/WeBWorK/ContentGenerator/`. There are not as many as I thought. Maybe when I did an initial count there was something wrong with the sytnax I used.

In a few places, things like variable names are changed. I believe they are changed consistently over their scope and the changes should cause no trouble. But they will lighten future scans for typos.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
@@ -306,13 +306,13 @@ sub process_search ($c, @dbsearch) {
 	# Build a hash of MLT entries keyed by morelt_id
 	my %mlt = ();
 	my $mltind;
-	for my $indx (0 .. $#dbsearch) {
-		$dbsearch[$indx]{oindex} = $indx;
-		if ($mltind = $dbsearch[$indx]{morelt}) {
+	for my $index (0 .. $#dbsearch) {
+		$dbsearch[$index]{oindex} = $index;
+		if ($mltind = $dbsearch[$index]{morelt}) {
 			if (defined($mlt{$mltind})) {
-				push @{ $mlt{$mltind} }, $indx;
+				push @{ $mlt{$mltind} }, $index;
 			} else {
-				$mlt{$mltind} = [$indx];
+				$mlt{$mltind} = [$index];
 			}
 		}
 	}
@@ -325,12 +325,12 @@ sub process_search ($c, @dbsearch) {
 		if (scalar(@idlist) > 1) {
 			my $leader = WeBWorK::Utils::ListingDB::getMLTleader($c, $mltid) || 0;
 			my $hold   = undef;
-			for my $subindx (@idlist) {
-				if ($dbsearch[$subindx]->{pgid} == $leader) {
-					$dbsearch[$subindx]->{children} = [];
-					$hold = $subindx;
+			for my $subindex (@idlist) {
+				if ($dbsearch[$subindex]->{pgid} == $leader) {
+					$dbsearch[$subindex]->{children} = [];
+					$hold = $subindex;
 				} else {
-					$dbsearch[$subindx]->{noshow} = 1;
+					$dbsearch[$subindex]->{noshow} = 1;
 				}
 			}
 			do {    # we did not find the leader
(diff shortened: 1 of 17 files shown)


pstaabp (approved):

lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
```diff
@@ -306,13 +306,13 @@ sub process_search ($c, @dbsearch) {
...
-		if ($mltind = $dbsearch[$indx]{morelt}) {
+	for my $index (0 .. $#dbsearch) {
+		$dbsearch[$index]{oindex} = $index;
+		if ($mltind = $dbsearch[$index]{morelt}) {
 			if (defined($mlt{$mltind})) {
-				push @{ $mlt{$mltind} }, $indx;
+				push @{ $mlt{$mltind} }, $index;
 			} else {
-				$mlt{$mltind} = [$indx];
+				$mlt{$mltind} = [$index];
```
drgrice1, line 315: I would say that technically this is a false positive.  I suspect this was an intended short form of the `$index` used for the variable name.  This is caught by crate-ci/typos, but `$subindx` below is not, and that variable was spelled that way, I think clearly, for the same reason.

In any case, if this is changed, then `$subindx` below should also be changed for consistency.

drgrice1 (approved):
I think this is okay.

I have done some testing with `crate-ci/typos`, and have concluded that it would not be a good idea to enable it as a work flow.  In testing both with the webwork2 and PG code as well as with several other projects, I am seeing way to many false positives.  So I don't think it would actually be safe to use in automated checking.  It is fine to use it manually though.

Alex-Jordan:
Did you see how to whitelist false positives? When building up these PRs, I had a `_typos.toml` file. I'm not pushing for this, just want to point out that possibility. That the repo would have a `_typos.toml` file, and new entries would need to be added occasionally to that file.

lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
Alex-Jordan: This is one that I almost left alone. But I just liked the readability of `index` over `indx`. I could have added it to the `_typos.toml` file but I opted to replace it here. I missed `subindx` though. I'll get to that.

drgrice1:
Yeah, but that seems like it would be tedious to maintain.

For instance, `crate-ci/typos` sees the variable `$pn` in `templates/ContentGenerator/GatewayQuiz.html.ep`, and says that should be `on`.  So would you whitelist `pn`?  There are numerous cases of two letter variable names like that that it says should be changed.

In the file `templates/ContentGenerator/Instructor/Scoring.html.ep` it sees the string `ful` and say it should be `full` (but yet it ignores the other string `scr` there).  You wouldn't want to whitelist `ful`.

In the file `bin/OPLUtils.pm` there is a `writeJSONtoFile` method.  `crate-ci/typos` says that should be `writeJSONNotFile`.  Perhaps the method could be better named with camel case, but it is not an entirely invalid way to name the method.

In other projects I saw some really dangerous things that it was making suggestions for.  For example, one project has a list of YouTube video ids, and it was suggesting changes to parts of those.

In general, I think that using it for manual and double checked spell checking is good, and in fact I have enabled it as a language server in vim to get spell checking there.

lib/WeBWorK/ContentGenerator/Instructor/SetMaker.pm
drgrice1: But that is part of my issue with the whitelist.  I don't think that `indx` should be whitelisted.  In general, it would be better for the variable here to be named `$index` and not `$indx`, but it is well established that I don't like abbreviations, and some other developers do like to abbreviate.

Alex-Jordan: I agree. That's why I changed it to `index` instead of adding it to my local whitelist. But I am good with just making this a new annual thing, reviewing for typos this way. It's similar (but not quite as automated) as when you run the localizations.

I'm about to go change `subindx` too.

Alex-Jordan: `s/subindx/subindex/` completed

drgrice1: I think that if you (or someone) were to get a configuration setup that works okay, then it wouldn't hurt to try a workflow.  The worst case scenario we decide it is isn't good and remove it.

PR #3046, opened 2026-07-08
Add error handling in the pg problem editor

pstaabp:
This adds error handling for the PGML converter to the PGProblemEditor.  This mimics the pg tidy interface and code. 

If there is an error in parsing the PG problem--currently only the loadMacros command will throw an error--then the error is shown in the right pane of the PG problem editor.  Upon fixing the error and rerunning the "Convert to PGML" in the "Code Maintenance" table, the code is updated and pre-rendered.  Note, previously, the code was not rerendered after code conversion. 

Also, openwebwork/pg#1425 is needed to test this.


pstaabp:
Close in favor of #3101.

PR #3054, opened 2026-07-13
Build the library browser alert toast with DOM methods to avoid DOM XSS

xcompass:
## Summary

The `alertToast` helper in `htdocs/js/SetMaker/setmaker.js` (the library browser) interpolated its `title` and `msg` arguments directly into an `innerHTML` template. All current callers pass plain text, but several of those strings incorporate server-provided content (e.g. RPC error messages), and interpolating DOM text into `innerHTML` lets it be reinterpreted as HTML. CodeQL flags this pattern as a `js/xss-through-dom` alert.

Per review, this now removes the `innerHTML` usage entirely: the toast is built with `document.createElement` and the title and message are assigned with `textContent`, so caller-supplied strings are always treated as text and can never be parsed as markup. The construction mirrors the toast already built this way in `htdocs/js/TagWidget/tagwidget.js`.

The separate `alertToast` in `htdocs/js/GatewayQuiz/gateway.js` is intentionally left alone — it renders HTML from server-controlled data attributes by design.

## Testing

- `node --check htdocs/js/SetMaker/setmaker.js` passes (Node 20).
- `prettier --check` on the modified file passes (matches the check-formats CI workflow).
- Behavior is unchanged for all existing callers: the rendered toast markup is identical (same elements, classes, and attributes as the previous template); a title or message containing markup characters displays literally instead of being parsed as HTML.

### DIFF
--- a/htdocs/js/SetMaker/setmaker.js
+++ b/htdocs/js/SetMaker/setmaker.js
@@ -19,16 +19,37 @@
 			'p-3'
 		);
 		toastContainer.style.zIndex = 20;
-		toastContainer.innerHTML =
-			'<div class="toast bg-white" role="alert" aria-live="assertive" aria-atomic="true">' +
-			'<div class="toast-header">' +
-			`<strong class="me-auto">${title}</strong>` +
-			'<button type="button" class="btn-close" data-bs-dismiss="toast" aria-label="close"></button>' +
-			'</div>' +
-			`<div class="toast-body alert ${good ? 'alert-success' : 'alert-danger'} mb-0 text-center">${msg}</div>` +
-			'</div>';
+
+		const toast = document.createElement('div');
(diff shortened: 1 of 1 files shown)


drgrice1 (commented):
I was a bit lazy when I implemented this and used `innterHTML`.  It would be better to remove that entirely and build the HTML with `document.createElement` rather than using `innterHTML` and then `querySelector` to find the contents.

xcompass:
Done — removed the `innerHTML` usage entirely; the toast is now built with `document.createElement` and `textContent`, following the same construction as the toast in `htdocs/js/TagWidget/tagwidget.js`.

drgrice1 (approved):
This looks good now.

pstaabp (approved):
👍

PR #3057, opened 2026-07-13
Allow overriding the OPL metadata release API URL via an environment variable

xcompass:
## Summary

`bin/download-OPL-metadata-release.pl` hardcodes the GitHub API URL used to locate the latest OPL metadata release (`https://api.github.com/repos/openwebwork/webwork-open-problem-library/releases/latest`). Sites that maintain a fork of the webwork-open-problem-library (for example, to carry local problems and publish their own metadata releases) cannot use the script against their fork without patching it.

This change allows the release API URL to be overridden by setting the `OPL_METADATA_RELEASE_API_URL` environment variable:

```
OPL_METADATA_RELEASE_API_URL=https://api.github.com/repos/<org>/webwork-open-problem-library/releases/latest \
    ./bin/download-OPL-metadata-release.pl
```

When the variable is not set, the script behaves exactly as before, using the openwebwork URL. The rest of the fetch/decode/restore logic is untouched, and a brief comment documents the override at the point of use.

## Testing

- Ran `perltidy` v20260204 (the version pinned in `.github/workflows/check-formats.yml`) with the repository `.perltidyrc`; the committed file is idempotent under it.
- `perl -c` passes on the full script (checked with the script's WeBWorK/Helper module dependencies stubbed, since compiling those requires a full runtime environment).
- Verified the default (variable unset) resolves to the same openwebwork URL string as before the change, so existing installations see no behavior change.

Dependencies: none.

### DIFF
--- a/bin/download-OPL-metadata-release.pl
+++ b/bin/download-OPL-metadata-release.pl
@@ -32,8 +32,10 @@ my $ce = WeBWorK::CourseEnvironment->new({ webwork_dir => $ENV{WEBWORK_ROOT} });
 die "The WeBWorK temporary directory $ce->{webworkDirs}{tmp} does not exist or is not writable."
 	if (!-d $ce->{webworkDirs}{tmp} || !-w $ce->{webworkDirs}{tmp});
 
-my $releaseDataFF =
-	File::Fetch->new(uri => 'https://api.github.com/repos/openwebwork/webwork-open-problem-library/releases/latest');
+# Set the OPL_METADATA_RELEASE_API_URL environment variable to override the release API URL,
+# for example, for sites that use a fork of the OPL.
+my $releaseDataFF = File::Fetch->new(uri => $ENV{OPL_METADATA_RELEASE_API_URL}
+		// 'https://api.github.com/repos/openwebwork/webwork-open-problem-library/releases/latest');
 my $file        = $releaseDataFF->fetch(to => $ce->{webworkDirs}{tmp}) or die $releaseDataFF->error;
 my $path        = Mojo::File->new($file);
 my $releaseData = decode_json($path->slurp);


bin/download-OPL-metadata-release.pl
```diff
@@ -32,11 +32,14 @@ BEGIN
...
-my $path        = Mojo::File->new($file);
-my $releaseData = decode_json($path->slurp);
+# Set the OPL_METADATA_RELEASE_API_URL environment variable to override the release API URL,
+# for example, for sites that use a fork of the OPL.
+my $releaseApiUrl = $ENV{OPL_METADATA_RELEASE_API_URL}
+	// 'https://api.github.com/repos/openwebwork/webwork-open-problem-library/releases/latest';
+my $releaseDataFF = File::Fetch->new(uri => $releaseApiUrl);
+my $file          = $releaseDataFF->fetch(to => $ce->{webworkDirs}{tmp}) or die $releaseDataFF->error;
+my $path          = Mojo::File->new($file);
+my $releaseData   = decode_json($path->slurp);
```
drgrice1: Please change this to

```suggestion
# Set the OPL_METADATA_RELEASE_API_URL environment variable to override the release API URL,
# for example, for sites that use a fork of the OPL.
my $releaseDataFF = File::Fetch->new(uri => $ENV{OPL_METADATA_RELEASE_API_URL}
		// 'https://api.github.com/repos/openwebwork/webwork-open-problem-library/releases/latest');
my $file        = $releaseDataFF->fetch(to => $ce->{webworkDirs}{tmp}) or die $releaseDataFF->error;
my $path        = Mojo::File->new($file);
my $releaseData = decode_json($path->slurp);
```

There is no need for the intermediate `$releaseApiUrl` variable that is only used once.  Just put its value in the one place it is used.

drgrice1 (changes requested):
It is questionable if this should be accepted.  Should we support using external repositories for the OPL?  Those that are maintaining these repositories should submit their problems to the actual OPL instead.  There are other methods for using external libraries for problems that can be used.

This will need discussion.

xcompass:
We have instructors who are not conformable sharing their libraries publicly. So we have to host our own repo. For those are OK with sharing, we let them open PRs to the OPL directly.

bin/download-OPL-metadata-release.pl
xcompass: Done — applied your suggestion; the `$ENV{OPL_METADATA_RELEASE_API_URL} // ...` default is now inlined in the `File::Fetch->new` call with no intermediate variable.

drgrice1:
We have discussed this, and have decided that we are only going to support using the official OPL, and not some fork elsewhere.  There are other ways for maintaining local libraries that are not publicly shared in any case.

So I am closing this pull request.

Issue #1156, opened 2020-11-05
Feature requests for Gateway quizzes


taniwallach:
- auto-save / auto-submit during GW quizzes to prevent data loss
  - save answers button per question [@dlglin ]
 - create a method for a PG problem to be aware it is in GW vs homework usage mode [Edward Sternin]

[initially collected from participant comments during the "Lessons From the Major Online Systems" session at https://rederly.com/finals-conference.]

Alex-Jordan:
* an easier way for instructors to review submitted answers on test versions and edits scores
* reduce load on the server: make it so that when a student clicks to grade, the only problems that are processed are those for which the student has submitted a new answer. (We have students who will enter an answer to question 1 and click submit, and all 20 problems in the quiz get processed. And they do this as they go. Perhaps only an issue when they are allowed more than one graded submission)

> create a method for a PG problem to be aware it is in GW vs homework usage mode [Edward Sternin]

I have a hack for this. Not that I want to promote hacks, but until the feature is built, it has worked for me. You can look at the environment variables and see if the GW suffix is defined. As in `if($envir{QUIZ_PREFIX}) {do stuff};`.

taniwallach:
As a reminder to myself and others: Tom (?) from Rederly mentioned using some sort of timer on changes to input boxes to trigger their auto-save of answers.

esternin:
Actually the more relevant approach would be to determine whether the hints are enabled or not.  One of the differences between using the same problem in homework and in a gateway/quiz is to enable/disable hints, so displaying all text that is referring to hints ("You can get a hint if you do not get the answer in three tries" or similar) should be conditional on whether hints are enabled.  And, this will take care of those cases where hints are actually allowed during GW as well. 

But I am not clear: *is there* such an envir variable already, or is this a "hack" in the sense that you need to add the variable somewhere yourself?  Surely, the renderer knows whether hints are enabled or not, but how?

Alex-Jordan:
Regarding the hints issue.

[...]

PR #3059, opened 2026-07-13
Periodically autosave answers on timed gateway tests

xcompass:
Periodically submit the gwquiz form in the background as a preview request on timed gateway tests, so the server saves the student's latest answers and work is not lost to a crash or dropped connection.

Per review: disabled by default. Set `$gatewayAutosaveInterval` (seconds) in the course environment to enable and control the period. Saves are skipped when answers have not changed since the last save, and the interval is randomized upward by up to a third to stagger load. Not active when an instructor acts as another user.

No server code changes; the preview pipeline already stores last answers.


drgrice1 (changes requested):
This is not the right way to do this.  Submitting a test already causes a lot of server overhead, and so making a test preview auto submit on a regular basis is really not going to work.  Particularly with no option to disable it, and no options to set the period or anything.

This will need more thought and better development.  In any case, this is certainly something that will need to wait until the next release.  Please re-target to the develop branch.

xcompass:
Changed target to develop. Happy to discuss it further. I'll add an option and have it disabled by default.

xcompass:
Added the option: `$gatewayAutosaveInterval` in defaults.config, in seconds, 0/off by default — nothing renders and no timer is started unless a site or course sets it. The period is fully configurable through it, and saves only happen when answers actually changed. Also rebased onto develop as requested.

dlglin:
There have been many discussions about how to improve the capture of student answers if they don't submit the exam (including #1156).

At this point having one question per page has almost entirely solved the issue for me since answers are saved when a student navigates between pages.  Historically this caused other problems because students would submit the entire exam after answering the first question, but there are now multiple features that have reduced that to a very rare occurrence; namely the popup that appears warning students of this, and in the 2.21 release the "previous page/next page" buttons that are more prominent than the "Grade Test" button.

Here are a couple of alternative ideas for saving answers:

1. Have a "save answer(s)" button next to each question.  The easy way to implement this would be to just have this button trigger an answer preview since that saves the student's answers.
2. Set up a javascript trigger that saves an answer as soon as a student navigates out of an answer blank.  This would be complicated to implement since there are things like multianswer to account for, and it leads to similar load issues to what @drgrice1 described.  I'm also not convinced that it would provide much more benefit than one question per page.

xcompass:
Thanks, @dlglin, for the suggestions. Our primary use case for this feature is to prevent students from losing all their answers if their computer crashes during an exam. We’ve encountered this several times in the past, particularly during large exams with more than 1,300 students, where the likelihood of such an unfortunate event is much higher. Losing their work during a final exam is extremely frustrating and stressful for students.

drgrice1:
We will try work on implementing a better approach for this that can save student answers in a test that does not require a hack to submit the test via a preview in the background for the next release.  This approach is not acceptable though, so I am closing this pull request.

PR #3062, opened 2026-07-15
Make the color scheme chooser active theme check work.

drgrice1:
In the HTML markup for the color scheme chooser in `templates/layouts/system.html.ep` each of the theme buttons in the dropdown have a check icon.  However, that check icon has `d-none` set, and it is always that way.  So the checks are never shown. The javascript is supposed to make the check for the active theme visible, but does not. So make that happen.

This is a result of copying the HTML that Bootstrap actually uses for the theme chooser on their documentation page, and using the JavScript they demonstrate on their documentation page, but their implementation and their demonstration do not quite match.


pstaabp (approved):

somiaj (approved):
Oh hey, there is a check there!!

Issue #794, opened 2017-07-21
Conditional Release sets


pstaabp:
If Conditional Release is turned on for a course and for a problem set, a set is added that doesn't exist, a warning will popup, but that set is still set in the database.  A better way to do this is to remove the nonexistent set before writing to the DB. 

In addition, I don't think one should be able to include the current set as in the conditional release.  It appears that it will tell you that you can access it once you score a certain percent on the following sets (which includes itself).  

Lastly, I wonder if we should also check if the user is assigned the sets in the list?  --- This may be difficult to do in general.  

I unearthed these by adding conditional release to webwork3, and will probably build these into webwork3.

Alex-Jordan:
When setting conditional release for a set, we should not ask the instructor to type a comma-separated list. This should be some sort of multi-select input that only offers valid existing sets in the first place.

I'm not sure if conditional release is also written to set definition files. But if it is, then when those are imported, you could also have this issue arise. And using a multi-select would not solve the problem entirely.

Alex-Jordan:
If we changed to using a multiselect list, would we continue with `restricted_release` being a comma-separated string in the database? If so, I am unsure how to get the form to submit the multiple selected items as a comma-separated string.

PR #3080, opened 2026-07-23
Make conditional release in the set details page use a select.

Alex-Jordan:
This is the first thing I've tried using Claude for. It addresses #794.

For this, I let Claude do it's thing. I monitored the process, and approved each request it made to alter code or run a test. I did not change anything that Claude did. I reviewed the code, and it seems fine to me. Although in places, conditional logic is maybe not the way I would have done it.

The feature works. Now instead of typing a comma-separated list of set IDs, you just select from a multiselect.

A few things are impressive, but one that stands out to me is that Claude knew to exclude the current set from the list of sets to conditionally release upon. Unless I'm missing something, I don't see that there was previously a check for that. And I did not prompt it to do this.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -14,7 +14,7 @@ use WeBWorK::Utils             qw(cryptPassword x);
 use WeBWorK::Utils::Files      qw(surePathToFile readFile);
 use WeBWorK::Utils::Instructor qw(assignProblemToAllSetUsers addProblemToSet);
 use WeBWorK::Utils::JITAR      qw(seq_to_jitar_id jitar_id_to_seq);
-use WeBWorK::Utils::Sets       qw(format_set_name_internal format_set_name_display);
+use WeBWorK::Utils::Sets       qw(format_set_name_display);
 require WeBWorK::PG;
 
 our @EXPORT_OK = qw(FIELD_PROPERTIES);
@@ -197,7 +197,7 @@ use constant FIELD_PROPERTIES => {
 		override  => 'any',
 		help_text => x(
 			'This set will be unavailable to students until they have earned the "Score Required for Release" on the '
-				. 'sets specified in this field.  The sets should be written as a comma separated list.'
+				. 'sets selected in this field.  Hold down Ctrl (or Cmd on Mac) to select or deselect multiple sets.'
 		)
 	},
 	restricted_status => {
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Also I should note that the time taken to fire up Claude, ask it to work, review its work, and test it out a few ways was about 20 minutes. Of course, this is not a super complicated feature.

BTW, while testing, I notice that when we export a set definition file, it does not carry the conditional release data. Should it? I rarely transfer set definition files anymore. I can't decide if it would be good for them to carry conditional release data or if that should be left for the new instructor to carry out.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -886,14 +886,31 @@ sub fieldHTML ($c, $userID, $setID, $problemID, $globalRecord, $userRecord, $fie
...
 					class => 'form-control form-control-sm',
-					$field eq 'restricted_release' || $field eq 'source_file' ? (dir => 'ltr') : ()
+					$field eq 'source_file' ? (dir => 'ltr') : ()
 				);
-				if ($field eq 'problem_seed') {
+				if ($field eq 'restricted_release') {
+					# Only sets that currently exist in the course are valid choices.
+					my @setIDs = sort grep { $_ ne $setID } $db->listGlobalSets;
+					my %isSelected =
+						map { $_ => 1 } split(/\s*,\s*/, $value =~ s/^\s*|\s*$//gr);
```
drgrice1: This fits nicely onto one line. Prefer that if possible.  So make this

```suggestion
					my %isSelected = map       { $_ => 1 } split(/\s*,\s*/, $value =~ s/^\s*|\s*$//gr);
```

The spacing in that accounts for what `perltidy` does to line up with the line above it.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -1411,7 +1428,14 @@ sub initialize ($c) {
...
 					my @paramValues = $c->param("set.$setID.$field");
-					my $param = @paramValues > 1 && $paramValues[0] eq 'numeric' ? $paramValues[1] : $paramValues[0];
+					my $param;
+					if ($field eq 'restricted_release') {
+						$param = join(',', grep {$_} @paramValues);
+					} elsif (@paramValues > 1 && $paramValues[0] eq 'numeric') {
+						$param = $paramValues[1];
+					} else {
+						$param = $paramValues[0];
+					}
```
drgrice1: I would rather see this left as a ternary expression.  If done carefully with `perltidy` as in the following suggestion it gives a nice organized and quite readable layout.

```suggestion
					my $param =
						$field eq 'restricted_release'                       ? join(',', grep {$_} @paramValues)
						: (@paramValues > 1 && $paramValues[0] eq 'numeric') ? $paramValues[1]
						:                                                      $paramValues[0];
```

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -1488,13 +1512,16 @@ sub initialize ($c) {
...
 				my @paramValues = $c->param("set.$setID.$field");
-				my $param       = @paramValues > 1 && $paramValues[0] eq 'numeric' ? $paramValues[1] : $paramValues[0];
+				my $param;
+				if ($field eq 'restricted_release') {
+					$param = join(',', grep {$_} @paramValues);
+				} elsif (@paramValues > 1 && $paramValues[0] eq 'numeric') {
+					$param = $paramValues[1];
+				} else {
+					$param = $paramValues[0];
+				}
```
drgrice1: Also make this a ternary expression again.  This is almost identical to the other case.

drgrice1 (approved):
This generally looks good.  I just have some minor quibbles on the code formatting.

Alex-Jordan:
I made those changes. I noted that there were places where Claude's conditional logic code was not the way I would have done it, and these ternary expressions were those places. For a first run, I thought I'd leave things exactly as Claude made them. Just to see. But if I continue using Claude I'll be making more changes. I already needed that with #3081.

pstaabp (approved):
Much nicer interface.

PR #3081, opened 2026-07-23
Allow users (with high enough permission) to set their answer date to…

### LINKED ISSUE #975: Feature request; student's closing their access to a set
From a teacher in my department:

I have several online students that want me to change their due dates so they can see the correct answers once they have made their final submission, final in the sense they don't want to do any more.  This would be OK in a class of 19 students but it is a pain in the neck in a class of 100 students.  It would be nice if there was a final submission button for the webwork sections so it would do so automatically.  That is, the answers would be given then.

Alex-Jordan:
… right now

This is another Claude powered PR. This would address #975. This time, prpobably owing to my bad prompting, I had to go through several iterations. But the code looks good afaict and it tested well when I tried it out.

### DIFF
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -739,6 +739,7 @@ $authen{admin_module} = ['WeBWorK::Authen::Basic_TheLastOption'];
 	change_password            => "student",
 	change_email_address       => "student",
 	change_pg_display_settings => "student",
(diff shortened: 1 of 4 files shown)


pstaabp (approved):
Looks good.

drgrice1 (approved):
Looks good.

PR #3082, opened 2026-07-23
Guidance for an instructor who has not set the LMS context ID

Alex-Jordan:
When the table is presented with an LMS context ID (after a failed attempt to use the deep linking tool) this adds some explicit guidance for an instructor about what to do with that information.


drgrice1 (commented):
It may also be worth noting that the LTI tab may not be enabled.  If the LTI tab is not enabled, and this message is shown, then the message may be confusing and not helpful.

Alex-Jordan:
OK, it's a bid more verbose now. Let me know if you have ideas for improving this. I just want to increase instances of faculty getting things sorted without calling for help.

drgrice1 (approved):
Looks good now.

somiaj (approved):

Issue #1078, opened 2020-02-21
Feature request: Forcing password change


taniwallach:
See: http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2750

It would be helpful if there was a manner to force users issued a default password to change it or to mark passwords as "expired".

PR #3083, opened 2026-07-23
Force password reset.

Alex-Jordan:
This is another Claude-powered PR (the third I have tried). It would address #1078. This one required the most interventions yet. My initial prompt basically had the feature working, but there were UI issues when I tried to use the feature. There were about ten adjustments that I asked it to make.

There is a new `must_reset_password` field in the `password` table for a course. For users where this is 0, null, or the user simply did not have a record in that table, there should be no changes.

If this record is 1 for a user, then they can't go anywhere in the course until they reset their password. (Even if a user is logged in and happily plodding a long, if the instructor flags them as needing to reset their password, the user will be sent to the reset password screen on their next page load.) The reset password screen is modeled after the 2FA screen.

The instructor can manually flag (or unflag) this for a user via a new checkbox when editing users in the Accounts Manger. It's right next to the existing checkbox for deleting a user's password.

Also when editing users in the Accounts Manager, if a user had no password but the instructor types one in, the checkbox for forcing a password reset will automatically check. (The instructor can uncheck if they want.) If the user already had a password and the instructor is typing in a new one, that will not automatically check the box for forced password reset. (But the instructor can check it if they want.)

When a new user is added, maybe no password record is created for that user. In that scenario, nothing happens regarding forced password reset. Otherwise, a password record is created. And:

- if the user was added with a classlist file that already had a crypted password, the forced password reset will not be flagged.
- in all other cases (added with an unencrypted password in a classlist file, added via the Accounts Manager with some sort of password) then the flag will be set to force a password reset. Note that the instructor could unset this after the new user account is created if they want.

I think that fully describes the changes here. Since we are studying the use of AI coding assistants, I thought it might be interesting to post all my prompts here. I find it interesting that several of my prompts are not hyper-specific, but it knew what to do. For example, the way it handled alignment of the checkboxes in the Accounts Manager editor. And also, it did smart things that I did not even ask for. Like the details of how it auto-checks and auto-unchecks the new checkbox in the scenario when a user has no password record.

## What I prompted

1. Add a new column to a course's password table in the database to prompt password reset. Whenever a user successfully logs in to a course, if this value is set for that user, they cannot visit any pages in the course. Instead, they are presented with a page to change their password. They must actually change their password to a new password, and there should be a confirmation input field for the user to enter their password a second time. Once their password is changed, the "prompt for password reset" value in the database should be unset. The "prompt for password reset" value should be set whenever a new user is added to a course using a method other than importing a classlist file, except not in situations where the user doesn't even have a password. Also it should be set for a user imported with a classlist file, but the crypted password field was not set in that classlist file. Laslty, when a user is in the Accounts Manager and is editing users, the Password column should have an additional checkbox that is used to set the "prompt for password reset" value in the database.

2. When I edited a user in Accounts MAnager and used the checkbox to delete their password, I was given multiple instances of the error "Argument "" isn't numeric in numeric ne (!=) at /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm line 864". That appears to be related to the new changes. Can you account for that error and fix it?

3. A user had no password. And then in the Accounts Manager, I assigned a password to them. This should set the new must_reset_password database field to 1, but it remains set to 0.

4. In the Accounts Manager, when editing users, when the

5. In the Accounts Manager, when editing users, when the "delete password" checkbox is hidden, then the new "force password reset" box is the first checkbox, and it aligns with other rows badly. It aligns with the "delete password" box in other rows. Can we align the "force password reset" checkboxes better?

6. after making changes and testing, you can let me handle regenerating npm assets and restarting the webwork2 service

7. When I go to the Accounts Manager, edit page, and a user has no password, the "force password reset" checkbox is checked by default. However, I think it should not be checked unless/until I enter a new password for that user. Mainly, if I do not enter new passwords, and I click Save Edit, I do not want that to cause the must_reset_password field to be set.

8. When a user has the must_reset_password field set and are taken to the page where they set a new password, there should be an option to return to the original login page and log in as someone different.

9. The new "Log in as a different user" link is good, but instead of taking the user to the logout page, can it bypass that and take the user all the way to the login page?

10. When a user is required to change their password, they should not be allowed to change it to the same thing as their old password.

11. Can you make a text file that lists all of the prompts I gave you this session?

### DIFF
--- a/htdocs/js/AddUsers/add-users.js
+++ b/htdocs/js/AddUsers/add-users.js
@@ -13,4 +13,25 @@
 
 	passwordSelect.addEventListener('change', setPlaceholders);
 	setPlaceholders();
+
+	// Each "force password reset" checkbox starts out unchecked. Automatically check it once a password is typed
+	// into that row's password field, or once a fallback password source other than "None" is selected (since that
+	// fallback value will be used as the password for any row left blank). If the instructor manually toggles a
+	// checkbox, stop overriding that row's checkbox for the rest of the page's lifetime.
+	const manuallySet = new Set();
+	for (const autoCheckbox of document.querySelectorAll('input[data-auto-check="password"]')) {
+		const passwordField = autoCheckbox.closest('tr')?.querySelector('.new_password');
+		if (!passwordField) continue;
+
+		autoCheckbox.addEventListener('change', () => manuallySet.add(autoCheckbox));
+
+		const updateCheckbox = () => {
+			if (!manuallySet.has(autoCheckbox))
+				autoCheckbox.checked = passwordField.value !== '' || passwordSelect.value !== '';
+		};
+
+		passwordField.addEventListener('input', updateCheckbox);
+		passwordSelect.addEventListener('change', updateCheckbox);
+		updateCheckbox();
+	}
 })();
--- a/htdocs/js/UserList/userlist.js
+++ b/htdocs/js/UserList/userlist.js
@@ -145,7 +145,26 @@
 		});
 	}
 
-	for (const deleteCheck of document.querySelectorAll('input[name$="password_delete"]')) {
-		new bootstrap.Tooltip(deleteCheck, { placement: 'top', fallbackPlacements: [] });
+	for (const tooltipCheck of document.querySelectorAll(
+		'input[name$="password_delete"]:not([disabled]), input[name$="password_must_reset"]'
+	)) {
+		new bootstrap.Tooltip(tooltipCheck, { placement: 'top', fallbackPlacements: [] });
+	}
+
+	// For a user with no password, the "force password reset" checkbox starts out unchecked (there is nothing to
+	// reset yet). Automatically check it once a new password is actually typed into the corresponding password
+	// field, and uncheck it again if that field is cleared back out. If the instructor manually toggles the
+	// checkbox themselves, stop overriding their choice for the rest of the page's lifetime.
+	for (const autoCheckbox of document.querySelectorAll('input[data-auto-check="password"]')) {
+		const passwordField = autoCheckbox.closest('.row')?.querySelector('input[name$=".password"]');
+		if (!passwordField) continue;
+
+		let manuallySet = false;
+		autoCheckbox.addEventListener('change', () => {
+			manuallySet = true;
+		});
+		passwordField.addEventListener('input', () => {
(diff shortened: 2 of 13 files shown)


drgrice1:
A few notes on this.

I think that this may need to wait until the next release.  With the release imminent at this point, I don't think it is a good idea to add something like this that makes such a change to authentication.  It will take some work to test that it works correctly with all of the authentication modules.

The ability to set if a user must reset their password on the next login is not that important on the page for editing users.  It should be there, but it is more important that it be on the add users page.  So it can be set for the users first login without needing to go to another page.  It is when creating a user that this would normally be done.  Not when editing a user later.

Alex-Jordan:
I'll make that change about the Add Users page. I might try it on my own without Claude.

I don't object to waiting on any of these Claude-built changes until the next release. Or even rejecting them, if there's good reason. Since we are trying Claude out for a month, I will probably keep popping more onto the stack until the trial month runs out. If they are targeted to the 2.21 branch and don't make it to the release, I'll retarget at that time to develop.

Alex-Jordan:
> I might try it on my own without Claude.

Nope. I let Claude do it.

pstaabp:
Functionally, this looks good.  One minor quibble is the UI for the checkboxes.  Here's what this looks like:
image

the two buttons aren't labelled at all.  I do appreciate the tool tip that pops up right away, but an instructor would have to know where to look and mouse over it to find it. Even a label "Password (delete/reset)" might even help with this.

Alex-Jordan:
> UI for the checkboxes

Have you compared that to what we have now? There is just the one checkbox for deletion, and there is no header. IIRC, there is no column header just because we are already taking up so much horizontal space. I can add the column header, but I'd want to check that others won't object to how that will take up so much space.

It's too bad the word "Accommodation" is so long. That column takes up way more width than it needs to.

Alex-Jordan:
I was thinking you meant to have a separate column header, but I see that you just meant to extend the existing column header. So I did that:

Screenshot 2026-08-03 at 8 33 49 PM

I'm not sure what the best words would be. This particular string doesn't cause any wrapping on my screen, which I guess is good.

pstaabp (approved):
I think that looks better.

Alex-Jordan:
This now has the checkbox vertically centered as discussed in a meeting.

drgrice1 (approved):
Lets merge this.

PR #3087, opened 2026-07-26
Fix a cut and paste error in the reset password template.

drgrice1:
The label should be for the `otp_code` input, and not be for the `confirmPassword` input above which already has a label.


Alex-Jordan:
I'm going to just merge this.

PR #3088, opened 2026-07-26
move the test navigation table into a nav element

Alex-Jordan:
A medium level accessibility issue that Claude flagged is that the table here had role="navigation". This addresses that, by removing the aria labeling from the table and enclosing it all in a nav.

### DIFF
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -516,12 +516,12 @@
 			% end
 		% }
 		% my $jumpLinks = begin
-			<div class="table-responsive">
-				<table class="gwNavigation" role="navigation" aria-label="problem navigation">
+			<nav class="table-responsive" aria-label="<%= maketext('problem navigation') %>">
+				<table class="gwNavigation">
 					<%= content 'gw-navigation-cols' =%>
 					<%= content 'gw-navigation-table-rows' =%>
 				</table>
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Er, actually this was one of two "high" level accessibility issues it found. The other two are (1) addressed by #3087 and (2) the fact that PDF hardcopies are not tagged PDFs (which of course we can't fix). So even our "high" level accessibility issues per Claude are pretty small.

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -516,12 +516,14 @@
...
 		% }
 		% my $jumpLinks = begin
-			<div class="table-responsive">
-				<table class="gwNavigation" role="navigation" aria-label="problem navigation">
-					<%= content 'gw-navigation-cols' =%>
-					<%= content 'gw-navigation-table-rows' =%>
-				</table>
-			</div>
+			<nav aria-label="<%= maketext('problem navigation') %>">
+				<div class="table-responsive">
```
drgrice1: This could be condensed into just the `nav` element.  Move the `class="table-responsive"` onto the `nav`, and delete the `div`.

Alex-Jordan: Done!

drgrice1 (approved):

pstaabp (approved):

PR #3089, opened 2026-07-26
add role="alert" to various alerts

Alex-Jordan:
This is a "medium" level accessibility issue identified by Claude. Adds `role="alert"` to several alerts so they will be announced.


drgrice1 (approved):
Looks good.  I can confirm that these elements with the `alert` role added are now announced when they appear, and that they were not announce previously.  I tested with Gnome Orca.

pstaabp (approved):

PR #3090, opened 2026-07-26
buttons for reordering problems in a set details page

Alex-Jordan:
When in the Set Details page, there will now be buttons on each problem to move a problem up/down in sequence. The mouse click-and-drag tool is still there, but these buttons give a keyboard accessible way to rearrange the problems.

If the set is a JITAR set, there are also buttons for indenting/outdenting problems.

Buttons are disabled when appropriate (like you can't move problem #1 up in the list).

### DIFF
--- a/htdocs/js/ProblemSetDetail/problemsetdetail.js
+++ b/htdocs/js/ProblemSetDetail/problemsetdetail.js
@@ -42,10 +42,189 @@
 		}
 	};
 
+	const getSourceFilePath = (id) =>
+		document.getElementById(`problem.${id}.source_file_id`)?.value ||
+		document.getElementById(`problem_${id}_default_source_file`)?.value ||
+		'';
+
+	// Compute the "same source file" alerts for every problem based on the current order of the list.
+	const updateRepeatFileAlerts = () => {
+		if (!container) return;
+		const repeatFileText = container.dataset.repeatFileText;
+		if (!repeatFileText) return;
+
+		const shownYet = new Map();
+		for (const item of container.querySelectorAll('.psd_list_item')) {
+			const row = item.querySelector(':scope > .problem_detail_row');
+			const alertContainer = row?.querySelector('.pdr_repeat_file_alert');
+			if (!alertContainer) return;
+
+			alertContainer.innerHTML = '';
+
+			const problemID = item.id.replace('psd_list_item_', '');
+			const sourceFile = getSourceFilePath(problemID).replace(/^\//, '').replace(/\.\./g, '');
+			if (!sourceFile || sourceFile.startsWith('group:')) return;
+
+			if (shownYet.has(sourceFile)) {
+				const alert = document.createElement('div');
+				alert.className = 'alert alert-danger p-1 mb-2 fw-bold';
+				alert.textContent = repeatFileText.replace('[_1]', shownYet.get(sourceFile));
+				alertContainer.append(alert);
+			} else {
+				shownYet.set(sourceFile, row.querySelector('.pdr_problem_number')?.textContent ?? '');
+			}
+		}
+	};
+
 	const setProblemNumberFields = () => {
 		container.querySelectorAll('.psd_list_item .pdr_problem_number').forEach((num) => (num.textContent = ''));
 		recursiveRenumber(Sortable.get(container).toArray());
+		updateRepeatFileAlerts();
 	};
+
+	// Buttons to rearrange problems. Either up/down for rearranging order, or buttons for nestin/de-nesting JITAR
+	// problems. Buttons are disabled when the corresponding action isn't possible.
+	const updateReorderButtonStates = () => {
+		if (!container) return;
+		for (const list of [container, ...container.querySelectorAll('.sortable-branch')]) {
+			const items = list.querySelectorAll(':scope > .psd_list_item');
+			for (const [index, item] of items.entries()) {
+				const upButton = item.querySelector(':scope > .problem_detail_row .psd_move_up');
+				const downButton = item.querySelector(':scope > .problem_detail_row .psd_move_down');
+				if (upButton) upButton.disabled = index === 0;
+				if (downButton) downButton.disabled = index === items.length - 1;
+
+				// Nesting nests the item under its previous sibling, so it needs a previous sibling to nest under.
+				// Denesting moves the item up to its parent's list, so it isn't possible at the top level.
+				const nestButton = item.querySelector(':scope > .problem_detail_row .psd_nest');
+				const denestButton = item.querySelector(':scope > .problem_detail_row .psd_denest');
+				if (nestButton) nestButton.disabled = index === 0;
+				if (denestButton) denestButton.disabled = list === container;
+			}
+		}
+	};
+
+	// Recompute the nestDepth property (used by the drag-and-drop "put" rule) for a sub-list and all of its
+	// descendant sub-lists after the sub-list has moved to a new position in the tree.
+	const recomputeNestDepth = (list) => {
+		if (!list) return;
+		list.nestDepth = list.parentNode.closest('.sortable-branch').nestDepth + 1;
+		list.querySelectorAll(':scope > .psd_list_item > .sortable-branch').forEach(recomputeNestDepth);
+	};
+
+	// Show or hide each row's expand/collapse button depending on whether its sub-list currently has any children.
+	const updateCollapseButtonVisibility = () => {
+		if (!container) return;
+		for (const list of container.querySelectorAll('.sortable-branch')) {
+			if (list.firstElementChild) list.parentNode.collapseButton?.classList.remove('d-none');
+			else list.parentNode.collapseButton?.classList.add('d-none');
+		}
+	};
+
+	// Bookkeeping after any keyboard-driven reorder/nest/denest: update the hidden problem number/parent fields,
+	// re-enable/disable fields whose relevance depends on tree position, update this row's button states, and
+	// manage focus.
+	const finishReorder = (item, button, fallbackSelectors) => {
+		setProblemNumberFields();
+		disableFields();
+		updateReorderButtonStates();
+		updateCollapseButtonVisibility();
+
+		if (!button.disabled) {
+			button.focus();
+			return;
+		}
+		for (const selector of fallbackSelectors) {
+			const fallback = item.querySelector(selector);
+			if (fallback && !fallback.disabled) {
+				fallback.focus();
+				return;
+			}
+		}
+	};
+
+	const prefersReducedMotion = window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+
+	// Animate elements that just moved from where they used to be to where they ended up (the "FLIP" technique),
+	// so a move up/down reads as the two rows swapping places instead of an unexplained instant jump. Only the
+	// vertical position can change here, since these are stacked list items.
+	const animateReorder = (elementsAndFirstRects) => {
+		if (prefersReducedMotion) return;
+		for (const [el, firstRect] of elementsAndFirstRects) {
+			const deltaY = firstRect.top - el.getBoundingClientRect().top;
+			if (!deltaY) continue;
+			el.style.transition = 'none';
+			el.style.transform = `translateY(${deltaY}px)`;
+			el.getBoundingClientRect(); // Force a reflow so the starting position above is rendered before animating.
+			requestAnimationFrame(() => {
+				el.style.transition = 'transform 150ms ease-in-out';
+				el.style.transform = '';
+			});
+			el.addEventListener('transitionend', () => (el.style.transition = ''), { once: true });
+		}
+	};
+
+	const moveItem = (button, direction) => {
+		const item = button.closest('.psd_list_item');
+		const sibling = direction === 'up' ? item.previousElementSibling : item.nextElementSibling;
+		if (!sibling?.classList.contains('psd_list_item')) return;
+
+		const firstRects = [item, sibling].map((el) => [el, el.getBoundingClientRect()]);
+
+		if (direction === 'up') item.parentNode.insertBefore(item, sibling);
+		else item.parentNode.insertBefore(sibling, item);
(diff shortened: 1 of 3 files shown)


drgrice1:
Conceptually I think that this is okay.  However, I have been thinking about working on this, and this is not how I would implement this.  There are several problems with this approach.

One issue is that it involves numerous buttons, which there really isn't room for.  This is what I see on a narrow window.  I took it to the extreme to really emphasize the point in this screenshot, but it happens already even when the window is not nearly this narrow.
move-buttons

Another issue is that when you click on one of the up/down buttons for a problem the switch of problems is almost imperceivable.  In most sets the settings are usually all the same except for the source file.  So since the numbering is automatically updated (and that is correct to do), the only way you can see the change is by watching the source file.

Another issue that I am seeing has to do with the tooltips.  I think that this can be fixed with this pull request and its design though.  The issue is that if you click on one of the up/down buttons the tooltip is revealed and the problem is switched with the one above or below.  The keyboard focus goes with the switch to where the original problem now is, but the tooltip stays where it was.  This seems to confuse Bootstrap's javascript, and the tooltip never seems to get closed. In the extreme you can end up with numerous tooltips open at a time as in the following picture.
tooltips-not-closing
I have also noticed that one of those tooltips gets completely stuck, and there seems to be no way to actually get it to go away.  It seems to be the one on the first button used. I believe this can be fixed by taking care to close the tooltips in the javascript when the button is clicked.

What I envisioned, and had planned to implement when I got around to it, was an approach similar to what is now implemented for the PG drag and drop in the `dragndrop.js` file for the `DragNDrop.pm` module.  The existing move buttons would serve as the focus point.  The mouse could still be used to drag those around as it currently is, but the arrow keys could also be used to move the problems around when the move button is focused.  The up and down arrows clearly would move a problem up or down, and in a JITAR set the left and right arrows increase or decrease the nesting level.  This would all be aria announce as it is for the PG drag and drop.  One advantage of this approach over the implementation in this pull request is that if you use the down arrow repeatedly, you would be able to quickly move a single problem down several positions.  If you want to move a problem all the way from the top of the list to bottom, you could just hold the down arrow down, and keyboard repeat would take the problem down quickly.  Also, with the css transitions similar to those used in the PG drag and drop it would make the movement of the problem more clear.

Alex-Jordan:
I'm happy to defer to your recommendations here. I could withdraw this, or feed what you have described here to Claude. I probably won't directly intervene with coding changes myself on this one though.

Whatever things here (and with the other PRs I've made today) that we can complete, it will affect the Accessibility Guide I'm trying to complete before the release. It's getting down to a pretty short list of known issues (as far as webwork2 goes, that is).

drgrice1:
We can perhaps go with this for now, and then change to what I was thinking of later when I have time to work on that.

Alex-Jordan:
OK. I'll see what Claude can do for the smaller tweaks you identified.

Alex-Jordan:
The issues you found are at least mostly addressed now. The method I directed for the first issue (about overcrowding) is maybe not the best though.

drgrice1:
I find the "indent"/"outdent" terminology a bit ugly.  Also, in a JITAR set it isn't really indentation.  It is nesting, and in fact the wording everywhere else calls it nesting.

Alex-Jordan:
There is another problem with this. Say problem 1 and 2 use the same file. Problem 2 will have an alert "This problem uses the same source file as number 1.". Then if I use the buttons to swap #1 with #2, now the new #1 still says "This problem uses the same source file as number 1.". I'm working on that.

Alex-Jordan:
OK, the language (both user-facing and internal) is now about nest/denest instead of indent/outdent.

Also I moved the alerts for "This problem uses the same source file as number 1." from the template to the javascript, so it can be recalculated as things move. This is probably already an issue with the current mouse-driven rearrangement tool.

htdocs/js/ProblemSetDetail/problemsetdetail.js
```diff
@@ -42,11 +42,191 @@
...
+		'';
+
+	// Compute the "same source file" alerts for every problem based on the current order of the list.
+	const updateRepeatFileAlerts = () => {
+		if (!container) return;
+		const repeatFileText = container.dataset.repeatFileText;
+		if (!repeatFileText) return;
+
+		const shownYet = {};
+		container.querySelectorAll('.psd_list_item').forEach((item) => {
```
drgrice1: Please change this to

```suggestion
		for (const item of container.querySelectorAll('.psd_list_item')) {
```

and remove the `);` on line 76 below.

Make the same change for the `forEach` uses on line 89 and 117.

The similar `forEach` calls in this file that already existed should be updated at some point.  The `for (const ...)` construct is generally clearer for multi-line for loops.  The `forEach` method is largely a remnant of the old days when a `var` automatically hoisted.  With `let` and `const` that is no longer an issue.  The `forEach` usage is still nice for named functions such as on line 112 below, and brief one liners such as line 80 below (although that returns the result of `num.textContent = ''` which is really not correct, and eslint would complain about that).

htdocs/js/ProblemSetDetail/problemsetdetail.js
```diff
@@ -42,11 +42,191 @@
...
+		updateRepeatFileAlerts();
+	};
+
+	// Buttons to rearrange problems. Either up/down for rearranging order, or buttons for nestin/de-nesting JITAR
+	// problems. Buttons are disabled when the corresponding action isn't possible.
+	const updateReorderButtonStates = () => {
+		if (!container) return;
+		[container, ...container.querySelectorAll('.sortable-branch')].forEach((list) => {
+			const items = list.querySelectorAll(':scope > .psd_list_item');
+			items.forEach((item, index) => {
```
drgrice1: Please change this to

```suggestion
			for (const [index, item] of items.entries()) {
```

and delete the `);` on line 103.

I originally included this line in the previous comment, but realized this one might need more explanation because it works differently to iterate over the `index` and element.

htdocs/js/ProblemSetDetail/problemsetdetail.js
```diff
@@ -42,11 +42,191 @@
...
+			if (!sourceFile || sourceFile.startsWith('group:')) return;
+
+			if (Object.prototype.hasOwnProperty.call(shownYet, sourceFile)) {
+				const alert = document.createElement('div');
+				alert.className = 'alert alert-danger p-1 mb-2 fw-bold';
+				alert.textContent = repeatFileText.replace('[_1]', shownYet[sourceFile]);
+				alertContainer.append(alert);
+			} else {
+				shownYet[sourceFile] = row.querySelector('.pdr_problem_number')?.textContent ?? '';
+			}
```
drgrice1: Anytime that `prototype` hacking is being done, generally, you should question the code.  There are advanced techniques that require accessing the `prototype`, but `hasOwnProperty` usage by itself is not one of them.  It is highly questionable if the `hasOwnProperty` check is needed for a basic object that has no other properties or methods than the ones set locally.  There is always the possibility of a conflict with a builtin method or property, and so it is good to check.  Although, why would you use `Object.prototype.hasOwnProperty.call` for a locally defined object that can have no overrides that required the safety of that method.  Calling `showYet.hasOwnProperty` is perfectly safe in this instance.

Although, the real problem is that using a bare object is the wrong thing to do, and the reason for the need of such mysterious incantations as `hasOwnProperty` at all.  `showYet` should be a `Map`.  A `Map` was introduced into JavaScript for exactly this purpose and has been around in all browsers for more than 10 years.  So instead change line 56 above to `const shownYet = new Map();`, and change this to

```suggestion
			if (shownYet.has(sourceFile)) {
				const alert = document.createElement('div');
				alert.className = 'alert alert-danger p-1 mb-2 fw-bold';
				alert.textContent = container.dataset.repeatFileText.replace('[_1]', shownYet.get(sourceFile));
				alertContainer.append(alert);
			} else {
				shownYet.set(sourceFile, row.querySelector('.pdr_problem_number')?.textContent ?? '');
			}
```

drgrice1 (commented):
This is going to take some time to review and test properly.  But here are some things I see at this point.

Alex-Jordan:
OK, those code changes are in.

htdocs/js/ProblemSetDetail/problemsetdetail.js
Alex-Jordan: Er wait. I missed this somehow. Sorry, stand by and let me look it all over again.

Alex-Jordan: OK, it's good now.

htdocs/js/ProblemSetDetail/problemsetdetail.js
```diff
@@ -211,7 +392,9 @@
 
 	// Initialize tooltips.
 	document
-		.querySelectorAll('.psd_view,.psd_edit,.pdr_render,.pdr_grader,.pdr_handle > i')
+		.querySelectorAll(
+			'.psd_view,.psd_edit,.pdr_render,.pdr_grader,.pdr_handle > i,.psd_move_up,.psd_move_down,.psd_nest,.psd_denest'
```
drgrice1: I forgot to mention this before.  Please change this to

```suggestion
			'.psd_view,.psd_edit,.pdr_render,.pdr_grader,.pdr_handle > i,' +
				'.psd_move_up,.psd_move_down,.psd_nest,.psd_denest'
```

Alex-Jordan: Made this change.

drgrice1 (approved):
I think this looks good at this point.

pstaabp:
I think the responsiveness is a bit odd for JITAR sets.  Here's the large size:
image

and for a medium size: 
image

and then for a narrow screen:
image

The 2nd four-pack of buttons (old ones) start on the top row, then in the medium size, go to the 2nd row and then on the narrow screen, back to the top row.  

It may be less jarring to bring the source input box to the 2nd row on the medium screen.

somiaj (approved):
Looks good.

One thing I noticed (no fix needed) is when you copy/paste the name of a pg file into another problem, the 'duplicate warning' message doesn't appear until the order of the problems is changed. Further if you edit the filename of a duplicate file to a unique thing, the warning doesn't get removed until the problems are reordered. Would it be worth also updating the duplicate messages when the filename of a problem is edited?

Alex-Jordan:
> It may be less jarring to bring the source input box to the 2nd row on the medium screen.

This was fairly nontrivial. The top two lines of each problem's box use bootstrap grid, and then below that is a table with limited options for setting widths with bootstrap. But I like what I came up with better. See what you think.

@somiaj For what you noticed, I did not want to add even listeners for edits to the text field. But it could be done in the future.

Alex-Jordan:
I think that this should get another pass by @drgrice1, since some things changed structurally to respond to @pstaabp's observations.

drgrice1:
I think that the number of buttons that this adds, particularly for JITAR sets, is a challenge that I am not sure can be fully remedied.

This is what I see on a mobile device in a JITAR set:
jitar-narrow

Compare that to what is seen with the current release candidate branch:
jitar-narrow-rc

Both screenshots were taken after collapsing all details and expanding all nesting.

The buttons don't fit on the first line, and moving the details collapse button down makes it so that the source file label and input get meshed as well.

This was not great before as certain screen widths were already a problem, but now things are worse.  This is one of the reasons that I had mentioned the approach of just using the existing move button with arrow key controls.  It also seems odd to me to have a move button that is for mouse cursor use only, and then to have completely separate buttons for keyboard controls.

We can go with this for now, but I think this needs a complete redesign in the future.

htdocs/js/ProblemSetDetail/problemsetdetail.js
```diff
@@ -42,11 +42,192 @@
...
+
+		// Since focus stays on (or moves to another) reorder button after the click is handled below, the
+		// tooltip's own focus/hover triggers won't naturally hide it. Hide it explicitly so it doesn't get stuck.
+		bootstrap.Tooltip.getInstance(button)?.hide();
+
+		if (button.classList.contains('psd_move_up')) moveItem(button, 'up');
+		else if (button.classList.contains('psd_move_down')) moveItem(button, 'down');
+		else if (button.classList.contains('psd_nest')) nestItem(button);
+		else denestItem(button);
+	});
```
drgrice1: I just noticed this.  It does not make sense that this click handler is on the container instead of directly on the buttons.  That is not a good way to deal with this.  This should be

```suggestion
	for (const button of document.querySelectorAll('.psd_move_up, .psd_move_down, .psd_nest, .psd_denest')) {
		button.addEventListener('click', () => {
			// Since focus stays on (or moves to another) reorder button after the click is handled below, the
			// tooltip's own focus/hover triggers won't naturally hide it. Hide it explicitly so it doesn't get stuck.
			bootstrap.Tooltip.getInstance(button)?.hide();

			if (button.classList.contains('psd_move_up')) moveItem(button, 'up');
			else if (button.classList.contains('psd_move_down')) moveItem(button, 'down');
			else if (button.classList.contains('psd_nest')) nestItem(button);
			else denestItem(button);
		});
	}
```

It is really a hack to use a click event listener on the container and then find the button that was clicked in the listener, when really you want a click on the button.  Note that you don't even need to check that the button is disabled with this because disabled buttons do not receive the click event at all.

Alex-Jordan: OK, this change is made (not yet pushed). I'm going to see if there's a quick way to address the spacing issues you saw. (Yes, the whole thing should be refactored. In my opinion, also replacing the table in the lower part.)

Alex-Jordan:
Pushed, with that click event listener change.

Also with changes that keep the collapse button at the top row. However, at the tiny screen size (smaller than the sm breakpoint), there is not enough room for all the things you mentioned for a nested JITAR. So only below the sm breakpoint, I made the buttons narrower. Then it all fits. Even if it's JITAR with a wide number like 8.20.8.

drgrice1:
I think that is better, and good enough for now.

PR #3084, opened 2026-07-24
Prettier format template files

drgrice1:
The `@openwebwork/prettier-plugin-mojolicious-html-template` package is a `prettier` plugin that formats HTML with `prettier` and uses `perltidy` to format Perl content inside the Mojolicious template regions.

The `@openwebwork/prettier-plugin-mojolicious-html-template` project repository is housed at https://github.com/openwebwork/prettier-plugin-mojolicious-html-template.  I am giving ownership of the repository to webwork, because the project was developed with the usage of Claude code, and that usage was funded by the webwork project.

So this is my first Claude powered pull request.

Note that the first commit adds the framework for using `prettier` with the plugin on template files, the second applies `prettier` with the plugin to all template files, and the third does some manual clean up.  The manual clean up is mostly of strings in the help files because the new formatting ends up indenting them another level and makes the strings exceed the 120 character limit.

I am targeting this to develop.  There is certainly no urgency for adding this to the 2.21 release as it is a development only thing.

Note that the `bin/dev_scripts/pod-templates/main-index.mt` file is not formatted at this point.  The `.mt` extension is not handled by the plugin, and is in fact not even a valid extension for a Mojolicious template file.  That extension was introduced by @pstaabp at some point, and we just kind of went with it.  Really any Mojolicious template file should have the extension `.[output-format].ep` or `.[output-format].epl` where `[output-format]` is the type of file generated when the template is rendered.  Note that a file rendered by the `Mojo::Template` package directly or rendered by the `Mojolicious::Plugin::EPLRenderer.pm` package (which is just a thin wrapper around the `Mojo::Template` package) should use the `.epl` extension, and any rendered via the `Mojolicious::Plugion::EPRenderer.pm` package should use the `.ep` extension (this is how anything rendered by a `Mojolicious:Controller` is rendered).  So the `bin/dev_scripts/pod-templates/main-index.mt` should be ``bin/dev_scripts/pod-templates/main-index.html.epl`.

Also, the `courses.dist/modelCourse/templates/achievements/notifications/default.html.ep` is incorrectly named.  It should be `courses.dist/modelCourse/templates/achievements/notifications/default.txt.ep` (or really `courses.dist/modelCourse/templates/achievements/notifications/default.txt.epl` since it is rendered via `Mojo::Template` directly).  This file certainly should not be formatted since it is not HTML, and the plugin only handles HTML template formatting.  Text can't be formatted in any case (other than perhaps line wrapping).  Due to this

Fixing those file naming issues is left for later.

PR #3091, opened 2026-07-26
use aria-labels, not alt text, on icons

Alex-Jordan:
The icons we use for indicating problem set type are icons, not img elements. They can't have alt text. Instead, apply the img role and an aria-label.

### DIFF
--- a/templates/ContentGenerator/Instructor/ProblemSetList/set_list_row.html.ep
+++ b/templates/ContentGenerator/Instructor/ProblemSetList/set_list_row.html.ep
@@ -19,7 +19,7 @@
 	<td><%= check_box apply_date_sets => $set_id, class => 'form-check-input' =%></td>
 	<td dir="ltr">
 		% if ($iconClass) {
-			<i class="<%= $iconClass =%>" title="<%= $iconTitle =%>" alt="<%= $iconTitle =%>"></i>
+			<i class="<%= $iconClass =%>" title="<%= $iconTitle =%>" role="img" aria-label="<%= $iconTitle =%>"></i>
 		% }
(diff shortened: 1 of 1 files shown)


templates/ContentGenerator/Instructor/ProblemSetList/set_list_row.html.ep
```diff
@@ -34,7 +34,7 @@
 				<span class="set-label set-id-tooltip <%= $visibleClass %>" data-bs-toggle="tooltip"
 					data-bs-placement="right" data-bs-title="<%= $set->description %>">
 					% if ($iconClass) {
-						<i class="<%= $iconClass =%>" title="<%= $iconTitle =%>" alt="<%= $iconTitle =%>"></i>
+						<i class="<%= $iconClass =%>" title="<%= $iconTitle =%>" role="img" aria-label="<%= $iconTitle =%>"></i>
```
drgrice1, line 37: This is too long for the 120 character limit.  Please change this too

```suggestion
						<i class="<%= $iconClass =%>" title="<%= $iconTitle =%>" role="img"
							aria-label="<%= $iconTitle =%>"></i>
```

at least for now.  Although prettier will format this differently with #3084.

drgrice1 (approved):
This looks good other than a minor code formatting issue.

templates/ContentGenerator/Instructor/ProblemSetList/set_list_row.html.ep
Alex-Jordan: Done!

pstaabp (approved):

PR #3092, opened 2026-07-26
announce remaining time for test at certain times

Alex-Jordan:
Even moreso than other recent PRs, this one could wait until after 2.21 is released and I'll retarget to develop. (That's true for all of these recent PRs, but this one seems like extra testing is warranted.)

An accessibility flag raised by Claude is that the countdown timer for a test is not announced. But of course we don't want it constantly announcing every tick of the clock. This makes it announce the timer at certain times. Those times are just what I came up with off the top of my head: at 1 second, 2 seconds, 3 seconds, 4 seconds, 5 seconds, 15 seconds, 30 seconds, 1 minute, 5 minutes, 15 minutes,  30 minutes, and every multiple of 30 minutes. Different times could be used.his 

Also, this probably clashes with the existing thing where there is a dialog that pops up to warn you that you only have 45 seconds left, etc. So more thought should be put into something like this working well with that feature (which is a better feature for sighted users).

This should be tested with a screen reader, and I've never had much success trying that myself.

### DIFF
--- a/htdocs/js/GatewayQuiz/gateway.js
+++ b/htdocs/js/GatewayQuiz/gateway.js
@@ -10,6 +10,7 @@
 
 	// Gateway timer
 	const timerDiv = document.getElementById('gwTimer'); // The timer div element
+	const timerAnnounceDiv = document.getElementById('gwTimerAnnounce'); // Hidden live region for time announcements
 	let actuallySubmit = false; // This needs to be set to true to allow an actual submission.
 	// The 'Grade Test' submit button.
 	const submitAnswers =
@@ -58,8 +59,30 @@
 		bsToast.show();
 	};
 
+	// Only announce the remaining time via the hidden live region at these checkpoints (seconds remaining).
+	// Announcing every second would be overwhelming, so this only fires at the last few seconds (when the test is
+	// about to auto-submit) and at a handful of longer-interval checkpoints on the way down from 30 minutes.
+	const shouldAnnounceRemainingTime = (t) =>
+		(t >= 1 && t <= 5) ||
+		t === 15 ||
+		t === 30 ||
+		t === 60 ||
+		t === 300 ||
+		t === 900 ||
+		(t >= 1800 && t % 1800 === 0);
+
+	// Update the hidden live region with the current remaining time, either because it just hit one of the
+	// checkpoints above, or because force is true (used to always announce the initial remaining time on page load).
+	const announceRemainingTime = (remainingTime, force = false) => {
+		if (!timerAnnounceDiv || (!force && !shouldAnnounceRemainingTime(remainingTime))) return;
+		timerAnnounceDiv.textContent =
+			remainingTime >= 0
+				? `${remainingTimeString}${formatTime(remainingTime)}`
+				: `${remainingTimeString}00:00:00`;
+	};
+
 	// Update the timer
-	const updateTimer = () => {
+	const updateTimer = (forceAnnounce = false) => {
 		const dateNow = new Date();
 		const browserTime = Math.round(dateNow.getTime() / 1000);
 		const remainingTime = serverDueTime - browserTime + timeDelta;
@@ -72,6 +95,7 @@
 		}
 
 		if (!timerDiv.dataset.acting) {
+			announceRemainingTime(remainingTime, forceAnnounce);
 			// Check to see if we should put up a low time alert, or submit the test if
 			// the time is near the end of the grace period.
 			const alertStatus = sessionStorage.getItem('gatewayAlertStatus');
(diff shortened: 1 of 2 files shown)


htdocs/js/GatewayQuiz/gateway.js
```diff
@@ -72,6 +95,7 @@
 		}
 
 		if (!timerDiv.dataset.acting) {
+			announceRemainingTime(remainingTime, forceAnnounce);
```
drgrice1, line 98: Please add an empty line below this to offset it from the comment below it.

htdocs/js/GatewayQuiz/gateway.js
```diff
@@ -58,8 +59,30 @@
...
+	// Only announce the remaining time via the hidden live region at these checkpoints (seconds remaining).
+	// Announcing every second would be overwhelming, so this only fires at the last few seconds (when the test is
+	// about to auto-submit) and at a handful of longer-interval checkpoints on the way down from 30 minutes.
+	const shouldAnnounceRemainingTime = (t) =>
+		(t >= 1 && t <= 5) ||
+		t === 15 ||
+		t === 30 ||
+		t === 60 ||
+		t === 300 ||
+		t === 900 ||
```
drgrice1, line 71: I have frequently seen Claude miss many obvious things.  This is one of them (unless this is your code).  When I was not hearing any of the announcements, the first thing that I checked was that the text in the visually hidden div was actually changing, and I saw that it was not until it got down to 5 seconds or less.  The `===` comparisons are not going to work.  The `remainingTime` is not an integer.  This is because the server time that is received in the `updateTimeDelta` method is not an integer (it is the `submitTime` for the request which is set using `Time::HiRes`).  So the comparison `t === 15` is highly unlikely to actually be true.  So you will probably need to use something like `(t >= 15 && t <= 16)` instead.  Either that or I suppose we could round the entire `timeDelta` value on line 168 instead of only the client time obtained from `Date().getTime() / 1000`, and then the `===` comparisons will work.  I don't think that the sub second precision is needed.

Note that even when the `div` content is being set, the content is not announced (at least with Gnome Orca).

drgrice1 (commented):
Conceptually this is fine.

I am not hearing any of the announcements in testing with Gnome Orca.  Furthermore, when in browse mode and navigating around the page with the arrow keys, now both of the timer divs are stopped at and announced.  So you get repeated content.

With NVDA on Windows the announcements are made.  The browse mode issue is the same, and both timer divs are stopped at and announced.  Also, the last 5 seconds is not good.  It announces "remaining time 5 seconds", then "remaining time 4 seconds", and by the time that is done the test is out of time.  So the notification stating that pops up and it is read.  Then it continues to read "remaining time 3 seconds", "remaining time 2 seconds", and "remaining time 1 second", meanwhile the timer on the screen is at 0 through all of that.  I think that in the last five seconds the "remaining time" text will need to be removed so that it just reads "5 seconds", "4 seconds", ...  Although I haven't tested that to see how well that works.

drgrice1:
Actually, I just noticed the browse mode issue is even worse, because the first timer div has the actual time, and the visually hidden time div sits at 1 minute until the 30 second mark.

Also, I think that there is going to need to be though about this in combination with the pop up alerts, because NVDA announces those also (Gnome Orca does not).  So the last minute is filled with the time announcements, and it seems that would get really annoying if you are trying to work a test, and all of that is interrupting announcements of actual test content.

drgrice1:
Maybe the current notifications are enough, and just put aria hidden on the timer div?  I think the count down at the end even would interfere with any last second effort you may be trying to make so much that you could not do anything useful.

So I think this pull request will need more thought in general.

Alex-Jordan:
> Maybe the current notifications are enough, and just put aria hidden on the timer div?

Maybe. If you are not sighted, and you are curious how much time is left, is there a way to look? Of course, these announcements don't do that either. But if we have a solution that problem, then I think you might be right about just hiding the timer.

drgrice1:
That is true.  So adding `aria-hidden` to the time div is not the solution.  But I don't think that this approach of adding a second timer div that is also readable by an unsighted user and reads the wrong time most of the time is either.

drgrice1:
Would adding the `aria-live` and `aria-atomic` attributes momentarily to the existing div work to announce the remaining time at some intervals?  Perhaps in the end, only the 5 second mark should be announced instead of attempting a count down?

Alex-Jordan:
This one will probably be closed. I'll see if I can circle back later today, but I have to step away from WW development for a while now (today).

Alex-Jordan:
I'm going to close this and ask Claude for a suggestion as to how a user can effectively and accessibly ask "How much time do I have left?"

Alex-Jordan:
> Also, I think that there is going to need to be though about this in combination with the pop up alerts, because NVDA announces those also (Gnome Orca does not). 

Claude may have identified why this happens. I'm working it into changes for #3099.

Alex-Jordan:
For the record (assuming a correct diagnosis) it is not reliable to insert a live region into the DOM with its content already there. This could be why Gnome Orca wasn't catching the updates. Instead, first insert the live region. Then once it is in the DOM, insert the content.

PR #3093, opened 2026-07-26
make escape close navigation on narrow screen

Alex-Jordan:
On a narrow screen, when the sidebar navigation is open, this change makes it so that the escape key closes the navigation. 

From Claude:
>   This also replaced the ad hoc backdrop
>   creation/removal scattered across the click handler, resize handler, and
>   inline toggle logic with a single `openMobileDrawer`/`closeMobileDrawer` pair,
>   which incidentally fixes a latent bug where clicking the toggle button while
>   the drawer was already open left a stale extra backdrop in the DOM.

This was all a "medium" accessibility issue that Claude found.

### DIFF
--- a/htdocs/js/System/system.js
+++ b/htdocs/js/System/system.js
@@ -5,6 +5,49 @@
 		const threshold = 768;
 		let currentWidth = window.innerWidth;
 		const content = document.getElementById('content');
+		const masthead = document.querySelector('.masthead');
+		const toggleButton = document.getElementById('toggle-sidebar');
+		const skipLink = document.getElementById('skip-to-main-content');
+
+		// Make masthead, content, and skip to main content link inert while the drawer is open.
+		let sidebarBackdrop = null;
+
+		const handleEscapeKey = (e) => {
+			if (e.key !== 'Escape') return;
+			navigation_element.classList.remove('toggle-width');
+			content.classList.remove('toggle-width');
+			closeNarrowScreenDrawer();
+			toggleButton?.focus();
+		};
+
+		const closeNarrowScreenDrawer = () => {
+			if (!sidebarBackdrop) return;
+			sidebarBackdrop.remove();
+			sidebarBackdrop = null;
+			document.body.classList.remove('no-scroll');
+			content.inert = false;
+			if (masthead) masthead.inert = false;
+			if (skipLink) skipLink.inert = false;
+			document.removeEventListener('keydown', handleEscapeKey);
+		};
+
+		const openNarrowScreenDrawer = () => {
+			content.inert = true;
+			if (masthead) masthead.inert = true;
+			if (skipLink) skipLink.inert = true;
+			document.body.classList.add('no-scroll');
+			document.addEventListener('keydown', handleEscapeKey);
+
+			sidebarBackdrop = document.createElement('div');
+			sidebarBackdrop.classList.add('sidebar-backdrop');
+			document.body.append(sidebarBackdrop);
+			sidebarBackdrop.addEventListener('click', () => {
+				navigation_element.classList.remove('toggle-width');
+				content.classList.remove('toggle-width');
+				closeNarrowScreenDrawer();
+				toggleButton?.focus();
+			});
+		};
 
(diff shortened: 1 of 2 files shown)


htdocs/js/System/system.js
```diff
@@ -5,27 +5,52 @@
 		const threshold = 768;
 		let currentWidth = window.innerWidth;
 		const content = document.getElementById('content');
+		const masthead = document.querySelector('.masthead');
+		const toggleButton = document.getElementById('toggle-sidebar');
+
+		// On narrow viewports the sidebar behaves as a modal drawer: a backdrop covers the rest of the page (both
+		// the masthead and the main content are visually obscured by it, see system.scss z-indexes), so that
+		// background content is also made inert (unreachable by keyboard or screen reader) while the drawer is
+		// open, and Escape or clicking the backdrop closes the drawer again.
```
drgrice1: I recognize this sort of comment by now.  Why is Claude so bad a making readable commentary.  Please simplify this and make it more readable.  There is not need for the reference to `system.scss`, so delete that.  I keep telling Claude not to use semicolon separated sentences.  Claude thinks that every single sentence must have a semicolon, a hyphen, or a parenthetical, and often all of those.  These are what we call run on sentences, and clarity is not contained therein.  Notice this comment is one sentence and four lines long.

Look at the code in the new https://github.com/openwebwork/prettier-plugin-mojolicious-html-template repository.  That commentary is after I repeatedly told Claude to clean up and reduce comments.  I did get Claude to stop using the semicolons, hyphens, and parentheticals so much at least.  At one point the `src/mojo/printer.js` file had more comments than code in it and it was much longer, but it still has too many.  I plan to clean that up at some point.

htdocs/js/System/system.js
```diff
@@ -52,11 +77,20 @@
...
+
+		// Let Escape close the drawer on narrow viewports, same as clicking the backdrop.
+		document.addEventListener('keydown', (e) => {
+			if (e.key === 'Escape' && sidebarBackdrop) {
+				navigation_element.classList.remove('toggle-width');
+				content.classList.remove('toggle-width');
+				closeMobileDrawer();
+				toggleButton?.focus();
+			}
+		});
```
drgrice1: This is not good.  Having a `keydown` listener that is always active on the page even when not needed is not a good idea.  This listener should be attached when the sidebar is shown (in the `openMobileDrawer` function), and removed when it is hidden (in the `closeMobileDrawer` function).

htdocs/js/System/system.js
Alex-Jordan: OK, I changed this, and I'll work on better commentary.

htdocs/js/System/system.js
Alex-Jordan: I asked it to make that change. Javascript issues like this are not my forte.

htdocs/js/System/system.js
```diff
@@ -5,27 +5,59 @@
...
+
+		const handleEscapeKey = (e) => {
+			if (e.key !== 'Escape') return;
+			navigation_element.classList.remove('toggle-width');
+			content.classList.remove('toggle-width');
+			closeMobileDrawer();
+			toggleButton?.focus();
+		};
+
+		const closeMobileDrawer = () => {
```
drgrice1: I think that using names such as `closeMobileDrawer` and `openMobileDrawer` is misleading here.  This is not for mobile devices, but for anytime the browser window width is small.  This code is used on a desktop/laptop computer as well if the window is narrow.

drgrice1 (commented):
I just noticed something else that will probably need to be made `inert` while the sidebar is open on a narrow screen.  That is the "Skip to main content" link.  I noticed that if the sidebar is open, that link is still in the tab order, and if you focus it with the keyboard and hit enter it still attempts to focus the page content, and then things get weird with the tab order.

Alex-Jordan:
OK, I made some changes here that I think address your comments.

drgrice1 (approved):
This is ready I think.

somiaj (approved):

PR #3094, opened 2026-07-26
put aria-sort attributes on tables in Instructor Tools

Alex-Jordan:
The set list and user list tables now put `aria-sort="ascending"/"descending"` on the `<th>` of the primary sort column. And `aria-sort="other"` on any other (seondary or ternary) currently-sorted column's `<th>`, since `aria-sort` can only represent one column as strictly ascending/descending at a time.

### DIFF
--- a/lib/WeBWorK/ContentGenerator.pm
+++ b/lib/WeBWorK/ContentGenerator.pm
@@ -1261,6 +1261,33 @@ sub read_scoring_file ($c, $fileName) {
 	return parse_scoring_file($c->ce->{courseDirs}{scoring} . "/$fileName");
 }
 
+=item ariaSortAttr($c, $field)
+
+Returns the C<aria-sort> attribute (as C<< aria-sort="..." >>) to use on the
+C<< <th> >> for the sortable column C<$field> of a table, based on the
+C<primarySortField>/C<primarySortOrder>, C<secondarySortField>, and
+C<ternarySortField> values for the content generator C<$c>.  Returns an
+empty string if C<$field> is not currently a sort field for the table.
+
+Some content generators store these values directly on C<$c>, and others
+store them in the stash, so both locations are checked.
+
+=cut
+
+sub ariaSortAttr ($c, $field) {
+	my $primarySortField   = $c->{primarySortField}   // $c->stash->{primarySortField}   // '';
+	my $primarySortOrder   = $c->{primarySortOrder}   // $c->stash->{primarySortOrder}   // '';
+	my $secondarySortField = $c->{secondarySortField} // $c->stash->{secondarySortField} // '';
+	my $ternarySortField   = $c->{ternarySortField}   // $c->stash->{ternarySortField}   // '';
+
+	if ($primarySortField eq $field) {
+		return $primarySortOrder eq 'ASC' ? 'aria-sort="ascending"' : 'aria-sort="descending"';
+	} elsif ($secondarySortField eq $field || $ternarySortField eq $field) {
+		return 'aria-sort="other"';
+	}
+	return '';
+}
+
 =back
 
 =cut
--- a/templates/ContentGenerator/Instructor/ProblemSetList/set_list_table.html.ep
+++ b/templates/ContentGenerator/Instructor/ProblemSetList/set_list_table.html.ep
@@ -80,7 +80,8 @@
 						<% end =%>
 				</th>
 				% for (@$fieldNames) {
-					<th id="<%= $_ %>_header">
+					<th id="<%= $_ %>_header"
+						<%== !$c->{editMode} && $sortableFields->{$_} ? $c->ariaSortAttr($_) : '' %>>
 						% if (!$c->{editMode} && $sortableFields->{$_}) {
 							<div class="d-flex justify-content-between align-items-end gap-1">
 								<%= link_to $fieldHeaders{$_} => '#', class => 'sort-header',
--- a/templates/ContentGenerator/Instructor/UserList/user_list.html.ep
+++ b/templates/ContentGenerator/Instructor/UserList/user_list.html.ep
@@ -14,7 +14,7 @@
 					<i class="fa-solid fa-check-double" aria-hidden="true"></i>
 				<% end =%>
 		</th>
-		<th>
+		<th <%== $c->ariaSortAttr('user_id') %>>
 			<div class="d-flex justify-content-between align-items-end gap-1">
 				<%= link_to maketext('Login Name') => '#', class => 'sort-header',
 						data => { sort_field => 'user_id' } =%>
@@ -23,14 +23,14 @@
 		</th>
 		<th><%= maketext('Login Status') %></th>
 		<th><%= maketext('Assigned Sets') %></th>
-		<th>
+		<th <%== $c->ariaSortAttr('first_name') %>>
 			<div class="d-flex justify-content-between align-items-end gap-1">
 				<%= link_to maketext('First Name') => '#', class => 'sort-header',
 					data => { sort_field => 'first_name' } %>
 				<%= include 'ContentGenerator/Instructor/UserList/sort_button', field => 'first_name' =%>
 			</div>
 		</th>
-		<th>
+		<th <%== $c->ariaSortAttr('last_name') %>>
 			<div class="d-flex justify-content-between align-items-end gap-1">
 				<%= link_to maketext('Last Name') => '#', class => 'sort-header',
 					data => { sort_field => 'last_name' } =%>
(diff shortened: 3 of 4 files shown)


templates/ContentGenerator/Instructor/UserList/user_list.html.ep
```diff
@@ -1,4 +1,13 @@
 % unless ($c->{editMode}) {
+	% my $ariaSortAttr = sub {
+		% my $field = shift;
+		% if ($c->{primarySortField} eq $field) {
+			% return $c->{primarySortOrder} eq 'ASC' ? 'aria-sort="ascending"' : 'aria-sort="descending"';
+		% } elsif ($c->{secondarySortField} eq $field || $c->{ternarySortField} eq $field) {
+			% return 'aria-sort="other"';
+		% }
+		% return '';
+	% };
```
drgrice1: I think it would be cleaner to move the `$ariaSortAttr` function into the `UserList.pm` module, instead of defining it in the template.

Just define
```
sub ariaSortAttr ($c, $field) {
	if ($c->{primarySortField} eq $field) {
		return $c->{primarySortOrder} eq 'ASC' ? 'aria-sort="ascending"' : 'aria-sort="descending"';
	} elsif ($c->{secondarySortField} eq $field || $c->{ternarySortField} eq $field) {
		return 'aria-sort="other"';
	}
	return '';
}
```
at the end of `UserList.pm` and adapt the calls below to be of the form `$c->ariaSortAttr('student_id')`.

drgrice1 (approved):
This generally looks good.

templates/ContentGenerator/Instructor/UserList/user_list.html.ep
Alex-Jordan: Done, and force pushed.

somiaj (approved):
Works as expected.

I noticed that the "Job Manager" table doesn't include `aria-sort` in its table if you wanted to fix this there too.

Alex-Jordan:
Good catch, I pushed the same changes for the Job Manager.

lib/WeBWorK/ContentGenerator/Instructor/JobManager.pm
```diff
@@ -222,4 +222,13 @@ sub byDescStartedTime  { local ($b, $a) = ($a, $b); return byStartedTime(); }
...
 
+sub ariaSortAttr ($c, $field) {
+	if ($c->stash->{primarySortField} eq $field) {
+		return $c->stash->{primarySortOrder} eq 'ASC' ? 'aria-sort="ascending"' : 'aria-sort="descending"';
+	} elsif ($c->stash->{secondarySortField} eq $field || $c->stash->{ternarySortField} eq $field) {
+		return 'aria-sort="other"';
+	}
+	return '';
+}
+
```
somiaj: Should this be moved into a Utils file since this function is now in two locations?

Alex-Jordan: OK, I moved them into `lib/WeBWorK/Utils/Instructor.pm`.

templates/ContentGenerator/Instructor/ProblemSetList/set_list_table.html.ep
```diff
@@ -80,7 +80,15 @@
...
-					<th id="<%= $_ %>_header">
+					% my $ariaSort;
+					% if (!$c->{editMode} && $sortableFields->{$_}) {
+						% if ($c->{primarySortField} eq $_) {
+							% $ariaSort = $c->{primarySortOrder} eq 'ASC' ? 'ascending' : 'descending';
+						% } elsif ($c->{secondarySortField} eq $_) {
+							% $ariaSort = 'other';
+						% }
+					% }
+					<th id="<%= $_ %>_header" <%== $ariaSort ? qq{aria-sort="$ariaSort"} : '' %>>
```
somiaj: Maybe use the utility here too. I would include it at the top of this file:

```
% use WeBWorK::Utils::Instructor qw(ariaSortAttr);
%
```

Then change this code to:

```
<th id="<%= $_ %>_header" <%== ariaSortAttr($c, $_) %>>
```

templates/ContentGenerator/Instructor/UserList/user_list.html.ep
```diff
@@ -14,7 +14,7 @@
 					<i class="fa-solid fa-check-double" aria-hidden="true"></i>
 				<% end =%>
 		</th>
-		<th>
+		<th <%== $c->ariaSortAttr('user_id') %>>
```
somiaj, line 17: I am unsure what is better here, but attaching this to the `$c` object seems like what isn't done else where. Instead I would have done it like above. Include `ariaSortAttr` at the top of the template file:

```
% use WeBWorK::Utils::Instructor qw(ariaSortAttr);
```

Then call it as `<%== ariaSortAttr($c, 'user_id') %>`.

And remove the includes from the `.pm` files.

I don't know if one is preferred or not, but it isn't uncommon in template files to include and use the utilities directly vs attach it to the controller object.

templates/ContentGenerator/Instructor/ProblemSetList/set_list_table.html.ep
somiaj: Oh wait, I missed there is a `!$c->{editMode}` that you are also checking for, so what I have above won't work directly. Maybe:

```
<th id="<%= $_ %>_header" <%== !$c->{editMode} && $sortableFields->{$_} ? ariaSortAttr($c, $_) : '' %>>
```

Alex-Jordan: Good call, and it's done.

templates/ContentGenerator/Instructor/UserList/user_list.html.ep
Alex-Jordan: Everywhere else in this PR calls `ariaSortAttr` as a method on `$c`. Are you saying they should all be changed to `ariaSortAttr($c, ... )`?  I couldn't tell if your comment is only about this one instance.

somiaj: Yea, I twas about not adding the method to the controller object and just including it directly in the template file it is needed, and update all cases that do that.

somiaj: Note to me this is more what is done elsewhere, and I just thought there was no reason to add this method to the controller object. But I'm unsure if there is any reason that one way might be preferred over another, it just originally looked odd to me compared to what I've seen done elsewhere.

Alex-Jordan: Is it worth leaving it this way simply to not have to import the subroutine at the top of these files?

Alex-Jordan: BTW, something went wrong with my IT team's OS patches to my development server. I can't really test things right now, so I'm not trying any code changes at the moment.

somiaj: I would probably see if @drgrice1 can clarify if such a change is worth it, or if it just a matter of taste.

drgrice1: I would rather see the method imported in the file it is used. That makes it clear why the import is there. So I agree with @somiaj to import the `ariaSortAttr` in the template.

Unfortunately I only have my phone right now, and can't effectively look closely at the code to give a proper analysis though.

Alex-Jordan: I'll wait to hear the final word on this. My basic (untrained, not to be trusted) understanding of object oriented problem is that the point of having a method for `$c` is so you can just call it in this way. But I'm happy to change it.

drgrice1: The issue is that the `ariaSortAttr` is not technically a method of the controller object `$c`.  It is a method that is exported by the `WeBWorK::Utils::Instructor` package, and then imported into the controller module packages in which it is being used.  So calling it as `$c->ariaSortAttr` is misleading as to what it actually is.  It makes it look like it is a method of the controller module package.  Thus you would expect to find its code in that controller module package or one of its ancestor packages, and not some other package somewhere else entirely.

On that note I think that the right thing to do is not to put the `ariaSortAttr` method in the `WeBWorK::Utils::Instructor` package, but to put it instead in the `WeBWorK::ContentGenerator` package.  So it would then be something that any content generator package can use.  Then you can call it the way that you are calling it, and the method would not be exported or imported anywhere.  This method is directly pertaining to content generation in any case, so that is an appropriate place to put this method.

Alex-Jordan: As I look at this, moving the subroutine to `ContentGenerator.pm`, the end of this file has:

```
sub read_scoring_file {
    ...
}

=back

=cut

1;
```

Are those POD directives extraneous?

drgrice1: The `=back` ends the `=over` that begins on line 1112.  Note that there is another over that starts after that on line 1119, and that is ended on line 1149.  The `=cut` after that is needed to end the POD section that the `=back` starts.

Alex-Jordan: Nevermind about the POD. I think I get it.

Pushed now with the discussed move.

lib/WeBWorK/ContentGenerator.pm
```diff
@@ -1261,6 +1261,35 @@ sub read_scoring_file ($c, $fileName) {
...
+empty string if C<$field> is not currently a sort field for the table.
+
+Some content generators store these values directly on C<$c>, and others
+store them in the stash, so both locations are checked.
+
+=cut
+
+sub ariaSortAttr {
+       my ($c, $field) = @_;
+
```
somiaj: I believe this should be:

```suggestion
sub ariaSortAttr ($c, $field) {
```

somiaj (approved):
Looks good now.

@drgrice1 would you like to look this over one more time before merging?

drgrice1:
Looks good.  Merging.

PR #3096, opened 2026-07-26
make a few anchors with 'role="button"' respond to space

Alex-Jordan:


Alex-Jordan:
This was the last item Claude found for accessibility concerns with webwork2. The only unaddressed item after all these PRs is the one about PDF harcopies not being tagged.

htdocs/js/JobManager/jobmanager.js
```diff
@@ -1,4 +1,15 @@
...
+	// automatic activation on Enter, not Space like a real button would. Add that here.
+	document.querySelectorAll('.result-btn').forEach((btn) => {
+		btn.addEventListener('keydown', (e) => {
+			if (e.key === ' ') {
+				e.preventDefault();
+				btn.click();
+			}
+		});
+	});
+
```
drgrice1: Please move this code to the end.  Why is new code added to the beginning of the file when the order relative to the existing code does not matter?  In my opinion this is lower priority code than the existing code, and it should be at the end.

templates/ContentGenerator/CourseAdmin/manage_lti_course_map_form.html.ep
```diff
@@ -1,3 +1,7 @@
+% content_for js => begin
+	<%= javascript getAssetURL($ce, 'js/CourseAdmin/manage_lti_course_map.js'), defer => undef =%>
+% end
```
drgrice1: Did you forget to add the `htdocs/js/CourseAdmin/manage_lti_course_map.js` file to this pull request?

Alex-Jordan: Yeah, sorry. I was rushing to get to something else at home. It's there now.

htdocs/js/JobManager/jobmanager.js
Alex-Jordan: Moved. And the comment is toned down now too.

drgrice1 (commented):
As I am saying with all of these JavaScript changes, please change the `forEach` usage to `for of` loops.

Then this pull request is good.

Alex-Jordan:
OK, they are now "for of" loops.

drgrice1 (approved):
This looks good.

somiaj (approved):
Looks good.

I see the same JS code is added in multiple files. Would it be worth adding this to system.js and adding a class that can be used for other pages as well if there are other buttons in the future on other pages you want to respond to space?

drgrice1:
I had the same thought, but decided not to ask that at this point. It would be good though.

Alex-Jordan:
OK, now the duplicate code is in `system.js`.

Alex-Jordan:
@somiaj Does this still look good? With two approvals, would you be OK to merge? Just trying to reduce the PR list length.

somiaj:
I'll take a look in a bit. I haven't checked since you put the code into `system.js`.

templates/ContentGenerator/CourseAdmin/manage_lti_course_map_form.html.ep
```diff
@@ -1,3 +1,4 @@
+%
```
somiaj: This blank line with just `%` seems to be left from switching over to using `system.js`.

htdocs/js/System/system.js
```diff
@@ -87,12 +52,24 @@
...
-				closeNarrowScreenDrawer();
+				document.body.classList.remove('no-scroll');
+				document.querySelectorAll('.sidebar-backdrop').forEach((overlay) => overlay.remove());
 			}
 			currentWidth = window.innerWidth;
 		});
 	}
 
+	// Make elements with the role="button" attribute (that are not actual buttons, and so do not have this behavior
+	// built in) activate upon use of the spacebar.
```
somiaj: Comment says make elements with `role="button"`, but the code only cares about class and doesn't check role at all. I'm okay with it as is, but figured I'd mention it feels the comment doesn't state what the code is actually doing.

somiaj: I think it is "Make" that feels off to me. Again not an issue, just something that stuck out. I'll merge once you remove the blank line with `%` you left below.

templates/ContentGenerator/CourseAdmin/manage_lti_course_map_form.html.ep
drgrice1: Note that this `%` symbol will be removed by prettier and checked by the formatting workflow once #3084 is merged!  This is exactly the sort of thing that I wanted that for.

templates/ContentGenerator/Instructor/JobManager.html.ep
```diff
@@ -183,7 +183,7 @@
 							<div class="d-flex justify-content-between gap-1">
 								<%= maketext($jobs->{$jobID}{state}) =%>
 								% if (defined $jobs->{$jobID}{result}) {
-									<a role="button" class="result-btn" tabindex="0" data-bs-toggle="modal" href="#"
+									<a role="button" class="result-btn spacebar-activatable" tabindex="0" data-bs-toggle="modal" href="#"
```
drgrice1: I just noticed that this needs some clean up.  `tabindex="0"` is unnecessary with `href="#"`.  I am not sure why I put both of those on this `a` tag.  Also, the line is too long with the added class.  So please change this to

```suggestion
									<a role="button" class="result-btn spacebar-activatable" data-bs-toggle="modal"
										href="#" data-bs-target="<%= "#result-$jobID" %>">
```

htdocs/js/System/system.js
Alex-Jordan: OK, I changed the comment to `Make elements with role="button" (that have been given the class below) activate upon use of the spacebar.`

templates/ContentGenerator/CourseAdmin/manage_lti_course_map_form.html.ep
Alex-Jordan: The blank line is removed.

somiaj:
Everything addressed seems fixed, so merging.

PR #3097, opened 2026-07-27
show links to sets even when you are not assigned to those sets

Alex-Jordan:
Suppose you are an instructor and there is a set that is not assigned to you. If you go to the Sets Manager, and go to the Set Details page for that set, you won't see links to that set in the side navigation (under Sets Manager, Statistics, and Student Progress). This changes it so that you do. For example, it facilitates going to the Statistics page for that set even though it's not assigned to you as a user.

### DIFF
--- a/templates/ContentGenerator/Base/links.html.ep
+++ b/templates/ContentGenerator/Base/links.html.ep
@@ -125,14 +125,14 @@
 			% # Sets Manager
 			<li class="list-group-item nav-item"><%= $makelink->('instructor_set_list') %></li>
 			% # Editor link.  Only shown for non-versioned sets
-			% if (defined $setID && $setID !~ /,v\d+$/) {
+			% if (defined $globalSetID || defined $setID && $setID !~ /,v\d+$/) {
 				<li class="list-group-item nav-item">
 					<ul class="nav flex-column">
 						<li class="nav-item">
 							<%= $makelink->(
 								'instructor_set_detail',
 								text       => $prettySetID,
-								captures   => { setID => $setID },
+								captures   => { setID => $globalSetID // $setID },
 								link_attrs => { dir => 'ltr' }
 							); %>
 						</li>
@@ -162,7 +162,7 @@
 									$c->tag('span', dir => 'ltr', $prettySetID),
 									$prettyProblemID
 								)),
-								captures => { setID => $setID, problemID => $problemID },
+								captures => { setID => $setID // $globalSetID, problemID => $problemID },
 								target   => 'WW_Editor'
 							) %>
 						</li>
@@ -177,7 +177,7 @@
 								text => param('file_type') eq 'set_header'
 									? b(maketext('[_1]: Set Header',      $c->tag('span', dir => 'ltr', $prettySetID)))
 									: b(maketext('[_1]: Hardcopy Header', $c->tag('span', dir => 'ltr', $prettySetID))),
-								captures          => { setID => $setID, problemID => 0 },
+								captures          => { setID => $setID // $globalSetID, problemID => 0 },
 								systemlink_params => { file_type => param('file_type') },
 								target            => 'WW_Editor'
 							) %>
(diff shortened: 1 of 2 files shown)


drgrice1:
I think that a little more granularity is needed for which links specifically are shown.  If an instructor is not assigned to the set, then I agree with showing the sub links for `Sets Manager`, `Statistics` and `Student Progress`, but I don't think that the sub link for `Assignments` should be shown.  That just leads to a useless page with an error alert showing that the set is not valid for the user.

Either that or this should be taken further.  Currently as noted above if the instructor is not assigned to a set, you get the error alert that the set is not valid.  That should perhaps still be shown, but lets assume that the set is assigned to students.  If the student navigation were still shown when the instructor goes to this page, then this page still is useful.  Although, that would take some work to implement since it would probably involve needing to make changes to the `WeBWorK::checkSet` method.

Alex-Jordan:
I pushed changes so that the link is not shown under Assignments (if you are not actually assigned to the set).

drgrice1 (approved):
Looks good now.

pstaabp:
In SetDetails, I removed myself from the set, but a student has been assign it. 

I'm seeing:
```
Warning messages
Use of uninitialized value $setID in substitution (s///) at template ContentGenerator/Base/links.html.ep line 227.
Use of uninitialized value $setID in substitution (s///) at template ContentGenerator/Base/links.html.ep line 281.
```

at the bottom and this seems to make sense if on line 224, `$setID` is not defined for me but `$globalSetID` is, then I'll get the link now, but line 227 (actually 231) has a 
```
{ setID => $setID =~ s/,v\d+$//r }
```

Alex-Jordan:
Good catch. I caught something similar at a different place, but overlooked this one. Can you try it now? I changed the value of that key-value pair to be `$globalSetID // what-it-was-before`.

templates/ContentGenerator/Base/links.html.ep
pstaabp: I think this also causes the problem.  

This might fix it. 
```
captures => { setID => ($globalSetID // $setID ) =~ s/,v\d+$//r }
```

somiaj:
I am also getting a warning on the page to assign a set to a checkbox list of users:

```
Use of uninitialized value $setID in substitution (s///) at template ContentGenerator/Base/links.html.ep line 227.
```

Alex-Jordan:
Dang, I thought I thoroughly reviewed for this after Peter's finding. But this is another one of the same thing. I will fix it soon and push.

Alex-Jordan:
OK, that instance is addressed now in the same way as the previous one.

drgrice1:
I think all of the undefined warnings are resolved now.  My approval still stands.

somiaj:
The problem editor doesn't show the link to the set/problem when not assgined. Here is a set I'm assigned to:

image

That link doesn't show up when editing problems in a set I am not assigned to.

somiaj:
The link to the problem stats are also not shown. Here is what a set I'm assigned to shows while in the problem editor:

image

I am okay with this as is, and not asking to add those links, just pointing this out in case you wanted to also include them when not assigned to a set.

Alex-Jordan:
Yeah, I didn't do anything at the problem level, and IIRC it was intentional because there is some complication in the logic that I wanted to avoid dealing with. I'll take a look though.

somiaj:
@Alex-Jordan I am okay with it as is, just double checking it was intentional and you don't think there is a use case for that too.

Alex-Jordan:
OK, I believe those links show up and work now. I tested by vising a Set Detail page for a set that I was not assigned, and using the edit problem icon on a problem. The links appeared and I was able to follow all of them to their destination.

pstaabp (approved):
This looks good now.

somiaj:
The problem links are now shown, but are not selected and don't work in the problem editor (the stats links work). Here is what broke for me:

Sets Manager -> Click on "X/Y" to go to the set editor for a set you are not assigned to -> Click edit button for any problem in the set -> Problem editor opens, the links for the problem show up correctly in the menu but I first noticed that the "SetID: Problem X" link under "Problem Editor" is not highlighted as selected. Further when I actually click on that link I get "Page not found"

Looking closer the reason I get that is because the link ends with `CourseID/instructor/pgProblemEditor//4` -- note the `setID` is missing from the link (it should be there between `pgProblemEditor` and `4`.

Alex-Jordan:
@somiaj Man I don't know what's up with this one. I can't even blame Claude for all the mistakes.  But I think this might finally be ready.

drgrice1:
You can still blame Claude!  I would!

somiaj (approved):
I wasn't able to break it anymore.

Alex-Jordan:
I mean I could blame Claude. But this was a PR that was 100% me, 0% Claude. It was an issue I noticed while testing some other Claude-powered PR though, so I guess that could count.

PR #3099, opened 2026-07-27
button to announce remaining time on test

Alex-Jordan:
See revised description two posts below.

~~First, this adds a button under the Gateway Test countdown clock that says "Announce time remaining". If that button is pushed, a visually hidden div is updated with the current amount of remaining time. This div is an aria-live region, and a screen reader user will hear the time.~~

~~The button itself should not be visually hidden. So there is a bit of an oddity here for sighted users who see this button and might think it does nothing. Are there any suggestions for improving that?~~

~~Now we have a button for a user to ask "How much time do I have left?"   This means that the timer itself can be safely hidden to screen readers. So it is.~~

~~As usual, I haven't had success actually trying a screen reader to hear that this works. Related, I am not 100% sure that the time amount is formatted in a helpful way (e.g. "5 minutes", not "300 seconds"). But I am guessing that `formatTime` handles that well.~~

### DIFF
--- a/htdocs/js/GatewayQuiz/gateway.js
+++ b/htdocs/js/GatewayQuiz/gateway.js
@@ -30,6 +30,46 @@
 		return date.toISOString().substring(11, 19);
 	};
 
+	// Raise a toast to report the remaining time.
+	const showTimeToast = (text) => {
+		const toastContainer = document.createElement('div');
+		toastContainer.classList.add(
+			'gwAlert',
+			'toast-container',
+			'position-fixed',
+			'top-0',
+			'start-50',
+			'translate-middle-x',
+			'p-3'
+		);
+		toastContainer.innerHTML =
+			'<div class="toast bg-white" role="status" aria-live="polite" aria-atomic="true">' +
+			'<div class="toast-body alert alert-info mb-0 text-center d-flex align-items-center justify-content-center">' +
+			`<span class="me-2">${text}</span>` +
+			'<button type="button" class="btn-close" data-bs-dismiss="toast" aria-label="close"></button>' +
+			'</div>' +
+			'</div>';
+		document.body.prepend(toastContainer);
+		const bsToast = new bootstrap.Toast(toastContainer.firstElementChild, { delay: 4000 });
+		toastContainer.addEventListener('hidden.bs.toast', () => {
+			bsToast.dispose();
+			toastContainer.remove();
+		});
+
+		setTimeout(() => bsToast.show());
+	};
+
+	// Show the current amount of time remaining in a toast, in response to a click on the announce button.
+	const announceTimeRemaining = () => {
+		const remainingTime = serverDueTime - Math.round(new Date().getTime() / 1000) + timeDelta;
+		const text =
+			remainingTime >= 0
+				? `${remainingTimeString}${formatTime(remainingTime)}`
+				: `${remainingTimeString}00:00:00`;
+
+		showTimeToast(text);
+	};
+
 	const alertToast = (message, delay = 5000) => {
 		const toastContainer = document.createElement('div');
 		toastContainer.classList.add(
@@ -41,6 +81,7 @@
 			'translate-middle-x',
 			'p-3'
 		);
+		// Start the body empty and populate its content later, to have reliable capture of the update by screen readers.
 		toastContainer.innerHTML =
 			'<div class="toast bg-white" role="alert" aria-live="assertive" aria-atomic="true">' +
 			'<div class="toast-header">' +
@@ -55,7 +96,8 @@
 			bsToast.dispose();
 			toastContainer.remove();
 		});
-		bsToast.show();
+
+		setTimeout(() => bsToast.show());
 	};
 
 	// Update the timer
@@ -155,6 +197,8 @@
 
 		updateTimeDelta();
 		setInterval(updateTimeDelta, Math.min((parseInt(timerDiv.dataset.sessionTimeout) - 60) * 1000, 2147483646));
+
+		document.getElementById('gwAnnounceTimeBtn')?.addEventListener('click', announceTimeRemaining);
 
 		const remainingTime = serverDueTime - browserTime + timeDelta;
 
--- a/htdocs/js/GatewayQuiz/gateway.scss
+++ b/htdocs/js/GatewayQuiz/gateway.scss
@@ -1,17 +1,25 @@
 /* gateway styles */
 
-#gwTimer {
+#gwTimerContainer {
 	position: sticky;
 	width: 15em;
 	top: 2.75rem;
 	left: calc(100% - 15em - 0.75rem);
 	right: 0.75rem;
-	text-align: center;
 	z-index: 31;
+}
+
+#gwTimer {
+	text-align: center;
 }
 
 .gwAlert {
 	z-index: 32;
+
+	.toast-header + .toast-body {
+		border-top-left-radius: 0;
+		border-top-right-radius: 0;
+	}
 }
 
 #gwScoreSummary {
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -224,33 +224,42 @@
 	%
 	% # Print the timer if there is less than 24 hours left.
 	% if ($timeLeft < 86400) {
-		<%= tag('div',
-			id    => 'gwTimer',
-			class => 'alert alert-warning p-1',
-			data  => {
-				server_time     => int($submitTime),
-				server_due_time => $c->{set}->due_date,
-				session_timeout => $ce->{sessionTimeout},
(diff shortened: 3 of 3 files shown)


Alex-Jordan:
Claude suggests that clicking the button could trigger the toast that we already have. I'm gong to try that and update here if it seems good.

Alex-Jordan:
First, this adds a button under the Gateway Test countdown clock that says "Announce time remaining". If that button is pushed, there is a toast with the current amount of remaining time. This is an aria live region, and a screen reader user will hear the report.

Now we have a button for a user to ask "How much time do I have left?" This means that the timer itself can be safely hidden from screen readers. So it is.

This also potentially resolves an issue with the toasts that already exist (at 90 seconds, 45 seconds, and 0 seconds) where some screen readers were not announcing them. The issue might have been that the toast was inserted into the DOM fully constructed. Instead, inserting it empty, then updating it with content is the reliable way to get all (working) screen readers to read the update.

As usual, I haven't had success actually trying a screen reader to directly hear that this works.

Alex-Jordan:
And this makes one small styling improvement about rounded corners on the alert toasts that already existed.

somiaj:
How do other platforms deal with this? I think a button is good, but there are already lots of buttons on tests. Is there a way to make the button only visible to screen readers? Would that be something worth considering or would it be nice to have that button there for everyone?

drgrice1:
The toast that announces the time is not good in dark mode.
bad-in-dark-mode

Alex-Jordan:
> Is there a way to make the button only visible to screen readers? 

No, this is not possible without violating a different accessibility issue. Users still need to get focus onto the button to use it. If you let focus land on an invisible thing, it is confusing to a sighted user.

Alex-Jordan:
> The toast that announces the time is not good in dark mode. 

I added `alert alert-info` classes to it, and it appears to respect dark mode now.

htdocs/js/GatewayQuiz/gateway.js
```diff
@@ -30,6 +30,49 @@
...
+		const bsToast = new bootstrap.Toast(toastContainer.firstElementChild, { delay: 4000 });
+		toastContainer.addEventListener('hidden.bs.toast', () => {
+			bsToast.dispose();
+			toastContainer.remove();
+		});
+
+		requestAnimationFrame(() => {
+			toastContainer.querySelector('.toast-body span').textContent = text;
+			bsToast.show();
+		});
```
drgrice1: This doesn't fix Gnome Orca not announcing the content.  I wouldn't adjust code to try to get Gnome Orca to work.  It often is not quite fully reliable for testing.  That is why I try to test with NVDA.  Unfortunately, that requires booting to windows.

Alex-Jordan: Ah, well that's too bad. It still sounds right to me that we should not insert the fully constructed div when it's an aria-live region. The explanation from Claude sounds reasonable, that some screen readers might only respond to an existing live region getting an update and not to a live region being inserted.

Are you recommending changes to the code here?

drgrice1: I am not sure what is different with our code, but the live example at https://getbootstrap.com/docs/5.3/components/toasts/ is announced by Gnome Orca.

Alex-Jordan: Claude has a few theories about why Gnome Orca is struggling here. I'll wait to discuss them with you sometime in the future.

htdocs/js/GatewayQuiz/gateway.js
```diff
@@ -41,21 +84,26 @@
...
 		toastContainer.addEventListener('hidden.bs.toast', () => {
 			bsToast.dispose();
 			toastContainer.remove();
 		});
-		bsToast.show();
+
+		requestAnimationFrame(() => {
+			toastContainer.querySelector('.toast-body').innerHTML = message;
+			bsToast.show();
+		});
```
drgrice1: If you change this to

```suggestion
		setTimeout(() => bsToast.show());
```

and revert line 94 above, then these messages are announced by Gnome Orca.  If you leave line 94 as it is, and set the text in the timeout, it also works, but is unnecessary.  Using `requestAnimationFrame` does not work at all.

Note that for the `showTimeToast` method, the equivalent changes by themself do not make Gnome Orca announce it.  For that the `role` has to be changed to `alert` to make it be announced.  Gnome Orca does not seem to announce with `role="status"` at all.  I suspect this is an issue with Gnome Orca.  I haven't tested with NVDA, but I suspect that it will announce those.

Alex-Jordan: Do you think we should leave `showTimeToast` and the role on that element as is? We don't want to change its role, correct?

Alex-Jordan: Those changes at lines 94, 103-106 are now pushed.

drgrice1: I think the role of "status" is correct.

drgrice1: Bootstrap documents that role as a valid thing to use for this.

htdocs/js/GatewayQuiz/gateway.js
```diff
@@ -30,6 +30,49 @@
...
+			'</div>' +
+			'</div>';
+		document.body.prepend(toastContainer);
+		const bsToast = new bootstrap.Toast(toastContainer.firstElementChild, { delay: 4000 });
+		toastContainer.addEventListener('hidden.bs.toast', () => {
+			bsToast.dispose();
+			toastContainer.remove();
+		});
+
+		requestAnimationFrame(() => {
```
drgrice1: This should also be changed to `setTimeout`.  There is no point in using `requestAnimationFrame`, since it isn't helping anyway, and to be honest I thought that was an odd choice to begin with.  I would also recommend making this the same as the `alertToast` method, and just put the content in place from the start, and then this would just be `setTimeout(() => bsToast.show());`.

htdocs/js/GatewayQuiz/gateway.js
```diff
@@ -30,6 +30,49 @@
...
+			toastContainer.querySelector('.toast-body span').textContent = text;
+			bsToast.show();
+		});
+	};
+
+	// Show the current amount of time remaining in a toast, in response to a click on the announce button.
```
drgrice1, line 62: This comment is not accurate anymore.  What is really happening is that the fully constructed DOM content is inserted first, and then the toast is shown in the next tick via `setTimeout`.  That allows the DOM to update before the toast is shown, and emulates what happens if the toast were statically in the page to begin with as on the Bootstrap documentation page.

htdocs/js/GatewayQuiz/gateway.js
drgrice1: Hmm.  So I tested with NVDA now, and it is also not announcing any toasts with the role of `status`.

htdocs/js/GatewayQuiz/gateway.js
Alex-Jordan: Part of this was Claude's belief that some screen readers will not respond to an aria live region being added to the DOM, rather they only respond to an aria live region that is already in the DOM being updated. Do you think there is no validity to that?

drgrice1: I don't think that is valid, and testing does not support that.

htdocs/js/GatewayQuiz/gateway.js
Alex-Jordan: I removed the comment. I cant think of a concise way to say what you said here.

htdocs/js/GatewayQuiz/gateway.js
Alex-Jordan: I changed this one from `requestAnimationFrame` to `setTimeout` and then what I see with the toast is missing the message. I see this:

Screenshot 2026-08-03 at 8 45 24 PM

What am I missing?

drgrice1: You need to also add the `text` to the initial construction above.  So line 48 would become `` `<span class="me-2">${text}</span>` ``.

Alex-Jordan: Ah, sorry I looked over it so many times and missed that. OK, those changes are made now.

drgrice1 (approved):
I think this is good for now.

I think that there is further work to do on some things with the layout noted below, but perhaps they can wait.

There is an existing overlap issue with the positioning of the score summary that is shown when a test allows multiple submissions and the timer div (and now the "Announce remaining time" button).

I think that there is also an issue added with this with the placement of the original toast alerts and the new timer announce toast when both are present at the same time.  In general, all of these toasts really should share the same `toast-container`.  That is actually one of the purposes (in addition to placement) for the `toast-container`.  All toasts on the page should go into it, and then they all stack up nicely and don't overlap.

PR #3103, opened 2026-07-31
easy entry to a course from admin course

Alex-Jordan:
This is marked draft. Even though I'm targeting WeBWorK-2.21 right now, that is only so that the diff is clearly visible in GitHub. Later this will be re-targeted to develop, following the 2.21 release.

This (optional feature, off by default) makes it so that if you are using cookies for session management (not keys) and if you have a valid active session in the admin course, then that will smoothly grant you access into any other course. Some conditions are needed, of course:

- There is a user in that other course with the same username as your admin course user.
- That admin course user has to have high level permissions (`create_and_delete_courses`).
- The password hash is the same for both users (the one in the admin course and the one in the other course you intend to enter).

The main feature here (from my perspective) is that you can click links in the admin course and just be granted a session in the course you clicked on. This even works if that other course only allows users to enter through an LMS. You can also just click any link to any course, like say one in a student help email, and gain a session cookie. And you won't need to type a password.

All of this still requires 2FA for the course you are entering, assuming 2FA is enabled for that course, for a user of your level. That's actually something I would prefer not to have to do if I'm already authenticated in the admin course. But that could be changed later if this PR is not too objectionable.

Technical note: just because your user in the admin course and user in some other course have the same password, they would still have different password hashes if passwords were set independently. This really only works if the user in the other course were added to that other course as an admin user at the time the other course was initialized.

### DIFF
--- a/lib/WeBWorK/Authen.pm
+++ b/lib/WeBWorK/Authen.pm
@@ -40,6 +40,8 @@ use Scalar::Util         qw(weaken);
 use Mojo::Util           qw(b64_encode b64_decode);
 use Math::Random::Secure qw(irand);
 
+use WeBWorK::CourseEnvironment;
+use WeBWorK::DB;
 use WeBWorK::Debug       qw(debug);
 use WeBWorK::Utils       qw(x runtime_use utf8Crypt cryptPassword);
 use WeBWorK::Utils::Logs qw(writeCourseLog);
@@ -199,7 +201,9 @@ sub verify {
 	} else {
 		$self->write_log_entry("LOGIN FAILED $self->{log_error}") if defined $self->{log_error};
 		$self->maybe_kill_cookie;
-		$c->stash(authen_error => $self->{error}) if $self->{error} && $self->{error} =~ /\S/;
+		# A failed admin cross-course login attempt should not surface an error to the user.
+		$c->stash(authen_error => $self->{error})
+			if $self->{error} && $self->{error} =~ /\S/ && ($self->{credential_source} // '') ne 'admin_cross_course';
 	}
 
 	my $caliper_sensor = Caliper::Sensor->new($c->ce);
@@ -261,6 +265,8 @@ sub do_verify {
 
 	if (defined $self->{login_type} && $self->{login_type} eq 'guest') {
 		return $self->verify_practice_user;
+	} elsif (($self->{credential_source} // '') eq 'admin_cross_course') {
+		return $self->verify_admin_cross_course_user;
 	} else {
 		return $self->verify_normal_user;
 	}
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
This is now retargeted to `develop`. However at the moment, `develop` has not yet been updated with the WeBWorK-2.21/main branch.

drgrice1:
Develop is now up to date. So you can rebase onto it now.

Alex-Jordan:
Rebased and pushed.

lib/WeBWorK/Authen.pm
```diff
@@ -40,6 +40,8 @@ use Scalar::Util         qw(weaken);
 use Mojo::Util           qw(b64_encode b64_decode);
 use Math::Random::Secure qw(irand);
 
+use WeBWorK::CourseEnvironment;
+use WeBWorK::DB;
 use WeBWorK::Debug;
```
drgrice1: This needs to be

```suggestion
use WeBWorK::Debug       qw(debug);
```

and is the reason for the conflict.

lib/WeBWorK/Authen.pm
```diff
@@ -378,7 +384,89 @@ sub get_credentials {
...
+		$ce_admin = WeBWorK::CourseEnvironment->new({ courseName => $ce->{admin_course_id} });
+		$db_admin = WeBWorK::DB->new($ce_admin);
+	};
+	return 0 if $@ || !$db_admin || ($ce_admin->{session_management_via} // '') ne 'session_cookie';
+
+	# Decode the admin course's session cookie
+	my $sessions        = $c->app->sessions;
+	my $adminCookieName = 'WeBWorKCourseSession.' . $ce->{admin_course_id};
+	my $cookieMethod    = $sessions->encrypted ? 'encrypted_cookie' : 'signed_cookie';
+	my $rawValue        = $c->$cookieMethod($adminCookieName);
```
drgrice1: This should be

```suggestion
	my $sessions = $c->app->sessions;
	my $rawValue = $c->signed_cookie('WeBWorKCourseSession.' . $ce->{admin_course_id});
```

Webwork uses `signed_cookies` (the default).  So there is not need to check if the session cookie is encrypted or not.  It isn't.  Also the `encrypted` cookie feature was not added to Mojolicious until version 9.39, and we currently allow version 9.34 or newer (except a few bad versions) of Mojolicious.  Even if we do move to requiring newer versions of Mojolicious and use encrypted cookies, this check would not be necessary.  This would then use the `encrypted_cookie` method instead.  We know which one we are using, so no need to check.

lib/WeBWorK/Authen.pm
```diff
@@ -378,7 +384,89 @@ sub get_credentials {
...
+	return 0 unless $rawValue;
+	$rawValue =~ y/-/=/;
+	my $adminSession = eval { $sessions->deserialize->(b64_decode($rawValue)) };
+	return 0 unless $adminSession;
```
drgrice1: There is no need for the `eval` here.  Mojolicious does not use an `eval` in their code for this, so I see no reason that we should.  In fact, might as well use essentially  their code here and make this

```suggestion
	return 0 unless my $adminSession = $sessions->deserialize->(b64_decode($rawValue));
```

lib/WeBWorK/Authen.pm
```diff
@@ -378,7 +384,89 @@ sub get_credentials {
...
+
+	my $expiration = $adminSession->{expiration} // $sessions->default_expiration;
+	my $expires    = delete $adminSession->{expires};
+	return 0 if !$expires && $expiration || defined $expires && $expires <= time;
+
+	my ($adminUserID, $adminKey, $adminTimestamp) = @{$adminSession}{qw(user_id key timestamp)};
+	return 0 unless $adminUserID && $adminKey;
+
+	# Confirm the admin session is still valid
+	my $AdminKey = $db_admin->getKey($adminUserID);
```
drgrice1: Do not use Pascal case for variable names.  Use camel case.  I have been working on removing that sort of thing in the webwork2 code.  I recommend using the variable `$adminKeyRecord` here, since `$adminKey` is already used above, and that makes the difference clear here.  The actual key in the database is the `key` column, and this is the database record that contains that column.

lib/WeBWorK/Authen.pm
```diff
@@ -378,7 +384,89 @@ sub get_credentials {
...
+	return 0
+		unless defined $AdminKey
+		&& $AdminKey->key eq $adminKey
+		&& time <= ($adminTimestamp // $AdminKey->timestamp) + $ce_admin->{sessionTimeout};
+
+	return 0 unless _admin_course_has_create_delete_permission($ce_admin, $db_admin, $adminUserID);
+
+	my $AdminPassword = $db_admin->getPassword($adminUserID);
```
drgrice1: This also should not be Pascal case.  I recommend using `$adminPasswordRecord` for consistency with the `$adminKeyRecord` variable above.

lib/WeBWorK/Authen.pm
```diff
@@ -378,7 +384,89 @@ sub get_credentials {
...
+# full Mojolicious controller bound to the admin course, and here the controller is bound to the course being
+# entered instead.
+sub _admin_course_has_create_delete_permission {
+	my ($ce_admin, $db_admin, $user) = @_;
+
+	my $activity_role = $ce_admin->{permissionLevels}{create_and_delete_courses};
+	return 0 unless defined $activity_role && exists $ce_admin->{userRoles}{$activity_role};
+	my $role_permlevel = $ce_admin->{userRoles}{$activity_role};
+
+	my $PermissionLevel = $db_admin->getPermissionLevel($user);
```
drgrice1: I realize this variable was copied from `Authz.pm`, but this also should not be Pascal case.  The `Authz.pm` file needs a lot of clean up.  It is a mess.

lib/WeBWorK/Authen.pm
```diff
@@ -594,6 +682,37 @@ sub verify_normal_user {
...
+Compares this course's password hash for the user against the admin course's hash
+captured by C<try_admin_cross_course_credentials>.
+
+=cut
+
+sub verify_admin_cross_course_user {
+	my $self = shift;
+	my $c    = $self->{c};
+
+	my $user_id = $self->{user_id};
```
drgrice1: There is no need for a local `$user_id` variable.  Just use `$self->{user_id}` in the two places the `$user_id` variable is used in this method.

lib/WeBWorK/Authen.pm
```diff
@@ -594,6 +682,37 @@ sub verify_normal_user {
...
+
+	my $coursePassword = $c->db->getPassword($user_id);
+	unless (defined $coursePassword
+		&& $coursePassword->password =~ /\S/
+		&& $coursePassword->password eq $self->{admin_cross_course_password})
+	{
+		$self->{log_error} = 'admin cross-course login: no matching password for this user in this course';
+		$self->{error}     = $c->maketext(GENERIC_ERROR_MESSAGE);
```
drgrice1: Delete this line.  The only case in which this method is called is if the `credential_source` is `'admin_cross_course'`, and in that case `$self->{error}` is also ignored (see line 206).  So there is no point in setting it.

drgrice1 (commented):
These are just code suggestions.  I am still working on analyzing the effect of this on other authentication modules.

I have created a pull request to this branch with these suggested code changes.

lib/WeBWorK/Authen.pm
```diff
@@ -378,7 +384,89 @@ sub get_credentials {
...
+	my $cookieMethod    = $sessions->encrypted ? 'encrypted_cookie' : 'signed_cookie';
+	my $rawValue        = $c->$cookieMethod($adminCookieName);
+	return 0 unless $rawValue;
+	$rawValue =~ y/-/=/;
+	my $adminSession = eval { $sessions->deserialize->(b64_decode($rawValue)) };
+	return 0 unless $adminSession;
+
+	my $expiration = $adminSession->{expiration} // $sessions->default_expiration;
+	my $expires    = delete $adminSession->{expires};
+	return 0 if !$expires && $expiration || defined $expires && $expires <= time;
```
drgrice1: I forgot to mention this.  I included it in the pull request.

Most of this code seems to be copied from the `load` method of the `Mojolicious::Sessions` package.  They `delete` the `expires` key from the session, but this shouldn't.  This could be cleaned up and changed to

```suggestion
	return 0
		if !$adminSession->{expires} && ($adminSession->{expiration} // $sessions->default_expiration)
		|| defined $adminSession->{expires} && $adminSession->{expires} <= time;
```

drgrice1:
So as far as the other authentication modules go, here is my assessment.

- This is compatible with LDAP authentication.  That is assuming the admin course user has a password set in the admin course and that is copied to other courses.
- This is not compatible with Saml2 authentication, but probably could be made to work with it.
- This is not compatible with Shibboleth authentication.  It might be possible to make it compatible with that module, but this would be challenging to say the least.
- I didn't really test it, but this will work fine with the LTI authentication modules since they completely fall back to the basic authentication module in the cases of interest anyway.
- I don't have a way to test CAS authentication, but that probably doesn't matter.  It is broken anyway.

So probably for now, you could just add comments in the documentation stating that this does not work with Saml2 and Shibbolith.  Perhaps later, if there is demand, this could be extended to work for those.

Alex-Jordan:
Thanks for your review of this. I merged your PR and updated the comments in defaults.config and localOverrides.conf.dist. I wrote that it won't work for Sam2, Shibboleth, or CAS. But maybe I should not mention CAS?

lib/WeBWorK/Authen.pm
```diff
@@ -378,7 +384,86 @@ sub get_credentials {
...
+# entered instead.
+sub _admin_course_has_create_delete_permission {
+	my ($ce_admin, $db_admin, $user) = @_;
+
+	my $activity_role = $ce_admin->{permissionLevels}{create_and_delete_courses};
+	return 0 unless defined $activity_role && exists $ce_admin->{userRoles}{$activity_role};
+	my $role_permlevel = $ce_admin->{userRoles}{$activity_role};
+
+	my $permissionLevel = $db_admin->getPermissionLevel($user);
+	return 0 unless defined $permissionLevel && defined $permissionLevel->permission;
```
drgrice1: This should be 
```suggestion
	return 0
		unless defined $permissionLevel && defined $permissionLevel->permission && $permissionLevel->permission ne '';
```
Although, if you want to you could remove the `defined $permissionLevel->permission` since that literally can never happen.

Alex-Jordan: OK, that change is pushed. And I removed the `defined $permissionLevel->permission` bit.

drgrice1 (approved):
I think this looks good now.

somiaj (approved):

PR #3110, opened 2026-08-05
Cache course archives in the admin course archives directory. (hotfix of #3109)

drgrice1:
Anytime the `listArchivedCourses` method of the
`WeBWorK::Utils::CourseManagement` package is called a cache of the course archive files in the admin course archives directory is created (if it does not exist) or updated (if new archive files are found, archive files are removed, or archive files are modified).

Also, anytime a course archive is created and the archive file is stored in the archives directory of the admin course, the cache is updated. Note that when an archive file is added to the cache this way it is actually much faster than when it is added to the cache via the `listArchivedCourses` method since the tar archive does not need to be inspected at all (it was just created, so the course id is already known).

Note that the `listArchivedCourses` approach for updating the cache for deleted archive files is really fast, so there is no need to handle that elsewhere.  Unfortunately, if archive files are added or modified in other ways than creating a course archive via the `archiveCourse` method, there is no way to detect that.  So the slow `listArchivedCourses` cache update is the only recourse. If there are only a few added archive files this is still relatively fast though.

The archive cache is stored in the file `archives/archive-cache.json` in the admin course directory.

The `listArchivedCourses` method is called when the "Course Listings" or "Unarchive Course" pages in the admin course are loaded. So if you have a lot of course archives in the admin course archives directory that are not yet cached, the load time will still be slow. However, once those archives are in the cache, the load time will be pretty much as fast as before.

A `getHumanReadableFileSize` method was added to the `WeBWorK::Utils::Files` package since that is used both in the `listArchivedCourses` method and in the `archiveCourse` method, and might be useful elsewhere.

### DIFF
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -15,6 +15,7 @@ use DBI;
 use String::ShellQuote;
 use UUID::Tiny            qw(create_uuid_as_string);
 use Mojo::File            qw(path);
+use Mojo::JSON            qw(decode_json encode_json);
 use File::Copy::Recursive qw(dircopy);
 use File::Spec;
 use Archive::Tar;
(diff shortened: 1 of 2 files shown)


Alex-Jordan (approved):
This looks good. And it is working on my production server.

somiaj (approved):
Works and drastically speeds up viewing the course list page after the cache was built.

Should we add to the release notes that the first time logging into the admin course could take a bit while the cache is built?

PR #3111, opened 2026-08-05
Net::SAML2 0.86 security fix now requires cacert

ionparticle:
Net::SAML2 0.86 is a security release that fixes CVE-2026-18092, CVE-2026-18089.

CVE-2026-18089 fix requires adding the (previously optional) cacert param to the Net::SAML2::Protocol::Assertion->new_from_xml call. This ensures that if there are embedded certificates in the assertions, they are verified to be trusted by the IdP before being used.


drgrice1 (approved):
I can confirm that this is needed with the latest versions of `Net::SAML2`.

drgrice1:
I created a pull request of this to go into develop.

Alex-Jordan (approved):

PR #3115, opened 2026-08-05
Show both the course id and the archive file for duplicates when listing course archives in the admin course. (hotfix of #3114)

drgrice1:
The previous setup did not take into account what would happen if multiple archive files contained the same course id.  So instead of the `listArchivedCourse` method returning a hash of the form
    
```perl
{
    myTestCourse => {
        filename => 'myTestCourse.tar.gz',
        size     => '605 KB'
    }
}
```

it now returns a hash of the form

```perl
{
    'myTestCourse.tar.gz' => {
        courseID     => 'myTestCourse',
        size         => '605 KB',
        lastModified => 1778667472
    }
}
```

Note that the `lastModified` key is only included because that is what is saved in the cache file, and that is just directly returned if the archive file has not been modified. Although, at this point it is not used elsewhere.

In addition when duplicate courseIDs are seen each of those in the hash will contain the `duplicateCourseID` key.

Then both the courseID and the filename are then displayed for those archives that are duplicate courseIDs.  The format displayed is

```
notDuplicatedCourseID (2 GB)
myTestCourse (myTestCourse.tar.gz, 605 KB)
myTestCourse (myTestCourseAlt.tar.gz, 1.6 MB)
```

### DIFF
--- a/lib/WeBWorK/Utils/CourseManagement.pm
+++ b/lib/WeBWorK/Utils/CourseManagement.pm
@@ -123,6 +123,9 @@ containing the C<courseID>, file C<size>, and C<lastModified> time. For example,
         }
     }
 
+Note that archive files that contain the same C<courseID> will also have the
+C<duplicateCourseID> key set to 1.
+
 =cut
 
 sub listArchivedCourses {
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
Closing in favor of #3117.

PR #3116, opened 2026-08-06
Show both course id and the archive file when they differ when listing course archives in the admin course.

drgrice1:
This is an alternate to #3114.


somiaj (approved):

somiaj:
Would adding some help to describe when the filename is listed be helpful for those who may not understand why sometimes the filename is there and others it isn't?

drgrice1:
I don't know. I think that most of the time the file names will not be shown at all. Those that do should be able to figure it out.

drgrice1:
Also, there isn't really a good way to add help here.  There is not a help for the course listings page, and I wouldn't want the popover help here.

Alex-Jordan (approved):

Issue #3102, opened 2026-07-30
Custom logout for SAML2


mackensen:
We're using WeBWorK 2.20 with the SAML2 authentication method. Our identity provider does not include a logout URL in its metadata. It would be nice to have a way to configure a custom logout URL with SAML2.

ionparticle:
Usually when the metadata doesn't advertise support for single logout, it's because the IdP doesn't want to support it. Single logout doesn't just sign users out of Webwork, it signs users out of every other application they were authenticated into, and that's not a use case everyone wants. Are you sure that your IdP will properly respond to a single logout request?

mackensen:
> Usually when the metadata doesn't advertise support for single logout, it's because the IdP doesn't want to support it. Single logout doesn't just sign users out of Webwork, it signs users out of every other application they were authenticated into, and that's not a use case everyone wants. Are you sure that your IdP will properly respond to a single logout request?

I've been testing a solution similar to what @drgrice1 implemented in [d6888e4](https://github.com/drgrice1/webwork2/commit/d6888e44b94b0414ad2105e97e994dc55af1bf55). In our environment we have CAS handling the actual login/logout (we don't use the CAS module because it's effectively broken/abandoned). Redirecting the user to the CAS logout after local logout seems to accomplish what we're looking for.

drgrice1:
This was added in #3120 and will be in the next release.  Unfortunately that is pretty much a year away though.  So those wanting to use this now will need to patch #3120 in.

I am closing this issue in any case, since it is done.

PR #3120, opened 2026-08-06
Add a configuration option to provide a custom logout URL for SAML2 authentication.

drgrice1:
If a SAML2 identity provider does not provide a logout URL in its metadata, and yet one is available for the identity provider, then the `$saml2{sp}{logout_url}` option can be set to that logout URL, and it will be used instead of the logout URL from the metadata.

This was requested in issue #3102.


somiaj (approved):
Code looks good. Unable to test.

drgrice1:
To be honest, I can't really test it either. My local saml2 test instance provides the logout URL, and so does the provided docker test instance in the repository.

Alex-Jordan (approved):

PR #3123, opened 2026-08-10
Extend prettier formatting to all files prettier can format.

drgrice1:
This simplifies the `package.json` script because it can just format every file type that prettier can handle.  Anything ignored by git will also be ignored by prettier.  If there are additional files in the repository that need to be skipped, they can be added to the `.prettierignore` file.  At this point the `htdocs/library-tree.json` file is the only file that is skipped.

Note that the `bin/dev_scripts/pod-templates/main-index.mt` file was renamed to `bin/dev_scripts/pod-templates/main-index.html.epl` so that it will be formatted by prettier, and because the `.mt` extension is not a valid Mojolicious template extension.  The valid Mojolicious template extensions are of the format `outputformat.ep` or `outputformat.epl` where output format is the type of file that is generated by the template, and `.ep` is used for templates rendered via the `Mojolicious::Plugin::EPRenderer` module and `.epl` is used for templates rendered via the `Mojolicious::Plugin::EPLRenderer` module or rendered directly with the `Mojo::Template` module.

Also, achievement notification templates should have the extension `.txt.epl` instead of `.html.ep`.  Since these generate only plain text for an email, and not html, the `.html` extension for the output format is incorrect.  Also, since these are rendered directly with the `Mojo::Template` module, the `.epl` extension is the correct one for that.  Note that this is backward compatible.  If an achievement is set to use a template with the `.html.ep` then that will still work. However, the default template distributed with webwork2 is the `default.txt.epl` file. When a template is saved in the achievement notification editor it will save it with the `.txt.epl` extension, so future templates that are created will have the correct extension.

Most files that prettier can format were already being formatted. The few that were not have now been formatted.  These will also be checked by the formatting workflow.


somiaj:
Unsure if it matters, but `bin/dev_scripts/pod-templates/main-index.html.epl` is shown as a new file and the old file as being deleted, not that the file moved.

drgrice1:
Unfortunately, there is nothing that I could do to stop that with one commit.  I guess there are to many changes and git does not recognize it as a rename.  I did use `git mv` to move the file initially.  But then when `prettier` formatted it, it dropped the `rename` status.  I suppose if I rework this into two commits with the first commit just moving the file, and the second reformatting, then it would probably work.

somiaj:
Probably not a big deal to track the history. I was hoping `git mv` would notice it as a move/edit vs delete old / create new. I guess I was under the impression that could be done in a single commit. Nice to know that it cannot for the future if we want to better track history across file moves.

drgrice1:
That usually does work, but only if the changes are not significant enough.  Due to the new indentation in the file, almost all lines were changed, and I think that even though they were only whitespace changes, that was too much for git to still realize it as a rename.

somiaj:
It also just might be github ui vs git. Maybe the history was there, but github is still showing delete/create in the files changed summary.

somiaj:
If you look at the individual commits you see the move, then the edit in the two commits. But when they are seen together github ui doesn't clearly show it was a move/edit.

drgrice1:
I reworked this into two commits.  The first is basically the formatting setup, and renames the two files.  The second is only formatting with prettier.  Note that first commit does show the rename for both files.  The second just shows a modification.  However, here in GitHub when reviewing files, it does not show the rename for the `bin/dev_scripts/pod-templates/main-index.html.epl`. but the git history still shows the rename.

somiaj:
When running `npm run prettier-check` from `htdocs/`, I noticed that files in `.git/` were being checked. Should those be ignored too?

drgrice1:
I saw that it is looking there. Prettier is supposed to ignore that directory by default. I don't think it is actually formatting anything there.

somiaj:
It seems to take time (~15 seconds here) going though listing a bunch of files. It is probably ignoring them (it goes fast), but it might speed it up a tiny bit to explicitly ignore it so it doesn't waste time searching through the directory was my thought.

somiaj (approved):

drgrice1:
I went ahead and added `.git` to the `.prettierignore` file.  The issue is that the `package.json` file and `node_modules` are not at the root level in the repository.  That causes some odd behavior, and prettier does not ignore the `.git` directory.

somiaj:
Thanks, I don't see prettier trying to search through or check files in `.git`.

Alex-Jordan (approved):

PR #3124, opened 2026-08-10
Make Caliper an optionally loaded plugin that interfaces with webwork2 via hooks.

drgrice1:
This means that the Caliper code does not need to be directly in the primary webwork2 code, and those that don't use Caliper don't have to even have it checking to see if it is enabled. Just leave the `Mojolicious::WeBWorK::Plugin::Caliper` plugin commented out in `webwork2.mojolicious.yml`. If you want to allow the usage of Caliper on your server, then uncomment that plugin. Caliper can be configured per course in `course.conf` files if the plugin is enabled.

This `hook` approach is also extendable. Other plugins, even those not directly in the webwork2 repository could utilize these hooks.  The hooks added in this pull request are when user logs in or out, when an answer is submitted in a regular assignment, and when a page change, preview, or submission occurs in a test.  Additional hooks could also be added for further extendability with plugins in the future.

The `$caliper{base_url}` option no longer exists.  Those using Caliper just need to make sure that the `$server_root_url` and `$webwork_url` are set correctly in `site.conf`.  The point is that anyone using webwork2 should do that anyway, so the `$caliper{base_url}` setting was redundant.

The Caliper packages are heavily updated.  The packages all use signatures.  Instead of passing both the course environment and database handle everywhere, just pass the controller which has both of those. In addition, doing so gives access to do things better. For example, the environment variable usage (`HTTP_X_FORWARDED_FOR`, `REMOTE_ADDR`, `HTTP_CLIENT_IP`, `HTTP_USER_AGENT`, and `HTTP_HOST` in `Caliper/Entity.pm` and `HTTP_REFERER` in `Caliper/Event.pm`) was clearly broken and not working since the switch to Mojolicious. Since the controller is now available in those places the things those environment variables used to provide with modperl and apache can now be obtained from the controller.

`Mojo::UserAgent` is used instead of `HTTP::Async`. This is the only place that the outdated and unmaintained `HTTP::Async` package is used. So drop that dependency.

The UUIDs used now use `create_uuid_as_string` from the `UUID::Tiny` package, instead of the `Data::UUID` package and essentially `Data::UUID->new->create_str`. This is done for two reasons.  First, this is the only use of this dependency.  So that is another dependency not needed by webwork2.  Second, the Caliper specification recommends using version 4 UUIDs and `Data::UUID` does not generate version 4 UUIDs, but `UUID::Tiny` does.

The problem source is no longer sent in Caliper events. That really should never have been done.  That is a large amount of data and is not needed. This means that even those that do not use Caliper have to endure the additional server load required to transmit that back from the rendering process to the main process.  It also makes the Caliper events much larger in size than they should be. The source file is still sent and that really should be all that is needed.  The source can be looked up in the file.

Other than not sending the problem source, everything else sent by the Caliper implementation is the same. So the implementation still meets the 1EdTech Caliper specification as much as it did before (the problem source is certainly not something the specification insists on or is even considered in the specification). See https://www.imsglobal.org/spec/caliper/v1p2.

I have attached a Perl Mojolicious app that can be used to test the webwork2 Caliper implementation.  It will work with both this pull request and the current implementation in the develop or main branches.  To use it extract the zip archive, and in the directory created execute `morbo script/CaliperDashboard -l http://*:5000`.  You can change the port if needed.  You will also need to set the Caliper settings in `localOverrides.conf` appropriately.  See the `README.md` file it includes for more details.  Its only dependencies are `Mojolicious` and `Mojo::SQLite`.
[CaliperDashboard.zip](https://github.com/user-attachments/files/30908815/CaliperDashboard.zip)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,7 +109,6 @@ RUN apt-get update \
 	libgd-perl \
 	libhtml-scrubber-perl \
 	libhtml-template-perl \
-	libhttp-async-perl \
 	libio-compress-perl \
 	libiterator-perl \
 	libiterator-util-perl \
@@ -133,7 +132,6 @@ RUN apt-get update \
 	libnet-ip-perl \
 	libnet-ldap-perl \
 	libnet-oauth-perl \
-	libossp-uuid-perl \
 	libpadwalker-perl \
 	libpandoc-wrapper-perl \
 	libpath-class-perl \
@@ -231,18 +229,18 @@ COPY --from=base /opt/base/pg $APP_ROOT/pg
 RUN echo "PATH=$PATH:$APP_ROOT/webwork2/bin" >> /root/.bashrc \
 	&& mkdir /run/webwork2 /etc/ssl/local \
 	&& cd $APP_ROOT/webwork2/ \
-		&& chown www-data DATA ../courses logs tmp /etc/ssl/local /run/webwork2 \
-		&& chmod -R u+w DATA ../courses logs tmp /run/webwork2 /etc/ssl/local \
+	&& chown www-data DATA ../courses logs tmp /etc/ssl/local /run/webwork2 \
+	&& chmod -R u+w DATA ../courses logs tmp /run/webwork2 /etc/ssl/local \
 	&& echo "en_US ISO-8859-1\nen_US.UTF-8 UTF-8" > /etc/locale.gen \
-		&& /usr/sbin/locale-gen \
-		&& echo "locales locales/default_environment_locale select en_US.UTF-8\ndebconf debconf/frontend select Noninteractive" > /tmp/preseed.txt \
-		&& debconf-set-selections /tmp/preseed.txt \
+	&& /usr/sbin/locale-gen \
+	&& echo "locales locales/default_environment_locale select en_US.UTF-8\ndebconf debconf/frontend select Noninteractive" > /tmp/preseed.txt \
+	&& debconf-set-selections /tmp/preseed.txt \
 	&& rm -f /etc/localtime /etc/timezone && echo "Etc/UTC" > /etc/timezone \
-		&& dpkg-reconfigure -f noninteractive tzdata \
+	&& dpkg-reconfigure -f noninteractive tzdata \
 	&& cd $WEBWORK_ROOT/htdocs \
-		&& npm install \
+	&& npm install \
 	&& cd $PG_ROOT/htdocs \
-		&& npm install
+	&& npm install
 
 # ==================================================================
 # Phase 7 - Final setup and prepare docker-entrypoint.sh
--- a/DockerfileStage1
+++ b/DockerfileStage1
@@ -71,7 +71,6 @@ RUN apt-get update \
 	libgd-perl \
 	libhtml-scrubber-perl \
 	libhtml-template-perl \
-	libhttp-async-perl \
 	libio-compress-perl \
 	libiterator-perl \
 	libiterator-util-perl \
@@ -94,7 +93,6 @@ RUN apt-get update \
 	libnet-ip-perl \
 	libnet-ldap-perl \
 	libnet-oauth-perl \
-	libossp-uuid-perl \
 	libpadwalker-perl \
 	libpandoc-wrapper-perl \
 	libpath-class-perl \
--- a/bin/check_modules.pl
+++ b/bin/check_modules.pl
@@ -129,12 +129,6 @@ my %modulesList = (
 			ubuntu => 'libdata-structure-util-perl'
 		}
 	},
-	'Data::UUID' => {
-		package => {
-			ubuntu => 'libossp-uuid-perl',
-			rhel   => 'perl-Data-UUID'
-		}
-	},
 	'Date::Format' => {
 		package => {
 			ubuntu => 'libtimedate-perl',
@@ -281,11 +275,6 @@ my %modulesList = (
 		package => {
 			ubuntu => 'libhtml-parser-perl',
 			rhel   => 'perl-HTML-Parser'
-		}
-	},
-	'HTTP::Async' => {
-		package => {
-			ubuntu => 'libhttp-async-perl'
 		}
 	},
 	'IO::File' => {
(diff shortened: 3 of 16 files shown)


Alex-Jordan:
Do we need to ping Caliper users about this for testing?

drgrice1:
It wouldn't hurt. Although the point of providing the test instance is that with that we can test it. Caliper is just a 1edtech specification, just like LTI authentication. So what is important is that the specification is met.

xcompass (commented):
Nice cleanup overall — the plugin/hook split is a good direction, and replacing the blocking `HTTP::Async` `wait_for_next_response` loop with `Mojo::UserAgent` promises is a real win. Dropping `Data::UUID` and `HTTP::Async` looks safe: `UUID::Tiny` is already a dependency (used by `WeBWorK::Authen::LTIAdvanced::SubmitGrade`, `WeBWorK::ContentGenerator::LTIAdvanced`, and `WeBWorK::Utils::CourseManagement`) and is already present in `bin/check_modules.pl` and both Dockerfiles.

I went through this against a checkout of the branch, tracing each hook back to its emit site. Four things I think need addressing before this lands — two functional regressions and two silent-breakage risks for sites currently running Caliper.

---

### 1. `tags => \%$tags` no longer unblesses the Tags object

In `lib/Caliper/Entity.pm`, both `problem` and `problem_user` previously did an explicit hash copy specifically to strip the blessing:

```perl
my %tags_ref       = %$tags;
my $unblessed_tags = \%tags_ref;
```

That has been replaced with `tags => \%$tags`, but `\%$ref` returns a reference to the *same* hash and keeps the blessing:

```
$ perl -e 'my $o = bless {}, "Foo"; print ref(\%$o), "\n"'
Foo
$ perl -e 'my $o = bless {}, "Foo"; my %c = %$o; print ref(\%c), "\n"'
HASH
```

`WeBWorK::Utils::Tags` has no `TO_JSON`, so Mojo::JSON will stringify it and `extensions.tags` ends up as `"WeBWorK::Utils::Tags=HASH(0x…)"` instead of the tag data. `tags => {%$tags}` restores the old behaviour.

Related, in `problem` the `group:` branch now sets `$tags = { keywords => [] }`, so the two branches emit structurally different `tags` values (a plain `{keywords => []}` vs. a Tags object). Previously the group branch left `tags` undef entirely.

### 2. The `startTime` reset now runs after the page has been rendered

`Mojolicious::Plugins::emit_hook` ignores subscriber return values, so the `async sub` hook runs synchronously up to its first `await`, returns a promise that is discarded, and control goes straight back to `process_and_log_answer` / `pre_header_initialize`. Rendering then happens on the same tick, while the LRS POST is still in flight.

Both templates read the parameter at render time:

- `templates/ContentGenerator/Problem.html.ep:110` — `hidden_field(startTime => param('startTime') || time)`
- `templates/ContentGenerator/GatewayQuiz.html.ep:413` — `my $startTime = param('startTime') || time;`

So the `$c->param('startTime', '')` placed *after* `await $caliper_sensor->sendEvents(...)` in both hooks lands too late to affect the response — it is effectively dead code. The practical effect is that the hidden `startTime` field keeps its original value across submissions, so every subsequent `startedAtTime`/`duration` is measured from the first page load and grows without bound.

Moving the reset to before the `await` fixes it, and then the `my $tx = $c->render_later->tx;` line is no longer needed for that purpose (`WeBWorK::ContentGenerator::go` already retains the transaction).

### 3. Existing installs will silently stop emitting Caliper events

`lib/Mojolicious/WeBWorK.pm` only falls back to `conf/webwork2.mojolicious.dist.yml` when `conf/webwork2.mojolicious.yml` does not exist. Every existing site has its own copy, which will not gain the new `plugins:` block on upgrade. Those sites will keep `$caliper{enabled} = 1` in `localOverrides.conf`, see no error anywhere, and simply stop sending events.

Would it be worth emitting a warning at startup (or on the first event attempt) when `$caliper{enabled}` is true but the plugin has not been registered? Failing that, this needs to be called out prominently in the release notes, since there is no other signal that anything changed.

### 4. Removing `$caliper{base_url}` silently changes every resource IRI

The setting existed precisely because of the note that was removed:

```
# base_url should ideally be hard coded to a persistent url pointing to the webwork root
# (important to keep it consistent over time)
```

For any install where `$caliper{base_url}` differs from `$server_root_url . $webwork_url`, every actor, course, membership, set, problem, answer and attempt IRI changes after this upgrade. From the LRS's perspective those are new entities, so longitudinal analytics on the existing data break — and it happens with no error and no output difference that an admin would notice until they query the LRS.

A leftover `$caliper{base_url}` in `localOverrides.conf` is now just silently ignored. Could `Caliper::ResourceIri::new` (or startup) warn if `$ce->{caliper}{base_url}` is still set, pointing at `$server_root_url`/`$webwork_url`? That would give affected sites a chance to verify the two produce an identical string before the identifiers change underneath them.

drgrice1:
@xcompass: I fixed the first two issues.  The second two are standard upgrade procedure.  It is the expectation that webwork administrators compare their configuration files to the `.dist` versions and relegate changes.  There will be no warnings for this.

Note that `$ce->{server_root_url} . $ce->{webwork_url}` is a hard coded persistent URL.  Installs should never have a case where the old `$caliper{base_url}` differed from that.  That means your server was not set up correctly to begin with.

Ideally, the `Caliper` code would not be in this repository at all.  It would be in some other repository for those that want to use it.

drgrice1:
You should also note that the fallback to using the `webwork2.mojolicious.dist.yml` file is only for development purposes.  In production usage, you should always be using a `webwork2.mojolicious.yml` file.

xcompass:
Thanks for the clarification. Looks good to me. Feel free to merge. 

It would be nice to mention this breaking change somewhere so admin knows they need to make the configuration change to re-enable Caliper. 

Sorry for the delayed review. just came back from vacation.

drgrice1:
We will put notes about this in the release notes, including comments in the upgrade instructions.  The same as we would do for any notable change in the configuration files.  I added the "Put comment in release notes" tag to remind us.

somiaj (approved):
I haven't tested, code looks good, adding an approval since @xcompass stated it worked.

pstaabp (approved):
Also, I'll approve based on @xcompass review, since I haven't used this feature at all.

somiaj:
Two approvals, merging.

PR #3126, opened 2026-08-10
Make code meet perlcritic severity level 4.

drgrice1:
This adds a GitHub workflow that checks that Perl files in the `lib` directory satisfy perlcritic with the `.perlcriticrc` file included in this pull request.

I have been working toward this for years now (since December of 2022 according to the date on this commit).  Originally there were are large number of changes to files.  At this point, there are not so many, so I think it is time for this to go in.

Note there are some things that can't be helped (particularly with safe mode code).  So those are disabled with `## no critic` annotations. Almost every method in the `lib/WeBWorK/DB.pm` file breaks the `Subroutines::RequireArgUnpacking` rule, and so that is disabled for the entire file.

Note that if `no critic` annotations are added to the code, they should be specific and disable only the rule or rules that are to be ignored, and those rules should usually be enabled again later in the file after the rule breaking code. Other than the annotations already added in this pull request, this is generally something that should be avoided.


somiaj (approved):
Looks good, all files pass `perlcritic ./lib`.

I notice the work flow isn't testing perl (`.pm` and `.pl`) files in `./bin` and only files in `./lib`. Should the perl files in `./bin` be tested as well?

drgrice1:
Yes, it is intentionally only testing files in `./lib`.  That is because many of the scripts in the `./bin` directory still need a lot of work.  You can test it.  Just run `perlcritic ./bin`.  You will see lots of issues on some files.  Eventually, we will fix those and add those to be checked.

Alex-Jordan (approved):

PR #3128, opened 2026-08-12
Remove `images` display mode and even the `displayMode` option generally

drgrice1:
This is split into two commits.

The first just removes the `images` display mode, but generally leaves all behavior intact. This updates webwork2 for the corresponding change to PG.  The `images` display mode has been removed, and so remove the option to select it and the `refreshMath2img` translation option that is no longer used.
    
Note that the `equation_display` route that no one knows about and that can only be accessed by entering the URL `https://yourserver.edu/webwork2/courseID/equation_display` in the browser has been removed since it depends on the `ImageGenerator` package which was removed on the PG side.
    
All of the equation cache variables have been removed from the course environment, since those are no longer used.  The `remove_stale_images` script is no longer needed, and was deleted.
    
The non-native `depths` table that was only used for these images has been removed.  Note that the `depths` table will need to be manually deleted from the database. Although it won't hurt if it is still there.  It will quietly sit there unused if it is not deleted.

The second commit takes things further.  I realized that once the `images` display mode is removed, there really is no point in a user interface choice of display modes.  There is no real reason to actually enable the `plainText` display mode which has always been disabled by default, and so there is no real choice for the user anymore.  So it removes the `displayMode` and related options. Note that this only removes the selection as pertains to HTML display mode choices in the user interface.  Internally, there are still display mode decistions that are made to determine the type of problem render.

There is one case where the display mode choice is left, and that is in the problem editor.  That is because that is a different type of choice that is much the same as the internal decisions made elsewhere.

In short the displayMode is now only used to determine if it is hardcopy, html, or PTX.  In some cases plainText is used to save processing (such as in determining a new variant for a "Show Me Another" problem).


somiaj (approved):
Initial tests shows everything is working. I even tested leaving the displayMode settings in localOverrides, and it didn't affect the behavior.

Just out of curiosity what is `Math_tth`?

somiaj:
Opps, I meant `HTML_tth`.

drgrice1:
Do you mean `HTML_tth`?  It is an old display mode.  It is essentially obsolete and does not work on its own anymore.  However, it is a fallback used with the `HTML_MathJax` display mode.  The only reason it is still there is because of problems that incorrectly use some of the variables defined in `PGbasicmacros.pl`.  See https://wiki.openwebwork.org/wiki/Installation_Manual_for_2.21#PGbasicmacros.pl for details.

Alex-Jordan (approved):

PR #3129, opened 2026-08-12
Remove the script tag processing of MathJax.

drgrice1:
The injection of `<script type="math/tex">...</script>` tags by PG is removed by https://github.com/openwebwork/pg/pull/1499.  So the script tag handling in the `mathjax-config.js` file is no longer needed.

### DIFF
--- a/htdocs/js/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/MathJaxConfig/mathjax-config.js
@@ -143,34 +143,7 @@ if (!window.MathJax) {
 				});
 			}
 		},
-		options: {
-			renderActions: {
-				findScript: [
-					10,
-					(doc) => {
-						if (doc.processed.isSet('findScripts')) return;
-						const containers = doc.adaptor.getElements(doc.options.elements || [doc.document.body], doc);
-						for (const container of containers) {
-							for (const node of container.querySelectorAll('script[type^="math/tex"]')) {
-								const math = new doc.options.MathItem(
-									node.textContent,
-									doc.inputJax[0],
-									!!node.type.match(/; *mode=display/)
-								);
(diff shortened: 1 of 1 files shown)


somiaj:
I think you linked the wrong PR in the commit and description.

somiaj (approved):

drgrice1:
Should this be a hotfix also?  It is actually not necessary for https://github.com/openwebwork/pg/pull/1501.

somiaj:
I could go either way. I doubt not processing script tags that aren't there won't speed things up in a noticeable fashion. But couldn't hurt.

Alex-Jordan:
I won't have time to check this (and the associated PR in pg) until later today. But the one thing on my mind to check is how/if PTX inserts any math (specifically for defining an author's LaTeX macros) into the iframe srcdoc for an embedded webwork file. If there's a real issue there, it's possible that it would be better to leave this be on main and give me a year to work out something different for 2.22 and beyond.

somiaj:
@Alex-Jordan well this doesn't need to be a hotfix, and is independent of the PG one, in the sense we could remove using of script tags from PG (PGML isn't using them already), but leave this alone so webwork can still process them if something like pretext needs that processing.

Alex-Jordan (approved):

Alex-Jordan:
If there will be a hotfix for this, I think I need to know as soon as possible. I have some work to do (with PreTeXt) so that PreTeXt works with a 2.21 server. And if this script tag happens for 2.21, I should know about that just in case it matters. I have reason to believe that it would.

somiaj:
I don't think a hotfix is needed, specifically it does effect pretext. Though maybe you should try to update things now to not use script tags if possible to prepare for 2.22.

PR #3131, opened 2026-08-13
Add a missing `debug` import in `lib/WeBWorK/Authen/LTIAdvantage/SubmitGrade.pm`.

drgrice1:
This was missed in #3126.  The `WeBWorK::Debug` package no longer exports the `debug` method by default, and it must be explicitly imported where needed.


somiaj (approved):

drgrice1:
I figured I better check all of these, and found one more in the `lib/WeBWorK/File/SetDef.pm` file.  That is all of them though.

Alex-Jordan (approved):

PR #3132, opened 2026-08-13
Fix an issue with the `navigation_allowed` permission.

drgrice1:
Do not let a user view another set other than the one in the session if a user does not have the `navigation_allowed` permission.  Generally, the only way this will happen is if the user modifies the URL in the browser to try to change to a different set. Currently that works, and it shouldn't.

### DIFF
--- a/lib/WeBWorK/Authz.pm
+++ b/lib/WeBWorK/Authz.pm
@@ -434,6 +434,18 @@ sub checkSet {
 		return $c->maketext("Requested set '[_1]' is not available yet.", $setName);
 	}
 
+	if (!$self->hasPermissions($userName, 'navigation_allowed') && ($c->authen->session->{set_id} // '') ne $setName) {
+		$c->{viewSetCheck} = 'restricted';
+		if ($ce->{LTI}) {
+			# Note that this content is HTML escaped in the template, and so this may not contain the link to
+			# $ce->{LTI}{ $ce->{LTIVersion} }{LMS_url} as is done with similar such messages for this.
+			return $c->maketext('You must access this assignment from your Course Management System ([_1]).',
+				$ce->{LTI}{ $ce->{LTIVersion} }{LMS_name});
+		} else {
+			return $c->maketext('You do not have permission to access this set.');
+		}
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):

somiaj:
I don't have LTI fully setup on my develop server, but I noticed the following as an instructor:

In course configuration set `navigation_allowed` to `nobody`.

Go to set manager, and click on the problem number for a set to get into the set details editor, from there the link to the set appears under the greyed out assignments page. Click on that to get to the link to the set (or typing in the url directly) gives a warning:

```
Use of uninitialized value in string ne at /home/jaimos/git/openwebwork/webwork2/lib/WeBWorK/Authz.pm line 437.
```

Now this is not something anyone should be doing, but the error states that there could be a chance that either `$c->authen->session->{set_id}` could be undefined. Unsure if there is a way that could happen.

lib/WeBWorK/Authz.pm
```diff
@@ -434,6 +434,18 @@ sub checkSet {
 		return $c->maketext("Requested set '[_1]' is not available yet.", $setName);
 	}
 
+	if (!$self->hasPermissions($userName, 'navigation_allowed') && $c->authen->session->{set_id} ne $setName) {
```
somiaj: Changing this to:

```
if (!$self->hasPermissions($userName, 'navigation_allowed') && ($c->authen->session->{set_id} // '') ne $setName) {
```

Fixes the warning I ran into.

drgrice1: I made that change, but I am certain that if you start changing various permissions to `nobody` that really shouldn't be, and that don't make sense to be `nobody`, you will find lots of other more serious problems with the code not being prepared to deal with that.

somiaj:
Agreed, this permission doesn't work well for anything above `login_proctor` which the help states to use if you want to restrict student navigation.

I was just worried their maybe other cases where `$c->authen->session->{set_id}` could be undefined for a student user.

somiaj (approved):

PR #3133, opened 2026-08-15
Fill in missing translations with AI-assisted drafts

drgrice1:
Fill in empty msgstr entries in the 13 non-English locale .po files using Claude, following each file's own pre-existing terminology conventions where they existed.  Also remove leftover dummy `"tr: <English>"` placeholder stubs that had been inserted.

This matches the corresponding pull request to PG, and is an approach to obtaining full translation. I obviously cannot attest to how good the translations are, and these should be reviewed by native speakers.

Confidence notes per locale, for reviewers:

- en-GB: highest confidence. Not a translation — a deterministic American-to-British spelling/terminology substitution (colour, behaviour, etc.) applied via word-boundary matching against the source corpus. 98 of 3189 strings actually changed; everything else (placeholders, HTML, punctuation) is byte-identical to en.po.
- fr-CA, tr, ko, zh-CN, ru-RU, cs-CZ, he-IL: moderate-to-high confidence. Each file already had substantial pre-existing human-translated coverage (roughly 15-45% of strings) before this pass; new strings were translated to match that file's own established terminology rather than inventing fresh vocabulary.
- fr, es, el, hu, de: standard confidence. These files were nearly or fully empty of prior translation, so terminology (e.g. fr's usager/courriel/devoir house style) was chosen fresh this session and has no prior human precedent to anchor it — worth extra scrutiny from a native reviewer.
- zh-HK: standard confidence for the ~3120 newly-translated strings, with two caveats worth flagging to reviewers. First, of the 96 strings that already had non-blank msgstr values before this session, 27 turned out to be Turkish/French text (some further garbled into nonsense CJK-range characters), almost certainly a legacy Transifex cross-locale data bleed unrelated to this project; those were identified and retranslated into Chinese rather than left as false "already covered" entries. Second, the remaining ~69 legitimately-Chinese pre-existing entries are in Simplified script, while all new entries added this session are in Traditional script (matching what "zh-HK" is supposed to mean) — this pre-existing Simplified/Traditional split was deliberately left alone rather than silently rewritten, and should be reconciled by a native reviewer.
- he-IL: standard confidence, with one caveat already fully resolved. A latent bug in the batch-fill tooling (it only checked the first line of a multi-line-wrapped msgstr field to decide "is this blank") transiently corrupted 29 pre-existing entries during the first batch; all 29 were recovered from git history and verified, and the tooling was fixed before continuing. Confirmed via placeholder audits that no other locale in this project was affected by the same bug.


Alex-Jordan (approved):
The diffs are all too large to view in GH. I'm just marking approval. I wouldn't catch anything that you wouldn't have caught.

somiaj (approved):
I too don't have any way to verify these. They only affect translations and if users have issues with them, they can improve them.

somiaj:
I decided to check some local changes, and saw something not part of this PR. I noticed that various translations remove leading/trailing spaces. Could this be something we could ask claud or some script to double check?

Example:

```
+#: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/ProblemSet.pm:318
+#: /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/ProblemSet.pm:337
 msgid " Answers Available."
 msgstr "Lösung verfügbar."
```

I assume that space might be needed, a script could go double check that leading or ending spaces are not being stripped by translations.

somiaj:
I am going to merge this as is. If someone wants to address translations taking away leading/trailing spaces, that can be another PR.

drgrice1:
I am sure that AI could help to make sure the spaces remain.  Although we really need to fix those things in the code.  Spaces should not be part of the translations if they are needed.

Alex-Jordan:
About the leading/trailing spaces. Maybe if there is a structural need for spaces like that, they should be moved out of the `maketext`.

Issue #3138, opened 2026-08-18
Set edtior user overrides settings table columns too small.


somiaj:
When overriding the settings for a single user in the set editor, the columns for the general parameters and set defaults are too small, causing the set default date to be cut off.

<img width="763" height="479" alt="Image" src="https://github.com/user-attachments/assets/198b9cb2-bd55-40ea-9314-db57f1808f70" />

somiaj:
Fixed in #3142.

PR #3142, opened 2026-08-19
Improve the width of the field table on the set detail page.

drgrice1:
Remove the `w-75` class from the third column in that table.  That width specifier is not right for a column in this table, and results in the column being to wide.  The width of the column is good on its own.

Also, for the problem field tables switch to using `col-lg-6 col-xl-5 col-xxl-4` for the field table and `col-lg-6 col-xl-7 col-xxl-8` the problem rendering area when editing a set for all users, and use `col-lg-7 col-xxl-6`` for the field table and `col-lg-5 col-xxl6` for the problem rendering area when editing the set for users.  This was always `col-lg-5` and `col-lg-7` prior to WeBWorK 2.21, and was switched to always `col-lg-8` and `col-lg-4` in #3090 for WeBWorK 2.21. The pre 2.21 setting often made the field table to narrow, but the 2.21 setting went to far makes it too wide and leaves to small of a width for problem rendering.

This fixes issue #3138.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -761,7 +761,7 @@ sub fieldTable ($c, $userID, $setID, $problemID, $globalRecord, $userRecord = un
 		);
 	}
 
-	unshift @$rows, $c->tag('col') x 2, $c->tag('col', class => 'w-75'), $forOneUser ? $c->tag('col') : '';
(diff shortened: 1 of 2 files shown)


somiaj (approved):

somiaj:
While we are at it, would it be worth adding some bigger break points too? Maybe `col-12 col-lg-6 col-xl-5 col-xxl-4` with appropriate pairing for the two columns (and adding 1 column for individual set detail page).

drgrice1:
@somiaj: I added your suggestion.  I also added one additional breakpoint when editing for users.

Alex-Jordan (approved):

PR #3143, opened 2026-08-19
Improve the width of the field table on the set detail page. (hotfix of #3142)

drgrice1:
Remove the `w-75` class from the third column in that table.  That width specifier is not right for a column in this table, and results in the column being to wide.  The width of the column is good on its own.

Also, for the problem field tables switch to using `col-lg-6 col-xl-5 col-xxl-4` for the field table and `col-lg-6 col-xl-7 col-xxl-8` the problem rendering area when editing a set for all users, and use `col-lg-7 col-xxl-6`` for the field table and `col-lg-5 col-xxl6` for the problem rendering area when editing the set for users.  This was always `col-lg-5` and `col-lg-7` prior to WeBWorK 2.21, and was switched to always `col-lg-8` and `col-lg-4` in #3090 for WeBWorK 2.21. The pre 2.21 setting often made the field table to narrow, but the 2.21 setting went to far makes it too wide and leaves to small of a width for problem rendering.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
@@ -761,7 +761,7 @@ sub fieldTable ($c, $userID, $setID, $problemID, $globalRecord, $userRecord = un
 		);
 	}
 
-	unshift @$rows, $c->tag('col') x 2, $c->tag('col', class => 'w-75'), $forOneUser ? $c->tag('col') : '';
+	unshift @$rows, $c->tag('col') x ($forOneUser ? 4 : 3);
 
 	return $c->tag(
 		'table',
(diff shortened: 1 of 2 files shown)


somiaj (dismissed):

Alex-Jordan (approved):

Alex-Jordan:
@somiaj You approved the non-hotfix counterpart to this, but not this hotfix one. I've approved this one, so you could approve it and merge it.

somiaj (approved):

somiaj:
@Alex-Jordan I did approve this one, but the approval was removed with the change, which I didn't realize was a feature in the main branch.

drgrice1:
> @Alex-Jordan I did approve this one, but the approval was removed with the change, which I didn't realize was a feature in the main branch.

Yeah, I set that feature for pull requests to the main branch.  I did that a while ago after we talked about it, and I think it was actually you that suggested that this be set.

PR #3144, opened 2026-08-19
Update `WWSafe.pm`.

drgrice1:
The updated `WWSafe.pm` is based on version 2.47 of the upstream `Safe.pm` module. It is essentially the upstream module except that some reentrancy issues with `reval` and `rdo`, and an issue with sharing an entire package to the safe compartment are fixed.

These are basically the same changes as to the WWSafe.pm module for PG.

Note that webwork2 uses this for three things:

- to evaluate the configuration files and construct the course environment.
- to evaluate achievement evaluators
- to generate achievement notification email text from the template set for the achievement notificaiton

So clearly those are the things to test.

### DIFF
--- a/lib/WeBWorK/WWSafe.pm
+++ b/lib/WeBWorK/WWSafe.pm
@@ -3,9 +3,9 @@ package WeBWorK::WWSafe;
 
 #use 5.003_11;
 use 5.12.0;
-use strict;
-use utf8;
-$WeBWorK::WWSafe::VERSION = "2.16";
+use Scalar::Util qw(reftype refaddr blessed);
+
+$WeBWorK::WWSafe::VERSION = "2.47";
 
 # *** Don't declare any lexicals above this point ***
 #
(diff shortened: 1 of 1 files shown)


somiaj:
I am testing these as I write problems. I did notice that both need to be rebased onto develop as they are a few commits behind.

drgrice1:
I rebased this pull request.  What did you mean by both of these?  The corresponding PG pull request was on the tip of develop.

somiaj:
Ahh I hadn't fetched your branches in my local PG, and got a slightly older version of the branch.

somiaj:
Do you have any suggested way to test this? I tested a few of what I consider my more complicated custom checkers and saw answers are still be accepted as correct. I have also gone through the library browser and viewed a decent chunk of my local problems (no actual testing the graders work). So far I see no errors/warnings. I am going to approve so this can get in, but unsure what should be tested. Note I'm using both `safe-update` branch here (which is for achievements I haven't tested that much) and on the PG side for testing problems.

somiaj (approved):

drgrice1:
What you have done is basically it for testing the PG side.  Look for problems that might do something different code wise (different macros, different underlying module usage, etc.) and make sure that they work.

For the WeBWorK side, any page you load is testing this pull request.  That checks that the course environment loads and works as it should.  Then test achievements, and make sure those are earned when they should be.  Do what you can with achievement notifications, if you can test emails.

pstaabp (approved):
Maybe merge and we can test as we run the develop branch over the next bunch of months.

PR #3145, opened 2026-08-19
Setup for PG dark mode.

drgrice1:
Rework all of the problem rendering to support dark mode for PG.

This is straightforward for problems in sets.  All of the code added to force light mode for problems just needed to be removed.

For the problem editor and all of the places that problems are rendered in an iframe via the `render_rpc` end point, a little work was needed to allow the parent to control the color scheme. The child (whose content is rendered via `templates/RPCRenderFormats/default.html.ep`) now sends a message to the parent to signal that it has loaded. This occurs at the beginning of the page load (it is important that this is earlier than the iframe `load` event or flickering will occur), and then the parent posts a message back to the child with the data `{ theme: 'mode' }` where `mode` is either `dark` or `light` to set the mode of the iframe.

The problem editor is generally reworked so that everything is properly rendered in dark mode such as hardcopy theme xml and PG critic results.

### DIFF
--- a/htdocs/js/GatewayQuiz/gateway.scss
+++ b/htdocs/js/GatewayQuiz/gateway.scss
@@ -108,9 +108,8 @@ div.gwDivider {
 	margin: 0px 0px 10px 0px;
 }
 
-/* Override the pg style so that the problem-content is not offset in gateway quizzes and force a light color scheme. */
+/* Override the pg style so that the problem-content is not offset in gateway quizzes. */
 .problem-content {
-	color-scheme: light;
 	padding: unset;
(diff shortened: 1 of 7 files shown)


somiaj (approved):

somiaj:
Another thing I noticed with this is MathJax menu isn't in dark mode, did that need to be updated to honor the darkmode setting?

somiaj:
Update, the MathJax menus seem to only honor the browser dark mode setting, not the site light/darkmode setting.

pstaabp (approved):

PR #3146, opened 2026-08-20
Update student orientation feedback.

somiaj:
Update this problem to match the changes in PG to use custom classes for the colors and add the option of unknown correctness.

Also try to keep lines under 80 characters.


drgrice1 (approved):
Looks good.

This should probably be a hotfix as well, since these are the buttons in the 2.21 release.

Alex-Jordan (approved):

PR #3161, opened 2026-08-20
Fix make sure key is defined before comparing it.

somiaj:
Make sure `$localData->{set_id}` is defined before comparing it in a `ne` comparison in still_no_right achievement.

Found another achievement that compared a value that could be undefined.

### DIFF
--- a/courses.dist/modelCourse/templates/achievements/still_not_right.at
+++ b/courses.dist/modelCourse/templates/achievements/still_not_right.at
@@ -6,6 +6,7 @@
 
 # Initialize $localData.
 $localData->{incorrect_streak} //= 0;
+$localData->{set_id}           //= '';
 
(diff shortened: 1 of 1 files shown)


courses.dist/modelCourse/templates/achievements/still_not_right.at
```diff
@@ -6,6 +6,7 @@
 
 # Initialize $localData.
 $localData->{incorrect_streak} //= 0;
+$localData->{set_id}           //= '';
```
drgrice1, line 9: You could also change line 18 below to
`if ($problem->set_id ne ($localData->{set_id} // '')`.

But this works too.  Either way is fine.  Just noting.

drgrice1 (approved):

Alex-Jordan (approved):

PR #3165, opened 2026-08-21
Test time announce icon

somiaj:
Place a small button with a speaker icon next to the remaining time on tests instead of a full button on the second row. This makes the remaining time + button overlay take up less space.

I went with using a "sound wave" icon, but am open to what ever icon is preferred. Here is what the remaining time overlay looks like with this:

image

I originally tried a megaphone, but the bootstrap one didn't look right to me (maybe a megaphone with some sound waves coming from it would be useful). Here is what the megaphone looks like.

image

@Alex-Jordan any thoughts? Also is `aria-label` enough to make this accessible?

### DIFF
--- a/templates/ContentGenerator/GatewayQuiz.html.ep
+++ b/templates/ContentGenerator/GatewayQuiz.html.ep
@@ -220,11 +220,13 @@
 	%
 	% # Print the timer if there is less than 24 hours left.
 	% if ($timeLeft < 86400) {
-		<div id="gwTimerContainer" class="d-flex flex-column gap-2 mb-2">
+		<div
+			id="gwTimerContainer"
+			class="alert alert-warning d-flex justify-content-center align-items-center gap-2 py-1 mb-0"
+		>
 			<%= tag(
-				'div',
+				'span',
 				id            => 'gwTimer',
-				class         => 'alert alert-warning p-1 mb-0',
 				'aria-hidden' => 'true',
 				data          => {
 					server_time     => int($submitTime),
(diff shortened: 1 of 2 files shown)


somiaj:
@drgrice1 is there a way to format a single file, vs `npm run prettier-format` having to search through all files? Specifically format a template file.

drgrice1:
I use the formatter in vim via the coc.nvim coc-prettier extension.

drgrice1:
Maybe it is time that we stop using Font Awesome, and switch to using Bootstrap icons.  Then we can use their `i` tag syntax for them instead of pasting the svg as you have done here.  Bootstrap's icons are entirely free, and they have a good selection of icons.  Font Awesome has become annoying with their very limited free selection.  I am sure we could find decent Bootstrap replacements for all of the Font Awesome icons we are using.  In some cases, there are probably even better Bootstrap icons available.

somiaj:
I am using bootstrap icons here, but I copied the svg code for them, not just the `<i>` code as I wasn't sure if we load the libraries for it or not. I'm okay with whatever change. I mostly go with bootstrap icons as I think they fit in bootstrap better.

drgrice1:
The Font Awesome icons go well with Bootstrap also.  In fact, historically, the Font Awesome icon set was created for usage with Bootstrap.

Alex-Jordan:
Just one quick thing until I can test for real. Could there be more space between the text and the button?

drgrice1:
I thought the same thing.

somiaj:
Added some space, only `me-1` (I thought `me-2` was too much, but could make it more if you prefer).

image

Alex-Jordan:
I'd go more. Also, should it be margin at the beginning of the button, rather than margin at the end of the text?

Does removing the p-1 make the button too small? Asking because presently the button seems large up against the yellowish behind it. I'm not looking at a live build though.

Last nitpick. The icon in the button does not appear to be centered. I see it as slightly to the right of center, as if there is some extra padding to its left.

somiaj:
The actual SVG I am using was 16x16 (thought it was a bit small) and I increased the size to 20x20, I think the center is due to scaling.

I don't know if it matters where the margin is. I went with putting it at the end of the text, because in most places the margin was applied to the bottom (end) of the previous object, not the object itself. I can update that if preferred.

Padding is worse without `p-1`, and I think with `p-0` it doesn't look good. I instead used some CSS to make the horizontal padding `2px` and vertical `1px`. Changed the margin of the time span to `m-2`. This is what it looks like.

image

somiaj:
I added `align-middle` to the time span so it centers vertically better instead of being slightly above vertical.

image

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -220,11 +220,11 @@
...
 	% # Print the timer if there is less than 24 hours left.
 	% if ($timeLeft < 86400) {
-		<div id="gwTimerContainer" class="d-flex flex-column gap-2 mb-2">
+		<div id="gwTimerContainer" class="alert alert-warning text-center py-1 mb-0">
 			<%= tag(
-				'div',
+				'span',
 				id            => 'gwTimer',
-				class         => 'alert alert-warning p-1 mb-0',
+				class         => 'align-middle me-2',
```
drgrice1: Change this to


```suggestion
		<div
			id="gwTimerContainer"
			class="alert alert-warning d-flex justify-content-center align-items-center gap-2 py-1 mb-0"
		>
			<%= tag(
				'span',
				id            => 'gwTimer',
```

That fixes the issues with the icon in the button not being aligned correctly.  Also, using a flex `gap` instead of a margin either at the end of the text span or beginning of the button is always more reliable.

drgrice1: Note that the `text-center` class was the primary problem.  Using that with non-text such as a button never quite works right.

drgrice1:
> Maybe it is time that we stop using Font Awesome, and switch to using Bootstrap icons. Then we can use their `i` tag syntax for them instead of pasting the svg as you have done here. Bootstrap's icons are entirely free, and they have a good selection of icons. Font Awesome has become annoying with their very limited free selection. I am sure we could find decent Bootstrap replacements for all of the Font Awesome icons we are using. In some cases, there are probably even better Bootstrap icons available.

Actually, I messed with this a bit.  The Font Awesome icons are generally just better, and Font Awesome has some things that you can't do at all with Bootstrap icons without a lot of effort (such as sizing classes and spinners).

templates/ContentGenerator/GatewayQuiz.html.ep
somiaj: Updated.

drgrice1 (approved):

Alex-Jordan (approved):

drgrice1:

drgrice1:
> @drgrice1 is there a way to format a single file, vs `npm run prettier-format` having to search through all files? Specifically format a template file.

@somiaj: I forgot to answer this.  To run `prettier` on a single file use `npx prettier --write relative/path/to/file` from the `htdocs` directory.  For example, `npx prettier --write ../templates/ContentGenerator/GatewayQuiz.html.ep` or `npx prettier --write js/GatewayQuiz/gateway.scss`.  Or do both of those files at once with `npx prettier --write ../templates/ContentGenerator/GatewayQuiz.html.ep js/GatewayQuiz/gateway.scss`.

Issue #2761, opened 2025-06-30
How to move webworkConfig out of inline JavaScript in templates/layouts/system.html.ep


taniwallach:
I am interested in getting the core WW2 functionality to work with a Content-Security-Policy which does not use `unsafe-inline` for `script-src`.

I'm not sure what is the best approach to replace the `script` block in `templates/layouts/system.html.ep` which sets the value of `webworkConfig` (and the value of `webwork_url` inside that object).

It seems one option would be to create a ` .js.ep` template file, and have the controller handle requests for that file.

I'm not sure if it feasible to just add the standard value into `htdocs/js/System/system.js` and then somehow update that value if needed via a value passed in somehow in the HTML code (via a hidden form field?).

At present, I am working around this by using a hash value for the script block generated in the CSP. It could be that might remain the best approach, at least for now.

@drgrice1 Your advice on how to best approach this would be appreciated.

drgrice1:
A `script-src` nonce value would probably be the only real way that this can be done.  Then there is not need for something like `.js.ep` template.  Basically, for each request there would be a nonce generated for that request.  Then all inline scripts and css style tags that we want to allow would need to have the `nonce` attribute added with that nonce value, and content security policy `script-src` would need to be set to that nonce value for that request.

The script tag in `templates/layouts/system.html.ep` where the `webworkConfig` is defined would just need that `nonce` attribute to work.  Note that you could not replace the `webworkConfig` with a hidden form field.  There is a lot more to the `webworkConfig` variable than what is set in `system.html.ep`.  It is used several places in JavaScript and other values added to it that are not just text.

That approach could be made to go all the way for PG.  A new translation option would need to be added to PG that is the value of this nonce.  Then everywhere in PG that we want to allow inline javascript or css, that nonce attribute would need to be added.

taniwallach:
Thanks! For now - I'm handling this script using a hash in the CSP. Without that - various things fail to work, in particular the problem display in the set-editor and the preview display in the problem editor.

The nonce based approach would also require creating some reasonable manner to manage the dynamic creation of the CSP.

That is fare more than I think is wise to try to do before the pending release. It can wait for after the release when develop is up-to-date.

About the matter of using a nonce for styling. Right now, our info-security people are willing to allow me to keep `unsafe-inline` for `style-src`. They complained only about it being a high severity problem in `script-src`. We have some problems which "broke" mid-semester due to the security demand, but for now - hopefully almost everything we are using is still working.

[...]

PR #3169, opened 2026-08-26
Add csrf token protection to API calls.

drgrice1:
Anytime a page is rendered a csrf token is added to the `webworkConfig` variable. This is then retrieved in JavaScript anytime an API call is going to be performed and is attached to the request. The API will now refuse any request that does not have this token or if the token does not match the token in the session. This utilizes the Mojolicious `csrf_token` helper method.

The point of this is that a user that has properly logged into webwork in the browser may be tricked into doing something in another tab or window by an attacker that would cause a request to be sent to the API. That request would automatically have the valid WeBWorK session cookie attached to it by the browser, and so the request would succeed and could do whatever the user has permission to do via the API. Adding the csrf token prevents that because the other tab or window would not have access to the `csrf_token` in the `webworkConfig` JavaScript variable.

This was one of the vulnerabilities that @Alex-Jordan found with Claude before.

This does mean that using the API via a script will not work as easily. You would have to extract the `csrf_token` either from the cookies in the response or by parsing the HTML to find the `csrf_token` in the `webworkConfig` variable. But it can still be done.  If we really want a proper API for usage in scripts we could make it possible for users to generate an API token in the UI (those with sufficient permissions at least), and the API token added as an `Authorization: Bearer` header could be used as an alternative to the `csrf_token`. At this point I don't think this is needed though.  I doubt anyone really uses the API in this way.

Note that this does not apply to the `render_rpc` endpoint. It probably should, but that would cause issues for PreTeXt an others using this endpoint in the same way (although this could be worked around in much the same way that cookie disabling is done if the `allow_unsecured_rpc` option is set). At this point I think this is also not necessary as this endpoint generally can't do anything destructive. It just renders problems.


lib/WeBWorK/ContentGenerator.pm
```diff
@@ -695,7 +695,8 @@ sub webwork_js_config ($c, $showMathJaxErrors = 0) {
 	return encode_json({
 		webwork_url             => $c->location,
 		mathJaxBSColorSchemeUrl => getAssetURL($c->ce, 'js/MathJaxConfig/bs-color-scheme.js'),
-		$showMathJaxErrors ? (showMathJaxErrors => true) : ()
+		csrf_token              => $c->csrf_token,
+		$showMathJaxErrors ? (showMathJaxErrors => true) : (),
```
somiaj: Not that it matters, but any reason for adding the extra `,` at the end of the last line here?

drgrice1: That is probably because I originally added the `csrf_token` last, and the comma, and then moved the `csrf_token` up, and forgot to delete the comma.  I removed it now.

taniwallach:
This is likely to make significant trouble for anyone using a `Content-Security-Policy` which does not allow `unsafe-inline` for `script-src` (or `script-src-elem`). Since the `csrf_token` is not static (and should not be) - using a fixed hash value to permit the inline code setting `webworkConfig` will no longer work.

The best option if the current approach is used seems to be to teach WW to manage the generation of the `Content-Security-Policy` and to then use a nonce value to permit use of inline JavaScript for each specific call. @drgrice1 suggested that in https://github.com/openwebwork/webwork2/issues/2761#issuecomment-3019098152.

I am reopening https://github.com/openwebwork/webwork2/issues/2761.

**Other options:**

Would it be sufficiently secure and feasible to send the `csrf_token` via a hidden form field, or somewhere else in the HTML page?

Would a server-side approach based on "Fetch Metadata header" work?
See:
  * [OWASP check sheet on Cross-Site Request Forgery Prevention](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#fetch-metadata-headers)
  * [Blog post about the method](https://blog.miguelgrinberg.com/post/csrf-protection-without-tokens-or-hidden-form-fields)

---

BTW There is a CVE for the Mojolicious `csrf_token` support before version 9.48.
  * https://www.sentinelone.com/vulnerability-database/cve-2026-15747/
  * https://ubuntu.com/security/CVE-2026-15747

drgrice1:
While the `csrf_token` could be set in a hidden input, and I did consider that approach, this approach is much cleaner.  At this point, I have to say that WeBWorK does not support using the `Content-Security-Policy` with `unsafe-inline`. That would break lots of other things as well beside the `webworkConfig` variable.  If you (or someone) would like to implement a nonce hash that would be added to all of WeBWorK/PG's inline scripts, then we could support that.

taniwallach:
@drgrice1 Thanks for the response. I this case I am going to try to argue that the current development team needs to consider this matter further.

I suspect that the proposed policy / PR may make the next release of WeBWorK no longer feasible to use in some universities and colleges, and in particular in my own.

Modern security practice both strongly encourage use of a CSP, and avoiding inline JavaScript. Where needed, the inline Javascript is recommendation is to use nonces.

Given the current web security situation, I am very skeptical that the information security team here will allow me to drop the CSP or to permit `unsafe-inline` for the `script-src`. If there are a few pieces of static inline code here and there in WW 2.21/PG 2.21 or future versions, I can still use WW by adding necessary fixed hash values. I suspect that what is being done here will really make using several core features essentially impossible. I'm not sure to what extent student use would be impacted.

I'm not sure what other current "core" WW/PG functionality depends on inline scripts. I suspect that for core functionality most of any inline code is static code which could be moved into external files. If there are other features which really need dynamically generated inline code, that it would seem to me to be in the teams interest to add the CSP nonce support. If this is one of the only real uses of dynamic inline JS, maybe it is worth considering a less elegant but more "security recommendation" compliant approach for the `csrf_token`.

About the PG problem side of things, my recommendation would be to consider if it remains a defensible policy to allow problem authors to trigger the inclusion and execution of JavaScript code they can provide. Since PG provides problem authors unlimited ability to place content into the HTML, a CSP without `unsafe-inline` on `script-src` is essentially the only way to prevent a webwork server from distributing malicious code which will be run by the client browsers. A standard reasonably restrictive CPS will also disallow loading untrusted external libraries. For better or worse, allowing any staff who can edit problems so much "power" is probably not something that university information security teams are likely to be happy with. How to attach a nonce value to inline JavaScript added by PG macros/libraries shipped as part of the system is something which can be considered and handled.

Note: As of WW 2.19, the only thing other than the `webworkConfig` variable which was critical to everything we are using was the LTI repost logic, which I moved to a separate JavaScript file. There may be some specific advanced PG features which use inline scripts (the GeoGebra support provided by edFinity did). However, due to the security restrictions imposed by adding a pretty strict CSP, we only needed to stop using or modify under 10 questions. The ones I recall were either using GeoGebra applets in a non-critical way (so they applets were dropped or had some custom problem code adding inline JS for some reason. For WW 2.21 - I am in the middle of a transition (swapping between WW 2.19 and WW 2.21 while trying to find stable and functional settings for WW 2.21), and some issues may come up.

At present, I certainly cannot do the work to add CSP management and add nonces to WeBWorK/PG's inline scripts. 

If it helps to get some perspective, my current CSP for WW 2.21 looks roughly like:
```
    Content-Security-Policy: "default-src 'none'; script-src 'self' ; script-src-elem 'self' 'sha256-HASH-for-WebworkConfig' https://cdn.jsdelivr.net/npm/@mathjax/mathjax-newcm-font@4.1.3/chtml.js https://cdn.jsdelivr.net/npm/@mathjax/mathjax-newcm-font@4.1.3/svg/dynamic/hebrew.js ; style-src 'self' 'unsafe-inline'; img-src 'self' data: ; font-src 'self' data: ; connect-src 'self' ; media-src 'self' ; object-src 'self'  ; frame-ancestors 'self' https://moodle25.technion.ac.il ; frame-src 'self' ; form-action 'self' https://moodle25.technion.ac.il  ; upgrade-insecure-requests; base-uri 'self' ;  worker-src 'self' blob: https://moodle25.technion.ac.il ;"
```
  - I think the `blob` for `worker-src` was needed for something MathJax related.
  - I dropped the long list of moodle URLs from different academic years which remain active and in the CSP.

drgrice1:
You are missing a lot of PG functionality that uses inline scripts.  The `GraphTool`, images injected via the `plots.pl` macro (which is the leading method for adding images to pages at this point), etc.  Those do NOT use static code that can be moved to a separate JavaScript file.  Also, there are documented approaches to adding buttons for MathQuill inputs that use inline JavaScript (see the `tutorial/sample-problems/ProblemTechniques/NamedAnswerRules.pg` problem in PG).  There is a lot of modern functionality that has been added that needs interaction between the Perl code and JavaScript.  The only way to do that is to use inline scripts.  As time progresses and OPL problems are updated, you are going to find a lot more problems that stop working with the `Content-Security-Provider` headers you are using.

You are also missing that the `webworkConfig` variable is already not static, even without this pull request, and there are quite a few things that will fail with your fixed hash.

Yes, MathJax uses a `blob` worker from the `tex-svg.js` file.

drgrice1:
Closing this in favor of #3192.

PR #3177, opened 2026-08-29
Fix the test student nav to show all versions of tests.

drgrice1:
In the rework of the test student nav there was a flaw in design that led to only one version of a test being shown for each student.  It is not the right thing to iterate over the users For which a test version exists.  Instead the actual test versions need to be iterated over.  The users are looked up in a hash in that loop.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1376,86 +1376,82 @@ sub nav ($c, $args) {
 		# Find all versions of this set that have been taken (excluding those taken by the current user).
 		my @userVersions =
 			$db->listSetVersionsWhere({ user_id => { '!=' => $userID }, set_id => { like => "$setID,v\%" } });
+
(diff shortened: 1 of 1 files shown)


somiaj (approved):

somiaj:
In testing this out, I had some thoughts (nothing to do with this PR). But if there are multiple versions it might be worth having a "best version" filter, so if grading students and moving through them only the best version is shown. Of course this may mean what about a "best version for section XXX" too. Could be the filters could be a checkbox list vs a link. This way the filter could do other things. Also thinking a dropped students filter that is off by default could also be useful.

Anyways, that is far beyond this PR, just something to consider in the future.

drgrice1:
Perhaps, but definitely not for a hotfix or for this corresponding pull request.

One comment I have regarding that, is that the test with the best score may not be the test with the best score once an instructor's adjustments to the scores are in effect.  So I do not think that would be a good practice.  An instructor should consider all versions when grading a test.  So in my opinion that should NOT be implemented.

Alex-Jordan (approved):

PR #3178, opened 2026-08-29
Fix the test student nav to show all versions of tests. (hotfix of #3177)

drgrice1:
In the rework of the test student nav there was a flaw in design that led to only one version of a test being shown for each student.  It is not the right thing to iterate over the users For which a test version exists.  Instead the actual test versions need to be iterated over.  The users are looked up in a hash in that loop.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
+++ b/lib/WeBWorK/ContentGenerator/GatewayQuiz.pm
@@ -1378,86 +1378,82 @@ sub nav ($c, $args) {
 		# Find all versions of this set that have been taken (excluding those taken by the current user).
 		my @userVersions =
 			$db->listSetVersionsWhere({ user_id => { '!=' => $userID }, set_id => { like => "$setID,v\%" } });
+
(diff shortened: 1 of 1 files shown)


drgrice1:
This definitely needs to be merged before instructors start grading tests this term!

drgrice1:
I want to point out that the code in develop is quite wrong.  For example, assume that you have assigned a test that allows students to take at least 2 attempts, and there are two students that have taken the test. Say that student1 has taken 2 attempts, but student2 has taken only one attempt.  Then the student nav will list student1's first version, but not the second version, and will list student2 with version 2, but will not even list student2's actual only version.  If you click on the link for student2, version2 it will give you ` The selected test (setID,v2) is not a valid test for student2. Requested version (2) of set "setID" is not assigned to user student2.`

drgrice1:
So yes, I really dropped the ball in that rework of the test student nav.

somiaj (approved):
Fixes the issue.

Alex-Jordan (approved):

PR #3179, opened 2026-08-31
Move the version_id into its own field.

drgrice1:
The `version_id` field has been added to the `set_user`, `problem_user`, and `past_answers` tables. If a user set has type "default" or "jitar" then it has version 0.  If a user set has type "gateway" or "proctored_gateway" then it has version 0 if it is the user's template set.  Otherwise it is a versioned set. The same is true for user problems associated with those sets.

When a course database is upgraded the old `<setID>,v<versionID>` in the `set_id` field is separated into the `set_id` without the version, and the `version_id` in its own field. This is done anytime it is detected that the previous database table did not have the `version_id` and it is one of the tables that field has been added to.

This makes database interaction for versioned sets much cleaner and straight forward. Since a `version_id` is an actual database table column, you can use it directly in where and order clauses without hackery.  All of the "groking" for a set version is eliminated.

Note that it is now the case that when a user with the `proctor_quiz_login` permission attempts to enter a proctored test (either their own or to view another user's test when acting as that user), the user will not need to enter the proctor credentials (and will skip the proctor login page entirely).  Also, a user with the `proctor_quiz_grade` permission will not see the proctor login page when grading a test (again for either their own test or when allowed to grade a test when acting as another user).

Note that the `$authen{proctor_module}` option has been removed from the `defaults.config` file.  There is only one proctor authentication module `WeBWorK::Authen::Proctor`, and it is now hardcoded.  No need for the option.

Other than the change to the proctor login noted above, everything else should work the same as before.

### DIFF
--- a/bin/dump-past-answers.pl
+++ b/bin/dump-past-answers.pl
@@ -159,7 +159,7 @@ sub write_past_answers_csv {
 					my $setID    = $_->set_id;
 					my @versions = $db->listSetVersions($userID, $setID);
 					for my $version (@versions) {
-						push(@sets, $db->getUserSet($userID, "$setID,v$version"));
+						push(@sets, $db->getSetVersion($userID, $setID, $version));
 					}
 				} else {
 					push(@sets, $_);
--- a/conf/defaults.config
+++ b/conf/defaults.config
@@ -680,10 +680,6 @@ $modelCoursesForCopy = ["modelCourse"];
 # responds affirmatively will be used.
 $authen{user_module} = 'WeBWorK::Authen::Basic_TheLastOption';
 
-# Select the authentication module to use for proctor logins.
-# A string or a hash is accepted, as above.
-$authen{proctor_module} = "WeBWorK::Authen::Proctor";
-
 # List of authentication modules that may be used to enter the admin course.
 # This is used instead of $authen{user_module} when logging into the admin course.
 $authen{admin_module} = ['WeBWorK::Authen::Basic_TheLastOption'];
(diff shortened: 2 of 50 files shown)


drgrice1:
Note that in testing this you should create an archive of the course you test with and revert back to that archive after switching back to the develop branch.  Once the course is upgraded, there is not downgrade!

Note also that the admin course will work without upgrading (since it never uses versioned sets or problems).

drgrice1:
Note that this will conflict with <s>#3175 and</s> #3177, but already includes fixes for those issues.  I will fix this pull request if those are merged first.  If this is merged first, then those pull requests will be closed and not merged.

somiaj:
I tested this out, and everything seemed to work in my tests: Creating new test versions, hard copy of different versions, grades page, stats, problem grader, student progress.

Do you think we should get this in now, or wait until maybe closer to the release next year for switching between development and main branches and testing on our development servers, since this will be a major changes to all course databases.

drgrice1:
I think that getting this in earlier will lead to better testing of it before the next release.  I recommend that those of use that have dev servers keep classes around with both versions of the database for switching back to main or branches with this not yet merged in.

somiaj (approved):

pstaabp (approved):
Tested it with a couple of different tests including some problem grading.

somiaj:
Going to merge. Be aware that you'll need archives of old courses if you want to be able to test stuff on main after this, or don't upgrade all your courses on your development server.

PR #3181, opened 2026-08-31
Fix LMS roster synchronization for large classes.

drgrice1:
The names and role provisioning services specification states that a platform does not need to return all users in a single request to the names and roles URL, and apparently Canvas (and probably others) does not return all of them at once if there are enough of them (50 seems to be the limit that Canvas uses in my testing).  If the first request contains a `rel="next"` URL in its `link` header, that means that there are more users to be fetched.  So the LMS roster synchronization handler needs to follow that returned link as many times as it is returned from the previous request until all users are obtained.

Since the current code just stops after the first request, and for a large class does not have all users, this results in the remaining users being dropped from the webwork course.

I tested this with my Canvas instance with 300 users in a course. That took about 25 seconds for the initial roster import.  So this also reveals that this really should be a job queue process, and not in the current request.  Although, that is not done at this point, and is left for another pull request and the next release. However, if we feel this is pressing enough, it can be done now as a hotfix.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
+++ b/lib/WeBWorK/ContentGenerator/Instructor/UserList.pm
@@ -551,245 +551,260 @@ async sub lms_roster_sync_handler ($c) {
 		|| $ce->{LTIVersion} ne 'v1p3'
 		|| !$ce->{LTI}{v1p3}{preferred_source_of_username};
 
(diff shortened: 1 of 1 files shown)


somiaj (approved):
Here is approval to get the hotfix in. I can't fully test.

Alex-Jordan (approved):

PR #3186, opened 2026-09-01
Move LMS roster synchronization to the job queue. (Hotfix of #3183)

drgrice1:
The roster synchronization can take quite some time and be somewhat resource intensive if the course in the LMS happens to be large. There is also no way to determine how long it will take before starting the task.  So this moves that task to the job queue.  When the "Synchronize LMS Roster" form on the Accounts Manager page is submitted, it will quickly respond with "Roster synchronization queued". The job status can then be seen on the Job Manager page.  Once completed, the instructor can see the changes on the Accounts Manager page. Note that the job queue info button shows a complete report stating every user that is added, updated, or dropped. There is a summary at the end of the report.


Alex-Jordan:
I will be able to test this with a production setup that uses Canvas late tonight.

somiaj (dismissed):

Alex-Jordan (dismissed):

Alex-Jordan (dismissed):

Alex-Jordan:
Everything worked in the testing I did with a production Canvas server. If @somiaj can approve this again and merge, we are caught up on hotfixes.

drgrice1:
The review "dismissals" is a bit annoying.  I don't understand why it even dismissed @Alex-Jordan's last review already. Nothing changed as far as I can tell after that review was granted.  I may need to look at the branch settings again.

drgrice1:
I turned of the review dismissing rule.  So now, once a review is given, it will stay.

somiaj:
Assuming there is no way to tell the difference between a rebase and an actual change in the commit.

somiaj:
Grr, I missclikced and closed it, now @Alex-Jordan needs to approve again.

somiaj (approved):

drgrice1:
> Assuming there is no way to tell the difference between a rebase and an actual change in the commit.

As far as I can tell there is not.

Alex-Jordan (approved):

PR #3188, opened 2026-09-01
Move the RPC render problem iframe theme chooser JavaScript to its own file.

drgrice1:
This inline script was added in #3145 to support PG dark mode, and most likely @taniwallach will want this not to be inline with the `Content-Security-Policy` headers he is using.

### DIFF
--- a/htdocs/js/RenderProblem/iframe-child.js
+++ b/htdocs/js/RenderProblem/iframe-child.js
@@ -1,0 +1,7 @@
+(() => {
+	window.parent.postMessage('render-iframe-ready');
+	window.addEventListener('message', (event) => {
+		if (!event.data || !event.data.theme) return;
+		document.documentElement.dataset.bsTheme = event.data.theme;
+	});
+})();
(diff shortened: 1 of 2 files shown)


somiaj:
This isn't an issue with this PR as it also affects develop, but something I noticed. When I have webwork set to "auto" and my browser set to "dark", and go to the library browser, and render a collection of problems, they all render in light mode.

I am able to change it to dark mode or render in dark mode if I choose dark mode in the webwork chooser.

somiaj (approved):
Approving the moving of the javascript.

drgrice1:
Actually I forgot that I had noticed issues when the selector is explicitly set to "auto" and the browser in dark mode earlier, and was going to look into it before the PG dark mode was merged.

Alex-Jordan (approved):

somiaj:
Is there a way to turn off the requirement to rebase branches before merging them, or are we going to have to slow down merging of PRs and can only get one at a time until the branch is rebased? I know I could update the branch, but that would add another merge commit.

drgrice1:
I turned that off. I think that is a new setting that is on by default. I don't think I turned it on.

Issue #3189, opened 2026-09-02
Brightspace LTI makes large keys, can fix by changing Course_settings from TEXT to MEDIUMTEXT


josephmaher:
When using the LTI integration with Brightspace, Brightspace seems to make very large keys, in some cases bigger than the 64K limit for a TEXT field.  Changing these fields to MEDIUMTEXT fixes this, be great if you could do this for future versions of webwork2!

ALTER TABLE `Course_setting` MODIFY `value` MEDIUMTEXT;

Here's the error:

WeBWorK error
An error occurred while processing your request.
 
For help, please send mail to this site's webmaster, including all of the following information as well as what what you were doing when the error occurred.
 
Error record identifier
XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX::XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
Error messages
DBD::MariaDB::st execute failed: Data too long for column 'value' at row 1 at /opt/webwork/webwork2/lib/WeBWorK/DB/Database.pm line 137.
 
Context
132: 		if ($errmsg =~ /Unknown column/) {
133: 			warn 'It looks like the database is missing a column. You may need to upgrade your course tables. '
134: 				. 'If this is the admin course then you will need to upgrade the '
135: 				. 'admin tables using the upgrade_admin_db.pl script.';
136: 		}
137: 		die $errmsg;
138: 	}
139: 
140: 	return;
141: }
142: 
Call stack
in Mojolicious::_die called at line 137 of /opt/webwork/webwork2/lib/WeBWorK/DB/Database.pm
in WeBWorK::DB::Database::handle_error called at line 614 of /opt/webwork/webwork2/lib/WeBWorK/DB/Schema/NewSQL/Std.pm
in WeBWorK::DB::Schema::NewSQL::Std::update_where called at line 737 of /opt/webwork/webwork2/lib/WeBWorK/DB.pm
in WeBWorK::DB::setSettingValue called at line 327 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/LTIAdvantage.pm
in WeBWorK::ContentGenerator::LTIAdvantage::get_lms_public_keyset called at line 382 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/LTIAdvantage.pm
[...]

PR #3190, opened 2026-09-02
Cache only the LTI 1.3 signing key actually needed from the LMS public keyset.

drgrice1:
Only one key is ever actually needed to verify a given incoming JWT. Fetch the full keyset from the LMS when there's no cached key or the currently cached key does not work (presumably because it has expired), but don't cache it yet. Once the JWT has been decoded, pull the key from the keyset that is actually used to decode that JWT, and in the case that the keyset was just retrieved from the LMS, then cache just that one key.

D2L Brightspace can serve a large number of keys in its keyset. It most likely uses a different public/private key for each external tool, and it seems that some D2L instances are even shared by multiple institutions in which case each institution has its own set of public/private keys in the keyset. When there are enough keys in the keyset its size can exceed the 64KB limit of the `TEXT` value field in the settings table. So caching only the needed key keeps that from happening.

This should fix issue #3189.  This was also reported in https://forums.openwebwork.org/mod/forum/discuss.php?d=8834 (although I suspect those are really the same D2L instance as in the issue) and https://forums.openwebwork.org/mod/forum/discuss.php?d=8832.

### DIFF
--- a/lib/WeBWorK/ContentGenerator/LTIAdvantage.pm
+++ b/lib/WeBWorK/ContentGenerator/LTIAdvantage.pm
@@ -3,7 +3,7 @@ use Mojo::Base 'WeBWorK::ContentGenerator', -signatures, -async_await;
 
 use Mojo::UserAgent;
 use Mojo::URL;
-use Mojo::JSON           qw(decode_json);
+use Mojo::JSON           qw(decode_json encode_json);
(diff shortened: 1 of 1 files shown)


dlglin (approved):
See my notes on #3191.

somiaj (approved):

PR #3192, opened 2026-09-02
Add CSRF protection with the `Sec-Fetch-Site` header to API calls.

drgrice1:
Any API request verifies that the request is from the same origin by first checking that the `Sec-Fetch-Site` header is `same-origin`. Since this has only been widely supported since 2023, a fallback check that the `Origin` or `Referer` header is same origin is used if the `Sec-Fetch-Site` header is not present. Note that those headers are set by the browser and none of them can be changed by JavaScript. Note that there are some really old browsers that do not set any of those headers, and API requests will simply fail for those browsers.  However, those are browsers that we will just have to say we do not support anymore. Those browsers are quite old (pre 2015), so generally this is not a concern.

The point of this is that a user that has properly logged into webwork in the browser may be tricked into doing something in another tab or window by an attacker that would cause a request to be sent to the API. That request would automatically have the valid WeBWorK session cookie attached to it by the browser, and so the request would succeed and could do whatever the user has permission to do via the API. However, that request would not have the `Sec-Fetch-Site` header set to `same-origin` and the `Origin` and `Referer` headers would also not have the same origin.

This was one of the vulnerabilities that @Alex-Jordan found with Claude before.

This approach also is easy to work with if using the API via a script. Scripts can add the `Sec-Fetch-Site` header, and can set it to whatever they want it to be.  Obviously they just need to set it to `same-origin`.

Note that this does not apply to the `render_rpc` endpoint. It probably should, but that would cause issues for PreTeXt (although this could be worked around in much the same way that cookie disabling is done if the `allow_unsecured_rpc` option is set). At this point I think this is also not necessary as this endpoint generally can't do anything destructive. It just renders problems.

This is an alternate approach to #3169 that does not modify the `webworkConfig` variable to pacify @taniwallach.


taniwallach:
@drgrice1 Thanks - Yes I certainly prefer this approach to the one in https://github.com/openwebwork/webwork2/pull/3169.

pstaabp (approved):
Didn't test, but appears to be an improvement.

somiaj (approved):
Things seem to still work on my dev server, tested the library browser, the problem grader to change scores, and the problem editor. I also tested my use of the rpc end point for my embedded problems, and all still works.

somiaj:
I'm going to merge this. Would it be worth adding CSRF tokens to the various instructor forms as well?

drgrice1:
What do you mean by adding CSRF tokens to instructor forms?  Do you really mean all forms aside from the API requests, since those are all the same in the way that they work?

As long as `$CookieSameSite` is set to "Lax" that isn't needed.  However, if `$CookieSameSite` is "None" then the form posts are exposed. Although we already document the "None" setting as potentially allowing for certain types of cross-site attacks.

somiaj:
Yea, I meant is it worth also adding csrf_tokens to all of the POST forms used for instructors (though I guess all forums would work too, I was just thinking about the instructor pages).

I didn't realize `$CookieSameSite='Lax'` was another approach to prevent JavaScript from sending cross-site POST requests using the session cookie in the browser. I guess there are some edge cases where using both could help, at least in a quick search on the topic. So seems the answer is it probably isn't needed as long as people don't use "None".

drgrice1:
Looking into this, there is also a complication with LTI 1.1.  Any login request is a cross-site request, and those requests can go to any page.  So you have to take care to allow those and only ensure requests using the cookie are same origin.

PR #3193, opened 2026-09-02
Add a workflow for checking spelling.

drgrice1:
A few more typos in the code were fixed.  Note that the typo in the `htdocs/library-tree.json` was fixed. We don't actually use that file anymore anyway.

The following code changes were made so that the words involved do not need to be blanket ignored.  They weren't good variable names anyway.

* The variable `$pn` in `templates/ContentGenerator/GatewayQuiz.html.ep` was renamed to `$problemNumber`.
* The variable `%visibles` in `lib/WeBWorK/Utils/FilterRecords.pm` was renamed to `%visibleSets`.
* The wording `select alls` was changed to `select all checkboxes` in the comments in `htdocs/js/SelectAll/selectall.js`.

crate-ci/typos is configured via the .typos.toml file with tightly scoped extend-identifiers/extend-ignore-re entries for the remaining legitimate non-dictionary terms (crypted, DNE, ANDed, LANGuage, etc.) so the checker stays useful for catching real future typos.

This was suggested by @Alex-Jordan, and I thought I would experiment with it.  I had Claude help with creating the configuration.  If we don't want to go with this the typos in the `doc/AccessibilityGuide.md` should be fixed.


somiaj (approved):
I am okay with a spell checking work flow.

Alex-Jordan (approved):

Issue #3194, opened 2026-09-03
Error with SingleProblemGrader when acting as student viewing a problem with an error.


somiaj:
If for some reason a problem doesn't process correctly due to an error in the code, and you are viewing it in a homework set while acting as a student (for me it was due to a random seed breaking a problem), the new single problem grader errors out not giving the actual error.

Any bad problem should cause this, but here is a MWE (which causes a processing error):

```
DOCUMENT();
ENDDOCUMENT();
```

Error given by the SingleProblemGrader:

```
Can't use an undefined value as an ARRAY reference at template HTML/SingleProblemGrader/grader.html.ep line 20.

    15: 	<div class="problem-grader-table">
    16: 		% my $currentScore    = 0;
    17: 		% my $rawCurrentScore = 0;
    18: 		%
    19: 		% # Subscores for each answer in the problem.
    20: 		% if (@{ $grader->{pg}{flags}{ANSWER_ENTRY_ORDER} } > 1) {
    21: 			% # Determine the scores and weights for each part of the problem.
    22: 			% my $total = 0;
    23: 			% my (@scores, @weights);
    24: 			% for my $ans_id (@{ $grader->{pg}{flags}{ANSWER_ENTRY_ORDER} }) {
    25: 				% push(@scores,  $grader->{pg}{answers}{$ans_id}{score} * 100);
```

PR #3195, opened 2026-09-03
Make sure to account for an undefined `ANSWER_ENTRY_ORDER` in the single problem grader.

drgrice1:
Previously the problem grader could only be made visible when the problem successfully rendered.  Now that it is always visible, a little more care is needed.  The `ANSWER_ENTRY_ORDER` array in the PG flags will not be defined if the problem does not successfully render.

This fixes issue #3194.

### DIFF
--- a/templates/HTML/SingleProblemGrader/grader.html.ep
+++ b/templates/HTML/SingleProblemGrader/grader.html.ep
@@ -17,7 +17,7 @@
 		% my $rawCurrentScore = 0;
 		%
 		% # Subscores for each answer in the problem.
-		% if (@{ $grader->{pg}{flags}{ANSWER_ENTRY_ORDER} } > 1) {
+		% if (@{ $grader->{pg}{flags}{ANSWER_ENTRY_ORDER} // [] } > 1) {
 			% # Determine the scores and weights for each part of the problem.
 			% my $total = 0;
 			% my (@scores, @weights);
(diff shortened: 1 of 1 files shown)


drgrice1:
It seems to me that the fact that the problem
```
DOCUMENT();
ENDDOCUMENT();
```
does not render reveals a fundamental flaw in PG.

The `PG.pl` macro certainly should not call any method that is defined in another macro.  Either the `MODES` method should not be called in `PG.pl`, or the `MODES` method should be defined in `PG.pl`.

somiaj (approved):
Fixes the issue.

somiaj:
When testing this out, I noticed that some users the problem grader has a comment field and others don't, what is the logic for if it is possible for the instructor to add a comment with the single problem grader?

somiaj:
> does not render reveals a fundamental flaw in PG.

Note, @dlglin pointed out a few other issues with PGbasicmacros.pl requiring stuff from PGauxiliaryFunctions.pl, and basically PG has issues and isn't well tested in the case PGstandard.pl is not loaded. Either we start just loading PGstandard.pl by default, or we fix those issues is my thought.

drgrice1:
The comments are saved in the past answer table. So in order for there to be a place to save the comment, there must be a past answer. So if the student has not submitted an answer, there is no comment field.

somiaj:
Using @Alex-Jordan approval on #3196 as an approval on this too, and merging.

PR #3196, opened 2026-09-03
Make sure to account for an undefined `ANSWER_ENTRY_ORDER` in the single problem grader. (Hotfix of #3195)

drgrice1:
Previously the problem grader could only be made visible when the problem successfully rendered.  Now that it is always visible, a little more care is needed.  The `ANSWER_ENTRY_ORDER` array in the PG flags will not be defined if the problem does not successfully render.

This fixes issue #3194.


somiaj (approved):

Alex-Jordan (approved):

Alex-Jordan:
I can't merge this and #3195 without rebasing first. Can you check on things? I don't want to cause any git headaches.

PR #3197, opened 2026-09-04
Update the `prettier-plugin-mojolicious-html-template` package.

drgrice1:
I made some improvements to the package to fix some formatting issues seen in the template files, and published the updated version.  The formatting issues can be seen in the changes to the template files.  I think they are self explanatory.

### DIFF
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -21,7 +21,7 @@
 				"sortablejs": "^1.15.7"
 			},
 			"devDependencies": {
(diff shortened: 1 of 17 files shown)


drgrice1:
Also, it is now an error to not have `perltidy` installed.  Previously it would format the HTML and ignore Perl code in that case.  However, that would result in formatting that does not match the workflow.  So that is not allowed anymore.

somiaj (approved):

Alex-Jordan (approved):

PR #3199, opened 2026-09-06
Get all sets when performing LTI date synchronization.

drgrice1:
After the issue with the names and role service URL and realizing that the `next` link needs to be utilized to get all names and roles from the URL, I figured I had better check the spec for the `lineitems` URL used for set date synchronization to see if it works the same, and in fact, it does.  See https://www.imsglobal.org/spec/lti-ags/v2p0#container-request-filters. So this makes sure that all sets are obtained from the `lineitems` URL in the same way.

In testing this with Canvas, it turns out that Canvas only sends 10 sets at once.  So if a course has more than 10 sets, then only 10 sets will be retrieved from the lineitems URL in the first request.  Note that date synchronization will not necessarily fail though, as in the next phase of the job queue task, requests are sent to the individual lineitem URLs for the sets if the data was not received from the lineitems URL for the set assuming that the lineitem URL for the set has been obtained (webwork will have it if content item selection was used, or if a user has entered the set directly from the LMS).  However, this results in extra network communication that is unnecessary.

### DIFF
--- a/lib/Mojolicious/WeBWorK/Tasks/LTISetDateSync.pm
+++ b/lib/Mojolicious/WeBWorK/Tasks/LTISetDateSync.pm
@@ -52,25 +52,37 @@ async sub synchronizeSetDates ($job, $setIDs, $syncToLMS) {
 
 	my $ua = Mojo::UserAgent->new;
 
-	my $lineitemsRequest =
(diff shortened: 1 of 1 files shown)


somiaj (approved):
I can't test LTI stuff, but this fix is similar as before and needs to get in.

Alex-Jordan (approved):

PR #3201, opened 2026-09-06
Fix PG color scheme selection with explicit 'Auto' mode set.

drgrice1:
When the color scheme selector at the top of the page is used and "Auto" is selected, and the browser is set to dark mode, problems are rendered incorrectly in light mode.  This is because the `renderproblem.js` and 'pgproblemeditor.js' JavaScript send the "auto" setting from local storage and the iframe child JavaScript sets the `data-bs-theme` attribute of the iframe page to that.  Since that value means nothing in terms of color scheme selection for bootstrap, the page renders in the default bootstrap light mode. So if the theme in local storage is 'auto', detect the browser setting and use that to select 'dark' or 'light' to send to the iframe child.

### DIFF
--- a/htdocs/js/PGProblemEditor/pgproblemeditor.js
+++ b/htdocs/js/PGProblemEditor/pgproblemeditor.js
@@ -595,10 +595,14 @@
 
 	window.addEventListener('message', (event) => {
 		if (event.data !== 'render-iframe-ready' || iframe.contentWindow !== event.source) return;
+		const storedTheme = localStorage.getItem('WW.color-scheme');
 		iframe.contentWindow.postMessage({
 			theme:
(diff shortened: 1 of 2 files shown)


somiaj:
On my site, I just use the `data-bs-theme` attribute off the html tag value directly (that should always be set correctly?), vs go through that logic each time. Would that work here and simplify things?

drgrice1:
That might work, but I am not sure that it simplifies the logic any.  Technically you would still need to validate that the attribute you get is defined, and this logic is not actually complicated.

somiaj (approved):
Fixes the issue.

Alex-Jordan (approved):

PR #3202, opened 2026-09-07
Add links to test problems on Grades page.

somiaj:
Add links to the problems for tests on the Grades page if a student can access the test version. This also removes any links to a proctored test version on the grades page. This page is links for students, and they cannot access a proctored test without a proctor password.

In testing this I did notice some issues with linking to test problems directly. First as an instructor, the student navigation bar at the top covers the link anchors when linking to an individual problem on a test with more than one problem with page. Second when linking to a problem directly (such as from the grades page), the div that holds the actual problem isn't positioned correctly until after all the MathJax is fully rendered and the content is made visible, causing the div to move. Those aren't really related to this PR, but become a bit more obvious when testing this PR, and I think this is now the only case that we link to problems directly from a different page.


lib/WeBWorK/ContentGenerator/Grades.pm
drgrice1: This uses a very sloppy and hackish string manipulation approach to obtaining a proper URL for the link.  This is changed in #3179 and will conflict with that pull request.  What is done in that pull request a much better way of doing this.

drgrice1 (commented):
You are mistaken that this is the only place that links directly to a problem in a test from a different page.  This is also done when editing a set version for a user on the set detail page and when grading via the manual problem grader.

lib/WeBWorK/ContentGenerator/Grades.pm
somiaj: This was inherited from the old grades page, and I'll fix any conflicts once #3179 is merged.

somiaj:
Ahh I didn't realize that there was direct links elsewhere I could look at. I switched from `newPage` to `currentPage` to match what is done there. The issues I describe still happen with those links and wasn't something I was looking to fix, just commenting on the behavior.

lib/WeBWorK/ContentGenerator/Grades.pm
somiaj: Though does seem there is some dead code here that in my changes I didn't notice needed to be cleaned up.

somiaj: This was pointless code so I removed it here, though I don't think I can fully update this to not use a similar hack below until #3179 is merged, at which time I'll fix any conflicts.

somiaj:
A few more tweaks to link logic, removing all the link hacks. This will have to be adjusted as other PRs it conflicts with are merged.

somiaj:
@drgrice1 Claude suggested to add the window scroll to the position computations will make it so the overlay correctly covers the problem content when linking to an anchor. It seems to work in my testing, but instead of making a PR, I will just add the diff here to let you look at it first. We can probably add this to one of the PRs we are working on if you think it is reasonable, unless you think this should be a hotfix.

```diff
diff --git a/htdocs/js/MathJaxConfig/mathjax-config.js b/htdocs/js/MathJaxConfig/mathjax-config.js
index 31de3d420..0f67bd580 100644
--- a/htdocs/js/MathJaxConfig/mathjax-config.js
+++ b/htdocs/js/MathJaxConfig/mathjax-config.js
@@ -161,8 +161,8 @@ if (!window.MathJax) {
                loaderOverlay.classList.add('problem-content');
                const bodyRectangle = problemContent.getBoundingClientRect();
                loaderOverlay.style.position = 'absolute';
-               loaderOverlay.style.top = `${bodyRectangle.y}px`;
-               loaderOverlay.style.left = `${bodyRectangle.x}px`;
+               loaderOverlay.style.top = `${bodyRectangle.y + window.scrollY}px`;
+               loaderOverlay.style.left = `${bodyRectangle.x + window.scrollX}px`;
                loaderOverlay.style.width = `${bodyRectangle.width}px`;
                loaderOverlay.style.height = `${bodyRectangle.height}px`;
                loaderOverlay.style.overflow = 'clip';
@@ -177,8 +177,8 @@ if (!window.MathJax) {
                problemContent.after(loaderOverlay);
                const resizeObserver = new ResizeObserver(() => {
                        const bodyRectangle = problemContent.getBoundingClientRect();
-                       loaderOverlay.style.top = `${bodyRectangle.top}px`;
-                       loaderOverlay.style.left = `${bodyRectangle.left}px`;
+                       loaderOverlay.style.top = `${bodyRectangle.top + window.scrollY}px`;
+                       loaderOverlay.style.left = `${bodyRectangle.left + window.scrollX}px`;
                        loaderOverlay.style.width = `${bodyRectangle.width}px`;
                        loaderOverlay.style.height = `${bodyRectangle.height}px`;
                });
```

drgrice1:
That seems fine.  Go ahead and add it to this pull request.  I don't think that it is crucial as a hotfix.  This does not affect the usage for students, and only happens with the instructor links that are not used that often.

somiaj:
@drgrice1 looking at the suggested change, I just now noticed, why does the `ResizeObserver` update use `bodyRectangle.top` while the initial sizing using `bodyRectangle.y`?

somiaj:
Updated this to fix the overlay issue, and removed modification of the gateway css to not conflict with the changes in #3204.

drgrice1:
> why does the `ResizeObserver` update use `bodyRectangle.top` while the initial sizing using `bodyRectangle.y`?

`top` and `y` and `left` and `x` are actually equal (except in the case that the `problemContent` has negative height or width which won't happen here).  I remember that I was using `x` and `y` originally in both places, but switched to using `top` and `left` to match the style attributes which do not have the `x` and `y` alternates.  I must have forgot to make the change in both places which should have been done for consistency.

somiaj:
Updated to use `top`/`left` in both places.

drgrice1 (dismissed):
I think this looks good.

pstaabp (approved):
Looks good. 

I'm guessing this is going to conflict with #3179.  If so, what order to bring them in?

somiaj:
I say move the bigger one in first, and I'll fix the conflicts.

somiaj:
Rebased and fixed the conflicts. The removal of `$setTemplateID` in #3179 was a mistake. This broke showing scores and links to tests on the Grades page. This is because the grades page at this point still uses the old version names for the setID. I removed that change and updated the use of this variable to ensure links work correctly on the Grades page.

@drgrice1 I removed your approval and you need to look this over in relation to #3179.

lib/WeBWorK/ContentGenerator/Grades.pm
```diff
@@ -248,30 +253,26 @@ sub displayStudentGrades ($c) {
...
+			# Only show link if the problems can be seen, and it is not a proctored test
+			# since students cannot directly access them until after it is graded, at
+			# which time it reverts to a normal test.
+			$item->{link} = ''
+				if $set->assignment_type eq 'proctored_gateway'
+				|| (
+					defined $set->hide_work
+					&& ($set->hide_work eq 'Y'
+						|| $set->hide_work eq 'BeforeAnswerDate' && time < $set->answer_date)
+				);
```
drgrice1, line 265: Perhaps this should show the link for a problem in a proctored test in the case that the user viewing the grades has the `proctor_quiz_login` permission.  In that case, the user has access to the test anyway (and no longer even needs to enter proctor authentication to do so).

drgrice1 (approved):
I think this looks good.

lib/WeBWorK/ContentGenerator/Grades.pm
somiaj: The overall design of the grades page in my redesign was to only show what a student would see and not have lots of conditions to show users acting as a student additional stuff. I could do that here, but there are lots of other places information is shown based on only what a student can see (for instance there is no condition for when the user has the `view_hidden_work` permission here).

If someone wants to go add all those additional conditions and permissions, they could, but my original design left that out. The instructor's student progress page can be used to gain access to all that information if needed.

drgrice1: Okay. That is fine.

PR #3203, opened 2026-09-08
Adjust the `scroll-margin-top` for anchors on pages with the student nav.

drgrice1:
This makes the `scroll-margin-top` setting for elements with an `id` attribute (used as anchors via a URL fragment) account for the height of the student/problem nav if it is present on the page.  Basically, the value of the `--sticky-nav-height` css variable is always added to the `scroll-margin-top` for these elements.  If the sticky student/problem nav is present on the page, then JavaScript sets the value of that css variable to the height of the sticky nav plus `1rem`.  Thus when a URL fragment is used the id element is not obscured behind the sticky nav.

One place where this is prominent is when viewing a problem in a test from either the set detail page for the test when editing a set version for a user, or from the manual problem grader when grading a problem in a test.

@somiaj mentioned this in #3202, and I have observed this for a while now. Note that @somiaj adds `outline: none` to these test div elements with an `id` which also improves the initial appearance when the page loads.

### DIFF
--- a/htdocs/js/System/system.js
+++ b/htdocs/js/System/system.js
@@ -93,6 +93,19 @@
 		});
 	}
 
+	// Account for the height of the sticky nav (if present) in scroll-margin-top, so
+	// that utilizing a URL fragment doesn't obscure anchors behind the sticky nav.
+	const stickyNav = document.querySelector('.sticky-nav');
+	if (stickyNav) {
+		const updateStickyNavHeight = () =>
+			document.documentElement.style.setProperty(
+				'--sticky-nav-height',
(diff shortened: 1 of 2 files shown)


drgrice1:
I realized that there is another case where this improves things.  That is when a test has more than one problem per page, and a problem number jump link is used.  That case also scrolls down to far, and that is fixed with this pull request.

somiaj (approved):
This fixes the scroll issue for problems in test, and probably elsewhere anchor links are used and the navigation bar is shown. Since this affects problem scrolling in tests with more than one problem and other links that exist to particular problems, might be worth a hotfix.

pstaabp (approved):

PR #3204, opened 2026-09-08
Rework the test problem header.

drgrice1:
Currently the message that reports if answers are recorded or not is the first thing in a problem in a test, and that also has the `id` of the problem.  That `id` element is what is used as the anchor when a problem jump link is used or if entering the test directly to a problem from a link else where.  Furthermore, in most of those cases when those fragment URLs are used, that recorded answer div is empty. This results in odd behavior such as that empty div being outlined.

So move that div down after the actual problem header and wrap the problem header (and sub header) in a div that now contains the `id` of the problem. This is much more intuitive when one of these fragment URLs is used.

Note that this needs #3203 to work better, and will conflict with #3202 (but the changes in #3202 that this conflicts with should probably be removed with this change).


templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -617,28 +617,29 @@
...
-							<%= "($problemValue $points)" %>
-						% }
-						%
-						% # This uses the permission level and user id of the user assigned to the set.
-						% if ($authz->hasPermissions($effectiveUserID, 'print_path_to_problem'))
-						% {
-							<%= $problems->[ $probOrder->[$i] ]->source_file %>
-						% }
-					</span>
+					<div><%= $recordMessage %></div>
```
somiaj: looking above, it appears that `$recordMessage` is already inside a `div`, so I don't think it needs to be wrapped in a `div` here.

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -617,28 +617,29 @@
...
+							% }
+							%
+							% # This uses the permission level and user id of the user assigned to the set.
+							% if ($authz->hasPermissions($effectiveUserID, 'print_path_to_problem'))
+							% {
+								<%= $problems->[ $probOrder->[$i] ]->source_file %>
+							% }
+						</span>
+					</div>
 					% # Show the jump to anchor.
```
somiaj: This comment feels off, the anchor is now above the comment, and this is now talking about the `$recordMessage`.

templates/ContentGenerator/GatewayQuiz.html.ep
drgrice1: I removed the `div` around the `h2` problem header and the `problem-sub-header`.  However, with that, it turns out the `div` aroudn the `$recordMessage` is still needed.  The `$recordMessage` div has `d-inline-block` (so display `inline-block`) set, and that means that the record message alert will be on the same line as the problem header.  The `inline-block` display is needed to prevent the `div` from being made the full width of the problem.

somiaj: Maybe only show this div if `$recordMessage` is not the empty string to avoid putting an empty div into the DOM?

templates/ContentGenerator/GatewayQuiz.html.ep
```diff
@@ -617,10 +617,9 @@
 				% }
 				%
 				<div class="gwProblem">
-					% # Show the jump to anchor.
-					<div id="<%= "prob$i" %>" tabindex="-1"><%= $recordMessage %></div>
-					% # Output the problem header.
-					<h2 class="gw-problem-number"><%= maketext('Problem [_1].', $i + 1) %></h2>
+					<h2 id="<%= "prob$i" %>" class="gw-problem-number" tabindex="-1">
```
somiaj, line 620: Do we want to avoid tabbing to the header? I think that might be useful to move between problem headers and isn't needed now that the id/anchor isn't a separate div?

templates/ContentGenerator/GatewayQuiz.html.ep
drgrice1: It really doesn't matter.  An empty div is not allocated any space.  I suppose I will add that though.

templates/ContentGenerator/GatewayQuiz.html.ep
drgrice1: `tabindex="-1"` does not allow tabbing to the header.  It only allows JavaScript to focus the header, and that is still needed by the jumplink JavaScript code.

somiaj (approved):

somiaj:
I did notice that when this is combined with #3202 (or other ways to link to directly to a problem header), the "Problem X" header gets an outline around it when linking directly to the problem from a different page. Is this behavior we want?

drgrice1:
Yeah, you could add back the outline none css if you want. That would fix the issue.

somiaj:
Yea, I was just curious what behavior we would prefer. I would suggest adding it to this PR, since this is the header rework. To me I don't like it, but it is less of an issue for me since it isn't an empty div. So I could go either way.

drgrice1:
I added `outline: none` to the `gw-problem-number` class.

pstaabp (approved):

PR #3208, opened 2026-09-09
Fix how parserRadioMultiAnswer.pl answers are counted as unanswered when grading a test.

drgrice1:
Currently the javascript considers all of the inputs.  However, the inputs in the unchecked parts of a `RadioMultiAnswer` will not have an answer, and should not affect whether a question is determined to be answered or not. So the part sub answers for the unchecked radio buttons are ignored, and only the checked radio button answers considered.

To be frank I think that this unreliable JavaScript approach of attempting to determine unanswered questions is going to have to be ditched.  JavaScript is never going to be able to reliably make this determination.


somiaj (approved):
This method may not be accurate, but is a good guess. The bigger issue is students blindly hitting "Okay" without reading, which we cannot fix.

pstaabp (approved):

PR #3211, opened 2026-09-11
Fix a couple of LTI issues. (hotfix of #3210)

drgrice1:
First, for both versions of LTI, there are several columns in the user table that should never be modified in the `maybe_update_user` methods. Those are the `displayMode`, `useMathQuill`, `useMathView`, and most importantly the `accommodation_time_factor` columns.  However, since the `$tempUser` object is created without those ever initialized, these values get set to defaults in the database when an update does occur. The worst case here is that the `accommodation_time_factor` gets set back to 1 in the case that the instructor changed it.

Second, when grades are passed back for a test and yet there are no test versions taken by the student, there was an error in the `WeBWorK::Authen::LTI::GradePassback::earliestGatewayDate` method that makes the return value incorrect, and as a result grades end up being passed back when they should not.  Basically, the method should not only check that the `$setVersions` passed in is an array reference, but also that it is nonempty.


drgrice1:
By the way, see https://forums.openwebwork.org/mod/forum/discuss.php?d=8836 for the first issue.  The second issue is something I have been noticing for the last few weeks since starting to use LTI 1.3 in production.

somiaj (approved):
I haven't noticed this issue, but I don't think any of my users change data often enough that I would have seen it. Code is easily understandable.

pstaabp (approved):

PR #3212, opened 2026-09-14
Add option to show/hide preview button.

somiaj:
The preview button can sometimes provide students with information about the correctness of their answers, and with MathQuill is not needed to preview how their answer is formatted. This new option configures if the "Preview My Answers" button is shown or not.

If the preview button is not shown, then this also changes the behavior of the "enterKey" option.

The `pageChangeHack` and its interaction with the preview button was removed, and instead test page links now submit the form directly and no longer rely on clicking the preview button.


lib/WeBWorK/ConfigValues.pm
```diff
@@ -805,18 +805,31 @@ sub getConfigValues ($ce) {
...
-						. 'instructor pages like the PG Editor and the Library Browser.'
+						. '"Preview My Answers" button, if it is shown, otherwise this acts like "submit". If '
+						. 'this is set to "submit", then the enter key activates the "Submit Answers" button '
+						. 'instead. Or if that button is not present, it will activate the "Check Answers" button. '
+						. 'Or if that button is also not present, it will activate the "Preview My Answers" button. '
+						. 'A third option is "conservative". In this case, the enter key behaves like "preview" when '
+						. 'the "Submit" and "Preview My Answers" buttons are available and there are only finitely '
+						. 'many attempts allowed. Otherwise the enter key behaves like "submit". Note that this is '
+						. 'only affects homework problem pages, not test/quiz pages, and not instructor pages like '
+						. 'the PG Editor and the Library Browser.'
```
drgrice1: There are several issues with this documentation.

The first issue is a problem with the first line as it is in develop.  It is not true that "hitting the enter key on a homework problem page" does anything at all based on this setting.  This setting applies to hitting the enter key when an answer input is focused.  If the page is focused, but no answer input is focused, hitting the enter key does nothing.

The rest is that this documentation as adapted for this change is not entirely correct (or at least not complete).  It states that when the setting is "submit" then the enter key activates "Submit Answers", and if that button is not present then it activates "Check Answers", and if that button is also not present then it activates "Preview My Answers".  However, with this pull request the "Preview My  Answers" button may also not be present.  Yes, it is possible for there to be no submit buttons at all depending on the `check_answers_after_due_date`, `check_answers_after_answer_date`, and the new `$pg{options}{showPreviewButton}` setting and the current state of the set relative to the due date and answer date.  There are other permissions and problem set states that can make this occur as well.

drgrice1 (approved):
I think this is generally good.

drgrice1 (changes requested):
Actually, I take that back.  This is not complete for tests.  There are more than one preview button, and this does not remove all of them.  Furthermore, the JavaScript for the page change links is broken with the preview button removed.  So that will take more work.

somiaj:
I don't think the `pageChangeHack` is needed. Instead of clicking the preview button, just submit the form directly seems to work in my limited testing. I'll look more later. Do you have any suggestions on this approach.

drgrice1 (approved):
I think this is generally good now.  That is aside from the documentation issue for the `pg{options}{enterKey}` variable.

somiaj:
I tried to clean up the description. I feel it is a bit wordy, but there are a lot of different possibilities. Maybe @Alex-Jordan can look it over, as this is their feature.

lib/WeBWorK/ConfigValues.pm
```diff
@@ -802,21 +802,36 @@ sub getConfigValues ($ce) {
...
-					'If this is set to "preview", hitting the enter key on a homework problem page activates the '
-						. '"Preview My Answers" button. If this is set to "submit", then the enter key activates '
-						. 'the "Submit Answers" button instead. Or if that button is not present, it will activate '
-						. 'the "Check Answers" button. Or if that button is also not present, it will activate '
-						. 'the "Preview My Answers" button. A third option is "conservative". In this case, the '
-						. 'enter key behaves like "preview" when the "Submit" button is available and there are '
-						. 'only finitely many attempts allowed. Otherwise the enter key behaves like "submit". '
-						. 'Note that this is only affects homework problem pages, not test/quiz pages, and not '
-						. 'instructor pages like the PG Editor and the Library Browser.'
+					'This affects the action when the enter key is pressed while inside an answer input in a '
```
drgrice1: I think that I would word this first sentence, "This affects the action when the enter key is pressed while an answer input is focused."  That is more accurate, and saying "inside" seems a bit off, particularly when the input is a checkbox or radio button.  Note that this option does apply to those as well.

somiaj: I went with "inside" as I felt "focused" might not be as well understood. I will update it to your suggestion.

drgrice1: I think "focused" should be a rather common terminology for this sort of thing, and it shouldn't cause confusion.

lib/WeBWorK/ConfigValues.pm
```diff
@@ -802,21 +802,36 @@ sub getConfigValues ($ce) {
...
-						. 'the "Submit Answers" button instead. Or if that button is not present, it will activate '
-						. 'the "Check Answers" button. Or if that button is also not present, it will activate '
-						. 'the "Preview My Answers" button. A third option is "conservative". In this case, the '
-						. 'enter key behaves like "preview" when the "Submit" button is available and there are '
-						. 'only finitely many attempts allowed. Otherwise the enter key behaves like "submit". '
-						. 'Note that this is only affects homework problem pages, not test/quiz pages, and not '
-						. 'instructor pages like the PG Editor and the Library Browser.'
+					'This affects the action when the enter key is pressed while an answer input is focused in a '
+						. 'homework problem. There are three possible action buttons, "Preview My Answers", '
+						. '"Submit Answers", and "Check Answers", that maybe available. Each setting configures a '
```
pstaabp: "maybe" should be "may be"

somiaj: Updated.

pstaabp (approved):

PR #3213, opened 2026-09-15
Move all authen module config to localOverrides.conf

dlglin:
This removes the setting of `$authen{user_module}` and `$authen{admin_module}` from the config files for each of the custom authentication methods, and moves the setting of these variables after the includes of the additional config files.

This should help with the perennial issues of people including multiple authentication config files, and having the later one overwrite the authentication modules configured in the earlier one.


somiaj (approved):
Looks good.

drgrice1 (approved):
Looks good.

PR #3214, opened 2026-09-15
Fix calls using $db->listUserProblemsWhere.

somiaj:
With the addition of `version_id` in #3179, This DB call now returns the data in the order `[ user_id, set_id, version_id, problem_id, ... ]`. This changed the order of the list by adding `version_id` between the `set_id` and `problem_id`, so everywhere this call was used, the code needed to be updated due to this change.

In looking at this, I noticed that there are a bunch of database calls, such as `getUserProblems`, `listUserProblems`, `existUserProblems` that do not take `version_id` into account. I'm unsure if those may cause other issues.

I noticed this issue with the homework problem page Previous and Next buttons no longer working.


lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
```diff
@@ -1783,15 +1782,16 @@ sub initialize ($c) {
...
-			# there is an argument for being able to get() all problem versions for all users in one database call.  The
-			# current code may be slow for large classes.
+			# there is an argument for being able to get() all problem versions for all users in one database call.
+			# The current code may be slow for large classes.
 			if ($setRecord->assignment_type !~ /gateway/) {
+				my @userProblemIDs =
+					$forUsers
+					? (map { [ $_, $setID, $problemID ] } @editForUser)
+					: (map { [ $_->[0], $_->[1], $_->[3] ] }
+						$db->listUserProblemsWhere({ set_id => $setID, version_id => 0, problem_id => $problemID }));
```
drgrice1: `version_id => 0` is not needed here.  That is added by default for the `problem_user` table.

drgrice1 (commented):
Good catch.

lib/WeBWorK/ContentGenerator/Instructor/ProblemSetDetail.pm
somiaj: Thanks. Updated.

drgrice1:
Note that there will not be a similar problem with the `getUserProblems`, `listUserProblems`, and `existsUserProblems` calls.  This would only be a problem with the where variants, since those do not go through the `box` method.

drgrice1:
Actually, `getUserProblems` goes through the `box` method, and so this will not happen with that method.  The `listUserProblems` method directly specifies the field (the `problem_id`) it gets, so this also will not happen with that method.  The `existsUserProblem` method is just more specific to begin with, and so this will not happen with that method either.

somiaj:
My comment was more that `getUserProblems` etc could have been sent a set_id which use to contain the version string before the version was seperated, so wasn't sure if this was ever used to get problems for a set version. Since the method doesn't take that into account, I wasn't sure if they might get all problems for all versions, or only version 0. Just something I noticed. I did notice there was some calls for `getProblemVersionsWhere`, so it might not be an issue. Only commented in case it was.

somiaj:
I mostly noticed because `getUserProblems` failed when I sent it the output directly from `listUserProblemsWhere`, so that is why I adjusted it via a map.

drgrice1:
I am going to go ahead and break the two review rule with this.  This is going to get annoying quickly (at least the next/previous problem button issue), and this is only for develop anyway.

drgrice1:
> I mostly noticed because `getUserProblems` failed when I sent it the output directly from `listUserProblemsWhere`, so that is why I adjusted it via a map.

That is because `getUserProblems` expects the `user_id`, `set_id`, and `problem_id` in that order, but `listUserProblemsWhere` returns the key fields of the user_problem table which is now the `user_id`, `set_id`, `version_id`, and `problem_id`.

It should not be the case that `getUserProblem` would have ever been given a `set_id` that contained the version.  The `getProblemVersion` method should have been used for that.

PR #3218, opened 2026-09-22
Fix bug in ProblemGrader when only points are shown.

somiaj:
The hidden field score was incorrectly set (by me) to zero, and not the actual score of the student initially. Which means saving the grades without changing a student's score would result in a grade of zero.

I must admit, I'm unsure why setting the parameter didn't work, but elsewhere in the code this was the method used to undefine the parameter and then set the value inside the input, so I must have overlooked that when implementing this. Could it be how mojolicious deals with parameter values and hidden inputs?


drgrice1:
Mojolicious does not set the values of hidden inputs from the parameters.  This is different than for text inputs and other visible form fields for which Mojolicious does set the values from parameters.

somiaj:
That it was I thought was the case, but wasn't sure. I was just so use to visible inputs, I didn't think it would be different for hidden inputs when I implemented this, and it escaped my testing until it bit me recently.

templates/ContentGenerator/Instructor/ProblemGrader.html.ep
```diff
@@ -240,9 +240,9 @@
 								% my $thisStepSize = $useUserValue ? points_stepsize($problemValue) : $stepSize;
 								<td class="restricted-width-col">
 									% if ($ce->{problemGraderScore} eq 'Point') {
-										% param("$userID.$versionID.score", wwRound(0, $_->{problem}->status * 100));
+										% param("$userID.$versionID.score", undef);
```
drgrice1: You should probably actually delete this line. It certainly isn't needed for the hidden input. So unless setting it is need for something else (and it really shouldn't be), this line does nothing useful.

somiaj: Deleted.

drgrice1 (approved):
Looks good.  Fixes the problem.

